PR #2, opened 2022-10-19
MISC: Fix alias not working in test suite

hydroflame:


Snarling (approved):
Nice, tests pass now.

PR #132, opened 2022-10-19
INFILTRATION: Fix SlashGame scaling

### LINKED ISSUE #124: Slash minigame timing off - development
[steveturri]
When infiltrating, the slash minigame seems to have broken a bit.  In order to get success I need to click in preparing mode and not in attack mode.  

Development branch

Snarling:
Address / Fix #124

Slash game was previously changed so that the guard is vulnerable during "Preparing?" instead of "ATTACKING!" phase, but the time scaling was not changed to use the difficulty setting for the preparing phase, it was still using a hardcoded 250ms.

* Difficulty now scales time for correct phase
* Player loses game immediately after guard attacks (attack lasts for 250ms)

### DIFF
--- a/src/Infiltration/ui/SlashGame.tsx
+++ b/src/Infiltration/ui/SlashGame.tsx
@@ -40,18 +40,18 @@ export function SlashGame(props: IMinigameProps): React.ReactElement {
     }
   }
   const hasAugment = Player.hasAugmentation(AugmentationNames.MightOfAres, true);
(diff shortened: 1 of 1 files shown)


Totto16:
I think to really fix this one line is still wrong, ```js
   if (phase !== 1) {
      props.onFailure();
    } else {
      props.onSuccess();
    }

```

here a 2 has to be inserted instead of a 1, since otherwise it fails when pressing in the attack state, and succeeds in the preparing state!

PR #138, opened 2022-10-20
NETSCRIPT: Crimes streamlining / expanded enum use

### LINKED ISSUE #45: Toasts: use const objects (enums) instead of magic strings
[BrianLDev]
This came up in a broader discussion on the Discord dev channel about moving away from "magic strings" in general.

To start, hydroflame recommended converting Toasts, and then later we can look into enums for Gang names, crimes, universities, courses, etc.

image

[waffleattack]
> Could I have an example for what the magic strings are in toasts? 

[BrianLDev]
> @waffleattack since hydroflame was the source of this request, they're the best one to ask for more details.
> 
> But if I had to guess, I'd say that the toast variants should be converted from magic strings into const objects/enums.  Screenshot below:
> image
> 

[waffleattack]
> ah so instead of inputting 
> ```js
> ns.toast("Alert!", "warning", 10000)
> ```
> 
> you would instead be able to do
> ```js
> ns.toast("Alert!", TOASTS.WARNING, 10000)
> ```
> if I understand it correctly?

[BrianLDev]
> @waffleattack Yup, that's exactly how I'd do it.  But you might want to check with hydroflame to see if they had anything else in mind.

[phyzical]
> nah thats correct 👍 

[hydroflame]
> Yeah it's not that easy
> How do you feed `TOASTS` to the scripts?

[waffleattack]
> could we do it similar to ns.args? so instead of using `TOASTS.WARNING` you would use `ns.TOASTS.WARNING`?

[hydroflame]
> https://github.com/danielyxie/bitburner/blob/dev/src/ScriptEditor/NetscriptDefinitions.d.ts#L6426
> 
> If someone can uncomment this line and make it compile ...
>

### LINKED ISSUE #46: Suggestion: Add constants (enums) to singularity API to avoid "magic strings"
[BrianLDev]
This was talked about in depth on the Bitburner discord today (Feb 24, 2022), but to summarize:

There are a lot of Singularity API functions that require "magic strings" like University name, course name, Crime name, Faction name, etc.  "Magic strings" are generally considered to be no bueno when it comes to good programming habits, so it would be great if we could add some simple constant objects that contain the names of related items.

*note: these are usually called enums, but vanilla JS doesn't have enums, so they're implemented in JS as constant objects* 

Here's an example of all the crimes in a constant object (aka an enum):

```js
export const Crimes = {
	shoplift: "shoplift",
	rob: "rob store",
	mug: "mug someone",
	larceny: "larceny",
	dealDrugs: "deal drugs",
	bondForgery: "bond forgery",
	traffickIllegalArms: "traffick illegal arms",
	homicide: "homicide",
	grandTheftAuto: "grand theft auto",
	kidnap: "kidnap",
	assassination: "assassination",
	heist: "heist",
	none: "none"
};
```

Generally, enums allow for autocomplete which is one of the big benefits to avoid typos, misspellings, future changes to strings, etc.  The autocomplete in Bitburner doesn't always seem to correctly look up object properties, so we would have to also make sure that functionality works as well.  

Since autocomplete might create some spoilers, I highly recommend that these const objects (aka enums) get added to the Singularity API and marked as **FREE** (no RAM cost).  When players have reached the Singularity API, they won't be spoiled by having a full list of University names, Crime names, etc.  The other benefit of having these in the Singularity API is that players will probably run into problems with magic strings along the way, and experiencing that pain will show the value of using enums/const objects.

One nice thing about this change is that it would be "backwards compatible", meaning that the const enums will just provide a better way to input strings to function args, and all previously created functions and scripts will be unaffected.

Since this would be a moderate sized change, Hydroflame suggested that we start by converting the Toasts over to enums, so there's a separate issue created for that specific one (3032).  Once the Toasts issue is resolved, we can move on to this one.

**SUGGESTIONS FOR CONST OBJECTS/ENUMS:**

**No big spoiler concerns:**
- City names
- City.location names (locations within a city)
- University names
- University course names
- Gym names
- Work type
- Crimes
- Program names
- Hack type (weaken, grow, hack)

**Possible spoiler concerns (will have to discuss before implementing)**
- Company names
- Faction names
- Augmentation names

[BrianLDev]
> For the possible spoiler concern items: one way to minimize that is to keep the full enum hidden (e.g. Augmentation names), and create a separate object that contains only the items that the player has accessed so far.
> 
> That will prevent the player from seeing any spoilers, and new items can be added added every time a player accesses something new (e.g. joins a new faction with a list of augmentations not seen before)

Snarling:
Will fix #45, fix #46

Example of ingame documentation when using a function that expects a crime name or a toast variant:
image
image

Ingame documentation for ns.enums:
image
image
image


TheMas3212 (commented):
looks good to me

PR #140, opened 2022-10-21
DEVELOPMENT: Remove bundle files from git tracking

### LINKED ISSUE #129: Ignore `dist` folder from git version contorl
[Tanimodori]
Currently the `dist` folder is under git version control. The js bundle size of the project is ~17MB (`main.bundle.js`, `main.bundle.js.map`, `vendor.bundle.js`, `vendor.bundle.js.map`) and the `.git` folder is ~345MB now which is growing bigger and bigger and leads to performance drop of IDEs. Here is a list of commit numbers and `.git` folder size of popular project on github for comparisons.

| project                                            | `.git` size | #commit | size per commit |
|----------------------------------------------------|-------------|---------|-----------------|
| [`vite`](https://github.com/vitejs/vite)           | ~25MB       | ~4800   | ~5KB            |
| [`rollup`](https://github.com/rollup/rollup)       | ~24MB       | ~5000   | ~5KB            |
| [`electron`](https://github.com/electron/electron) | ~126MB      | ~26800  | ~5KB            |
| this project                                       | ~345MB      | ~7200   | ~48KB           |

This project has roughly 10x size of `.git` folder compared to other project. I see no benefits put them under git version control so I suggest removing them from git (and history maybe) before this project become unmaintainable.

Snarling:
bitburner-src repo is not running github pages, so a built version of the web app is not needed in git.

fix #129


Tanimodori:
why not just remove the entire `dist`? And the old dist is still here in the `.git` without changing the git history fyi.

Snarling:
@Tanimodori I don't think the entire dist directory is built from source (I could be wrong there, I can test later to find out). Anything that does not get built from source but is needed for the working game needs to remain (e.g. images) because otherwise checking out the game and building will not create a working game.

hydroflame (approved):

TheMas3212 (approved):
seems like good progress

PR #142, opened 2022-10-22
CORPORATIONS: Fix performance issue with lots of employees

### LINKED ISSUE #141: Fix performance with lots of corporation employees
Performance was really bad for me when I started having thousands of employees in each city.   I found a place to speed it up by only calculating research modifiers once per Industry.process() call instead of 4 times for each employee every cycle.

JasonGoemaat:
# CORPORATIONS: Fix performance issue with lots of employees

Using chrome devtools profiling I found that it took about 4 seconds to run the process and almost all of it was in the four methods called for each employee in 'calculateProductivity' like 'industry.getEmployeeCreMultiplier' along with the same method for Cha,  Int, and Eff, about 1/4 for each.   These in turn call researchTree.getEmployeeCreMultiplier().  Each of these then processes the whole research tree looking for multipliers and creating lots of objects along the way.

This change memoizes `getMultiplierHelper(propName: string)`, and adds a call to `Industry.process()` to reset it each time process() is called.

closes #141

### DIFF
--- a/src/Corporation/Industry.ts
+++ b/src/Corporation/Industry.ts
@@ -383,6 +383,11 @@ export class Industry {
   }
 
   process(marketCycles = 1, state: string, corporation: Corporation): void {
+    const researchTree = IndustryResearchTrees[this.type];
+    if (researchTree) {
+      researchTree.clearHelpers();
+    }
+
     this.state = state;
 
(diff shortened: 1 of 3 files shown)


hydroflame:
Let's just nuke employees and avoid caching issues like this

PR #148, opened 2022-10-24
MISC: remove trailing whitespaces

quacksouls:
The command `npm run format` reports some trailing whitespaces.  Remove those whitespaces.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,7 +18,7 @@ permissions:
   contents: read
   pages: write
   id-token: write
-  
+
 concurrency:
   group: "pages"
   cancel-in-progress: true
@@ -35,16 +35,16 @@ jobs:
         uses: actions/checkout@v2
         with:
           fetch-depth: 0
-          
+
       - name: Use Node.js 16.13.1
         uses: actions/setup-node@v2
         with:
           node-version: 16.13.1
           cache: "npm"
-          
+
       - name: Install NPM dependencies
         run: npm ci
-        
+
       - name: Build Production App
         if: ${{ github.event.inputs.buildApp == 'true' }}
         run: npm run build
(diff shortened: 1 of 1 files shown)


Snarling:
Usually these format or lint fixes will just make it into any other PR since every PR should theoretically be running `npm run format` and `npm run lint`. So there is usually no need to include these changes in a separate PR

quacksouls:
> Usually these format or lint fixes will just make it into any other PR since every PR should theoretically be running npm run format and npm run lint. So there is usually no need to include these changes in a separate PR

You mean when a PR is made, the linter and/or formatter that is set to run on GitHub would also remove trailing whitespaces?  Previously, when there were linting/formatting issues with the dev branch, it would take a while for the issues to be fixed.  And during the wait, anyone who runs the formatter/linter on their own fork would get formatter/linter errors whenever they want to test their code.  They could wait for the issues to be fixed in the dev branch and/or include the fixes in their WIP PR.  A minor problem, I think, but can be annoying when you want to make a PR (and its commits) as self-contained and independent as possible, and revolve around a central theme.

Snarling:
> Previously, when there were linting/formatting issues with the dev branch, it would take a while for the issues to be fixed. 

I think the main reason for that is that there were long periods of time between PRs getting merged, because PR merges were only occurring in large batches separated sometimes by weeks. In some cases, this also affected separate PRs that people created just to fix formatting or lint errors, like this one. So by the time the next batch of merging comes along, instead of 5 PRs that do other things but also fix the formatting (because they ran `npm run format` like they're supposed to), now there would be 5 PRs that include the formatting fixes and then an extra 1 PR that is only formatting fixes.

When there are multiple merge-ready PRs addressing formatting fixes, I would prefer merging a PR that does something else but just includes the fixes because they ran the formatter. The changelog is based on PR titles, and "MISC: remove trailing whitespaces" doesn't make sense/isn't useful on the changelog.

If there aren't any other PRs ready to merge, it's fine to just fix formatting in a separate PR like this. This might have even been the case earlier, but by the time I saw your PR there was also another PR ready to merge that included the fix, so that's why I closed this without merging.

PR #149, opened 2022-10-24
CORPORATION: Added check to buy amount

### LINKED ISSUE #127: Buying Infinity Corp Materials (ui breaks)
[Draco18s]
Due to a bug in my scripting, I accidentally divided by zero, assigning the resulting value to the parameter that went into a buyMaterial command.

This was the result:

image

Should only need this amount of code to replicate:

        ns.corporation.buyMaterial("Computers", "Sector-12", "RealEstate", 1/0);
    }

Result should be the Buy/s value being `NaN` trillions and the game will purchase "as many as you can afford/fit in the warehouse" AFAICT.

- [x] Save file [bitburnerSave_1665807278_BN3x1.zip](https://github.com/danielyxie/bitburner/files/9791140/bitburnerSave_1665807278_BN3x1.zip)
- [x] Minimal scripts to reproduce the issue
- [x] Steps to reproduce
- [x] Version of the game, e.g. Bitburner v2.1.0

G4mingJon4s:
# CORP: Added check to buy amount

The code now checks wether or not the given buy amount is a finite number, before changing the buy amount
# Linked issues

Closes #127

### DIFF
--- a/src/Corporation/ui/MaterialElem.tsx
+++ b/src/Corporation/ui/MaterialElem.tsx
@@ -125,7 +125,7 @@ export function MaterialElem(props: IMaterialProps): React.ReactElement {
           <Tooltip
             title={
               <Typography>
-                Buy: {numeralWrapper.format(mat.buy, nfB)} <br />
+                Buy: {mat.buy >= 1e33 ? mat.buy.toExponential(3) : numeralWrapper.format(mat.buy, nfB)} <br />
                 Prod: {numeralWrapper.format(mat.prd, nfB)} <br />
                 Sell: {numeralWrapper.format(mat.sll, nfB)} <br />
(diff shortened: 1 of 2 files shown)


Snarling:
Does not fix the linked bug. Linked bug of NaNt being displayed occurs with 1e33 or higher as the number to buy, it's not actually due to Infinity (which is not NaN)

G4mingJon4s:
The code now checks, wether or not the number is convertable by numeral js (1e33 above gets converted to its exponential string)

G4mingJon4s:
The last change didn't change the /s and button value.

Snarling:
Looks good, but needs an `npm run format` to fix line length issue for src\Corporation\ui\MaterialElem.tsx.
Should also be fixed with alt-shift-f from that file in vscode if you have it set to use prettier.

PR #151, opened 2022-10-24
CORPORATION: Give Employees some RNG

Caldwell-74:
Giving new Employees their old range of 50-100 for each of the avg stats to fulfill the "random" part of the Function Name


Snarling:
Fails build and prettier checks

For build, it's failing because getRandomInt is no longer being imported in that file, so it's not defined. You would need to re-add the import for it. You can also see build errors prior to submitting a PR by running `npm run start:dev`

For prettier, you can clear any formatting issues with `npm run format`, or if you know what files have an issue you can use alt+shift+f from vscode to format the file, as long as you have prettier set up as your code formatting in vscode.

Snarling:
Moved to #152 (your commits are still there), I just move onto a separate branch to add some test fixes & merged.

Issue #121, opened 2022-10-19
MP-100 is not a valid expression for Products


TheMas3212:
[Draco18s]
Setting a sell price of `MP-2` or smaller generates the notice `Error: Invalid value or expression for sell price field.`

`MP-2` *is* valid for Materials, however.

`MP-1` works, but any other negative value seems to fail.

- [x] Save file [bitburnerSave_1665085206_BN3x1.zip](https://github.com/danielyxie/bitburner/files/9728135/bitburnerSave_1665085206_BN3x1.zip)
- [x] Steps to reproduce
  1. Go to corp
  2. Go to any division that produces products (Software, Robots, etc)
  3. Set the sell price of any product to `MP-2`
- [x] Version of the game, e.g. Bitburner v2.1.0 (8f4636cb)

G4mingJon4s:
I checked the code:
When issueing a new sell price, it converts MP from the string to 1 to test the calculation with eval. When the calculated price is negative it throws the error. see [here](https://github.com/bitburner-official/bitburner-src/blob/dev/src/Corporation/Actions.ts#L159) (was edited)
To solve this you would need to replace MP with the current value, rather than 1.
This works with materials as its class has a property bCost, which is the market price.
The market price for a product however is not defined on its class AFAIK.
This also allows input like `100-MP`, which gets evaluated as 99, but in reality is a negative number most of the time.
The MP is calculated on every cycle: see [here](https://github.com/bitburner-official/bitburner-src/blob/dev/src/Corporation/Industry.ts#L902)
That shows that we can't just replace the MP with the price, as the values could be 0 or possibly undefined.
(ex. would be having the fulcrum dashboard upgrade, which allows market input before actual rating values etc. are given.)

One fix could be just replacing [line 903](https://github.com/bitburner-official/bitburner-src/blob/dev/src/Corporation/Industry.ts#L903) with `sCost = Math.min(0, eval(sCost));`
That however would give the user no information on why their product sell makes no profit.

PR #158, opened 2022-10-25
CORP: Better fix for MP validation

Snarling:
Improve the fix for #121.

* Minimum product sell price bound to product cost.
* Validation strings for MP-based sell price formula updated to provide better validation no matter the product or material's actual cost / market price. Should not be possible to get syntax errors for any MP on a string that was accepted as the formula.
* Validation string evaluation no longer required to be >0.


Snarling:
@G4mingJon4s FYI

PR #162, opened 2022-10-27
TUTORIAL: Fix Ram Text

jaculler:
See commit:
f756f09dbe Player starts with 8GB of RAM on home.

This updates the tutorial (and tests) to match the updated value.


quacksouls (approved):
Looks good.  As shown in the attached image, the home server starts with 8GB RAM.
tutorial

PR #164, opened 2022-10-28
DOCS: De-uglify ns.print examples

LJNeon:
I saw this example and it seems to completely misuse template strings, there's no need to concat within a template string.
I also noticed that the first example use an unnecessary amount amount of variables and so I fixed it to what the former second example seemed to be trying to do.


quacksouls (approved):
All good and much more beautiful now.

PR #165, opened 2022-10-28
add formulas support for companies gain

AlexeyKozhemiakin:
add formulas support for companies gain

### DIFF
--- a/src/Netscript/RamCostGenerator.ts
+++ b/src/Netscript/RamCostGenerator.ts
@@ -600,6 +600,7 @@ export const RamCosts: RamCostTree<Omit<NSFull, "args" | "enums">> = {
       crimeGains: 0,
       classGains: 0,
       factionGains: 0,
+      companyGains : 0
     },
   },
 } as const;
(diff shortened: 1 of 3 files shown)


Snarling:
This is not the correct way to implement this. Formulas API stuff should work no matter the current status of the real player, but this function requires the player to actively be employed by the company being checked (and does not allow the player to specify a particular position at the company, it can only give info about the current position).

It should be similar to the factionGains function where it just takes in a job type and a favor (rather than caring about which company it's at).

AlexeyKozhemiakin:
I just wrapped the existing calculateCompanyWorkStats, the implementation was as-is. Apparently it depends on Company-level attrributes expMultiplier and salaryMultiplier, so when i started to refactor similar to FactionWork using position type and favor - found this.

```
{
    name: LocationName.AevumBachmanAndAssociates,
    info: "",
    companyPositions: Object.assign({}, AllTechnologyPositions, AllBusinessPositions, AllSecurityPositions),
    expMultiplier: 2.6,
    salaryMultiplier: 2.6,
    jobStatReqOffset: 224,
  }
```,

Snarling:
Added via #195 which was built off this one.

PR #168, opened 2022-10-28
DOC: `ns.singularity.workForFaction()`: update doc and examples

### LINKED ISSUE #166: Singularity.workForFaction() documentation remarks
Bitburner version: 2.1.0

The remarks in the documentation for `ns.singularity.workforfaction()` in `markdown/bitburner.singularity.workforfaction.md` seem to be incorrect?

The remarks say you will not receive your reputation until you finish the task and provides a workaround which keeps restarting `workforfaction()` every 60 seconds. But I just tried it out in bitburner and I do get the reputation right away, and `getFactionRep()` reflects it correctly as well.

quacksouls:
Fixes #166.  Since v2.0, while working for a faction the player immediately gets their rewards, e.g. faction reputation and XP.  Update the documentation of `ns.singularity.workForFaction()` to reflect this immediate gain.  Update the examples accordingly.

### DIFF
--- a/src/ScriptEditor/NetscriptDefinitions.d.ts
+++ b/src/ScriptEditor/NetscriptDefinitions.d.ts
@@ -1837,27 +1837,29 @@ export interface Singularity {
    *
    * This function will return true if you successfully start working for the specified faction, and false otherwise.
    *
-   * Note that when you are working for a faction, you will not actually receive your earnings (reputation, experience) until you FINISH the action.
-   *
    * @example
    * ```ts
-   * // NS1:
-   * //If you only want to work until you get 100,000 faction reputation. One small hack to get around this is to continuously restart the action to receive your earnings:
-   * while (getFactionRep(FACTION NAME) < VALUE) {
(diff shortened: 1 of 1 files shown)


Snarling:
I think the only reason it needed a disclaimer section about when/how you should check your current faction rep is because you had to go through a weird process (ending faction work / checking current rep / restarting work) in order to get updated numbers for your current rep.

Because rep is now rewarded continuously while you work, there is no longer any special process needed, you will always get an accurate result when you check your current rep.

I think the correct documentation change here is to remove this verbose disclaimer/explanation section (rather than just modifying it), because the way this works no longer needs a special disclaimer.

quacksouls:
> I think the correct documentation change here is to remove this verbose disclaimer/explanation section (rather than just modifying it), because the way this works no longer needs a special disclaimer.

Yes, this makes more sense: let the UI show the work gains rather than explain it in the doc.  Please see the updated commit 1bf44ef76d3b2f4d7915d6494c52d35416c00515.  Everything else in the original commit is the same.

Snarling:
Included in #180

PR #171, opened 2022-10-30
DOCUMENTATION: UPDATE TO EXPLAIN ARGS USAGE IN NS2

ModestMC:
First, this repo does not seem to be where the site I'm reading is generated from; thus, this is not the documentation I saw.

Second, unless the tail function does something odd I have yet to see, when I attempt to run the NS2 code shown in this file it does not work. This PR does not remove said code.

The [documentation](https://bitburner.readthedocs.io/en/latest/netscript/netscriptscriptarguments.html) does not explain how NS2 scopes the game arguments, I wasted 30 minutes trying to figure out why I couldn't get args to work as this file stated. It's explained in other files, but this one looks like it missed the updating.

Simple fix, I suspect this is just a minor oversight, code now includes an example of how to use it with ns2.

### DIFF
--- a/doc/source/netscript/netscriptscriptarguments.rst
+++ b/doc/source/netscript/netscriptscriptarguments.rst
@@ -34,3 +34,16 @@ ns2 script:
 const args_obj = arguments[0]
 const argument1 = (args_obj.server.args[0])
 const argument2 = (args_obj.server.args[1])
+
+When using ns2, the main function takes as input the game's ns object
+containing the argument array you are trying to pass into the script,
+so the arguments can be referenced from it like this::
+
+    export async function main(ns) {
+	    if (ns.args[0]){ 
+		    while (true){
+			    await ns.hack(ns.args[0]);
+		    }
+	    }
+	    else{ns.print("No target given")}
+    }


quacksouls:
1. Use "NS2" instead of "ns2".

2. Format code example in RST as
```js
.. code:: javascript

    line 1;
    line 2;
    line 3;
```

3. I started a new game at

https://danielyxie.github.io/bitburner/

I copied the code in the commit, saved it to the file `demo.js`, and ran it from the Terminal as:

```sh
$ run demo.js n00dles
```

I got the error as shown in the attached image.
args

Perhaps you want to replace the `while` loop that hacks a target server, provided as a command line argument, with something simpler such as printing the argument provided.  If no arguments are provided, then the script would immediately quit and you would not be able to see its log.  In this case, using `ns.tprint()` would be preferable to `ns.print()`.  A simpler example would be something like the following:

```js
.. code:: javascript

        if (ns.args.length > 0) {
            ns.tprint("First argument: " + ns.args[0]);
        } else {
            ns.tprint("No arguments provided.");
        }
    }
```

The line `ns.tprint("First argument: " + ns.args[0]);` can be written using string template as

```js
ns.tprint(`First argument: ${ns.args[0]}`);
```

but this is beginner's guide so no string template necessary.

Snarling:
Modified and merged via #220

PR #182, opened 2022-11-02
STOCK:Adds function to return the organization for a stock symbol

### LINKED ISSUE #181: Feature: Add a way to map stock symbols to organization
There is no way to map the stock symbols to an organization (useful for deciding to have grow/hack impact stock price).

Others have statically mapped them out, but that doesn't future proof, and manual data entry doesn't seem like the "bitburner" way. It seemed like a simple oversight, especially after seeing the field is already included in the stock class.

SamuraiNinjaGuy:
Currently there is no way to map stock symbols to the organization. This fixes that.

# Linked issues
closes #181

# Bug fix
Tested NS1 and NS2, requires WSE and TIX.

NS1
tprint(stock.getOrganization('SLRS'))

NS2
ns.tprint(ns.stock.getOrganization('ECP'))

### DIFF
--- a/src/ScriptEditor/NetscriptDefinitions.d.ts
+++ b/src/ScriptEditor/NetscriptDefinitions.d.ts
@@ -1121,6 +1121,46 @@ export interface TIX {
   getPrice(sym: string): number;
 
   /**
+   * Returns the organization associated with a stock symbol.
+   *
+   * @remarks
+   * RAM cost: 2 GB
+   *
+   * The organization associated with the corresponding stock symbol. This function
+   * requires that you have the following:
+   *
+   * 1. WSE Account
+   *
+   * 1. TIX API Access
+   *
+   * @example
+   * ```ts
+   * // NS1
+   * stock.getOrganization("FSIG");
+   *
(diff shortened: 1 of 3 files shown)


src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -1120,6 +1120,46 @@ export interface TIX {
...
+   * @example
+   * ```ts
+   * // NS1
+   * stock.getOrganization("FSIG");
+   *
+   * // Choose the first stock symbol from the array of stock symbols. Get the
+   * // organization associated with the corresponding stock symbol
+   * var sym = stock.getSymbols()[0];
+   * tprint("Stock symbol: " + sym);
+   * tprint("Stock price: " + stock.getOrganization(sym));
```
Snarling, line 1145: Should be "Stock organization: ", not price

Snarling (changes requested):
Sorry, never got around to looking at this one until now.

Snarling:
Merged as #234

PR #186, opened 2022-11-03
UI: SoftResetButton.tsx Tooltip changed to make more sense

rai68:
Changed tooltip to be more fitting.

bitburn

Changed to `"Perform a soft reset. This will perform the same action as installing Augmentations."`

### DIFF
--- a/src/ui/React/SoftResetButton.tsx
+++ b/src/ui/React/SoftResetButton.tsx
@@ -28,7 +28,7 @@ export function SoftResetButton({
 
   return (
     <>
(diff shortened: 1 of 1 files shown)


quacksouls:
The proposed change seems misleading to me.  The new sentence:

`"This will perform the same action as installing Augmentations."`

implies that a soft reset also installs any Augmentations you have bought but are waiting to be installed.

This makes more sense to me:

`"Resets everything as if you had just installed an Augmentation."`

It tells you that a soft reset would result in a reset, like what you would get after installing one or more Augmentations.  Furthermore, it does not imply that a soft reset would install any Augmentations you have bought.

PR #188, opened 2022-11-04
UI: use newline to separate sentences in popup message

quacksouls:

> One more presentation layer issue
>
> Bitburner v2.1.0 (6ac04717) 
break

After the player has completed the creation of a program, a popup message appears to inform that the program is completed.  ~We should use whitespace to separate the sentences in the popup message, as done in these files:~

1. ~[src/DarkWeb/DarkWeb.tsx](https://github.com/bitburner-official/bitburner-src/blob/6ac04717fe363b2054bb8b7b28211f9e6f2294e9/src/DarkWeb/DarkWeb.tsx#L83)~
2. ~[src/NetscriptFunctions/Singularity.ts](https://github.com/bitburner-official/bitburner-src/blob/6ac04717fe363b2054bb8b7b28211f9e6f2294e9/src/NetscriptFunctions/Singularity.ts#L546)~

Before:
break-before

~After:~
break-after

### DIFF
--- a/src/Work/CreateProgramWork.ts
+++ b/src/Work/CreateProgramWork.ts
@@ -85,7 +85,7 @@ export class CreateProgramWork extends Work {
           `You've finished creating ${programName}!`,
           "The new program can be found on your home computer.",
         ];
-        dialogBoxCreate(lines.join("<br>"));
+        dialogBoxCreate(lines.join("\n"));
       }
 
       if (!Player.getHomeComputer().programs.includes(programName)) {


Snarling:
Should be `\n` instead of a space I think. Preserves how the dialog originally looked and seems like it would be more readable.

quacksouls:
> Should be `\n` instead of a space I think. Preserves how the dialog originally looked and seems like it would be more readable.

Good catch!  The commit is updated.  Here's what the dialog box looks like when a newline character is used for line break.

program

PR #190, opened 2022-11-04
DOC: `ns.args`: improve documentation and examples

quacksouls:
Elaborate on the documentation of the arguments array `ns.args`.  Provide examples to illustrate how to pass arguments from the Terminal and extract the provided arguments within a script.


Snarling:
Modified and merged via #220

PR #191, opened 2022-11-04
API: Align hostname validation in port hack programs

Voileexperiments:
`ns.brutessh` is missing the `hostname === undefined` validation.

(Also a full stop was missed in `ns.httpworm` validation fail message.)


Snarling:
Neither of these error messages will ever be reached. helpers.string will already create an error dialog with the needed info, the correct fix is to just remove the undefined check because after helpers.string we already know that it's a string (so it can't be undefined).

image

Snarling:
Removed the extra error sections on #194

PR #192, opened 2022-11-04
UI: `cat`: proper line breaks when showing .js, .script, or .txt files

quacksouls:

> I just found another instance of similar issue.
> Maybe it is time to check every usage of `<br>`.
>
> How to reproduce: type cat <any_js_file> on terminal 
dialog

Currently, the HTML line break sequence `<br /><br />` is hardcoded into the dialog box message when showing the content of these file types: ".js", ".script", and ".txt".  By default, the function `dialogBoxCreate()` currently assumes that its first parameter is not HTML, but a text string, so whatever is in the string will appear in the dialog box.  

### Before patch

Here are the current dialog boxes.

**cat [.js file]**
cat-js-before

**cat [.script file]**
cat-script-before

**cat [.txt file]**
cat-txt-before

### How to fix

Use the newline character instead for line break.  If we still want HTML, then the following code also works:

**src/Terminal/commands/cat.ts**
```js
const isHtml = true;
dialogBoxCreate(`${script.filename}<br /><br />${script.code}`, isHtml);
```

**src/TextFile.ts**
```js
const isHtml = true;
dialogBoxCreate(`${this.fn}<br /><br />${this.text}`, isHtml);
```

### After patch

Here are dialog boxes after the patch.

**cat [.js file]**
cat-js-after

**cat [.script file]**
cat-script-after

**cat [.txt file]**
cat-txt-after

### DIFF
--- a/src/Terminal/commands/cat.ts
+++ b/src/Terminal/commands/cat.ts
@@ -49,7 +49,7 @@ export function cat(args: (string | number | boolean)[], server: BaseServer): vo
   } else if (filename.endsWith(".script") || filename.endsWith(".js")) {
     const script = Terminal.getScript(relative_filename);
     if (script != null) {
-      dialogBoxCreate(`${script.filename}<br /><br />${script.code}`);
+      dialogBoxCreate(`${script.filename}\n\n${script.code}`);
       return;
     }
   }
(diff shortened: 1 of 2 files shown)


Snarling:
Thanks, looks like I missed some of these when I changed dialogBoxCreate.

PR #213, opened 2022-11-12
NETSCRIPT: Allow concurrent calls to instant functions

N00byEdge:
This allows delay free calls like `getPlayer()` during time consuming functions like `hack()` or `installBackdoor()`.

Note that this **does NOT** allow multiple calls at the same time to the latter category of functions (you still can only call `hack()` once at a time from a script), so it should not affect balance or anything, just make the API much less cumbersome to work with.


Snarling:
Scripts only being able to perform one active ns function at a time is a purposeful design choice.

PR #224, opened 2022-11-23
UI: Fix keyboard shortcuts for other keyboard layouts

d0sboots:
A prior change changed from event.key to event.code. I'm not sure why this commit was made, since there's no further explanation in the commit message. But using event.code breaks non-QWERTY keyboard layouts, so reverting improves my (Dvorak) experience greatly.
Also, it's strongly warned against at MDN:
https://developer.mozilla.org/en-US/docs/Web/API/Element/keydown_event#event_properties

This also fixes ScriptEditorRoot.tsx, where some code snuck in that wasn't using *any* of the constants.

This reverts commit 016a9a873fba71388bdbaa1423134fd1f2335a15.

## Testing

Tried the various hotkeys, they work correctly for Dvorak now. In particular, Ctrl-X (which is physically Ctrl-B) now cuts text instead of exiting to the terminal in the script editor.

Switching to Qwerty, the hotkeys match their "usual" behavior. I didn't detect any weird cases, and I tried a bunch in both layouts.

### DIFF
--- a/src/ScriptEditor/ui/ScriptEditorRoot.tsx
+++ b/src/ScriptEditor/ui/ScriptEditorRoot.tsx
@@ -4,6 +4,7 @@ import * as monaco from "monaco-editor";
 
 type IStandaloneCodeEditor = monaco.editor.IStandaloneCodeEditor;
 type ITextModel = monaco.editor.ITextModel;
+import { KEY } from "../../utils/helpers/keyCodes";
 import { OptionsModal } from "./OptionsModal";
 import { Options } from "./Options";
 import { isValidFilePath } from "../../Terminal/DirectoryHelpers";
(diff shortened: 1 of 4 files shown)


d0sboots:
Thanks to @TheMas3212, I discovered https://github.com/danielyxie/bitburner/issues/3592 which was the original motivation for the change. I don't have a mac to test with, but https://github.com/danielyxie/bitburner/pull/3605 claims that "Windows/Mac/Linux with US-intl/US/en_GB/NL keyboard layouts (these were all affected in the same way)"

I have tested specifically with Windows US-intl. The behavior I see is that **currently** (on the live version of the game), if I use *right*-Alt+S, I get ß(eszett). This is because on intl (and presumably the others), *right*-Alt is mapped to AltGr. It does *not* show up as event.altKey (nor any other flags), so there is no way to trap for it - you will always get ß.

In contrast, *left*-Alt is normal alt. It works properly, both currently and with my revert (and presumably, when the original bug was filed).

TL;DR: I don't think the "fix" ever actually fixed anything.

d0sboots:
I'm not sure what's up with the checks. These files are not generated, as far as I'm aware, and lint was clean when I ran it myself.

Snarling:
Yeah I have no idea what's up with these check failures either, I think this is fine for merging.

PR #225, opened 2022-11-24
DOC: Sleeve: link the Sleeve API in the list of APIs

quacksouls:

> In head branch, it seems that there is type named sleeve (can get by ns.sleeve), and there is type named Sleeve (can get by ns.sleeve.getSleeve(0)).
The markdown document doesn't have former, only latter. i.e. I can't see markdown document of list of function of ns.sleeve.
https://github.com/bitburner-official/bitburner-src/blob/dev/markdown/bitburner.sleeve.md

The URL

https://github.com/bitburner-official/bitburner-src/blob/dev/markdown/bitburner.sleeve.md

currently points to this page:

sleeve

The expected page is something like this:

https://github.com/danielyxie/bitburner/blob/dev/markdown/bitburner.sleeve.md

The interface that defines the Sleeve object clashes with the interface that defines the Sleeve API.  The former is named `Sleeve` (uppercase), whereas the latter is named `sleeve` (lowercase).  API Extractor does not like this and inserts a link to the Sleeve object in place of where the Sleeve API should be.  A fix is to rename the interface for the Sleeve object as `SleeveObject`.  Or something other than using lowercase and uppercase to differentiate the Sleeve object interface from the Sleeve API interface.

### DIFF
--- a/src/ScriptEditor/NetscriptDefinitions.d.ts
+++ b/src/ScriptEditor/NetscriptDefinitions.d.ts
@@ -55,7 +55,7 @@ interface Player extends Person {
 }
 
 /** @public */
-interface Sleeve extends Person {
+interface SleeveObject extends Person {
   shock: number;
   sync: number;
   memory: number;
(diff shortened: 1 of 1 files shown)


quacksouls:
Closed because this causes a build failure, which I do not know how to fix.

Snarling:
@quacksouls Remade as #227 , if you're curious what needed to be changed to fix the build errors, some changes also needed to be made in NetscriptFunctions/Sleeve.ts and PersonObjects/Sleeve/Sleeve.ts

PR #223, opened 2022-11-22
MISC: Significantly improve the speed of script creation

d0sboots:
In my browser (Chrome 107.0.5304.107 Windows 10) this improves noop script execution from 8281.063/sec to 13942.606/sec - a 68% speedup!

The 'ns' API has to be wrapped, so that functions can access the context (primarily the WorkerScript) without leaking it to the player. Before, this happened by walking the whole API tree and creating new wrapping functions for every script, which is understandably expensive.

Now we take advantage of class #private fields to securely hold the context, allowing the wrapping functions to exist on the prototype and only need to be created once. These are widely available (>93% according to caniuse) and work with Electron.

# Benchmarks
benchmark.js
```js
/** @param {NS} ns */
export async function main(ns) {
    const resolved = Promise.resolve();
    const perf = globalThis.performance;
    let speedMax = 0;
    for (let i = 0; i < Number(ns.args[0]); ++i) {
        const start = perf.now();
        let end, cycles;
        for (cycles = 0; (end = perf.now()) - start < 1000; cycles++) {
            ns.exec("/lib/noop.js", "home");
            await resolved;
        }
        const speed = (cycles * 1000) / (end - start);
        speedMax = speed > speedMax ? speed : speedMax;
        ns.tprintf("Iter %2d: %.3f/sec", i, speed);
        await new Promise(resolve => setTimeout(resolve));
    }
    ns.tprintf("Best: %.3f/sec", speedMax);
}
```
/lib/noop.js
```js
/** @param {NS} ns */
export function main(ns) {}
```
### Before
```
[home ~/]> run benchmark.js 30
Running script with 1 thread(s), pid 128630 and args: [30].
Iter  0: 7246.000/sec
Iter  1: 7737.905/sec
Iter  2: 6598.361/sec
Iter  3: 6433.427/sec
Iter  4: 7428.571/sec
Iter  5: 7370.000/sec
Iter  6: 7113.155/sec
Iter  7: 7682.927/sec
Iter  8: 8137.167/sec
Iter  9: 7428.029/sec
Iter 10: 7266.000/sec
Iter 11: 7976.000/sec
Iter 12: 7094.581/sec
Iter 13: 7966.000/sec
Iter 14: 7235.018/sec
Iter 15: 7076.754/sec
Iter 16: 7199.560/sec
Iter 17: 7373.051/sec
Iter 18: 7654.469/sec
Iter 19: 7538.738/sec
Iter 20: 8281.063/sec
Iter 21: 7820.218/sec
[...]

PR #229, opened 2022-11-25
NETSCRIPT: Faster API wrapping on script launch.

Snarling:
Credit to @d0sboots , this is basically his work from #223 just refactored a bit.

* ns API is wrapped once
* reference to workerscript is kept in a private class field, and this is all that needs to be updated when running a new script. All layers of ns are stamped with this private class field.
* Significantly increases maximum number of scripts before out-of-memory crash (for performance reasons it is still not recommended to run more than a few thousand scripts at once).
* Significantly speeds up the time needed to generate a new WorkerScript.

**_BREAKING CHANGE:_** if you take an ns function and put it into another variable off of ns, you will need to bind the function to ns (or to its parent layer) for it to continue to work. e.g. 
```js
// No longer works
let myTprint1 = ns.tprint

// Still works
let myTprint2 = ns.tprint.bind(ns);

myTprint2("Hello world"); // Works
myTprint1("Goodbye world"); // Error due to function not having an ns layer as a this value
```

There are still types that should be improved, todos have been added in those locations.

### Benchmark1:
10k script executions, both one-at-a-time and all-at-once. Over 6x faster in one-at-a-time execution and almost 2x as fast for all-at-once.
image

### Benchmark2:
Running as many scripts as possible before browser crashes due to out-of-memory error. Also include how long it took to launch each of the first 60 sets of 1000 scripts.
**Dev**: browser crashed at 63000
image

**This PR**: At least 344k scripts at once. Test had ran for a very long time at this point. so I just closed it.

https://user-images.githubusercontent.com/84951833/204103009-7071bf49-7f6b-4b89-a014-b4e5960291a1.mp4

PR60kScriptTimes

### DIFF
--- a/src/NetscriptFunctions.ts
+++ b/src/NetscriptFunctions.ts
@@ -1910,16 +1910,6 @@ export const ns = {
 };
 
 const wrappedNS = wrapAPILayer({} as ExternalAPI<NSFull>, ns, []);
-function freezeEnums(obj: object, isEnums = false) {
-  if (isEnums) Object.freeze(obj);
-  for (const [k, v] of Object.entries(obj)) {
-    if (isEnums) Object.freeze(v);
-    else if (typeof obj === "object") {
-      freezeEnums(v, k === "enums");
-    }
-  }
-}
-freezeEnums(wrappedNS);
 
 export function NetscriptFunctions(ws: WorkerScript): ExternalAPI<NSFull> {
   const instance = new StampedLayer(ws, wrappedNS) as any;
@@ -1932,7 +1922,7 @@ function stampLayers(ws: WorkerScript, stamped: any, unstamped: any) {
   for (const [k, v] of Object.entries(unstamped)) {
     if (typeof v === "object" && k !== "enums") {
       stamped[k] = new StampedLayer(ws, v);
-      stampLayers(ws, stamped[k], unstamped[k]);
+      stampLayers(ws, stamped[k], v);
     }
   }
 }
--- a/src/NetscriptFunctions.ts
+++ b/src/NetscriptFunctions.ts
@@ -1909,22 +1909,32 @@ export const ns = {
   ...NetscriptExtra(),
 };
 
-const wrappedNS = wrapAPILayer({} as ExternalAPI<NSFull>, ns, []);
+export const wrappedNS = wrapAPILayer({} as ExternalAPI<NSFull>, ns, []);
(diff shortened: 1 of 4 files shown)


src/NetscriptFunctions.ts
```diff
@@ -1913,6 +1909,34 @@ export const ns = {
...
+
+export function NetscriptFunctions(ws: WorkerScript): ExternalAPI<NSFull> {
+  const instance = new StampedLayer(ws, wrappedNS) as any;
+  instance.args = ws.args.slice();
+  stampLayers(ws, instance, wrappedNS);
+  return instance;
+}
+
+function stampLayers(ws: WorkerScript, stamped: any, unstamped: any) {
+  for (const [k, v] of Object.entries(unstamped)) {
```
d0sboots: Suggestion: We've already figured out the nature of all the keys in wrapAPILayer. If that returned a secondary object that was just a pruned tree of API names, we could traverse just that and avoid checking the whole API tree at script creation time.

src/NetscriptFunctions.ts
```diff
@@ -1913,6 +1909,34 @@ export const ns = {
...
+    if (typeof v === "object" && k !== "enums") {
+      stamped[k] = new StampedLayer(ws, v);
+      stampLayers(ws, stamped[k], unstamped[k]);
```
d0sboots: `unstamped[k]` is just `v`.

src/NetscriptFunctions.ts
```diff
@@ -1913,6 +1909,34 @@ export const ns = {
   ...NetscriptExtra(),
 };
 
+const wrappedNS = wrapAPILayer({} as ExternalAPI<NSFull>, ns, []);
+function freezeEnums(obj: object, isEnums = false) {
```
d0sboots: Why not do this in wrapAPILayer, which is already doing this traversal *and* checking for enums?

d0sboots (commented):
I hope you don't mind me leaving a few comments, since you rewrote my code. :)

src/NetscriptFunctions.ts
Snarling: Moved enum freezing to wrapAPILayer

src/NetscriptFunctions.ts
Snarling: Addressed in 1dc7bea

PR #230, opened 2022-11-25
DOC: `ns.codingcontract.attempt()`: how to use the optional parameter `opts`

quacksouls:
Explain how to use the optional parameter `opts` of the function `ns.codingcontract.attempt()`.  It is not obvious that an object should be passed in, using the key/value format of the interface `CodingAttemptOptions`.  I have seen players passed in `true` for this fourth parameter.  The Bitburner server of Discord have had many questions about the parameter `opts`.

The correct way to use the fourth parameter `opts` is to pass in an object along the key/value format of `CodingAttemptOptions`.  Also update the documentation of `opts` and the return value.  I have formatted the documentation to use at most 90 characters because it is no fun to scroll across to read a very long line.  Others might disagree with this formatting and that's OK.  I can change the doc back to very long lines.

### DIFF
--- a/src/ScriptEditor/NetscriptDefinitions.d.ts
+++ b/src/ScriptEditor/NetscriptDefinitions.d.ts
@@ -3155,11 +3155,42 @@ export interface CodingContract {
    *
    * Attempts to solve the Coding Contract with the provided solution.
    *
+   * @example
+   * ```ts
+   * // NS1
+   * // The result is a string, the default behavior. The code is the same as this:
+   * // var result = codingcontract.attempt(yourSolution, filename, hostname, { returnReward: true });
+   * var result = codingcontract.attempt(yourSolution, filename, hostname);
+   *
+   * // The result is a boolean.
+   * var result = codingcontract.attempt(yourSolution, filename, hostname, { returnReward: false });
+   * ```
+   * @example
+   * ```ts
+   * // NS2
+   * // The result is a string, the default behavior. The code is the same as this:
+   * // const result = ns.codingcontract.attempt(yourSolution, filename, hostname, { returnReward: true });
+   * const result = codingcontract.attempt(yourSolution, filename, hostname);
+   *
+   * // The result is a boolean.
+   * const result = ns.codingcontract.attempt(yourSolution, filename, hostname, { returnReward: false });
+   * ```
+   *
    * @param answer - Solution for the contract.
    * @param filename - Filename of the contract.
-   * @param host - Hostname of the server containing the contract. Optional. Defaults to current server if not provided.
-   * @param opts - Optional parameters for configuring function behavior.
-   * @returns True if the solution was correct, false otherwise. If the returnReward option is configured, then the function will instead return a string. If the contract is successfully solved, the string will contain a description of the contract’s reward. Otherwise, it will be an empty string.
+   * @param host - Hostname of the server containing the contract. Optional. Defaults to
+   *     current server if not provided.
+   * @param opts - Optional parameters for configuring function behavior. Pass in an
+   *     object following the key/value format of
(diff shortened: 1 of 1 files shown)


Snarling:
@quacksouls I thought this was way too verbose, but part of that is just the weirdness that comes from the CodingAttemptOptions parameter which is difficult to document properly / isn't easy to know what you need to provide when using the ingame IDE.

So in building off of this, I also just changed how the function works, so it takes in a boolean now instead of an options object with one possible option. Way simpler to use and way simpler to document.

Also the default behavior is supposed to be a boolean return (a couple months ago I accidentally reversed it so string return was default) so I restored that as well.

Replaced with #231

Snarling:
Also regarding line width for the documentation, the standard line width cap should be 120 because that's what we have set in .prettierrc for the rest of the code. So the previous documentation was indeed way too wide.

quacksouls:
Honestly, I feel like I don't know how to document anything in this game.  There is no set guideline for documenting functions and parameters.  When I try to clarify the behaviour of a parameter from whatever cryptic documentation is currently available and by reading the  source code, my PR seems to be way off from what is needed.  I don't seem to know what sort of documentation is needed for a function or parameter or what to write as an example.  I really don't want to waste any more of the reviewers' time in reading my PRs that seem to mostly be rejected anyway.  I think I'll work on typo PRs from now on.  Sorry to have wasted your time in reading my work.

Snarling:
@quacksouls In general, I do tend to have a more spartan take on documentation than you, and I have either kicked back or manually thinned out documentation PRs when I see them (not just from you). Maybe that's my bad, I am probably being way too picky about the documentation.

Here's my philosophy. NetscriptDefinitions.d.ts is technical programming documentation, so it should be concise, technical, and doesn't need to teach the player how to use js (e.g. stuff like "if you want to specify the 4th argument, you can't just skip the 3rd" shouldn't need to be explicitly stated in a function's documentation). There is other less technical documentation in the docs folder (though it's not at the function-specific level).

You've contributed a lot and I hope you continue to contribute.

--

In this case part of the issue was that I introduced a bug a few months ago where the default value was backwards, so I already needed to make a PR to fix that. Then I would need to have you totally change the documentation again because your defaults would be backwards. If there wasn't also a bug in the game function, I would probably have just merged your PR.

So because of needing to fix that bug, I just built a new PR off of yours because I had to make one anyway. Once I have a PR in progress for something I tend to make more related changes. Since I was fixing the bug and would probably make other related changes, I figured I would fix the docs as well. I built off your PR so you would still be credited in the commits for the original doc update, even though some of it got removed afterwards.

As expected from there, I end up making more changes, at first I just got rid of the whole CodingContractOptions interface because it just makes it difficult to tell what you're supposed to provide when you're a player just trying to write a script in the IDE, e.g. on live you would see this:
image
Which is pretty impossible to know what you're supposed to provide for `opts`. The player should not be required to open up a browser and check the documentation to know what to provide for a parameter.

The extra description from your PR does make it easier to know inside the IDE that you're supposed to provide something in a {returnReward: boolean} format, but that's type information so it really should just be the type. So I tried changing the type to { returnReward: boolean } which was fine, but I was having trouble trying to add documentation for `@param opts.returnReward` in a way that would build the docs. 

Then I realized an options object doesn't even make sense when there's a single option, so I thought about just changing it to a standalone boolean flag. But in the current draft, I'm considering just getting rid of returnReward altogether and making it always return a string (which can already be used as a boolean since a failure is empty string which is falsy). There doesn't seem to be a downside to just making it always return the string.

So since I'm jumping all around with these game-side changes, it would not have been fair to expect you to keep this PR open and constantly shift the documentation to match whatever I'm doing on the game function side.

PR #231, opened 2022-11-26
CONTRACT: codingcontract.attempt always returns a string

Snarling:
**SORT OF BREAKING CHANGE**: codingcontract.attempt always returns a reward string, or an empty string on attempt failure.

This is breaking for anyone who was directly comparing e.g. `returnValue === true`. However, anyone who was using booleans before and was simply using the booleans directly as a conditional should be unaffected, e.g. `if (ns.codingcontract.attempt(...))`

Adapted from #230

Fails lint and generated file checks because this includes updated documentation.


Snarling:
Considering just making this always return a reward string for success / empty string for failed. The string can already be used as a boolean because empty string is falsy.

PR #241, opened 2022-12-02
MISC: Refactor "Page" enum and surrounding functions (no observable changes)

d0sboots:
Instead of mapping to opaque numbers, the enum now holds useful string keys. This eliminates the need for translation code (or another map) elsewhere.

Also removed are the host of Router.toX() functions. They are replaced with Router.toPage(page), a single generic function. For the cases where the transition is more complicated (requires extra params), the functions are kept. The Page enum is split into SimplePage and ComplexPage, so that type-checking keeps this safe. For the most part, this is an implementation detail and everyone will continue to use Page.

In SidebarRoot, this allows for significant reduction in trivial clickPage() functions, and there's also a consolidation of the tutorial "flash" handling logic. The real payoff for this comes in the next PR.

An additional change that snuck in: Responsibility for Settings.isSidebarOpened moved entirely into SidebarRoot, it was being duplicated oddly before.

The large number of files touched were done with regex, to reduce chance of error. (Also, because it's a lot.)

# Testing
I verified the tutorial still works all the way through (including flash highlighting)

### DIFF
--- a/src/Augmentation/AugmentationHelpers.tsx
+++ b/src/Augmentation/AugmentationHelpers.tsx
@@ -21,6 +21,7 @@ import {
   initUnstableCircadianModulator,
 } from "./data/AugmentationCreator";
 import { Router } from "../ui/GameRoot";
+import { Page } from "../ui/Router";
 import { mergeMultipliers } from "../PersonObjects/Multipliers";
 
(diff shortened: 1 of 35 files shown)


Snarling:
This needs a rebase due to merging the keyCode revert revert

d0sboots:
done

PR #246, opened 2022-12-04
UI: Break SidebarRoot into smaller components, and memoize

### LINKED ISSUE #238: SidebarRoot can take >50ms to render, and rerenders every 200ms
- Steps to reproduce
  - Open the profiling tab and take a profile. The sidebar will clearly appear as a large timesink, although it might not be as dramatic if you have a faster PC.
- Bitburner v2.1.0 (8f4636cb), probably all earlier versions

Needless to say, this severely impacts things like batching that depend on stable timing.

I am actively working on this issue.
Sidebar Slow

d0sboots:
The sidebar was much too big to be one component. By splitting it up, the code is more modular, and it also opens up a lot of opportunities for memoization. In dev I now see render times <6ms, where they were >50ms before. It should be better in production builds, and also on faster computers.

closes #238

# Testing
Ran through the tutorial completely
I have **NOT** exhaustively checked all menu items for workingness. I'm not that far in the game, and I don't want to spoil myself.

Checked alignment
image
Note that there's extra space in prod at the top and bottom of the Hacking menu. This change removes that and makes it the same for everything.

There are a few other minor semi-intentional changes. For instance, in locations, the bar used to show up but "City" wasn't highlighted:
image
Now that is fixed.

### DIFF
--- a/src/Sidebar/ui/SidebarAccordion.tsx
+++ b/src/Sidebar/ui/SidebarAccordion.tsx
@@ -1,0 +1,90 @@
+import React, { useMemo, useState } from "react";
+import Collapse from "@mui/material/Collapse";
+import ListItem from "@mui/material/ListItem";
+import ListItemIcon from "@mui/material/ListItemIcon";
+import ListItemText from "@mui/material/ListItemText";
+import Tooltip from "@mui/material/Tooltip";
+import Typography from "@mui/material/Typography";
+import ExpandLessIcon from "@mui/icons-material/ExpandLess";
+import ExpandMoreIcon from "@mui/icons-material/ExpandMore";
+
+import { SidebarItem, ICreateProps as IItemProps } from "./SidebarItem";
+import type { Page } from "../../ui/Router";
+
+interface IProps {
+  key_: string;
+  page: Page;
+  clickPage: (page: Page) => void;
+  flash: Page | null;
+  items: (IItemProps | boolean)[];
+  icon: React.ReactElement;
+  sidebarOpen: boolean;
+  classes: any;
+}
+
+// We can't useCallback for this, because in the items map it would be
+// called a changing number of times, and hooks can't be called in loops. So
+// we set up this explicit cache of function objects instead.
+// This is at module scope, because it's fine for all Accordions to share the
+// same cache.
+// WeakMap prevents memory leaks. We won't drop slices of the cache too soon,
+// because the fn keys are themselves memoized elsewhere, which keeps them
+// alive and thus keeps the WeakMap entries alive.
+const clickFnCache = new WeakMap();
+function getClickFn(toWrap: (page: Page) => void, page: Page) {
+  let first = clickFnCache.get(toWrap);
+  if (first === undefined) {
+    first = {};
+    clickFnCache.set(toWrap, first);
+  }
+  // Short-circuit: Avoid assign/eval of function on found
+  return (first[page] ??= () => toWrap(page));
+}
+
+// This can't be usefully memoized, because props.items is a new array every time.
+export function SidebarAccordion(props: IProps): React.ReactElement {
+  const [open, setOpen] = useState(true);
+  // Obnoxious, because we can't modify props at all.
+  const li_classes = useMemo(() => ({ root: props.classes.listitem }), [props.classes.listitem]);
+  const icon = Object.assign({}, props.icon);
+  icon.props = Object.assign({ color: "primary" }, icon.props);
+
+  // Explicitily useMemo() to save rerendering deep chunks of this tree.
+  // memo() can't be (easily) used on components like <List>, because the
+  // props.children array will be a different object every time.
+  return (
+    <>
+      {useMemo(
+        () => (
+          <ListItem classes={li_classes} button onClick={() => setOpen((open) => !open)}>
+            <ListItemIcon>
+              <Tooltip title={!props.sidebarOpen ? props.key_ : ""} children={icon} />
+            </ListItemIcon>
+            <ListItemText primary={<Typography>{props.key_}</Typography>} />
+            {open ? <ExpandLessIcon color="primary" /> : <ExpandMoreIcon color="primary" />}
+          </ListItem>
+        ),
+        [li_classes, props.sidebarOpen, props.key_, open, props.icon.type],
+      )}
+      <Collapse in={open} timeout="auto" unmountOnExit>
+        {props.items.map((x) => {
+          if (typeof x !== "object") return null;
+          const { key_, icon, count, active } = x;
+          return (
+            <SidebarItem
+              key={key_}
+              key_={key_}
+              icon={icon}
+              count={count}
+              active={active ?? props.page === key_}
+              clickFn={getClickFn(props.clickPage, key_)}
+              flash={props.flash === key_}
+              classes={props.classes}
+              sidebarOpen={props.sidebarOpen}
+            />
+          );
+        })}
+      </Collapse>
+    </>
+  );
+}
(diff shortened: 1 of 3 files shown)


d0sboots:
This actually should be totally ready to go, it just needs the refactor to land first for the commits to work properly

src/Sidebar/ui/SidebarItem.tsx
```diff
@@ -0,0 +1,47 @@
...
+  clickFn: () => void;
+  flash: boolean;
+  classes: any;
+  sidebarOpen: boolean;
+}
+
+export const SidebarItem = memo(function (props: IProps): React.ReactElement {
+  // Obnoxious, because we can't modify props at all.
+  const icon = Object.assign({}, props.icon);
+  icon.props = Object.assign({ color: props.flash ? "error" : !props.active ? "secondary" : "primary" }, icon.props);
```
Snarling:
```suggestion
  const icon = {
    type: props.icon.type,
    key: props.icon.key,
    props: { ...props.icon.props, color: props.flash ? "error" : props.active ? "primary" : "secondary" },
  };
```
Avoids `any` type from `Object.assign`, and should also be a more efficient way of copying the icon object with a modified color.

src/Sidebar/ui/SidebarRoot.tsx
```diff
@@ -59,6 +55,39 @@ import { InvitationsSeen } from "../../Faction/ui/FactionsRoot";
...
+const checkIcon = <CheckIcon />;
+const helpIcon = <HelpIcon />;
+const settingsIcon = <SettingsIcon />;
+const developerBoardIcon = <DeveloperBoardIcon />;
+const emojiEventsIcon = <EmojiEventsIcon />;
+const accountBoxIcon = <AccountBoxIcon />;
+const publicIcon = <PublicIcon />;
+const liveHelpIcon = <LiveHelpIcon />;
+const chevronLeftIcon = <ChevronLeftIcon color="primary" />;
+const chevronRightIcon = <ChevronRightIcon color="primary" />;
```
G4mingJon4s, line 86: Why should the root know about the icons of its children?
Wouldn't it be much better, if it instead gave something the children could use to grab it themselves?
It would stop any rerenders to re-define all of these values, but instead each child that needs to be rerendered grabs it for themselves.
One other way of doing it, would be to put each accordion in an useMemo hook and put the icon in there.
Yes, it would be a bit harder to read and changing the icons would be not as straight forward, but it would stop any unwanted rerenders, because the reference to the icon element changed.

G4mingJon4s (changes requested):
Putting each accordion in their own useMemo allows the game to not rerender everything, when it rerenders the sidebar.
At the moment, all icons will be redefined, which leads to rerenders of all children.
Defining all of these icons on each rerender is bad practice.

src/Sidebar/ui/SidebarRoot.tsx
d0sboots: Well, the root has to define the data in *some* format - it is the only component that is not "generic," i.e. that knows the concrete data to assign to each place. Using actual ReactElements is very convenient, because it allows for the root to customize the icons in arbitrary ways. This is needed for the Augmentations icon, which has a transform on it.

No changes are needed here to avoid rerenders: There basically aren't any that aren't completely necessary.
image

It took me a *long* time to fully wrap my head around how React works with children elements, in order to get this straight. Because the icons are always the same objects when being passed *in* to SidebarItem, `memo` will see this and allow them to be memoized. This means that all the rendering from SidebarItem down gets eliminated, *even though* it creates new versions of the icon elements internally.

The stuff *above* SidebarItem can't really be memoized. (It could, potentially, but only with a ton of effort and it's not really worth it.) In particular, SidebarRoot has to do various calculations and create a fresh array of items to pass to each SidebarAccordian; trying to express that dependency structure would be a really big pain either for `memo` or `useMemo`. And as you can see, the total time taken up by each `SidebarAccordian` render is now very small (although collectively, they are a big chunk of the remaining time).

src/Sidebar/ui/SidebarItem.tsx
Snarling: Actually better solution is to not send the built jsx icon as a prop at all and just send the function for creating a new icon of the given type, I will send a different suggestion later this morning that avoids this whole thing where we break down a prop icon to rebuild a clone and display the clone.

d0sboots: Done, and rebased/ready to go

d0sboots: > Actually better solution is to not send the built jsx icon as a prop at all and just send the function for creating a new icon of the given type, I will send a different suggestion later this morning that avoids this whole thing where we break down a prop icon to rebuild a clone and display the clone.

This doesn't work, unless you know of a snazzy way to package other props with it. That was my initial attempt, but it foundered on the situation with `doubleArrowIcon`.
You could just send the props I suppose, but the syntax of specifying an icon as JSX is nicer.

src/Sidebar/ui/SidebarAccordion.tsx
```diff
@@ -0,0 +1,90 @@
...
+  // Explicitily useMemo() to save rerendering deep chunks of this tree.
+  // memo() can't be (easily) used on components like <List>, because the
+  // props.children array will be a different object every time.
+  return (
+    <>
+      {useMemo(
+        () => (
+          <ListItem classes={li_classes} button onClick={() => setOpen((open) => !open)}>
+            <ListItemIcon>
+              <Tooltip title={!props.sidebarOpen ? props.key_ : ""} children={icon} />
```
Snarling, line 61:
```suggestion
  const li_classes = useMemo(() => ({ root: props.classes.listitem }), [props.classes.listitem]);

  // Explicitily useMemo() to save rerendering deep chunks of this tree.
  // memo() can't be (easily) used on components like <List>, because the
  // props.children array will be a different object every time.
  return (
    <>
      {useMemo(
        () => (
          <ListItem classes={li_classes} button onClick={() => setOpen((open) => !open)}>
            <ListItemIcon>
              <Tooltip title={!props.sidebarOpen ? props.key_ : ""}>
                <props.icon.type color={"primary"}></props.icon.type>
              </Tooltip>
```

src/Sidebar/ui/SidebarItem.tsx
```diff
@@ -0,0 +1,53 @@
...
+      className={props.active ? props.classes.active : ""}
+      onClick={props.clickFn}
+    >
+      <ListItemIcon>
+        <Badge badgeContent={(props.count ?? 0) > 0 ? props.count : undefined} color="error">
+          <Tooltip title={!props.sidebarOpen ? props.key_ : ""} children={icon} />
+        </Badge>
+      </ListItemIcon>
+      <ListItemText>
+        <Typography color={props.flash ? "error" : !props.active ? "secondary" : "primary"} children={props.key_} />
```
Snarling, line 49:
```suggestion
  const color = props.flash ? "error" : props.active ? "primary" : "secondary";
  return (
    <ListItem
      classes={{ root: props.classes.listitem }}
      button
      key={props.key_}
      className={props.active ? props.classes.active : ""}
      onClick={props.clickFn}
    >
      <ListItemIcon>
        <Badge badgeContent={(props.count ?? 0) > 0 ? props.count : undefined} color="error">
          <Tooltip title={!props.sidebarOpen ? props.key_ : ""}>
            <props.icon.type color={color} />
          </Tooltip>
        </Badge>
      </ListItemIcon>
      <ListItemText>
        <Typography color={color} children={props.key_} />
```

Snarling (changes requested):
This really isn't functionally all that different, but at least it gets rid of the manual recreation of the icon.

I did run into some type issues when trying to change the icon prop to no longer be a constructed JSX element. I'll probably mess with it some more after merging this.

PR #248, opened 2022-12-05
CORP: Major changes to material quality/product rating calculations, products' prices can be set separately in each city

Mughur:
> Reorganize and regroup bitnode multipliers

The node multiplier lists had no internal consistency, grouped them like the default node multiplier list.

> Change quality calculation of produced materials and products to take used materials' quality into account

Previously produced materials' and products' quality/rating didn't take the materials used into account at all. changed it so that Quality = MIN(quality, avg used material quality * log_10(quality))
Effective rating = MIN(rating, avg used materials * log_10(quality))

and previously stored materials/products and bought/imported materials effect the quality/effRating:
Quality = (old quality * old stored amount + new quality * produced amount + import quality * import amount + 0 * bought amount) / (old + produced + imported + bought amounts) (tho since purchasing, producing and importing are all done in different corp states, each of them change the quality sequentially instead of at once)
Rating = (old effective rating * old amount + produced rating * produced amount) / (old + produced amounts)

material production uses small (1) base value for quality of materials used to prevent a catch-22 situation (to get >0 quality materials, you'd need other >0 quality materials, but you can't get >0 quality materials without >0 quality materials)

> products' prices can be set separately in each city

Changed the product sCost from a single value per product to an array so that each city has it's own value.
issues: The sell product modal doesn't show the right things for some reason :/ and any pre-change products will keep their old single value sCost. somebody else has to figure out how to fix that.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "bitburner",
-  "version": "2.1.0",
+  "version": "2.2.0",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "bitburner",
-      "version": "2.1.0",
+      "version": "2.2.0",
       "hasInstallScript": true,
       "license": "SEE LICENSE IN license.txt",
       "dependencies": {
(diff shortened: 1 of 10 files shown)


Mughur:
I'll split this into multiple branches / PRs

PR #254, opened 2022-12-12
NETSCRIPT: Reduce cost of two multipliers functions

d0sboots:
The data for both of these is available from getPlayer() for 0.5 GB. Since both of them together only return a small subset of the data from getPlayer(), it makes sense that their combined cost should be much less.

Tested that `getHackingMultipliers` still works and costs less.


Snarling:
0.1 seems really low. Really getPlayer should be more than 0.5GB for how much data it provides but I'll leave that alone for now. I'm merging this but I'll probably raise the costs up a bit from 0.1.

PR #255, opened 2022-12-13
NETSCRIPT: Add ns.pid variable for accessing the current script's PID

jeek:
API: Add "ns.pid" to allow scripts to fetch their own pid without contributing to the script's RAM size.


Snarling:
The idea here is fine but it also needs to be added to the documentation for ns. This does cause a bunch of other type stuff to break though so I'll just go ahead and take care of that part in a separate commit.

PR #258, opened 2022-12-20
Tutorial: Remove NS1 tutorial, change getting started guide to .js

Mughur:
Time to start phasing out NS1/.script

This will steer new players to using NS2/.js from the start, making NS1/.script eventually feasibly deprecable.


doc/source/guidesandtips/gettingstartedguideforbeginnerprogrammers.rst
```diff
@@ -143,35 +144,35 @@ The :js:func:`getServerMaxMoney` Netscript function is used to find this value
...
+        } else if (ns.getServerMoneyAvailable(target) < moneyThresh) {
             // Otherwise, if the server's money is less than our threshold, grow it
-            grow(target);
+            await ns.grow(target);
         } else {
             // Otherwise, hack it
-            hack(target);
+            await ns.hack(target);
         }
     }
```
d0sboots, line 178: Suggestion: Add a paragraph here, below the one that follows. Suggested wording:

The `await` keyword is needed for `ns.hack`/`ns.grow`/`ns.weaken` because these commands take time to execute, unlike the others. If you forget to await these commands, you will get an exception saying you tried to do multiple things at once, because your code will immediately finish the function call without waiting for the operation to be done. Also important is that `await` can only be used in functions marked `async` (which main() is).

doc/source/guidesandtips/gettingstartedguideforbeginnerprogrammers.rst
```diff
@@ -401,33 +402,38 @@ Netscript functions:
...
-            //  4. Increment our iterator to indicate that we've bought a new server
-            var hostname = purchaseServer("pserv-" + i, ram);
-            scp("early-hack-template.script", hostname);
-            exec("early-hack-template.script", hostname, 3);
-            ++i;
+    /** @param {NS} ns */
+    export async function main(ns) {
+        // How much RAM each purchased server will have. In this case, it'll
+        // be 8GB.
+        var ram = 8;
```
d0sboots: Optional: You could switch the `var`s to `let`/`const` in the examples, to lead people down the proper path.

PR #260, opened 2022-12-20
NETSCRIPT: Added ns.corporation.getConstants, replacing many other corporation getter functions

Mughur:
Removed 7 "getX" functions from corp API, added "getConstants" function, which contains the information that was in the 7 "getX" functions and some additional information, like material sizes, required materials and their amounts for each industry type.

Added cityName enum to be used everywhere.

Moved material sizes to materialInfo, which include material name, size and whether it's production multiplier material.

Moved product rating weights to industry data.

Fixed a bug which allowed running multiple script instances with the same arguments.


src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -7612,6 +7597,88 @@ interface CorporationInfo {
...
+  /** Amount of acquirable shares. */
+  warehousePurchaseCost: number;
+  /** Price of the shares */
+  coffeeCost: number;
+  /** Fraction of profits issued as dividends */
+  materials: Record<string, materialInfo>;
+  /** Tax applied on your earnings as a shareholder */
+  products: Record<string, productInfo>;
+  /** Your earnings as a shareholder per second this cycle */
+  divisions: Record<string, divisionInfo>;
```
Snarling: Record<string, X> is a bad type to provide to the Player (and just in general). Even though it's a fixed object with specific keys, Record<string, X> doesn't know what keys it has and doesn't provide autocompletion assistance to the user, and doesn't know when an invalid key was provided.

I don't mind tweaking these types myself, but it will probably wreck your other gameplay PR with a bunch of merge conflicts if I merge this one and then mess around with a bunch of it afterwards. So I'd rather merge them both together so my subsequent changes won't cause an issue. Will probably just wait until after 2.2.0 full release to merge these.

Snarling: Almost all the documentation comments are referencing the incorrect data, e.g. "Price of the shares" documentation for coffeeCost.

Snarling (changes requested):

src/ScriptEditor/NetscriptDefinitions.d.ts
Mughur: hmm.. I thought I already fixed all those :/

Mughur: To be honest I don't know how to make those better :/

Snarling: No worries, I don't mind tweaking types myself, I just don't want to make it hell to rebase your other major PR that touches all the same stuff. Is the gameplay changes PR ready ish? If so I'll just merge both and mess with types and whatnot afterwards.

Mughur: yeah, they are merge-ready as far as I'm concerned

Snarling: Wait where's the gameplay branch though? Did you submit a PR for that one?

Snarling (approved):

src/ScriptEditor/NetscriptDefinitions.d.ts
Mughur: I removed it, the corp rework stuff need some work before being pushed. Maybe it'll be good enough for 2.4 or 2.3.1 :)

PR #261, opened 2022-12-21
NETSCRIPT: Added ns.corporation.issueNewShares function

Mughur:
Added API call for issuing new shares, takes in share amount as an optional parameter, defaults to max amount issuable. Returns the amount of funds generated from the action.

Uses the same (mostly copypasted) logic as the UI action.

### DIFF
--- a/src/Corporation/Actions.ts
+++ b/src/Corporation/Actions.ts
@@ -16,6 +16,7 @@ import { ResearchMap } from "./ResearchMap";
 import { isRelevantMaterial } from "./ui/Helpers";
 import { checkEnum } from "../utils/helpers/enum";
 import { CityName } from "../Locations/data/CityNames";
+import { getRandomInt } from "../utils/helpers/getRandomInt";
 
 export function NewIndustry(corporation: Corporation, industry: IndustryType, name: string): void {
   if (corporation.divisions.find(({ type }) => industry == type))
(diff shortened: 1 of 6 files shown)


src/Corporation/Actions.ts
```diff
@@ -90,6 +91,27 @@ export function IssueDividends(corporation: Corporation, rate: number): void {
   corporation.dividendRate = rate;
 }
 
+export function IssueNewShares(corporation: Corporation, amount: number, max: number): number {
```
Snarling: If this is in Corporation/Actions.ts, shouldn't the actual corporation UI use this function? It looks like this is only being used by the new netscript function. Is there a reason both are not using the same function?

src/Corporation/Actions.ts
```diff
@@ -90,6 +91,27 @@ export function IssueDividends(corporation: Corporation, rate: number): void {
...
+  if (isNaN(amount) || amount < 0 || amount > max) {
```
Snarling: Doesn't issuing new shares also have to be in increments of 10m? That doesn't seem to be enforced here.

Snarling (changes requested):

src/Corporation/Actions.ts
Mughur: aw shucks. I calculated the 10m increment amount for the max new shares only, you're absolutely right

src/Corporation/Actions.ts
Mughur: They do now

Snarling:
I'll merge this after the main gameplay affecting corp PR (you were doing a balance changes PR right?) and the corp consts ones. I foresee possible merge conflicts between the PRs and it should be easier to resolve them on this one since it's smaller.

Snarling (approved):

Issue #10, opened 2022-10-19
Various issues relating to leading slashes in file names


TheMas3212:
[Xynrati]
Should be standard support and handled the same everywhere

**Comments From Original Issue**

[rabirabirara]
> To be specific, in the documentation (https://bitburner.readthedocs.io/en/latest/basicgameplay/terminal.html?highlight=backdoor#netscript), it is noted that files not in the root directory _need_ to be referred to by absolute path.  This is a limitation that is documented as part of the game; it is not a bug per se.
> 
> However, the documentation does not mention that you *cannot* refer to files in the root directory by absolute path in arguments passed to functions like `fileExists()` or `exec()`.  This is probably the main problem (besides the inherent inconsistency of NetScript's filepath behavior) and why a few people comment on the forward-slash in file paths.
> 
> So for future users with this problem, you can do the following to standardize file paths so that they are correctly treated.  This is a NetScript 2 function.
> 
> ```js
> function standardize(path) {
>     // remove all slashes at the start of the string
>     let s = path.replace(/^\/*/g, "");
>     if (s.includes("/")) {
>         // the file is not in the root directory; prepend a single slash
>         return "/" + s;
>     } else {
>         // the file is in the root directory (no slashes in the filename that aren't in front); prepend no slashes
>         return s;
>     }
> }

[Xynrati]
> Not exactly. If you do this using wget for files in the root folder, you'll cause problems. 

[rabirabirara]
> I see. Since your original post is quite bare, I had thought you were referring to a different problem.  Could you elaborate what your issue actually is?  After all, your choice of the word "oddness" leaves a lot to the imagination.
> 
> EDIT: I see; you're a contributor, which means this issue is probably something you're referring to yourself.  My bad!  I was thinking that this was along the lines of #1634, which was closed without comment.  I was thinking I could be helpful to people with that problem, but I suppose I missed the mark on your issue.

Snarling:
This issue is being used to track all ongoing issues with leading slash inconsistency in filenames.

For all instances of players interacting with file names (scp, exec, imports, etc):
1. The file name should never include a leading slash. Resulting file names from processes that create files should never end up with a leading slash in the actual filename.
2. A leading slash in a filepath should act as if the player is referencing the root directory of the server.
3. Not using a leading slash or dots should act as if the player is referencing the current working directory (or for scripts, the same folder as that script)
[...]

PR #262, opened 2022-12-21
DOC: Add information about absolute path requirements to scp doc

lucebac:


Snarling:
I'll go ahead and merge this for now because it is correct for the moment, but the "absolute path requirement" is really a bug. The player should be able to use any standard means of specifying a file path for any function or terminal command that needs a file path. (issue #10)

PR #263, opened 2022-12-21
NETSCRIPT: strip leading slashes from filenames if file is in root

lucebac:


Hoekstraa:
Has this been tested to ensure the in-game API and RFA still function?

lucebac:
Not really, no. I barely was able to compile and test my changes because of a scuffed setup on my end. However, I do believe that I did not break anything since my change only affects `ns.scp` which, as far as I am aware of, is only used in netscript but not by the RFA or any internal functions (assuming that RFA is short for remote file API).

Snarling:
Hey sorry this PR has been in limbo for so long, I didn't merge it initially because filename consistency is really in need of a bigger overhaul, but really that's no reason this couldn't have been merged. Merging now and if we need to make big changes that override this in the future we will.

PR #266, opened 2022-12-26
CORPORATION: getCorporation() returns division names not division objects

Caldwell-74:
getCorporation() is the only way to get the names of current divisions via script
but it returns the whole division Object this makes getDivision() redundant

with this change getCorporation().divisions is an array of strings of division names without the extra information 
this will makes it easier for the player to handle the data (string[] instead of object[]) and gives getDivision() the purpose it should have


d0sboots:
(recap of discussion in #endgame-3)
I have a dissenting opinion: I think the array of division Objects is more useful than just names. Personally, I would end up reconstructing the array (or probably a map). It's a mild objection because it's not hard to reconstruct, but then again getting the names is a simple `map()` call as well.

Snarling:
Yeah with everything being on getCorporation, it just makes a getDivision function kind of pointless. The array of divisions would be more _useful_ but the player can use getDivision to get that, and we don't need to provide all of the corporation API usefulness in as few functions as possible.

PR #268, opened 2022-12-28
DOC: Clarifying that netscript ports are across servers

AdityaHegde:
I assumed ports are local to a server. Adding a clarification in the docs for other like me.


Snarling:
Hey sorry for not merging this way sooner, I was slightly irked by the weird difference in line length from the original: 
image, but it's already inconsistent through the rest of the file and it wasn't really a good reason not to merge.

PR #279, opened 2023-01-01
SLEEVE: added storedCycles to getSleeve()

### LINKED ISSUE #249: A way to find sleeve remaining cycles/bonus time
- [ ] Version of the game v2.2.0

Could we get an api function that returns the remaining cycles/bonus time a given sleeve has? Could be in sleeve skills or its own function call. Right now there is no way outside of exploits to get these values. Hopefully easy to add, but would be very useful for many.

Caldwell-74:
fixes #249 

added Bonus Time in milliseconds to getSleeve() 
added info about Bonus Time and how it is obtained to getSleeve() Description

### DIFF
--- a/src/NetscriptFunctions/Sleeve.ts
+++ b/src/NetscriptFunctions/Sleeve.ts
@@ -162,6 +162,7 @@ export function NetscriptSleeve(): InternalAPI<Sleeve> {
         shock: sl.shock,
         sync: sl.sync,
         memory: sl.memory,
+        storedCycles: sl.storedCycles,
       };
 
       return data;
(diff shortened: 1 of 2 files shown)


Snarling:
As written bonusTime would not appear in the IDE when the player types `ns.sleeve.getSleeve().` and is looking for autocompletion assistance for what is available. It also would not show up in documentation when a player looks at the return type for ns.sleeve.getSleeve. Anyone trying to use this property in a typescript environment (a lot of people write their scripts in ts and then transpile to js) will get errors.

This is because getSleeve return type is SleevePerson which does not have bonusTime as a property. bonusTime cannot be added to the SleevePerson interface because that interface is also used by the game internally (the actual game's sleeve object implements the SleevePerson interface). You won't be able to just add bonusTime as a property on SleevePerson because that interface is used by the game's internal sleeve object as well (this is to allow direct use of getSleeve and getPlayer in formulas functions that also handle the actual internal Player and Sleeve objects) and the internal Sleeve type does not have a bonusTime property either.

Caldwell-74:
so the easiest solution would be a new function to getBonusTime(sleeveNum) ?

PR #300, opened 2023-01-05
REFACTOR: Major enums

WorldMaker:
Motivation:

- Typescript enums are confusing in NetScriptDefinitions: they are both Type and Value. An ambient declaration to make the Type easily accessible confuses the user as to where to access the Values.

- Typescript enums are "deprecated" or "legacy". They predate many modern Typescript conveniences such as type unions and don't always play nicely with them and don't always provide some of the nicer typing experiences.

- Typescript string enums are especially recommended *against* as a Typescript best practice. Number enums provide some extra features in their emitted JS, but Typescript string enums are stuck emitting an old bit of JS that doesn't look much like modern JS.

- Migrating to something new can provide modern JS niceties such as `[Symbol.iterator]` so that they can be used directly in `for (of)` loops.

Implementation:

I tried to keep the boilerplate to a minimum and to keep the various enum definitions relatively similar to their definitions as Typescript enums. To do this I use a tiny bit of JS and TS metaprogramming.

I moved the existing "enum helpers" into a base class so that the updated enums would support them all directly. The base class also implements `[Symbol.iterator]()` and a reverse lookup (`getKey`)

I split these Typescript enums into separate Type and Value instances. For consistency, in the main codebase all the updated types are now singular (such as `CityName`) and all the "picker instances" are now plural (such as `CityNames`).

I did not change pluralization of ns.enums in the NetScriptDefinitions.d.ts however to keep things roughly backward compatible with existing user scripts.


WorldMaker:
I am terribly sorry that I realize this is exactly the sort of "first PR" that is frowned against by most project maintainers because it is a massive refactor that touches almost every source file. I did not mean for it to snowball quite so much, my motivation did stem primarily from trying to help NetScriptDefinitions confusion around Typescript enums. I think this work helps the rest of the code base too, and I thought it might help illustrate how by pushing this change through to its semi-automated conclusion. You'll see some nice `for (of)` loops and I think most of the code that had been using "enum helpers" looks slightly cleaner than it had. I hope it is appreciated and there is interest in merging these changes. Thanks for trying to review this massive number of small changes. (Most of them really are just pluralizing a lot of things that previously weren't plural.)

Snarling:
Alright so first, the enthusiasm is appreciated and I agree with a lot of what's behind this PR. This being a first PR isn't really an issue, but this sort of broad and opinionated change is something that should really be discussed before you start working on it, either in an issue here or the best place would be at the development channel on the discord.

I will probably be able to implement some of this stuff or at least the ideas behind them, but cannot merge the whole thing. Some things to note:

* There is essentially no time we need the keys of an enumlike object, we really only care about the values. The keys only usefulness is getting us to the associated value on this one object, there's no other same-keyed object to use the key on separately.

* There is already a plan to thin out our use of enums and enumlike objects, because in most cases the key-value syntax doesn't add anything compared to just using typed strings and using the strings directly in the code. In fact, what's usually perceived as an upside of using enums/as const objects is that it's super easy to change the name of something. However, that's something we want to avoid doing, because changing the name of something often results in breaking savegames.

* For iterating/typechecking/fuzzy matching/whatever other runtime needs, an array of typed strings is typically the simplest object to use. A lot of the codebase already uses plain typed strings and associated arrays - see Corporation/data/Constants.ts for an example.

One of the benefits of the enum helper functions (like checkEnum, findMember) were that they work equally well for actual enums, for typed string arrays, and for `as const`/enumlike objects. Moving these functions onto a class doesn't actually add any functionality, it just changes the usage to a syntax that perhaps you prefer (object.function instead of function(object)) while locking the functionality to only members of this new class.

* I see you're providing the types to the player as Interfaces. This does not provide type safety in ensuring that what we provide to the player matches the game internals 1 to 1. e.g. the most likely change we could make internally to one of these types is that we add a new item (e.g. we add a new city). If we add a new entry but forget to update NetscriptDefinition.d.ts, typescript will see no issue with that because the interface is still satisfied. One way to enforce this without enums is with declared `as const` object types, but unfortunately those do not seem to be recognized correctly by the documentation tool we use to autogenerate the .md files. That's really the only reason we're still using enums anywhere instead of `as const` objects.

Snarling:
Another thing to note is that modern object types can't really be passed to the player, because of the limitations of the ns1 interpreter. e.g.

```js
tprint(enums.CityName.Sector12);
```
prints out `undefined` in ns1

WorldMaker:
The suggestion did come out of the Discord, just not the Development channel.

> There is essentially no time we need the keys of an enumlike object

I thought this at first too, but found a usage of the Keys in a TSX file.

> There is already a plan to thin out our use of enums and enumlike objects, because in most cases the key-value syntax doesn't add anything compared to just using typed strings and using the strings directly in the code

That's what this PR does. The end result of the TS metaprogramming is that the types themselves (the singular named ones like `CityName` or `LocationName`) *are* just the simple union type of strings. You don't have to use the plural type "locators" and "helpers" anymore and things will type check just fine.

> For iterating/typechecking/fuzzy matching/whatever other runtime needs, an array of typed strings is typically the simplest object to use. A lot of the codebase already uses plain typed strings and associated arrays

I considered using this as a basis instead of `as const` objects, because I did see that pattern elsewhere, but I wanted to preserve the various enum "Helpers" in a somewhat similar state to avoid major logic changes in this PR. Including the JS/TS metaprogramming to support that style of "enum" with similar helpers as well would be easy enough to add.

> One of the benefits of the enum helper functions (like checkEnum, findMember) were that they work equally well for actual enums, for typed string arrays, and for as const/enumlike objects. Moving these functions onto a class doesn't actually add any functionality, it just changes the usage to a syntax that perhaps you prefer (object.function instead of function(object)) while locking the functionality to only members of this new class.

Those helpers made it easy to hide accidental `O(n)` behaviors. The class implements a couple of small improvements to that (several common ones move to `O(log n)`). 

The functionality isn't necessarily locked to just this class, you can duck-type interfaces for any similar class. There's an unnamed such duck type interface in one of the asserts in the PR (it just cares about the `{ has(v): v is T }`).

> I see you're providing the types to the player as Interfaces. This does not provide type safety in ensuring that what we provide to the player matches the game internals 1 to 1. e.g. the most likely change we could make internally to one of these types is that we add a new item (e.g. we add a new city). If we add a new entry but forget to update NetscriptDefinition.d.ts, typescript will see no issue with that because the interface is still satisfied.

The important types also in the NetScriptDefinitions are the singular named types which are simple string type unions. Those *do* cause type errors if a change is made that isn't reflected in both. (A lot of this PR got so big in part from fixing them.)

WorldMaker:
I had not considered NS1 and am not familiar enough with its implementation yet. One possibility would be for NS1 only just doing a `JSON.parse(JSON.stringify(enumsObject))`. It would lose the nice Iterator but still be useful for dotted property reference.

Also, the NetScriptDefinitions.d.ts move to using the string union types mean that NS1 users would no longer "have to" use the enumsObjects and could just pass strings without autocomplete problem.

WorldMaker:
It looks like a large PR with a "lot of work" that is difficult to merge, but the basics are sound and it is easy to adjust now to however the team would prefer "enums" to look moving forward. It does move all the APIs to string type unions, which gives a lot of flexibility. That was most of the effort was making sure that the string union types were solid. With that out of the way, the raw core of the "enums" can be easily shifted to anything else (ArrayLikes instead of as const objects, sure).

WorldMaker:
Aside:

> In fact, what's usually perceived as an upside of using enums/as const objects is that it's super easy to change the name of something.

I *did* notice that some of these enums were being used for pseudo-l10n which is contrary to some l10n best practices. I'm not an l10n expert by any means, but I could maybe offer some suggestions.

Snarling:
> The important types also in the NetScriptDefinitions are the singular named types which are simple string type unions. Those do cause type errors if a change is made that isn't reflected in both. (A lot of this PR got so big in part from fixing them.)

I was able to add a new CityName and LocationName to Enums.ts runtime object, and add a new location in LocationsMetadata for the added location at the added city, without any build errors even though NetscriptDefinition.d.ts was not aware of the new city/location. The new items would have been provided to the player at runtime, but they aren't part of either the hardcoded string type or the enumlike object type provided to the player.

I think we'll definitely be able to build off this, just a lot will get scrapped in converting things towards arrays (and the wrapper will probably be tweaked to be an array wrapper, or as a wrapper that can handle either arrays or enumlike objects).

> I did notice that some of these enums were being used for pseudo-l10n which is contrary to some l10n best practices. I'm not an l10n expert by any means, but I could maybe offer some suggestions.

Is that so? I'm not aware of any localization at all in the game at the moment. There might be a few inconsistent uses of localized numbers.

Snarling:
> Also, the NetScriptDefinitions.d.ts move to using the string union types mean that NS1 users would no longer "have to" use the enumsObjects and could just pass strings without autocomplete problem.

You can already use the strings, and that's the expected use. The documentation is written so that it will take an EnumMember, or any of the values of the enum member. Unfortunately, the documentation tool that builds our .md files does not expand the options out like the ingame IDE or VSCode does.

image
image

But yeah simplifying this documentation is one of the motivators for wanting to go to typed strings.

WorldMaker:
> Is that so? I'm not aware of any localization at all in the game at the moment. There might be a few inconsistent uses of localized numbers.

Lots of places across the TSX files where specific values called from the enums are used solely for names in documents/user-facing strings instead of just the raw name/words. Fake example:

```ts
return (
  <p> `A funny thing happened on the way to ${CityNames.Aevum}`</p>
)
```

It kind of makes sense if you expect the names to change at some point and generally the biggest and always most likely reason for such things to change is localization, so I see this as pseudo-localization.

What I've learned from localization best practices is that often localizers in that in many cases they would most prefer the "uncut" string because that isn't actually a variable in this case (it is fixed; it doesn't change) and they can get the best localization with the full string.

In cases where the enum is actually used as a variable, my understanding of best practices is "never leak API/model keys to user strings, always expect a translation layer". While `.toString()` seems like an easy enough translation layer, I've learned over time that often you should always start from at least a skeleton of:

```ts
function getUiString(value: SomeEnumOrLike) {
  switch (value) {
    default:
      return value.toString()
  }
}
```

Requiring that `switch` even with such a lame default in the skeleton makes it clearer from the beginning that it should handle all cases and *not* rely on the default, the default is a stopgap/worst case scenario. (And at some point if you move to actual localization the default must switch to `never` or `throw`.)

WorldMaker:
> I was able to add a new CityName and LocationName to Enums.ts runtime object, and add a new location in LocationsMetadata for the added location at the added city, without any build errors even though NetscriptDefinition.d.ts was not aware of the new city/location. The new items would have been provided to the player at runtime, but they aren't part of either the hardcoded string type or the enumlike object type provided to the player.

I could experiment with that to see if I could make it better.

My biggest suggestion though, and I didn't want to do it in this PR would be to move to more directly shared .ts files and a tsconfig with a declaration file build instead of entirely hand-managing it. Let typescript compile the final NetscriptDefinition.d.ts. It is good at that.

> I think we'll definitely be able to build off this, just a lot will get scrapped in converting things towards arrays (and the wrapper will probably be tweaked to be an array wrapper, or as a wrapper that can handle either arrays or enumlike objects).

Like I said, the bulk of this work was just splitting the string union type and the "locator"/"helper" type. Now that that is out of the way, it's a quick next change to move all of them to arrays (since the hard work is splitting the old enum types), though that will start to affect behavior and add much more potential breaking changes. Do you want me to go ahead and migrate one or two of these in this branch to see what it looks like?

Snarling:
My intent is to merge this for v2.3, which is probably over a month out, since I intend to include a bunch more type fixes in 2.3 (e.g. bladeburner has a bunch of plain old string types that need proper types added). During those changes I will probably get rid of ns.enums altogether at that point if possible, since they're not really enums if we change them to arrays.

Some of the ns enums don't need to be provided to the player as a runtime object at all, since there isn't a conceivable reason to need to iterate through them (e.g. ToastVariant). They will still need an internal object to typecheck player input against though. Things that are provided to the player I intend to rework into being part of a getConstants call on whatever API makes sense (similar to how corp runtime objects for types are provided on corporation.getConstants). So for most of the enums that would be just ns.getConstants, or for new blade ones it would be ns.bladeburner.getConstants, etc.

But if you want to do some work converting any of these to arrays, that would certainly be welcome (you can just leave them on the enums object for now). When it gets closer to merge time and I start doing serious work on types for 2.3, I will probably merge whatever's here at the time and just work from there.

> My biggest suggestion though, and I didn't want to do it in this PR would be to move to more directly shared .ts files and a tsconfig with a declaration file build instead of entirely hand-managing it. Let typescript compile the final NetscriptDefinition.d.ts. It is good at that.

Maybe something like this could work, but NetscriptDefinition.d.ts is very integrated into the dev process right now and it may be very difficult to even get something like this to build properly. NetscriptDefinition.d.ts is also the **only** thing defining the parameter types for the entire player-facing API functions, rather than other projects where maybe you would build a .d.ts file out of the types of existing functions that are type-defined where they are actually located in the code.

All API functions internally are required to take `unknown` as the type for every parameter because we don't know what the player is going to put in, whereas we never want to tell the player-facing API that a function input is `unknown` (because the player has to know what they ought to provide).

WorldMaker:
I've had some success building API declarations (and then documentation from those) where the API is still a subtly different facade from the internals. I imagine they'd share a lot of constants and stuff, but you'd probably still want a somewhat hand managed facade of interfaces for more complex stuff, but it can live "closer" to the code that acts on it because a build process can manage all the lookups across the code base. I wouldn't expect the declaration build to be the *same* build and tsconfig as the rest, I'd just expect using a few more `import` or `export { Thing } from 'shared/thing.ts'` type stuff would be very handy.

WorldMaker:
I migrated one of the easier enums for a comparison.

Snarling:
Awesome, thanks.

I've been talking with some people who use enums currently in their scripts, and I've been persuaded that there's some value in keeping the Object.member syntax around for players to use. I think walking back a lot of what I said above, we probably will not remove the object enums already provided.

Just for consistency within the corp API (and because they're not being provided as enums), may keep EmployeePositions as an array and move IndustryType to one as well. Not 100% sure there. But will definitely be merging this with most of it intact at this point. Let me know if there's still any significant work you want to do on it before I move it over to a 2.3 prep branch and mess around with it / use it for adding types for bladeburner stuff etc

WorldMaker:
@Snarling Yeah, that makes some sense. People do like the "locator" pattern even when it gets in the way sometimes.

I don't have any remaining "TODO" items in this branch, other than I've been trying to be good about keeping up with merge conflicts.

If it weren't apparent from the above discussion, this PR is entirely about: please tell me what you need help with on this stuff. If it isn't clear by now, I have some Typescript metaprogramming skills at trying to keep things "DRY" (Don't Repeat Yourself) but also try to build nice, proper types as the result of that. If you have other things you'd like me to take a look at or migrate for you related to this to try to make things better in this branch, please let me know. I'm happy to keep lending a hand on this and I think I made it clear my interest here started and ended with helping the player experience and I know I still don't know exactly what the expected ("non-player") developer experience is desired to be in this repo so do let me know if there is more that I can help with.

Snarling:
Moved over to, squashed (to make rebases easier), and rebased at https://github.com/bitburner-official/bitburner-src/tree/newEnumBranch

I know this has been waiting a long time, but I think we still need to have some further discussion regarding what actually gets implemented. While looking through this closer during the rebase, I noticed some additional cons that I hadn't considered before. I'll start a new thread soon either on the development discord or on a new PR comment thread, and I'll make sure to tag you.

PR #307, opened 2023-01-06
CORPORATION: added Checks to setAutoJobAssignment

Caldwell-74:
added two checks to setAutoJobAssignment
-> that amount is a positive Integer
-> that amount is smaller or equal to the amount of unassigned employees
the second one is already implemented in the OfficeSpace autoAssignJob but without throwing an error it just returns false 
i think the player should get alerted when the calculation of amount is wrong but if not iam fine to change it back

### DIFF
--- a/src/NetscriptFunctions/Corporation.ts
+++ b/src/NetscriptFunctions/Corporation.ts
@@ -608,7 +608,18 @@ export function NetscriptCorporation(): InternalAPI<NSCorporation> {
       const job = helpers.string(ctx, "job", _job);
 
       if (!checkEnum(EmployeePositions, job)) throw new Error(`'${job}' is not a valid job.`);
+      if (amount < 0 || !Number.isInteger(amount))
+        throw helpers.makeRuntimeErrorMsg(
+          ctx,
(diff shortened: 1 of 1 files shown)


Snarling:
Ooooh actually, this might have had a bug in it. Isn't setAutoJobAssignment used to unassign employees as well (by setting the assignment number for unassigned?). In this case the second check is going to throw errors when it shouldn't

Snarling:
Sorry can you resubmit? I didn't mean to just insta revert, I was trying to see if it would let me reopen the PR...

Mughur:
IMO setAutoJobAssignment **should** allow trying to assign more employees than are available. In my code I have the job assignments like
```
autoassign(div, city, jobs[0], floor(size/x))
autoassign(div, city, jobs[1], floor(size/x))
autoassign(div, city, jobs[2], floor(size/x))
autoassign(div, city, jobs[3], floor(size/x))
autoassign(div, city, jobs[4], size)
```
that way I **know** that all of my employees get assigned to *something* without having to do a bunch of checks before assigning

PR #312, opened 2023-01-08
INFRA: Update everything except React

d0sboots:
### This is tha big one!

React and MUI require a lot more work, they will be dealt with separately. Everything else *seems* like it was able to be updated without much trouble.

Also includes the removal of Cypress, since no one runs it and therefore it is not useful as a testing tool.

Also includes various fixes for React warnings, as preparation for the future React upgrade.

Also also includes fixups for Electron, it also now doesn't have errors on startup and puts steam_appid.txt in the correct place. (On Windows at least, I can't verify other platforms.)

### Testing
I've been playing with this myself for a while, both with serve:dev and the electron version. Everything seems startlingly stable, but there's no way I could check everything. Things I *have* checked:

- Console logs are quiet, except for some things that are unambiguously preexisting and too large to deal with here
- MathJax works
- Steam works? It doesn't throw any errors on startup, at least.
- The Electron app fully shuts down when closing it via the 'X'. (That was broken before)
- Live reloading works
- Lint/format work
- Tests still test (faster than they used to!)
- Basic stuff like travel, running scripts, etc. all works


d0sboots:
If you can retry the failed check, it's a weird error that's nothing to do with us. It lints fine locally.

Snarling:
Honestly I'd kind of like to just remove Mui at some point, but that's also an incredible amount of work.

I think as a programming game, we should expose styling in a way that's easier for players to plug into: class names that make sense and are actually re-used throughout the codebase for common ui elements, css variables for storing colors / font / line spacing / etc. Mui is like the opposite of that, where everything is obfuscated to the point where a player who wants to interact with the styling basically can't.

e.g. changing a theme should be as easy as changing the values of some css variables and that's it, but right now it requires regenerating all the css, regnerating classnames for everything on the page, and causing react to redo its entire render from top to bottom so it can assign these brand new class names to every ui element. Also as I'm sure you noticed, this makes trying to memoize something way more clunky, because even ui elements that should be stateless and unchanging from React's perspective now have to manage state because of useStyles and variable className.

Maybe some of that can be fixed by just using mui in a different way though.

d0sboots:
Good to hear your thoughts on this!

It's important to mentally separate MUI from the styling library. MUI is a component library; if we got rid of it, we would be getting rid of *all* the components we build on - `<Box>`, `<Typography>`, `<Table>`, etc. That would be a ludicrous amount of work, and for very little gain.
However, MUI builds on top of/around a styling library, which is what you're talking about. We're currently using JSS, which was the default solution for MUI4. This is awkward, because we're actually *using* MUI5. The default styling solution for MUI5 is Emotion, but it might not be the best fit for us - fortunately, it's not hard to use a different one of our choosing. TSS would be very similar to JSS, but more supported and easy to migrate to, at least.

I get you on not liking the autogenerated css classes and use of `className`. It makes me sad when looking at the DOM, too. XD However, it comes with one *very* large advantage, which is why it's adopted by so many styling libraries: You can keep your style declarations with the UI code that they're styling. The "plain CSS" approach requires you to keep all your styling in a separate file, and (in the worst case) all lumped together.
Also, you won't get away from React redoing its render from top-to-bottom. That's the inherent nature of how React works. Short of completely going outside the system and basically writing our own thing, and at that point, why are we even using React?

But, there is good news: The most annoying part of this is the inability to target items using sensible, stable names. Fortunately, this is something that we can accomplish (even today, with JSS, or with other systems), we just need to put in the effort to define those names. The example is kind of buried in the docs, but the key is to use `slotProps` to add custom `className`s. https://mui.com/material-ui/guides/interoperability/#deeper-elements

src/Locations/ui/GenericLocation.tsx
```diff
@@ -41,49 +41,47 @@ export function GenericLocation({ loc }: IProps): React.ReactElement {
...
+      return <UniversityLocation loc={loc} />;
     }
 
     if (loc.types.includes(LocationType.Casino)) {
-      content.push(<CasinoLocation />);
+      return <CasinoLocation />;
     }
 
-    return content;
+    return null;
```
koenenalexander, line 81: This may have introduced a bug for locations with multiple location types.
For example, a location that is a Tech Vendor and a Company will only have the Company component returned.

d0sboots: Thanks for the heads up, fix PR is out now.

Snarling:
@d0sboots So just to preface this, I don't actually expect that we'll move towards using regular html elements and a non-framework-based styling solution, I'm just explaining why I think it's more elegant than how we're doing it now. A bit of a rant. Before I was working on this project, I was just a player who did a lot of document manipulation stuff including custom UIs, so that's where I'm coming from - the current implementation is extremely unapproachable for players like that.

> if we got rid of it, we would be getting rid of all the components we build on - `<Box>, <Typography>, <Table>`

Are those worthwhile components? Table is just a basic html element already, Box is just literally div (unlike Paper I'm pretty sure Box doesn't even apply any styles in our project - but if it did it could easily be replaced by a single classname as could Paper), and Typography is just defining font face/color styles that should have been defined way higher up in the hierarchy and cascaded down. And it also leads to us unnecessarily nesting more elements because we don't just put text in an `li` or a `td` or `th`, we add a new Typography component inside them to make a whole new element that wasn't necessary. Also the lack of style cascading is what leads to new content added with tprintRaw or printRaw being completely unstyled which in my opinion is a terrible user experience.

We have one font for the entire page (other than monaco and the "font" for icons), but instead of defining it in one place and cascading it down to the entire page, we're defining the same font on 100s of different individual text elements. Cascading styles is something people make mistakes on a lot. So I think frameworks encourage this "non-cascade" way of doing styles because it's kind of idiotproof and much harder to get unexpected interactions. And while that approach might be great for a lot of projects because it lets developers not need to really worry about how to use css correctly, I don't think it's a good fit when we have a good idea that many of our users will be attempting to add their own content to the page and we should want that content to come with basic styles. We should also want the dom to be approachable, sanely-structured, and in accordance with the "normal" usage of web technologies like css, since the nature of the game is that many of our players are probably going to be looking around there and actually attempting to learn in some way from what they see.

Back on components, one thing a lot of people like about mui is the Button. Sure it looks cool, with the little click wave thing when you press it. But if a player wants to make their own button, even cloning an existing Button from the game won't get one that behaves the same because it relies on event handlers to actually make that visual effect work. So again it leads to a disappointing user experience for that subset of users who are ui modifiers and want to make additions that are consistent with the game.

> The "plain CSS" approach requires you to keep all your styling in a separate file, and (in the worst case) all lumped together.

Not true at all. This approach is possible with basic document functions instead of using some framework e.g. in CharacterOverview.tsx we could have something at the top level like (just example style content, probably would look way different):

```js
const styleContent = `
#overview-table{
  //I'm not actually writing out a whole stylesheet, just some examples
}
#overview-table :is(td,th){color: var(--overviewTextColor)}
#overview-hp-row{--overviewTextColor: var(--hpColor)}
#overview-money-row{--overviewTextColor: var(--moneyColor)}
#overview-hack-row{
  --overviewTextColor: var(--hackColor);
  border-bottom-width: 1;
}`
document.head.insertAdjacentHTML("beforeEnd", `<style>${styleContent}</style>`);
```

To make it less ugly, we could even make a simple oneliner utility function injectStyles that does that last line. Or to have it behave like the current implementation (where we keep the class names as variables and define styles on a per-class basis), it would be slightly more complex but honestly not much more.

With a proper stylesheet that actually uses the cascading nature of css, the html layout of the overview table could be as simple as this:
```html
<table id="overview-table">
  <tr id="overview-hp-row">
    <th>HP</th>
    <td>10 / 10</td>
    <td id="overview-hp-hook" />
  </tr>
  <tr id="overview-money-row">
    <th>Money</th>
    <td>$1.041k</td>
    <td id="overview-money-hook" />
  </tr>
  ...etc for other rows
</table>
```

> Also, you won't get away from React redoing its render from top-to-bottom. That's the inherent nature of how React works. Short of completely going outside the system and basically writing our own thing

The only reason React needs to be involved in a theme rerender is when the dom needs to be mutated, which is only true for us because of the dynamic class names. Without dynamic class names, React is not part of the rerender.

The slow part of the rerender is the actual page rerendering though, and this needs to occur for a theme change regardless of the implementation. But React does not inherently need to be involved. While removing react from the theme change workflow doesn't make the theme change faster, it does make memoization a lot easier, since dom content does not change based on theme anymore and purely-ui portions of components would now be completely constant and would never need to be rerendered.

In the example above, even the overview style sheet would not need to be regenerated for a theme change, because it uses css variables for all of the variable parts of the theme. The only thing that would need to be updated is a single, small stylesheet that holds the values of the top-level "theme" css variables (e.g. hackColor, hpColor, moneyColor, fontFamily, etc.). But again the rerender would still be just as slow, as the whole page would need to rerender (but it's not often you need to modify the theme).

The main issue with using css variables is that they don't work in `::before` and `::after` pseudoelements, and since we're using component libraries we have no control or visibility over whether those are being used or not.

d0sboots:
Sorry for being late getting back to you on this, I wasn't able to reply at the time and then I forgot about it :(

Overall, I agree with a lot of your rant. I'm not the biggest fan of React either, and I don't think it's the best fit for what we're doing.

**_However_**, I *do* think it's important to have a framework of some kind, even if it's a homebrewed framework. (But unless you really know what you are doing, stick to an existing one.) A framework imposes structure and unity on the UI: By forcing you to do things a certain way, it prunes away a lot of the "bad ways" from the tree of possible decisions XD.

Also, I don't think switching frameworks (at this point) would ever recoup the cost put into it; I've been involved in those kinds of overhauls, and they're a *ton* of work. React may be a bit of an awkward fit, but it's just not *that* bad that it would be worth sinking in the time to remove it. IMO. And more importantly, we'd need a concrete plan of what to move *to*, beyond vague ideas of "we could do this with plain HTML." After all, that's how it was before, and Hydroflame says it was a total mess. I believe him.

Snarling:
I'm not suggesting getting rid of React at all, I'm just saying there's no reason for React to be involved in a theme change. What I had above wasn't meant to be raw HTML even if it looks like it, that was meant to represent the actual JSX structure. You can use html tags as JSX tags, you don't need to just use SpecialNamedComponents. And when those SpecialNamedComponents don't add anything useful to the basic html tags they are wrapping, why not just use the html tags directly?

Theme is just controlled by css. Changing a theme should *just* involve changing out a stylesheet in the document head (not controlled by React), and should not involve any changes in the body. This causes the actual page to rerender, but the ReactDOM would have no changes, which means nothing for React to do. Right now that is still happening, but *also* all the classnames are changing, so React needs to go through and strip away all references to the old classnames and add back references to all the new classnames. The changing classnames are the only ReactDOM changes and the only thing requiring a "React rerender".

I'm pretty sure MUI doesn't require autogenerated, dynamic classnames, so React's invovement will already be removed if we just go to static classnames, because at that point there's no changes for React to make on a theme change. I do still have qualms with Mui in general though (all those extra Mui-WhateverBase etc. classnames that have no style content and just make the dom hard to read, plus stuff I've mentioned before).

As a human developer who knows how to use CSS, I would go a step further than just static classnames - all of the main structural css should in stylesheets that are static. There are only a few things that change on a theme change: colors, font, line spacing. Those would be defined as css variables in one single "dynamic" stylesheet, which would be the only thing to change on a theme change. Having css variables for theme properties also means the player can add their own content and have it automatically change colors to match a new theme if a theme change is performed (not currently possible without constantly polling for theme changes and then making changes to the DOM). It's just a much better experience to interact with as a player.

I think if we can present the game in such a way that examining it also shows a good example of suggested web design practices for human beings, then that would be great. Because of the nature of the game, we have a playerbase that includes plenty of people interested in web development, but if they examine the DOM now, the example they see is not something they should emulate. Instead they see something that doesn't use one of the main features of css (cascading), that creates an extremely unreadable dom, doesn't expose the style in a way that the player can reliably reuse, etc. A player with a budding interest in web design isn't going to be inspired when they take a look under our hood - if anything it'll be discouraging since it makes web design look alien and incomprehensible.

Alien and incomprehensible output isn't normally a problem for framework developers, since usually you don't care about the HTML/CSS output or how easy it is to hook into your classes and theme, just that your page looks the way you want it to, performs well, and that's it. But I think because of the nature of the game and how we can expect our playerbase to want to interact with the page, that stuff should uniquely matter to us.

d0sboots:
I'm not sure if you're talking about a hypothetical state - "It would be a lot nicer if the code was like this" - or are advocating for an actual change in this direction. Although I tentatively agree with you on the former, I *don't* think this is a direction we should be pushing towards.

The reason is that what you are talking about would essentially require a rewrite of every part of BitBurner's UI. It wouldn't need to be a single atomic update (thankfully), but it would still be a huge undertaking. And I don't think the goals (the few people who use DOM inspector see better classnames) are worth that kind of cost.

> A player with a budding interest in web design isn't going to be inspired when they take a look under our hood - if anything it'll be discouraging since it makes web design look alien and incomprehensible.

I actually take a very different lesson from this. I think people are *far* more likely to look at our source code than to look at the rendered DOM. As a result, I place a much higher priority on the *source code* being human and readable. In this regard, React definitely wins - it has tons of spooky magic under the covers, but the layout of stuff in JSX is pretty comprehensible.

And our use of MUI similarly works to make the source more readable: By using `<Progress>`, `<Snackbar>`, `<Drawer>` etc. we not only save ourselves the work of re-implementing all these components, but the names are much clearer as to their function than if we just threw a bunch of HTML elements into the JSX.

PR #314, opened 2023-01-09
Dist clean

### LINKED ISSUE #313: Cleanup dist/ext folder

tiziodcaio:
Fixes #313


electron/gameWindow.js
```diff
@@ -37,6 +37,7 @@ async function createWindow(killall) {
     minHeight: 400,
     webPreferences: {
       nativeWindowOpen: true,
+      nodeIntegration: true,
```
d0sboots: This is a security hole... why are you enabling it?

tiziodcaio: Yes sorry, I've seen it on stackoverflow like a fix to a problem which appeared... 😐 But now it's working without this security hole!

Snarling:
Those assets are by default fetched from a CDN at runtime, but that doesn't work for us because the game can be played offline. That's why they're bundled with the app in dist/ext.

Does monaco-editor-webpack-plugin create a fully packaged version of monaco? If so we may be able to implement that part, but we will still need to bundle MathJax.

tiziodcaio:
I'm studying the question... Because mathjax is downloading from the CDN for the mathjaxcomponent (and there's an issue opened for that...)

The monaco-editor codebase needs more analysis, because I didn't find out where it calls the cdn and why, if it's bundled it shouldn't, right?

PR #324, opened 2023-01-16
NETSCRIPT: Added cyclesWorked to certain sleeve.getTask returns, and added bladeburner.getActionSuccesses

### LINKED ISSUE #318: A way to find a sleeves current cycles worked for current task and BB contract/operation successes.
Version of the game, v2.2.1

Sort of like how bonus cycles wasn't available the current cycles worked on a sleeves current task isn't available. These could be slid into getTask() I would think. ns.sleeve.getTask(0).cyclesWorked for example feels right.

Also a way of finding current successes with BB Contracts or Operations via the API is not yet possible, this would also be a great help. *Blackops might be useful too, as a way of knowing which ones you have completed, but I don't think it is really necessary tbh.

- [x] Add Sleeve Cycles Worked for current task to API
- [x] Add BB Contract/Ops successes to API
- [x] Add BB BlackOps successes to API

Zelow79:
Add access to some values not originally available via API.
 -Sleeve crime cycles worked
 -Sleeve bladeburner work cycles worked
 -Sleeve infiltration cycles worked
 -Bladeburner action successes

The above mentioned applicable sleeve tasks could be found with ns.sleeve.getTask(sleeveNum).cyclesWorked and if the task is crime, BBwork, or Infiltration it will return how many cycles have passed since it started. If it is not doing one of those actions it will return null.

ns.bladeburner.getActionSuccesses(type,name) returns the current amount of successes you have with that action.

closes: https://github.com/bitburner-official/bitburner-src/issues/318

Here is getActionSuccesses in action:
image

### DIFF
--- a/src/Netscript/RamCostGenerator.ts
+++ b/src/Netscript/RamCostGenerator.ts
@@ -249,6 +249,7 @@ const bladeburner = {
   getActionMaxLevel: RamCostConstants.BladeburnerApiBase,
   getActionCurrentLevel: RamCostConstants.BladeburnerApiBase,
   getActionAutolevel: RamCostConstants.BladeburnerApiBase,
+  getActionSuccesses: RamCostConstants.BladeburnerApiBase,
   setActionAutolevel: RamCostConstants.BladeburnerApiBase,
   setActionLevel: RamCostConstants.BladeburnerApiBase,
   getRank: RamCostConstants.BladeburnerApiBase,
--- a/src/NetscriptFunctions/Bladeburner.ts
+++ b/src/NetscriptFunctions/Bladeburner.ts
@@ -172,6 +172,13 @@ export function NetscriptBladeburner(): InternalAPI<INetscriptBladeburner> {
       const action = getBladeburnerActionObject(ctx, type, name);
       return action.autoLevel;
     },
+    getActionSuccesses: (ctx) => (_type, _name) => {
+      const type = helpers.string(ctx, "type", _type);
+      const name = helpers.string(ctx, "name", _name);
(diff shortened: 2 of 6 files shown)


Snarling:
NetscriptDefinition.d.ts also has definitions for the sleeve task types, e.g. SleeveBladeburnerTask. For the 3 task types that have cyclesWorked, that needs to be added to the type definition, otherwise people using ts will get errors when they attempt to reference task.cyclesWorked, and it will not be provided as an autocomplete option (for all users). e.g. there should be a cyclesWorked autocomplete entry when a player does this in the ingame IDE, but there's not because the task's type definition doesn't have it:
image

Other than that, looks good to me

Snarling:
Also when a PR is going to close an issue, you can write e.g. "closes #318" in the description of your PR, and it will autoclose that issue when your PR gets merged. There are a few other keywords you can use instead of closes: close, fix, fixes, and some more.

Oh and the title check is failing because the title should be in the format `CATEGORY: Description of fix`

Zelow79:
Wait, what isn't it doing?
image

Zelow79:
Would this be the fix I need? (I only applied it to Bladerburner so far)
image

Snarling:
So the ingame API doesn't just suggest actual properties of the object for autocompletion, it also suggests all sorts of other keywords or variable names.

On that BladeburnerTask image, it looks like you probably have it fixed correctly since it is suggesting cyclesWorked as a property of your object (different icon than all the other suggestions)

Zelow79:
New PR up.

Zelow79:
Ooof, I missed a step.

PR #330, opened 2023-01-17
NETSCRIPT: Add growThreads to formulas, improve docs for hacking functions

d0sboots:
Switch usage in ns.growthAnalyze and remove numCyclesForGrowth everywhere, since it's a trap for the unwary.

Also add a hook for numCyclesForGrowthCorrected in formulas.hacking, since there was none before and it's complicated. There's no reason to force players to re-write this for themselves.

This has an observable change on ns.growthAnalyze, one that should be better for most people but *will* mess up scripts that are relying on the specific results currently returned (for instance, using it as a replacement for formulas).


Snarling:
In what way is growthAnalyze a trap? It tells you a number of threads that is sufficient to grow a server by a particular factor regardless of the current money on the server. So if a player uses the number of threads suggested by growthAnalyze, they will never be led to ruin by that because the threadcount can be used at any amount of money.

Compare to growthAnalyze in your revision: if a server is close to empty or close to full, your revision will suggest threadcounts which are far *lower* than what is actually needed in general for a certain multiplicative factor. At the extreme end, it suggests 0 threads for any factor if the server is full or empty. This lower-than-needed thread suggestion is what I would actually call a trap, as if players use this thread count they could end up with an empty server, confused at why the number of threads they were told was needed ended up not being enough. In your revision there no longer is any way to get a generalized thread count that is sufficient for a multiplicative factor at any money amount. If you fully prep a server before calculating batch thread counts with growthAnalyze, you end up unable to calculate them at all without formulas, since you're suggested to use 0 threads now (since the server is already at max).

This would break countless hacking scripts out there.

d0sboots:
It's a trap because, depending on the amount of money you're growing from, the result from `growthAnalyze` can be an *extreme* overestimate. In the limit case where the server has $0, it would take Infinity threads to grow to any non-zero value, which is clearly impossible. This leads players to think that servers **can't** reach $0, or that if they do they're unrecoverable. (I've heard both of these sentiments in #batching.)

However, I am quite sensitive to causing breakage for players, and your point about being unable to use the new `growthAnalyze` after prepping is true. So, I've changed it to leaving `growthAnalyze` alone (aside from doc changes) and just introducing new functions.

As I mention in the commit, I believe it's important to have this in the base API so that players without formulas access can still get exact results (with the trade-off that they must prep their servers, first). Also, I don't like trend of saying "just get formulas" - for something as central as hacking, which you start doing at the very beginning of the game, I think the player should have the tools to do a good job from the start (but possibly they have to jump through some extra coding hoops to make use of them).

Snarling:
@d0sboots I made some changes, can you review the changes from 0a03d4d?
The biggest change is that growthAnalyzeCorrected no longer allows specifying a hypothetical startMoney (hypothetical calculations are the realm of the formulas API).

src/NetscriptFunctions.ts
```diff
@@ -288,43 +288,45 @@ export const ns: InternalAPI<NSFull> = {
...
+      if (mult < 1 || !isFinite(mult)) {
+        throw helpers.makeRuntimeErrorMsg(ctx, `Invalid argument: multiplier must be finite and >= 1, is ${mult}.`);
+      }
+      // TODO 2.3: Add assertion function for positive integer, there are a lot of places everywhere that can use this
+      if (cores < 1 || !isFinite(cores)) {
+        throw helpers.makeRuntimeErrorMsg(ctx, `Invalid argument: cores must be finite and >= 1, is ${cores}.`);
       }
 
-      return numCycleForGrowth(server, Number(growth), cores);
+      return Math.ceil(numCycleForGrowth(server, mult, cores));
```
d0sboots: Please **don't** do this.

Not only is it a breaking change, which we'd like to avoid, but it destroys much of the utility that growthAnalyze has over growthAnalyzeCorrected. Despite the warning, most users actually *want* the full precision, and probably would prefer it if growthAnalyzeCorrected gave decimal results too. (But it can't, not at full accuracy, without being needlessly slower.)

For example, a common recommendation is to hardcode "Math.E" as the value to pass for `mult`, since this makes future math simpler. If you do this for n00dles, you get a pretty small number, well less than 50. Truncating this to an integer would kill this method; you'd have to pass a a value near infinity just to get acceptable precision back.

src/NetscriptFunctions.ts
```diff
@@ -288,43 +288,45 @@ export const ns: InternalAPI<NSFull> = {
...
     },
   growthAnalyzeCorrected:
     (ctx) =>
-    (_hostname, _targetMoney, _startMoney, _cores = 1) => {
```
d0sboots: I contemplated this myself when I added the function, but I didn't do it this way because it makes the function mostly useless.

To make effective use of these functions, the server needs to already be prepped. However, a prepped server has max money, and a server with max money will always return 0 from the new version of `growthAnalyzeCorrected`.

A compromise that I just thought of is to make the single parameter `startMoney` instead. This reflects the question you actually want to ask: How many threads are needed to re-grow the server to max? If the server is not prepped, you'll get a different answer that, with skill, you can still interpolate to the question you wanted to ask.

This also makes the function useless (only) when the server has 0 money, which seems appropriate.

If we go with this idea, I think we should change the signature of the one in `formulas` to match.

src/NetscriptFunctions/Formulas.ts
```diff
@@ -165,6 +165,7 @@ export function NetscriptFormulas(): InternalAPI<IFormulas> {
         checkFormulasAccess(ctx);
         return calculatePercentMoneyHacked(server, person);
       },
+      // TODO 2.3: Remove growPercent, add growMultiplier function. Much better name given the output.
```
d0sboots: Although that is a better name, I don't think the churn from breaking scripts is worth it.

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -3931,50 +3931,49 @@ type HackingFormulas = {
...
-   * @param player - Player info from {@link NS.getPlayer | getPlayer}
+   * @param player - Player info, typically from {@link NS.getPlayer | getPlayer}
    * @param cores - Number of cores on the computer that will execute grow.
    * @returns The calculated grow percent.
    */
   growPercent(server: Server, threads: number, player: Person, cores?: number): number;
   /**
    * Calculate how many threads it will take to grow server to targetMoney.
-   * This calculation is the same as what the game uses, but is only accurate
-   * if all the server details are correctly set, *including* moneyAvailable.
```
d0sboots: Why'd you remove this? I think it's worth calling out that `moneyAvailable`, which usually doesn't matter when calling formulas functions, is very important here.

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -4776,77 +4775,60 @@ export type NS = {
   hackAnalyzeChance(host: string): number;
 
   /**
-   * Estimate the number of grow threads needed to grow a server by a certain multiplier.
+   * Calculate the number of threads needed to multiply funds on a target host by a given multiplier.
```
d0sboots: Please leave this as "Estimate." I feel pretty strongly about this. "Calculate" implies that this gives the correct, exact answer, and it does not.

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -4776,77 +4775,60 @@ export type NS = {
...
-   *   for each thread used. In reality, the number of threads is also *added* to the money first, and this
-   *   has a large effect when the server is low or at 0$. Use {@link NS.growthAnalyzeCorrected | growthAnalyzeCorrected }
-   *   for an exact result.
-   * - The result is only valid for the server's current status. Changes in
-   *   security will change the result. Use functions in `ns.formulas` to find
-   *   results for arbitrary situations.
-   * - Calling grow N times with 1 thread will *not* be exactly the same effect as
-   *   calling it once with N threads, because of security changes.
-   * - The result will likely not be an integer.
+   * This function returns the integer number of {@link NS.grow | grow} threads needed in order to multiply the money
```
d0sboots: remove integer and/or change to decimal

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -4776,77 +4775,60 @@ export type NS = {
...
+   * available on the specified server by a given multiplier, at the server's current security level. This does not
+   * take into account the server's current money.
+   *
+   * Note that {@link NS.grow | grow} will add $1 to the host's money for each thread before applying this multiplier.
```
d0sboots: Nice rewording! I like the flow of these paragraphs better.

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -4776,77 +4775,60 @@ export type NS = {
...
-   * //For example, if you want to determine how many grow threads you need to double the amount of money on foodnstuff, you would use:
-   * var growThreads = growthAnalyze("foodnstuff", 2);
-   * //If this returns 100, then this means you need to call grow once with at most 100 threads in order to double the money.
-   * ```
-   * @example
-   * ```ts
-   * // NS2:
-   * //For example, if you want to determine how many grow threads you need to double the amount of money on foodnstuff, you would use:
+   * ```js
+   * // calculate number of grow threads needed to apply a 2x multiplier server money.
```
d0sboots: "multiplier *to*" perhaps?

d0sboots (commented):
Two things that I consider blockers/may require further discussion: The `ceil` on growthAnalyze and the signature of growthAnalyzeCorrected.

Overall, looks very nice!

src/NetscriptFunctions/Formulas.ts
Snarling: Yeah still not sure. It also has a wacky signature compared to the other hacking formulas so I'm leaning towards changing the name, the removedFunction dialog can let the player know the replacement function to use and know about the param order swap.

src/ScriptEditor/NetscriptDefinitions.d.ts
Snarling: Added something back to make that clearer

src/ScriptEditor/NetscriptDefinitions.d.ts
Snarling: It is calculating. It's calculating an exact number of threads so that the multiplicative portion of growth is exactly what the player requested.

It's not calculating the overall effective multiplier because there is an additive step and a multiplicative step to server growth, and this is just calculating an exact threadcount for the actual multiplying portion.

I redid some more of the doc, including ns.grow, to make it even more clear that growthAnalyze is calculating only for the multiplier, and that there is also an additive portion of growth that goes into the overall effective multiplier.

src/NetscriptFunctions.ts
Snarling: I didn't change this because different server starting characteristics are exactly what formulas allows players to simulate.

Snarling:
See last commit 6929131.

I added a growthAnalyzeMult function that may be able to provide what you're looking for. It allows directly calculating growth multipliers for various thread counts. Combined with knowing that thread count is added to money before the multiplier is applied (which is now documented across many different functions), the player can do their own simulations regarding different starting conditions.

To get the easy-to-use one that just plops out a thread count right away, the player will need formulas. But without formulas, this should be sufficient if the player just does a bit of work (e.g. you can use growthAnalyze to determine the maximum threads that would be needed, then use growthAnalyzeMult + math to check if any lower threadcounts would still be sufficient).

In its current state, it's hard for me to see the usefulness in growthAnalyzeCorrected (which for now I renamed to growthAnalyzeTargetMoney but can rename it back or to something else). Does it still provide enough usefulness that it should be kept in if no other changes are made to it? It does still seem to be the easiest way to calculate how many threads are needed to rescue an accidentally-emptied server.

d0sboots:
So, I feel like this is going in the wrong direction. (Aside from the doc changes - those are great!)

I started this PR to fix the issue that `growthAnalyze` is inaccurate. I'm convinced now that it can't be "fixed" - there are some legitimate uses for it as-is, and more importantly *lots* of existing scripts are relying on it working the way it does. OK.

But the meta-reason I wanted to "fix" it has to do with the confusion and difficulty people in #batching have with computing grow threads for hacking. Using `growthAnalyze` is *not* safe, because you *will* get an answer of Infinity if/when you accidentally drain the money. This means players have to manually implement `growthAnalyzeCorrected`, and that's both silly and miserable.

The proposed `growthAnalyzeMult` is redundant; I wouldn't add it at all. It's a relatively simple formula to convert between the result of `growthAnalyze` and `growthAnalyzeMult`, and since we need to keep the first anyway, there's no need for the 2nd.

This leaves `growthAnalyzeCorrected`. In its current proposed form, as `growthAnalyzeTarget`, it is largely useless, for the reasons I explained up-thread. I don't think changing it to operate on `startMoney` instead of `targetMoney` is allowing for formulas-like "what-if" questions. Rather, it lets you ask a question with a single parameter (`startMoney`) just like `growthAnalyze` lets you ask a question with a single parameter (`mult`). You are still constrained by current reality on all other fronts - `targetMoney`, `hackDifficulty`, `skills.hacking`, etc.
The truth is that `startMoney` effectively *is* `mult`: You're asking how many threads it takes to re-grow a certain percentage of lost money, to get back to where you are right now.

Snarling:
> I started this PR to fix the issue that `growthAnalyze` is inaccurate.

It's not inaccurate at all, it's just that it doesn't cover the whole picture of what growth is. Growth isn't "just" a multiplier.

When you perform a grow, there is an addition and there is a multiplication (and then it is capped to max money). GrowthAnalyze calculates *exactly* how many threads you need to obtain a certain multiplier (which is only the multiplication part).

You are I think expecting it to provide the overall total effective "multiplier" which is just finalMoney/startingMoney. That's not a multiplier though - it's just the effective overall change from the additive growth and then the actual multiplier, and since it's different for every combination of startingMoney and threadCount, it's not something that is really useful to calculate in a general sense because it would need to be recalculated every time you find yourself at a different startMoney than expected (e.g. you had a desync and accidentally emptied your server, or you now are hacking slightly more money because hacking skill went up, etc).

> But the meta-reason I wanted to "fix" it has to do with the confusion and difficulty people in #batching have with computing grow threads for hacking. Using `growthAnalyze` is _not_ safe, because you _will_ get an answer of Infinity if/when you accidentally drain the money.

And if you accidentally drain the money (not hypothetically), you would be able to use growthAnalyzeTargetMoney.

If you accidentally drain the money and then you ask growthAnalyze how many threads you need in order to get an infinite multiplier... that's just user error. growthAnalyze doesn't ask for finalMoney/startingMoney, it asks for a desired multiplier.

Obviously no multiplier is going to rescue you from $0. To know you even can rescue a server just requires knowledge of the $1/thread additive growth, which is very well documented now.

> This means players have to manually implement `growthAnalyzeCorrected`, and that's both silly and miserable.

Is it? I don't think the game needs to just hand you direct answers to every question you would ever want to ask. It's a programming game, some complex tasks actually need to be left as an exercise for the player to implement. That's why there isn't just a "give me all server names" function, for example. Instead there's a "scan" function and the player needs to do it themselves if they want that.

Reaching 0 money on a server you're targeting isn't intended. It's not ideal play, and it should not be encouraged. If the player finds themselves in this situation however, there is a rescue mechanic to make it so it's not literally impossible to recover, but the API doesn't need to provide ns.howDoIRescueTheServerIfIAccidentallyEmptyIt because that isn't even a thing that the player should want to do.

> This leaves `growthAnalyzeCorrected`. In its current proposed form, as `growthAnalyzeTarget`, it is largely useless, for the reasons I explained up-thread. I don't think changing it to operate on `startMoney` instead of `targetMoney` is allowing for formulas-like "what-if" questions. Rather, it lets you ask a question with a single parameter (`startMoney`) just like `growthAnalyze` lets you ask a question with a single parameter (`mult`). You are still constrained by current reality on all other fronts - `targetMoney`, `hackDifficulty`, `skills.hacking`, etc. The truth is that `startMoney` effectively _is_ `mult`: You're asking how many threads it takes to re-grow a certain percentage of lost money, to get back to where you are right now.

Formulas is the only thing that lets you simulate starting conditions for a server that are hypothetical. startMoney is a starting condition of a server. Other functions let you ask "what if I did this" from the current starting conditions, but they don't let you simulate different starting conditions.

d0sboots:
> When you perform a grow, there is an addition and there is a multiplication (and then it is capped to max money). GrowthAnalyze calculates exactly how many threads you need to obtain a certain multiplier (which is only the multiplication part).

I understand this, but to me it's not a relevant distinction. There is nothing useful you can *do* with "only the multiplication part," aside from using it to calculate the full thing. For any other purpose, it is irrelevant data - and if you can only use it as an intermediate result to calculate the thing you actually want, why not just get the thing you actually want?

> Reaching 0 money on a server you're targeting isn't intended. It's not ideal play, and it should not be encouraged.

Untrue. There comes a point where you get enough RAM to completely drain and re-grow servers as part of every batch. This is pretty expensive to do, but even before that point the additive nature of the grow formula starts making a significant impact on the grow threads needed.

----

I haven't seen you hanging out in #batching at all, so I'm going to assume that it's not particularly your passion (whereas it very much is to me.) Forgive me if some of what I say here is redundant, but I want to ensure we're talking from the same page:

Coming up with *some* sort of batching system is the key to making reasonable amounts of money from hacking. I'm defining "batching" pretty broadly here: It doesn't necessarily even need to align things at specific time intervals, but you will need a way to launch worker jobs for `hack`, `grow` and `weaken` (either on demand or as permanent workers) with varying numbers of threads.

One of the key challenges is figuring out the proper *number* of threads to use for each of those. An optimal system will use all the available RAM; other systems at least need a proper ratio of threads so that `hack`/`weaken`/`grow` can run in a steady state.

At the core of this is understanding the relationship between hack and grow. One of the simplest, and thus most common strategies is to pick a particular "hack %", i.e. how much money will be hacked from a server on each iteration. The number of hack threads falls out directly from this choice via `hackAnalyze` and some simple-ish math with logarithms.

However, finding the number of grow threads is *not* an easy problem. At this point, people generally take one of two approaches:
1. Just use the result of `growthAnalyze` directly. It's an overestimate, but not by much if your hack% isn't high.
2. If you're writing your code to be general purpose, you need to actually solve the problem. That means you have to reimplement `numCycleForGrowthCorrected`, since you are solving that exact problem. Except, if you've looked at that code, you know that it's 130 lines long, and most of it is comments explain the *incredibly* dense math that goes into it. It cites academic sources!
   - As a result, re-implementing it from scratch is well out of the skill level of almost everyone. What some have done is instead reimplement `calculateServerGrowth` (the inverse function) and then binary search for the answer. This works well enough, but still requires source-diving to have a chance of success. The last option is just to copy someone else's code, who's already done the work.
   - This is why I say that asking this of people is "silly and miserable." It's silly, because the function already exists in the code, we're just refusing to give access to it. It's miserable, because writing this utility function is **much, much** harder than all the other coding challenges we provide, since it requires advanced numerical techniques and/or a willingness to source-dive, which many consider borderline cheating.

I'll note that "just use formulas" is *not* an answer here. Hacking is literally the first mechanic introduced in the game, and it is often the first way to make money, *long* before you have $5b to throw at a bunch of frankly overpriced functions. You need a strategy that can work out of the gate, even if that gate is BN1.2, BN1.3, BN11, etc.

Snarling:
Yes I have a very good understanding of batching. I have not been actively playing recently so I am not active in any of the channels right now, but I was relatively active in that channel in the past. For instance I uploaded the batching reference save that was being used for a while to test batchers (and still might be), and I was one of the early adopters of the "undying workers" style of batching (using global communication in part because at the time, you couldn't wait for a port write).

> I understand this, but to me it's not a relevant distinction (re: growth being addition and multiplication). There is nothing useful you can do with "only the multiplication part," aside from using it to calculate the full thing.

You can literally use it to accurately calculate thread counts, except when you empty a server. There might be a difference of like... one thread in extremely rare instances, usually only if you're nearly emptying a server and there's a close floating point that rounds in favor of the additive bonus.

You have said you think there is a "significant" difference even before you get to the point of literally emptying a server. So let's look at a test case as an example.

BN1.1, all augs. Targeting rho-construction with a 99% hack target vs a 100% hack target.
First, obviously not a realistic scenario, since you would blow past rho as a meaningful target with all augs in bn1.1. It's just an easy to set up scenario in dev menu for targeting rho, and I figured it would be more fair to your position than using ecorp which has even more money. I think rho is the earliest target where you might potentially "have more ram than you can meaningfully use" at the already extremely-inefficient 99% hack target.
```
test.js: growthAnalyze threads for hackportion = 0.99 (multiplier 100): 394.7023694836944
test.js: Formulas growThreads for hackPortion = 0.99: 395
test.js: Formulas growThreads for hackPortion = 1: 1396
```

So at least in this case, there's literally no difference in thread count. And 99% is an extremely generous target for me to choose for your position. Even hacking 99% is almost never optimal play until this same. If you have an example in realistic gameplay where there is any significant divergence in thread count, other than emptying a server, I'd be interested to see that.

It's not too complicated to see why threadcount isn't different between the two methods. This server has a max money of around 13 billion. That means when it's hacked down to 1% of money, it's still got $130 million on it. The only inaccuracy in a calculation that only uses growthAnalyze is that instead of starting from 130_000_000 before applying the multiplier, the actual calculation when you run a 395 thread grow essentially has the server start from 130_000_395, so you end up with effectively 1.000003 as much money as you are expecting. But being off by a factor of 1.000003 is almost never going to cause threadcount to change by even 1.

When it's hacked down to 100%, now your effective starting money isn't 130 million, it's just however many threads you use (which is orders of magnitude lower than 130 million), so you need to grow the money many more orders of magnitude to make up for that.

The additive growth is just a rescue mechanic so you don't brick your save by accidentally emptying every server. It's meant to be negligible in normal gameplay, and it is. Standard guidance is that emptying a server is a bad idea (which as shown above, it remains a bad idea even through end-of-node conditions). The basic API is not written for the player that wants to make a tiny optimization in an endgame corner case of limitless ram, it's written for general use. Using just the growth multiplier for calculations is not only very accurate (see example calc above), but it's way more usable for the player as they don't need to constantly recalculate their growth based on different starting money (since the multiplier doesn't care about the starting money).

> If you're writing your code to be general purpose, you need to actually solve the problem. That means you have to reimplement numCycleForGrowthCorrected, since you are solving that exact problem. Except, if you've looked at that code, you know that it's 130 lines long, and most of it is comments explain the incredibly dense math that goes into it. It cites academic sources!

For anything other than an empty server, even this unnecessary solution is way simpler than the numCyclesForGrowthCorrected, because you can use growthAnalyze for an extremely close starting value and then iteratively check if you are able to lower the threads (usually this is one iteration because growthAnalyze is accurate for non-empty servers, or at least for servers at 1% money as shown above). It's only literally when you're emptying a server that the OCD must-account-for-adding-$200 method actually needs a more in depth solution than that (if you want to be processor-efficient).

If a player doesn't want to implement that sort of function themselves, they can also just use the formulas API like any other player who wants an omniscient look-ahead hacking calculations (e.g. how much time will a hack take after I gain a hacking level, or if I have a desync and this process launches from elevated security during this window, when will it end, etc.).

"Additive growth is something you should account for" is just not something we should communicate to the players through the API because in normal gameplay, that's bad guidance. If additive growth is ever significant in a run, it's because the player is being extremely ram inefficient. Especially with the original wording, your PR was basically saying if you're not accounting for additive growth, then you are totally lost and what are you even doing with your script.

As a counterpoint to myself, 100% hackpercent targeting does have one major advantage for batching over anything other than 100% - it's a more resilient to h/g desyncs in regular hgw or ghw batching because a second hack before grow won't take any extra money (also won't raise sec) and you'll get right back to 100% immediately on the next grow. But batching on its own is an advanced hacking strategy as far as the API is concerned (even if it's more commonplace now since a higher proportion of the playerbase are discord regulars than in the past). And batching with a target of 100% is an extremely ram-inefficient corner case that the basic API just really doesn't need to be written around.

I think what I will do here is just get rid of the growthAnalyzeMult (redundant like you said) and potentially also get rid of the growthAnalyzeTargetMoney (not very useful as-is because like other basic hacking functions you need to "prep" to the desired starting conditions, and I think bypassing that is not allowable for anything outside of Formulas API). Also instead of pointing the player to the now-removed growthAnalyzeTargetMoney function, I think the documentation should change to just reinforce to the player that they should not empty servers (but if they do, here's how the rescue mechanic works). The formulas version is great because it fits right in with the rest of formulas.

I know a lot of that is just about the opposite of the message you submitted this PR for ("growthAnalyze bad"), but ignoring additive growth is the general advise that is actually applicable for basic gameplay. The API is not written for advanced, corner-case usage, even if this usage may have become "meta" among active discord members.


> I'll note that "just use formulas" is not an answer here. Hacking is literally the first mechanic introduced in the game, and it is often the first way to make money, long before you have $5b to throw at a bunch of frankly overpriced functions. You need a strategy that can work out of the gate, even if that gate is BN1.2, BN1.3, BN11, etc.

There's a big difference between "out the gate" and "I have so much ram I don't care if I have to use 3x as much of it to get 1.01x as much hacking income."

d0sboots:
> For instance I uploaded the batching reference save that was being used for a while to test batchers (and still might be)

Oh yeah! That is still the gold standard, even if it only works as a reference for very low RAM situations.

> I think what I will do here is just get rid of the growthAnalyzeMult (redundant like you said) and potentially also get rid of the growthAnalyzeTargetMoney

I was actually going to suggest the same if we couldn't come to some agreement: Keep the things that we both agree are unambiguous wins - `growThreads` in formulas, the better docs - and cut the rest. After all, it's much easier to add things later than it is to change/remove things in the API that weren't quite right. With the numbers you posted, I'm convinced this is the right plan in general, not just as a compromise.

>> I understand this, but to me it's not a relevant distinction (re: growth being addition and multiplication). There is nothing useful you can do with "only the multiplication part," aside from using it to calculate the full thing.

> You can literally use it to accurately calculate thread counts, except when you empty a server. There might be a difference of like... one thread in extremely rare instances, usually only if you're nearly emptying a server and there's a close floating point that rounds in favor of the additive bonus.

I think this gets to the core of the argument, right here. It's still the case that `growthAnalyze` is an estimate - and using it directly is a heuristic that works for the majority of cases. But that distinction matters a lot less in practice when, as you showed, it works *so* well for *such* a wide range of cases.

----

I wanted to run my own numbers, since I think the scenario you chose is not the most favorable possible. I chose hacking `n00dles` (low money) at the start of BN9 (bad multipliers) with only SF 1.1-1.3 (semi-realistic). The unrealistic part is the hack percentage of 99%.

```
[home ~/]> run eval.js 'ns.getServer("n00dles")'
Running script with 1 thread(s), pid 10 and args: ["ns.getServer(\"n00dles\")"].
eval.js: {"contracts":[],"cpuCores":1,"ftpPortOpen":false,"hasAdminRights":false,"hostname":"n00dles","httpPortOpen":false,"ip":"45.6.2.4","isConnectedTo":false,"maxRam":4,"messages":[],"organizationName":"Noodle Bar","programs":[],"ramUsed":0,"runningScripts":[],"scripts":[],"serversOnNetwork":[],"smtpPortOpen":false,"sqlPortOpen":false,"sshPortOpen":false,"textFiles":[],"purchasedByPlayer":false,"backdoorInstalled":false,"baseDifficulty":2.5,"hackDifficulty":2.5,"minDifficulty":1,"moneyAvailable":7000,"moneyMax":17500,"numOpenPortsRequired":0,"openPortCount":0,"requiredHackingSkill":1,"serverGrowth":3000}

[home ~/]> run eval.js 'ns.growthAnalyze("n00dles", 100)'
Running script with 1 thread(s), pid 11 and args: ["ns.growthAnalyze(\"n00dles\", 100)"].
eval.js: 34.324587357968866

[home ~/]> run eval.js 'let server=ns.getServer("n00dles");server.moneyAvailable=server.moneyMax/100;ns.formulas.hacking.growThreads(server, ns.getPlayer(), server.moneyMax)'
Running script with 1 thread(s), pid 12 and args: ["let server=ns.getServer(\"n00dles\");server.moneyAvailable=server.moneyMax/100;ns.formulas.hacking.growThreads(server, ns.getPlayer(), server.moneyMax)"].
eval.js: 34
```

So in this scenario, there *is* a difference. Playing around with the numbers a bit, the difference is roughly 1.28 threads, so more than 1, which is relevant when threads are this low in general.

*However*, this is the most extreme scenario I could think of, and even here `growthAnalyze` will work, it will just be slightly less efficient. I think this supports the points you made about people who want to completely optimize being able to do something different, so better documentation should be fine to cover the "have a backup plan if you drain the server" case.

d0sboots:
I removed the 2 functions we agreed we didn't want, and rebased everything onto `dev` since there were conflicts.

Edit: That'll teach me not to run lint/format first...

PR #333, opened 2023-01-20
CORP: Revert mandatory coffeeparty

d0sboots:
Steady-state Morale/Happiness/Energy drain is not a good gameplay mechanic. It forces the player to do a tedious task for no real reason. Worse, it disproportionally punishes new players, who already have an uphill battle dealing with corp and trying to play it manually.

This goes back to the 2.1 model of steady-state morale increase. The multiplier is much lower than it was then; for best results, coffeeparty should still be used, but it no longer is required on a recurring basis.

This also fixes a bug where only revenue was being checked; now the multiplier will properly decrease when profits are negative.

### DIFF
--- a/src/Corporation/OfficeSpace.ts
+++ b/src/Corporation/OfficeSpace.ts
@@ -112,10 +112,14 @@ export class OfficeSpace {
     if (this.totalEmployees > 0) {
       /** Multiplier for employee morale/happiness/energy based on company performance */
       const perfMult = Math.pow(
-        0.999 - (corporation.funds < 0 ? 0.002 : 0) - (industry.lastCycleRevenue < 0 ? 0.002 : 0),
+        1.002 -
+          (corporation.funds < 0 ? 0.002 : 0) -
+          (industry.lastCycleRevenue < industry.lastCycleExpenses ? 0.002 : 0),
         marketCycles,
       );
-      /** Flat reduction per cycle */
+      // Flat reduction per cycle.
+      // This does not cause a noticable decrease (it's only -.001% per cycle), it only serves
+      // to make the numbers slightly different between Happiness and Morale.
       const reduction = 0.001 * marketCycles;
 
       if (this.autoCoffee) {


Snarling:
Isn't it just a non-mechanic if it just keeps itself topped upfor anyone who doesn't completely ruin their corp? Like there's no point to the mechanic at all if you never have to actually interact with it, and make the autobrew/autoparty researches useless again.

Mughur:
autobrew/autoparty researches wouldn't be useless, they never were. Sure it wouldn't have a big effect, but when you hire more employees the stats drop, causing the city to be less effective for some time. Using autobrew/autoparty or coffeeparty can help make the time that they are less effective lower (/zero)

d0sboots:
Initially I thought it was a non-mechanic too, but then I had it explained to me by the serious Corp people. :) Besides what Mughur said, it's also important for speedrunners: The initial value of morale/happiness is low, and it takes a while to rise naturally, so an initial coffeeparty is a must for speedruns. It's also a nice-to-have for regular runs, but no longer required.

More critically, **_not every mechanic is a good mechanic_**. If you feel that morale was useless before, that doesn't justify changing it so it becomes required. That's an argument that it should probably be cut entirely. In this case, it's not actually useless, but the autoparty research probably was (and will be again). IMO, they should probably just be removed (the same goes for a *lot* of research - it's strictly worse to buy than just letting your points boost quality).

Snarling:
So originally, Corp was intentionally a mechanic that could not be automated. There was no API exposed in part because the exponential nature of it could lead to extreme optimization that makes all other game mechanics obsolete.

However, as a coding game, of course players wanted an API. And some players were already making homebrew APIs for it using document manipulation. So it was decided to "give the players what they want" and give it an API. Once the playerbase "solved" corporation, this caused by far the largest game balance issue the game has ever had (at least using API provided by the game - auto infiltration is probably more broken). Prior to the API being provided, to have a successful corporation _at all_ you needed to perform a huge amount of manual interaction.

Players always want the game to be easier. Players never ask for any part of the game to be more difficult, and they always will complain if anything is made more difficult. But I think we can agree that making a game endlessly easier and easier isn't really good game design.

One thing I think about is whether something makes sense. Does it make sense that companies with employees who do hardly anything and aren't being made to pump out production end up with extremely unhappy and tired employees, whereas companies that extract a huge amount of productivity from their employees end up with super happy, energized employees automatically? Doesn't make sense to me at all. And I'm not saying everything needs to make sense (e.g. "grow"), but in the case of auto-rising employee needs it's not a matter of us needing to suspend disbelief in order for a real game mechanic to exist, instead we're bending reality in order to justify a mechanic not needing any interaction at all (and needing just a tiny bit of interaction for more optimal play).

But I think I see a way forward that would allow for eliminating the tedium for players who want to, while imposing a cost for compared to manual needs management (which could still be used in an optimized run). A new job can be created to handle employee needs ("morale officer" or something like that, maybe even just HR). As long as at least 10% of employees are assigned to this role, needs would trend up. Assigning employees to this role leads to higher overall employee wages depending on office productivity (part of maintaining morale is actually giving employees raises if they're more productive), and also of course you lose out on whatever productivity those 10% of employees would have been providing themselves in other roles. To make that 10% thing easier to manage, there would just be a checkbox for whether to use HR or not, and checking it would require at least 10% of employees unassigned. Checking it would then assign those 10% rounded up employees to the HR job, which would automatically make all stats trend upward (regardless of company productivity), and any time the player expands the office and the 10% amount rises, it would autoassign the right number of new employees to HR as well.

Or for potentially a smaller change, maybe just having idle employees have a positive effect on happiness/morale/energy, as it represents allowing your employees some downtime as opposed to squeezing every last drop of productivity out of them.

@d0sboots @Mughur thoughts?

Also @hydroflame I'd be interested to hear your take on this as well.

d0sboots:
So, I don't give a hoot about realism, and (strongly depending on the game) I don't think it particularly matters. What I think matters for game design is, "is this mechanic *fun*?"

BB is a programming incremental/idle game. That'll mean slightly different things to different people, but the main shared bits are:
1. Monkey sees beeg number -> neuron activation
2. You should be able to program to do stuff (i.e., don't rely on manual actions to gate progress)
3. You shouldn't *have* to program everything; especially on the first time interacting with a mechanic, manual experimentation is encouraged
4. The programming tasks should be *fun* (this is the big subjective one)

Right now coffeeparty is failing 3 and (I'd argue) 4. It is much too painful to do corps manually right now because of it, and automating it is also not an interesting task, but rather something you are doing because you are forced to.

Personally, I think the best "fix" would be to rip out morale entirely. I'll say it again: **Corp has too many mechanics**, it is waaay too complicated. And most of the mechanics don't matter, or matter only in trivial ways. Every one of these mechanics is a mental tax, both when playing manually *and* when automating.

So, if we keep it, I'd want it to be something that is *strictly optional*, in the sense that you can gain some advantage (speedrun faster, etc.) by managing morale, but ignoring it totally won't stop you from winning corp in a reasonable timeframe.

To that end, I like your simpler idea of just having idle employees contribute to morale the best. To make that work, it would need to be clear (probably via tooltips both on unassigned workers and the morale numbers) that this is actually happening. I'll also note that employee salaries are cheaper than throwing parties, so leaving some on break works out better from a steady-state perspective. But the exponential nature of employee hiring means that in practice, the main cost is hiring and not salary, so coffeeparty still fills an important role if you are optimizing.

Snarling:
> To that end, I like your simpler idea of just having idle employees contribute to morale the best. To make that work, it would need to be clear (probably via tooltips both on unassigned workers and the morale numbers) that this is actually happening. I'll also note that employee salaries are cheaper than throwing parties, so leaving some on break works out better from a steady-state perspective. But the exponential nature of employee hiring means that in practice, the main cost is hiring and not salary, so coffeeparty still fills an important role if you are optimizing.

Yeah I would like to add a checkbox on the office for "automatic needs management" that is only checkable if you have at least 10% of employees unassigned, and it prevents reducing the unassigned employees to anything lower than 10%. Then that would automatically make employee needs trend up at the previous rate if it's enabled. Should also have an API to enable/disable. The checkbox makes it a lot easier for casual play, no need to micromanage assignments.

Although potentially it could cause confusion in player scripts, when they see employees unassigned and try assigning them jobs and it just keeps failing because those employees are the 10%. So there are issues too.

I agree that the employee needs system in general also just sucks. In the major corp rework, it can either get completely reworked to something more interesting or maybe even removed.

I might just cave and revert it back to the nonmechanic where employees are just happy to be working for you. But corp is already way too strong of a game mechanic so the idea of "buffing" it puts a bad taste in my mouth.

d0sboots:
I do overall like the idea of the automatic needs management checkbox. I think your two ideas can be merged: Have workers on break provide morale (always), the checkbox just requires that you keep 10% unallocated, which guarantees that morale will go up. You can disable that to manage it more manually.

But I think that's a bigger change that should be worked on elsewhere. IMO we should fix the bleeding first. I don't see this fix as a "buff" at all: Changing it does *not* affect how broken corps are, you'll still be able to beat a node in a few hours (or 20 minutes with fraud/exports) just like you can now. I'd also like to fix that (or have Mughur fix that XD), but that's *also* a separate discussion.

What this changes is how *annoying* and *tedious* corps are, especially for new players who are playing manually and don't have any script automation.

Snarling:
Not thrilled about it but I'm merging this because it seems like no one else really sees employee needs as anything other than a nuisance.

PR #341, opened 2023-01-27
MISC: reduced formatReallyBigNumber trailing 0s and updated formatSkill() to just use formatReallyBigNumber

Zelow79:
Reduced `formatReallyBigNumber()` trailing 0s and updated `formatSkill()` to just use `formatReallyBigNumber()` this helps prevent whole numbers like `1` returning as `1.000`. This change also prevents numbers like `1.040` returning when formatted with `formatReallyBigNumber()`.

Examples:
image
image
image

### DIFF
--- a/src/ui/numeralFormat.ts
+++ b/src/ui/numeralFormat.ts
@@ -62,14 +62,16 @@ class NumeralFormatter {
     if (n === Infinity) return "∞";
     for (let i = 0; i < extraFormats.length; i++) {
       if (extraFormats[i] <= nAbs && nAbs < extraFormats[i] * 1000) {
-        return this.format((n as number) / extraFormats[i], "0." + "0".repeat(decimalPlaces)) + extraNotations[i];
+        return (
+          this.format((n as number) / extraFormats[i], "0." + "[" + "0".repeat(decimalPlaces) + "]") + extraNotations[i]
+        );
       }
     }
     if (nAbs < 1000) {
-      return this.format(n, "0." + "0".repeat(decimalPlaces));
+      return this.format(n, "0." + "[" + "0".repeat(decimalPlaces) + "]");
     }
-    const str = this.format(n, "0." + "0".repeat(decimalPlaces) + "a");
-    if (str === "NaNt") return this.format(n, "0." + " ".repeat(decimalPlaces) + "e+0");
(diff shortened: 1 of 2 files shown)


Zelow79:
Hmm. Not really sure what to say about this. Does that test really need to force "987.000"? Obviously the whole point of this PR is to reduce redundant 0s is there no way around this?

Zelow79:
Updated with expected values for nFormat.test.ts

Zelow79:
Missed a value in nFormat.test.ts and corrected it.

Snarling:
Yeah updating the test is the correct way to deal with an expected test failure change.

I'm good with this personally, but I could see some people being bothered about how "jumpy" it could sometimes be if a value is changing between some decimals and none. E.g. for `X/Y hashes`, if you're capped out on hashes, it'll show a relatively short display string for both values, but then you spend a few and get a tiny bit of accumulation and X becomes a much longer string and pushes Y to the right, and then X shrinks again when you cap out again.

I'll hold off on merging this until we get a couple other opinions on it. But like I said I'm fine with it.

Zelow79:
No worries either way, I knew this one could be controversial, however I will point out it is all UI and shouldn't interfere with any scripts. I assume you know this, but just thought I would mention to be safe. I admit there is some UI cases where having a uniform x.xxx value is nice even if that value is something like `0.000` however, seeing things like 1.000 urks my OCD too lol. Just thought I would give it a shot and if it passes, then cool. If not, all good too. Thanks for looking it over so soon.

Snarling:
Can you check out the nFormat branch (https://github.com/bitburner-official/bitburner-src/tree/nFormat) and mess around with it to see if you have any major concerns with the way different types of values are displayed?

Basically going even further beyond the changes you have here, it's normalizing basically all number displays (other than ram, that's still todo) to use the same formatter (just with different options for certain things, e.g. integers never show decimals for numbers <1000). Should be more consistent just using one way of formatting, and it's done in a way that's Locale aware.

Options/Interface has a new option for showing trailing decimal zeroes (defaults to false). Should satisfy people who prefer it either way.

Snarling:
This is a built dev version of the current build, for testing

[game.zip](https://github.com/bitburner-official/bitburner-src/files/10541090/game.zip)

PR #342, opened 2023-01-29
UI: Added iron man theme.

MattiYT:
Warm red with a faded gold accent.


Snarling:
Title should be in the format CATEGORY: Description of change, so for example `UI: Added Iron Man theme`
Looks like it also needs an `npm run format`

PR #348, opened 2023-01-31
MISC: added formatBBNumber() and applied to BB UI

Zelow79:
added formatBBNumber() and applied to BB UI elements to properly display digits without unnecessary trailing 0s and as to not effect formatReallyBigNumber() in the process.

Images below to show effect:
image
image
image
Before with old PR:
image
After this PR:
image


Zelow79:
Closed due to nFormat rework PR coming.

PR #354, opened 2023-02-03
UI: Added new locale-aware and configurable number formatting

Snarling:
Also totally unrelated to the main focus of the PR, this also removes dist/bitburner.d.ts and the theme images. The theme images are regenerated on a build.

Numeral had bugs and was not being actively maintained. Instead of trying to get another external package to work, I just implemented a custom formatter that supports our specific formatting (extended suffix list).

The new formatter is also aware of locale changes in Options/Interface, and several new numeric display options are included to customize how numbers are displayed:

https://user-images.githubusercontent.com/84951833/217584707-d69b483b-00c6-4968-ab6c-5451cdb0d655.mp4

Includes 3 API functions: ns.formatNumber, formatRam, and formatPercent.

nFormat is still included for now as a numeral wrapper but it is deprecated and will be removed in 2.3, which will be announced near the top of 2.2.2 changelog.


src/NetscriptFunctions.ts
```diff
@@ -1681,15 +1681,21 @@ export const ns: InternalAPI<NSFull> = {
...
+  nFormat:
+    (ctx) =>
+    (_n, format = {}) => {
+      const n = helpers.number(ctx, "n", _n);
+      // Provide legacy
+      if (typeof format !== "object") {
+        throw helpers.makeRuntimeErrorMsg(
+          ctx,
+          "Incorrect usage of nFormat. The formatOptions for nFormat is no longer a string.\n" +
+            "Please review the latest documentation and adjust usage accordingly.\n" +
```
d0sboots: Can we provide a direct link to the docs here?

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -1,6 +1,14 @@
 /* TODO: remove ns1-specific documentation for all functions, and just create a basic doc somewhere that says how to
  *       convert examples for use in .script files (e.g. no async/await, var instead of let/const, etc). */
 
+type NFormatOptions = {
```
d0sboots: This type needs docs.
(The names are semi-self-explanatory, but it's not completely clear what their exact function is. What, exactly, does suffixStart do? What happens when isRam is true? Can multiple of the bools be true at once? Should they actually be an enum? Etc.)

src/ui/nFormat.ts
```diff
@@ -87,6 +87,10 @@ export function nFormatNew(n: number, options: NFormatOptions = {}) {
   if (!suffixList[suffixIndex]) return exponentialFormatter.format(n).toLocaleLowerCase();
   // Suffixed form
   n /= logBase ** suffixIndex;
+  if (Math.abs(n) >= logBase - 1 + parseFloat("0." + "9".repeat(fractionalDigits)) && suffixList[suffixIndex + 1]) {
```
d0sboots: This probably deserves a comment, since I'm looking at both the code and test together, and I'm still not sure why you had to add it.

src/ScriptEditor/NetscriptDefinitions.d.ts
Snarling: Addressed in d64dfe0

src/NetscriptFunctions.ts
Snarling: Addressed in d64dfe0

src/ui/nFormat.ts
Snarling: Yeah this unfortunately was the best way I could think of to force it to use the next highest suffix if it was going to round to something that shouldn't display (e.g. 1,000.000m instead of 1.000b). But it's probably terrible for performance...

Snarling: I added a comment but I'm not marking this as resolved yet as I'd like to improve this before merging.

Zelow79:
Thought regarding the removal of `numeral.js`, why not keep it, just for `ns.nFormat()` and label nFormat as depreciated and continue forward with all other UI using your custom function? Then one day way down the road when it feels like no one is using nFormat anymore poof it. Sounds like the safest bet to me and it isn't like keeping `numeral.js` hurts anything in the meantime.

d0sboots:
I agree that that would be the most user-friendly path.

I'm sort of ambivalent on it here though, because anyone who's using `nFormat` now is not super-likely to change unless/until it's deprecated, at which point you encounter the same pain. It also means you need to introduce a new function (that will stick around forever) for the new formatting. *But*, you could call it `formatNumber`, which is a nicer name.

For those reasons, although I think it's the better path, I'm only half-thumbs-up for it. XD

Zelow79:
> I agree that that would be the most user-friendly path.
> 
> I'm sort of ambivalent on it here though, because anyone who's using `nFormat` now is not super-likely to change unless/until it's deprecated, at which point you encounter the same pain. It also means you need to introduce a new function (that will stick around forever) for the new formatting. _But_, you could call it `formatNumber`, which is a nicer name.
> 
> For those reasons, although I think it's the better path, I'm only half-thumbs-up for it. XD

Renaming it would sort of defeat the whole purpose though, `ns.nFormat()` currently exists and by keeping `numeral.js` you don't have to change anything about it. It would unfortunately leave some people dependent on it for a while, but like mentioned labeling it depreciated would give people more time to move away from it. Perhaps we could even reference `Intl` as a better alternative on the documents page as well to help people move away from it. I am also not sure what the pain is you encounter here, by leading players to using `Intl` you teach them a more proper why to format their numbers in JS while also removing a format number function all together. Was just a thought though and whatever majority thinks is best we should probably go with.

d0sboots:
I guess I didn't explain myself properly.

`ns.nFormat()` currently exists. In the current version of the PR, it is changed/repurposed to use the new numeral formatting, so there remains only one public function (named `nFormat`.)

If we went with your suggestion, then the *new* functionality needs a new function in order to be properly exposed. I was suggesting we could call this `ns.formatNumber`, which is a better name than `ns.nFormat`. This is good, because the new function will live forever - when `nFormat` is deprecated, we wouldn't be changing the name of the new function.

Zelow79:
> I guess I didn't explain myself properly.
> 
> `ns.nFormat()` currently exists. In the current version of the PR, it is changed/repurposed to use the new numeral formatting, so there remains only one public function (named `nFormat`.)
> 
> If we went with your suggestion, then the _new_ functionality needs a new function in order to be properly exposed. I was suggesting we could call this `ns.formatNumber`, which is a better name than `ns.nFormat`. This is good, because the new function will live forever - when `nFormat` is deprecated, we wouldn't be changing the name of the new function.

While true, this suggestion still introduces a potential API break and defeats the purpose imo.

d0sboots:
> While true, this suggestion still introduces a potential API break and defeats the purpose imo.

Wait why? One of us is misunderstanding the other but I no longer know who ^_^;;

Zelow79:
> > While true, this suggestion still introduces a potential API break and defeats the purpose imo.
> 
> Wait why? One of us is misunderstanding the other but I no longer know who ^_^;;

Wouldn't changing the name of the function to anything other than `nFormat()` brick peoples scripts that are currently using `nFormat()`? Unless I have the wrong assumption on what an API break means.

d0sboots:
Yes, but the entire point is that we *wouldn't* be changing `nFormat`, and *therefore* that would require introducing a new function to expose the new number formatting.

Zelow79:
> Yes, but the entire point is that we _wouldn't_ be changing `nFormat`, and _therefore_ that would require introducing a new function to expose the new number formatting.

Why? We could just tell them we use `Intl` which is JS native and let them build their own from there. I think that is probably better than too much hand holding personally. Or perhaps add it to formulas or something if they want QoL for not making their own.

d0sboots:
Have you looked at the new impl? It's pretty complicated (and requires a bunch of tests), I wouldn't wish its reimplementation on our players. The point here is to expose the same formatting function that the game uses, so you can display things the same way (for instance, the k/m/b suffixes). If you just want a generic formatting function, then yes use `Intl` or any of the other JS-native options out there.

Zelow79:
> Have you looked at the new impl? It's pretty complicated (and requires a bunch of tests), I wouldn't wish its reimplementation on our players. The point here is to expose the same formatting function that the game uses, so you can display things the same way (for instance, the k/m/b suffixes). If you just want a generic formatting function, then yes use `Intl` or any of the other JS-native options out there.

Then assuming you're just speaking of making this new formatter as `ns.formatNumber()` then yes I would agree. I guess personally I don't see much of a reason for them to try to match the format 1:1 with the UI, but if it is something others need then that could work and wont risk bricking the old `nFormat()` with invalid args. Although if it is mimicking the games UI format maybe it should have a name that matches that use? `ns.ui.format()` or `ns.uiFormat()`?

Snarling:
The point of nFormat has always been to provide the player with the same formatting the game internally uses, it's even in the current doc "This is the same function that the game itself uses to display numbers." although that's really a lie because on live it doesn't provide suffixes past t and also doesn't work around some of the random NaN display bugs that numeral has (e.g. `ns.nFormat(1e33) === "NaNt"`) which the game had special handling for. 

I do think it's fine to keep nFormat using a numeral-like library for now, and then at 2.3 we do the API break to just get rid of nFormat altogether (I'd just as soon fully remove it now but I get the aversion to API breaking on a minor version).

I was fine with nFormat but formatNumber is probably a better name overall. Does clash with the existing tFormat though.

Snarling:
@d0sboots @Zelow79 I think the remaining concerns are now addressed in 209ed93

My plan is to remove nFormat in 2.3 when an API break is more acceptable. I don't like keeping an unmaintained package in the game just to provide legacy support for bad functionality (giving player a buggy formatter)

Should formatRam be provided as a separate ns function, or does it make sense to keep it rolled in to formatNumber?

d0sboots:
formatRam: I'm good with leaving it rolled in, *if* having the boolean flag makes sense in combination with the other flags. I.e. is it really a flag, or is it an enum? (This potentially applies to the design of the underlying function.)

API break in 2.3 makes sense, gotta drop it sometime and at least that gives some players the chance to stop playing, and therefore not be mad XD

PR #357, opened 2023-02-10
UI: Revealed faction augments remaining in faction invite.

Zelow79:
Revealed faction augments remaining from faction invite. Changed text to display "No Augmentations left" when they're out, instead of "0 Augmentations left".

image

### DIFF
--- a/src/Faction/ui/FactionsRoot.tsx
+++ b/src/Faction/ui/FactionsRoot.tsx
@@ -12,20 +12,10 @@ import { Factions } from "../Factions";
 
 export const InvitationsSeen: string[] = [];
 
-const getAugsLeft = (faction: Faction): number => {
-  const augs = getFactionAugmentationsFiltered(faction);
-
-  return augs.filter((augmentation: string) => !Player.hasAugmentation(augmentation)).length;
-};
-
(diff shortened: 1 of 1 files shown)


Zelow79:
I forgot to mention, here I update the tooltip to match the rest of the game:
image

Like this:
image

Which used to look like this:
image

Snarling:
Not really a fan of the tooltip that has the same content as the text the player is hovering, really only has a point when the player squishes their screen tiny enough to not see the faction name. But it does improve it making it match the rest of the game's tooltips.

I made a couple small non-functionality changes, see adc493f

Zelow79:
> Not really a fan of the tooltip that has the same content as the text the player is hovering, really only has a point when the player squishes their screen tiny enough to not see the faction name. But it does improve it making it match the rest of the game's tooltips.
> 
> I made a couple small non-functionality changes, see [adc493f](https://github.com/bitburner-official/bitburner-src/commit/adc493fb13afffd880764f10c19d27434e43ff4d)

Those changes seem fine to me. I don't disagree about the silly tooltip displaying what you're hovering over, but that was already there before I got here lol. Could have just removed the tooltip there all together though if you felt it was necessary.

PR #361, opened 2023-02-12
CODEBASE: Combine the title and dist checks

Snarling:
Submitting this with a valid title, then changing the title to something invalid to show that a comment gets generated, then changing it back to show that the comment gets edited.

This PR combines the two checks that generate PR comments (title and dist checks) so that they can edit the comment if needed. It also removes some previous guidance to not include markdown changes, since `npm run doc` no longer flags all markdown files as changed and it no longer causes an issue with reviewing a PR.


Snarling:
No comment was generated on initial test run because both the title check and dist change check passed.

Snarling:
After changing to an invalid title, the above comment was autogenerated. It will be edited once I change the title back, here is the initial comment.

image

PR #362, opened 2023-02-12
CORP: Change/fix TA.II to use configured sell amounts

d0sboots:
It used to set prices based on current production only.


Snarling:
@Mughur Can you take a look at this? If you don't have concerns with it I think we can just merge this.

Mughur:
Looks fine, I'll have to edit a few lines in the overhaul but otherwise I see no issues.
This was discussed beforehand in discord, so I already know details/reasons of the change

PR #368, opened 2023-02-14
CORP: gameplay overhaul dont merge

Mughur:
playable here: https://mughur.github.io/bitburner-src/

currently implemented:
```
- material/product quality/effective rating depends on the quality of materials used *and* the "expected" quality/rating (which was previously the only factor)

- export and sell states switched places (@jeek jumped the gun, but that's fine)

- export amounts can use "[E/I]INV" and "[E/I]PROD" for calculating "E"xport/"I"mport city's "INV"entory/"PROD"uction of the given material

- energy is removed as material and division

- you can purchase multiples (1,5,10,50,100,max) of upgrades in the corp screen as well as through the API (the UI for this is still a bit WIP)

- bulk purchasing in unlocked from the get-go

- getProduct() is city specific

- sell price/amount can/have to be set separately on each city

- removed the input box in market-ta2

- unlocks remain visible after purchase

- setting product price to "MP" now doesn't add extras, so that TA1 is actually worth something

- capping sell share amount to 100t or rewriting the sell shares functionality, also sold share amount must be an integer

- coffee -> tea

- removed happiness as stat

- "training" position is now "intern" position. exp gained in this position is 10x the other positions (previously 5x)

- stats drop if less than 1/9 of the employees in a city are interning, approx stay stable if exactly 1/9 are interning and increasing if 1/9 is interning. stats can still be increased even if funds and profit <0, but at a reduced rate

- smart supply has 3 options: "leftovers" which acts as the "use leftovers" option did, "imports" which uses the previous cycle's imported material into account when calculating purchase amounts and "none" which buys the required materials

- rewriting corpHandbook.lit (partially done, will be redone at the end of the rework process)

- added corp restart button to the corp overview. allows restarting the corp without restarting the node or using dev menu

- nerfed investors and increased valuation period (ya'll knew this was coming)

- nerfed advertising and wilson

- corp can have up to total of 20 divisions (BN multiplier can change this) of any type, meaning you could have 20 agriculture divisions if you so desire.

- divisions can be sold for (cost of starting division + office openings + warehouse purchases)/2

- player cannot sell all their corp shares

- added bunch of division types, changed materials production/consumed of some types
```

on the list:
```
- tinkering with numbers
  - cost of things
  - scaling of things, most importantly advert/awareness/popularity/advertising multiplier

- tie ins with other mechanics

- sleeve/player jobs
```


Snarling:
Already merged via #428

PR #369, opened 2023-02-14
CODEBASE: Bundle monaco dynamically

### LINKED ISSUE #313: Cleanup dist/ext folder

Snarling:
Fixes #313 to the extent we currently can. For now MathJax remains statically packaged in dist/ext.

* Updated monaco-editor
* use monaco-editor-webpack-plugin.
* monaco-vim also moved out of dist.
* Fixed a monaco-vim bug caused by one of the changes above. I only very briefly tested vim functionality due to it being vim.
* Also changed location of some generated files, and had to change .gitignore.
* Removed existing package we were using for Editor React component, due to a version issue. Made a replacement component.
* Bumped game version and started 2.3 dev changelog.


Snarling:
The test is failing due to monaco-editor being installed as a dev package, even though this doesn't cause the regular build process to fail. Discrepancy is probably due to the tests not being aware of the webpack configs for using monaco-editor-webpack-plugin. Not sure best way to resolve, maybe would have no impact just making it a non-dev package.

I thought the dist change test would fail, clearly the logic for detecting dist changes in that test needs some work.

PR #371, opened 2023-02-17
NETSCRIPT: Add `additionalMsec` to BasicHGWOptions

d0sboots:
This option adds additional time to the sleep in hack/grow/weaken before the command takes effect. The critical difference between this and doing your own sleep is that it creates a single, uninterruptible sleep: This opens up multiple new avenues of gameplay for batching.

Note that use of this new feature is theoretically always suboptimal, since extending the sleep time wastes RAM.

----

It's time to finally send this out. I don't have an example script (writing one was what held this up for so long), but I've tested it and it works.

A strategy that this enables is what I'm calling "Ultra-protobatching." It works as follows:
1. Calculate number of batches, number of threads, etc.
2. Launch *all* the batches simultaneously. Use `additionalMsec` to cause them to all land at the same time. Round-robin the launching so that they go H,G,W,H,G,W or H,W1,G,W2,H,W1,G,W2 depending on your preferred pattern.
3. It will actually take some amount of time to launch everything, but that's OK. The execution of all the batches will be smeared across this time window, which will probably only be 10-20ms. Despite this, the nature of the timing guarantees that everything will land in order, with no possibility of desyncs.
4. Once all the batches have landed, go to step 2.

This is substantially simpler than traditional batching - you still need to pick a good target, calculate threads and number of batches, etc., but all the worrying about timing and desyncs is gone. It's not *optimal* by a long shot - a lot of RAM is being wasted by sleeping for extra time in `hack` and `grow`, but it's more than adequate for most people to get by.

I view this as a good thing - I think batching is too arcane and inaccessible. Others disagree.

Another aspect is that this levels the playing field substantially for batchers - right now, batching is the only part of the game where the real-life performance of your computer matters significantly. Having a faster computer means you can run batches with tighter timing, both because you can afford to run more and because you'll have lower risk of timing mishaps. This leads to more profit.
By being able to take the timing tolerance largely out of the picture, slower machines gain more than fast machines.

This also opens up exciting new avenues of exploration at the high end, in the area of optimal batching research. I've given it a fair amount of thought myself, and I *still* don't know what the best algorithm will be, but I know that this allows for a lot of stuff that couldn't be done before.


Snarling:
I think this is a terrible change, as it trivializes almost all of the complexity of batching.

Removing the entire timing constraint means the player will never be forced to target inefficient hack% due to not being able to squish enough batches close together. At extremely high ram, the best solution is just going to be exactly what you posted, which is just the natural evolution of a collision wall batcher when you remove the constraint of ramp-up time and intra-batch spacing.

If the player does put in the extra effort to calculate everything out and avoid collisions and pack their function completions as close as feasible, etc. at high ram they will just be rewarded with less than you can make with your batcher that just asks the game "hey can you please make these end in the correct order? thanks". At lower ram an omniscient batcher that actually takes on the challenges of batching (timing, collisions, etc) will outperform that due to not including sleep-inside-grow/hack. But once that batcher has to target higher ram%, guaranteeing correct execution order and continuing to execute efficient batches at infinitesimal spacing is going to win out for approximately zero effort (and performance will continue to be very much CPU dependent, maybe even moreso than now, as you try to pack more and more scripts together)

d0sboots:
If you've crossed the threshold where packing in zero-spaced batches is worth the RAM hit, then you're at the point where you can't effectively use your existing RAM - i.e. you're nearing max RAM on the personal servers, and you're well past the point where you should have won the node. At that point, you're just doing it for kicks, so I'm not overly concerned about the optimal strategy past that point.

d0sboots:
I'll note that this function can also be used to make traditional batching extremely reliable. This means that you can use traditional batcher logic but with very short timings, down to 1ms potentially. This significantly raises the floor on how much RAM you need before the "just pack them in with 0 delay" approach becomes best.

PR #375, opened 2023-02-19
MISC: Sleeve Int applies to active/passive shock recovery

Tyasuh:
Minor change to the shock recovery formulas to allow a small buff from a sleeve's intelligence.

### DIFF
--- a/src/PersonObjects/Sleeve/Sleeve.ts
+++ b/src/PersonObjects/Sleeve/Sleeve.ts
@@ -172,7 +172,10 @@ export class Sleeve extends Person implements SleevePerson {
     this.storedCycles += numCycles;
     if (this.storedCycles < CyclesPerSecond || !this.currentWork) return;
     const cyclesUsed = Math.min(this.storedCycles, 15);
-    this.shock = this.shock - 0.0001 * calculateIntelligenceBonus(this.skills.intelligence, 0.75) * cyclesUsed;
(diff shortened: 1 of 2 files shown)


src/PersonObjects/Sleeve/Sleeve.ts
```diff
@@ -171,7 +172,7 @@ export class Sleeve extends Person implements SleevePerson {
     this.storedCycles += numCycles;
     if (this.storedCycles < CyclesPerSecond || !this.currentWork) return;
     const cyclesUsed = Math.min(this.storedCycles, 15);
-    this.shock = Math.max(0, this.shock - 0.0001 * cyclesUsed);
+    this.shock = this.shock - 0.0001 * calculateIntelligenceBonus(this.skills.intelligence, 0.75) * cyclesUsed;
```
Snarling: The `Math.max(0, value)` here was preventing shock from going negative, it's still needed.
image

Snarling (changes requested):

src/PersonObjects/Sleeve/Sleeve.ts
Tyasuh: Thank you for pointing that one out, and my bad on that, much appreciated.

Snarling (approved):
Will be merged once we push a full release for 2.2.2

PR #376, opened 2023-02-20
BN9: QoL improvements.

Zelow79:
`Bn9` now starts with the same node that `bn9.3` rewards (the idea here is, it helps showcase the BNs new feature and just speeds up a lot of what made it so slow early on).


Mughur:
I think adding the free node is a good idea, but I'm not so sure about increasing the hash production and WD difficulty.
Also using pi just feels out of place :/

Zelow79:
> I think adding the free node is a good idea, but I'm not so sure about increasing the hash production and WD difficulty. Also using pi just feels out of place :/

Updated to return BN values to original values.

PR #385, opened 2023-02-24
SOURCEFILE: Fix calculation of decMult

d0sboots:
This change was prompted because someone was confused by the discrepancy on Discord.

----

The decrease was being calculated via subtraction instead of reciprocal. This was an artifical buff to things that improved through decrease, since they decreased "more" than they should.

Since this only affected hacknet server-related costs, and those are rather weak anyway, I buffed the bonus from BN9 to compensate. The overall effect on decrease is now rougly the same if you have BN1.3 + BN9.3: (1 - .28) * (1 - .14) = .6192 ~= .6104 ~= 1 / (1.28 * 1.28)

The overall effect is a buff for BN9 hacknet server production, and more rational formulas.

## Testing

Dev menu (after much fiddling) agreed with the expected numbers

### DIFF
--- a/src/BitNode/BitNode.tsx
+++ b/src/BitNode/BitNode.tsx
@@ -322,7 +322,7 @@ export function initBitNodes() {
         Augmentations)
         <br />
         <br />
-        This Source-File also increases your hacknet multipliers by:
+        This Source-File also increases hacknet production and reduces hacknet costs by:
         <br />
         Level 1: 16%
         <br />
--- a/src/BitNode/BitNode.tsx
+++ b/src/BitNode/BitNode.tsx
@@ -119,11 +119,11 @@ export function initBitNodes() {
         BitNodes will disable this mechanic) and level 3 permanently unlocks the full API. This Source-File also
         increases your charisma and company salary multipliers by:
         <br />
-        Level 1: 16%
+        Level 1: 8%
         <br />
-        Level 2: 24%
+        Level 2: 12%
         <br />
-        Level 3: 28%
+        Level 3: 14%
       </>
     ),
   );
@@ -324,11 +324,11 @@ export function initBitNodes() {
         <br />
         This Source-File also increases your hacknet multipliers by:
         <br />
-        Level 1: 8%
+        Level 1: 16%
         <br />
-        Level 2: 12%
+        Level 2: 24%
         <br />
-        Level 3: 14%
+        Level 3: 28%
       </>
     ),
   );
--- a/src/BitNode/BitNode.tsx
+++ b/src/BitNode/BitNode.tsx
@@ -119,11 +119,11 @@ export function initBitNodes() {
         BitNodes will disable this mechanic) and level 3 permanently unlocks the full API. This Source-File also
         increases your charisma and company salary multipliers by:
         <br />
-        Level 1: 8%
+        Level 1: 16%
         <br />
-        Level 2: 12%
+        Level 2: 24%
         <br />
-        Level 3: 14%
+        Level 3: 28%
       </>
     ),
   );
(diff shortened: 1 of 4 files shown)


Snarling:
> The overall effect is a buff for BN9 hacknet server production, and more rational formulas.

Did hacknet need a buff? I know as a player it feels nicer to have something buffed, but that's all that ever gets pushed through, and it just makes the game easier and easier ad infinitum which is not healthy for the game. BN9 already just got a huge buff merged in (during BN9 you get the SF9.3 server).

Also it's up in the air which way makes sense for it to work. I agree that the reciprocal version is the same effective power level as the listed % increase, but neither way actually matches up with the description "increases your multipliers by X" so really it doesn't need to work one way or the other. One way preserves the listed % being relevant (e.g. 16% increase turns into 16% decrease) the other way preserves the same effective mathematical worth of the bonus.

d0sboots:
I was going based on the perception that hacknet is still a bit weak, as a mechanic. Maybe this perception is wrong.

I'm not sure what you mean by "neither way actually matches the description." We had this conversation in the channel already, but: If something says "Increases your multipliers by 100%", it would make no sense for it to make increasing things be 100% + 100% = 200% and decreasing things be 100% - 100% = 0%, i.e. free. That would be a 2x buff for increased things and an *infinite* buff for decreased things.

Any time something is a buff with a multiplicative effect, cost-based things need to be divided by the same amount. That's how it works everywhere else in the game (BN12 for instance), here is the only place where it got screwed up.

That said, I'm not super attached to this, I just felt moved to try and fix it because it confused someone else.

Snarling:
> I'm not sure what you mean by "neither way actually matches the description." We had this conversation in the channel already, but: If something says "Increases your multipliers by 100%",

What I mean is that "increases your multipliers by X" does not imply anything about how it works with regards to multipliers that actually get a decrease and not an increase. Do the multipliers that actually get a decrease get the same percentage change (current implementation), or do they get a decrease that gives the player the same mathematical equivalent bonus as the increasing multipliers (your PR implementation)? Neither is directly implied and both interpretations are just assumptions about how it ought to work.

But you're right, the mathematical-equivalent way is how everything else already works. For BN1 the change probably just makes sense, but for BN9 I think the wording still sucks. "Increases your hacknet-related multipliers by XX%" is not great wording when there's literally one hacknet-related multiplier that actually gets increased and 4 mutlipliers that instead get a decrease. It could probably be changed to something to the effect of "Increases hacknet production and reduces hacknet costs by:" and keep the existing functionality.

src/BitNode/BitNode.tsx
```diff
@@ -119,11 +119,11 @@ export function initBitNodes() {
         BitNodes will disable this mechanic) and level 3 permanently unlocks the full API. This Source-File also
         increases your charisma and company salary multipliers by:
         <br />
-        Level 1: 8%
+        Level 1: 16%
```
Snarling: Wrong node (this is BN3)

d0sboots: whoops

Snarling:
Lowered buff to 12%
SF1 still applies the effective reduction to grant the same effective "worth" as the increasing multiplier.
Wording of SF9 now says it's an X% reduction, so the previous functionality was correct for that (21% reduction is 0.79 multiplier, not 1/1.21).
The result is still an overall slight buff with SF1.3 + SF9.3:

Pre-PR:
Hacknet prod: 1.28 from SF1.3, 1.14 from SF9.3 = 1.4592
Hacknet costs: 0.72 from SF1.3, 0.86 from SF9.3 = 0.6192

After PR:
Hacknet prod: 1.28 from SF1.3, 1.21 from SF9.3 = 1.5488 (1.061 multiplier compared to before PR)
Hacknet costs: 0.78125 from SF1.3, 0.79 from SF9.3 = 0.6171875 (0.9967 compared to before PR)

d0sboots:
I think the final result, where one of these is dividing and one is subtracting, is worse/more confusing than what we started with. Oh well.

PR #388, opened 2023-02-24
DOC: fix typo

Mughur:
Unfortunately the checks don't yell at you for making typos in the docs


Snarling:
Header info and example on formatNumber still say formatRam
image

PR #391, opened 2023-02-24
TERMINAL: Fixed speechmarks not making numeric terminal arguments into strings

### LINKED ISSUE #349: In terminal arguments, numbers in quotes are interpreted as numbers instead of strings.
Per the documentation, wrapping a number in quotes is supposed to treat it as a string instead of a number, but they are converted to numbers.

Simple one line code to test: 
/** @param {NS} ns */
	ns.tprint(typeof ns.args[0]);
}

Running that script with the first argument as "5" for example, outputs "number".

On Bitburber v2.2.1

LiamGeorge1999:
fixes #349

tested in local hosted instance via chrome:

```/** @param {NS} ns */
  for (let arg of ns.args){
		ns.tprint(typeof arg);
  }
}
```
image

image

### DIFF
--- a/src/Terminal/Parser.ts
+++ b/src/Terminal/Parser.ts
@@ -1,8 +1,9 @@
 import { KEY } from "../utils/helpers/keyCodes";
 import { substituteAliases } from "../Alias";
 // Helper function to parse individual arguments into number/boolean/string as appropriate
-function parseArg(arg: string): string | number | boolean {
+function parseArg(arg: string, stringOverride: boolean): string | number | boolean {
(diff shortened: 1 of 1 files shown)


Snarling (changes requested):
Need to run `npm run format` since this is failing the prettier check. It'll probably just remove this one extra bit of whitespace that made it in 
image.

You don't need to manually find formatting issues like this, just running `npm run format` before committing should fix all formatting issues.

Other than that, looks great!

LiamGeorge1999:
ran prettier, vsc insisted it was an empty commit but low and behold, that space has disappeared.

Snarling (approved):

PR #399, opened 2023-02-27
BUGFIX: Memory Leak in NetscriptPorts

jeek:
Prior to making this change, a script that uses a non-trivial amount of Netscript Ports once each and then never uses them again would cause the empty ports to hang around, taking up a growing amount of memory. This patch deallocates empty ports from the NetscriptPorts map.

Tested by running my script that uses a non-trivial number of ports once each and watching the Memory tab in the debug console. Prior to the patch, it would grow until game crashed, at a rate of about 2MB/s. After the patch, memory usage grew to about 300MB and stayed there.


Snarling:
@jeek When you get a chance, can you test the new changes I just pushed affect your memory usage with high ports usage? If this doesn't work satisfactorily I have one more idea that may work without breaking portHandle references.

PR #406, opened 2023-03-03
FIX: Quick followup to #404

Tyasuh:
Many apologies if this steps on anyone's toes. But here's a quick bit that was missed in the following 
https://github.com/bitburner-official/bitburner-src/pull/404


Snarling:
Thanks, got a little sloppy there.

PR #417, opened 2023-03-10
Adding `factionReputationGain` to `GangGenInfo` interface

Perlkonig:
# Addition of `factionReputationGain` to `GangGenInfo` interface

CATEGORY: API

After unlocking the gang mechanic, I discovered that knowing the rate of faction reputation gain mattered to how my singularity script decided what action to take next. Instead of just calculating it myself, especially given that it's dependent on an internal constant, I thought adding it to the API might be appropriate.

# Documentation

When running `npm run doc`, it ended up altering a *ton* of files. It did correctly create the new MD file for the new property, but I didn't commit all those changes because (a) it would clutter an otherwise simple PR and (b) I assume the documentation will be regenerated as a normal part of release. Let me know if you want me to submit the updated MD files.

# Testing

I loaded a save game with an active gang. Then I did a simple dump of `GangGenInfo`, resulting in the following:

GangGenInfo

Then I opened up the faction screen and checked that faction was accruing as expected:

Reputation1

Reputation2


Snarling:
Since this is not something already visible ingame on the UI, it doesn't make the most sense to have it provided directly on the API.

It is a useful thing to calculate, but that does not always mean it's something that should be provided directly to the player. The issue here is that the relevant constant for calculating this is not currently provided to the player by the API. There are already plans for including more robust constants to the player in the 2.3 update, so that players can calculate things like this without needing to refer to source code.

So probably will not merge this, but we will be providing a way to get all of the gang constants (probably `ns.gang.getConstants()`) so that players can implement something like this on their own more easily.

Another way you can check faction per second is with a polling method instead of a calculation method, e.g. check your faction and then check it again 10s later. But we'll still be providing the constant so players can implement a calculation method easier.

> When running npm run doc, it ended up altering a ton of files.

The documentation build should add a bunch of updates for every single file, but then the git add command that occurs right after that (inside of npm run doc) should remove all the unnecessary/unchanged files from the commit since it should fix the line endings. `npm run doc` updating every single .md file was an issue in the past but it should be fixed now. e.g. 057aeae was a commit done using `npm run doc` and only includes changed files.

Would you mind just doing an `npm run doc` and committing to this branch? If it is including changes for every .md file, I need to figure out why and correct that, since the workflow now is supposed to be that branches that affect markdown should include the markdown changes.

Perlkonig:
Thanks for the reply and rationale. 

Pushed the doc change. You're right. It was much cleaner than it first looked. Now I know.

I look forward to 2.3. Thank you for such an engaging game.

PR #421, opened 2023-03-11
ENGINE: Use setTimeout() instead of requestAnimationFrame()

d0sboots:
This changes the main loop to use setTimeout() instead of requestAnimationFrame(): This is both appropriate (we are updating game logic, not painting the screen), and also necessary for any kind of background processing: requestAnimationFrame is always *completely* descheduled when the tab loses focus.

### DIFF
--- a/src/engine.tsx
+++ b/src/engine.tsx
@@ -407,7 +407,7 @@ const Engine: {
       Player.lastUpdate = _thisUpdate - offset;
       Engine.updateGame(diff);
     }
-    window.requestAnimationFrame(Engine.start);
+    window.setTimeout(Engine.start, CONSTANTS.MilliPerCycle - offset);
   },
 };
 


Snarling:
Running a sound that's not supposed to be heard in the background to keep the game running is jank. Jank stuff like that can be implemented by players in their own scripts, does not need to be provided by the game.

The other change seems good though.

d0sboots:
It *is* a bit jank, but it's jankiness that's forced on us by the web ecosystem. There's literally no way to ensure that things like batchers continue to run without either doing this or running in a WebWorker (and we already discussed how the latter isn't really practical).

Rather than looking at absolute jankiness, I think the better metric is "is there a better/nicer way to do this" and "is this something that a significant number of people want?" For the former, the answer is no, and I believe the answer to the latter is yes. Also, this requires some relatively obscure technical knowledge to pull off, so I don't think it makes sense to expect people to implement it for themselves.

d0sboots:
Another primary motivation for having it be built-in is so that it can be saved as a setting. If you did it as a user script, you'd either need to remember to run it every time you started the game, or it would need to run in the background, taking up RAM. Or you could make it a library, and hook it into something that you intend to always run, if you have a script that is always running (but not everyone has a script like that.)

All of these have various downsides compared to a setting that always works.

Snarling:
The game can already be run with no browser based throttling, by running the electron version provided through steam.

If a player wants to play in a regular browser but also wants their browser to not act the way a browser normally acts, that is for them to either perform configuration on their browser or implement a jank solution like playing a constant meaningless sound.

Playing a sound has the potential to cause actual issues with the player's user experience. E.g. a bluetooth headset that is paired to multiple devices may continue playing a meaningless 1hz unwanted sound from a browser on the player's computer instead of playing an actually important sound from the other paired device. The ability to play an arbitrary number of audio streams simultaneously is also not guaranteed. IMO something like a constant hum has a high likelihood of causing negative impact on the player's experience in the rest of their operating system.

It is also just not necessary - the game does not require unthrottled operation, it has bonus time mechanics to make up for missed cycles.

> Or you could make it a library, and hook it into something that you intend to always run, if you have a script that is always running (but not everyone has a script like that.)

If a player doesn't have a script they're running, then they definitely don't need a sound playing nonstop. The only reason would be someone who specifically wants to batch but also wants to break the rules of how execution works in browser tabs. If you want to break those rules, do it, but don't expect the game to do it for you. Players implenenting it themselves also means it's super easy for them to just turn it off during times when it's no longer needed, via their own code instead of needing to manually navigate a settings menu.

d0sboots:
> If a player wants to play in a regular browser but also wants their browser to not act the way a browser normally acts, that is for them to either perform configuration on their browser or implement a jank solution like playing a constant meaningless sound.

If browsers provided a way to disable this behavior *per-site*, then that would be an option. But, they don't. Firefox only has knobs for changing the behavior browser-wide: https://superuser.com/questions/1500289/how-to-aggressively-throttle-background-tabs-in-firefox-using-dom-min-background. Chrome doesn't have any way to disable the throttling, at all. For better or worse, it is up to websites to take measures to ensure that they don't get throttled, instead of relying on the browser.

I also reject the solution of "just use Electron." We provide offline bundles for a reason. There are plenty of legitimate cases where someone wouldn't use steam, from it not being supported on their platform, to intermittent internet access, wanting to run custom or development builds *that others have built*, etc. The browser should not be a 2nd-class citizen.

Things like "there aren't unlimited audio streams" and "it might cause issues with bluetooth" are reasons we shouldn't *force* a sound to play. But they don't explain why you're against providing the *option*.

Also, bonus time is not enough. For instance, without focusing on the tab for a certain amount of time each day, you don't have enough runtime to burn down the bonus time for corp. It's even more skewed for stocks. You're basically saying that it's fine if tabbing away is equivalent to closing the tab.

Snarling:
> But they don't explain why you're against providing the option.

I have literally explained this already and I don't really have time to just keep repeating myself.

1. It's jank
2. If players who "need" a no-throttling environment want to implement this, they can. There is an option, it's just not first party supplied option because it's jank.
3. It has the potential to cause complications for the player outside of Bitburner.
4. "You're basically saying that it's fine if tabbing away is equivalent to closing the tab." Yes and every single web browser agrees. If you don't want your web page to stop processing, don't tab away. If you want to be able to tab away and still keep processing the tab unthrottled in a game that gives you all of js to play with, be prepared to implement a workaround. The game does not need to hand the players weird janky workarounds to defeat how web browsers work.

I'm not merging this, and I have explained why.

TheMas3212:
i would also agree that this is outside the scope of what the game should be doing

d0sboots:
OK, it's just the 1 line now.

PR #425, opened 2023-03-15
TEST: Add a test for save/load

d0sboots:
I'm working on some stuff touching savefiles, it's high time we had a test.

### DIFF
--- a/test/jest/Save.test.ts
+++ b/test/jest/Save.test.ts
@@ -1,0 +1,144 @@
+import "../../src/Player";
+
+import { loadAllServers, saveAllServers } from "../../src/Server/AllServers";
+
+// Direct tests of loading and saving.
+// Tests here should try to be comprehensive (cover as much stuff as possible)
+// without requiring burdensome levels of maintenance when legitimate changes
+// are made.
+
+// Get a stable clock so we don't have time-based diffs.
+function freezeTime() {
+  // You're about to hack time, are you sure? YES/NO
+  const RealDate = Date;
+  Date = function () {
+    return new RealDate(1678834800000);
+  };
+  Date.now = () => 1678834800000;
+}
+
+// Savegame generated from dev on 2023-03-12, mostly empty game with a few
+// tweaks. A RunningScript was added in-game to test the one bit of
+// non-trivial machinery involved in save/load.
+//
+// Most of the Servers have been removed to reduce space. Default values have
+// been removed both for space, and to test that they are added correctly.
+function loadStandardServers() {
+  loadAllServers(String.raw`{
+  "home": {
+    "ctor": "Server",
+    "data": {
+      "hasAdminRights": true,
+      "hostname": "home",
+      "ip": "67.4.8.1",
+      "isConnectedTo": true,
+      "maxRam": 8,
+      "messages": [
+        "hackers-starting-handbook.lit"
+      ],
+      "organizationName": "Home PC",
+      "programs": [
+        "NUKE.exe"
+      ],
+      "ramUsed": 1.6,
+      "runningScripts": [
+        {
+          "ctor": "RunningScript",
+          "data": {
+            "args": [],
+            "filename": "script.js",
+            "logs": [
+              "I'm a log line that should be pruned",
+              "Another log line"
+            ],
+            "logUpd": true,
+            "offlineRunningTime": 0.01,
+            "onlineRunningTime": 7.210000000000004,
+            "pid": 2,
+            "ramUsage": 1.6,
+            "server": "home",
+            "dependencies": [
+              {
+                "filename": "script.js",
+                "url": "blob:http://localhost/302fe9e5-2ec3-4ed7-bb5a-4f8f4a85f46d",
+                "moduleSequenceNumber": 2
+              }
+            ]
+          }
+        }
+      ],
+      "scripts": [
+        {
+          "ctor": "Script",
+          "data": {
+            "code": "/** @param {NS} ns */\\nexport async function main(ns) {\\n  return ns.asleep(1000000);\\n}",
+            "filename": "script.js",
+            "module": {},
+            "dependencies": [
+              {
+                "filename": "script.js",
+                "url": "blob:http://localhost/e0abfafd-2c73-42fc-9eea-288c03820c47",
+                "moduleSequenceNumber": 5
+              }
+            ],
+            "ramUsage": 1.6,
+            "server": "home",
+            "moduleSequenceNumber": 5,
+            "ramUsageEntries": [
+              {
+                "type": "misc",
+                "name": "baseCost",
+                "cost": 1.6
+              }
+            ]
+          }
+        }
+      ],
+      "serversOnNetwork": [
+        "n00dles"
+      ],
+      "purchasedByPlayer": true
+    }
+  },
+  "n00dles": {
+    "ctor": "Server",
+    "data": {
+      "hostname": "n00dles",
+      "ip": "61.6.6.2",
+      "maxRam": 4,
+      "organizationName": "Noodle Bar",
+      "serversOnNetwork": [
+        "home"
+      ],
+      "moneyAvailable": 70000,
+      "moneyMax": 1750000,
+      "numOpenPortsRequired": 0,
+      "serverGrowth": 3000
+    }
+  }
+}`); // Fix confused highlighting `
+}
+
+test("load/saveAllServers", () => {
+  // Feed a JSON object through loadAllServers/saveAllServers.
+  // The object is a pruned set of servers that was extracted from a real (dev) game.
+
+  freezeTime();
+  loadStandardServers();
+
+  // Re-stringify with indenting for nicer diffs
+  const result = saveAllServers(/*excludeRunningScripts=*/ false);
+  expect(JSON.stringify(JSON.parse(result), null, 2)).toMatchSnapshot();
+});
+
+test("load/saveAllServers pruning RunningScripts", () => {
+  // Feed a JSON object through loadAllServers/saveAllServers.
+  // The object is a pruned set of servers that was extracted from a real (dev) game.
+
+  freezeTime();
+  loadStandardServers();
+
+  // Re-stringify with indenting for nicer diffs
+  const result = saveAllServers(/*excludeRunningScripts=*/ true);
+  expect(JSON.stringify(JSON.parse(result), null, 2)).toMatchSnapshot();
+});
--- a/test/jest/__snapshots__/Save.test.ts.snap
+++ b/test/jest/__snapshots__/Save.test.ts.snap
@@ -1,0 +1,257 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`load/saveAllServers 1`] = `
+"{
+  "home": {
+    "ctor": "Server",
+    "data": {
+      "contracts": [],
+      "cpuCores": 1,
+      "ftpPortOpen": false,
+      "hasAdminRights": true,
+      "hostname": "home",
+      "httpPortOpen": false,
+      "ip": "67.4.8.1",
+      "isConnectedTo": true,
+      "maxRam": 8,
+      "messages": [
+        "hackers-starting-handbook.lit"
+      ],
+      "organizationName": "Home PC",
+      "programs": [
+        "NUKE.exe"
+      ],
+      "ramUsed": 1.6,
+      "runningScripts": [
+        {
+          "ctor": "RunningScript",
+          "data": {
+            "args": [],
+            "dataMap": {},
+            "filename": "script.js",
+            "logs": [],
+            "logUpd": true,
+            "offlineExpGained": 0,
+            "offlineMoneyMade": 0,
+            "offlineRunningTime": 0.01,
+            "onlineExpGained": 0,
+            "onlineMoneyMade": 0,
+            "onlineRunningTime": 7.210000000000004,
+            "pid": 2,
+            "ramUsage": 1.6,
+            "server": "home",
+            "threads": 1,
+            "dependencies": [
+              {
+                "filename": "script.js",
+                "url": "blob:http://localhost/302fe9e5-2ec3-4ed7-bb5a-4f8f4a85f46d",
+                "moduleSequenceNumber": 2
+              }
+            ]
+          }
+        }
+      ],
+      "scripts": [
+        {
+          "ctor": "Script",
+          "data": {
+            "code": "/** @param {NS} ns */\\\\nexport async function main(ns) {\\\\n  return ns.asleep(1000000);\\\\n}",
+            "filename": "script.js",
+            "url": "",
+            "module": {},
+            "dependencies": [
+              {
+                "filename": "script.js",
+                "url": "blob:http://localhost/e0abfafd-2c73-42fc-9eea-288c03820c47",
+                "moduleSequenceNumber": 5
+              }
+            ],
+            "dependents": [],
+            "ramUsage": 1.6,
+            "queueCompile": false,
+            "server": "home",
+            "moduleSequenceNumber": 5,
+            "ramUsageEntries": [
+              {
+                "type": "misc",
+                "name": "baseCost",
+                "cost": 1.6
+              }
+            ]
+          }
+        }
+      ],
+      "serversOnNetwork": [
+        "n00dles"
+      ],
+      "smtpPortOpen": false,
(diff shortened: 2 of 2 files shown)


Snarling:
I agree that we should have a test for saves, but I don't think this test as written is really useful in any way.

1. It would need to be updated every version at least, because version number of the save would change. If the input save is already fully in the format of the output save for the current version, then it's not really testing a whole lot, because we already know that the current version's saveObject creates save data in that exact format.
2. Because it is testing the actual output of the save to match an exact expected result, it becomes very difficult to maintain this test. Any change that affects the save would also need to make a change to this test, and it wouldn't even be clear where exactly the changes should be made (both the provided and expected save data, or just the expected save data?). Even just adding a new setting to the game would require modifying this test as written.
3. The important thing to test with saves is that saves from *different* versions still load gracefully with no errors in the current version. Such a test could include a battery of test saves (e.g. test saves from 1.7.0 / 2.0 / 2.1 / 2.2 with all endgame mechanics unlocked), and then make sure the engine loop does not cause any errors after loading each save. Not sure if that would be rigorous enough or what complications we might run into getting that to run, but at least it would be testing an actual point of failure and wouldn't be a nightmare to maintain the test. A test failure would mean we broke an old version of saves, not that we just forgot to update the exact expected savestring when we made some internal change.

d0sboots:
There's a sliding scale of coverage vs. sensitivity to unimportant changes when it comes to tests like this. At one end you have targeted unit tests, which are unlikely to be randomly broken but also only cover a small set of behaviors (and often not the ones that actually break). On the other end, you have integration tests and golden tests like this one, which cover a lot, but also need more maintenance because they are more sensitive.

I'd like to get some more of the latter type going, because our test coverage is very low and they tend to have more "bang for the buck," in my experience.

However, your point #1 about how this would need to be changed (at least) on every version means this particular approach won't fly. Back to the drawing board.

I don't entirely agree with #3. It's important to test that different versions work, especially because this is not something that gets as much manual attention, but it's not the *only* important thing. It's also important to test that regular save/load continues to work.

Snarling:
> I don't entirely agree with number 3. It's important to test that different versions work, especially because this is not something that gets as much manual attention, but it's not the _only_ important thing. It's also important to test that regular save/load continues to work.

Yeah that's true.

An idea, not sure if it can be implemented simply:

1. Simulate starting a new game
2. Using code, add a bunch of features to the player (e.g. jobs/augs/SF/corp/blade/sleeves performing work/ etc) that way we are testing as much of the possible savedata as possible.
3. Create a savedata based on this playerstate.
4. Start a new test that has access to the savedata from step 3 and load that save.
5. Make sure all the features we added to the player in step 2 are still present in the loaded save.
6. Save it again and make sure the new generated savedata remains identical to the savedata from step 3.

This would prevent manual upkeep of the test in most cases since the savedata is dynamically generated on the same version in both steps 3 and 6.

I'm not sure if that can be easily implemented though, having dynamically generated content from step 3 be available at the start of what pretty much needs to be a new test at step 4 (so that it is loading the page fresh).

d0sboots:
OK, I've reduced scope to just `loadAllServers`/`saveAllServers`. This removes all the issues you brought up, despite it being the same type of test:
1. It's not touching the core metadata, so things like game versions won't affect the test.
2. This is not a hot-path for changes: Adding settings or other stuff like that won't affect this test. It's only scoped to server-related changes.
3. Because it's testing the inner property, we're dealing with JSON strings instead of JSON escaped strings now. This makes it possible to format the results, so diffing and maintenance become much nicer.

The only downside is that there's still a lot of "fluff" surrounding the "meat" of the test: It's not visually obvious that the logs are being pruned, for instance. But I think the simplicity of directly testing strings is worth it.

For testing other sections of the save, we can either use the same approach, or a different one where this approach would have too much churn.

Snarling:
I still don't really see the point of this type of hardcoded string test. It's extremely difficult to look at it and tell what it's even testing, and it also prevents changing anything about the server object without also changing an arcane test.

e.g. Potentially there are changes soon to the way files are stored on a server, so that finding whether a script already exists doesn't require iterating through each entry of a scripts array and checking if the filename is equal (necessary because it's stored as an array and not an objectlike structure). Changing this to a more performant data structure would completely break a test like this. A server's running scripts being stored in an array is one of the main reasons there are huge performance losses in launching thousands of scripts from the same server, something I'm sure you're aware people do a lot in situations where they would like the game to be as performant as possible. So it's likely some stuff will change in the structure of server objects and therefore a hardcoded string test isn't desirable.

Did you read the suggestion I had above? This would test save fidelity (step 5 verifies that the data from step 2 can be saved and then loaded), and would also test savestring equivalence on load+save (savestring in step 3 is equal to savestring in step 6 that has gone through an additional load/re-save). As far as I can tell that approach can cover anything we would want to test without requiring huge, difficult to maintain hardcoded strings that lock us out of making any internal changes. This type of test may be useful for projects where it's vitally important that the shape of an output must remain constant, but that's just not true with save data in bitburner. There do not need to be barriers to changing how data is organized, other than ensuring that saves from previous versions still convert over gracefully to the current version.

d0sboots:
If we did something like changing `scripts` from an array to an object, I don't think it would actually break this test. It *ought* to, but that's a different matter. The reason is that our save/load routines are just loading/storing JSON usually with no validation, so if we're stuffing the wrong type of object in they don't care.

Oh, the test would actually break because the default value would change for n00dles. That's good.

I did read your suggestion, but it didn't seem like it was buying anything to me. The issue is in step 5: in order to validate that everything is present, we have to traverse all the internal data structures of `Server`. This is just as prone to breakage as checking the savestring; in fact, checking the JSON dump of an object is the simplest and most readable way of asserting the total state of the object!

Ultimately, because we are testing internal details of how Servers are stored, we *want* and *expect* to update the tests if those internal details change.

d0sboots:
Oh, and like I said above: This same approach wouldn't necessarily be right for something like settings, which churns a lot more heavily. Although you should often expect to be updating tests when you update code, for settings I don't think the trade-off would be right. To me at least, testing is a pragmatic as opposed to principled thing.

Snarling:
> I did read your suggestion, but it didn't seem like it was buying anything to me. The issue is in step 5: in order to validate that everything is present, we have to traverse all the internal data structures of `Server`.

Step 6 is to validate that everything that is present in the save from step 3 is still present and unchanged after a fresh load+re-save. So step 6 is the only "check everything" step, and it's just checking that everything that was originally saved is saved identically again after a fresh load. This is a full data structure check, but it's checking that a dynamically saved data structure from step 3 is identical to the dynamically created data structure created when loading and then re-saving it.

Step 5 is just to check for existence and basic functionality of things that were added to the player in step 2, e.g. for bladeburner we would do Player.bladeburner = new Bladeburner() or whatever during step 2, and then in step 5 we would just check that Player.bladeburner is defined and that Player.bladeburner.process() does not lead to any errors. We don't need to do a full shape check on the bladeburner object. Same goes for pretty much all the other features we're testing. So instead of hardcoding the entire shape of a feature, we're just creating it before saving and then making sure it still exists and works after loading.

My suggestion does not include anything that amounts to a full shape check against a hardcoded expected structure. That's what your test is, but it's a type of test I don't think is fitting for save data. So if you still think testing against a hardcoded expected structure is something useful for us to implement for saves, that is where we're not seeing eye to eye. In my eyes a hardcoded structure test is a prison we don't need to put ourselves in, especially if we spend our test-improvement efforts on implementing a test to make sure saves from previous versions load gracefully in the current version, and the advanced features in that savegame (blade, corp, gang, etc) are still present in the loaded version and process without error.

With the suggestion I provided, we still could end up with errors in the test when something like server scripts data structure changes (e.g. we might add a test script to a server in step 2, then check for the existence and maybe try copying it to another server in step 5 to check for runtime errors during usage). When changing scripts from an array to a string-keyed object, we'd end up with a type error or runtime error on this test if we forgot to update the test. But fixing this wouldn't require manually editing hardcoded jsons, it would just be changing the one line where we access the script to be the correct way (e.g. server.scripts["test.js"] instead of server.scripts[0]). Way more elegant and frictionless IMO. I certainly don't want to have to be hand-editing json strings to fix a test that was expected to break on any internal data shape changes.

Quick correction on the 6-step thing, really we would perform step 6 prior to step 5. Step 5 has the potential to cause changes in the data as we run process functions as part of the test. An immediate save after load with no processing should not have any mutations (if it's a save from the same game version, which is guaranteed with step 3). Since I've already referred to the steps by number so many times, I'll just keep them in this mixed up order and add a note to step 6.

d0sboots:
Yes, I do think a full check of the shape is useful. Or at the very least, checking *most* of the shape is useful, and at that point it's easiest to check the whole thing.

Where this is most useful is "negative" tests: checks that certain things are *not* included in the save. Saving the logs of a RunningScript will not cause issues (usually, it's a different story now that they can contain React elements), but it is a performance problem. One of the things the test is currently explicitly checking is that RunningScript logs are dropped.

Looking at the whole shape has revealed other issues, or potential issues. We're saving the pid, and the blob info, and the compiled blob info for scripts. I'm willing to bet all of those things are bugs, and it's only apparent because I was forced to *look* at the full data. If/when we fix it, testing the full shape is the best way to make sure they stay gone.

I've worked with this type of test for my job; it's not uncommon or unusual. Yes, there is a higher rate of spurious failures due to the higher sensitivity. However, this is balanced by the greater ease in fixing them: you literally copy-paste the new result to expected, it's entirely mechanical. As long as they're used *appropriately* (and I'm specifically only wanting to do this for AllServers, I don't know that it would be the right approach for Bladeburner or Corp or anything else), it's effective and not a burden.

d0sboots:
<https://blexin.com/en/blog-en/golden-master-pattern-dont-fear-the-legacy-code/> has a good overview of the technique. It also raises a good point, which is that full coverage of the app using traditional unit testing would be better, but generally people looking to write golden master tests are very far from that point. Golden master testing (aka Characterization Testing) is attractive as a pragmatic technique because it gets a lot of coverage very cheaply.

I will look into moving the expected output to its own file, which allows the test framework itself to update the test on failure if you flag it as "correct."

d0sboots:
I realized some more context would be helpful.

I'm working on refactoring `saveAllServers` to improve its speed. Once I'm done, *nothing* about its output should be changed. This test is the type of test that provides that assurance. That's why the 6-part test you outlined, while a good test for certain things, is ultimately uninteresting to me: It can't provide the surety that I haven't accidentally changed something (because it's not checking *everything*), so I'd have to write this test anyway.

So this test *will* be useful, the question is whether it's useful only to me. XD

d0sboots:
It turns out that Jest already had the infrastructure for this built-in, in the form of `toMatchSnapshot`! The error messages it generates on failure are very nice, both highlighting where the changed lines are and also providing the command to run to update the snapshots if the change is intentional.

Snarling:
> We're saving the pid, and the blob info, and the compiled blob info for scripts. I'm willing to bet all of those things are bugs

Yeah I also noticed this, I can't imagine there's a good reason the blob url should be saved as that needs to get re-created after loading anyway.

> I'm working on refactoring saveAllServers to improve its speed. Once I'm done, nothing about its output should be changed.

Yeah that is useful context. Personally I think this type of test is good for certifying an individual change like that, but almost every other time that we're making a change that affects the saving of server objects in any way, it's because the actual structure is changing. It's not something that an unrelated change somewhere else in the code is going to accidentally affect and we'll need a test to tell us "hey programmer, you also affected the shape of server saves with that change, did you mean to do that?"

> I've worked with this type of test for my job; it's not uncommon or unusual. 

I think sometimes people try over-applying things from their traditional programming projects/jobs to a project like this. Like in most projects, a full shape check is super useful because you actually want data shapes that are never-changing, because maybe other teams or other parts of your application are relying on that shape remaining constant. But part of the draw of a project like this is that we control the whole internal codebase, and there are no other teams that we need to maintain stability on internal code for. So if we want to make a change to servers (e.g. change scripts from an array to an object keyed by filepath), we just have to make all the changes throughout the codebase to make that work (which typically ts will force us to do anyway before runtime, or basic runtime functionality checks would check in the cases where ts is still happy with bad code).

For instance that extra data you noticed is being saved but probably doesn't need to be: with a test like this in place, anyone who wants to cull that unneeded data from the save would need to modify an extra test in order to do so. But they know they're removing that data, they don't need a test to yell at them for it. I think that would be the case for basically every PR that affects this test, the submitter already knows they are altering the shape of the server save.

> It turns out that Jest already had the infrastructure for this built-in, in the form of toMatchSnapshot! The error messages it generates on failure are very nice, both highlighting where the changed lines are and also providing the command to run to update the snapshots if the change is intentional.

Depending on how frictionless this is to use when there are intentional shape changes, maybe this is something we could use. But again I don't see a particular usefulness in this project of doing shape checks when the changes that cause the test to fail are almost always going to be intentional (because we're testing that something we don't expect to remain unchanging doesn't change).

d0sboots:
> I think sometimes people try over-applying things from their traditional programming projects/jobs to a project like this. Like in most projects, a full shape check is super useful because you actually want data shapes that are never-changing, because maybe other teams or other parts of your application are relying on that shape remaining constant. But part of the draw of a project like this is that we control the whole internal codebase, and there are no other teams that we need to maintain stability on internal code for. So if we want to make a change to servers (e.g. change scripts from an array to an object keyed by filepath), we just have to make all the changes throughout the codebase to make that work (which typically ts will force us to do anyway before runtime, or basic runtime functionality checks would check in the cases where ts is still happy with bad code).

This is how my job was too. It was very frictionless to change code. The only real difference was that we had much better tests overall. XD

> I think that would be the case for basically every PR that affects this test, the submitter already knows they are altering the shape of the server save.

You're missing the point of the test, which is to catch the times when the data changes and it *wasn't* supposed to. There are lots of changes where you don't *think* it will affect something, but it actually does. This type of test lets you know, "Hey! Whoa there! You actually changed this over here, did you mean to?"

A lot of the time, you *did* mean to. That's why it's especially easy to update the test with new data. But sometimes, you didn't, and that's the whole point. Just like with regular unit tests, if they only ever failed when we were intentionally changing that code, there would be no point. Tests exist to catch the *unintentional* things.

Edit: And yeah, you could say "we'll only use this when we know we're refactoring that code." That has some value. But again, it misses the part where we make mistakes at times when we don't think we have to worry about making mistakes.

d0sboots:
Here's another thing that's worth pondering: Even as low as our existing test coverage is, I know that several of them have needed changes (sometimes multiple times) due to legitimate reasons. Have any of them ever actually caught a *real* problem, as opposed to a false-positive? And if so, how many, as compared to the number of times they've needed updating?

"Regular" unit-tests skew heavily towards false-positive detection, as well.

Snarling:
So that other PR was nice and it made it a lot easier for me to review based on you having a test showing that the output was exactly as expected, but I think it's probably the first PR ever where I might have though to myself "it would be nice to have a test so I could make sure the server save is still working right." Very specific and something I don't think any PR has affected for ages.

But after thinking about it some more, I do think it might be worth implementing an overall savefile test of some sort. Changes to the save file should be allowable and not directly discouraged, but they should be rare enough that alerting a contributor that their changes modify the save file output is not going to be too cumbersome, especially if updating the snapshot is fairly straightforward and frictionless.

Something I don't like at all is any of that test being hardcoded jsons that someone manually generated. Snapshot should start empty, then we can dynamically generate a save (by e.g. adding features to the player, simulating as many of the save file objects as possible), then save that and compare with the snapshot. Obviously the first run will fail since it won't be an empty save, but we'll just use that as the first replacement snapshot. 

We can also check separately that loading the snapshot and then re-saving creates output identical to the save content created by dynamically generating a new save. In cases where the save file format has actually changed, this test ensures that there is sufficient safe-loading code to convert the old save into the new format.

I get that your test as written is specifically also testing e.g. that logs don't get saved and that's why you have two different hardcoded jsons for input and expected output because you were specifically interested in that one interaction, but that's not really a broad test that matters for the codebase, it's literally just a test for your one PR that implements excluding runningscript logs from the save. That sort of thing does not need to be tested overall.

d0sboots:
> I get that your test as written is specifically also testing e.g. that logs don't get saved and that's why you have two different hardcoded jsons for input and expected output because you were specifically interested in that one interaction, but that's not really a broad test that matters for the codebase, it's literally just a test for your one PR that implements excluding runningscript logs from the save. That sort of thing does not need to be tested overall.

You say that, and yet I'm about to send you PR 2 that hits exactly these areas XD

Snarling:
Already merged through #432

PR #432, opened 2023-03-18
NETSCRIPT: Add "temporary" as a RunOption to run/exec/spawn

d0sboots:
This is blocked on the resolution of the other PR, since it updates that test. (You can use this PR as an example of what updating that test looks like, for an intentional change.)

----

This creates the RunOptions interface, to pass as a parameter to
run()-type functions. For backwards compatibility, you can pass a thread
number or an options struct, and the struct has a "threads" member.
(This mirrors how JS itself has been extended.)

The new option is "temporary": It marks a RunningScript to be excluded
from the save. This is both a (significant) performance improvement, as
well as a useful semantic: There are lots of times where you have a
central "manager" script that you want to be restarted on load, but it
spawns helper or "worker" scripts that shouldn't be restarted.

It's possible to write your code so that the helpers will immediately
exit if they detect that they are running from a fresh load, but it's
both simpler, smaller, and more performant to never save them in the
first place.


Snarling:
Seems like a useful feature.

I see it's still draft and you haven't handled this part yet, but I'm curious how you want to handle the NetscriptDefinition entries given the complex parameter. One option I guess would be just soft-deprecating threads, another would be multiple documentation pages for the two function signatures (like ns.kill).

d0sboots:
Oh, I had only marked it as a draft because of the test issue; this PR depends on (and thus includes) the other one, so that should either be merged first or we should decide that we're not testing AllServers. Other than that, I'd thought it was ready.

I'll generate the docs and see how they look in the current state. (Really generating docs should be included with every PR, no?)

d0sboots:
K, I'd missed the docs on the functions, that's fixed now. Still needs deciding on the tests.

Snarling:
Really any test that mainly consists of hardcoded json is just not getting merged.

IMO the only hardcoded json should be in the snapshot, that way it can be updated without having to manually edit jsons. Also I don't think a test specifically for this one portion of the save file is warranted. It's useful for demonstrating that this PR works, but it's not really useful outside that in my eyes. Some tests for the overall save file would be useful but there's no need to include that in this PR if you'd just like this merged sooner.

Snarling:
Also that documentation turned out a lot less gross looking than I expected.
I would have just gotten rid of the explicit statement that you'll get a runtime error if you use less than 1 thread. But that was already there.

d0sboots:
> Also I don't think a test specifically for this one portion of the save file is warranted. It's useful for demonstrating that this PR works, but it's not really useful outside that in my eyes.

OK. I strongly disagree, but I'll pull the test out and keep it local, to run against my changes. It'll reduce visibility into why my changes are safe, and it's a little awkward to manage a permanent local branch with rebasing and stuff, but not unmanageable.

Snarling:
I'll just merge it for now and we can rework it later into a full savegame test.

PR #435, opened 2023-03-20
CODEBASE: Stricter types and added an enum helper

Snarling:
The stricter types are surrounding companies and factions.

Adding types required a lot of refactoring, and while I was doing that I also streamlined a few sections of code.

Started adding more categories of enums, but more types will need to be added into these categories.


Snarling:
Replaced by #596

PR #440, opened 2023-03-22
NETSCRIPT: Greatly speed up script launching, and remove the limitation unique args per script

d0sboots:
This turned out to be a major refactor. Under the covers, the largest change is removing `runningScripts` from servers and replacing it with `runningScriptMap` of type `Map<ScriptKey, Map<number, RunningScript>>`. The ScriptKey is a string that encodes the script name + arguments. A two-level map is needed, because script name + arguments are no longer enough to uniquely identify a script. I went through a lot of alternative designs, this was the simplest one I could think of that allows O(1) lookup/deletes both by PID and by arguments.

The original purpose of this change, which is still retained, is removing the limitation of unique arguments per script invocation. This is now gone for running from the Terminal, and (by default) for `ns.run()`, `ns.exec()` etc. However, `preventDuplicates` can be specified as a RunOption to get the old behavior back, if you were depending on it in scripts. (IMO the new behavior is a much nicer default for most people, so it is worth the potential of breakage.)

Another significant change is adding `filePath` and `scriptPath` as helpers in NetscriptHelpers, this simplified and regularized a lot of functions in NetscriptFunctions, and can be considered a follow-on to the FilePath change.

After the change:
```
[home ~]> run benchmark_save.js 
Running script with 1 thread(s), pid 500006 and args: [].
Launching 100000 copies...
Launched in 0.524 seconds
Scripts started in 0.825 seconds. Saving game...
Saving took 0.037 seconds
```
```js
/** @param {NS} ns */
  ns.disableLog("run");
  ns.disableLog("kill");
  const orig = React.createElement;
  let resolve;
  // Pull the save function from the button on the character menu
  React.createElement = function (...args) {
    const props = args[1];
    if (props && props.save && props.killScripts) {
      React.createElement = orig;
      resolve(props.save);
    }
    return orig.call(this, ...args);
  };
  const resultP = Promise.race([
    new Promise((res) => (resolve = res)),
    ns.asleep(1000),
  ]).finally(() => {
    React.createElement = orig;
  });
  // Force a rerender
  ns.ui.setTheme(ns.ui.getTheme());
  const saveFunc = await resultP;
  if (!saveFunc) {
    ns.tprint("ERROR: Couldn't find save function!");
    return;
  }

  const COPIES = 10000;
  ns.write(
    "benchmark_save_helper.js",
    `
/** @param {NS} ns */
export function main(ns) {
  ns.writePort(ns.args[0], "");
  return ns.asleep(300000);
}`.trim(),
    "w"
  );
  ns.tprintf("Launching %d copies...", COPIES);
  const beforeLaunch = performance.now();
  ns.atExit(() => {
    const host = ns.getHostname();
    for (let i = 0; i < COPIES; ++i) {
      ns.kill("benchmark_save_helper.js", host, ns.pid, i);
    }
  });
  const promise = ns.getPortHandle(ns.pid).nextWrite();
  for (let i = 0; i < COPIES; ++i) {
    const pid = ns.run("benchmark_save_helper.js", 1, ns.pid, i);
    if (pid < 1) {
      ns.tprintf("Failed to launch copy %d, benchmark failed", i);
      return;
    }
  }
  const afterLaunch = performance.now();
  ns.tprintf("Launched in %.3f seconds", 0.001 * (afterLaunch - beforeLaunch));
  await promise;
  const afterStart = performance.now();

  ns.tprintf(
    "Scripts started in %.3f seconds. Saving game...",
    0.001 * (afterStart - afterLaunch)
  );
  await ns.asleep(0);
  const before = performance.now();
  saveFunc();
  await ns.asleep(0);
  const after = performance.now();
  ns.tprintf("Saving took %.3f seconds", 0.001 * (after - before));
}
```


steven-r:
This will possibly break a couple of scripts as they rely on the "do not start again" feature.

d0sboots:
> This will possibly break a couple of scripts as they rely on the "do not start again" feature.

Ah, you mean scripts are blindly launching something that might already be running, to guarantee it will run, but they only want one copy?

That makes sense. *sigh* I'll put the new behavior in ExecOptions.

d0sboots:
Alright, I added a new option and also adjusted docs. I think this is ready now.

The new option `preventDuplicates` is false by default, meaning everyone gets the faster behavior but also it may break some people. (But it's much easier to fix in user code, since it's a small addition to get the old behavior back.) If we want to be conservative, we can start with it true and (maybe) change it later.

steven-r:
I still like the idea and how to circumvent the new behavior.
But as a user for some years now and running over all struggles with the 2.2 API changes: Can we introduce breaking changes in a non "on the fly" manner?
How about having a release policy to allow breaking changes only on Major releases and keep existing APis "as-is" , if not fixing a security incident or a major bug.
I have dozens of scripts that might need an adaption which is something I do only if I have enough spare time.

Maybe it's not the right place to discuss this as part of a PR and we should move this to discord.

steven-r:
I still like the idea and how to circumvent the new behavior.
But as a user for some years now and running over all struggles with the 2.2 API changes: Can we introduce breaking changes in a non "on the fly" manner?
How about having a release policy to allow breaking changes only on Major releases and keep existing APis "as-is", if not fixing a security incident or a major bug.
I have dozens of scripts that might need an adaption which is something I do only if I have enough spare time.

Maybe it's not the right place to discuss this as part of a PR and we should move this to discord.

d0sboots:
> How about having a release policy to allow breaking changes only on Major releases and keep existing APis "as-is" , if not fixing a security incident or a major bug.

Personally, I agree with you: I think we're too eager to make breaking changes, on the whole. This coming release *will* be a Major release like you describe, because corporations have been significantly overhauled (along with accompanying API changes), but I don't think that lets us off the hook completely.

The only reason I implemented it as default on (and thus breaking) was because this is a significant performance improvement for everyone, especially people who run a lot of scripts. The opposite would be default-slow, which isn't a situation I want if I can help it. Also, it's confusing: Why does launching scripts sometimes fail, even if you have the RAM? Regular OSes don't work like that.

But, just having the option is a big improvement and would allow savvy players a way to get the speedup, so I'm fine with whatever Snarling decides on this one.

steven-r:
I agree on the performance aspect and having this change together with the corporation overhaul as a major release.

Then we as the users should have a long enough beta time to adjust our scripts accordingly (or even having to versions of BB active for some time (if steam allows this)

Snarling:
There should be no needed script changes from this, unless you were relying on your script failing to launch a duplicate script.

Edit: Oh I see that's what's already mentioned above, that some scripts probably do rely on this feature (e.g. just using exec/run as a proxy for an isRunning check)

Snarling:
As far as performance improvements, another way we can accomplish the same thing might be through a clever data structure.

One idea without changing game behavior is that we could make server.runningScripts into a Map keyed by the script's filename. The values of that map would themselves be maps, keyed by the stringified args array with the RunningScript as value.

Launching with an array with no checks is super quick - just push onto the array. With these maps it wouldn't be too much more - it needs to stringify its args array and do a couple of map operations. Still a huge improvement compared to iterating through potentially thousands of runningscripts to compare filename and then compare all args for any with matching filenames.

Looking up whether a script is running or not (for instance, to prevent player from deleting a file that is currently running) becomes extremely simple - just check if the runningscripts map has the script filename. Compared to iterating potentially the entire array of runningScripts in the expected case that it's not running and the deletion/moving is allowed.

Finding a particular runningScript with filename+args remains possible, and is now that same stringify + couple of map lookups instead of iterating through a massive array and doing comparisons on every entry.

When a runningScript dies, with an array it would have to splice itself out which is slow on big arrays. With maps, removing itself is super quick, just a couple of map operations (one to clear itself, then checking the parent map to clear the filename key if the size is now 0). In cases where there are a lot of scripts running, this should average way faster than splicing an array.

Thoughts?

d0sboots:
> As far as performance improvements, another way we can accomplish the same thing might be through a clever data structure.

I thought of that as well, but I didn't want to go that route because ultimately, I think "all arguments must be unique" is a bad/confusing limitation. However, if we're worried about breaking scripts to the point where the default of the preventDuplicates would be "true", it would probably be worth being fancy to speed up the common case.

One thing I'll note is that due to memory layout and access patterns, splicing an element out (if you know it's location) is much faster than searching for it.

Snarling:
>  I think "all arguments must be unique" is a bad/confusing limitation.

Yeah to an extent I agree with this. So on a gameplay perspective I could kind of go either way on it - one way to favor stability, the other way to favor a change that makes the game make more sense.

>One thing I'll note is that due to memory layout and access patterns, splicing an element out (if you know it's location) is much faster than searching for it.

Yeah so I did some testing, and splice on its own is surprisingly more performant than I thought it would be, considering all elements after the splice need to have their indices adjusted. Probably some browser optimization magic going on behind the scenes for that. Notable though that a runningScript can't really know its own index because that info is not static, so Search + Splice is the more realistic comparison for how a runningScript has to remove itself from the array. That's true for script death whether duplicates are allowed or not. Allowing duplicates makes scripts quick to launch with arrays, but they will still be slow to die.

Here's the test results for different methods of clearing a large runningScripts array/map one RunningScript at a time:

image

May have to re-run these tests. For the arrays, I was deleting from the middle of the array, but I think my setup was deleting from the earliest entries in the map which may have a performance advantage over deleting random/middle entries (wasn't able to find definitive info on Map.prototype.delete time complexity).

d0sboots:
Good point on needing to search + splice on delete. I guess I'll go ahead and do the fancy thing.

I thought about moving `runningScripts` into `scripts`, but that has issues because you can have multiple versions of a script with the same name on a server, when you look across a span of time. Making a two-level map would work, but it looks a bit weird, and also has two lookups for no purpose in the vast majority of cases. I think I'll just make a single-level map that's directly keyed by `JSON.stringify` of the script name + args.

> Yeah so I did some testing, and splice on its own is surprisingly more performant than I thought it would be, considering all elements after the splice need to have their indices adjusted.

The indices aren't stored in the elements or anything like that. An array of objects (or an array of primitive types, like "all strings" or "all numbers") is just a flat slab of memory, so adjusting it by one is just a `memcpy`. It's still O(N), but it's much faster than just about anything else you can do that's O(N).

Edit: You can also see a clear performance cliff at both 100000 and 200000 for Search+Splice, as the elements stop being able to fit in L3 cache.

Snarling:
> I think I'll just make a single-level map that's directly keyed by JSON.stringify of the script name + args.

I thought about this and I don't think it works for determining if a particular script is running in the first place, which is used right now to prevent deletion or moving of a file that is currently running. With that kind of key, you would need to iterate through all keys and do some string analysis on each one to determine if a particular script is running.

If you want to implement it can you commit to the branch from #473 where we already have JSONMaps and server.scripts as a JSONMap? I also don't mind implementing it.

Since RunningScripts prevent the parent Script from being removed/renamed, it actually would be a possibility to make RunningScripts a child of Script instead of doing a new double-layer map (well this is still a double layer map, just the first layer is the actual scripts map instead of a new map that is keyed the same as the scripts map and is just for holding the second-level maps). The runningScript might refer to an older compiled version of the script, but that's fine.

d0sboots:
> I thought about this and I don't think it works for determining if a particular script is running in the first place, which is used right now to prevent deletion or moving of a file that is currently running.

I was thinking of getting rid of that restriction too, since it's dumb. You can completely overwrite a script (or set it to empty string) while it's running, but you can't remove it? Makes no sense, and also (up until now) was totally unneeded for the implementation.

> If you want to implement it can you commit to the branch from https://github.com/bitburner-official/bitburner-src/pull/473 where we already have JSONMaps and server.scripts as a JSONMap? I also don't mind implementing it.

I'll keep using this PR, since I don't want to pull all of it into the other one. I'll rebase it off yours for now, and wait for it to be submitted.

Snarling:
> I was thinking of getting rid of that restriction too, since it's dumb. You can completely overwrite a script (or set it to empty string) while it's running, but you can't remove it? Makes no sense, and also (up until now) was totally unneeded for the implementation.

I think it's primarily so that when you reload the page, you can't have RunningScripts that are specifying an invalid filename (although since the player is allowed to modify the file, you can end up with scripts that reload but then fail due to ram usage, so...).

Also while I'm not sure that Linux does this, Windows typically prevents users from deleting/renaming/(or even modifying) files that are currently in use, so the behavior does correlate to reality as well. Although without enforcing no-modifications, it does seem kind of contradictory. Enforcing no-modifications to scripts isn't something we should do, because it would be a huge pain for the player to deal with (it's common to have a script running and be revising it at the same time). So either keeping the existing restriction (but not extending to no-modification) or getting rid of the restriction would be fine. Just will need some graceful handling for page reloads where running scripts are targeting an invalid filename, maybe with a toast or terminal message to inform the player that some runningscripts couldn't be loaded because file X no longer exists on server Y or something.

d0sboots:
I'll make sure out has some kind of toast. I think the standard "couldn't find X file" should be sufficient, it should already be obvious why the game is trying to run it.

Snarling:
@d0sboots Do you plan to get the runningScripts-as-map rework into 2.3? Thinking about pausing major commits until after 2.3 release, but if you have something in progress and close to ready, we can squeeze it in during the next day or two. I'd like to actually release 2.3 in about a week.

d0sboots:
I think I can have it ready. All the code is done, I'm just debugging at this point.

d0sboots:
I have rebased onto `dev`, in particular the filepath change.

Having done all that: Please hold the train for this. I'm seeing a lot of issues that I'm fixing incidentally as I continue to debug this, that are probably fallout from it.

Currently uploaded version is a ways from working, but it doesn't crash on startup (at least).

Note that we both rewrote `kill` and `tail`, differently. I'm not sure what to do about that.

src/NetscriptFunctions.ts
```diff
@@ -814,24 +814,15 @@ export const ns: InternalAPI<NSFull> = {
...
-        if (!path) throw helpers.makeRuntimeErrorMsg(ctx, `Invalid filepath: ${file}`);
+      const path = helpers.filePath(ctx, "files", file);
+      if (hasScriptExtension(path) || hasTextExtension(path)) {
         contentFiles.push(path);
         continue;
       }
-      if (!file.endsWith(".lit")) {
+      if (!checkEnum(LiteratureName, path)) {
         throw helpers.makeRuntimeErrorMsg(ctx, "Only works for scripts, .lit and .txt files.");
       }
```
Snarling: If the player provides a .lit filename that doesn't happen to be one of the valid enum members (e.g. they misspell the filename or make one up) they will get a bad error message. I think this is a case where we don't need to check the enum. If they provide a bad .lit filename, the error should be "that file doesn't exist", not "you need to provide the right file extension"

d0sboots:
I think this is ready to go now. I tested script stuff pretty thoroughly, to the extent that I found some bugs that were pre-existing instead of caused by this change. XD

This is a big change though, and there's no way I was able to manually test everything.

Full description is updated in the first post.

src/Netscript/WorkerScript.ts
```diff
@@ -33,7 +33,7 @@ export class WorkerScript {
   delay: number | null = null;
 
   /** Holds the Promise reject() function while the script is "blocked" by an async op */
-  delayReject?: (reason?: ScriptDeath) => void;
+  delayReject: ((reason?: ScriptDeath) => void) | undefined = undefined;
```
Snarling, line 36: Is there a reason for changing this to specifically define the variable as undefined, instead of just not defining it?

src/Terminal/commands/tail.ts
```diff
@@ -1,66 +1,38 @@
...
+      const [rawName, ...args] = commandArray;
+      const path = Terminal.getFilepath(rawName);
+      if (!path) return Terminal.error(`Invalid filename: ${rawName}`);
+      if (!hasScriptExtension(path)) return Terminal.error(`Invalid file extension. Tail can only be used on scripts.`);
+
+      const candidates = findRunningScripts(path, args, server);
+
+      // if there's no candidate then we just don't know.
+      if (candidates === null) {
+        Terminal.error(`No script named ${path} is running on the server`);
```
Snarling: Isn't this error message inaccurate? All that's checked is whether path+args is running, the script with that path might be running with different args.

src/Netscript/killWorkerScript.ts
```diff
@@ -14,40 +14,21 @@ import { AddRecentScript } from "./RecentScripts";
 import { ITutorial } from "../InteractiveTutorial";
 import { AlertEvents } from "../ui/React/AlertManager";
 import { handleUnknownError } from "./NetscriptHelpers";
+import { roundToTwo } from "../utils/helpers/roundToTwo";
 
 export type killScriptParams = WorkerScript | number | { runningScript: RunningScript; hostname: string };
 
-export function killWorkerScript(params: killScriptParams): boolean {
+export function killWorkerScript(ws: WorkerScript): boolean {
```
Snarling: Not a big deal but the killScriptParams type is probably unused now and can be removed. Lint won't recognize that because it's an export

src/NetscriptFunctions.ts
```diff
@@ -741,21 +738,23 @@ export const ns: InternalAPI<NSFull> = {
...
+        if (byPid === null) {
           helpers.log(ctx, () => helpers.getCannotFindRunningScriptErrorMessage(ident));
           return false;
         }
 
-        res = killWorkerScript({ runningScript: runningScriptObj, hostname: server.hostname });
+        res = true;
+        for (const pid of byPid.keys()) {
+          res = killWorkerScriptByPid(pid);
+        }
```
Snarling: The resulting value for `res` will only depend on the status from the last pid processed. If there was an error of some sort in the middle of the map, the status will be overwritten by the next iteration of that for loop and it will look like there were no errors.

src/NetscriptFunctions.ts
```diff
@@ -1414,34 +1405,30 @@ export const ns: InternalAPI<NSFull> = {
...
     const hostname = helpers.string(ctx, "hostname", _hostname);
     const server = helpers.getServer(ctx, hostname);
     let suc = false;
-    for (let i = 0; i < server.runningScripts.length; i++) {
-      if (server.runningScripts[i].filename == scriptname) {
-        killWorkerScript({ runningScript: server.runningScripts[i], hostname: server.hostname });
-        suc = true;
-        i--;
+
+    const pattern = matchScriptPathExact(escapeRegExp(path));
```
Snarling, line 1420: Just by name, I would expect a function named matchScriptPathExact to take in a path, not for it to require an already regex-escaped path.

src/Terminal/commands/kill.ts
```diff
@@ -1,25 +1,50 @@
...
-  if (typeof args[0] === "number") {
-    const pid = args[0];
-    if (killWorkerScript(pid)) return Terminal.print(`Killing script with PID ${pid}`);
-  }
-  // Shift args doesn't need to be sliced to check runningScript args
-  const fileName = String(args.shift());
-  const path = Terminal.getFilepath(fileName);
-  if (!path) return Terminal.error(`Could not parse filename: ${fileName}`);
-  if (!hasScriptExtension(path)) return Terminal.error(`${path} does not have a script file extension`);
+  try {
```
Snarling, line 9: Re: your other comment, rewriting tail/kill is obviously necessary for this rework so it makes complete sense to override my changes. But I think now the only two terminal commands with this weird top-level try/catch structure are your two rewrites. That structure is a remnant from when certain Terminal helper functions would throw if they were not able to succeed, instead of e.g. optionally returning null to indicate failure

src/Terminal/commands/killall.ts
```diff
@@ -1,10 +1,12 @@
...
 
 export function killall(_args: (string | number | boolean)[], server: BaseServer): void {
   Terminal.print("Killing all running scripts");
-  for (const runningScript of server.runningScripts) killWorkerScript(runningScript.pid);
-  WorkerScriptStartStopEventEmitter.emit();
+  for (const byPid of server.runningScriptMap.values()) {
+    for (const runningScript of byPid.values()) {
+      killWorkerScriptByPid(runningScript.pid);
+    }
+  }
```
Snarling, line 11: Maybe still need this WorkerScriptStartStopEventEmitter emission. I'm not sure though, I am reviewing from web interface and haven't dug into whether it was needed there.

Snarling (commented):
Just looking over some stuff via the web interface, these were some things that popped out at me. Overall looks pretty good, will try to take a closer look tomorrow

src/Netscript/WorkerScript.ts
d0sboots: Monomorphism. <https://www.builder.io/blog/monomorphic-javascript>

For objects that are hot, we want them to only have a single shape. I came across this at random while debugging something else, and decided to fix it while I was here. (Normally I'd use `null` instead of `undefined`, but this way is guaranteed to execute the same.)

src/NetscriptFunctions.ts
d0sboots: I struggled (and failed) to come up with a good name here. Do you have a suggestion?

src/Terminal/commands/kill.ts
d0sboots: Can't some of the underlying functions still throw? I thought I saw that when I checked. If it's really all no-throw, I'll take that out.

src/Terminal/commands/killall.ts
d0sboots: The underlying killWorkerScript always does the emit, so doing it in other places is just extra events.

src/NetscriptFunctions.ts
d0sboots: You were the one who put that in there. XD I think this was a case of overzealous type-checking, but I was able to work around it.

src/Terminal/commands/tail.ts
d0sboots: Done.

src/Netscript/killWorkerScript.ts
d0sboots: Done

src/NetscriptFunctions.ts
d0sboots: Done

src/Netscript/WorkerScript.ts
Snarling: Informative link, thanks

src/NetscriptFunctions.ts
Snarling: I would keep the name the same but rewrite it to actually take in a ScriptFilePath directly, do the escaping, and return the regex, instead of expecting the input to be already-escaped. Then the name would make sense with the function, and it would also be a far more usable function (would not be required to import a regex escaper every time you want to use one of these functions).

Same with the unanchored one - it should take in the filter pattern, then escape it, and then return an appropriate regex. This does mean the unanchored one would have to separately return a regex instead of wrapping the exact function.

d0sboots: I can make the ExactPath version take a FilePath, but the unanchored one can't, since at least some of its callers provide a pattern as opposed to a path. So then there's that weird dichotomy.

src/Terminal/commands/kill.ts
Snarling: Yeah it looks like you're correct. I don't think it's any of the functions you're using here, but no need to change this for now.

src/NetscriptFunctions.ts
Snarling: Yeah it's just a bit cumbersome to use the functions right now. But not a big deal, I'm just going to merge this in so we can get more playtesting time in on this before 2.3 goes live.

d0sboots: Makes sense, if you can think of better naming/a less awkward way to arrange this later, I'm all for it. It's mostly just `--grep` in ps that's messing things up.

PR #441, opened 2023-03-22
NETSCRIPT: Add ramOverride as a RunOption

d0sboots:
This lets you launch scripts with an arbitrary amount of RAM. If you use this, it's up to you to ensure that the script has enough RAM to pass the dynamic RAM check.

This is useful for large manager scripts where you might want to launch them with a lower cost when you aren't using their expensive functions (like corp). Also useful for a single HGW worker script that does a similar thing for each operation, but wants to use 1.7GB for hack.

Testing revealed issues with `WorkerScript.ramUsage` (it was possible to bypass the dynamic RAM check), so I removed it. There should be a single source of truth.


Snarling:
So I didn't carefully review the previous typechecking helper and only just now saw the `any` types. IMO `any` doesn't belong anywhere in a typechecker so I added a new utility type that should make it easier to typecheck objects without resorting to any and I refactored that typechecking helper. While I was searching to make sure I eradicated the `any`s I also changed a small amount of unrelated jsdoc.

At some point we'll probably want to add more supporting documentation, as right now the only player facing doc stating that you can override ramcost is the documentation on the RunOptions object.

d0sboots:
Maybe I'm just old-school, but I think it's kinda rude to change someone's PR in a way that's beyond correcting typos or other non-controversial stuff, and then merge it before they have a chance to respond.

Yes, you are the maintainer so you have the ability to submit whatever you want, but that would go in with your name instead of the PR author's.

Snarling:
Figured you just would have liked it merged sooner instead of me telling you I don't like the helper, and then you having to figure out a way to do it without `any` or starting an argument that `any` is fine. Especially given that the second scenario seemed more likely, I just made the changes and then merged. I guess I could also have merged the bad helper and then just fixed it after.

d0sboots:
See, that's what happens when you assume. :) I actually completely agree that `any` should be removed, it was some of the other changes I was objecting to.

PR #447, opened 2023-03-25
 BUGFIX: Fixed issue where duplicate programs could be created

### LINKED ISSUE #346: BUG : Can have multiple copies of the same program
Bitburner v2.2.1 (c46cedd5) (Steam)

So I was creating the FTP program, then I started doing contracts for Cybersec, then I returned to finish the FTPCrack.exe, but it finished without finishing...

image

image

Attached save file...

[bitburnerSave_1675041486_BN1x1.zip](https://github.com/bitburner-official/bitburner-src/files/10540832/bitburnerSave_1675041486_BN1x1.zip)

update: It was probably an issue deleting the temporary incomplete file at some point I guess. I manually removed the file and no error happened, the incomplete message was gone from the programs list and the finished exe worked ok.

Minzenkatze:
# Linked issues

fixes #346

# Bug fix

- The issue was that clicking "doing something else" after starting, and then "create program" would not add the program to the Player.getHomeComputer().programs array, before the new CreateProgramWork Instance would check the array. The fix was to check if the current work was still on a program, and if so, call finishWork before creating a new instance when clicking the create program button.

### DIFF
--- a/src/Programs/ui/ProgramsRoot.tsx
+++ b/src/Programs/ui/ProgramsRoot.tsx
@@ -10,7 +10,7 @@ import { Player } from "@player";
 import { Settings } from "../../Settings/Settings";
 
 import { Programs } from "../Programs";
-import { CreateProgramWork } from "../../Work/CreateProgramWork";
+import { CreateProgramWork, isCreateProgramWork } from "../../Work/CreateProgramWork";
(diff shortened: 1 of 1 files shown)


Snarling:
Test failure was not due to anything in the PR, looks to maybe be a service being down. Merging.

Great first PR!

PR #449, opened 2023-03-26
UI: Changes to UI on Create Program tab if a program is currently being created.

Minzenkatze:
I think it would look nicer if you could see your current progress on the Create Program tab instead of the way it is now. I also changed the text of the "Create Program" Button if that Program is already being created. 
Here's how the changes look:
Resume_focus_button_gif


Snarling:
Looks great!

Only critique I have is that when it's a "resume focus" button, that's all the button should do, not the whole process it does for create program. Right now the create program routine might functionally do the same as resuming focus, but that's not guaranteed to remain the same (e.g. if a dialog is added for when you stop working on a program to let you know how much progress you made, that dialog will pop up in the "resume focus" case when it shouldn't).

So I would stuff the entire button inside a ternary and just conditionally make an entirely different button if it's the currently worked program.

Minzenkatze:
That's a good point, i added those changes.

PR #450, opened 2023-03-26
NETSCRIPT: reduce RAM cost for renamePurchasedServer to 0

Zelow79:
I personally feel this cosmetic of a function at 2gb of ram makes it pretty much unusable. I personally see no harm in making it cost 0 RAM or if it must cost something 1 or less RAM. Just a suggestion, feel free to shoot it down if it doesn't align with your personal vision for the game.

image

image

Post Update:
image

### DIFF
--- a/src/Netscript/RamCostGenerator.ts
+++ b/src/Netscript/RamCostGenerator.ts
@@ -494,7 +494,7 @@ export const RamCosts: RamCostTree<NSFull> = {
   getPurchasedServerCost: RamCostConstants.GetPurchaseServer,
   getPurchasedServerUpgradeCost: 0.1,
   upgradePurchasedServer: 0.25,
-  renamePurchasedServer: 2,
+  renamePurchasedServer: 0,
   purchaseServer: RamCostConstants.PurchaseServer,
   deleteServer: RamCostConstants.PurchaseServer,
   getPurchasedServers: RamCostConstants.PurchaseServer,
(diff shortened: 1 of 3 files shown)


Snarling:
> image

Agreed.

> True if the upgrade succeeded, and false otherwise.

I'll merge it but only if you fix that jank return value description, which seems to have been copied from upgradePurchasedServer

Zelow79:
> > image
> 
> Agreed.
> 
> > True if the upgrade succeeded, and false otherwise.
> 
> I'll merge it but only if you fix that jank return value description, which seems to have been copied from upgradePurchasedServer

I noticed that too after I submitted it, but was too lazy to go back and fix it. It was there before though. I will look into it now.

Zelow79:
Uhh... lol
image

Zelow79:
Maybe just do this?
image

Snarling:
I wasn't very clear there sorry, I didn't mean remove the return altogether (technically an API break if anyone was relying on the return value), just that the description "True if the upgrade succeeded, and false otherwise." was bad because it's not performing an upgrade, it's just renaming the server.

But like you said it wasn't your mistake, it was an existing issue. I'm just going to merge this and then perform some additional changes afterwards.

Zelow79:
Oof, so in hindsight I did a lot of extra work for nothing. I could have just changed `upgrade` to `rename` succeeded.... /facepalm. So, if I were you, I would just exclude that last commit I sent and just do that. Change "upgrade" to "rename"

Snarling:
Yeah it was easy enough to just restore it. Sorry for the miscommunication

Zelow79:
No worries, probably more my fault lol. Just happy to see this merged tbh. I can have a lot more fun with it now that it will be reasonably priced.

PR #452, opened 2023-03-28
REFACTOR: Break the import cycle of Player

d0sboots:
This reorders the dependency between `Player` (the global object) and `PersonObjects/Player/PlayerObject` (the implementation) so that instead of going `Player` -> `PlayerObject`, it now goes `PlayerObject` -> `Player`.

Since `PlayerObject` is a massive class, it depends on most things throughout the game, and at the same time most things depend on `Player`. This massive dependency cycle permeated most things, and prevented a lot of classes from being imported standalone without ordering issues.

Added 3 tests on important mid-level classes that used to fail to import.

----
## Testing

This change exposed a latent bug in `AllServers.ts`: It is importing `RunningScript.ts` for the type `RunningScript`, but also depending on execution of `RunningScript` for the side-effect of setting the property on the Reviver so that it loads correctly.
Since it is using `import type`, that side-effect doesn't happen since the import doesn't exist at run-time. Interestingly, changing to `import` doesn't change things: Both TypeScript and Babel prune imports if it can identify that you are using them to import types only, and this saves *tons* of circular import headaches.

Before, the large circular import caused by `Player` caused the side-effect to happen due to an import of `RunningScript` through another path. Now, that doesn't happen. In the full game, `RunningScript` will also be imported through another path, but it is not good to depend on this.

What's especially of note is that this was caught *only* by the new Save test, so it is already showing its worth...

### DIFF
--- a/src/Achievements/Achievements.ts
+++ b/src/Achievements/Achievements.ts
@@ -1,4 +1,3 @@
-import { PlayerObject } from "../PersonObjects/Player/PlayerObject";
 import { AugmentationNames } from "../Augmentation/data/AugmentationNames";
 import { SkillNames } from "../Bladeburner/data/SkillNames";
 import { Skills } from "../Bladeburner/Skills";
@@ -25,6 +24,8 @@ import { FactionNames } from "../Faction/data/FactionNames";
 import { BlackOperationNames } from "../Bladeburner/data/BlackOperationNames";
 import { isClassWork } from "../Work/ClassWork";
 import { BitNodeMultipliers } from "../BitNode/BitNodeMultipliers";
+
+import type { PlayerObject } from "../PersonObjects/Player/PlayerObject";
 
 // Unable to correctly cast the JSON data into AchievementDataJson type otherwise...
 const achievementData = (<AchievementDataJson>(<unknown>data)).achievements;
--- a/src/PersonObjects/Player/PlayerObject.ts
+++ b/src/PersonObjects/Player/PlayerObject.ts
@@ -6,6 +6,7 @@ import * as generalMethods from "./PlayerObjectGeneralMethods";
 import * as serverMethods from "./PlayerObjectServerMethods";
 import * as workMethods from "./PlayerObjectWorkMethods";
 
+import { setPlayer } from "../../Player";
 import { Sleeve } from "../Sleeve/Sleeve";
 import { PlayerOwnedSourceFile } from "../../SourceFile/PlayerOwnedSourceFile";
 import { Exploit } from "../../Exploits/Exploit";
@@ -171,5 +172,7 @@ export class PlayerObject extends Person implements IPlayer {
     return player;
   }
 }
+
+setPlayer(new PlayerObject());
 
 Reviver.constructors.PlayerObject = PlayerObject;
--- a/src/PersonObjects/Player/PlayerObjectAugmentationMethods.ts
+++ b/src/PersonObjects/Player/PlayerObjectAugmentationMethods.ts
@@ -1,6 +1,7 @@
 /** Augmentation-related methods for the Player class (PlayerObject) */
-import { PlayerObject } from "./PlayerObject";
 import { calculateEntropy } from "../Grafting/EntropyAccumulation";
+
+import type { PlayerObject } from "./PlayerObject";
 
 export function applyEntropy(this: PlayerObject, stacks = 1): void {
   // Re-apply all multipliers
--- a/src/PersonObjects/Player/PlayerObjectBladeburnerMethods.ts
+++ b/src/PersonObjects/Player/PlayerObjectBladeburnerMethods.ts
@@ -1,5 +1,6 @@
 import { Bladeburner } from "../../Bladeburner/Bladeburner";
-import { PlayerObject } from "./PlayerObject";
+
+import type { PlayerObject } from "./PlayerObject";
 
 export function canAccessBladeburner(this: PlayerObject): boolean {
   return this.bitNodeN === 6 || this.bitNodeN === 7 || this.sourceFileLvl(6) > 0 || this.sourceFileLvl(7) > 0;
--- a/src/PersonObjects/Player/PlayerObjectCorporationMethods.ts
+++ b/src/PersonObjects/Player/PlayerObjectCorporationMethods.ts
@@ -3,8 +3,9 @@ import {
   CorporationUnlockUpgradeIndex,
   CorporationUnlockUpgrades,
 } from "../../Corporation/data/CorporationUnlockUpgrades";
-import { PlayerObject } from "./PlayerObject";
 import { resetIndustryResearchTrees } from "../../Corporation/IndustryData";
+
+import type { PlayerObject } from "./PlayerObject";
 
 export function canAccessCorporation(this: PlayerObject): boolean {
   return this.bitNodeN === 3 || this.sourceFileLvl(3) > 0;
--- a/src/PersonObjects/Player/PlayerObjectGangMethods.ts
+++ b/src/PersonObjects/Player/PlayerObjectGangMethods.ts
@@ -1,9 +1,10 @@
 import { Factions } from "../../Faction/Factions";
 import { Faction } from "../../Faction/Faction";
 import { Gang } from "../../Gang/Gang";
-import { PlayerObject } from "./PlayerObject";
 import { GangConstants } from "../../Gang/data/Constants";
 import { isFactionWork } from "../../Work/FactionWork";
+
+import type { PlayerObject } from "./PlayerObject";
 
 export function canAccessGang(this: PlayerObject): boolean {
   if (this.bitNodeN === 2) {
--- a/src/PersonObjects/Player/PlayerObjectGeneralMethods.ts
+++ b/src/PersonObjects/Player/PlayerObjectGeneralMethods.ts
@@ -1,4 +1,3 @@
-import { PlayerObject } from "./PlayerObject";
 import { applyAugmentation } from "../../Augmentation/AugmentationHelpers";
 import { PlayerOwnedAugmentation } from "../../Augmentation/PlayerOwnedAugmentation";
 import { AugmentationNames } from "../../Augmentation/data/AugmentationNames";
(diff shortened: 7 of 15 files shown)


Snarling:
> This reorders the dependency between Player (the global object) and PersonObjects/Player/PlayerObject (the implementation) so that instead of going Player -> PlayerObject, it now goes PlayerObject -> Player

This is a good thought but it will be more consistent if we initialize Player inside the engine loop, instead of in top-level code that now has its order of execution depend on when the PlayerObject file happens to get imported. 

To prevent circular imports having an effect, the rest of the codebase has typically moved all initialization code out of top-level scope in a file and into an initialization function, and then have that initialization function get called in the engine loop. Not sure why we never did this with Player.

> This change exposed a latent bug in AllServers.ts: It is importing RunningScript.ts for the type RunningScript, but also depending on execution of RunningScript for the side-effect of setting the property on the Reviver so that it loads correctly.

Yeah so this is a big issue with a lot of these classes. A bunch of files with classes use top-level code to set their revivers, and top-level code is what causes import cycle issues. I think an ideal solution here (doesn't need to be changed in this PR, just a thought for the future) would be to move the top level code out of these class files, have the reviver import every class that has a reviver, and then set the revivers in an initialization function that could be called by the engine (or called by a test that requires the revivers to be initialized).

> Since it is using `import type`, that side-effect doesn't happen since the import doesn't exist at run-time.

Yeah I think there are a lot of places throughout the codebase where we should be using `import type` instead of a regular import for classes and enums. I think classes and enums are the only times it actually matters that we use that syntax, since they both can be used just as types but also might be used for runtime.

> Adding three tests that just make sure 3 specific files can be imported

Somewhat bizarre... but we can keep these for now and once we deal with more potential import issues by getting rid of as much top-level code as possible, these can probably be retired at that point.

d0sboots:
> This is a good thought but it will be more consistent if we initialize Player inside the engine loop, instead of in top-level code that now has its order of execution depend on when the PlayerObject file happens to get imported.

That *is* how this is working. `engine.tsx` imports PlayerObject directly for its side-effect of initialization. Just like in C++, we should be assuming that anything that is initialized at the top-level is "unordered" with respect to any other top-level initialization, and thus there should be no ordering required between them. Player and Revivers is fine for this (see below).

> To prevent circular imports having an effect, the rest of the codebase has typically moved all initialization code out of top-level scope in a file and into an initialization function, and then have that initialization function get called in the engine loop. Not sure why we never did this with Player.

That's nearly what's going on here. The difference is that there's no explicit `initPlayer()` function that you can forget to call; instead the initialization happens as a side-effect of importing PlayerObject.

Part of the reason for this is that the Player object is *so* fundamental, almost nothing else works if it isn't present. I'm not sure if even initializing it at the very beginning of `load()` would be enough; it's possible that it's already accessed before that point. Certainly `startExploits` uses it, so it would have to be the very first thing at minimum.

> Yeah so this is a big issue with a lot of these classes. A bunch of files with classes use top-level code to set their revivers, and top-level code is what causes import cycle issues. I think an ideal solution here (doesn't need to be changed in this PR, just a thought for the future) would be to move the top level code out of these class files, have the reviver import every class that has a reviver, and then set the revivers in an initialization function that could be called by the engine (or called by a test that requires the revivers to be initialized).

I think this would be a step backwards. Setting the reviver is logically "part" of the class; it belongs in the same file where the class is defined. Doing it elsewhere greatly increases the risk that one will be missed.

More importantly, if the reviver file is setting the revivers for all the classes, it needs to *import* all the classes. But the classes also need to import the reviver, in order to deserialize themselves. This makes the reviver a giant dependency cycle, nearly as big as the one I'm getting rid of here for Player.

Snarling:
Fair enough with the player stuff, it's close enough that it's fine. The changes you made are certainly an improvement from how it was before, so even though it's not consistent with the rest of the codebase I'm not going to push on this one.

Merging because this is all the PR actually changes, the reviver discussion is a separate thing.

> > Classes shouldn't run top level code
> 
> I think this would be a step backwards. Setting the reviver is logically "part" of the class; it belongs in the same file where the class is defined. Doing it elsewhere greatly increases the risk that one will be missed.
> 
> More importantly, if the reviver file is setting the revivers for all the classes, it needs to _import_ all the classes. But the classes also need to import the reviver, in order to deserialize themselves. This makes the reviver a giant dependency cycle, nearly as big as the one I'm getting rid of here for Player.

Import cycle does not matter if there's no top level code being ran that actually uses those imports. Having top level code (in this case, every single class having top level code referencing the reviver import) creates a potential for an import cycle issue to occur, where there would be absolutely zero chance of an import cycle issue if an initializer function is used.

Also, imagine a full-savegame test. This test would currently need to ensure that it positively imports every single file that has a reviver function, otherwise it doesn't work. If there's a reviver initializer, it just needs to import that and perform the initialization.

Edit: Unless for some reason just defining a class counts as top level code that actually requires those imports to be fully loaded, which is possible. Stuff referenced inside a function is not required to be fully loaded during the import cycle (until that function gets ran), but maybe classes work differently.

d0sboots:
> Edit: Unless for some reason just defining a class counts as top level code that actually requires those imports to be fully loaded, which is possible. Stuff referenced inside a function is not required to be fully loaded during the import cycle (until that function gets ran), but maybe classes work differently.

This is part of it. The class definition *itself* is top-level code, just like function definitions are top-level code. You need to import the class definition if you are extending or making `new` instances of the class, just like you need to import a function if you will call that function.

> Import cycle does not matter if there's no top level code being ran that actually uses those imports.

No, it absolutely matters because the cycle means that anything that imports *anything* from *any part* of the cycle is now bringing in the entire cycle, and everything it depends on. Due to how BB works, that essentially means the entire game for high-level cycles like `Player` or this hypothetical `Reviver` cycle. It also means that you can run into order-of-import issues for stuff like `Server` that depends on `BaseServer` already being defined; this is why I wrote those tests. (They didn't pass before this PR.)

> Having top level code (in this case, every single class having top level code referencing the reviver import) creates a potential for an import cycle issue to occur, where there would be absolutely zero chance of an import cycle issue if an initializer function is used.

No, because the imports run the same direction. All of these classes need to import the Reviver already in order to implement `fromJSON`. By having the reviver namespace be set in the same file, it creates a new dependency where that namespace must be imported, but that exists in the same file: No new imports are added.

Import cycles are strictly avoided, currently, because Reviver is a utility class that sits at the edge of the import graph, depending on very little.

d0sboots:
Forget to respond to this part:

> Also, imagine a full-savegame test. This test would currently need to ensure that it positively imports every single file that has a reviver function, otherwise it doesn't work. If there's a reviver initializer, it just needs to import that and perform the initialization.

The correct answer here is that the game code itself should ensure the imports are correct, and the test would test for this by not importing anything extra (and asserting that it works correctly). This is how the current Save test is working, on a more limited scale.

This isn't that hard to arrange; In most cases, the "container" of a class also creates instances of that class, and so its import will not be type-only, and thus the import will naturally guarantee that the reviver is set properly. The issue with `RunningScript` was that its container, `Server`, never directly constructs `RunningScript` and so the import gets elided. (And this was fixed by adding an explicit side-effect import with a comment.)

Also, this isn't a concern for the actual game, since all these classes are imported for real *somewhere*, and so the reviver is always set in practice. It's more of a test/code-hygiene issue.

Snarling:
> The correct answer here is that the game code itself should ensure the imports are correct, and the test would test for this by not importing anything extra (and asserting that it works correctly). This is how the current Save test is working, on a more limited scale.

Agreed that the game needs to make sure imports are correct. The simplest way to do this is an initializer in one location instead of import-for-effect statements littered throughout the codebase. There was no issue in the actual game, but there was an issue in your test because you were working with an incomplete reviver which the game never does.

Here are the competing solutions for the right revivers possibly not being initialized:

1. Your solution: require "import for side effects only" code in various places to ensure that every class also imports any files with classes that could be included in that reviver. With this solution, for instance, player should import e.g. every single type of player work because those might be included in a player. This solution requires judgement calls all over the place, and if we get those judgement calls wrong then we could end up with an incomplete reviver in a test.
2. My solution: a reviver initializer. This does mean any test that uses the reviver will end up importing a lot of stuff from the game - specifically everything that is needed to construct a complete reviver. This is desirable, not undesirable. It ensures that any test using the reviver uses one that is fully set up just like the one in game. It eliminates issues like the one you encountered here where you had to precision-place an import-for-side-effects statement to initialize the right reviver that was getting missed. A test writer can never fail to import part of the reviver if it's all done at once every time.

Especially as we address more of the instances where "import" should be changed to "import type" more of these issues will potentially pop up if we do not move to a reviver initializer. Right now much of the initializing code is probably getting activated by accident just because we're indiscriminately importing the real class in places where we only need to import the type. A full savegame test with solution 1 needs to import everything that activates reviver initialization code before it can safely say the reviver is initialized. A full savegame test with solution 2 just needs to import one initializer and after running that, it's all there just like in game.

I don't believe that a reviver initializer would cause "used before defined" errors which are really the only potential problem with circular imports. These used before defined issues are prevented by not referencing the imported stuff in top level code, solved by using an initializer function. It's true that this solution casts a wide import net, but that net is exactly the size and shape required to fully initialize the reviver, which should be exactly what we want. We should not need to surgically select the exact things that need to be imported for a specific initializer like you did in this PR. It's okay if we use a fully initialized reviver even on tests that aren't related to the entire save file.

d0sboots:
> Your solution: require "import for side effects only" code in various places to ensure that every class also imports any files with classes that could be included in that reviver. With this solution, for instance, player should import e.g. every single type of player work because those might be included in a player. This solution requires judgement calls all over the place, and if we get those judgement calls wrong then we could end up with an incomplete reviver in a test.

There's no judgement call involved, because the tests will tell you directly if you got it wrong. Since this is strictly a code hygiene issue/testing issue, and not one that can/will cause problems for the actual game, I don't see a reason to take action on it prematurely. 

Or, put differently: If we started adding imports for side effects in advance of tests that showed they were needed, that *would* be making arbitrary judgement calls. But it would also be putting the cart before the horse: The tests are what have value, and the code restructuring is just to allow it to be more testable.

d0sboots:
I should also say: I don't think having a reviver initializer is terrible. It would just need to be in its own file to avoid cycles, and it would be extra setup for any test dealing with loading.

d0sboots:
> Right now much of the initializing code is probably getting activated by accident just because we're indiscriminately importing the real class in places where we only need to import the type.

This is a misconception. Both TypeScript and Babel prune imports of types (I.e. classes) that are used only for their types. This means that although it would be better to use "import type" more, to be explicit, it's not strictly necessary.

The only places the imports actually exist are where `new` instances are being made, or the class is extended, or something else (constants/functions) is also imported from the file. But, since every class must be constructed *somewhere*, that's enough to guarantee a real import exists. (And generally where we want it.)

PR #453, opened 2023-03-28
SINGULARITY: Added run and install identifier

G4mingJon4s:
# Added run and install identifier

The player object now holds a run and install identifier. These get updated when entering a new bitnode or augmentations were installed. This allows the user to distinguish between different runs and installs easily, where as before the way of keeping track of such changes was unreliable.

# Actions
- [X] Added fields on player object
- [X] Added updating fields
- [X] Added api function to singularity
- [X] Added ram costs
- [X] Added documentation
- [X] Tested various cases, such as recovery and bitflumes
- [X] Ran `npm run format`
- [X] Ran `npm run lint`
- [x] Ran `npm run doc`

# Todo
- run `npm run doc` ? Not sure if this is necessary.


Snarling:
Yes you should `npm run doc`, it should only create those two new docs and modify the singularity doc to include references to the new functions. It'll look like it's modifying a bajillion things in the process but all the other changes should disappear.

As far as the PR content, I like the idea of providing some sort of identifier so a script can more easily tell if it's being reinitialized due to a page reload or due to an aug reset or a node reset. Not super keen on using a random string identifier though.

I think the ideal way to provide this would be through something like lastAugReset and lastNodeReset, each holding a numeric date (from Date.now()). Then it's not only an identifier, but also provides another use and you can imagine what it is just by the property name. The problem there is that the functionality overlaps with playtimeSinceLastX properties, which are currently given to the player in ns.getPlayer (although IMO they shouldn't be). So if adding these extra date properties, it would make sense to remove the playtimeSinceLastX properties but that would be an API break. possibly we could

I do agree that identifiers in whatever form they come should be provided through singularity, especially since the existence of bitnodes is a spoiler for basegame API (which is one of the issues with playtimeSinceLastBitnode being provided on ns.getPlayer).

Ignoring the possibility of using last reset times instead of random IDs, I'm not really happy with the property name runIdentifier. I know we commonly call each bitnode attempt a "run" but to me that property name does not sufficiently explain what it is, especially when "run" has other usage in the API.

Open to any thoughts anyone else has about this.

G4mingJon4s:
Yeah, providing the identifier through time is also a great idea.
Moving the properties in `ns.getPlayer` to their own functions in the singularity interface is probably the best choice here.
Making these accessable before beating the first node would spoil a lot...

Since there is already `ns.getTimeSinceLastAug`, there is already an identifier for each install.
For the current run, `ns.getTimeSinceLastNode` might be an option. This would have the same basic functionality of `ns.getTimeSinceLastAug`, but provides the time since the last node.
The question here is where to put it. Making it a singularity function weirdly separates these two functions. Whilst making the other a singularity function too would break a lot.

Snarling:
The only problem with "getTimeSince" functions is that this is a number that changes as time goes on, so it's less useful as an "identifier". Even if we provide direct access to lastUpdate, I don't even think `lastUpdate - ns.getTimeSinceLastAug()` is going to always result in the same number 100% of the time, I think it could wiggle by up to 200ms. So in that sense it's not as much of an "identifier" as it should be since it isn't remaining the same.

If we are okay with removing the two times from getPlayer, then we can shift internally to a lastAugReset/lastNodeReset that would be numbers that remain the same during the entire aug/node run, instead of the playtimeSinceLastX properties that update every engine tick. We could keep the getTimeSinceLastAug function (it can return lastUpdate - lastAugReset) and then on singularity provide something like getLastResetTimes() that provides both of the last reset times. Then even though it's partially duplicated functionality between the main API and singularity, it at least doesn't look like as much of an inconsistency to do that. Removing the playtimeSinceX properties from getPlayer is still an API break, but I don't think those properties should be on there in the first place (especially since one of them is a spoiler).

G4mingJon4s:
So calculating the time when the install happened is actually pretty accurate, but any lags gamewise might make it inaccurate.
I tried this, and it gave me the same value every time (without any lags):
```ts
const times = [];
for (let i = 0; i < 1000; i++) times.push(Date.now() - ns.getTimeSinceLastAug());

const v = times[0];
ns.tprint(times.every(t => t === v)); // true
```

Snarling:
It's because you're running every one of those at the same time, so as long as those 1000 calls take less than 1ms, the times will all be identical. Actually surprising that all 1000 of them can run in the same ms though.

Here's what I got with a similar test but running the 1000 calls spaced out across a span of ~10s:
```js
export let main = async ns => {
  const times = [];
  for (let i = 0; i < 1000; i++) {
    await ns.asleep(10);
    times.push(Date.now() - ns.getTimeSinceLastAug());
  }
  const maxTime = Math.max(...times);
  const minTime = Math.min(...times);
  ns.tprint(`Wiggle: ${maxTime - minTime}`);
}
```
image

G4mingJon4s:
Ah, good to know. Well that means we can't really use these functions as identifiers...

So, we either use a random string like I implemented or give the player the exact timestamp when the install/node happened/began...
Or someone has a better alternative ofc...

G4mingJon4s:
So, I added the markdown from `npm run doc`. I know this is not really a good implementation right now, but I'm happy to implement any changes. I will look into getting the exact timestamp of the install/node reset.
If I should add anything, just contact me here or on discord (`G4mingJon4s#6950`, ping on server or dm).

Snarling:
@G4mingJon4s Sorry, haven't followed up on this one in a bit.

I think it makes sense to:
1. Remove player.playtimeSinceLastX properties
2. Make new player.lastXReset properties
3. in SaveObject.ts under the version <31 if statement, would need to add code to convert the old time entries to the new entries (lastUpdate-playtimeSinceLastX) and delete the old entries. Since we're not incrementing the version yet this code also needs to be able to just do nothing if a player's save has already been converted to the new format.
4. Remove the old properties from player interface in NetscriptDefinition, only adding back the lastAugReset property. I have an idea for making this less of a hard API break but I'd need to mess around with it so I'll try that out after.
5. Provide lastNodeReset via some singularity function
6. keep ns.getTimeSinceLastAug(), which would be deprecated and would return lastUpdate (or now) - lastAugReset
7. Whatever places the playtimeSinceLastX properties are directly used would of course need to be reworked to use the new data. If this results in basically calculating playtimeSinceLastX a bunch of separate times, it may be worth doing the calculation once in the engine loop and then updating a augPlaytime exported variable or something, to use it in other places where it's needed.

Then after all that, lastReset times would essentially also be aug/node run identifiers. Sorry, it's kind of a lot of stuff to change. Let me know if you need/want assistance.

G4mingJon4s:
So, this is the basic change. I'm not quite sure how to implement the version change...
I couldn't find a `if (ver < 31)` anywhere, so I guess it has been implemented after I made the branch.
And I'm not quite sure if using `Date.now()` is the right way...

I'll lint and format after the version change and any further changes...

G4mingJon4s:
So I realized some things...
I shouldn't have removed the playTime properties...
I'll revert the last change...

G4mingJon4s:
I screwed up, big time, let me redo the pr completely

PR #458, opened 2023-04-01
TEST: Add a NS2 test (finally)

d0sboots:
This was much harder than it should have been, since it required working around https://github.com/nodejs/node/issues/35889 : Node.js has a bug open for 1.5 years (and it's been present for longer) where it just segfaults if you dynamic import() in certain circumstances. These circumstances are hit by basically all test runners, and certainly by Jest.

Also sped up tests by reducing the number of traversed directories (the "roots" option).

# Testing

Scripts still run (sanity check of the very small amount of game code touched)


Snarling:
Very nice!

PR #460, opened 2023-04-02
SINGULARITY: getCompanyPositionInfo

jeek:
SINGULARITY: getCompanyPositionInfo

Returns the requirements to apply for a specific position at a specific company.


src/Work/CompanyPositionRequirements.ts
```diff
@@ -0,0 +1,45 @@
...
+    nextPosition: c.nextPosition,
+    requiredHacking: c.requiredHacking > 0 ? c.requiredHacking + n : 0,
+    requiredStrength: c.requiredStrength > 0 ? c.requiredStrength + n : 0,
+    requiredDefense: c.requiredDefense > 0 ? c.requiredDefense + n : 0,
+    requiredDexterity: c.requiredDexterity > 0 ? c.requiredDexterity + n : 0,
+    requiredAgility: c.requiredAgility > 0 ? c.requiredAgility + n : 0,
+    requiredCharisma: c.requiredCharisma > 0 ? c.requiredCharisma + n : 0,
+    requiredReputation: c.requiredReputation,
+  };
+};
```
Snarling: So for adding a new thing that is provided to the player for use, I like to think about, how can I shape this so it's most useful to the player. Following the existing shape of the internal code is much less important than giving something the most convenient shape for use by the player. Internal code can be freely reshaped however we want, but changing the shape of stuff provided in the API is very difficult because it is an API break for player scripts.

For instance, we already have the Skills interface, which is what players are given for their own skills. Providing the required skills using the same structure as this existing interface is convenient. e.g. players can check if they meet the skill requirements for a job without needing to hardcode the interface keys by manually comparing their own "hacking" skill value against a "requiredHacking" value and so on for each of the other skills.

So the way I would shape the interface is 
```ts
interface CompanyPositionInfo {
  name: CompanyPosName;
  nextPosition?: CompanyPosName; // could also be a required property with CompanyPosName | null
  requiredReputation: number;
  requiredSkills: Skills;
}
```

Additionally, it looks like this is probably based on WorkStats, which is the only thing in the game with this wonky structure where it's an interface sort of pretending it's a class. Not really a good example to use as a base.

In this case, I would define the interface directly in NetscriptDefinitions (and only there), and if you need it in the Singularity ns file you can import it. For the first implementation I would not use any helper functions and just manually construct the correct shape in getCompanyPositionInfo, later if we decide to reshape some internal code to match the api, we could add a helper function and use it both in the NS function as well as in the internal code for checking job requirements.

Changing the internal code to a new/better structure can also be done at the same time, but it doesn't need to be.

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -1698,6 +1698,29 @@ export interface Singularity {
...
+   *
+   *
+   * This function will return an object that contains the requirements for
+   * a specific position at a specific country.
+   *
+   * @example
+   * ```js
+   * // NS2:
+   * const COMPANY_NAME = "ECorp";
+   * const POSITION = "Chief Executive Officer";
```
Snarling: Extremely minor gripes, but the // NS2 comment will also be removed from the "ns2" (js) examples, since all examples will just be for regular js scripts (the script type formerly known as ns2). Also variable names in examples should use camelCase for consistency (I wouldn't normally kick back a PR for this, just adding this comment because there are already other changes requested)

src/NetscriptFunctions/Singularity.ts
```diff
@@ -687,6 +689,41 @@ export function NetscriptSingularity(): InternalAPI<ISingularity> {
...
+
+      // Make sure its a valid company
+      if (companyName == null || companyName === "" || !Companies[companyName]) {
+        throw helpers.makeRuntimeErrorMsg(ctx, `Invalid company: '${companyName}'`);
+      }
+
+      // Make sure its a valid position
+      if (positionName == null || positionName === "" || !CompanyPositions[positionName]) {
+        throw helpers.makeRuntimeErrorMsg(ctx, `Invalid position: '${positionName}'`);
+      }
```
Snarling: This doesn't correctly narrow the type for positionName from `string` to `CompanyPosName`.
For typechecking something that has an enum defined, currently the solution is using `checkEnum`, see NetscriptFunctions/Infiltration.ts function `calculateInfiltrationData` for an example. After throwing an error if checkEnum returned false, the type is narrowed to a member of the provided enum.

There's another PR in progress that will replace checkEnum with something else, but the basic idea will remain the same.

Snarling (changes requested):
Let me know if you need/want any assistance or clarification on the requested changes

Snarling (approved):
Looks good, thanks!

PR #461, opened 2023-04-03
SCRIPTS: Script modules are reused when they are imported

### LINKED ISSUE #444: Race condition in `compile`
The following script demonstrates the issue. Running it produces the following:
>RUNTIME ERROR
> import_test_generated2.js@home (PID - 35)
> 
> Failed to fetch dynamically imported module: blob:http://localhost:8888/757b1b83-6450-44fb-bcf6-8770dd28964a
stack:
> TypeError: Failed to fetch dynamically imported module: blob:http://localhost:8888/757b1b83-6450-44fb-bcf6-8770dd28964a

```js
/** @param {NS} ns */
function createScript(num, ns) {
  ns.write(
    `import_test_generated${num}.js`,
    (num == 2 ? 'import "import_test_generated1.js";\n\n' : '') +
    `
/** @param {NS} ns */
export function main(ns) {}
    `.trim(),
    "w"
  );
}

/** @param {NS} ns */
  createScript(1, ns);
  createScript(2, ns);
  ns.run("import_test_generated2.js", 1, 1);
  const promise = Promise.resolve().then(() => {
    ns.run("import_test_generated1.js");
    ns.run("import_test_generated2.js");
  });
  ns.run("import_test_generated2.js", 1, 2);
  await promise;
}
```

The problem is a race condition in `compile` (in NetscriptJSEvaluator.js). I think that whole function needs a rework, I don't see a way it can be patched and still have high certainty that it won't have other race conditions.

Snarling:
Simplified the compile process, while also allowing modules to be reused regardless of where they are imported. Changes here also fixed #444.

I also added some code related to slimming down the save file, by only saving or loading the desired keys of an object. The following changes were made to save/load:

Script: only filename, server, and code are saved or loaded.
RunningScript: Dependencies and logs are not saved.

On the save test, removed the time freeze portion. This was not needed for the test (time is never updated since the engine loop is not engaged during the test) and also was creating a ts error in my IDE (even though ts is not actually enforced in the tests yet).

Also rewrote parts of the reviver, mostly for typesafety.


src/utils/JSONReviver.ts
```diff
@@ -1,85 +1,100 @@
...
+/**
+ * A generic "smart reviver" function.
+ * Looks for object values with a `ctor` property and a `data` property.
+ * If it finds them, and finds a matching constructor, it hands
+ * off to that `fromJSON` function, passing in the value. */
+export function Reviver(key: string, value: unknown): any {
+  if (typeof value !== "object" || value === null) return value;
+  // @ts-ignore We know it's an object here, just checking type of object properties to type validate.
+  if (typeof value.ctor !== "string" || typeof value.data === "undefined") return value;
+  const typedValue = value as IReviverValue;
```
d0sboots: If you flipped these two lines and checked on `typedValue`, you wouldn't need the @ts-ignore.

src/utils/JSONReviver.ts
```diff
@@ -1,85 +1,100 @@
...
+  const obj = ctor.fromJSON(typedValue);
+  if (ctor.validationData !== undefined) {
+    validateObject(obj, ctor.validationData);
   }
-  return value;
+  return obj;
 }
 
 // eslint-disable-next-line @typescript-eslint/no-namespace
 export namespace Reviver {
```
d0sboots: Very optional: Could we move to a plain object instead of a namespace? (It's preexisting, so no harm leaving it, but you were cleaning up a lot of related stuff...)

src/utils/JSONReviver.ts
```diff
@@ -1,85 +1,100 @@
...
-  export const constructors: { [key: string]: any } = {};
+  export const constructors: Partial<
```
d0sboots: What job is the `Partial` doing here? Aren't all entries in a record optional?

src/utils/JSONReviver.ts
```diff
@@ -1,85 +1,100 @@
...
+  obj: T,
+  keys?: readonly (keyof T)[],
+): IReviverValue {
+  const data = {} as T;
+  // keys provided: only save data for the provided keys
+  if (keys) {
+    for (const key of keys) data[key] = obj[key];
+    return { ctor: ctorName, data: data };
   }
+  // no keys provided: save all own keys of the object
```
d0sboots, line 68: This comment is incorrect; `for..in` is all *enumerable* keys, whether or not they are own keys. I'd either change the comment, or use `for..of Object.entries(obj)` to enumerate own key/value pairs.

src/utils/JSONReviver.ts
```diff
@@ -1,85 +1,100 @@
...
+export function Generic_fromJSON<T extends Record<string, any>>(
+  ctor: new () => T,
+  // data can actually be anything. We're just pretending it has the right keys for T. Save data is not type validated.
+  data: Record<keyof T, any>,
+  keys?: readonly (keyof T)[],
+): T {
+  const obj = new ctor();
+  // If keys were provided, just load the provided keys (if they are in the data)
+  if (keys) {
+    for (const key of keys) if (key in data) obj[key] = data[key];
```
d0sboots: Very optional: You can write this as
```
const val = data[key];
if (value !== undefined) obj[key] = value;
```
to avoid a double-lookup. V8 *might* be able to optimize it away, but I'm unsure because `in` is a different type of lookup.

src/utils/JSONReviver.ts
```diff
@@ -1,85 +1,100 @@
...
+    return obj;
   }
+  // No keys provided: load every key in data
+  for (const key in data) obj[key] = data[key];
```
d0sboots: Same note about entries, although here there's less potential for bugs because it's a JSON object. But `entries` lets you save a lookup.

src/Script/Script.ts
```diff
@@ -75,44 +52,40 @@ export class Script {
...
-  /**
-   * Marks this script as having been updated. It will be recompiled next time something tries
-   * to exec it.
-   */
-  markUpdated(): void {
-    this.module = null;
-    this.moduleSequenceNumber = ++globalModuleSequenceNumber;
+  /** Invalidates the current script module and related data, e.g. when modifying the file. */
+  invalidateModule(): void {
+    delete this.module;
```
d0sboots: I'm pretty sure dynamically changing the properties of an object by adding/deleting them is more expensive (on at least some engines) than just changing the values; it negates optimizations based on using a single prototype for the in-memory layout of the class. I would set these to `null`, `undefined` or 0 instead.

src/Script/Script.ts
```diff
@@ -5,58 +5,35 @@
...
-interface ScriptReference {
-  filename: string;
-  server: string;
-}
+export type ScriptURL = string;
 
 export class Script {
-  // Code for this script
   code = "";
+  filename = "default.js";
```
d0sboots, line 21: Isn't empty-string a better default?

src/Script/Script.ts
```diff
@@ -75,44 +52,40 @@ export class Script {
...
+  /** Invalidates the current script module and related data, e.g. when modifying the file. */
+  invalidateModule(): void {
+    delete this.module;
+    if (this.url) URL.revokeObjectURL(this.url);
+    delete this.url;
+    delete this.ramUsage;
+    delete this.ramUsageEntries;
+    for (const [url, dependency] of this.dependencies) {
+      dependency.dependents.delete(this);
+      this.dependencies.delete(url);
```
d0sboots: You can just clear the map after the loop.

src/Script/Script.ts
```diff
@@ -75,44 +52,40 @@ export class Script {
...
+    }
+    this.dependents.forEach((dependent) => {
+      this.dependents.delete(dependent);
```
d0sboots: Same; you can just clear the set after.

Edit: Maybe not, there could be subtle effects due to the recursion. The recursion here makes everything very hard to analyze for correctness.

src/Script/Script.ts
```diff
@@ -75,44 +52,40 @@ export class Script {
...
-    this.moduleSequenceNumber = ++globalModuleSequenceNumber;
+  /** Invalidates the current script module and related data, e.g. when modifying the file. */
+  invalidateModule(): void {
+    delete this.module;
+    if (this.url) URL.revokeObjectURL(this.url);
+    delete this.url;
+    delete this.ramUsage;
+    delete this.ramUsageEntries;
+    for (const [url, dependency] of this.dependencies) {
+      dependency.dependents.delete(this);
```
d0sboots: Something subtle that is probably worth documenting: This is over-eager. Because the dependencies are both direct *and* indirect, this loop can remove items from other script's `dependents` when those items are still dependent through other paths.

I *think*, but am not 100% certain, that this is OK because it will be fixed up when `compile` is run.

src/Netscript/NetscriptHelpers.ts
```diff
@@ -248,20 +248,20 @@ function makeRuntimeErrorMsg(ctx: NetscriptContext, msg: string, type = "RUNTIME
...
+      if (ws.scriptRef.url && stackline.includes(ws.scriptRef.url)) return ws.scriptRef.filename;
+      // Also check urls for dependencies
+      for (const [url, script] of ws.scriptRef.dependencies) if (stackline.includes(url)) return script.filename;
+      // Check for filenames directly if no URL found
+      if (stackline.includes(ws.scriptRef.filename)) return ws.scriptRef.filename;
+      for (const script of ws.scriptRef.dependencies.values()) {
+        if (stackline.includes(script.filename)) return script.filename;
       }
     }
+    const filename = getFileName();
```
d0sboots: Probably clearer as an IIFE (Immediately Invoked Function Expression), that way you don't wonder why it's a named function.

src/Netscript/NetscriptHelpers.ts
```diff
@@ -248,20 +248,20 @@ function makeRuntimeErrorMsg(ctx: NetscriptContext, msg: string, type = "RUNTIME
...
-      }
-      for (const dependency of script.dependencies) {
-        if (stackline.includes(dependency.filename)) {
-          filename = dependency.filename;
-        }
+    function getFileName() {
+      // Filename is current file if url found
+      if (ws.scriptRef.url && stackline.includes(ws.scriptRef.url)) return ws.scriptRef.filename;
+      // Also check urls for dependencies
+      for (const [url, script] of ws.scriptRef.dependencies) if (stackline.includes(url)) return script.filename;
```
d0sboots, line 257: Style: I'm personally allergic to one-line if statements, but I accept they're probably actually fine, especially for cases like early return. One-line `for`s are a bit too far IMO, especially chained as a for-if.

d0sboots (commented):
I took the coward's way out and reviewed everything except NetscriptJSEvaluator.

I'm concerned that I'm going to find a potential design issue, because I gave this a lot of thought already and there are a *lot* of mostly-working solutions that fall down in various subtle edge cases; please give me another day to look at this closely.

src/utils/JSONReviver.ts
Snarling: Yeah I was actually going to do this, I only didn't because it would need to edit a bunch of files that are ultimately unrelated. There's no reason the constructors need to be a property of the Revivers function (that I'm aware of)

src/utils/JSONReviver.ts
Snarling: Record<string, X> tells ts that any string used as a key for the object will have an associated value of type X. Partial<Record<string, X>> says actually, some string keys do not have defined values (but for the ones that are defined, they will be of type X).
```ts
const a: Record<string, string> = {testKey: "testVal"};
const b: Partial<Record<string, string>> = {testKey: "testVal"};

let c = a[someRandomString] // ts will interpret this as string
let d = b[someRandomString] // ts will interpret this as string|undefined
```

src/utils/JSONReviver.ts
Snarling: Yes but then I would still need a comment explaining why I'm asserting the type as X before I actually typecheck that the type actually is X. In my view asserting type IReviverValue needs to be after we've already typechecked that it is in fact an IReviverValue

src/utils/JSONReviver.ts
Snarling: Good catch. I didn't realize Object.entries/keys limits to own properties but for...in does not. Will change this to an Object.entries.

src/utils/JSONReviver.ts
d0sboots: Hmm. What about defining the if-check as an (inline) guard function (https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates), and then after the early return TS itself will know that `value` is an IReviverValue? Then you can also dispense with the extra variable.

src/Script/Script.ts
Snarling: Actually it looks like clearing the set isn't even needed. This will be done automatically when the dependents invalidate their own modules (since the invalidation includes going through all dependencies and clearing self from the dependency's dependents set)

src/utils/JSONReviver.ts
Snarling: Yeah a typeguard function is the right way to handle it... ugh. Implemented

d0sboots: > ugh

It doesn't have to be a big deal, if you IIFE it it's only a little bit longer (basically adding `function` and the TS return guard type) :)

src/Script/Script.ts
Snarling: The only thing being removed from other scripts' dependents are scripts that are currently being invalidated, which will only be the edited script and anything higher in the import tree. No scripts that remain valid will be removed as dependents.

`dependents` is only used for passing invalidation up the import tree, so there is no need or desire to store dependents for any scripts that are already invalidated.

src/Netscript/NetscriptHelpers.ts
Snarling: Done c1183c8

src/Script/Script.ts
Snarling: I did change up the invalidation function a little in f86f6f5 to prevent re-invalidating the same script a bunch of times in instances where there's a complex import tree. Now there's an early return if the script has already been invalidated, and also over in NetscriptJSEvaluator we're now only storing direct dependents instead of deep-dependents. Since invalidation is passed to all dependents, it will naturally reach all deep dependents anyway, this just limits the amount that we try invalidating the same script multiple times when a file is edited.

src/utils/JSONReviver.ts
Snarling: Done in 42b7d09

src/utils/JSONReviver.ts
Snarling: Done in 34b5b77

Snarling:
> I took the coward's way out and reviewed everything except NetscriptJSEvaluator.
> 
> I'm concerned that I'm going to find a potential design issue, because I gave this a lot of thought already and there are a _lot_ of mostly-working solutions that fall down in various subtle edge cases; please give me another day to look at this closely.

No rush. I really appreciate the thorough review so far. NetscriptEvaluator.js should be a bit easier to review now too.
* Separate short function for setting dependency relationships for a script once its url is assigned.
* The compile function is really short now. Basically, if there's a module, return that. If not, import an existing URL if present or a newly generated one if not, and assign it as the script's module. Assigning a new URL is a synchronous operation, so no chance for two different urls to be generated.
* The first few lines and last few lines of generateScriptUrl are really the only new code there. The function used to be called _getScriptUrls and used to return a whole array, now it just returns the requested script's url. The whole middle section using Acorn to find and replace import text with script URLs was hardly changed. The whole function used to be in a giant try/catch which was no longer needed so the indentation changed, that's why the diff makes it look like way more changed than actually did.

src/utils/JSONReviver.ts
Snarling: Done

src/Script/Script.ts
Snarling: Yeah looks like you're right, I change it to set them undefined instead.

src/Script/Script.ts
Snarling: While the type right now is just "string" really the type for filename is/should be more constrained than that. If we had a ScriptFilename type validator, default.js would pass but empty string would not. Same reason I changed server from empty string to home, if there was a ServerName type validator, empty string would not pass.

src/Script/Script.ts
Snarling: Individual removing of dependents from set was removed on 42b7d09, since it happens in invalidateModule calls on the dependent scripts

src/Script/Script.ts
Snarling: Resolving this one, but if you do find that there's a failure case I'm just not seeing, either reopen or just start a new thread

src/Server/ServerHelpers.ts
```diff
@@ -256,7 +256,8 @@ export function prestigeHomeComputer(homeComp: Server): void {
 
   //Reset RAM usage calculation for all scripts
   homeComp.scripts.forEach(function (script) {
-    script.ramUsage = null;
+    // Don't need to invalidate the whole module, just the ram usage.
+    delete script.ramUsage;
```
d0sboots: Here too

d0sboots (commented):
OK, I looked at NetscriptJSEvaluator. I need to think about it some more, but I have two high level comments:

1. There are two separate "phases" here, determining the RAM cost and actually compiling. The way the code is set up, you can't compile without first determining the RAM cost, but the other way (only RAM cost with no compiling) is very possible. I feel like the way you are doing invalidation is linking these more tightly together; in particular, `saveScript` is only called by the script editor, and it used to do its extended `dependents` walk strictly to invalidate for RAM usage. (Invalidating the script alone was sufficient for compiling.)
2. I liked the globalModuleSequenceNumber better. It was a lazier system; writing only had to change one thing, instead of doing an entire walk. It also meant there was much less metadata that had to be carefully kept in the correct state to ensure correctness. The trade-off was that you had to check the seqNum of all dependents when running, but the number of dependencies is not typically high; it's the correctness I'm concerned about more than the performance here.

d0sboots:
I built the branch to verify: The following script (modified from my original code) demonstrates a race condition in this PR. A similar race condition affects current prod, but it takes a little more effort to trigger it.
```js
// A different race condition in execute.
// https://github.com/bitburner-official/bitburner-src/issues/444

/** @param {NS} ns */
function createScript(num, ns) {
  ns.write(
    `import_test_generated${num}.js`,
    (num == 2 ? 'import "import_test_generated1.js";\n\n' : '') +
    `
/** @param {NS} ns */
export function main(ns) {}
    `.trim(),
    "w"
  );
}

/** @param {NS} ns */
  createScript(1, ns);
  createScript(2, ns);
  ns.run("import_test_generated2.js", 1, 1);
  createScript(1, ns);
}
```

d0sboots:
I should also mention: I know how to solve that, as well as share scripts across *all* servers (not just sharing imports across the same server), but it takes things in a different direction than you've gone, so I'm not sure how I'd integrate it with what you've done.

src/Server/ServerHelpers.ts
Snarling: > There are two separate "phases" here, determining the RAM cost and actually compiling.

These are separate and are not intertwined. You should be able to compile a script without requiring a ram calculation (as long as you are providing a ramcost override), and you should be able to calculate a ramcost without requiring a compilation (e.g. using the mem command). Maybe right now it's still doing a ram calc even if an override is provided, but this can be fixed.

However, anything that invalidates ram usage, other than prestiging, should invalidate both ram usage and the module itself. Prestiging is the exception because singularity script costs can change based on node - the only effect that can invalidate a ramcost without invalidating the script's URL/module.

> I liked the globalModuleSequenceNumber better. It was a lazier system; writing only had to change one thing, instead of doing an entire walk.

Hard disagree here. Running a script with an already-initialized module is a common codepath and should be as lean as possible. Module invalidation, on the other hand, is only performed on infrequent code paths - overwriting scripts or their imports. If you don't "do the walk" at invalidation time, then you basically still have to do it at script run time to check if any imports have had their modules updated.

> it's the correctness I'm concerned about more than the performance here.

Invalidating during the event that causes an invalidation is more correct than invalidating when running.

Snarling:
>The following script (modified from my original code) demonstrates a race condition in this PR. 

Yeah, it looks like this is due to the write revoking the script url before it has actually had a chance to be imported by the first script's run. Should be able to fix this pretty easily.

>  share scripts across all servers

This is not desirable if you're talking about not requiring a different syntax to import a script across servers, where you could copy over a script that has imports and it magically still imports from the correct script even though that script doesn't exist on the destination server. Adding a new protocol syntax for importing a script from a specific server is something that could be added to this current implementation.

d0sboots:
> > The following script (modified from my original code) demonstrates a race condition in this PR.
> 
> Yeah, it looks like this is due to the write causing deletion of the script url before it has actually had a chance to be imported by the first script's run. Should be able to fix this pretty easily.

It's not simple to fix.

The fundamental issue is that it takes time to fetch, load and run a URL to process its imports, even when that URL is a blob. Furthermore, this process fundamentally runs in multiple steps if there are multiple imports: First A is fetched and loaded, then B is fetched and loaded, then C is fetched and loaded, then (C is run, then B is run, then A is run) as a single step.

Happening in multiple steps is what kills most simple approaches, because it means any kind of invalidation can happen *between* those fetch/load steps and cause the fetching to fail. The most bulletproof approach is to instead ensure that you never invalidate a Blob url unless/until nothing is using it - which means keeping it around until all RunningScripts that reference it are finished. There may be other things you can try, but they're more complicated, especially to ensure that you don't create more modules than you intend and that you have no race conditions.

There are two ways you can do that. One is reference counting, which is error-prone in a language like JS: You have to be very careful not only to (manually) cover all the places where references are taken/released, but also handle exceptions very carefully. The 2nd way is to use `WeakRef` and `FinalizationRegistry` to hook into the GC itself and invalidate urls when they are actually garbage collected, which is my preferred solution.

> > share scripts across all servers
> 
> This is not desirable if you're talking about not requiring a different syntax to import a script across servers, where you could copy over a script that has imports and it magically still imports from the correct script even though that script doesn't exist on the destination server. Adding a new protocol syntax for importing a script from a specific server is something that could be added to this current implementation.

I have a design that doesn't require any of that. It's essentially equivalent to a cache. You can use the same url for a script if it has the (exact) same code, post-transform. That implies that it also has all the same imports, since all the urls for the imports must match, and so on recursively.

Edit: To be clear, I didn't mean that I intended users to share scripts across servers. Rather, that as an implementation detail we can share the compiled form of identical scripts. (This is quite common for worker scripts that are copied to every server.)

Snarling:
> It's not simple to fix.

It is fixed though, and I would call the fix fairly simple.

> The 2nd way is to use WeakRef and FinalizationRegistry to hook into the GC itself and invalidate urls when they are actually garbage collected

URLs are never garbage collected unless you manually revoke them, so this would not have worked. You would just never have revoked any URLs. (Maybe I'm just misunderstanding what you were suggesting here though)

> You can use the same url for a script if it has the (exact) same code, post-transform

Currently they would never match unless we get rid of the extra code that is appended to the script code that places the script in a convenient location on the sources tab:

```js
newCode += `\n//# sourceURL=${script.server}/${script.filename}`;
```

Currently I added a rudimentary version that only works for scripts with no imports that have already been compiled before copying, but I like your idea here. Other than the developer tools I'm not sure if the sourceURL comment is doing anything for us. If not, we can get rid of that and then use a Map<CodeString, X> to store information. We need to store not just a url, but all of the Scripts that are using that URL, so the type of X might be `{url: ScriptURL, scripts: Set<Script>}`. When a new script decides to use that URL, it would add itself to the script set. When a script is invalidated, it would remove itself from that script set, and then if the set is empty it would queue the URL for deletion and also remove the entry from the map.

src/Server/ServerHelpers.ts
SamB: > Invalidating during the event that causes an invalidation is more correct than invalidating when running.

Would that make `mem` work properly when you update a module that a script imports, but not the script itself, and want to verify the effect on RAM usage? That'd sure be nice.

Snarling: > Would that make mem work properly when you update a module that a script imports, but not the script itself, and want to verify the effect on RAM usage? That'd sure be nice.

Does that not already work as expected?

If this didn't work as expected already, it should with this implementation. Modifying a script Y that is imported by script X will immediately remove existing ram calculations from both X and Y. Then running `mem` for script X would be forced to perform a fresh recalculation using the updated content of its imported script Y.

d0sboots:
> It is fixed though

Maybe. Maaaaybe. There's at least one potential issue that only works by either coincidence or an obscure corner of the standard. (By using scp in the right order, I can cause a URL to be revoked before it is fetched, but it seems like since that URL was already previously used to fetch a module, that takes precedence over the revoked object URL.) I feel like if I keep trying, I can probably find an issue with it.

It may be good enough for now though; at the least, it's probably more robust than what we have currently.

Oh, except for the fact that `Server.removeScript` doesn't invalidate its script. That's a hole that you can do fun things with, but it's an easy fix.

Snarling:
> I can cause a URL to be revoked before it is fetched

Edit: Oh actually I see the problem, it's in the new code I added to reuse modules/urls. Invalidating one of the scripts doesn't affect the stored url string on the other script, but the underlying url object gets revoked and the copied script still thinks everything is valid. There would have been an issue if you invalidate either script, and then try to use the other module as an import in a different file (it would try re-using the URL which is no longer valid but was never cleared).

Without that module/url reuse code there is no room for an issue. One solution is just to mark these two scripts as dependents of each other so they will invalidate each other, but that's not really the correct solution since it leads to creating a new module for one of the scripts that didn't change. Better solution if we want to keep url/module sharing is either a map of URLs to sets of scripts using them that will only revoke the when no more scripts are using it, or the weakmap which I think I get what you meant by now. For now I will just remove the url sharing feature since it can be implemented later and shouldn't be half-implemented like what I have here.

Original comment:

I don't think this is possible. A script url is only in the process of being used during a compile call, during the window where the number of ongoing compilations is greater than 0. During this time window, URLs are not revoked.

If an invalidating action occurs while there are no compilations ongoing, the url is revoked immediately and we know this is a safe time to revoke. If it occurs during compilations, the url will be revoked as soon as there are no more compilations ongoing, which is also a safe time to revoke. During the one time period where revoking could lead to any issues (compilations) revoking does not occur.

(Accidentally closed PR while typing comment on phone)

SamB:
> > The 2nd way is to use WeakRef and FinalizationRegistry to hook into the GC itself and invalidate urls when they are actually garbage collected
> 
> URLs are never garbage collected unless you manually revoke them, so this would not have worked. You would just never have revoked any URLs. (Maybe I'm just misunderstanding what you were suggesting here though)

It is true that the `blob:` urls have to be manually revoked before the browser can allow them, or their associated `Blob` data, to be removed from memory.

But if you had an object (say, the old ScriptUrl) by way of which the URL would be accessed, you could create a [FinalizationRegistry](https://tc39.es/ecma262/multipage/managing-memory.html#sec-finalization-registry-objects) using, say,
```typescript
const scriptUrlFinalizationRegistry = new FinalizationRegistry<string>(URL.revokeObjectURL);
```
at the TOP LEVEL.

And then you could do something like this to register the ScriptUrl later:

```typescript
    scriptUrlFinalizationRegistry.register(scriptUrl, blob);
```

... then, when you no longer had any strong references to the passed-in `scriptUrl`, the cleanup callback would get called with that `blob` URL value, so as long as you made sure to retain the `ScriptUrl`s during any operation that needs to fetch from those URLs, you'd be fine.

It has to be an object (or Symbol) because, as <https://tc39.es/ecma262/multipage/executable-code-and-execution-contexts.html#sec-canbeheldweakly> says,

> A language value without [language identity](https://tc39.es/ecma262/multipage/notational-conventions.html#sec-identity) can be manifested without prior reference and is unsuitable for use as a weak reference.

(In this case, these particular strings might or might not be retained in JavaScript form by the implementation of createObjectURL/revokeObjectURL; in many other cases, you might just happen to have an identical string lying around already, and you aren't permitted to know whether there's any difference between two strings with the same contents.)

Even in a hypothetical universe where the strings had language identity, you couldn't pass them in as the `target` (first parameter) for [FinalizationRegistry.prototype.register](https://tc39.es/ecma262/multipage/managing-memory.html#sec-finalization-registry.prototype.register) because you kind of need to pass them in as (part of) the `heldValue` so they can be passed to revokeObjectURL for cleanup, which would cause them to be retained forever. (FinalizationRegistry.prototype.register helpfully throws a `TypeError` if you try to pass the same exact value as both `target` and `heldValue` for this exact reason.)

Unfortunately, even with the restriction to objects, there's not really any coherent way for the spec to require that finalization actually happen under any circumstances, but I'm fairly certain that it can happen in v8.

d0sboots:
> Oh actually I see the problem, it's in the new code I added to reuse modules/urls. Invalidating one of the scripts doesn't affect the stored url string on the other script, but the underlying url object gets revoked and the copied script still thinks everything is valid. There would have been an issue if you invalidate either script, and then try to use the other module as an import in a different file (it would try re-using the URL which is no longer valid but was never cleared).
>
> Without that module/url reuse code there is no room for an issue.

Yes, that's what I meant with "by using scp in the right order." You gotta start trusting that when I say something is possible, it is XD

With the scp code removed, I think the compile side of things is looking pretty solid. (I wasn't going to suggest doing that, because it *seemed* like it would be OK with it in, but I feel more comfortable with it out for now too.)

d0sboots:
> But if you had an object (say, the old ScriptUrl) by way of which the URL would be accessed, you could create a [FinalizationRegistry](https://tc39.es/ecma262/multipage/managing-memory.html#sec-finalization-registry-objects) using, say...

SamB has basically outlined my plan. There's a few more details that I'd worked out, but I'd planned on working them out as a subsequent PR (done on top of this one), to work out the kinks on my own. IMO it's probably best to start with stabilizing this version as-is.

Snarling:
Gonna go ahead and merge this. URL/Module sharing can be implemented in a separate PR (possibly something for 2.4)

src/Server/BaseServer.ts
d0sboots, line 1: Still need to add an invalidate here, in `removeFile`.

Snarling: Thanks, missed that one.

Snarling (commented):
Invalidate script on removeFile (attempted commit from github.dev)

PR #468, opened 2023-04-07
NETSCRIPT: Compiled modules will be even more shared

d0sboots:
Turns out it was pretty straightforward to throw this together!

----

This creates greater memory savings by sharing compiled modules across servers, when it is possible to do so. This introduces a small user-visible change: Module-scoped variables will now be global to *all* instances of that script, instead of per-server global.

This also revokes Blob URLs *immediately* after they are created. Blob leakage should be a thing of the past!

# Testing

I tested basic script stuff, including checking that modules are being shared. I haven't done any in-depth memory tests yet.

### DIFF
--- a/src/Netscript/NetscriptHelpers.ts
+++ b/src/Netscript/NetscriptHelpers.ts
@@ -251,9 +251,7 @@ function makeRuntimeErrorMsg(ctx: NetscriptContext, msg: string, type = "RUNTIME
   const userstack = [];
   for (const stackline of stack) {
     const filename = (() => {
-      // Filename is current file if url found
-      if (ws.scriptRef.url && stackline.includes(ws.scriptRef.url)) return ws.scriptRef.filename;
-      // Also check urls for dependencies
+      // Check urls for dependencies
       for (const [url, script] of ws.scriptRef.dependencies) if (stackline.includes(url)) return script.filename;
       // Check for filenames directly if no URL found
       if (stackline.includes(ws.scriptRef.filename)) return ws.scriptRef.filename;
--- a/src/NetscriptFunctions.ts
+++ b/src/NetscriptFunctions.ts
@@ -900,8 +900,6 @@ export const ns: InternalAPI<NSFull> = {
 
         // Create new script if it does not already exist
         const newScript = new Script(file, sourceScript.code, destServer.hostname);
-        // If the script being copied has no dependencies, reuse the module / URL
-        // The new script will not show up in the correct location in the sources tab because it is just reusing the module from a different server
         destServer.scripts.push(newScript);
         helpers.log(ctx, () => `File '${file}' copied over to '${destServer?.hostname}'.`);
       }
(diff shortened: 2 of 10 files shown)


Snarling:
I missed it in mine, but we may want to invalidate all scripts on a Server other than home during prestigeAllServers. It might not be needed in this case though, I think those old servers from the previous run are able to be garbage collected, which would then trigger your cleanup.

> I haven't done any in-depth memory tests yet.

I don't really have a test meant to simulate "normal person playing the game" but I think in general, especially for people who copy 3 workers to every server every reset, there should be substantial improvements in memory usage. I did the same "how many unique modules before the page breaks" test and got up to 913k which is slightly lower but within a margin of error of what I was getting on other branches, but obviously a unique module test doesn't show off any of this branch's benefits.

One extremely minor negative as far as module count is that modules are now created for every import, even if those imports are not ran, whereas previously only blob urls were created which could then be revoked. But this is unavoidable based on the implementation, and on module count the benefits far outweigh the cost.

> This introduces a small user-visible change: Module-scoped variables will now be global to all instances of that script, instead of per-server global.

So there are a lot of issues with this general idea. First it kind of negates the usefulness of netscript ports. Second, the behavior can lead to unexpected player-script bugs if a player uses module-scoped variables. Third, there's really not a way to justify the way it works in-world.

Normally that would make me very against this sort of change. But the fact that modules eat up memory forever and never give it back, and normal gameplay leads to a bunch of extra modules every reset (copying scripts to servers), plus the benefit of it adding cool new functionality, wins out in this instance and I'm for it. I'm down to merge this if you feel like it's already where you want it.

d0sboots:
> I missed it in mine, but we may want to invalidate all scripts on a Server other than home during prestigeAllServers. It might not be needed in this case though, I think those old servers from the previous run are able to be garbage collected, which would then trigger your cleanup.

I think letting the GC handle it is correct; if there are any references still, then we should keep it in the cache.

> One extremely minor negative as far as module count is that modules are now created for every import, even if those imports are not ran, whereas previously only blob urls were created which could then be revoked.

The modules were actually always being created: they needed to, since they are referenced (transitively) as imports from the top-level script we are running. The only difference is we are creating the modules up-front, and keeping a reference to them.

> So there are a lot of issues with this general idea. First it kind of negates the usefulness of netscript ports. Second, the behavior can lead to unexpected player-script bugs if a player uses module-scoped variables. Third, there's really not a way to justify the way it works in-world.

Definitely agree with the last two points. I don't think it changes much about ports, though - the primary use of them is to communicate between *different* scripts, instead of different instances of the *same* script on different servers.

I'd be more worried about breaking user scripts, except I'm hoping that people generally aren't using them much. (And that when they are, it's in scripts that are already singletons.) Module-scoped variables are really awkward to use in BB: They get reset both on reload and on script change, so they're not good for storing any kind of long-term state. At the same time, it's not easy to reset them on demand, so they're not good for short-term state either. I've found that, even for a single script, global variables end up being much more ergonomic.

> I'm down to merge this if you feel like it's already where you want it.

I fixed a small thing, but I think it's ready.

PR #469, opened 2023-04-08
HASHNET: Company Favor

jeek:
HASHNET: Company Favor

Adds a new option to HashNet to spend hashes 5 Company Favor at a time.

### DIFF
--- a/src/Constants.ts
+++ b/src/Constants.ts
@@ -88,7 +88,7 @@ export const CONSTANTS: {
   LatestUpdate: string;
 } = {
   VersionString: "2.3.0",
-  VersionNumber: 30,
+  VersionNumber: 31,
 
   /** Max level for any skill, assuming no multipliers. Determined by max numerical value in javascript for experience
    * and the skill level formula in Player.js. Note that all this means it that when experience hits MAX_INT, then
--- a/src/Constants.ts
+++ b/src/Constants.ts
@@ -88,7 +88,7 @@ export const CONSTANTS: {
   LatestUpdate: string;
 } = {
   VersionString: "2.3.0",
-  VersionNumber: 31,
+  VersionNumber: 30,
 
   /** Max level for any skill, assuming no multipliers. Determined by max numerical value in javascript for experience
    * and the skill level formula in Player.js. Note that all this means it that when experience hits MAX_INT, then
--- a/src/SaveObject.tsx
+++ b/src/SaveObject.tsx
@@ -657,6 +657,9 @@ function evaluateVersionCompatibility(ver: string | number): void {
     // Prior to v2.2.0, sleeve shock was 0 to 100 internally but displayed as 100 to 0. This unifies them as 100 to 0.
     for (const sleeve of Player.sleeves) sleeve.shock = 100 - sleeve.shock;
   }
+  if (ver < 31) {
+    anyPlayer.hashManager.upgrades["Company Favor"] = anyPlayer.hashManager.upgrades["Company Favor"] ?? 0;
+  }
 }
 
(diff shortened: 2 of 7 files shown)


src/Constants.ts
```diff
@@ -88,7 +88,7 @@ export const CONSTANTS: {
   LatestUpdate: string;
 } = {
   VersionString: "2.3.0",
-  VersionNumber: 30,
+  VersionNumber: 31,
```
Snarling: Increasing the version number will just be done at the next actual version bump, also should not be required based on the implementation where you are only resetting the value to 0 if it is already nullish. Keeping this at 30 should continue to work even on games loaded before we actually bump the version to 31.

src/ui/React/CompanyDropdown.tsx
```diff
@@ -0,0 +1,42 @@
...
+interface IProps {
+  purchase: () => void;
+  canPurchase: boolean;
+  onChange: (event: SelectChangeEvent<string>) => void;
+  value: string;
+}
+
+export function CompanyDropdown(props: IProps): React.ReactElement {
+  const companies = [];
+  for (const company of companiesMetadata.sort((a, b) => a.name.localeCompare(b.name))) {
```
Snarling: Basically this means that every time the CompanyDropdown react element is rerendered, it will try to sort the entire companiesMetadata array again.
Could instead sort it in the file where it's initialized, and then there is no need to sort it again later.

src/SaveObject.tsx
```diff
@@ -657,6 +657,9 @@ function evaluateVersionCompatibility(ver: string | number): void {
     // Prior to v2.2.0, sleeve shock was 0 to 100 internally but displayed as 100 to 0. This unifies them as 100 to 0.
     for (const sleeve of Player.sleeves) sleeve.shock = 100 - sleeve.shock;
   }
+  if (ver < 31) {
+    anyPlayer.hashManager.upgrades["Company Favor"] = anyPlayer.hashManager.upgrades["Company Favor"] ?? 0;
```
Snarling: This will error out when loading a player that doesn't have a hash manager ("Can't access property `upgrades` on `undefined` (or null, whatever the does-not-have value is)"). I imagine the game only initializes a hashmanager for players who have SF9/in BN9 but I haven't dug into that and I'm not at my regular PC to check easily.

I would do something like:

```ts
const hashUpgrades = anyPlayer.hashManager?.upgrades;
if (hashUpgrades) hashUpgrades["CompanyFavor"] ??= 0;
```

Snarling (changes requested):
Just a couple issues on initial review. There are a couple other things I need to take a closer look at because I'm not familiar enough with how they work (e.g. does adding directly to favor have the potential to get undone by a recalculation of favor based on company rep)

src/SaveObject.tsx
jeek: You somehow reviewed an old version, I'd caught the error potential and had already submitted a patch for it. Thanks for the reminder that ??= exists, changed that line to use it.

src/ui/React/CompanyDropdown.tsx
jeek: Fixed.

src/Constants.ts
jeek: Reverted.

src/ui/React/CompanyDropdown.tsx
jeek: Note that ServerDropdown does the same thing, fix should also be applied there at some point.

Snarling (approved):

PR #470, opened 2023-04-08
NETSCRIPT: Add `sourceFiles` to getPlayer()

d0sboots:
There's currently no way to see what sourcefiles the player has. This is the logical place.


Snarling (changes requested):
A few things,

So first, whatever function provides this information would also need the associated type info added to NetscriptDefinitions. There's no type error that you didn't do this, because the returned type is still satisfying the required interface, but the extra info isn't part of the type and there will be no autocomplete + people who write scripts in ts will get errors trying to access it.

Whether this belongs on getPlayer at all is hard to determine. First, the property includes spoiler information, so that indicates against belonging on getPlayer (even though there is existing spoiler info on getPlayer in the form of bitNodeN and  playtimeSinceLastBitnode).

However it's possible that player sourcefiles could be used by formulas functions (e.g. companyGains could take into account the player's BN11 SF to account for favor contributing to salary), so since that functionality does seem to be necessary for accurate formulas, that indicates it needs to be included on getPlayer and maybe also on getSleeve if SF11 affects sleeve company gains as well, perhaps with a name that's not quite as overtly spoilery like sfLevels.

However, the game doesn't just look at Player.sourceFiles directly, it uses a helper function (Player.getSourceFileLevel or something like that) because the data structure of Player.sourceFiles is terrible for performing lookups (to lookup an SF level you need to iterate through an array and compare "n" values to find the right entry and then use the associated "lvl" value). So right now, formulas would not be able to take into account SF, because the interfaces passed in to formulas functions don't have the helper function, which is a class member of the actual Player object.

So this is an opportunity to also fix the format for sourceFiles to something that makes more sense to access data from (either an object or map keyed by the SF number and valued by the level) so we can get rid of that Player helper function and just get source file levels directly, and also rework some formulas functions to actually utilize the provided sourceFiles (might just be companyGains).

I don't mind implementing this if you're not super invested in it.

Snarling:
Also it looks like the information is already provided to the player via ns.singulairty.getOwnedSourceFiles. But if reworking formulas to use this info then we should include it on the person interface.

I think either way, it is worth converting it to a data structure that isn't terrible at its job, though this would be an API break for getOwnedSourceFiles (or if it's added to getPlayer, we may just get rid of getOwnedSourceFiles altogether)

d0sboots:
OK, I didn't want to propose a more major change but if you're good with it, I'm good with it. I'll keep getOwnedSouceFiles around to avoid breakage, but document it as legacy.

d0sboots:
OK, that was quite a bit of refactoring, but it's done now. I tested the save and upgrade paths as much as I could - in particular, the upgrade can't be in the usual place because the Player object gets used to show the "upgrade confirm" stats, and that doesn't go through the upgrade logic. (It didn't go through `loadPlayer` either, but I fixed that.)

steven-r:
One cent from my side:

`getOwnedSouceFiles` is marked as 5G memory consumption and is only available w/ SF4.

If you move this to `getPlayer()`, this will be reduced to .5G (AFAIK) and is available starting with SF1

Is this intentional?

d0sboots:
> If you move this to getPlayer(), this will be reduced to .5G (AFAIK) and is available starting with SF1
>
> Is this intentional?

Yes (on my part, at least). I don't care about the RAM cost, but if you want to make a truly general-purpose script (one that reacts properly regardless of what BNs the player has been in), you need to be able to *see* the sourcefiles without restriction. Otherwise, your script won't be able to run without BN4, and that's no good for something that's designed to operate even in BN1.1.

> First, the property includes spoiler information, so that indicates against belonging on getPlayer (even though there is existing spoiler info on getPlayer in the form of bitNodeN and playtimeSinceLastBitnode).

Side note: I don't think this is actually a spoiler. There are references to Bitnodes (particularly BitnodeMultipliers) throughout the API. Similarly, there are *lots* of references to sourcefiles. But these things come without the context that would make them spoilers.

When I was in BN1.1 and exploring the API, it was immediately apparent that "sourcefiles" gate access to most of the additional APIs, and also that something called "bitnodes" exist and have multipliers. It was *not* apparent that the two were linked, and if you had asked me for my best guess on what a bitnode was, I'd have said, "Maybe something like a hacknet node?" Something you could acquire for multipliers.

I don't think we need to worry about hiding the *existence* of these things (because in the API at least, it's impossible), but we're doing fine on preserving the actual spoiler info.

Snarling:
> Is this intentional?

No. This belongs gated behind SF4, and it belongs having a separate ramcost in my opinion. My only support for moving it to be included in getPlayer was if it's necessary for proper formulas support. Although actually, formulas doesn't need to receive 100% of its information from getPlayer/getSleeve. In fact in this current implementation, there's not even room to expand formulas to use the property, since it's not on the person interface it's only on player, and the related formulas need to work for information from getSleeve as well. Being on the person interface would also mean it needs to be on actual sleeves as well, which doesn't make a whole lot of sense really.

I think removing it from getPlayer and then when/if formulas functions are actually reworked to be able to use custom bitNode information, we would just accept in an additional optional parameter for a sourceFiles map (defaulting to the Player's actual SF). Without needing it for true formulas integration, I don't think this belongs on getPlayer (which is already an overloaded function for 0.5GB)

> Otherwise, your script won't be able to run without BN4, and that's no good for something that's designed to operate even in BN1.1.

The only thing a hypothetical BN1.1 script could use this for is for determining API access, which I get wanting to check before you actually have access (without needing to set up a separate script that errors out to get the info back to a main script). But I kind of like that checking this before you have the correct APIs requires outside the box thinking and isn't just handed to the player through the base API.

d0sboots:
> No. This belongs gated behind SF4, and it belongs having a separate ramcost in my opinion. My only support for moving it to be included in getPlayer was if it's necessary for proper formulas support. Although actually, formulas doesn't need to receive 100% of its information from getPlayer/getSleeve. In fact in this current implementation, there's not even room to expand formulas to use the property, since it's not on the person interface it's only on player, and the related formulas need to work for information from getSleeve as well. Being on the person interface would also mean it needs to be on actual sleeves as well, which doesn't make a whole lot of sense really.

Damnit Snarling, I put in all that work based on the understanding that we were on the same page here. XD

Oh well

PR #472, opened 2023-04-10
SINGULARITY: Added aug and node identifiers

G4mingJon4s:
This is a continuation of #453.

# Changes

- Added saving the timestamp for the aug and node install/finish.
- Added singularity functions that return these values.
- Removed the playtimeSinceX properties from the player object.

# Actions
- [X] Added the properties internally
- [X] Added the singularity functions for the properities.
- [X] Added the documentation for the functions.
- [X] Removed the player object properties.
- [x] Added the version change logic for adding the properties on older savefiles.


G4mingJon4s:
@Snarling 
So, I still have no idea how / where to implement the savefile logic.
Is it `(ver < 27)` in SaveObject.tsx?

Snarling:
> So, I still have no idea how / where to implement the savefile logic.

Oh so that's my bad. There was a recent PR that added more version logic, and originally it had the <31 check but I didn't notice it got removed.

You can replace this whole section: https://github.com/bitburner-official/bitburner-src/blob/751fe82f9255cb114207a54ad502101d13f2642e/src/SaveObject.tsx#L660-L662

with something like this to initialize the values.
```ts
if (ver < 31) {
  if (anyPlayer.hashManager !== undefined) {
    anyPlayer.hashManager.upgrades["Company Favor"] ??= 0;
  }
  anyPlayer.lastAugReset ??= anyPlayer.lastUpdate - anyPlayer.playtimeSinceLastAug;
  anyPlayer.lastNodeReset ??= anyPlayer.lastUpdate - anyPlayer.playtimeSinceLastBitnode;
}
```

Basically the version number is currently 30 (from Constants.ts). So on saves created now and anytime before now, it'll run this conversion logic. Once we bump version and release 2.3, the version will be 31, and any saves from that point forward will no longer need to trigger this savegame conversion code.

G4mingJon4s:
So, I implemented the version changes. I also did some testing and everything seems to work fine. Don't take my word for it though...

PR #473, opened 2023-04-11
CODEBASE: Add Jsonable Map and Set types, move player.sourceFiles to a map

Snarling:
This is followup from #470 which also included some some netscript API changes that we didn't go with in the end. Since that PR is now closed but had other great changes, I cherry picked the relevant commits over from @d0sboots other branch and reverted the netscript API changes.

Still todo on this branch: Mark the current getSourceFiles as deprecated, create a new function to provide a map to the player instead (getSourceFileLevels?). Potentially add another method for notifying the player that they are using a deprecated function (e.g. onetime terminal message per function per time the game is loaded).

Todo for separate PR is converting several other game data structures to map and set.
Good candidates for JSONMaps are server scripts, runningScripts, text files, player augmentations, jobs, etc.
Good candidates for JSONSets would be server messages, player achievements, exploits, faction invitations, joined factions, etc.


d0sboots:
My 2 cents: It's not worth migrating from the API of the existing `getSourceFiles` if you plan on keeping it in `ns.singularity`. The improvement in ergonomicness just isn't large enough to justify the churn involved in switching; as this PR demonstrates, it's very easy to store the result in a Map yourself if you want to.

Snarling:
Converted scripts to a map, and also added some more non-enum string types (IPAddress, ServerName, and ScriptFilename). Currently these types are just equal to string so there is no hard typechecking, that will be added before merging.

src/Types/strings.ts
```diff
@@ -0,0 +1,29 @@
+// Script Filename
+export type ScriptFilename = string /*& { __type: "ScriptFilename" }*/;
+/*export function isScriptFilename(value: string): value is ScriptFilename {
+  // implementation
+}*/
+/*export function sanitizeScriptFilename(filename: string): ScriptFilename {
+  // implementation
+}*/
+export function scriptFilenameFromImport(importPath: string, ns1?: boolean): ScriptFilename {
+  if (importPath.startsWith("./")) importPath = importPath.substring(2);
```
d0sboots, line 10: Why is our directory handling so bad. "./foo.js" in a script means "/foo.js" (I.e. it's absolute, despite looking relative) and "/foo.js" isn't even allowed.

Snarling: Right now this is just a reimplementation of some bad code that was being done separately in both NetscriptJSEvaluator and RamCalculation, so no behavior change yet.

But yeah that function needs a rework and it should take the originating file's ScriptFilename in as an input so it can do proper relative paths. Of course moving to correct relative paths is a behavior change because right now importing "myScript" or "./myScript" means you're importing "myScript.js" from the root directory, not from the same directory as the existing script. But if we only handle relative paths when the player provides a path starting with a `.` that shouldn't be too big of a breaking change.

PR #476, opened 2023-04-11
NETSCRIPT: Type correctness for getServer

### LINKED ISSUE #475: Hacknet servers (BN9) are missing property backdoorInstalled, leading `getServer()` API break
Be in BN9, have a hacknet server, run this:
```js
/** @param {NS} ns */

	ns.tprintf("%4j", ns.ui.getGameInfo());
	for(let target of ["n00dles", "home", "hacknet-server-0"]) {
		let s = ns.getServer(target);
		ns.tprintf("%s: %t", target, Object.keys(s).includes("backdoorInstalled"));
		let p = ns.getPlayer();
		try{ns.tprintf("%s: %s", target, typeof ns.formulas.hacking.growPercent(s, 1, p));}
		catch(e){ns.tprintf("%s",e)}
	}
	
}
```
results in
```
[home ~/]> run test.js
Running script with 1 thread(s), pid 431906 and args: [].
{
    "version": "2.2.2",
    "commit": "6eb5b5ab",
    "platform": "Steam"
}
n00dles: true
n00dles: number
home: true
home: number
hacknet-server-0: false
TYPE ERROR
test.js@home (PID - 431906)

formulas.hacking.growPercent: server should be a Server.

Stack:
test.js:L18@Module.main
```

I would have expected `ns.getServer("hacknet-server-0")` to return an object respecting the Server contract and the output for `hacknet-server-0` to look the same as for the other two.

Snarling:
Fixes #475 

Refactored netscript Server interface. Used new interface as base interface for actual BaseServer class, ensuring all actual servers comply with the interface.

Refactored getServer to provide specific info from the API type instead of also including dummy info that is not part of the API type.

Refactored netscript Server helper so that it will return a correct Server interface instead of pretending it is returning a Server class member,

Refactored hacking formulas functions to expect a Server interface instead of requiring a Server class object. For inputs that are invalid (missing keys needed for the given formula) worst-case return values are provided.

### DIFF
--- a/src/NetscriptFunctions.ts
+++ b/src/NetscriptFunctions.ts
@@ -985,31 +985,56 @@ export const ns: InternalAPI<NSFull> = {
 
       return Object.assign({}, getBitNodeMultipliers(n, lvl));
     },
-  getServer:
-    (ctx) =>
-    (_hostname = ctx.workerScript.hostname) => {
-      const hostname = helpers.string(ctx, "hostname", _hostname);
-      const server = helpers.getServer(ctx, hostname);
-      const copy = Object.assign({}, server) as Server;
-      // These fields should be hidden.
-      copy.contracts = [];
-      copy.messages = [];
-      copy.runningScripts = [];
-      copy.scripts = [];
-      copy.textFiles = [];
-      copy.programs = [];
-      copy.serversOnNetwork = [];
-      if (!copy.baseDifficulty) copy.baseDifficulty = 0;
-      if (!copy.hackDifficulty) copy.hackDifficulty = 0;
-      if (!copy.minDifficulty) copy.minDifficulty = 0;
-      if (!copy.moneyAvailable) copy.moneyAvailable = 0;
-      if (!copy.moneyMax) copy.moneyMax = 0;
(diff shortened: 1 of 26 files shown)


src/NetscriptFunctions.ts
```diff
@@ -985,31 +985,56 @@ export const ns: InternalAPI<NSFull> = {
...
+      minDifficulty: server.minDifficulty,
+      moneyAvailable: server.moneyAvailable,
+      moneyMax: server.moneyMax,
+      numOpenPortsRequired: server.numOpenPortsRequired,
+      openPortCount: server.openPortCount,
+      requiredHackingSkill: server.requiredHackingSkill,
+      serverGrowth: server.serverGrowth,
+    };
+
+    const copy = Object.assign({}, server) as Server;
```
d0sboots, line 1018: Isn't all this code unreachable?

Snarling: Yeah I meant to delete all that... not sure why it's not a lint error.

d0sboots: Probably because of that one bug, most of our lints don't work XD

PR #478, opened 2023-04-17
UI: Added hostname to LogBox title

unkwntech:
I added the hostname to the title bar of the log/tail window to indicate which server the script that is being viewed is running on.

Screenshot of log window showing the hostname "home" displayed as a prefix in the title bar.

### DIFF
--- a/src/ui/React/LogBoxManager.tsx
+++ b/src/ui/React/LogBoxManager.tsx
@@ -249,7 +249,12 @@ function LogWindow(props: IProps): React.ReactElement {
 
   function title(full = false): string {
     const maxLength = 30;
-    const t = `${script.filename} ${script.args.map((x: ScriptArg): string => `${x}`).join(" ")}`;
+    const server = GetServer(script.server);
+    let hostname = "";
+    if (server !== null) {
+      hostname = `${server.hostname}: `;
+    }
+    const t = `${hostname}${script.filename} ${script.args.map((x: ScriptArg): string => `${x}`).join(" ")}`;
     if (full || t.length <= maxLength) {
       return t;
     }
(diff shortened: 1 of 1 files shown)


src/ui/React/LogBoxManager.tsx
```diff
@@ -249,7 +249,12 @@ function LogWindow(props: IProps): React.ReactElement {
 
   function title(full = false): string {
     const maxLength = 30;
-    const t = `${script.filename} ${script.args.map((x: ScriptArg): string => `${x}`).join(" ")}`;
+    const server = GetServer(script.server);
```
Snarling: script.server here is already a hostname, is there a reason you're doing another lookup to fetch the actual server using that name?

src/ui/React/LogBoxManager.tsx
```diff
@@ -249,7 +249,12 @@ function LogWindow(props: IProps): React.ReactElement {
...
+    let hostname = "";
+    if (server !== null) {
+      hostname = `${server.hostname}: `;
+    }
+    const t = `${hostname}${script.filename} ${script.args.map((x: ScriptArg): string => `${x}`).join(" ")}`;
```
Snarling:
```suggestion
    const t = `${script.server}: ${script.filename} ${script.args.join(" ")}`;
```
It's pre-existing code, but that map function to turn arguments into strings is also not needed - [joining them](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/join#description) already does that.

Snarling (changes requested):

Snarling (approved):

brunnerh:
I feel like this should be an option, my titles already don't fit fairly often.
image

Snarling:
Yeah I can see how that would be a problem. I'll open an issue for it.

d0sboots:
I am undoing this in #508, but it adds the ability to both get the current title and set a custom title, so hopefully that makes it easy to reimplement for the scripts where you want it.

PR #479, opened 2023-04-18
FILES: Path rework & typesafety

Snarling:
* Added new types for various file paths, all in the Paths folder. Previously these were all just string so there was zero validation from ts, we just had to rely on not assigning the wrong thing by not making mistakes in code.
* TypeGuards/other helper functions related to these types
* Added basic globbing support with * and ?. Currently only implemented for Script/Text, on nano and download terminal commands which already supported some more limited pattern matching
* Enforcing the new types throughout the codebase, plus whatever rewrites happened along the way
* Server.textFiles is now a map
* TextFile no longer uses a fn property, now it is filename
* Added a shared ContentFile interface for shared functionality between TextFile and Script.
* related to ContentFile change above, the player is now allowed to move content from a text file to a script file and vice versa, there wasn't a great reason to disallow this.
* File paths now never start with slashes. All directory names other than root have ending slashes. The player is still able to provide paths starting with / but this now indicates that the player is specifying an absolute path relative to root instead of relative to the current directory.
* Note, relative imports have not been implemented yet, but should be much easier to implement now.
* Singularized the MessageFilename and LiteratureName enums
* Because they now only accept correct types, server.writeToXFile functions now always succeed (the only reasons they could fail before were invalid path).
* Remaining issues: tab completion for directories and the ./ command aren't correct.
* Once the tab completion test failures are fixed, the path test (formerly directory test) also needs attention.


src/Files/ContentFile.ts
```diff
@@ -0,0 +1,19 @@
+import type { BaseServer } from "../Server/BaseServer";
```
d0sboots: Why is this in "Files" but the others are in "Paths"? Seems like an artificial distinction.

src/InteractiveTutorial.ts
```diff
@@ -104,7 +104,10 @@ function iTutorialEnd(): void {
   ITutorial.isRunning = false;
   ITutorial.currStep = iTutorialSteps.Start;
   const messages = Player.getHomeComputer().messages;
-  const handbook = LiteratureNames.HackersStartingHandbook;
+  const handbook = LiteratureName.HackersStartingHandbook;
+  if (!isFilePath(handbook) || !isAbsolutePath(handbook)) {
```
d0sboots: Why do you need this check here (and places like here)? You're dealing with a constant.

src/Literature/Literature.ts
```diff
@@ -1,14 +1,21 @@
...
 export class Literature {
   title: string;
-  fn: string;
+  fn: LiteratureName & FilePath;
   txt: string;
 
-  constructor(title: string, filename: string, txt: string) {
+  constructor(title: string, filename: LiteratureName, txt: string) {
     this.title = title;
+    if (!isFilePath(filename) || !isAbsolutePath(filename)) {
```
d0sboots: This is happening enough that it feels like it should maybe be a helper, where you pass a literal string as part of an error message to use.

src/Literature/Literatures.ts
```diff
@@ -1,14 +1,14 @@
...
 import { FactionNames } from "../Faction/data/FactionNames";
 
-export const Literatures: Record<string, Literature> = {};
+export const Literatures = {} as Record<LiteratureName, Literature>;
 
 (function () {
   let title, fn, txt;
   title = "The Beginner's Guide to Hacking";
-  fn = LiteratureNames.HackersStartingHandbook;
+  fn = LiteratureName.HackersStartingHandbook;
```
d0sboots: Can I just say, that "fn" to me always means "function" and it is really bugging me when reading this code? XD

(No need to change it though)

src/Server/data/servers.ts
```diff
@@ -1251,7 +1258,7 @@ export const serverMetadata: IServerMetadata[] = [
   {
     hackDifficulty: 30,
     hostname: "silver-helix",
-    literature: [LiteratureNames.NewTriads],
+    literature: [Literatures[LiteratureName.NewTriads].fn],
```
d0sboots: This is kinda gross

src/Programs/data/ProgramsMetadata.ts
```diff
@@ -24,17 +25,24 @@ function bitFlumeRequirements() {
...
+  name: ProgramFilePath;
   create: IProgramCreate | null;
   run: (args: string[], server: BaseServer) => void;
 }
 
 export const programsMetadata: IProgramCreationParams[] = [
   {
     key: "NukeProgram",
-    name: "NUKE.exe",
+    name: assignProgramName("NUKE.exe"),
```
d0sboots: Couldn't you leave these as strings, and do the `assignProgramName` once, later?

d0sboots (commented):
Overall, I like the direction this is going. There's some extra boilerplate being introduced that I'm not as much a fan of (particularly in large static-data blocks), but I think that can be eliminated with a little restructuring.

src/Files/ContentFile.ts
Snarling: Yeah I was thinking about moving e.g. TextFile, Script, etc into "Files" (right now TextFiles is in root) but ended up not doing that. It is weird with nothing else in there.

src/InteractiveTutorial.ts
Snarling: This is to provide a runtime error in case someone changes something that makes that constant no longer obey the rules of the type it's supposed to be.

Edit: And also to just assert the right type. Without this check, it can't be added to messages because it's not the right type. I could just raw assert the type as `LiteratureName & FilePath` to allow this, given that it's already typechecked in Literatures, or assign it from that base object instead of just assigning the enum member directly.

src/Literature/Literatures.ts
Snarling: Yeah that edit I didn't manually look at that code, otherwise I would have changed it because I totally agree. The edit was just from using VSCode to rename the enum as a singular name

src/Server/data/servers.ts
Snarling: Yeah I agree. I'll try to think of a way that looks a bit better but especially with these Messages and Literatures it's difficult to get all the right type information to the Server object, since it's an enum that also has to go through typechecking functions (so directly assigning the value from the enum doesn't work).

src/Programs/data/ProgramsMetadata.ts
Snarling: Yeah potentially. I think the other option is basically how I'm doing it in Literatures, where the constructor takes in a parent type (in that case Enum, but for programs it would just be string) and then applies FilePath typechecking on top.

Once I'm done with the remaining autocompletion edits, I will try to make the type assertions for the Lits/Msg/Programs look nicer and all be handled similarly to each other.

Snarling:
Fixed autocompletion to correctly autocomplete things. I think the autocompletion side is 100% working now including a bunch of functionality that was missing before, but of course the tests are broken due to the huge rewrite.

Todos:
* Still need to make sure there are no holes in the command parsing side when it comes to accepting relative file paths.
* Clean up / standardize the way type assertions are being done between the "canonical filepaths' (programs / lit / messages)
* Get tests working again
* Get relative paths on imports working. I made a first attempt but the code for ram calculation parsing was hurting my brain and soul.

d0sboots:
> Get relative paths on imports working

You're probably already aware of this, but just in case: imports have even more cases than regular paths. There's absolute imports (starts with a slash), relative imports (starts with `.` or `..`) and "bare paths" (starts with neither.) For legacy reasons, we have to treat bare paths the same as absolute imports.

Snarling:
> You're probably already aware of this, but just in case: imports have even more cases than regular paths. There's absolute imports (starts with a slash), relative imports (starts with `.` or `..`) and "bare paths" (starts with neither.) For legacy reasons, we have to treat bare paths the same as absolute imports.

Yeah that's the default behavior for resolveFilePath - only treating a path as relative if it contains a `./` or `../`, same limitations apply for e.g. ns.run etc where we need to support absolute pathing as the default.
The terminal is the only place this doesn't apply, so Terminal.getFilePath is a wrapper that forces relative pathing unless the supplied path starts with a `/`

src/InteractiveTutorial.ts
Snarling: I went with just doing the strict typecheck in Literature, Message, and Program constructors, which are also forced to process every enum member due to the types for `Literatures`, `Messages`, and `Programs`.

Then I loosened up the type a little on servers.messages and servers.programs so the hardcoded filenames could be used directly without extra typechecking them every time.

src/Literature/Literatures.ts
Snarling: Went ahead and changed this, as well as the whole structure of Literatures.ts (to match how Programs / Messages initialize their values)

Snarling:
Still need to fix script imports to allow paths that include `./` or `../` directories to be treated as relative.

There doesn't seem to be a simple way to integrate this into the RamCalculation import walk, so this will be accomplished in a follow-on PR.

src/Script/Script.ts
```diff
@@ -32,9 +31,19 @@ export class Script {
    */
   dependencies: Map<ScriptURL, Script> = new Map();
 
-  constructor(fn = "", code = "", server = "") {
+  get content() {
+    return this.code;
+  }
+  set content(code: string) {
+    const newCode = Script.formatCode(code);
```
d0sboots, line 38: This change (minorly) broke how my scripts work. Let me explain.

I have an installer script that updates all my code. Yes, I should probably use Viteburner, but that's not the point here. Before it writes, it checks if the content is the same, and skips writing if it differs. I can use that as a quick sanity check to see that it's updating the expected set of files.

My code has a newline at EOF, because this is The Proper Convention. However, BB's editor strips that newline. No big deal, since I don't use the in-game editor, but it would show files as modified always, even when first opening them.

Now, however, *writing* the files changes the contents, and so my installer script flags them as out-of-date every time. It's impossible to actually write the code as it was written. IMO, this is a bug - only the editor should be reformatting the code.

Snarling: Yeah I see what you mean. Honestly there's not a great reason for the code editor to do it either. Sure we can trim off the trailing whitespace to save a few bytes on the save file, but it's not really meaningful and it's just a weird unexpected behavior in general, e.g. if a player uses a trailing newline, they should expect it's still going to be there when they save and reopen the script.

PR #480, opened 2023-04-19
UI: Enable bang ('!') history autocompletion in the terminal

Daniel-Barbera:
## Description

The user can now use the bang (`!`) token to fetch a command from the history, by one of two methods: 

- Typing `!N`, where `N` is the index of a command in the terminal `history`, and pressing `Tab`.
- Typing `!text` and pressing `Tab`. The terminal will find the most recent use of a command that starts with `text`, and fill it in for the user.

If the user does not press `Tab`, the terminal will attempt to convert the alias to its corresponding `history` entry before executing the commands.

Any feedback is more than welcome.

## Acceptance criteria

Thorough testing is required before merging this PR; any uncaught bugs will negatively impact UX, as the terminal is the core element of the interface.

### DIFF
--- a/src/Terminal/Terminal.ts
+++ b/src/Terminal/Terminal.ts
@@ -570,25 +570,67 @@ export class Terminal {
     }
   }
 
-  executeCommands(commands: string): void {
-    // Sanitize input
-    commands = commands.trim();
-    commands = commands.replace(/\s\s+/g, " "); // Replace all extra whitespace in command with a single space
-
+  executeCommands(commands: string[]): void {
     // Handle Terminal History - multiple commands should be saved as one
-    if (this.commandHistory[this.commandHistory.length - 1] != commands) {
-      this.commandHistory.push(commands);
+    const commandsAsString = commands.join("; ")
+    if (this.commandHistory[this.commandHistory.length - 1] != commandsAsString) {
+      this.commandHistory.push(commandsAsString);
       if (this.commandHistory.length > 50) {
         this.commandHistory.splice(0, 1);
       }
       Player.terminalCommandHistory = this.commandHistory;
     }
     this.commandHistoryIndex = this.commandHistory.length;
-    const allCommands = ParseCommands(commands);
 
-    for (let i = 0; i < allCommands.length; i++) {
-      this.executeCommand(allCommands[i]);
+    commands.forEach(cmd => this.executeCommand(cmd)); 
+  }
+
+  /**
+   * @param command A command to autocomplete from the terminal history. The string must be prepended with a `!` character,
+   * and may be followed by a number to specify the history index of the command to autocomplete from, or a string that will be 
+   * matched against the history.
+   * @returns A command string if there was a successful match, or `undefined` if no command was found.
+   */
+  expandCommandFromHistory(command: string): string | undefined {
+    if (command.charAt(0) !== "!") {
+      return undefined;
     }
+    // Remove the `!` character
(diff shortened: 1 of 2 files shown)


Daniel-Barbera:
Should I add last command auto-completion, too? (`!!`) 🤔

Snarling:
Hey sorry about this, but this will need to be reimplemented since command parsing and autocompletion were completely rewritten on #479 . I did mention when you messaged me on discord that a large-scale rework was in progress that could affect the same stuff you were talking about.

For the autocompletion, it works a bit differently than all other autocompletion, to where you are not just replacing the last argument, you are potentially replacing the entire last command with new text. So this would need some special handling added at TerminalInput. Should be able to just change a regex to capture either last argument (current regex) or the entire command if it starts with a bang.

I don't think it should be supported as a standalone command to be parsed, because this will cause the potential for infinite loops, e.g. if the commands string fetched from the history has another bang command, that then references another bang command etc. Also the bang command would be an individual command, but it would potentially be indicating an entire command array, so it complicates the parsing procedure by allowing travel both ways from command <-> commands. 

For the numeric replacements, note that the terminal commands are only stored as a fixed length array, and eventually older commands fall out of history. Once that starts happening, the indexes for each commands string shifts by 1 every time a new commands string is entered. So this could lead to bad/confusing user experience when using the numeric version.

Daniel-Barbera:
Hey, thanks for the feedback. It was very insightful, and I appreciate it.

>  I did mention when you messaged me on discord that a large-scale rework was in progress that could affect the same stuff you were talking about.

Yeah, that's totally on me. I somehow thought that a filepath rework would not affect the command parser. Shame on me. 🤣 

> Should be able to just change a regex to capture either last argument (current regex) or the entire command if it starts with a bang.

Agreed.

> I don't think it should be supported as a standalone command to be parsed, because this will cause the potential for infinite loops. e.g. if the commands string fetched from the history has another bang command, that then references another bang command etc

The way I handled this was to store the commands in the terminal history the way the terminal interprets them, as opposed to as to how the player writes the command. I am unsure whether this is the best approach, but I believe that's how it's handled in the UNIX shells that support `history`?

> Also the bang command would be an individual command, but it would potentially be indicating an entire command array, so it complicates the parsing procedure by allowing travel both ways from command <-> commands.

Correct. However, I am fairly certain that there is no way to add the feature without complicating the parsing procedure. I'd appreciate some pointers as to which is the correct direction to go.

> For the numeric replacements, note that the terminal commands are only stored as a fixed length array, and eventually older commands fall out of history. Once that starts happening, the indexes for each commands string shifts by 1 every time a new commands string is entered. So this could lead to bad/confusing user experience when using the numeric version.

Thanks for pointing that out. Should we remove the hard limit on the buffer's size, and maybe clear the buffer on soft & hard reset? Or should we just not implement the numeric expansion overall?

---
I should add that that my concern while making this PR is that new players who come from a UNIX background might feel frustrated with the limitations of the current shell, especially since the commands available (`mv`, `ls`, `nano`, `vim`) would suggest that our shell is trying to emulate a UNIX shell. I wanted to start working towards making it feel as UNIX-like as possible, but maybe I made a lot of assumptions in the way. Should the BitBurner shell ever behave like a "real" one, or is that kind of behavior out of scope?

Daniel-Barbera:
I'm closing this for now, as I have no intention of reimplementing it in the near future. I'll reopen when it's ready again.

PR #482, opened 2023-04-20
NETSCRIPT: Change error return of hackAnalyzeThreads

### LINKED ISSUE #102: hackAnalyzeThreads should return more appropriate value when hacking impossible due to security level
[TDSSS]
hackAnalyzeThreads claims it returns the number of threads you need to steal a certain amount of money from a Server.
There is a remark that it returns -1 when the money amount is less than 0 or greater than available money. However, when the server's security level is >= 100, it always returns 0.

**What I expected**
I expected some meaningful value to indicate an error or that no amount of threads can steal the desired amount of money. 0 does not seem to represent that to me and is a very surprising result.

**Suggested Fix**
I see two ways to address this:
a) Return some other meaningful value to indicate that no amount of threads can steal the desired amount of money. You could simply use -1 again, or some other value, like -2 or the like.

b) Update the documentation and function description to indicate that it returns 0 if no amount of threads can steal the desired amount of money.

Edit: hackPercent from the formulas package seems to show the same behavior, as well.

**Info and Reproduction Steps**
Version 2.0.2
Script to reproduce:
Make sure you have not hacked the 'blade' server so its security is >=100 (or use another sever with high initial security)
run:
```/** @param {NS} ns */
	let target = "blade";
	ns.tprint(ns.hackAnalyzeThreads(target, ns.getServerMoneyAvailable(target)));
}
```

Observe it prints 0

[MaWo2]
> I would like to add, that the "result" 0 is also quite unexpected, if the server cannot be hacked, because the player's hacking level is too low.

[Draco18s]
> These two lines would need to be updated to `return -1` instead of `return 0`:
> 
> https://github.com/danielyxie/bitburner/blob/be553f3548b0082794f7aa12c594d6dad8b91336/src/NetscriptFunctions.ts#L155
> 
> https://github.com/danielyxie/bitburner/blob/be553f3548b0082794f7aa12c594d6dad8b91336/src/NetscriptFunctions.ts#L161

[OneOfMany07]
> Why shouldn't it return infinity?  That seems the correct answer to the original situation.
> 
> I mean ignoring what the documentation says.  I don't know if they try to not use infinity in general inside the game.
> 
> As for why you're getting that answer, I assume some math is overloading the possible values of a variable.  Maybe using a higher range variable internally could hold it correctly?  Or changing how you calculate things to avoid a large intermediary value.

[OneOfMany07]
> > I would like to add, that the "result" 0 is also quite unexpected, if the server cannot be hacked, because the player's hacking level is too low.
> 
> What should the answer be instead of 0?  When there is no valid answer...  Either you give an obviously bad result like that (zero or negative or a string?), or throw an exception.  And I don't think I've seen anything use exceptions in here.

[Draco18s]
> > If hackAmount is less than zero or greater than the amount of money available on the server, then this function returns -1.
> 
> Because the docs say it returns -1.
>

d0sboots:
Fixes #102

If we don't do this, we should probably close that issue as WontFix.

### DIFF
--- a/src/NetscriptFunctions.ts
+++ b/src/NetscriptFunctions.ts
@@ -173,7 +173,7 @@ export const ns: InternalAPI<NSFull> = {
     const percentHacked = calculatePercentMoneyHacked(server, Player);
 
     if (percentHacked === 0 || server.moneyAvailable === 0) {
-      return 0; // To prevent returning infinity below
+      return -1; // To prevent returning infinity below
     }
 
     return hackAmount / (server.moneyAvailable * percentHacked);


Snarling:
I'd really rather return `Infinity` here as the "you can't do it" value, but -1 is definitely better than 0.

PR #490, opened 2023-04-24
NETSCRIPT: Move lastAugReset and lastNodeReset info back to main API

Snarling:
Created ns.getResetInfo function, removed the functions from singularity. Provides last reset times as well as current node number. That info is now deprecated on getPlayer.

Also adjusted some other deprecation stuff. This explicit deprecation warning is now provided for nFormat, and its full removal has been pushed to 2.4.


src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -67,6 +66,17 @@ interface SleevePerson extends Person {
   storedCycles: number;
 }
 
+/** Various info about resets
+ * @public */
+interface ResetInfo {
+  /** Numeric timestamp (from Date.now()) of last augmentation reset */
+  lastAugReset: number;
+  /** Numneric timestamp (from Date.now()) of last node reset */
```
d0sboots: Spnelling

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -67,6 +66,17 @@ interface SleevePerson extends Person {
...
+  lastNodeReset: number;
+  /** The current node */
```
d0sboots: I'm fine with calling these `lastNodeReset`/`currentNode` (i.e. being a bit coy in the naming, also it's a bit shorter).
However, I think in the comment we should actually call out that it is the current bitnode. If you don't know what it is yet, "node" and "bitnode" are equivalently opaque, but if you *do* know then "node" is a bit overloaded (Hacknet Node? BN? Something else?) whereas bitnode is unambiguous.

d0sboots (commented):
Looks good. 1GB seems like a reasonable cost.

PR #494, opened 2023-04-28
UI: Move pagination to top of active scripts, improve filtering

### LINKED ISSUE #27: Active scripts navigation UI bounces around when viewing last page
[Sleaker]
For control scripts that actively launch multiple other scripts in quick succession it becomes impossible to navigate the list once the last page has been reached as the footer UI which controls rows per page and back continually pops up and down as scripts are destroyed/started up.

This can make finding a specific script on servers with many active scripts difficult to find in the active scripts page.

UI elements that control navigation should be locked in place elements so they don't bounce around:
Fix 1) move the navigation elements to the header, 
Fix 2) Fill in empty rows until 'rows per page' has been reached.

Side feature: make the search box filter out any scripts that don't have the name, not just the servers with them running.

[phyzical]
> if this hasn't been addressed i think it would be a good idea. i know i've had a hard time clicking the right script sometimes.

Ookamiko:
# UI: Active script page

closes #27

## Change pagination display

Pagination display is now set on top of running servers and on top of active script for each server accordion.
The reason why put it on top rather than setting "blank" space to keep the pagination out of bouncing is because if the see limit is set to a high level (ie. if set to 50) then it may result in a lot of blank useless space.

Pagination bitburner

## Improve filtering of running script
The filtering system on the active script page has been improved. It can work over two possible ways:
- Filter is set to see a specific server: The filter will hide all other servers and show only those who match the filter based on their hostname. As such, all running scripts on those servers will be kept visible.
- Filter is set to see a specific script: The filter will hide all servers who don't run a script matching the filter based on their name (It works like the way the filter currently work). On top of that, the filter will hide all running scripts that don't match the filter based on its name on those servers.

Of course, if the filter matches multiple possibility, it will just show all those possibilities depending if it matches a server or a running script

Filter not set :

Without filter

Filter server :

Filter server

Filter script :

Filter script


ficocelliguy:
In your screenshots, the search bar is set above the pagination. However, they would fit more cleanly together if they were on the same line, leaving less empty space at the top of the page. Can they be aligned like that?

PR #495, opened 2023-04-29
DOC: Grammar & language fixes for v2.0.0 migration

RyanMcEntire:
# Documentation

Small changes were made to improve the grammar and punctuation with minimal changes made to the content.

Line 79 changed a term from `not-optional`, to `required`.

### DIFF
--- a/doc/source/v2.0.0_migration.rst
+++ b/doc/source/v2.0.0_migration.rst
@@ -100,7 +100,7 @@ Singularity
 stock.buy, stock.sell, stock.short
 ----------------------------------
     
-    These functions were renamed to stock.buyStock, stock.sellStock, and stock.buyShort because 'buy', 'sell', and 'short',
+    These functions were renamed to stock.buyStock, stock.sellStock, and stock.buyShort because 'buy', 'sell', and 'short'
     are very common tokens that would trick the ram calculation.
 
 corporation.bribe
--- a/doc/source/v2.0.0_migration.rst
+++ b/doc/source/v2.0.0_migration.rst
@@ -1,32 +1,32 @@
 v2.0.0 Migration Guide
 ======================
 
-In v2.0.0 a few more API have been broken.
+In v2.0.0 a few more APIs have been broken.
 
 Working
 -------
 
-    Working has been rebuilt from the grounds up. The motivation for that change is that all
+    Working has been rebuilt from the ground up. The motivation for this change is that all
     different types of work all required different cached variables on the main Player object.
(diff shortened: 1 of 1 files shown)


doc/source/v2.0.0_migration.rst
```diff
@@ -55,55 +55,55 @@ getPlayer
...
+    The top level singularity functions were deprecated a while ago in favor of the singularity namespace.
     This means calls like 'ns.connect' need to be changed to 'ns.singularity.connect'
 
 
 stock.buy, stock.sell, stock.short
 ----------------------------------
     
-    These functions were renamed to stock.buyStock, stock.sellStock, and stock.buyShort because 'buy', 'sell', and 'short'
+    These functions were renamed to stock.buyStock, stock.sellStock, and stock.buyShort because 'buy', 'sell', and 'short',
     are very common tokens that would trick the ram calculation.
```
Snarling: This added comma is not correct

Snarling (commented):
Good changes overall, just one issue

doc/source/v2.0.0_migration.rst
RyanMcEntire: Hah, I suppose the oxford comma is more a matter of taste. 



Some consider it to be not grammatically correct and redundant, others consider it to make more sense and reflect the parts of speech more clearly.

I tend to usually go for the oxford comma, but it won't bother me to leave it out either.

d0sboots: Ooh it's English pedantics time! :)

Snarling is correct here; this is a different issue than Oxford comma.

Oxford comma is about whether or not to have a comma between the second-to-last item in a list, as your source notes. In the given sentence, it would be the difference between:
"... because 'buy', 'sell', and 'short' are very common tokens..."
and
"... because 'buy', 'sell' and 'short' are very common tokens..."
(2nd one is without the Oxford comma)

Snarling objected to this, which is *not* the Oxford comma:
""... because 'buy', 'sell', and 'short', are very common tokens..."

You could argue for a comma after the first list (before the "because"), but here it feels awkward.

RyanMcEntire: Oof, you guys are right! Thanks for explaining how I was wrong. I got it twisted here, and I didn't re-read the sentence after this review very closely.

That's what I get for assuming 1am me was editing correctly.

RyanMcEntire: Sorry about that, I'll fix that up and remove the comma.

RyanMcEntire:
Ok, I believe it has been fixed after incorporating your feedback.

PR #498, opened 2023-04-29
UI: Fix (most) perf issues in Active Scripts

d0sboots:
This turned out to be remarkably simple: Just delete WorkerScriptStartStopEventEmitter, and let the UI refresh at a constant cadence that is fast enough to be rapid, but slow enough to not burn excessive CPU.

I said that a rework of Active Scripts would/should wait for 2.3, and that's still true. But this is small/simple enough, and (IMO) high value enough, that it's worth considering including earlier.

### DIFF
--- a/src/Netscript/WorkerScriptStartStopEventEmitter.ts
+++ b/src/Netscript/WorkerScriptStartStopEventEmitter.ts
@@ -1,4 +1,0 @@
-import { EventEmitter } from "../utils/EventEmitter";
-
-/** Event emitter that triggers when scripts are started/stopped */
-export const WorkerScriptStartStopEventEmitter = new EventEmitter<[]>();
--- a/src/Netscript/killWorkerScript.ts
+++ b/src/Netscript/killWorkerScript.ts
@@ -5,7 +5,6 @@
 import { ScriptDeath } from "./ScriptDeath";
 import { WorkerScript } from "./WorkerScript";
 import { workerScripts } from "./WorkerScripts";
-import { WorkerScriptStartStopEventEmitter } from "./WorkerScriptStartStopEventEmitter";
 
 import { GetServer } from "../Server/AllServers";
 import { AddRecentScript } from "./RecentScripts";
@@ -91,5 +90,4 @@ function removeWorkerScript(workerScript: WorkerScript): void {
 
   workerScripts.delete(workerScript.pid);
   AddRecentScript(workerScript);
-  WorkerScriptStartStopEventEmitter.emit();
 }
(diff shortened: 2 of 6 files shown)


Snarling (changes requested):
1. Recently killed scripts tab no longer rerenders without leaving and coming back to that tab. The recent scripts page should get a rerender, probably based on a debounced event (see number 3 below) 
2. Total production since last Augment Installation no longer updates without leaving and coming back to active script tab (ScriptProduction component). The ScriptProduction component should get a time-based rerender.
3. I'm not sure removing the event emitter is the right course of action here. Changing to a fixed time rerender is a performance improvement in the case where the player is constantly starting and stopping scripts, but it's a performance penalty when the player is not starting and stopping scripts (e.g. using immortal workers, or just running looping hack scripts). A debounced event-driven rerender would still prevent constant rerendering due to the debounce, and when scripts aren't being stopped or started there would be no need to rerender at all.

If removing the overall rerender, would also need a time-based rerender on the individual WorkerScriptAccordions because they show info based on time elapsed, earnings per second, etc). So maybe it's fine to just keep a time-based rerender and not bring back the event emitter. But the first two issues need addressed either way.

d0sboots:
> 1. Recently killed scripts tab no longer rerenders without leaving and coming back to that tab. The recent scripts page should get a rerender, probably based on a debounced event (see number 3 below)
> 2. Total production since last Augment Installation no longer updates without leaving and coming back to active script tab (ScriptProduction component). The ScriptProduction component should get a time-based rerender.

1 and 2 are done, that was a simple matter of moving the rerender (back) to ActiveScriptsRoot.

> 3. I'm not sure removing the event emitter is the right course of action here. Changing to a fixed time rerender is a performance improvement in the case where the player is constantly starting and stopping scripts, but it's a performance penalty when the player is not starting and stopping scripts (e.g. using immortal workers, or just running looping hack scripts). A debounced event-driven rerender would still prevent constant rerendering due to the debounce, and when scripts aren't being stopped or started there would be no need to rerender at all.
> 
> If removing the overall rerender, would also need a time-based rerender on the individual WorkerScriptAccordions because they show info based on time elapsed, earnings per second, etc). So maybe it's fine to just keep a time-based rerender and not bring back the event emitter. But the first two issues need addressed either way.

We always need/needed a time-based rerender, for stuff like 2. Total money earned can go up without any scripts being launched or killed. Before there was a time-based rerender at 200ms, I upped that to 400ms because 200ms is overkill (2.5 times a second is fast enough for humans, and also if the numbers are really changing constantly it gives you just enough time to read them.)

But the time-based render is also all you need; getting rid of all the other cruft eliminates the performance issues *and* simplifies everything.

Snarling:
Yeah in this case so much of the displayed info already needs to be rerendered based on time that the event approach doesn't save much. The code in ServerAccordions/ServerAccordion could be ran less often by making it run based on a debounced event, but that comes at the cost of needing the event in the first place. And actually there are a couple rare occurrences that would also need to trigger a rerender, like player renaming a purchased server, or buying more ram on home. So I agree that in this case the event-driven rerender approach doesn't offer much and the extra complication is not worth it. This wouldn't be true if WorkerScriptAccordion didn't have so much time-based display data that needs constant recalculating.

Snarling (approved):

PR #500, opened 2023-05-03
NETSCRIPT: Revert FormattedCode type

d0sboots:
Scripts are just strings, formatting should not be enforced upon them.

See the comment in the original PR for more justification


Snarling (approved):
Merging this but afterwards I'm just going to get rid of "formatCode" (which is just a trim) altogether, there's no need to forcibly remove whitespace even when the player saves a script from the script editor, we can just save what the player wrote directly (e.g. if they want to have a newline at the end of their code, they can and it won't be randomly undone by the script editor)

Issue #474, opened 2023-04-11
.eslintrc: duplicate `extends` key disables most of the lint rules


SamB:
It looks like `.eslintrc` enables a healthy chunk of rules here: <https://github.com/bitburner-official/bitburner-src/blob/751fe82f9255cb114207a54ad502101d13f2642e/.eslintrc.js#L7-L12>
Unfortunately, this is overridden by the duplicate key further down in the file: <https://github.com/bitburner-official/bitburner-src/blob/751fe82f9255cb114207a54ad502101d13f2642e/.eslintrc.js#L23>

The [`no-dupe-keys` ](https://eslint.org/docs/latest/rules/no-dupe-keys) rule would catch this, except:
  1. You have not set up eslint to actually lint .eslintrc.
  2. That rule isn't enabled: it's included in the `eslint:recommended` config that never actually got pulled in due to the duplicate key.

(It is somewhat unfortunate that JavaScript allows duplicate literal keys in object literals in the first place.)

I would submit a PR, but unfortunately the obvious fix leaves you with 3K+ warnings and errors to sort out one way or another, and while I expect you'd want to turn off (or configure) some of the relevant rules because of this, there may also be significant amounts of code that you would actually want to change as a result, so I'm not going to attempt to provide a warning-free pull request for this.

d0sboots:
Probably the right thing to do is to delete the duplicate key, and comment out the formerly ignored rules with an explanation. Then we can look at enabling them individually (or a subset of what they're made up of) in smaller chunks.

Hoekstraa:
@Snarling, does #501 close this issue, or is further work needed?

Snarling:
More work still needed on this one.

Alpheus:
Bump. Opened up new PR to make progress on this.

PR #501, opened 2023-05-03
CODEBASE: Expand lint rules, and Aliases are stored as maps

Snarling:
As mentioned in #474, the expected lint rules were not being applied. This PR is to enable the desired lint rules.

So far ada0e13 restores the `eslint:recommended` rules and makes the appropriate codebase changes to comply with the additional rules. Most of the violations were direct usage of hasOwnProperty from an object, so I created a helper function that instead calls hasOwnProperty from the Object prototype and replaced all usage. (also see 94fc8d6)

151daf0 contains mostly fixes the linter is able to automatically apply with `@typescript-eslint/recommended-requiring-type-checking` enabled. Most of these changes are removed type assertions that were unnecessary due to that type already being guaranteed.

94fc8d6 Switches out the custom helper hasOwnProps with Object.hasOwn

7dacaaa Changes aliases to use a map rather than a plain object. This streamlines usage and is just a better data structure for what it is. Worth noting that prior to this PR, there was a bug where setting an alias with the name hasOwnProperty would cause a crash, though that was already addressed by previous fixes.


src/utils/helpers/ObjectHelpers.ts
```diff
@@ -0,0 +1,4 @@
+/** A wrapper for Object.prototype.hasOwnProperty to avoid calling it from objects directly (lint rule). */
+export function hasOwnProp<T extends object>(obj: T, prop: string): prop is keyof T & string {
+  return Object.prototype.hasOwnProperty.call(obj, prop);
```
d0sboots: `Object.hasOwn` is the preferred replacement. (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn) It eliminates the need for this helper method. It has wide support, and if it isn't wide enough yet, Babel will transpile it for us.

Snarling: Done

src/Augmentation/AugmentationHelpers.tsx
```diff
@@ -41,16 +40,12 @@ function createAugmentations(): void {
...
-      Factions[name].augmentations = [];
-    }
-  }
+  for (const faction of Object.values(Factions)) faction.augmentations = [];
 }
 
 function initAugmentations(): void {
   resetFactionAugmentations();
-  clearObject(StaticAugmentations);
+  for (const augName of Object.getOwnPropertyNames(StaticAugmentations)) delete StaticAugmentations[augName];
```
d0sboots, line 48: Mental note: Good candidate for future JSONMap

src/data/codingcontracttypes.ts
```diff
@@ -270,14 +269,16 @@ export const codingContractTypesMetadata: ICodingContractTypeMetadata[] = [
...
-      while (true) {
+      let done = false;
+      while (!done) {
         // Up
         for (let col: number = l; col <= r; col++) {
           spiral[k] = data[u][col];
           ++k;
         }
         if (++u > d) {
-          break;
```
d0sboots, line 280: What was wrong with the breaks here? I liked the old way better.

d0sboots (approved):
Looks good. I'd submit this before it gets any bigger.

src/Augmentation/AugmentationHelpers.tsx
Snarling: Yeah I think StaticAugmentations isn't even from the save file, so it can just be a regular map I think.

The getOwnPropertyNames thing here is weird, I was just keeping the functionality the same from the removed helper function. Iterating through keys should be functionally identical in this case though. But yeah we should just redo it as a map soon.

src/data/codingcontracttypes.ts
Snarling: Linter didn't like us using a constant as the condition for a loop, this was my attempt at satisfying that. I'm not a huge fan of it either though.

PR #502, opened 2023-05-03
BUGFIX: Prevent crash that could occur with server named runningScripts

Snarling:
If a purchased server was named `runningScripts` and the option to exclude running scripts from the save was enabled, the server would be replaced with an empty array, which can lead to a variety of bugs (e.g. scan command crashes)

image

I included a potential fix but this does slow down the save process a bit and I still really don't like this replacer, so I am working on a better solution.

### DIFF
--- a/src/Server/AllServers.ts
+++ b/src/Server/AllServers.ts
@@ -200,10 +200,7 @@ export function loadAllServers(saveString: string): void {
 }
 
 function excludeReplacer(key: string, value: any): any {
-  if (key === "runningScripts") {
-    return [];
-  }
-  return value;
+  return key === "runningScripts" && Array.isArray(key) ? [] : value;
 }
 
 export function saveAllServers(excludeRunningScripts = false): string {
--- a/src/Server/AllServers.ts
+++ b/src/Server/AllServers.ts
@@ -199,10 +199,6 @@ export function loadAllServers(saveString: string): void {
   AllServers = JSON.parse(saveString, Reviver);
 }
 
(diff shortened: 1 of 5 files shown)


src/Server/AllServers.ts
```diff
@@ -200,10 +200,7 @@ export function loadAllServers(saveString: string): void {
 }
 
 function excludeReplacer(key: string, value: any): any {
-  if (key === "runningScripts") {
-    return [];
-  }
-  return value;
+  return key === "runningScripts" && Array.isArray(key) ? [] : value;
```
d0sboots: Shouldn't you be checking value, not key?

d0sboots:
I'm also interested to see your better solution... I'm glad I was able to get rid of the replacer for the normal case.

Snarling:
@d0sboots The new solution just handles it in the BaseServer toJSON, instead of going through the trouble of creating a runningScripts array just to replace it, we just early return with an empty array for runningScripts.

It does require some slight jank over in SaveObject.tsx though, to handle the one weird case where we're forcing the exclusion instead of using the setting

src/Server/AllServers.ts
Snarling: Yeah my bad. I see what happened here, I had it right in some staged changes on a separate branch but then I fat fingered it when I made a new branch for the change (since it wasn't really related to the other branch)
image

d0sboots:
It is a little jank, but the improvements everywhere else more than make up for it, IMO

PR #505, opened 2023-05-08
SETTINGS: Add an autoexec setting

d0sboots:
Instead of working on corp, I did this

This allows for scripts that can be run on game load, without needing a script that is always running to do that for you. Also makes the "don't save scripts" option more useful.

image
image
image
image
image

### DIFF
--- a/src/GameOptions/ui/AutoexecInput.tsx
+++ b/src/GameOptions/ui/AutoexecInput.tsx
@@ -3,7 +3,7 @@ import { Box, InputAdornment, TextField, Tooltip, Typography } from "@mui/materi
 import { Settings } from "../../Settings/Settings";
 import { parseCommand } from "../../Terminal/Parser";
 import { resolveScriptFilePath } from "../../Paths/ScriptFilePath";
-import { GetServer } from "../../Server/AllServers";
+import { formatRam } from "../../ui/formatNumber";
 import CheckCircleIcon from "@mui/icons-material/CheckCircle";
 import WarningIcon from "@mui/icons-material/Warning";
 import ErrorIcon from "@mui/icons-material/Error";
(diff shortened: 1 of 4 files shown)


src/GameOptions/ui/AutoexecInput.tsx
```diff
@@ -0,0 +1,113 @@
...
+        >
+          <CheckCircleIcon color="primary" />
+        </Tooltip>
+      );
+    } else {
+      return (
+        <Tooltip
+          title={
+            <Typography>
+              {cmd} costs {ramUsage}GB, you might only have {minRam}GB on home!
```
Snarling: Should not be directly using GB, need to use formatRam to show a ram number to player in case they are using GiB (not the only line affected)

src/GameOptions/ui/AutoexecInput.tsx
```diff
@@ -0,0 +1,113 @@
...
+        <Tooltip title={<Typography>{cmd} does not end in .js or .script!</Typography>}>
+          <ErrorIcon color="error" />
+        </Tooltip>
+      );
+    }
+    const home = GetServer("home");
+    if (!home) {
+      return (
+        <Tooltip title={<Typography>Internal error: Can't find home!</Typography>}>
+          <ErrorIcon color="error" />
```
Snarling: For this I have been using Player.getHomeComputer which always returns a server and handles throwing if it's somehow not there. It's not an error case that should ever happen so we don't need to gracefully present that error to the player

src/GameOptions/ui/AutoexecInput.tsx
```diff
@@ -0,0 +1,113 @@
...
+        <Tooltip title={<Typography>No script will be auto-launched</Typography>}>
+          <CheckCircleIcon color="primary" />
+        </Tooltip>
+      );
+    }
+    const cmd = String(args[0]);
+    const scriptPath = resolveScriptFilePath(cmd);
+    if (!scriptPath) {
+      return (
+        <Tooltip title={<Typography>{cmd} does not end in .js or .script!</Typography>}>
```
Snarling: This error could also be due to using a forbidden character, the error should just say that it's not a valid filepath

Snarling (changes requested):

Snarling:
Also not mentioned in review, using autoexec has the potential to cause "can't launch due to insufficient ram" errors even when the script has less than the player's home computer ram. The autoexec script will load, but later scripts may not and the error message will be confusing "This might be due to a change in ramcost etc." but the actual reason was that you shoved the script out of available ram with autoexec

src/GameOptions/ui/AutoexecInput.tsx
d0sboots: Done

src/GameOptions/ui/AutoexecInput.tsx
d0sboots: Done

src/GameOptions/ui/AutoexecInput.tsx
d0sboots: Done

d0sboots:
Good suggestions, all.
I removed the bit about "the RAM cost might have changed since the script was run" - that's no longer relevant (not sure when it was relevant, actually) since ramUsage is saved with RunningScripts.

image
image

PR #508, opened 2023-05-09
UI: LogBox overhaul

d0sboots:
This ended up being bigger than I expected.

----

This revamps a lot of the internals of LogBoxManager, although to users it should act entirely the same, and the most visible difference will be the use of icons instead of text for the buttons.

Added the ability to get/set the title, including having it be saved and setting it to arbitrary ReactElements for the brave.

Added the ability to read the log window's size/position, as well as determine if it is currently shown.

This shows how you can make multiple copies of the same script window by using rerun from the log window. It's maybe not the best behavior, but it's an interesting edge case and one I thoroughly tested things with. Making that not happen at all is tricky.
image

When running, new minimum size:
image

Here's my test script, it's very useful for testing a variety of behavior:
```js
/** @param {NS} ns */
  ns.tail();
  ns.printf("I am pid %d", ns.pid);
  const promise = ns.asleep(10000);
  ns.disableLog("ALL");
  doTitle(ns);
  await promise;
  ns.closeTail();
}

/** @param {NS} ns */
async function doTitle(ns) {
  let tp = null;
  while (true) {
    await ns.asleep(0);
    const newTp = ns.getRunningScript().tailProperties;
    if (newTp && (tp === null || tp.width !== newTp.width || tp.height !== newTp.height)) {
      tp = newTp;
      ns.setTitle(`${tp.width} ${tp.height}`);
      ns.printf(`${tp.width} ${tp.height}`);
    }
  }
}
```


d0sboots:
After taking a step back for a bit, I realized actually the solution was very simple.

Edit: For fun
```js
/** @param {NS} ns */
  ns.tail();
  ns.disableLog("ALL");
  ns.printf("Yes, this is a real script");
  let title = "                                                                                                                                                  WHEEE";
  while(true) {
    ns.setTitle(title);
    await ns.sleep(title.charAt(0) == " " ? 5 : 70);
    title = title.slice(1) + title.slice(0, 1);
  }
}
```

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -198,6 +210,15 @@ interface RunningScript {
...
   server: string;
+  /** Properties of the tail window, or null if it is not shown */
+  tailProperties: TailProperties | null;
+  /**
+   * The title, as shown in the script's log box. Defaults to the name + args,
+   * but can be changed by the user. If it is set to a React element (only by
+   * the user), that will not be persisted, and will be restored to default on
+   * load.
+   */
+  title: string | React.ReactElement;
```
Snarling: Can you try regenerating docs? I have a feeling these references to React will not do well, either erroring out or just making bad .md output. Will be able to test this in a little while.

Snarling (commented):
Haven't had a chance to do any testing, will try to take a closer look later tonight.

src/ScriptEditor/NetscriptDefinitions.d.ts
d0sboots: Whoops, forgot to do that.

Looks like it comes out fine. (I am also mildly surprised.) ReactElement isn't a link, but renders OK.

Snarling:
Nice, no issues with the doc generating. I figured it would complain about "what is React"

The React usability is pretty bad.

1. If the user has any error in their React element's code (e.g. a misspelled variable name) they are led to the recovery screen. In the case I ran into, I couldn't even close out of the recovery screen, I had to reload the page. If I had the script auto-launching with tail window up then I would need to load with noScripts to be able to load in.

2. Adding react content in general sucks, because there is zero cascading of styles (I intend to fix that during 2.3.1). If the player doesn't explicitly provide font/color, then the text is some black (invisible with the default theme) times new roman looking font.

3. The titlebar deforms and is no longer clamped to the right side of the titlebar when a react element is provided. Even an advanced player willing to use React-without-jsx should not have to provide extra styles for correctly positioning the existing titlebar elements.

https://github.com/bitburner-official/bitburner-src/assets/84951833/d9ea02b1-7592-4715-be93-ea5a7ae066c9

Here's the test code I used:

```js
/** @param {{title: string, fontFamily: string, color: string}} */
function RotatingTitleBar({ title, fontFamily, color }) {
  const [offset, setOffset] = React.useState(0);
  React.useEffect(() => {
    const interval = setInterval(() => setOffset(n => (n + 1) % title.length), 50);
    return () => clearInterval(interval);
  }, []);
  return React.createElement("span", { style: { color, fontFamily } }, title.substring(offset) + title.substring(0, offset));
}

/** @param {NS} ns */
  ns.setTitle(React.createElement(RotatingTitleBar, {
    title: "test title",
    fontFamily: ns.ui.getStyles().fontFamily,
    color: ns.ui.getTheme().primary
  }));
}
```

I think 1 could just be addressed by making it more clear in documentation that using the React option is use-at-own-risk and not supported. 3 can be addressed by wrapping the player's React content in a div that contains the right styles for positioning the rest of the titlebar. 2 is how all the other printRaw stuff already works so I don't think there's a need to address it specifically, and it should be fixed by style changes in (hopefully) 2.3.1

d0sboots:
For 1, what were you thinking for a more explicit warning? Because I literally already say it might break things and it's use at your own risk.

d0sboots:
Another option someone mentioned would be to drop the React support, but run the title through ANSITypography

Snarling:
> For 1, what were you thinking for a more explicit warning? Because I literally already say it might break things and it's use at your own risk.

Yeah that's true, my bad.

> Another option someone mentioned would be to drop the React support, but run the title through ANSITypography

I'm good with either way (we could even support both - if the player provides a string, display an ANSITypography of their text, if they provide their own React element, display that instead).

Snarling:
Okay I think I found where referencing React in the docs causes an issue:

image
image

The related types end up just presenting as `any` in the Script editor because the React stuff can't be resolved in the definition file. Could just typedocument them as string, but note in the description that for advanced users/unsupported behavior, a React element can be provided instead.

d0sboots:
I wasn't assigning any high priority to this, because it's just some UI stuff, but I just remembered that we do probably want to revert the title change before we push.

I can hustle on this if you want to get this in, or you can just revert that change on its own.

Snarling:
Good catch. For now will just do a revert on the title change, we can finish this up as one of the early 2.3.1 commits.

d0sboots:
I realized that the bug Jakob was having (unable to open tail windows for each script if the arguments are the same) was something that got fixed in this PR, so I should devote more energy to it.

I rebased the existing changes, but nothing beyond that yet. I plan to incorporate your React ErrorBoundary stuff.

Snarling:
All good, let me know if you want any assistance with the html/css stuff to get default react content displaying without shifting the titlebar buttons.

As far as the typedoc, I think we can just document it as a string and then have in the notes that for advanced usage, a React node can also be provided. Although that would mean people using ts would get errors if they try sending in react content, which sucks. But having it typedocumented as any in the ingame editor is also lame, seems we will have to deal with one or the other.

d0sboots:
OK, I think I've addressed all the outstanding issues

- Use CustomBoundary to catch React errors
- Wrap custom content in Typography, so it has reasonable default styling and font choice
- Introduce custom ReactElement type to NetscriptDefinitions, so in-game type hinting works properly

Here is your test code running:
image
image
image

Also if you are using custom react elements, the hover text will be the default name/args (even if the component errors)

Snarling:
Looks good!

PR #512, opened 2023-05-16
DOCS: Clarify exec()

svgPhoenix:
Explicitly state that the file to be executed needs to exist on the target server.


d0sboots:
Looks good to me, but Snarling has final call on merging. He might wait until after 2.3 is stabilized.

PR #513, opened 2023-05-16
UI: Better support for custom player React content

Snarling:
* Prevent erroring out the page / telling user to submit github issue when their own custom react content has errors
* Provide default styling for text content printed with printRaw (other content types will still not have the correct default styles applied)

Basically adds an extra error boundary to prevent sending errors in custom player code up to the main ErrorBoundary that handles other React errors in the game.

### DIFF
--- a/src/NetscriptFunctions/Extra.tsx
+++ b/src/NetscriptFunctions/Extra.tsx
@@ -21,7 +21,7 @@ export interface INetscriptExtra {
   bypass(doc: Document): void;
   alterReality(): void;
   rainbow(guess: string): void;
-  iKnowWhatImDoing(): void;
+  tprintRaw(value: React.ReactNode): void;
   printRaw(value: React.ReactNode): void;
 }
 
@@ -71,12 +71,10 @@ export function NetscriptExtra(): InternalAPI<INetscriptExtra> {
       Player.giveExploit(Exploit.INeedARainbow);
       return true;
     },
-    iKnowWhatImDoing: (ctx) => () => {
-      helpers.log(ctx, () => "Unlocking unsupported feature: window.tprintRaw");
-      // @ts-expect-error window has no tprintRaw property defined
-      window.tprintRaw = (value: React.ReactNode) => {
-        Terminal.printRaw(<CustomBoundary key={`PlayerContent${customElementKey++}`} children={value} />);
-      };
+    tprintRaw: () => (value) => {
+      Terminal.printRaw(
+        <CustomBoundary key={`PlayerContent${customElementKey++}`} children={value as React.ReactNode} />,
+      );
     },
     printRaw: (ctx) => (value) => {
       ctx.workerScript.print(
--- a/src/NetscriptFunctions/Extra.tsx
+++ b/src/NetscriptFunctions/Extra.tsx
@@ -1,4 +1,4 @@
-import type React from "react";
+import React from "react";
 import { Player } from "../Player";
 import { Exploit } from "../Exploits/Exploit";
 import * as bcrypt from "bcryptjs";
@@ -7,6 +7,10 @@ import { InternalAPI } from "../Netscript/APIWrapper";
 import { helpers } from "../Netscript/NetscriptHelpers";
 import { Terminal } from "../Terminal";
 import { RamCostConstants } from "../Netscript/RamCostGenerator";
+import { CustomBoundary } from "../ui/Components/CustomBoundary";
+
+// Incrementing value for custom element keys
+let customElementKey = 0;
 
 export interface INetscriptExtra {
   heart: {
@@ -70,11 +74,14 @@ export function NetscriptExtra(): InternalAPI<INetscriptExtra> {
     iKnowWhatImDoing: (ctx) => () => {
       helpers.log(ctx, () => "Unlocking unsupported feature: window.tprintRaw");
       // @ts-expect-error window has no tprintRaw property defined
-      window.tprintRaw = Terminal.printRaw.bind(Terminal);
+      window.tprintRaw = (value: React.ReactNode) => {
+        Terminal.printRaw(<CustomBoundary key={`PlayerContent${customElementKey++}`} children={value} />);
+      };
     },
     printRaw: (ctx) => (value) => {
-      // Using this voids the warranty on your tail log
-      ctx.workerScript.print(value as React.ReactNode);
+      ctx.workerScript.print(
+        <CustomBoundary key={`PlayerContent${customElementKey++}`} children={value as React.ReactNode} />,
+      );
     },
   };
 }
--- a/src/ui/Components/CustomBoundary.tsx
+++ b/src/ui/Components/CustomBoundary.tsx
@@ -35,7 +35,7 @@ export class CustomBoundary extends React.Component<PlayerBoundaryProps, PlayerB
  * correctly by the PlayerBoundary. This function checks the tree for a function before render. */
 function probeChildrenForFunction(tree: string[], children: unknown): undefined | Error {
   switch (typeof children) {
-    case "function":
+    case "function": {
       const error = new Error(
         `React content tree contains a function (invalid): ${[...tree, "function"].join(" -> ")}.`,
       );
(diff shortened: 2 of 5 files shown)


Snarling:
Here are some examples of the errors generated

**Error for function existing as a child anywhere in the tree of react children** (handled separately from other errors, as it's not caught the same way):
```js
const el = React.createElement;

function TestComponent(props) {
  return props.children;
}

/** @param {NS} ns */
  ns.tail();
  ns.printRaw(
    el(TestComponent, null,
      el("span", null,
        el("div", null,
          el(TestComponent, null,
            TestComponent // TestComponent incorrectly used as a child directly
          )
        )
      )
    )
  );
}
```
image
image

**Generic runtime error while rendering a component:**
```js
const el = React.createElement;

function TestComponent(props) {
  return somethingUndefined.property;
}

/** @param {NS} ns */
  ns.iKnowWhatImDoing();
  tprintRaw(el(TestComponent));
}
```
image
image

src/NetscriptFunctions/Extra.tsx
```diff
@@ -70,11 +74,14 @@ export function NetscriptExtra(): InternalAPI<INetscriptExtra> {
     iKnowWhatImDoing: (ctx) => () => {
       helpers.log(ctx, () => "Unlocking unsupported feature: window.tprintRaw");
       // @ts-expect-error window has no tprintRaw property defined
-      window.tprintRaw = Terminal.printRaw.bind(Terminal);
+      window.tprintRaw = (value: React.ReactNode) => {
+        Terminal.printRaw(<CustomBoundary key={`PlayerContent${customElementKey++}`} children={value} />);
+      };
     },
     printRaw: (ctx) => (value) => {
-      // Using this voids the warranty on your tail log
```
d0sboots, line 76: Aww, I kind of liked the warning. XD Do we not need it anymore?

src/ui/Components/CustomBoundary.tsx
```diff
@@ -0,0 +1,75 @@
...
+    if (this.state.error) {
+      // Typography is used because there are no default page styles.
+      // Span is used because it does not conflict with the DOM validation nesting (default Typography element of p is invalid at this location in dom tree)
+      return <Typography component={"span"}>Error in custom react content. See console for details.</Typography>;
+    }
+    return <Typography component={"span"}>{this.props.children}</Typography>;
+  }
+}
+
+/** Attempting to render a react tree containing a function as a child will cause an error that doesn't get displayed
```
d0sboots: Doesn't get displayed correctly how?
This seems like rather heavyweight error-checking to be doing on every call, so it would be really nice to not need to do it. It also seems like something that should be React's responsibility.

d0sboots (commented):
I'll update my change to use this once it gets submitted.

src/ui/Components/CustomBoundary.tsx
Snarling: Yeah so the Boundary is there to try and catch any errors, that way they can be communicated to the player by displaying their content as an error message and directing them to the console so they are aware of the error they have in their code.

Without this checking, if the player provides a function, the game will display an empty span (which is usually invisible) instead of a span telling the player they have an error, and will throw a different error that the player would just have to happen to check the console to ever see, that only displayed once instead of every time the player attempts to render invalid content:

```js
const el = React.createElement;

function TestComponent(props) {
  return props.children;
}

/** @param {NS} ns */
  ns.tail();
  ns.tprint("Test");
  console.log("Test");
  ns.iKnowWhatImDoing();
  tprintRaw(TestComponent)
  ns.printRaw(
    el(TestComponent, null,
      el("span", null,
        el("div", null,
          el(TestComponent, null,
            TestComponent // TestComponent incorrectly used as a child directly
          )
        )
      )
    )
  );
  console.log("Test2")
  ns.tprint("Test2");

}
```
image
image

Console on first run:
image

Console on subsequent runs (no error message at all):
image

d0sboots: I'm really torn about this. On one hand, React *does* provide a useful error message. On the other hand, it's only on console, and only the first time. But if you're using React components, you should be more willing to look at the console for errors. But, this is BitBurner, and not everyone is familiar with general webdev. *But*, use of React components *is* an advanced concept. Etc...

If there's any way to catch results like this and error-check the input for better messaging *after*, instead of doing a recursive walk of *every* component, that would make me a lot happier.

d0sboots: Oh, the one other thing: How do we know that functions are the only thing that errors like this? I'm guessing there are a number of other possible failure modes that act in a similar way. (But maybe they aren't as likely to be hit, so don't justify special handling?)

src/NetscriptFunctions/Extra.tsx
Snarling: The idea is that if we can provide adequate error checking / prevent react misuse from causing the actual game to crash, then it no longer needs to be an "advanced users only" feature with all sorts of warnings that it's unsupported functionality. Players could still have scripts that don't work if they don't use it right, but it isn't going to infect the rest of the game and send them to an error screen.

src/ui/Components/CustomBoundary.tsx
Snarling: >If there's any way to catch results like this and error-check the input for better messaging after, instead of doing a recursive walk of every component, that would make me a lot happier.

This would really only need to be checked on the initial creation of the react element, not on every rerender cycle.  May need to slightly rework it to function that way, I'm not solid enough on how class components work to know 100% if the constructor gets called every render cycle or just on initial render (I'm pretty sure it's called just once but I might be wrong)

> Oh, the one other thing: How do we know that functions are the only thing that errors like this? I'm guessing there are a number of other possible failure modes that act in a similar way. (But maybe they aren't as likely to be hit, so don't justify special handling?)

I sent every normal type of input in, and many of them did not display anything (boolean t/f, bigint, symbol, undefined, null) but didn't cause any errors. Object types that were not components or arrays caused an error, but the error was properly caught by the boundary and displayed normally. Functions were the only inputs in my testing that errored out but not in a way that was caught by the boundary and displayed to the user.

All I tested were types for `children` though. It's possible a similar error is created when sending some invalid type for `props` (although I have a feeling that will error out at the React.createElement call). Will do some testing on that tomorrow

d0sboots: > This would really only need to be checked on the initial creation of the react element, not on every rerender cycle.

But for function components, you *are* recreating the element every time the content changes. For instance, if you're using `printRaw` to add some extra styling to your logs, this checking will happen on every print. Potentially multiple times, if you construct your logs with multiple print statements. That's what I'm concerned about.

> I sent every normal type of input in, and many of them did not display anything (boolean t/f, bigint, symbol, undefined, null) but didn't cause any errors.
...
Functions were the only inputs in my testing that errored out but not in a way that was caught by the boundary and displayed to the user.

That's because the functions aren't actually erroring out. They're printing a *warning* to the console (despite the fact that React prints all its warnings with `console.error`) but otherwise ignoring it just like with other bad input such as bigint and symbols.

You could argue that this is poor design on React's part (I would), but I don't think the fact that React shows *extra* debugging info for this one case obligates us to treat it specially. If we are going to do hand-holding-type error checking (which increasingly I feel we shouldn't), it should focus on the bad inputs that React provides *no* feedback for.

d0sboots: Followup because I'm concerned I didn't state myself clearly:

I'm all for the new CustomBoundary component, I think that's great! I'm just distinguishing between errors that are *thrown* (and thus potentially crash the game) vs. errors (or in many cases, warnings) that are only printed to the console.

For the latter, there's no way we can detect them after-the-fact, but also no real need. They don't cause *direct* problems from a gameplay perspective; the only issue is that the code isn't doing what the user wanted, and they have to debug that. IMO React has *many* challenges in that department, and we won't be able to find and catch all of them ahead-of-time without requiring the user to watch the console.

Snarling: Yeah you're right about it just being a warning. Objects (that aren't components or arrays) produce an actual error with very similar wording which is part of what was tripping me up there (plus console.error being used for a warning).

image

I agree that there's no need to scour the react content to check for warnings, that whole section will be removed.

PR #515, opened 2023-05-17
DOC: Specify optional parameters in hacknet functions

CrafterKolyan:
Seems that this change is pretty trivial. These parameters are optional and default to 1. Found it confusing in help VS Code UI that it didn't state that these parameters are optional while functions work without specifying them.

Proof that arguments are optional and default to 1:
https://github.com/bitburner-official/bitburner-src/blob/75706d273a915063bb094011426c94ec33150466/src/NetscriptFunctions/Hacknet.ts#L94-L175

# Documentation

### DIFF
--- a/markdown/bitburner.hacknet.getcacheupgradecost.md
+++ b/markdown/bitburner.hacknet.getcacheupgradecost.md
@@ -9,7 +9,7 @@ Calculate the cost of upgrading hacknet node cache.
 **Signature:**
 
 ```typescript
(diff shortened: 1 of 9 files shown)


d0sboots:
Looks good to me, but Snarling will need to merge.

Snarling:
PR changes looked good, merged.

Not directly related to this PR, but it's weird that we round the input to the nearest integer. It would make sense on these to only accept positive integers instead of accepting any input and just rounding it. Might change that in 2.3.1

PR #520, opened 2023-05-19
DOC: Fix link on main page of documentation

CrafterKolyan:
Fix link on main page of documentation as https://bitburner-official.github.io/bitburner/ gives 404


d0sboots:
I think that link is *intended* to work, but hasn't been set up yet

Snarling:
Yeah I will see again about getting that other address set up, and we'll change it back if we are able to do that.

PR #525, opened 2023-05-23
DOC: Fix parameter name in documentation

CrafterKolyan:
VS code shows warning


d0sboots:
Unrelated to this change directly: The recommended workflow is to make a (throwaway) branch for each PR, and leave your dev branch as a clean fork of upstream. It's generally less hassle on your end, and keeps the commit history from steadily becoming more cluttered.

Snarling:
Yeah because it's showing all these other old commits as "part of this PR", it's also auto-suggesting their descriptions for a description of the merged PR:

image

Another way to avoid this is to just rebase your dev to the upstream dev after I've merged one of your previous PRs, which should remove the commits have already been merged upstream. But separate branches is usually what people go with since it would also allow you to have two independent PRs that are both submitted at the same time.

PR #526, opened 2023-05-24
NETSCRIPT: Basic implementation of ns.getFunctionRamCost

G4mingJon4s:
# NETSCRIPT: Basic implementation of ns.getFunctionRamCost

Added a netscript function that takes in a function name and returns the ram cost of that function.

## Actions
- [X] `npm run lint`
- [X] `npm run format`
- [X] `npm run doc`


src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -6510,6 +6510,18 @@ export interface NS {
...
 
+  /**
+   * Get the ram cost of a netscript function.
+   *
+   * If the given function does not exist, the function will return 0.
+   *
+   * @remarks
+   * RAM cost: 0 GB
+   *
+   * @param name - The function name. Also works with interfaces: `stock.getPosition`
```
d0sboots: I might reword this a little, because I'm already guessing people will get confused (regardless). Suggestion:

The fully-qualified function name, without the leading `ns`. Examples: `hack`, `tprint`, `stock.getPosition`.

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -6510,6 +6510,18 @@ export interface NS {
    * */
   getResetInfo(): ResetInfo;
 
+  /**
+   * Get the ram cost of a netscript function.
+   *
+   * If the given function does not exist, the function will return 0.
+   *
+   * @remarks
+   * RAM cost: 0 GB
```
d0sboots, line 6517: You should double-check as well, but RamCost currently has exactly one caller, which is APIWrapper. I think all the places where it is currently returning 0 should be throwing Errors instead; *if* it were currently ever possible to hit any of those cases, it would probably lead to a free RAM exploit.

This would also be much better now that you're adding an outside use, since 0 is indistinguishable from a real RAM cost.

d0sboots (commented):
We'll see if Snarling agrees with my plan for GetRam.

src/ScriptEditor/NetscriptDefinitions.d.ts
G4mingJon4s: Yeah, I think so too. I just hadn't had the to check if I would break anything when changing the function. First I thought of returning -1, but that could lead to ram being removed from the dynamic checker...
Throwing an error seems a bit harsh though...

src/ScriptEditor/NetscriptDefinitions.d.ts
G4mingJon4s: If I understand your suggestion correct, you mean something like this:
```ts
const ramCost = ns.getFunctionRamCost("stock.getPosition");
```
This is the current implementation, no `ns.` needed upfront

d0sboots: My suggestion is for the wording of the doc comment.

src/ScriptEditor/NetscriptDefinitions.d.ts
d0sboots: It's true that most ns functions tend to return error values instead of throwing. However, there is precedent: All the functions will throw if the arguments are messed up enough (missing argument, argument is the wrong type, etc), throw when trying to use functions you don't have the SF for, if you try to use port crackers without having them installed, etc.

Also, IMO it's better to throw when something is unambiguously a programmer error. In that case, it means you messed up and should be informed as visibly as possible, instead of having to hunt down "why am I getting this weird value?"

Snarling: Totally agree that throwing is better than just returning a numerical value if the player provides invalid data. e.g. imagine a player tries to check `ns.getFunctionRamCost("stock.getPositon")` (typo in function name), would that player rather get an error (telling them stock.getPositon is not a valid function name), forcing them to fix their code, or just get a mysterious numerical value, potentially causing some issue further in their code that then becomes hard to even troubleshoot.

PR #536, opened 2023-05-27
BUGFIX: Various fixes to our upgrade logic

### LINKED ISSUE #532: Recovery mode activated after loading old save
Not sure if this is an issue that is actionable, but the game suggested I upload something anyway!
The save is from around Feburary of 2022, and was renamed from .json because github refused it otherwise

[RECOVERY_BITBURNER_1685130768.txt](https://github.com/bitburner-official/bitburner-src/files/11579575/RECOVERY_BITBURNER_1685130768.txt)

Version: v2.3.0 (3e14a0839)

d0sboots:
Stuff broke over time, especially with the major changes we made leading up to 2.3. We should test with older saves if/when we make large changes in the future.

Fixes #532

### DIFF
--- a/src/utils/v2APIBreak.ts
+++ b/src/utils/v2APIBreak.ts
@@ -229,7 +229,7 @@ export const v2APIBreak = () => {
   }
 
   // API break function is called before the version31 2.3.0 changes, scripts are still an array.
-  for (const script of home.scripts as unknown as Script[]) {
+  for (const script of home.scripts.values() as unknown as Script[]) {
     processScript(rules, script);
   }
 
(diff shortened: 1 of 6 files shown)


src/utils/v2APIBreak.ts
```diff
@@ -229,7 +229,7 @@ export const v2APIBreak = () => {
   }
 
   // API break function is called before the version31 2.3.0 changes, scripts are still an array.
```
Snarling, line 231: This comment can be removed now that 2.3 migration code is actually called first, just noting it so we can remove at some point. Will go ahead and merge this -> cherry pick over to stable -> rerelease new hotfix

d0sboots: Whoops, missed that one

PR #543, opened 2023-05-29
BUGFIX: More savegame loading fixes

### LINKED ISSUE #540: Recovery mode on game start
- Save file: [RECOVERY_BITBURNER_1685276564.zip](https://github.com/bitburner-official/bitburner-src/files/11584553/RECOVERY_BITBURNER_1685276564.zip)
- Steps to reproduce: Load save file 
- Version of the game: v2.3.0 (037149a16)

It has been a bit since I played this game, booted it up and got this. Looking at when I got my last achievements, I was playing on v1.6.3 previously.

### LINKED ISSUE #541: Uncaught Promise Error
[bitburnerSave_1685308460_BN1x1.json.gz](https://github.com/bitburner-official/bitburner-src/files/11585894/bitburnerSave_1685308460_BN1x1.json.gz)

Havent played in a while and wanted to get back into it but anything I try to type in terminal return an uncaught promise error (ls, run, nano, etc.).
Also tried an older save with same result.
BitBurner v2.3.0

d0sboots:
* Fix loading issues back to pre-1.0
* Be more robust about issues with files not being maps
* Avoid non-fatal error when there's no LastExportBonus

Fixes #540, fixes #541
Also fixes a savegame reported in #bug-report

### DIFF
--- a/src/Server/BaseServer.ts
+++ b/src/Server/BaseServer.ts
@@ -316,14 +316,15 @@ export abstract class BaseServer implements IServer {
   static fromJSONBase<T extends BaseServer>(value: IReviverValue, ctor: new () => T, keys: readonly (keyof T)[]): T {
     const server = Generic_fromJSON(ctor, value.data, keys);
     server.savedScripts = value.data.runningScripts;
-    // Possibly migrate to using maps for scripts and textfiles.
-    // This is done here, directly at load, instead of the usual upgrade
-    // logic, for two reasons:
-    // 1) Our utility functions depend on it, so the upgrade logic itself needs
-    //    the data to be in maps, even the logic written earlier than 2.3!
-    // 2) If the upgrade logic throws, and then you soft-reset at the recovery screen (or
-    //    maybe don't even see the recovery screen), you can end up with a "migrated"
-    //    save that still has arrays.
+    // If textFiles is not an array, we've already done the 2.3 migration to textFiles and scripts as maps + path changes.
+    if (!Array.isArray(server.textFiles)) return server;
+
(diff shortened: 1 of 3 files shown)


src/Server/BaseServer.ts
```diff
@@ -313,9 +314,55 @@ export abstract class BaseServer implements IServer {
...
+        textFile.filename = newFilePath;
+        server.textFiles.set(newFilePath, textFile);
+      }
+    }
+    if (invalidScriptCount || invalidTextCount) {
+      // If we had to migrate names, don't run scripts for this server.
+      server.savedScripts = [];
+    }
+    return server;
   }
```
Snarling: Isn't all this migration code going to be ran every single time the game is loaded, instead of just when the version warrants the migration code to run?

We can choose to only run the migration code if textFiles is still an array, because if textFiles is a map already then the migration has already occurred.

Snarling (commented):
Looks good other than just the one issue noted. I'll probably just edit that fix in and merge this so it can be released

src/Server/BaseServer.ts
d0sboots: Yes, that was intentional. Because of various things like the different timing of introduction of maps vs path-checking code, I wasn't sure that there was a safe "it's definitely been correctly migrated" signal.

However, the presence of a map for text files is *probably* good enough.

One thing this whole migration has shown me is that we're *definitely* better off using the approach I took for RunningScripts: leave the save format the same (a simple array), and transform it to a map on save/load. This keeps backwards/forwards compatibility and just generally saves a lot of headaches.

PR #544, opened 2023-05-29
UI: Prevent terminal scroll on progress bar update (#94)

### LINKED ISSUE #94: Scrolling up while running backdoor is impossible
[kotenok2000]

When i run backdoor on n00dles and scroll up terminal resets down when progress bar moves.

- [ ] Save file
- [ ] Minimal scripts to reproduce the issue
- [x] Steps to reproduce
- [x] Version of the game:1.7.0

bezrodnov:
UI: Prevent terminal scroll on progress bar update

closes #94 

Tested locally in the browser after starting server via `npm run start:dev`

### DIFF
--- a/src/Terminal/Terminal.ts
+++ b/src/Terminal/Terminal.ts
@@ -6,7 +6,7 @@ import { BaseServer } from "../Server/BaseServer";
 import { Server } from "../Server/Server";
 import { CompletedProgramName } from "../Programs/Programs";
 import { CodingContractResult } from "../CodingContracts";
-import { TerminalEvents, TerminalClearEvents, TerminalProcessEvents } from "./TerminalEvents";
+import { TerminalEvents, TerminalClearEvents } from "./TerminalEvents";
 
 import { TextFile } from "../TextFile";
 import { Script } from "../Script/Script";
@@ -102,7 +102,6 @@ export class Terminal {
     if (this.action === null) return;
     this.action.timeLeft -= (CONSTANTS.MilliPerCycle * cycles) / 1000;
     if (this.action.timeLeft < 0.01) this.finishAction(false);
-    TerminalProcessEvents.emit();
   }
 
   append(item: Output | Link | RawOutput): void {
--- a/src/Terminal/Terminal.ts
+++ b/src/Terminal/Terminal.ts
@@ -6,7 +6,7 @@ import { BaseServer } from "../Server/BaseServer";
 import { Server } from "../Server/Server";
 import { CompletedProgramName } from "../Programs/Programs";
 import { CodingContractResult } from "../CodingContracts";
-import { TerminalEvents, TerminalClearEvents } from "./TerminalEvents";
+import { TerminalEvents, TerminalClearEvents, TerminalProcessEvents } from "./TerminalEvents";
 
 import { TextFile } from "../TextFile";
 import { Script } from "../Script/Script";
@@ -102,7 +102,7 @@ export class Terminal {
     if (this.action === null) return;
     this.action.timeLeft -= (CONSTANTS.MilliPerCycle * cycles) / 1000;
     if (this.action.timeLeft < 0.01) this.finishAction(false);
-    TerminalEvents.emit();
+    TerminalProcessEvents.emit();
   }
 
   append(item: Output | Link | RawOutput): void {
--- a/src/Terminal/ui/TerminalActionTimer.tsx
+++ b/src/Terminal/ui/TerminalActionTimer.tsx
@@ -1,5 +1,4 @@
 import React, { useEffect } from "react";
-import _ from "lodash";
 import Typography from "@mui/material/Typography";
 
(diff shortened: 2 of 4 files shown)


src/Terminal/ui/TerminalActionTimer.tsx
```diff
@@ -0,0 +1,26 @@
...
+export function TerminalActionTimer(): React.ReactElement {
+  const rerender = useRerender();
+
+  useEffect(() => {
+    const debounced = _.debounce(rerender, 25, { maxWait: 50 });
+    const unsubscribe = TerminalProcessEvents.subscribe(debounced);
+    return () => {
+      debounced.cancel();
+      unsubscribe();
+    };
```
Snarling, line 18:
```suggestion
import Typography from "@mui/material/Typography";

import { useRerender } from "../../ui/React/hooks";
import { Terminal } from "../../Terminal";
import { TerminalProcessEvents } from "../TerminalEvents";

export function TerminalActionTimer(): React.ReactElement {
  const rerender = useRerender();

  useEffect(() => {
    const unsubscribe = TerminalProcessEvents.subscribe(rerender);
    return unsubscribe;
```

Snarling:
```suggestion
import Typography from "@mui/material/Typography";

import { useRerender } from "../../ui/React/hooks";
import { Terminal } from "../../Terminal";
import { TerminalProcessEvents } from "../TerminalEvents";

export function TerminalActionTimer(): React.ReactElement {
  const rerender = useRerender();

  useEffect(() => {
    const unsubscribe = TerminalProcessEvents.subscribe(rerender);
    return unsubscribe;
```
Debounce is not needed here, it's needed for other terminal stuff because they could happen at arbitrarily fast rates (because the trigger is controlled by the player), but here we're only rerendering every time the terminal process function is ran, which should be a fixed time of approximately once every 200ms.

src/Terminal/Terminal.ts
```diff
@@ -102,7 +102,7 @@ export class Terminal {
     if (this.action === null) return;
     this.action.timeLeft -= (CONSTANTS.MilliPerCycle * cycles) / 1000;
     if (this.action.timeLeft < 0.01) this.finishAction(false);
-    TerminalEvents.emit();
+    TerminalProcessEvents.emit();
```
Snarling:
```suggestion
    else TerminalProcessEvents.emit();
```
Only need to send a process event if there is still an action underway. Without the debounce (see other suggestion), sending the event anyway leads to the action timer trying to rerender when there's no action.

Snarling (changes requested):
See suggested changes for removing the debounce from the timer. Another option would be to just remove the TerminalProcessEvents emitter entirely, and just do a useRerender(200) to handle rerendering the action bar. With that the action bar would need to check if the action was null and return `<></>` if so, since the rerender might hit during a time window where the action is null but the terminal hasn't rerendered yet (due to its own debounce) to remove the timer.

src/Terminal/ui/TerminalActionTimer.tsx
bezrodnov: Good catch! 👍
I did not even check how often this event is emitted when I created the PR.

It's indeed called approximately every 200ms. I've decided to use the alternative approach since it's cleaner (less code). And we don't need to check if the action is null there because in this case component is already unmounted.

bezrodnov: > And we don't need to check if the action is null there because in this case component is already unmounted.

we actually do need to do that 😅 - after some testing (with reduced rerender timeouts) I got an error because of that.
wip...

bezrodnov: @Snarling could you please review the latest version?

Snarling: https://github.com/bitburner-official/bitburner-src/assets/84951833/31acac28-5b83-48eb-932d-655349e343b4

Not sure if this can happen if the main process loop isn't delayed (e.g. by a script keeping the main process busy), but I was able to get this error by running a script that causes delays in the main engine loop and then just backdooring n00dles repeatedly. This will happen any time that the action has ended (so there's no action to make a bar out of) but the actual terminal has not rerendered yet because even though it got the event trigger from the action finishing, it's a debounced event so it won't happen right away.

Snarling: Just saw your update, that should be fine to avoid the issue.

Snarling (approved):

src/Terminal/ui/TerminalActionTimer.tsx
bezrodnov: yeah, I just made a wrong assumption that setting the action to null will unmount the progress bar before its rerender is called. The main reason here is that action is not a react state - so changing the action value does not trigger any re-renders directly.

Thanks for the review and valuable advice!

PR #546, opened 2023-05-29
UI: add close modal button (#497)

### LINKED ISSUE #464: improve accessibility for screen readers like nvda.
version of bitburner:
latest from website, 2.2.2d3f9554a
stepps to repro.
1.
start a game.
2.
save your game.
3.
go to options.
turn off aski.
quit your game.
4.
re launch your game.

when using nvda, I get stuck on the message below.

Offline for 12 hours 40 minutes 43 seconds. While you were offline:
• 
Your scripts generated $0.000
• 
Your Hacknet Nodes generated $136.929k
• 
You gained 0.000 reputation divided amongst your factions
 
pressing the buttons like hacking, city does nuthing.
still stuck on the message telling me how long i've been offline for.
to best debug this I would use nvda and try the stepps.
the only way i've found to get things semmi working is to clear my cookie but it cleares progress.
I am using the latest version of google chrome.

I would be greatful for any help.
Majid

### LINKED ISSUE #497: Dialogs could use a cancel button or "X"
At first I thought that there was no way to cancel certain actions.
In terms of usability it might be preferable to also have explicit buttons to close dialogs.

v.2.2.2 (d3f9554)

bezrodnov:
UI: add close modal button (#497)

closes #497
Edit: partially addresses #464 - needs some testing on real devices.

image

### DIFF
--- a/src/ui/React/Modal.tsx
+++ b/src/ui/React/Modal.tsx
@@ -1,11 +1,13 @@
+import React from "react";
 import { Theme } from "@mui/material";
 import Box from "@mui/material/Box";
+import IconButton from "@mui/material/IconButton";
 import Fade from "@mui/material/Fade";
 import M from "@mui/material/Modal";
 import createStyles from "@mui/styles/createStyles";
 import makeStyles from "@mui/styles/makeStyles";
 import { SxProps } from "@mui/system";
-import React from "react";
+import CloseIcon from "@mui/icons-material/Close";
 
 const useStyles = makeStyles((theme: Theme) =>
   createStyles({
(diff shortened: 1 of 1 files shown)


Snarling:
For #464 to get real accessibility, we should move towards using an actual [`dialog` element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog#accessibility_considerations) for modals. A close button is nice, but for instance it doesn't automatically bind the escape key to exiting the modal, whereas the native modal comes with that behavior.

But this is good. I just edited the description so this doesn't auto-close 464 since there's still more to be done on that.

PR #547, opened 2023-05-29
MISC: use prettier sort imports plugin

bezrodnov:
MISC: use prettier sort imports plugin

Just a QOL change to keep imports consistent across the project.


bezrodnov:
oh, as per madge there are 285 circular dependency issues...

I'll try to fix those and add an additional check to pipeline to avoid this kind of issues in future (of cause only if authors like this idea ^^ )

Snarling:
Re: import sorting I don't have a strong opinion on it. In many cases when we have imports grouped into sections, it is based on function, but in my mind it doesn't really matter what the order is.

If running prettier after installing the plugin is all you did, I'm not sure I can agree with using the plugin since it seems to have caused a test failure.

It also looks like sorting imports is a possible eslint rule, if so it would make more sense to just enable that rule instead of installing a new plugin.

Re: circular dependencies, these can sometimes cause issues, but they are not issues in and of themselves if the code is structured in a way that we don't end up with e.g. use before defined errors as a result. If the code in various files actually forms a dependency loop because it actually depends on each other, then that's not really a problem as long as the loop is able to be resolved before its contents are used.

There are probably some cases where we have circular imports just due to importing an actual class where we are only using that class for type information. In those cases we should be changing the `import` to an `import type`.

bezrodnov:
Hi @Snarling :)

As far as I know, ESLint does not support sorting imports out of the box (at least it's not capable of grouping different kinds of imports).
However, there are a few plugins that can do that and can be used as an alternative to this prettier plugin, e.g. [eslint-plugin-import](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/order.md).

I agree that sorting imports is a personal thing and if you don't see a value in it - that's fine, let's close the PR?

As for circular dependencies - as per my experience this is a very annoying thing. Sometimes it leads to hard-to-detect bugs which may leak into production after all the checks and tests (especially when test coverage is not so good).
Fixing this kind of issue can be time-consuming though. But when they are fixed - there's one less reason to worry and expect bugs from ^^

> There are probably some cases where we have circular imports just due to importing an actual class where we are only using that class for type information.
True, I've found quite a lot of those.
My personal preference on this is to keep types in independent files close to where they are used. E.g. `Player.types.ts` close to `Player.ts` and `SomeUIComponent.types.ts` close to `SomeUIComponent.tsx`.

P.S. if you are interested in this kind of topic (tools, utils, testing, tracing) - we can have a chat somewhere (I'm available on almost all platforms).
BTW, Great game! I enjoyed it :)

Snarling:
> sorting being unavailable with eslint

This makes it seem like it's available out of the box with eslint:
https://eslint.org/docs/latest/rules/sort-imports

> I agree that sorting imports is a personal thing and if you don't see a value in it - that's fine, let's close the PR?

It's not a one man show here, I'm not against the idea either and I'm sure there are others who appreciate the cleanliness/consistency, especially if it's basically free. But seeing the prettier and test fails does put me off on the idea a bit. But if whatever caused those to occur can be rectified I have no problem including something like this.

> Circular imports

Yeah we've actually had our share of issues with these in the past actually, mostly due to things in a dependency circle using top level code to reference content from other files in the circle. Historically our fix for that has been to move the offending top level code into an initializer function and just make sure that function runs before the point where we expect content to be initialized. But if there's a better way I'm open to it.

I think it makes sense to eliminate these as we can/in the easy cases. But like you said some of them may require large scale rewrites to eliminate the circular dependencies and I'm concerned out won't really be worth the effort in those cases.

bezrodnov:
> This makes it seem like it's available out of the box with eslint:
https://eslint.org/docs/latest/rules/sort-imports

As far as I understand It only sorts imports within a group or inside the import statement, but does not allow defining groups and custom sort order. E.g. it cannot be configured so that `react` import always goes on top, all other third-party modules go after react, then go all imports using aliases, and finally relative imports.

So I'll check what it takes to fix those circular dependency issues first cause changing import order might have broken some parts of the application.
For now, I'm closing this PR

d0sboots:
Re: Circular imports

Our codebase is a hell of circular imports. Not just cosmetic ones, but actual live functional ones. I've been working on cutting some of them as I can, but I am *not at all* surprised that sorting imports killed some tests.

PR #551, opened 2023-05-30
CORP: Add maxProducts property to division, consolidate maxProducts logic

kateract:
This PR add  a maxProducts property on the Division and also consolidates the logic behind determining the maximum number of products a division can support.


src/Corporation/Division.ts
```diff
@@ -35,6 +35,9 @@ export class Division {
 
   products = new JSONMap<string, Product>();
   makesProducts = false;
+  get maxProducts() {
```
d0sboots, line 38: What is the purpose of this getter? IMO, `getMaximumNumberProducts` is a fine name, there's no reason to have two names for the same thing.

kateract: I added the property because I wanted the exposed property in netscript to match on the object that it was representing.  Since the exposed division object has properties, but no methods, I implemented it this way.  

The logic could be moved to the getter, or the netscript object could expose the function instead of a property, but to limit the potential for breaking I left the method there and call it from the getter.

d0sboots: I kind of like it being a method on the internal object, it makes it clear that it's a computed property as opposed to one that is stored.

I can see the reasoning for wanting the netscript object to use the same name, but in that case just change the name of the existing method instead of adding a new one. You can keep it as a getter and use property access notation if you think it's really a win in clarity.

src/Corporation/Actions.ts
```diff
@@ -396,14 +396,8 @@ export function MakeProduct(
...
-  let maxProducts = 3;
-  if (division.hasResearch("uPgrade: Capacity.II")) {
-    maxProducts = 5;
-  } else if (division.hasResearch("uPgrade: Capacity.I")) {
-    maxProducts = 4;
-  }
-  if (division.products.size >= maxProducts) {
-    throw new Error(`You are already at the max products (${maxProducts}) for division: ${division.name}!`);
+  if (division.products.size >= division.maxProducts) {
+    throw new Error(`You are already at the max products (${division.maxProducts}) for division: ${division.name}!`);
```
Snarling, line 400: Not a big deal here as this is an error path and doesn't really need to be optimized, but this is calling the same function twice in a row to calculate max products (through multiple references to the getter variable).

I also like how nice getters can look, but in general they make it harder to notice inefficiencies like this, where if it was regular function call syntax, it would be clear here that this code path is calling getMaxProducts() or whatever twice in a row, which might normally prompt us to make a variable so we can reuse the value. Just mentioning it as this is a reason generally why we shouldn't broadly expand use of getters, but no need to revert the use here.

Issue #554, opened 2023-05-30
Script editor tabs don't look good with many scripts open


codernator:
When opening 7 or more tabs in Nano, the last tab will be cut off, such that the functions (refresh and close) are obscured or hidden. How much a tab is cut off seems to depend on file name length.

# Steps to reproduce
1) Open 6 scripts in Nano.  The tabs will look good.
2) Open a 7th script in Nano. The tabs will be resized and the text adjusted. The last tab will cut off.

While I have regularly observed this, I'm not entirely sure if it always the 7th tab that breaks. Of particular note, I had the window in Maximized Window mode (using Windows 11) with a 2560x1440 resolution, so there was plenty of room for more tabs. I've also observed it on my 1080p monitor with Maximized Window mode.  If the window is not maximized, then any tab could be cut off depending on the size of the window.

# Version
Bitburner v2.3.0 (43a3515ed) - and at least a few versions before.
on Windows 11 and Windows 10, if that is relevant.

bezrodnov:
One of the approaches could be to remove the max width calculation for script tabs (total tab container width divided by tab count) and use an invisible horizontal scroll (which is already there but not used xD). If this approach was used we could then add some scroll left/right buttons when needed in the beginning and end of the tab container (fixed or even on hover).

Also, mouse wheel support could be added, but as per my experience it's not that easy - that's only possible to achieve with JS and different browsers count wheel ticks differently. However, it seems to work fine in VS Code which is an electron app as well.
So maybe repeating VS Code approach would be the best solution here unless keeping chrome-like shortened-when-necessary tabs is a desired behavior.
wdyt?

bezrodnov:
One of the scroll approaches could be to place an active tab in the middle of the screen - refer to an example here (right above the States section): https://m2.material.io/components/tabs#scrollable-tabs

https://kstatic.googleusercontent.com/files/b5c47dfc7ed04a063dc68b1c9f6e4ffb8136ba62e0f1495cdeff833a69f0d07a446ec883079d167735d0f1743da0f9b8aa57da919415a9807a28a1dc252c72ff

VS Code approach is just to make sure that whichever tab is clicked/activated is fully visible on the screen... but the next tab is not, so without scrolling or closing previous tabs you can't really view the next ones

bezrodnov:
a bit more complex solution would be to support both current behavior and the one with scrolling and let user decide which one they like more (via game settings)
[...]

PR #560, opened 2023-06-02
REFACTORING: ScriptEditor

bezrodnov:
REFACTORING: ScriptEditor

### DIFF
--- a/src/ScriptEditor/ui/Editor.tsx
+++ b/src/ScriptEditor/ui/Editor.tsx
@@ -1,10 +1,9 @@
 import * as monaco from "monaco-editor";
 import * as React from "react";
 import { useEffect, useRef } from "react";
+import { useScriptEditorContext } from "./ScriptEditorContext";
 
 interface EditorProps {
-  /** Editor options */
-  options: monaco.editor.IEditorOptions;
   /** Function to be ran prior to mounting editor */
   beforeMount: () => void;
   /** Function to be ran after mounting editor */
@@ -13,10 +12,12 @@ interface EditorProps {
   onChange: (newCode?: string) => void;
 }
 
-export function Editor({ options, beforeMount, onMount, onChange }: EditorProps) {
+export function Editor({ beforeMount, onMount, onChange }: EditorProps) {
   const containerDiv = useRef<HTMLDivElement | null>(null);
-  const editor = useRef<monaco.editor.IStandaloneCodeEditor | null>(null);
+  const editorRef = useRef<monaco.editor.IStandaloneCodeEditor | null>(null);
   const subscription = useRef<monaco.IDisposable | null>(null);
+
+  const { options } = useScriptEditorContext();
 
   useEffect(() => {
     if (!containerDiv.current) return;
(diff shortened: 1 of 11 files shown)


bezrodnov:
Hi guys :)

I've spent a couple of hours tonight to refactor the ScriptEditorRoot.tsx and stuff around it a bit.
I'm not yet fully satisfied with the results, but it's already looking better imho.
I've quickly tested the game after these changes locally (start:dev, in browser) and haven't found obvious bugs so far.

Why am I doing this - I wanted to clean up this part of the app a bit before actually fixing the issue with tabs #554.

I hope my efforts won't get lost and that you will help me with the PR review and testing

d0sboots:
I looked it over at a high level; seems like a good change to me. ScriptEditorRoot was definitely getting too large.

I probably would have passed a few more props around rather than making a full-blown context, especially since this component hierarchy is pretty shallow, but I won't say it was the wrong call. There are definitely other places where a context would help matters.

Snarling will probably have other thoughts.

Snarling:
I also haven't done any testing yet, but just from an organizational standpoint I don't really have any issues with this. I agree that the ScriptEditorRoot was way too big and cluttered before, so the reorganization is warranted.

One gripe I do have just from the initial scan of the changes is I am not a fan of that useBoolean hook. setValue(false) and setValue(true) are already very clear / not confusing, and when someone unfamiliar with this custom hook sees it in use with it's slightly odd declaration usage, I believe it will cause more confusion than it solves.

bezrodnov:
@Snarling `useBoolean` hook is pretty straightforward and is very easy to get used to.
I agree that it looks very simple and basic. The main benefit it provides is that  `on`, `off` and `toggle` functions are wrapped in `useCallback` which is very convenient when you want to use them in `useEffect`, `useMemo` or in another `useCallback`.

It also looks nice in some situations, e.g.
```tsx
const MyComp = () => {
  const [isDialogOpen, { on: openDialog, off: closeDialog }] = useBoolean(false);
  ...
  return (
    <>
      ...
     <MyButton onClick={openDialog}>Open Dialog</MyButton>
     ...
     <MyDialog open={isDialogOpen} onClose={closeDialog} />
    </>
  )
```

Another point: one of the relevant "rules" I use in react is - wrap in `useCallback` functions passed down to child elements. This way child elements have the option to avoid unnecessary re-renders, establish effects, optimize performance with useMemo etc. It does not have to be used always, but I almost always do this in react contexts cause these props are accessible to every descendant. And this is where `useBoolean` hook comes in handy.

The first time I saw this hook I was also a bit skeptical (it was added by my colleague). But after giving it a try in a few projects at work I personally found it useful. And other colleagues were also happy to use it.

If you are still not convinced, I can easily remove it. In this case I'll add `useCallback`s right inside the context which will make it a bit less readable.

Snarling:
`React guarantees that setState function identity is stable and won’t change on re-renders. ` https://legacy.reactjs.org/docs/hooks-reference.html#usestate

~~As far as I can tell there is no reason to useCallback on a setState function. useCallback would be for ensuring that a function not created by a react hook is also stable and does not change between rerenders (or only changes based on a dependency list). There are some places in our code where it would be useful, but not for wrapping a setState function.~~

Edit: Actually I can see what you mean, in many cases we are not passing the setState function directly, but instead passing a new anonymous function that sets a particular state, and that would not remain stable. Feel free to keep it in.

Snarling:
Seems good, and I was not able to get anything to break in testing.

PR #562, opened 2023-06-03
CORPORATION: parse material Sell field once

Caldwell-74:
Material Sell Field was parsed twice and one time wrong when MAX was used
reduced it to one parse fixing the bug

when MAX was used it would replace it with the max possible based on the current price not the max currently present in the warehouse resulting in overselling when the input was MAX-10 or any other Amount based on and lower as MAX

### DIFF
--- a/src/Corporation/Division.ts
+++ b/src/Corporation/Division.ts
@@ -578,25 +578,7 @@ export class Division {
               corporation.getSalesMult() *
               advertisingFactor *
               this.getSalesMultiplier();
-            if (isString(mat.desiredSellAmount)) {
-              //Dynamically evaluated
-              let tmp = mat.desiredSellAmount.replace(/MAX/g, (mat.maxSellPerCycle + "").toUpperCase());
-              tmp = tmp.replace(/PROD/g, mat.productionAmount + "");
-
-              try {
-                sellAmt = eval(tmp);
-              } catch (e) {
-                dialogBoxCreate(
-                  `Error evaluating your sell amount for material ${mat.name} in ${this.name}'s ${city} office. The sell amount is being set to zero, sellAmt is set to ${sellAmt}`,
-                );
-                sellAmt = 0;
-              }
-              sellAmt = Math.min(mat.maxSellPerCycle, sellAmt);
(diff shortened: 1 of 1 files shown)


src/Corporation/Division.ts
```diff
@@ -836,7 +818,7 @@ export class Division {
           const desiredSellAmount = product.cityData[city].desiredSellAmount;
           if (isString(desiredSellAmount)) {
             //Sell amount is dynamically evaluated
-            let tmp: number | string = desiredSellAmount.replace(/MAX/g, (adjustedQty + "").toUpperCase());
+            let tmp: number | string = desiredSellAmount.replace(/MAX/g, adjustedQty.toString());
```
d0sboots, line 821: You changed this for products, but not for materials.

d0sboots (commented):
I agree that this is fixing a bug, but I don't understand your summary. AFAICT, there is no way that "MAX-10" would lead to *over*selling. Instead, it would have subtracted 10 from qty, giving a desiredSellAmt, and subtracted 10 from that again, leading to *under*selling.

src/Corporation/Division.ts
Caldwell-74: yeah i wasnt sure about that one didnt know if there was a functional difference between +"").toUpperCase and .toString()

d0sboots: Oh there definitely is, but I'm pretty sure it's only things we don't want.

Caldwell-74:
let me give you a small example
we have 100 of material X we want to sell MAX - 10 with a price that would allow us to sell at most 200
what we expect to happen is that it sells Math.min(stock, max possible for current price , sell amount)
in this example Math.min(100, 200, 100-10) = 90
what happens at the moment is that it uses that 200 as MAX and calculates
 Math.min(100, 200, 200-10) = 100
 selling 10 more than we want or ignoring the -10 


edit.: it would have led to underselling when the price is too high for selling MAX in the first place but thats alot harder to spot then selling all when it definitely shouldnt sell MAX/all

d0sboots:
I see, because maxSellPerCycle is never limited to anything other than the sell price. Very tricky.

PR #563, opened 2023-06-03
BUGFIX: script tab reordering

### LINKED ISSUE #553: Nano UI Glitch: Tab Movement
When rearranging tabs, the tab being dragged will appear to bounce back to where it came from.  However, as soon as you click any tab, the UI will be updated with the expected order.

# Steps to reproduce

1) Open two scripts in nano
2) Click and drag the tab from one script to reverse the visual order of the scripts.
3) On release, the tab you were dragging will appear to bounce back to where it came from.
4) Click any tab to observe an update. It seems any UI event will cause the expected update, such as using ALT-E to switch to nano editor.

# Version 
Bitburner v2.3.0 (43a3515ed) - and at least a few versions before.
on Windows 11 and Windows 10, if that is relevant.

bezrodnov:
BUGFIX: script tab reordering

closes #553 

Tested locally

### DIFF
--- a/src/ScriptEditor/ui/Editor.tsx
+++ b/src/ScriptEditor/ui/Editor.tsx
@@ -1,10 +1,9 @@
 import * as monaco from "monaco-editor";
 import * as React from "react";
 import { useEffect, useRef } from "react";
+import { useScriptEditorContext } from "./ScriptEditorContext";
 
 interface EditorProps {
-  /** Editor options */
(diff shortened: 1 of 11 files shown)


bezrodnov:
should be reviewed and merged after https://github.com/bitburner-official/bitburner-src/pull/560

Snarling:
Looks good, thanks!

I did have to do a rebase, normally I would recommend just including a fix like this in the existing PR

PR #564, opened 2023-06-03
CORPORATION: fixed Sell Buttons

Caldwell-74:
Product Sell Button displayed wrong price when MP was used
Material Sell Button displayed only the second character when a string was used as amount (MAX => A)

actual price used in division in SALE state is now cached to be reused in the UI (has been like this for TA2 now for all prices)

### DIFF
--- a/src/Corporation/Division.ts
+++ b/src/Corporation/Division.ts
@@ -552,7 +552,7 @@ export class Division {
             } else {
               sCost = mat.desiredSellPrice;
             }
-            mat.GUImarketPrice = sCost;
+            mat.uiMarketPrice = sCost;
 
             // Calculate how much of the material sells (per second)
             let markup = 1;
@@ -901,7 +901,7 @@ export class Division {
           } else {
             sCost = sellPrice;
           }
-          product.GUImarketPrice[city] = sCost;
+          product.uiMarketPrice[city] = sCost;
           let markup = 1;
           if (sCost > product.productionCost) {
             if (sCost - product.productionCost > markupLimit) {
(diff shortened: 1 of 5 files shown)


src/Corporation/Division.ts
```diff
@@ -553,6 +552,7 @@ export class Division {
             } else {
               sCost = mat.desiredSellPrice;
             }
+            mat.GUImarketPrice = sCost;
```
d0sboots: Treat acronyms like regular words for camel-case purposes. So this should be `guiMarketPrice`, or maybe just `uiMarketPrice`.

src/Corporation/Division.ts
```diff
@@ -902,7 +901,7 @@ export class Division {
           } else {
             sCost = sellPrice;
           }
-
+          product.GUImarketPrice[city] = sCost;
```
d0sboots: Same

src/Corporation/ui/MaterialElem.tsx
```diff
@@ -68,7 +67,7 @@ export function MaterialElem(props: IMaterialProps): React.ReactElement {
     if (isString(mat.desiredSellAmount)) {
       sellButtonText = (
         <>
-          Sell ({formatBigNumber(mat.actualSellAmount)}/{mat.desiredSellAmount[1]})
+          Sell ({formatBigNumber(mat.actualSellAmount)}/{mat.desiredSellAmount})
```
d0sboots, line 70: Wtf, was this just showing undefined?

d0sboots (approved):

src/Corporation/ui/MaterialElem.tsx
Caldwell-74: no it was showing the second character of the string
eg when input was MAX it shows A , PROD showed R 

(that code block handles the case of mat.desiredSellAmount beeing a string)

d0sboots: Oh of course. Still bad though

PR #565, opened 2023-06-03
GANG: Fix respectGainRate being stored incorrectly

Snarling:
* Removed default cycles of 1 from functions that are only called every 10 (normal) or 25 (bonus time) cycles.
* Added some more constants to represent number of cycles processed at once normally and in bonus time.
* Renamed some variables inside processGains to make it clearer what is per-cycle and what is total for the entire process
* Fix bug with gang.respectGainRate being stored as per-process even though it is documented as being per-cycle.

### DIFF
--- a/src/Gang/Gang.ts
+++ b/src/Gang/Gang.ts
@@ -33,8 +33,11 @@ export class Gang {
 
   isHackingGang: boolean;
 
+  /** Respect gain rate, per cycle */
   respectGainRate: number;
+  /** Wanted level gain rate, per cycle */
   wantedGainRate: number;
+  /** Money gain rate, per cycle */
   moneyGainRate: number;
 
   storedCycles: number;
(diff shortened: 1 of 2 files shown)


src/Gang/Gang.ts
```diff
@@ -104,50 +103,56 @@ export class Gang {
...
+      if (member.getTask().baseWanted < 0) justice++;
     }
-    this.respectGainRate = respectGains;
-    this.wantedGainRate = wantedLevelGains;
-    this.moneyGainRate = moneyGains;
-    const gain = respectGains;
-    this.respect += gain;
+
+    this.respectGainRate = respectGainsTotal / numCycles;
+    this.wantedGainRate = wantedLevelGainPerCycle;
```
d0sboots, line 123: Why aren't these adjusted?

d0sboots (commented):
👍

src/Gang/Gang.ts
bezrodnov: they are already "per cycle" and multiplied by `numCycles` later :)
I also was confused when I saw it first time as it's an eye-catching piece of code

PR #566, opened 2023-06-03
PRESTIGE: Make bitnode reset synchronous again

Snarling:
Fixes a bug where singularity callback scripts could be launched before the next engine cycle and therefore before the node change.

* Bitnode transfers are again performed without delay, reverts a change from #397 and solves the original issue a different way (delayed dialogs instead of delaying the entire node transition)
* Added 500ms timeout to the bitnode reset singularity functions (for consistency with the aug install functions)
* Not super related, but while I was in some functions I renamed instances of a `flume` variable to `isFlume` so it's more obvious that it's a boolean representing whether the reset was due to bitflume


d0sboots:
👍

PR #567, opened 2023-06-03
UI: script tabs management (#554)

### LINKED ISSUE #554: Script editor tabs don't look good with many scripts open

bezrodnov:
* removes script tab text and width calculation/limits
* enables vertical scroll with mouse wheel support (without shift).
* sticks the search field to the left so that it's not scrolled
* scrolls the current script tab into view
* fix an issue with activating, closing, reordering or updating a script tab when the script list is filtered

closes #554

### DIFF
--- a/src/ScriptEditor/ui/Tabs.tsx
+++ b/src/ScriptEditor/ui/Tabs.tsx
@@ -10,7 +10,7 @@ import Tooltip from "@mui/material/Tooltip";
 import CloseIcon from "@mui/icons-material/Close";
 import SearchIcon from "@mui/icons-material/Search";
 
-import { useRerender } from "../../ui/React/hooks";
+import { useBoolean, useRerender } from "../../ui/React/hooks";
 import { Settings } from "../../Settings/Settings";
 
 import { dirty, reorder } from "./utils";
@@ -18,7 +18,7 @@ import { OpenScript } from "./OpenScript";
 import { Tab } from "./Tab";
 
 const tabsMaxWidth = 1640;
-const searchWidth = 150;
+const searchWidth = 180;
 
 interface IProps {
   scripts: OpenScript[];
@@ -31,6 +31,7 @@ interface IProps {
 
 export function Tabs({ scripts, currentScript, onTabClick, onTabClose, onTabUpdate }: IProps) {
   const [filter, setFilter] = useState("");
+  const [isSearchTooltipOpen, { on: openSearchTooltip, off: closeSearchTooltip }] = useBoolean(false);
   const [searchExpanded, setSearchExpanded] = useState(false);
   const rerender = useRerender();
 
@@ -53,9 +54,10 @@ export function Tabs({ scripts, currentScript, onTabClick, onTabClose, onTabUpda
     setFilter(event.target.value);
   }
 
-  function handleExpandSearch(): void {
+  function toggleSearch(): void {
     setFilter("");
     setSearchExpanded(!searchExpanded);
+    closeSearchTooltip();
   }
 
   function handleScroll(e: React.WheelEvent<HTMLDivElement>): void {
@@ -64,7 +66,12 @@ export function Tabs({ scripts, currentScript, onTabClick, onTabClose, onTabUpda
 
   return (
     <Box display="flex" flexGrow="0" flexDirection="row" alignItems="center">
-      <Tooltip title={"Search Open Scripts"}>
+      <Tooltip
+        title={"Search Open Scripts"}
+        open={isSearchTooltipOpen}
+        onOpen={openSearchTooltip}
+        onClose={closeSearchTooltip}
+      >
         <span style={{ marginRight: 5 }}>
           {searchExpanded ? (
             <TextField
@@ -76,14 +83,14 @@ export function Tabs({ scripts, currentScript, onTabClick, onTabClose, onTabUpda
                 startAdornment: <SearchIcon />,
                 spellCheck: false,
                 endAdornment: (
-                  <IconButton onClick={handleExpandSearch}>
+                  <IconButton onClick={toggleSearch}>
                     <CloseIcon />
                   </IconButton>
                 ),
               }}
             />
           ) : (
-            <Button onClick={handleExpandSearch}>
+            <Button onClick={toggleSearch}>
               <SearchIcon />
             </Button>
           )}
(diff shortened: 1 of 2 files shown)


bezrodnov:
https://github.com/bitburner-official/bitburner-src/assets/5138959/f646907b-15b3-4a6b-ace8-0e86a8fea001

Snarling:
The changes are good. While testing, I noticed some bugs, but they mostly they appear to be existing bugs that precede this PR.

1. Console warnings when opening and closing the search input. These don't happen on live branch. I'm not sure if there's a real issue at play here or if it can be safely ignored, but I would rather avoid whatever is causing the warning if practical.
2. Closing the search input with the tooltip displayed leads to a persistent tooltip. This is probably due to the tooltip having two different child elements (either the button or the text input) instead of one persistent element. This happens on live as well, so doesn't need to be addressed on this PR, just mentioning it.
3. When the search is expanded, clicking a tab doesn't select the right one. e.g. clicking the 2nd tab in a filtered search of tabs will open the 2nd overall script, not taking into account the filter. I got the same behavior on live though.

See video:

https://github.com/bitburner-official/bitburner-src/assets/84951833/47636715-c953-486d-8ced-b554668f0025

bezrodnov:
I'll take a look and address them in the same PR

bezrodnov:
@Snarling issues above should be addressed. Could you please check?

src/ScriptEditor/ui/Tabs.tsx
```diff
@@ -1,20 +1,24 @@
 import React, { useState } from "react";
 import { DragDropContext, Droppable, Draggable } from "react-beautiful-dnd";
```
Snarling:
```suggestion
import { DragDropContext, Droppable, Draggable, DropResult } from "react-beautiful-dnd";
```
Import needed for suggested onDragEnd type fix

Snarling (commented):
I thought the extra span would fix the tooltip issue, but it unfortunately doesn't. It does clear up the anchorEl warning from mui.

I think it's just a problem with Mui's tooltip that we can't easily address ourselves. The event to clean up the tooltip does not appear to be firing if there are no further mouse events on the target element.

I wasn't able to find anything on MUI's issues list (open or closed) that was a complete match for this problem, but it's possible that updating to latest mui would fix the issue. I plan to replace our current usage of the mui legacy styles system with something that is more friendly for players who like to use custom content, and that should also allow us to update to latest mui (or at least make it easier).

Not caught on previous reviews, another bug that was already-existing before your refactor: drag and drop on filtered scripts was reordering the scripts using the filtered indexes but referencing the unfiltered array.

https://github.com/bitburner-official/bitburner-src/assets/84951833/ed13d2f3-09e3-4ff7-b983-b4812af7638a

While messing around with it, I got a rewrite for onDragEnd that fixes it (and also removes an `any` type), so I'm providing it here if you want to use or adapt it. It didn't let me suggest it directly because that section of the code is not touched by the PR yet.

```ts
  function onDragEnd(result: DropResult): void {
    // Dropped outside the list
    if (!result.destination) return;
    const sourceIndex = scripts.indexOf(filteredOpenScripts[result.source.index]);
    const destIndex = scripts.indexOf(filteredOpenScripts[result.destination.index]);
    if (sourceIndex < 0 || destIndex < 0) {
      console.error("Unexpected index on script editor tabs drag and drop");
      return;
    }
    reorder(scripts, sourceIndex, destIndex);
    rerender();
  }
```

src/ScriptEditor/ui/Tabs.tsx
bezrodnov: @Snarling I think it might look better this way:
```tsx
  const filteredScripts = Object.values(scripts)
    .map((script, originalIndex) => ({ script, originalIndex }))
    .filter(({ script }) => script.hostname.includes(filter) || script.path.includes(filter));

  function onDragEnd(result: DropResult): void {
    // Dropped outside of the list
    if (!result.destination) return;
    reorder(
      scripts,
      filteredScripts[result.source.index].originalIndex,
      filteredScripts[result.destination.index].originalIndex,
    );
    rerender();
  }
  ...
             {filteredScripts.map(({ script, originalIndex }, index) => {
                const { path: fileName, hostname } = script;
                const isActive = currentScript?.path === script.path && currentScript.hostname === script.hostname;

                const title = `${hostname}:~${fileName.startsWith("/") ? "" : "/"}${fileName} ${dirty(scripts, index)}`;

                return (
                  <Draggable
                    ...
                    index={index}
                  >
                    {(provided) => (
                      <Tab
                       ...
                        onClick={() => onTabClick(originalIndex)}
                        onClose={() => onTabClose(originalIndex)}
                        onUpdate={() => onTabUpdate(originalIndex)}
                      />
                    )}
                  </Draggable>
                );
              })}
```
wdyt?

bezrodnov: As for the tooltip issue - I'll take a look into it. Most probably it's not specific to script tabs and should be reproducible in other places with similar component behavior.

bezrodnov: https://github.com/mui/material-ui/issues/12299

bezrodnov: @Snarling should be fixed now, could you please check? :)

Snarling: Yeah your rewrite works a lot better. Tooltip workaround also seems to work, I couldn't trigger any unexpected behavior. Good stuff!

PR #570, opened 2023-06-04
MISC: optimize corp affordable upgrade level calculation

bezrodnov:
~~And consider affordable upgrade count for other batch purchase types than MAX (x1, x5, x10, x50, x100)~~


Snarling (commented):
Awesome, I was not happy with iteratively solving for the right number of upgrades since it's being constantly performed every rerender for every upgrade. However, the solution does have a few issues:

1. It doesn't handle zero or negative funds. Easily fixed by an early return 0 for these cases.
image. 

2. If less of the upgrade can be bought than the selected mult, the change to using max affordable means the player can't actually see how much their desired amount costs. The extreme case is where 0 of an upgrade can be bought, then the player doesn't even get a ballpark of what amount of money they will need in order to buy the upgrade.  
image

Snarling (commented):
Awesome, I was not happy with iteratively solving for the right number of upgrades since it's being constantly performed every rerender for every upgrade. However, the solution does have a few issues:

1. It doesn't handle zero or negative funds. Easily fixed by an early return 0 for these cases.
image. 

2. If less of the upgrade can be bought than the selected mult, the change to using max affordable means the player can't actually see how much their desired amount costs. The extreme case is where 0 of an upgrade can be bought, then the player doesn't even get a ballpark of what amount of money they will need in order to buy the upgrade.  
image

bezrodnov:
Zero funds should be fine according to the formula (`Math.log(1) = 0` which is then multiplied by some number). But negative - this is something I did not expect to happen :) Is it possible in the game, how did you reproduce this issue, by manually changing funds?

On the second point - yeah, it totally makes sense. However in this case this function does not need the `amount` argument cause it will always be `MAX`.

bezrodnov:
pushed a change to sanitize the value and to address the second point (so that behavior stays the same as before this PR)

bezrodnov:
I'll add some unit tests for this function

d0sboots:
You can get negative funds in a variety of ways. It's a standard (although not *common*) part of gameplay.

PR #578, opened 2023-06-06
CORP: Robotics industry NaN fix + better exports validation

Snarling:
1. Corrected the testing for player input on exports, now uses parentheses like the way the replacement is performed ingame.
2. Prevents exporting to the source city.
3. Prevents exporting to a city that does not have a warehouse for the target division. 
4. Provide better error messaging to player when an export fails.
    * Old error message, would display constantly during processing: 
image
    * New error message: Checks before setting export to prevent setting invalid amount: (outdated image, -IPROD is fine as an input)
image
5. Same validation is performed on gameload for saves older than 2.3.1 release.
    * Terminal error informs the player if any of their exports failed: 
image
    * Console error shows full details ("sellAmt" is just the source division's name): 
image
6. Can no longer have the same material get exported from one warehouse to another multiple times through duplicate exports objects.
7. Canceling an export no longer requires the exact stored amount for the export.

Separate from exports, also fixed a NaN bug that could affect produced materials in Robotics industry (or any other industry that uses AI Cores as a required material)

### DIFF
--- a/markdown/bitburner.basichgwoptions.additionalmsec.md
+++ b/markdown/bitburner.basichgwoptions.additionalmsec.md
@@ -4,7 +4,7 @@
 
 ## BasicHGWOptions.additionalMsec property
 
-Number of additional milliseconds that will be spent waiting between the start of the function and when it completes. Experimental in 2.2.2, may be removed in 2.3.
+Number of additional milliseconds that will be spent waiting between the start of the function and when it completes.
 
 **Signature:**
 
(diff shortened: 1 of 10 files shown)


d0sboots:
Wait, but the IPROD bug should be already fixed?

d0sboots:
I see. The issue is the the tests you're performing don't line up with how the expression is actually evaluated. (It's parenthesized when actually used).
We should do that generically instead of case-wise to be more safe.

Snarling:
Oh I guess I forgot that part was already addressed via a different fix, looks like I didn't backport it to 2.3 as a hotfix. I'll update the replacement to include parens (the updated validation is still needed to prevent exporting to same or nonexistent warehouse)

PR #583, opened 2023-06-07
CORP: Update exportMaterial amount to number | string

kateract:
`exportMaterial` in the typescript definitions was just a number, should be a string to match the interface (so you can use MAX, IPROD, etc)
Fix bug with exportMaterial amount MAX, looks like it was missed in the or condition when the last update for robotics was pushed.


Snarling:
Looks good. I thought you missed updating the actual exportMaterial function, but it was already accepting strings (or numbers, because helpers.string will convert numbers to string), just documented wrong. Good catch on the missing MAX

PR #586, opened 2023-06-08
MISC: tiny corp code style improvement

bezrodnov:
- extract IndustryDescription UI component from `data` directory to `ui`
- add `CorpSmartSupplyOption` type
- ~~clone corp constants only once~~
- other tiny improvements


src/Corporation/ui/IndustryDescription.tsx
```diff
@@ -0,0 +1,34 @@
+import React from "react";
+
+import Typography from "@mui/material/Typography";
+
+import { MoneyCost } from "./MoneyCost";
+import { Corporation } from "../Corporation";
+import { IndustryType } from "../data/Enums";
+import { IndustriesData } from "../data/IndustryData";
+
+interface IProps {
```
Snarling, line 10: Generally, I am not a fan of putting `I` in front of interface names, or making a type called props that doesn't say what it is the props for. Not a big deal especially with this being a file that only has one component in it, but IndustryDescriptionProps would be a better name IMO. Only mentioning it because there are other more important comments, something like this wouldn't be blocking for the PR to get merged.

src/NetscriptFunctions/Corporation.ts
```diff
@@ -677,13 +678,19 @@ export function NetscriptCorporation(): InternalAPI<NSCorporation> {
...
       // TODO: add functions for getting materialInfo and research info
+      const constants = cloneDeep(
+        omit(corpConstants, "fundingRoundShares", "fundingRoundMultiplier", "valuationLength"),
+      );
+
+      return () => {
+        checkAccess(ctx);
+        return constants;
+      };
     },
```
Snarling: If the player modifies that 1-time clone (e.g. they accidentally did an assignment instead of comparison in an if statement conditional), that alteration will persist every time they call getConstants which is not desirable.

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -7412,6 +7417,7 @@ interface CorpConstants {
   maxProductsBase: number;
   /** The minimum decay value for morale/energy */
   minEmployeeDecay: number;
+  smartSupplyUseOptions: CorpSmartSupplyOption[];
```
Snarling: Since this is being just now added to the player facing API, we can update the name to make a little more sense. I'm not sure why `Use` is in the title but it's a weird mismatch between the type name and the variable name. smartSupplyOptions makes more sense

src/NetscriptFunctions/Corporation.ts
```diff
@@ -415,7 +416,7 @@ export function NetscriptCorporation(): InternalAPI<NSCorporation> {
       const option = helpers.string(ctx, "option", _option);
       if (!hasUnlock(CorpUnlockName.SmartSupply))
         throw helpers.makeRuntimeErrorMsg(ctx, `You have not purchased the Smart Supply upgrade!`);
-      SetSmartSupplyOption(warehouse, material, option);
+      SetSmartSupplyOption(warehouse, material, option as CorpSmartSupplyOption);
```
Snarling: So here you're passing option to a function that requires a valid CorpSmartSupplyOption, but you're not doing any type validation on the input other than that it is a string. Either the function is meant to handle unvalidated strings (and should take in a string for that parameter) or it is meant to take in validated strings (and validation should be added before calling the function). Related to other comment

src/Corporation/Actions.ts
```diff
@@ -242,10 +245,14 @@ export function SetSmartSupply(warehouse: Warehouse, smartSupply: boolean): void
...
-  if (!corpConstants.smartSupplyUseOptions.includes(useOption)) {
-    throw new Error(`Invalid Smart Supply option '${useOption}'`);
+function assertSmartSupplyOption(value: string): asserts value is CorpSmartSupplyOption {
+  if (!corpConstants.smartSupplyUseOptions.includes(value as CorpSmartSupplyOption)) {
+    throw new Error(`Invalid Smart Supply option '${value}'`);
   }
+}
+
+export function SetSmartSupplyOption(warehouse: Warehouse, material: Material, useOption: CorpSmartSupplyOption): void {
+  assertSmartSupplyOption(useOption);
```
Snarling: If the function takes in a CorpSmartSupplyOption, we don't need to validate it again inside the function. It should either take in an unvalidated string (and do the validation here) or the validation should be done before calling this function.

It's better to do the validation before calling the function.

Snarling (changes requested):

src/Corporation/Actions.ts
```diff
@@ -251,7 +251,7 @@ function assertSmartSupplyOption(value: string): asserts value is CorpSmartSuppl
   }
 }
 
-export function SetSmartSupplyOption(warehouse: Warehouse, material: Material, useOption: string): void {
+export function SetSmartSupplyOption(warehouse: Warehouse, material: Material, useOption: CorpSmartSupplyOption): void {
   assertSmartSupplyOption(useOption);
```
Snarling: This also should either take in a string and perform validation itself, or it should take in a valid type and not perform its own validation.

Snarling: This also should either take in a string and do validation itself (worse option) or take in the correct type and expect validation to be performed before calling this function

src/Corporation/ui/IndustryDescription.tsx
bezrodnov: me too - I would never name it like that in my project. The only reason for this code style - is consistency in this project.

The way I prefer to implement things in react is:
1. have a directory named as your component (e.g. `IndustryDescription`)
2. add at least 3 files to this directory: `IndustryDescription.tsx`, `IndustryDescription.types.ts` and `index.ts` exporting everything from previous files (barreling). In case of MUI styles there would also be a `IndustryDescription.styles.ts` which exports `useStyles` hook.
3. `IndustryDescription.types.ts` exports a type (not interface) named `IndustryDescriptionProps`
4. `IndustryDescription.tsx` imports `IndustryDescriptionProps` from `IndustryDescription.types.ts` exports component named `IndustryDescription`:
```tsx
import { IndustryDescriptionProps } from './IndustryDescription.types.ts';
import { useStyles } from './IndustryDescription.styles.ts'

const IndustryDescription: React.VFC<IndustryDescriptionProps> = (props) => { // often props are destructured right away on this line
  const classNames = useStyles();
  //...
  return (
    <>something</>
  );
}
export { IndustryDescription } // export always in the end of file, never export default
```

However, applying my best practices in this project won't make sense if I do this only for a couple of places.
IMHO either I follow the current code style in the project or refactor everything (after agreement) and use a better way

src/ScriptEditor/NetscriptDefinitions.d.ts
bezrodnov: will it be ok to change? if it breaks someone's script (which I doubt) - will it still be fine?
this const was not mentioned in docs before, but it was still available.

src/NetscriptFunctions/Corporation.ts
bezrodnov: maybe we should punish players for this? xD

src/NetscriptFunctions/Corporation.ts
bezrodnov: This is something I considered as well. I had to choose between this function accepting any string as an argument in-game and having validation inside or accepting a strict set of strings but still having validation inside (just in case players call it with a wrong value). I found the second option more appealing

src/Corporation/Actions.ts
bezrodnov: We should still validate it because players who don't use TS can call it with whatever they want.

src/NetscriptFunctions/Corporation.ts
bezrodnov: sorry, the comment above was actually for `Actions.ts` file

bezrodnov: What we can do is - move assertion part from `Actions.ts` file here since the in-game code will not call `SetSmartSupplyOption` with an invalid option.
wdyt?

src/ScriptEditor/NetscriptDefinitions.d.ts
bezrodnov: one of the options is to add both `smartSupplyUseOptions` and `smartSupplyOptions` but mark the first one as deprecated.

Snarling: Oh, it looks like it was being actually provided in the object returned from getConstants, but just had not been added to the CorpConstants interface. It's a new addition from 2.3.0. It was never documented as being provided in the API until you just added documentation for it, I say go ahead and just change it.

bezrodnov: done 👍

src/NetscriptFunctions/Corporation.ts
Snarling: We validate player input using things like checkEnum, which also works for typed string arrays (not just enums). The other PR I'm working on will replace checkEnum with something similar that also handles making the error message, checkEnum just returns a boolean for whether the input was the right type or not.

You can search NetscriptFunctions/Corporation.ts for `checkEnum` and see what the normal usage is for type guarding player input with checkEnum. By the time we are calling an ingame function that has required types, we always should have already done the type validation.

bezrodnov: oh, I've already updated this solution with `assertMember`.

src/Corporation/Actions.ts
Snarling: Players don't call ingame functions. Players only call ns functions, which actually have all parameters as `unknown`. From there we do validation on their input to narrow the types to what is actually required.

bezrodnov: yeah, validation is already moved there :)

src/Corporation/ui/IndustryDescription.tsx
bezrodnov: btw `interface IProps` is used 201 times in this project 
lol

bezrodnov: I would love to restructure everything about react in the way I explained above. I would fee like home then xD

Snarling: Yeah before I started doing work on the game, these IProps or ITypeName were all over the place. There are a lot less IProps and similar now, but there are still some of them out there. I didn't go find and rename all of them because it's not a big deal and has no actual impact, usually I just change them when I come across them. Like I said it's not really an issue

bezrodnov: I could create a PR (or series of PR) to change this pattern if this is something that you also want (to me current pattern looks ugly as well)

src/NetscriptFunctions/Corporation.ts
bezrodnov: sorry, but I am a bit lost here - sometimes we say something like "let player try to hack the game and do weird stuff" but sometimes we try to protect them from something which does not even have impact on game (since this is a cloned value)

Snarling: It makes it so the game will never provide them with correct constants through getConstants ever again until they reload the page. We want our API to continue to provide the correct data to the player, no matter what they do with that data once they have it.

bezrodnov: @d0sboots, what do you think about it?
I am personally fine with any option you chose guys, you are maintainers anyway :)

d0sboots: I'm with Snarling here: Changing this creates a weird failure case, and requires more code, for a speedup that won't/shouldn't ever be observed.

bezrodnov: 👍

src/Corporation/ui/IndustryDescription.tsx
d0sboots: There's also the middle ground (that I think Snarling was suggesting): Fix these as you come across them, without doing a whole-project refactoring. Although if you have the energy for the latter, I guess go wild.

Snarling (approved):

src/Corporation/ui/IndustryDescription.tsx
bezrodnov: @d0sboots @Snarling 
which of the following styles do you prefer for react component props:

1. Inline props
```tsx
export const AchievementEntry: VFC<{
  achievement: Achievement;
  unlockedOn?: number;
  cssFiltersUnlocked: string;
  cssFiltersLocked: string;
}> = ({ achievement, unlockedOn, cssFiltersUnlocked, cssFiltersLocked }) => {
```
2. explicit props declaration in the same file
```tsx
type AchievementEntryProps = {
  achievement: Achievement;
  unlockedOn?: number;
  cssFiltersUnlocked: string;
  cssFiltersLocked: string;
};

export const AchievementEntry: VFC<AchievementEntryProps> = ({ achievement, unlockedOn, cssFiltersUnlocked, cssFiltersLocked }) => {...
```
3. Same as 2 but props are stored in a separate file, e.g. `AchievementEntry.types.ts`. In this case, each component will be nested into the same-name directory with props, component and an index file (for barreling)

And do you mind replacing a function declaration with an arrow function (like in the examples above)

d0sboots: I think the answer is none of those?

I was confused about what this `VFC` thing was that you were doing, since I hadn't seen it in our codebase before. Turns out, it's deprecated: <https://www.mydatahack.com/moving-away-from-fc-and-vfc/>

So, I think the right answer is like 2, but don't export the type (there's no need), and don't do the VFC thing. 1-like would also be fine. *Definitely* not 3.

bezrodnov: So the first one without VFC should be fine then?
I don't see much value in declaring a type, using it on the next line, and never seeing it anywhere else again.
However, there are cases when you would like to reference a type of one of the props inside a component (i.e. `MyComponent['onChange']`) - in those cases having an explicit type declaration could make sense (even though you could easily use something like `typeof onChange` instead).

bezrodnov: @d0sboots BTW, the link you shared references react 18 (in which functional components don't have `children` prop) - but this project is based on react 17, so it could still be an option

d0sboots: I took it to mean "`FC` and `VFC` are generally the way of the past, don't keep using them."

PR #596, opened 2023-06-11
ENUMS: Initial Enum Helper rework + Reorganization

Snarling:
This is a do-over of #435 which was broken during a rebase and not in a condition where it was easy to review or fix.

commit 53e31d7 : The really hard to review one with no functional changes.
* Moved enums that will need enum helpers to respective Enums files in their associated directories
* Changed the main Enums.ts to a barrel file for all enums
* Added an `@enums` alias for importing enums
* Singularized some plural enum names
* Changed LocationType from a numeric enum to a string enum. It's not involved in savegame so this should have no impact.
* Changed Locations/Cities initialization code, functionally identical

commit 86526d7
* Adds the enum helper

commit b896f8d
* Removes checkEnum and anything that relied on it.
* findEnumMember was not removed yet from the old helpers file, its functionality is not 1:1 preserved (it is extremely permissive in its matching). Still pondering whether to alter the behavior or replicate it.
* Also some additional enums needed to be added, but the way these were added resulted in a broken import loop.

commit 7303cb0, e8c52d8
* Fixes the broken import loop from the previous commit by defining corp enumlike data in the enums file, and just importing it into constants.
* Fixed a lint issue


Snarling:
Just gonna merge this now, further edits can be in a separate PR.

PR #598, opened 2023-06-11
MISC: i18n support (the beginning)

bezrodnov:
- added i18next and react-i18next
- added TS support for i18next
- extracted translation for LoadingScreen and IndustryDescription components, added RU translation for them just as an example


Snarling:
I didn't get a chance to mess around with this one, and now it wants a rebase due to conflicts with previous PRs (probably mine). I was ready to merge the other two so I just went ahead and rebased + merged, but since I'm not ready to merge this just yet I figure I should check with you whether you'd rather I just rebase it in cases like this, or leave it to you to rebase.

bezrodnov:
Rebased :)

hydroflame:
This is a really cool idea but we discussed several times in the past that translating is not really viable for a programming game.

There are several elements in the game and cannot or should not be translated and those elements makes translating everything else too tricky.

d0sboots:
Wait, why does leaving some things untranslated make everything else tricky? That just means the translation is the same across all the languages for those terms, which is really easy to accomplish (you just leave it unset, and it will always take the default value, in basically all translation frameworks.)

I don't disagree that it would be a ton of work, and hard to do at this stage, but I don't think there's anything about this being a programming game that makes it intrinsically more challenging.

PR #599, opened 2023-06-11
UI: show all skills in import save comparison tool

bezrodnov:
https://github.com/bitburner-official/bitburner-src/assets/5138959/76e8ddac-c0cb-49e7-ad29-3c147374763c

If the last commit is reverted, it will look like this:

image

### DIFF
--- a/src/ui/React/ImportSave/ImportSave.tsx
+++ b/src/ui/React/ImportSave/ImportSave.tsx
@@ -19,22 +19,22 @@ import makeStyles from "@mui/styles/makeStyles";
 import createStyles from "@mui/styles/createStyles";
 import { Theme } from "@mui/material/styles";
 
-import ThumbUpAlt from "@mui/icons-material/ThumbUpAlt";
-import ThumbDownAlt from "@mui/icons-material/ThumbDownAlt";
+import WarningIcon from "@mui/icons-material/Warning";
 import DirectionsRunIcon from "@mui/icons-material/DirectionsRun";
 import ArrowBackIcon from "@mui/icons-material/ArrowBack";
-import WarningIcon from "@mui/icons-material/Warning";
 
 import { Skills } from "@nsdefs";
 
-import { ImportData, saveObject } from "../../SaveObject";
-import { Settings } from "../../Settings/Settings";
-import { convertTimeMsToTimeElapsedString } from "../../utils/StringHelperFunctions";
-import { formatMoney, formatNumberNoSuffix } from "../formatNumber";
-import { ConfirmationModal } from "./ConfirmationModal";
-import { pushImportResult } from "../../Electron";
-import { Router } from "../GameRoot";
-import { Page } from "../Router";
+import { ImportData, saveObject } from "../../../SaveObject";
+import { Settings } from "../../../Settings/Settings";
+import { convertTimeMsToTimeElapsedString } from "../../../utils/StringHelperFunctions";
+import { formatMoney, formatNumberNoSuffix } from "../../formatNumber";
+import { ConfirmationModal } from "../ConfirmationModal";
+import { pushImportResult } from "../../../Electron";
+import { Router } from "../../GameRoot";
+import { Page } from "../../Router";
+
+import { ComparisonIcon } from "./ComparisonIcon";
 
 const useStyles = makeStyles((theme: Theme) =>
   createStyles({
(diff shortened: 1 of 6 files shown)


src/ui/React/ImportSave/ImportSave.tsx
```diff
@@ -1,38 +1,43 @@
...
+import IconButton from "@mui/material/IconButton";
+import Paper from "@mui/material/Paper";
+import Table from "@mui/material/Table";
+import TableHead from "@mui/material/TableHead";
+import TableRow from "@mui/material/TableRow";
+import TableBody from "@mui/material/TableBody";
+import TableContainer from "@mui/material/TableContainer";
+import TableCell from "@mui/material/TableCell";
+import Tooltip from "@mui/material/Tooltip";
+import Typography from "@mui/material/Typography";
```
Snarling, line 16: I think we decided there was no reason we were importing from individual files, and we can just import mui stuff all from the main mui/material. But it's not really a big deal so I'm not going to kick it back for that.

Snarling (commented):
I had to rebase for an unknown reason (there were no conflicts to resolve during rebase, even though it said there were complex conflicts in SaveObject.ts that couldn't be resolved in the web interface)

I also still think the file structure is a bit weird, but maybe it'll grow on me.

src/ui/React/ImportSave/ImportSave.tsx
bezrodnov: I actually could not remember what we decided xD
I've just checked again and d0sboots mentioned that our imports are mostly consistent, most of them are like in this PR (import from subdirectories).
I also prefer this kind of import for mui. In some cases, you also want to import some types (e.g. `SelectProps`) and it's better to have imports grouped in this case.

bezrodnov:
> I had to rebase for an unknown reason (there were no conflicts to resolve during rebase, even though it said there were complex conflicts in SaveObject.ts that couldn't be resolved in the web interface)
> 
> I also still think the file structure is a bit weird, but maybe it'll grow on me.

BTW, what's wrong with the file structure? To me it looks so much better - you have your main component in there and a bunch of child components. If you don't create a directory for those - it will be a huge pile of who-knows-what-they-do components at one directory level.

I would also remove `React` directory, it does not make any sense to me to have some components here and some outside.

d0sboots:
I'm also not a fan of extremely nested directories. If there actually *are* (or are going to be) a whole lot more files, that's different, but I'm also happy with keeping multiple components in a single file as opposed to breaking each one out into a separate file, as long as they're internal (non-exported) details of the main component, and that tends to reduce the overall number of files.

I think the index.ts is a bit redundant as well. Again, maybe not for a large directory, but this is not that right now.

d0sboots:
The flipside to that is that if we had things more finegrained per-file, we might actually be able to try and cut down on the circular dependency tangle...

bezrodnov:
An interesting read on this topic: https://steven-lemon182.medium.com/are-typescript-barrel-files-an-anti-pattern-72a713004250

bezrodnov:
>  I'm also happy with keeping multiple components in a single file as opposed to breaking each one out into a separate file, as long as they're internal (non-exported) details of the main component, and that tends to reduce the overall number of files.

This seems to be a very opinionated topic. I tried to find some good articles about it, but all I found are obvious recommendations like "if multiple components are exported - put them in separate files". As for putting child components in separate files or the same file (when they are not used anywhere else) - there's no clear answer.

To me it just feels harder to read the file when it has many components, I tend to collapse those child components first to find a "real boss" component I came for. I like the approach when every component file takes less than 100 lines of code (when possible) so that I don't scroll the file too much and can see almost everything on the same page. I find this approach more "structured" where everything sits perfectly where it should be, but I also understand that this is very opinionated.

One more advantage to this (a minor one) is when something goes wrong with your component and there's a warning or error in the console - you can see go to this component in IDE using the same file name as the component mentioned in the stacktrace.

Another advantage - it's easier to test your components. Even these small child components sometimes have some logic to test.
And if in some places you extract them but in others just put them inside the same file - it becomes inconsistent.

I'm still trying to figure out which approach is better.

d0sboots:
> Another advantage - it's easier to test your components. Even these small child components sometimes have some logic to test.

I'd be happy if we were testing *any* of our components XD

PR #600, opened 2023-06-11
MISC: fix warning caused by WIP component

bezrodnov:
There is a warning caused by the WorkInProgressRoot component saying that nesting p element inside the table body is wrong (or sm.th. like that)

While fixing it I found that gains in that table are unmounted and mounted to the DOM every game cycle (a small party in DOM with lights and music xD). So I added a key to each "gain" element instead of generating a unique key for every re-render iteration.

Tested locally.


Snarling (commented):
The rebase was just that we both messed with imports. I kept your ordering except I added the imports from `@enums` just below the `@player` import

PR #602, opened 2023-06-12
allowing empty Terminal new lines

kerpiaz:
CATEGORY: UI 
Allowing empty prompts in terminal so that someone who is used to the terminal can utilize the spamming of Enter to visually separate the commands he already ran and thus creating new lines.

### DIFF
--- a/src/Terminal/ui/TerminalInput.tsx
+++ b/src/Terminal/ui/TerminalInput.tsx
@@ -194,11 +194,15 @@ export function TerminalInput(): React.ReactElement {
   async function onKeyDown(event: React.KeyboardEvent<HTMLInputElement>): Promise<void> {
     const ref = terminalInput.current;
 
-    // Run command.
-    if (event.key === KEY.ENTER && value !== "") {
+    // Run command or insert newline.
+    if (event.key === KEY.ENTER) {
       event.preventDefault();
-      Terminal.print(`[${Player.getCurrentServer().hostname} /${Terminal.cwd()}]> ${value}`);
-      Terminal.executeCommands(value);
+      if (value !== "") {
+        Terminal.print(`[${Player.getCurrentServer().hostname} /${Terminal.cwd()}]> ${value}`);
+        Terminal.executeCommands(value);
+      } else {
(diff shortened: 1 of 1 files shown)


src/Terminal/ui/TerminalInput.tsx
```diff
@@ -194,11 +194,15 @@ export function TerminalInput(): React.ReactElement {
...
+    if (event.key === KEY.ENTER) {
       event.preventDefault();
-      Terminal.print(`[${Player.getCurrentServer().hostname} /${Terminal.cwd()}]> ${value}`);
-      Terminal.executeCommands(value);
+      if (value !== "") {
+        Terminal.print(`[${Player.getCurrentServer().hostname} /${Terminal.cwd()}]> ${value}`);
+        Terminal.executeCommands(value);
+      } else {
+        Terminal.print(`[${Player.getCurrentServer().hostname} /${Terminal.cwd()}]>`);
+      }
```
Snarling, line 205: Since the Terminal print is happening whether there's text or not, it doesn't need to be locked behind the conditional:
```suggestion
      Terminal.print(`[${Player.getCurrentServer().hostname} /${Terminal.cwd()}]> ${value}`);
      if (value !== "") Terminal.executeCommands(value);
```

kerpiaz: The terminal just sits there and does nothing if I only press ENTER with no commands in it. Only when a single character of any type is entered then it checks if it's a real command and does whatever it is coded for it. 
Here is without my update:

With no code change I press ENTER multiple times - nothing happens

When you press Enter in CMD or Linux terminal it just makes a new line with your dir path when no commands are there. My code simply adds this "realistic terminal experience" to the terminal. 
Here is with my update:

Code changed - I press enter and it makes a new line

Please excuse me if I did not explain this earlier in a better way and I hope these screenshots make it better to understand!

Snarling: Yes, I understood all that, I was just making a suggestion on better structured code for the PR. The suggested change still does the same thing.

Issue #372, opened 2023-02-17
Bladeburner/Sleeve - unequal naming GUI/scripts


NoCanDouCan:
There are some minor naming issues with the Bladeburner/Sleeve GUI and the script commands.
Some examples:

1. Bladeburner GUI
There are 5 major categories of tasks (General, **_Contracts_**, **_Operations_**, BlackOps,Skills)
If we use the function ns.bladeburner.getCurrentAction() the field type will be filled with
(General, **_Contract_**, **_Operation_**,BlackOps).

2. Sleeve GUI 
The task for the sleeve is called "Field analysis". In the Bladeburner GUI its called "Field Analysis".
And the script command would be ns.sleeve.setToBladeburnerAction(0,"Field analysis")

3. Sleeve script
ns.sleeve.getTask(i); will show "Contracts" and ns.sleeve.setToBladeburnerAction(i,"Take on contracts","Bounty Hunter");

I would recommend to use the same naming for all GUI and script commands .

- Version of the game: Bitburner v2.2.1 Steam

Snarling:
Totally agree. Unfortunately for things like this it will cause an API break if we standardize the internal code to match the UI names - ran into that with some similar changes in Corporation and Crime a couple patches back. Further renaming of stuff that affects API usage will be put off until 2.3

Zelow79:
For cases 2 and 3 the functions do actually match the GUI. The issue is you're considering the BB UI and not the Sleeve UI those come from.

![image](https://user-images.githubusercontent.com/32428876/220454750-8bc3da44-f26a-4107-a710-887080ed206f.png)
![image](https://user-images.githubusercontent.com/32428876/220454788-cb279924-9c30-45d0-bacd-ebff611df190.png)

d0sboots:
Perhaps the BB and Sleeve UIs should be standardized against each other?

d0sboots:
Also, it's *possible* to do this without an API break: You can continue to accept the old values as synonyms for the new values. Whether the compatibility is worth the hygiene hit is a judgement call.

Snarling:
If there is a name for a task or category of tasks in bladeburner that is also present for sleeves, it should definitely use consistent naming in both places.

On the topic of permissive processing of player input, see how player input for Crimes are currently handled (probably way too permissive):

```js
export function findCrime(roughName: string): Crime | null {
  if (checkEnum(CrimeType, roughName)) return Crimes[roughName];
  roughName = roughName.toLowerCase();
[...]

PR #604, opened 2023-06-12
MISC: fix sleeve university job

bezrodnov:
The "compute since" class type was mistyped as "study computer science".
Used the ClassType enum to fix this issue.
Ideally `takeUniversityCourse` should accept only ClassType as an argument, but `SleeveElem.tsx` will have to be rewritten to support this. So this is just a quick fix.

### DIFF
--- a/src/PersonObjects/Sleeve/Sleeve.ts
+++ b/src/PersonObjects/Sleeve/Sleeve.ts
@@ -214,23 +214,25 @@ export class Sleeve extends Person implements SleevePerson {
 
     // Set experience/money gains based on class
     let classType: ClassType | undefined;
+    // TODO: why lower case??? It's not effecient, not typesafe and in general a bad idea
     switch (className.toLowerCase()) {
-      case "study computer science":
+      case "study computer science": // deprecated, leave it here for backwards compatibility
+      case ClassType.computerScience.toLowerCase():
         classType = UniversityClassType.computerScience;
         break;
-      case "data structures":
+      case ClassType.dataStructures.toLowerCase():
         classType = UniversityClassType.dataStructures;
         break;
-      case "networks":
+      case ClassType.networks.toLowerCase():
         classType = UniversityClassType.networks;
         break;
-      case "algorithms":
+      case ClassType.algorithms.toLowerCase():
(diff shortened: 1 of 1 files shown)


bezrodnov:
Note: this fix might break scripts for those who used to pass `"study computer science"`  string

src/PersonObjects/Sleeve/Sleeve.ts
```diff
@@ -214,23 +214,24 @@ export class Sleeve extends Person implements SleevePerson {
 
     // Set experience/money gains based on class
     let classType: ClassType | undefined;
+    // TODO: why lower case??? It's not effecient, not typesafe and in general a bad idea
     switch (className.toLowerCase()) {
-      case "study computer science":
+      case ClassType.computerScience.toLowerCase():
```
d0sboots, line 220: You can (and probably should) keep the old case here as well for backwards compat. Very simple and harms nothing.

bezrodnov: I hope this piece of code will be rewritten one day cause it's ugly as hell xD

bezrodnov: This fix is a bit relevant to discussions in this ticket: https://github.com/bitburner-official/bitburner-src/issues/372.

Even though it's a game, it still has a publicly documented API which specifies the exact values to provide in scripts.
IMO it would be right to accept only these values and ask players to stick to docs. And maybe we should some time to adjust their scripts by deprecating invalid values and letting them know when there's an issue in a script.

d0sboots:
> I hope this piece of code will be rewritten one day cause it's ugly as hell xD

If you think that's ugly, you haven't looked at the savegame migration code yet (which can never be removed) XD

Snarling:
Yeah I think we can start enforcing strict types on the API in places like this where we are not, maybe at 2.4. Here's an example of the improved automatic error message we would get when the player passes in the wrong value for something that is a part of an enum/enumlike:

image 

I think that type of error is sufficient to point players in the right direction as far as correcting their code without too much frustration. As long as we make it easy for players to fix their scripts, I think we can force them to use exact types. But still should wait until the next major release.

PR #608, opened 2023-06-13
TYPESAFETY: Strict internal typing for AugmentationName

Snarling:
Did not provide an enum to the NS API yet or place strict types on the related API functions.

There was already an enum in place for this, it just wasn't being used properly.

Additionally, this PR attempts to initialize augmentations right out of the gate, instead of waiting for an initializer function, and to eliminate the deletion and recreation of augs every reset. However, on the first commit c8bd957, there is an import loop issue causing the game to not load, so this part may need to be reverted if the issue cannot be resolved.

c8bd957:
* Renamed StaticAugmentations to Augmentations (matches e.g. Factions / Companies / Programs etc)
* Changed keys of Augmentations, and name of each Augmentation, to be properly typed instead of just string. Lots of follow on changes from there in other files.
* Moved and restructured the metadata for initializing augs from AugmentationCreator.tsx to Augmentations.ts (in an IIFE).
* Removed or restructured functions related to reinitializing augmentations every reset. Only the CircadianModulator is reinitialized every time.

03033a5 and 1acf29a:
* Fix import loop issues that caused the game to fail to load (due to augmentations being initialized immediately on game load)

02b8098:
* Fix issue that displayed the wrong aug level for NFG when purchasing augs (showed the current level, not the next purchased level) 

b1eb2ca:
* Filter faction/player/sleeve augs on game load to remove any bad data.

1218456:
* Don't show a comprehensive list of possibilities in the error dialog when player misuses an enum with many options (like AugmentationName), instead show in the console


src/BitNode/BitNodeMultipliers.ts
```diff
@@ -1,168 +1,173 @@
...
-  [key: string]: number;
+  constructor(a: PartialRecord<keyof BitNodeMultipliers, number> = {}) {
+    for (const [key, value] of getRecordEntries(a)) this[key] = value;
+  }
 }
 
-/** The multipliers that are influenced by current Bitnode progression. */
-export const BitNodeMultipliers = Object.assign({}, defaultMultipliers);
+/** The multipliers currently in effect */
+export let currentNodeMults = new BitNodeMultipliers();
```
d0sboots, line 169: Although I like this better, I have a slight concern about it. If anything, anywhere in our codebase is keeping a reference to the BitNodeMultipliers, the switch away from const will cause an issue. I don't know of anything that *would*, but...

src/PersonObjects/Player/PlayerObjectGeneralMethods.ts
```diff
@@ -548,22 +549,16 @@ export function reapplyAllAugmentations(this: PlayerObject, resetMultipliers = t
     this.resetMultipliers();
   }
 
-  for (let i = 0; i < this.augmentations.length; ++i) {
-    //Compatibility with new version
-    if (this.augmentations[i].name === "HacknetNode NIC Architecture Neural-Upload") {
```
d0sboots, line 553: We don't still need this upgrade code somewhere?

src/SaveObject.ts
```diff
@@ -364,7 +363,7 @@ function evaluateVersionCompatibility(ver: string | number): void {
   if (typeof ver !== "number") return;
   if (ver < 2) {
     AwardNFG(10);
-    initAugmentations();
+    Player.reapplyAllAugmentations();
```
d0sboots, line 366: I'm 90% sure this was depending on initAugmentations() to set up factions. You should test with some of the old saves that were reported during the 2.3 transition to see if they will work.

d0sboots (commented):
Nice!

src/BitNode/BitNodeMultipliers.ts
Snarling: You still can only do a reassignment on a variable in the originating file, so the only place we need to worry about reassignments is in BitNodeMultipliers, all the other files still should essentially view it as a const

src/PersonObjects/Player/PlayerObjectGeneralMethods.ts
Snarling: Ah I had initially written that on a different branch where I had more safe load code (the other branch was converting person.augmentations to JSONMap, it ended up being way too huge before the first commit). Looks like I will need to re-add safe load code in (will just splice out any augs with invalid names)

src/BitNode/BitNodeMultipliers.ts
d0sboots: The difference is that something could have been holding a reference to BitNodeMultipliers, and before it would have still seen changes due to interior mutability.

Since you renamed the variable, we've seen all the callsites so it's probably fine.

src/SaveObject.ts
d0sboots: Wait, I had it turned around in my head. You're probably covered. Still, checking the old saves is a good idea.

(I have a collection of testing saves from various bugreports, maybe we should centralize them)

Snarling: Yeah it looks like that's why it was being called there, not for the Player effect (or maybe for both). That version probably had a lot of changes to faction augs (probably added a bunch of factions), and running that was the only way to get those augments to show up on the loaded factions without doing a soft reset first.

I'm fine with those players just not having the new augs available until they do a soft reset.

src/PersonObjects/Player/PlayerObjectGeneralMethods.ts
Snarling: Added safe load code to filter out invalid augs.

PR #615, opened 2023-06-15
UI: add cursor options to the script editor

bezrodnov:
https://github.com/bitburner-official/bitburner-src/assets/5138959/cd217b1d-f23c-4456-b781-0d7a63120a7b

Tested in the latest dev version, settings are persisted and loaded.

### DIFF
--- a/src/ScriptEditor/ui/Options.ts
+++ b/src/ScriptEditor/ui/Options.ts
@@ -1,4 +1,10 @@
+import type { editor } from "monaco-editor";
+
 export type WordWrapOptions = "on" | "off" | "bounded" | "wordWrapColumn";
+
+export type CursorStyle = editor.IEditorOptions["cursorStyle"];
+export type CursorBlinking = editor.IEditorOptions["cursorBlinking"];
+
 export interface Options {
   theme: string;
   insertSpaces: boolean;
@@ -9,4 +15,6 @@ export interface Options {
   fontLigatures: boolean;
   wordWrap: WordWrapOptions;
   vim: boolean;
+  cursorStyle: CursorStyle;
+  cursorBlinking: CursorBlinking;
 }
--- a/src/ScriptEditor/ui/OptionsModal.tsx
+++ b/src/ScriptEditor/ui/OptionsModal.tsx
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { ReactElement } from "react";
 
 import Button from "@mui/material/Button";
 import Typography from "@mui/material/Typography";
@@ -11,7 +11,10 @@ import EditIcon from "@mui/icons-material/Edit";
 import { useBoolean } from "../../ui/React/hooks";
 import { Modal } from "../../ui/React/Modal";
 import { ThemeEditorModal } from "./ThemeEditorModal";
-import { Options } from "./Options";
+import { CursorBlinking, CursorStyle, Options } from "./Options";
+
+const CURSOR_STYLES: CursorStyle[] = ["line", "block", "underline", "line-thin", "block-outline", "underline-thin"];
+const CURSOR_BLINKING_MODES: CursorBlinking[] = ["blink", "smooth", "phase", "expand", "solid"];
 
 export type OptionsModalProps = {
   open: boolean;
@@ -21,7 +24,7 @@ export type OptionsModalProps = {
   onThemeChange: () => void;
 };
 
-export function OptionsModal(props: OptionsModalProps): React.ReactElement {
+export function OptionsModal(props: OptionsModalProps): ReactElement {
   const [themeEditorOpen, { on: openThemeEditor, off: closeThemeEditor }] = useBoolean(false);
 
   const onFontSizeChange = (event: React.ChangeEvent<HTMLInputElement>) => {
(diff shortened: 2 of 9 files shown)


src/ScriptEditor/ui/Options.ts
```diff
@@ -1,4 +1,10 @@
+import { editor } from "monaco-editor";
```
Snarling: Should be an `import type` since it is not being used for any runtime code (we've been terrible about this in most of the codebase). It's possible webpack is already optimizing this out of the runtime code anyway.

src/ScriptEditor/ui/OptionsModal.tsx
```diff
@@ -13,14 +13,17 @@ import SaveIcon from "@mui/icons-material/Save";
...
+type OptionsModalProps = {
   options: Options;
   save: (options: Options) => void;
   onClose: () => void;
   open: boolean;
-}
+};
+
+const CURSOR_STYLES: CursorStyle[] = ["line", "block", "underline", "line-thin", "block-outline", "underline-thin"];
+const CURSOR_BLINKING: CursorBlinking[] = ["blink", "smooth", "phase", "expand", "solid"];
```
Snarling: Just a note, I'm not usually a fan of these ALL_CAPS_CONSTANTS but it's fine here and I see why you couldn't really use an enum here which normally would have been more appropriate (type is determined externally so enum is harder to use).

Snarling (commented):
Looks good, just see the one note about `import type` which I think we should use wherever possible to differentiate real imports from type-only imports

src/ScriptEditor/ui/OptionsModal.tsx
bezrodnov: Caps here totally make sense to me - when I see non-caps variable when reading a React component code I assume it's either a prop or a comp-scoped constant (sometimes it's also some 3rd party, but usually those are functions).

Caps help me (and others, I hope) understand, that the value is not expected to be modified in any way. And I don't have to use IDE or scroll back and forth the source file to understand this. But of course, I don't use caps everywhere where I have a const. This is just a place where I feel it makes the most sense.

bezrodnov: Just a note about enums - quite often it's a better choice in a similar situation. An exception could be that you want to display options in a different order or to define just a subset of all possible values.
But yeah, in this specific case it does not really matter - if there was an enum provided by monaco, I'd use it.

bezrodnov: Thank you for the review by the way!
Like your PR comments :)

bezrodnov:
@Snarling, I've just noticed one minor bug - when switching between vim and non-vim mode - cursor settings are kinda lost.
Looking into it...

bezrodnov:
and one more existing bug - if Save button is not clicked and Options are opened again - the values won't get reset.

d0sboots:
> Looks good, just see the one note about `import type` which I think we should use wherever possible to differentiate real imports from type-only imports

Webpack does indeed optimize these out, which has been saving our butt in terms of circular imports. There's a TS mode that would turn real imports used for type only into errors, which we could enable if we ever wanted to switch everything over and keep it clean.

bezrodnov:
Reverted the fix for #625 from this PR as it requires a better solution

src/ScriptEditor/ui/OptionsModal.tsx
```diff
@@ -61,7 +71,7 @@ export function OptionsModal(props: IProps): React.ReactElement {
 
   return (
     <Modal open={props.open} onClose={props.onClose}>
-      <ThemeEditorModal open={themeEditorOpen} onClose={() => setThemeEditorOpen(false)} />
+      <ThemeEditorModal open={themeEditorOpen} onClose={openThemeEditor} />
```
Snarling: Shouldn't this be onClose={closeThemeEditor}?

src/ScriptEditor/ui/OptionsModal.tsx
```diff
@@ -74,7 +84,7 @@ export function OptionsModal(props: IProps): React.ReactElement {
           <MenuItem value="one-dark">one-dark</MenuItem>
           <MenuItem value="customTheme">Custom theme</MenuItem>
         </Select>
-        <Button onClick={() => setThemeEditorOpen(true)} sx={{ ml: 1 }} startIcon={<EditIcon />}>
+        <Button onClick={closeThemeEditor} sx={{ ml: 1 }} startIcon={<EditIcon />}>
```
Snarling: Shouldn't this button be opening the theme editor instead of closing it?

Snarling (commented):
I didn't get a chance to test the functionality, but the changes make it look like there's some mismatched functions

bezrodnov:
let's merge this after https://github.com/bitburner-official/bitburner-src/pull/626 - there will be some conflicts which I plan to resolve. I believe in that branch I did not make these typos xD

src/ScriptEditor/ui/Toolbar.tsx
```diff
@@ -40,8 +40,13 @@ export function Toolbar({ editor, onSave }: IProps) {
   const { ram, ramEntries, isUpdatingRAM, options, saveOptions } = useScriptEditorContext();
 
   const onOptionChange: OptionsModalProps["onOptionChange"] = (option, value) => {
-    saveOptions({ ...options, [option]: value });
-    editor?.updateOptions(options);
```
bezrodnov, line 44: @Snarling there was a minor bug in a previous PR - instead of applying new options, the previous ones were applied.
this PR should fix it

it was not me xD

PR #619, opened 2023-06-16
FIX: singularity faction donation formula

### LINKED ISSUE #618: singularity.donateToFaction is cheaper than manual donating
Version 2.3.1 (Github release, non-steam zip, firefox)

I tried to figure out how much money it costs to gain a fixed amount of reputation. I'm currently in BN4x1 Singularity Level 1 (my first non-starter bitnode). I tried two approaches: Calculating from known constants, and calculating from a real donation.

`test.js`
```
const NODEGAIN = {4:0.75}
function calcDonationForRep(ns,rep){
  const p    = ns.getPlayer()
  const div  = 1e6 //DonateMoneyToRepDivisor (hardcoded)
  const mult = p.mults.faction_rep // 3.62285413582752 (All Normal Augs, No special faction, NeuroFlux L83)
  const gain = NODEGAIN[p.bitNodeN] //BitNodeMultipliers.FactionWorkRepGain
  ns.tprint('calcDonationForRep: ', (div) / (mult * gain))
  return (rep * div) / (mult * gain)
}

//how much does 1 rep cost?
function getRepCostFactor(ns, rep, faction = 'BitRunners'){
  const amount = 1e3
  const rep1 = ns.singularity.getFactionRep(faction)
  ns.singularity.donateToFaction(faction, amount)
  const rep2 = ns.singularity.getFactionRep(faction)
  const costPerRep = amount/(rep2-rep1)
  ns.tprint('getRepCostFactor: ', costPerRep)
  return costPerRep * rep
}

  try {
    ns.tprint(calcDonationForRep(ns,eval(ns.args[0])))
    ns.tprint(getRepCostFactor(ns,eval(ns.args[0])))
  } catch (err) {
    ns.tprint(err)
  }
}
```

However the terminal output gave weirdly different output, which i at first thought was a bug in my code.
`[home /]> run test.js 1e6`
```
test.js: calcDonationForRep: 368033.95426484034
test.js: 368033954264.8404
test.js: getRepCostFactor: 276025.3243434725
test.js: 276025324343.4725
```

However i noticed that the numbers are **both correct**! When i manually donate `368033954264.8404` to i.e. BitRunners i gain exactly 1.000m reputation, however if i automatically donate 
`ns.singularity.donateToFaction('BitRunners', 276025324343.4725)` 
i also gain exactly 1.000m reputation! The difference is exactly `0.75` which I believe is the `BitNodeMultipliers.FactionWorkRepGain` for Singularity Level 1 (I don't have API access to BitNodeMultipliers yet)

bezrodnov:
closes #618

### DIFF
--- a/src/Faction/formulas/donation.ts
+++ b/src/Faction/formulas/donation.ts
@@ -1,11 +1,11 @@
-import { Person } from "@nsdefs";
+import { Person as IPerson } from "@nsdefs";
+import { Player } from "@player";
 
 import { CONSTANTS } from "../../Constants";
 import { currentNodeMults } from "../../BitNode/BitNodeMultipliers";
-import { PlayerObject } from "../../PersonObjects/Player/PlayerObject";
 import { Faction } from "../Faction";
 
-export function repFromDonation(amt: number, person: Person): number {
+export function repFromDonation(amt: number, person: IPerson): number {
   return (amt / CONSTANTS.DonateMoneyToRepDivisor) * person.mults.faction_rep * currentNodeMults.FactionWorkRepGain;
 }
 
@@ -13,18 +13,18 @@ export function repNeededToDonate(): number {
   return Math.floor(CONSTANTS.BaseFavorToDonate * currentNodeMults.RepToDonateToFaction);
 }
 
-export function canDonate(amt: number, player: PlayerObject): boolean {
-  return !isNaN(amt) && amt > 0 && player.money >= amt;
+export function canDonate(amt: number): boolean {
+  return !isNaN(amt) && amt > 0 && Player.money >= amt;
 }
 
 /** Donates money to the faction provided and returns repuation gained */
-export function donate(amt: number, player: PlayerObject, faction: Faction) {
-  if (!canDonate(amt, player)) {
+export function donate(amt: number, faction: Faction) {
+  if (!canDonate(amt)) {
     return 0;
   }
 
-  const repGain = repFromDonation(amt, player);
-  player.loseMoney(amt, "other");
+  const repGain = repFromDonation(amt, Player);
+  Player.loseMoney(amt, "other");
   faction.playerReputation += repGain;
 
   return repGain;
--- a/src/Faction/formulas/donation.ts
+++ b/src/Faction/formulas/donation.ts
@@ -1,7 +1,31 @@
+import { Person } from "@nsdefs";
+
 import { CONSTANTS } from "../../Constants";
 import { currentNodeMults } from "../../BitNode/BitNodeMultipliers";
-import { Person as IPerson } from "@nsdefs";
+import { PlayerObject } from "../../PersonObjects/Player/PlayerObject";
+import { Faction } from "../Faction";
 
-export function repFromDonation(amt: number, person: IPerson): number {
+export function repFromDonation(amt: number, person: Person): number {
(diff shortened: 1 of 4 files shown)


d0sboots:
👍

bezrodnov:
PR is in the draft state cause I started working on tests to make sure both ways use the same value.

Snarling:
@bezrodnov Sorry it looks like #608 which I just merged caused some issues with this one requiring a rebase.

The major thing I'm seeing that might have caused an impact is that the currently active bitnode multipliers are now accessed via a variable named currentNodeMults (instead of BitnodeMultipliers, which is now the name of a class)

bezrodnov:
@Snarling no problem, it's draft anyway and should be very easy to rebase :)

bezrodnov:
I'll play with tests another time

src/Faction/formulas/donation.ts
```diff
@@ -1,7 +1,31 @@
+import { Person } from "@nsdefs";
+
 import { CONSTANTS } from "../../Constants";
 import { currentNodeMults } from "../../BitNode/BitNodeMultipliers";
-import { Person as IPerson } from "@nsdefs";
```
Snarling: The reason Person was named differently here is because there are two separate `Person` types.

`Person` from `@nsdefs` is an interface which just defines a minimum personlike structure for players passing in person objects to formulas functions.

`Person` is also the name of an actual game class, which does fulfill the Person interface but also has member functions and some other properties that aren't visible to players through the API.

So typically when we use the ns API versions of Person/Player/Sleeve, we don't refer to them using those names directly because we want the code to be clear that we're not talking about the ingame class, we're talking about the ns API interface version. The exact name doesn't really matter (could call it NSPerson, I have been fine keeping these specific instances as IPerson as well since the I is actually conveying information in this case)

src/Faction/formulas/donation.ts
```diff
@@ -1,7 +1,31 @@
...
+  if (!canDonate(amt, player)) {
+    return 0;
+  }
+
+  const repGain = repFromDonation(amt, player);
+  player.loseMoney(amt, "other");
+  faction.playerReputation += repGain;
+
+  return repGain;
+}
```
Snarling: For functions that use *the* player as the actor, we just use the imported player, there's no need to pass the player around as a function argument if there's exactly one valid thing you can pass as that argument. If it's allowed to be a Person in general, we would take in a Person instead of a PlayerObject.

Snarling (changes requested):

src/Faction/formulas/donation.ts
bezrodnov: Thanks for the explanation :)

It looks a bit strange that the class and interface have the same name which forces us to use an alias for the interface when importing it in-game.
Maybe naming the class as something like `AbstractPerson` would help both in not having to import as `IPerson`  the interface and also for code readability (when you see `AbstractPerson` you know, that this is a class with some basic implementation/attributes representing either a player or a sleeve)
wdyt?

d0sboots: I'm pretty OK with the way things are. There isn't a better name than Person, or Script, or Server, etc. At the same time, there's no reason to name the public interface that we derive from these class anything else, either. It only causes friction in the (few) places where both are used.

Snarling (approved):

PR #627, opened 2023-06-18
FIX: show ram calculation error reason to player

bezrodnov:
There was a discussion in Discord where people complained about not being able to understand where exactly syntax error was detected in their code. This PR should give players a bit more transparency on what went wrong.

Now when RAM button is clicked in the Code Editor toolbar player should see where syntax error was found:

image

And the same for running such script in the Terminal:

image

PR is also adding more info in case of import error:

image

image

### DIFF
--- a/src/Script/RamCalculations.ts
+++ b/src/Script/RamCalculations.ts
@@ -362,17 +362,17 @@ function parseOnlyCalculateDeps(code: string, currentModule: string): ParseDepsR
 
 /**
  * Calculate's a scripts RAM Usage
- * @param {string} codeCopy - The script's code
+ * @param {string} code - The script's code
  * @param {Script[]} otherScripts - All other scripts on the server.
  *                                  Used to account for imported scripts
  */
 export function calculateRamUsage(
-  codeCopy: string,
+  code: string,
   otherScripts: Map<ScriptFilePath, Script>,
   ns1?: boolean,
 ): RamCalculation {
   try {
-    return parseOnlyRamCalculate(otherScripts, codeCopy, ns1);
+    return parseOnlyRamCalculate(otherScripts, code, ns1);
   } catch (e) {
     console.error(`Failed to parse script for RAM calculations:`);
     console.error(e);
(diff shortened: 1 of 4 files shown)


bezrodnov:
Without hiding whitespace differences reviewing this is a bit painful

image

src/Script/RamCalculations.ts
```diff
@@ -80,115 +91,108 @@ function parseOnlyRamCalculate(otherScripts: Map<ScriptFilePath, Script>, code:
...
+  // Process additional modules, which occurs if the "main" script has any imports
+  while (parseQueue.length > 0) {
+    const nextModule = parseQueue.shift();
+    if (nextModule === undefined) throw new Error("nextModule should not be undefined");
+    if (nextModule.startsWith("https://") || nextModule.startsWith("http://")) continue;
 
-      resolvedRefs.add(ref);
+    // Using root as the path base right now. Difficult to implement
+    const filename = resolveScriptFilePath(nextModule, root, ns1 ? ".script" : ".js");
+    if (!filename) return { errorCode: RamCalculationErrorCode.ImportError }; // Invalid import path
```
bezrodnov: we could actually make this transparent to players too by adding fileName to `error` (and probably changing it to `errorMessage` (string | undefined). wdyt?

bezrodnov: I assume it totally makes sense and added it as well.

src/Script/Script.ts
```diff
@@ -17,6 +17,7 @@ export class Script implements ContentFile {
   // Ram calculation, only exists after first poll of ram cost after updating
   ramUsage: number | null = null;
   ramUsageEntries: RamUsageEntry[] = [];
+  ramCalculationError?: string | null = null;
```
d0sboots: This shouldn't be both optional *and* nullable. Nullable is preferred, for monomorphism.

d0sboots (approved):

Snarling (commented):
Seeing console error spam now when editing file (or any other time ram is calculated) (sorry about the weird noise on video)

https://github.com/bitburner-official/bitburner-src/assets/84951833/e8fc39f4-c818-4a90-a2da-29d1bd72cbca

Other than that, looks good.

src/Script/Script.ts
bezrodnov: forgot to remove that

PR #629, opened 2023-06-19
QUICKFIX: aug list

bezrodnov:
Quick fix for a bug reported in the Discord

image

see https://discord.com/channels/415207508303544321/415213413745164318/1120356609005207663


bezrodnov:
In the same save file after this change:

image

Snarling:
Just merging this as a bandaid, working on the correct fix for this.

Snarling:
Actually I had to revert it. Set is not compatible with JSON so is not compatible with save file.

PR #632, opened 2023-06-19
MISC: add log window size lock/unlock feature

bezrodnov:
There was a bit strange feature request in Discord: https://discord.com/channels/415207508303544321/415213435974975508/1120357898158080001

One of the options to address it could be
- add an icon to the tail window to lock/unlock window resize.
- add a new netscript function: `setTailSizeLocked`

Note: as per the current solution window won't be resized even with netscript functions if it's locked.

https://github.com/bitburner-official/bitburner-src/assets/5138959/50a873d1-cff5-44a6-b5f7-323a31a934e3

### DIFF
--- a/markdown/bitburner.ns.md
+++ b/markdown/bitburner.ns.md
@@ -9,7 +9,7 @@ Collection of all functions passed to scripts
 **Signature:**
 
 ```typescript
-export interface NS 
+export interface NS
 ```
 
 ## Remarks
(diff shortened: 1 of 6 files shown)


d0sboots:
I've probably missed discussion on Discord, but what problem is this solving? Can't you just... not drag the corner? No other UI acts like this.

bezrodnov:
> I've probably missed discussion on Discord, but what problem is this solving? Can't you just... not drag the corner? No other UI acts like this.

Updated description with a reference to the [discusson on Discord](https://discord.com/channels/415207508303544321/415213435974975508/1120357898158080001).
I was also a bit skeptical. And I'm still not sure if we need to do it and whether this solution is good enough.
Suggestions are welcome.

bezrodnov:
This is a very niche feature request from some of the players which probably does not make much sense.
In the end, they can use some workarounds like restarting their script or changing tail window size via script in a loop.
So I don't mind if this PR gets closed.

Snarling:
Yeah really just seems like a bad feature request.

PR #634, opened 2023-06-20
UPDATE bitburner.ns.scp.md

htmlh4cker:
"files" needs to be declared as const, tested it in a script without and it doesn't work without it.


Snarling:
Markdown files don't get edited directly, instead the definitions need to be updated in src/ScriptEditor/NetscriptDefinition.d.ts. The .md files are autogenerated from that file

htmlh4cker:
understood. Changed it in src/ScriptEditor/NetscriptDefinition.d.ts. and created a new pull request

d0sboots:
(In the future) you don't need to make a new PR, you can push changes to the one you have

Snarling:
Already merged as #636

Issue #630, opened 2023-06-19
Change alias to not substitute the same alias recursively


Hoekstraa:
As of now (dev build included), `alias` doesn't behave like POSIX spec says.
It naively repeats substitution 10 times if a recursive loop is found.
POSIX spec however mentions:
> To prevent infinite loops in recursive aliasing, if the shell is not currently processing an alias of the same name, the word shall be replaced by the value of the alias; otherwise, it shall not be replaced.

[source](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_03_01)

The current implementation can create a situation like the following (thank you, Noxebo on Discord):

![image](https://github.com/bitburner-official/bitburner-src/assets/19653491/5ef43890-4a88-4803-ad61-e0c886bc09d6)

How to resolve:
keep track of the last used alias, and if the same alias is used again directly after, don't substitute.

Snarling:
This was partially addressed by #635 but we may want to do more of a rewrite to provide full recursion-detection (instead of just checking the most recent alias substitution)

PR #635, opened 2023-06-20
QUICKFIX: Issue #630

shyguy1412:
Prevent recursive alias by checking if the last resolved alias is about to be resolved again. Includes check to make sure chaining the same alias still works as expected:

alias script="run script.js script";
"script" will now resolve to "run script.js script"
"script script" would still resolve to "run script.js script run script.js script"

### DIFF
--- a/src/Alias.ts
+++ b/src/Alias.ts
@@ -76,6 +76,7 @@ export function substituteAliases(origCommand: string): string {
 
     let somethingSubstituted = true;
     let depth = 0;
+    let lastAlias;
 
     while (somethingSubstituted && depth < 10) {
       depth++;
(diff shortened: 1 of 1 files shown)


Snarling:
Based on this discussion:
https://discord.com/channels/415207508303544321/459097632896188436/1120366677473906748

"When the shell performs alias substitution, it checks to see if value ends with a blank. If so, the shell also checks the next word of the command line for aliases. The shell then checks the new command line for aliases and expands them, following these same rules. This process continues until there are no aliases left on the command line, or recursion occurs in the expansion of aliases."

I think this may actually need to take into account more than just the 1 previous alias, really it should take into account the entire stack of aliases that led to the current substitution, that way we can accurately check whether a recursion has occurred during substitution, even if that recursion is more than 1 alias substitution deep.

However for now this fix seems to be a good improvement, and tracking an entire stack of alias substitutions may require a broader rewrite. Thanks!

PR #640, opened 2023-06-21
MISC: enforce eslint react checks

bezrodnov:
Added eslint react and react hooks plugins to improve code quality and safety.

Checklist before merge:
- [ ] I've tested the game with these changes
- [x] I've reviewed all the changes
- [ ] I like what I've done xD


src/DevMenu/ui/Bladeburner.tsx
```diff
@@ -18,13 +18,11 @@ import { Player } from "@player";
...
 import { SkillNames } from "../../Bladeburner/data/SkillNames";
+import { Bladeburner } from "../../Bladeburner/Bladeburner";
 
 const bigNumber = 1e27;
 
-export function Bladeburner(): React.ReactElement {
-  if (!Player.bladeburner) return <></>;
-  const bladeburner = Player.bladeburner;
-
+function BladeburnerUI({ bladeburner }: { bladeburner: Bladeburner }): React.ReactElement {
```
bezrodnov: I don't like this name tbh... maybe aliasing import for `Bladeburner` game feature would be better than renaming this UI component.

bezrodnov:
This one needs careful testing before merging.

src/DevMenu/ui/Bladeburner.tsx
Snarling: The whole structure change in this file is a bit confusing. 

Two better solutions for this file:
1. Just to move all the useState calls up to the top before the early return (the normal fix for when we are conditionally returning before all hooks have been called). While it means useState hook calls won't be organized next to the functions that are about those state variables, having the hook calls at the top separate from everything else is just normal react.

2. If we were going to pass bladeburner as a prop (something we usually avoid mainly due to react exposing props to global scope, but it's fine in the devmenu) then DevMenu.tsx would just pass bladeburner as a prop and we wouldn't need the weird wrapper or the early return check.

src/CotMG/ui/StaneksGiftRoot.tsx
```diff
@@ -19,7 +19,7 @@ interface IProps {
 
 export function StaneksGiftRoot({ staneksGift }: IProps): React.ReactElement {
   const rerender = useRerender();
-  useEffect(() => StaneksGiftEvents.subscribe(rerender), []);
+  useEffect(() => StaneksGiftEvents.subscribe(rerender), [rerender]);
```
Snarling, line 22: Since useRerender returns a useCallback now, don't we know that `rerender` will never change value, and there is no point in this change to the useEffect params?

src/ui/React/CharacterOverview.tsx
```diff
@@ -325,11 +329,11 @@ function WorkInProgressOverview({ tooltip, children, header }: WorkInProgressOve
...
 
 function Work(): React.ReactElement {
-  const setRerender = useState(false)[1];
+  const rerender = useRerender();
   useEffect(() => {
-    const clearSubscription = OverviewEventEmitter.subscribe(() => setRerender((old) => !old));
+    const clearSubscription = OverviewEventEmitter.subscribe(rerender);
     return clearSubscription;
-  }, []);
+  }, [rerender]);
```
Snarling, line 336: Same comment about rerender not being needed as a param on useEffect because it has a static value for each component.

Snarling (commented):
Just a partial review, I've only looked at 15/50 files so far. Most of the changes I've seen so far seem straightforward and safe (mostly moving hook usage before potential early returns)

src/CotMG/ui/StaneksGiftRoot.tsx
bezrodnov: That's true, but react hooks eslint plugin does not know that and we are forced to mention `rerender` as a dependency of `useEffect`, `useCallback` or `useMemo` (unless there's a way to configure this plugin to ignore it, just like it does for example in case of `useRef` variables)

src/DevMenu/ui/Bladeburner.tsx
bezrodnov: 1. I had to create this wrapper because just moving the "return empty fragment block" to the end of the component would cause NPEs in many places (`Player.bladeburner` can be null)

2. I like the second point though. In some of my previous projects there was even a rule that react component cannot return an empty fragment, null, or whatever else. I.e. if a component won't be able to render stuff - just don't ask it to do so, don't render it as a child of another component. But this approach has a downside in that a parent component needs to know how rendering is done by the child component. In our case, it should be fine though.

bezrodnov: BTW, it's even already the case - in `DevMenu.tsx`:
```tsx
{Player.bladeburner && <BladeburnerElem />}
```
So I'll just add it as a prop and wrapper won't be needed. But I still need a good name either for import alias or for UI component, since they are the same :)

bezrodnov: P.S. existing `BladeburnerElem` does not sound nice to me xD

src/CotMG/ui/StaneksGiftRoot.tsx
bezrodnov: More on this topic: https://github.com/facebook/react/issues/16873

Snarling: Ugly but I guess there's no harm in including it in the dependencies.

bezrodnov: Totally agree. I was looking for a solution to this, but it's just not provided in the official plugin :(

src/DevMenu/ui/Bladeburner.tsx
Snarling: Yeah there's a conflict between the BladeburnerRoot (root component of the actual bladeburner UI), Bladeburner (an ingame class), and whatever this is.

I think the most descriptive name is something like BladeburnerDevOptions. While it's long, it clearly explains in the name what its purpose is and it's clear that it's not related to either of the other two Bladeburners above. It doesn't need to have different names in the two files (previously Bladeburner in one file, BladeburnerElem in the other)

Snarling: ~~Re: the useStates not being able to be moved to the top, I don't think any of those rely on Player.bladeburner so I'm not sure I see what the conflict would have been~~ Ignore that, I missed that e.g. AllOperations was just something we're plucking from the bladeburner object.

bezrodnov: > I think the most descriptive name is something like BladeburnerDevOptions

I like having a name like this. The only thing I don't like is that it's inconsistent with other "dev" components.
Maybe we can rename them all (as well as the files they are located in) to something like `XxxDev` (BladeburnerDev, CompaniesDev, etc)?

src/Infiltration/ui/Game.tsx
```diff
@@ -49,32 +49,21 @@ export function Game(props: IProps): React.ReactElement {
...
 
-  function nextGameId(): number {
-    let id = gameIds.lastGames[0];
-    const ids = [gameIds.lastGames[0], gameIds.lastGames[1], gameIds.id];
-    while (ids.includes(id)) {
-      id = Math.floor(Math.random() * minigames.length);
-    }
-    return id;
-  }
+  const setupNextGame = useCallback(() => {
```
bezrodnov, line 52: Infiltration changes is one of the things I'm worried about the most in this PR

src/Infiltration/ui/SlashGame.tsx
```diff
@@ -39,24 +40,27 @@ export function SlashGame(props: IMinigameProps): React.ReactElement {
...
         id = window.setTimeout(() => props.onFailure(), attackingTime);
       }, preparingTime);
-    }, guardingTime);
+    }, guardingTimeRef.current);
+
     return () => {
       clearInterval(id);
     };
-  }, []);
+  }, [difficulty.window, props]);
```
bezrodnov: `props` in dependencies look wrong. I need to revisit this place and see if I can use refs to avoid it.

src/ScriptEditor/ui/ScriptEditorRoot.tsx
```diff
@@ -444,6 +436,19 @@ function Root(props: IProps): React.ReactElement {
     onOpenPreviousTab,
   });
 
+  const onTabClickRef = useRef(onTabClick);
+  onTabClickRef.current = onTabClick;
+  const parseCodeRef = useRef(parseCode);
+  parseCodeRef.current = parseCode;
```
bezrodnov: and this is also one of the ugliest things I've done this year xD

bezrodnov: That was a quick and dirty solution. I'm looking for something better here.

src/DevMenu/ui/Bladeburner.tsx
Snarling: Yeah that would be fine for consistency

src/ScriptEditor/ui/ScriptEditorRoot.tsx
Snarling: Is the entire point of these refs to allow you to pass the lint rule while not passing dependencies to the useEffect?

If so, two other possible solutions, both also ugly in their own way:
1. Just use an eslint ignore next line comment instead of making runtime changes
2. Default parameters in the useEffect function, e.g.

```js
useEffect((parseCode = parseCode, onTabClick = onTabClick) => {
    if (currentScript !== null) {
      const tabIndex = currentTabIndex();
      if (typeof tabIndex === "number") onTabClick(tabIndex);
      parseCode(currentScript.code);
    }
  }, []);
```

bezrodnov: I actually think there was a bug in this code. As far as I remember, these custom vim commands did not work, probably because of wrong hook deps. I need ro test it.

These refs allow you not to pass dependensies, but still use the latest values from the last rerender. They don't trigger effects or memoized hook recalculations, they solve both linter and runtime issues... but they look ugly xD

bezrodnov: > 2\. Default parameters in the useEffect function

That's something I haven't seen before. And I don't think it's even supported by React as effects have no any arguments.

Snarling: > These refs allow you not to pass dependensies, but still use the latest values from the last rerender. They don't trigger effects or memoized hook recalculations, they solve both linter and runtime issues... but they look ugly xD

The effect is only firing exactly once on the first render of the component, the effect doesn't happen again after that. So it's not using "the latest value from the last rerender", it's just using the value from the first render, because that effect never gets called again.

> (default parameters) That's something I haven't seen before. And I don't think it's even supported by React as effects have no any arguments.

It works because useEffects don't receive any arguments, which leads to the default value always being used. The default values would just be the values of those functions from the first render (so functionally identical to what it is now)

I got the idea essentially from here: https://stackoverflow.com/questions/55840294/how-to-fix-missing-dependency-warning-when-using-useeffect-react-hook/61183162#61183162

But making a ref only to use it in a one-time useEffect is pointless, on subsequent renders you are updating a ref but not even using the updated ref for anything.

bezrodnov: > > These refs allow you not to pass dependensies, but still use the latest values from the last rerender. They don't trigger effects or memoized hook recalculations, they solve both linter and runtime issues... but they look ugly xD
> 
> The effect is only firing exactly once on the first render of the component, the effect doesn't happen again after that. So it's not using "the latest value from the last rerender", it's just using the value from the first render, because that effect never gets called again.
> 
> > (default parameters) That's something I haven't seen before. And I don't think it's even supported by React as effects have no any arguments.
> 
> It works because useEffects don't receive any arguments, which leads to the default value always being used. The default values would just be the values of those functions from the first render (so functionally identical to what it is now)
> 
> I got the idea essentially from here: https://stackoverflow.com/questions/55840294/how-to-fix-missing-dependency-warning-when-using-useeffect-react-hook/61183162#61183162
> 
> But making a ref only to use it in a one-time useEffect is pointless, on subsequent renders you are updating a ref but not even using the updated ref for anything.

ok, instead of many words let me demonstrate to you why even with a single effect call it still matters: https://playcode.io/1513356
Here you can see two almost identical components in the bottom.
When you click the first one it prints the same value to console even though underlying function should increase the counter on each click.
When you click the second one (with refs passed down to use effect) - the result is different. And that's the kind of result we need.

Update: I added a third component using the approach you suggested (with default args in useEffect) - it has the same result as the first one (without refs)

bezrodnov: Explanation: refs work like class fields in the class-based react components. They are the same objects which are passed to every render call in the component (`ref.current` can be different, but `ref` says the same).

So if during the effect call, which happens only on component mount, you register some event handlers or some delayed action on timeout/interval, etc... using a ref - you will use the updated `ref.current` value whenever the handler you registered is actually triggered. But if you pass down a regular constant, not ref, then you will basically use a fixed value which will never get changed, unless you put this value as a dependency and re-register you handler with a new value.

Snarling: 1. Each of the 3 buttons does increase the actual counter, they are just console.logging stale values.
2. Your example code is not like the actual game code here.
    * Your code is running the ref'd function at a time other than immediately (by placing it as an onclick). This provides a possibility for the function to become "stale" and need to get an updated reference. Game code in this useEffect is running immediately in the useEffect and never again: there is no possibility for staleness.
    * Game code is referencing a function which is defined inside the same component, which means just directly referencing that function is already an "up to date reference" to the function even if the useEffect was ran more than once. Although if the useEffect was adding delayed code (which it's not) a ref might still make sense even with a function defined in the same component.

bezrodnov: > 1. Each of the 3 buttons does increase the actual counter, they are just console.logging stale values.

But that's exactly the point! Any non-memoized value (all except useCallback, useMemo, setState, useRef and static constants defined outside of component) you pass to useEffect/useCallback/useMemo without mentioning it as dependency will be "stale".

> 2. Your example code is not like the actual game code here.

Yes, but only in case of this specific place cause it does not have any kind of delays (timeouts/intervals, microtasks) or subscriptions (e.g. event listeners) and is executed immediately after the first render and only once.

My example just makes it obvious that without using ref (or passing down value to dependencies list) `useEffect` won't work as you expect it to in cases of delayed actions. As an example, you can check this code https://github.com/bezrodnov/bitburner-src/blob/enforce-eslint-rules/src/ScriptEditor/ui/useVimEditor.tsx#L27 where without using ref the code will not work properly as it will use stale values instead of the actual ones.


As for passing down arguments to useEffect's effect function - this is a very hacky thing that I have never seen since React hooks release. Not only it looks strange, but also does not solve the issue as react hooks plugin still complains about these arguments not being passed as hook dependencies:
image

bezrodnov: > Just use an eslint ignore next line comment instead of making runtime changes

This is also something I try to avoid at all costs. It's just not safe at the long run unless this code never changes again or whoever changes it in the future understands and notices that there's a lint rule being ignored here and knows how to fix it.

Snarling: > useEffect was needed for vim

That is because vim needs to be able to call fresh versions of the functions later on, when that functionality is invoked through vim keybindings in a delayed manner. This instance is not like that, there is no delayed manner in which the functions are potentially being used inside useEffect code, it's all immediate.

> default parameters

Yeah this doesn't even seem to work for clearing the lint failure, so it's off the table. Based on what I saw in that stackoverflow link I thought it was enough to satisfy the lint rule, my bad.


I get the aversion to using a lint ignore statement. If you want to keep the useRefs, at least get rid of the updating of ref.current every render - there is no point updating a ref when we know it is never going to be used again (although maybe this would also demand a variable rename, so it's clear that these refs are not kept current)

d0sboots: So, I've been following along here... It feels like this shouldn't be a separate effect at all? Can't it be put in beforeMount or onMount?

d0sboots (commented):
Looking over this... I'm not sure this is an improvement. The pervasive use of refs is *also* against React style, there just isn't a lint rule for it. Furthermore, an effect with no dependencies has a clear effect, whereas a ref whose value is never updated is quite a subtle footgun.

src/ScriptEditor/ui/ScriptEditorRoot.tsx
bezrodnov: > Can't it be put in beforeMount or onMount?

This is a type of change I tried to avoid in this PR cause it might introduce bugs. This PR was purely about enforcing lint rules without changing behavior (but I realize that in some places I've already broken this intention).
For now, I just added a lint ignore instruction.

src/Infiltration/ui/SlashGame.tsx
Snarling: The only difference between here and other infiltration files is that you're not destructuring props here. Most of the others have props in the dependency array as well, just in their destructured pieces.

src/PersonObjects/Sleeve/ui/StatsElement.tsx
```diff
@@ -137,21 +137,21 @@ export function EarningsElement(props: IProps): React.ReactElement {
       [`Dexterity Exp:`, `${formatExp(CYCLES_PER_SEC * rates.dexExp)} / sec`],
       [`Agility Exp:`, `${formatExp(CYCLES_PER_SEC * rates.agiExp)} / sec`],
       [`Charisma Exp:`, `${formatExp(CYCLES_PER_SEC * rates.chaExp)} / sec`],
-      [`Reputation:`, <ReputationRate reputation={CYCLES_PER_SEC * repGain} />],
+      [`Reputation:`, <ReputationRate key="reputation" reputation={CYCLES_PER_SEC * repGain} />],
```
Snarling: This key could be renamed to reputation-rate to match the one for sleeve company work. The money-rate keys already match between SleeveClassWork and SleeveCompanyWork

Snarling:
This will just need a very brief rebase due to #644 , I included the same name change to FactionsDev which I also had to change some code for due to type changes. I think the existing FactionsDev.tsx should already pass the lint checks.

Testing out infiltration on your branch now

src/Sidebar/ui/SidebarRoot.tsx
```diff
@@ -160,6 +163,24 @@ export function SidebarRoot(props: IProps): React.ReactElement {
...
   const canStaneksGift = Player.augmentations.some((aug) => aug.name === AugmentationName.StaneksGift1);
 
+  const clickPage = useCallback(
+    (page: Page) => {
+      if (page === Page.Job) {
+        Router.toJob(Locations[Object.keys(Player.jobs)[0]]);
+      } else if (page == Page.ScriptEditor) {
+        Router.toScriptEditor();
+      } else if ((Object.values(SimplePage) as Page[]).includes(page)) {
+        Router.toPage(page as SimplePage);
```
Snarling: ~~There's an "isSimplePage" function for this check, which also removes the need to do page `as` SimplePage~~ Disregard, it looked like this was new code but I see it was existing code at a different part of the file.

Snarling:
After reviewing and doing some tests, I'm pretty satisfied that this shouldn't break anything.

I'm not the hugest fan of some of the implementation (lots of refs that are just used for a one-time effect and that's it), but I think most of the issue is just the way react hooks work (they're not really meant to just be a drop-in replacement for the old lifecycle functions of class components).

If you're relatively satisfied with it I think we can just go ahead and merge it. Later PRs can touch on improvements if we find a better way to handle any of the stuff that had weird implementations here.

bezrodnov:
I agree that this is quite dirty PR in a few places. But if it does not break stuff - we should merge it so that we can start using eslint react rules and detect potential issues earlier as soon as possible.

As for refs - my vision: is that is mostly caused by the fact that we cannot wrap some parts of the code in `useCallback`/`useMemo`. Usually, when you want to avoid `useEffect` triggering in specific cases, you place related dependency in `useMemo`/`useCallback` and try to minimize the number of times it is re-evaluated.
The game state we use in react components is not that much compatible with React style - it can change on every render, and it cannot be used inside `useMemo` or `useCallback` (unless we reference a primitive value).

Also in some cases you just have to use refs. E.g. let's say you have a component where in `useEffect` you attach a listener to some dom element and this listener will be using a variable from the component (like state or prop). If you want to use the actual value of this variable you have a few options 
- add it to dependency so that old listener is de-attached and a new one is added every time this variable changes
- use react ref in the listener and update that ref in component code.

Quite often the first option is used and it's fine. But if the operation in the `useEffect` is costly and you want to optimize it a bit - you can use refs.

One of the in-game examples for this is a TextEditor where `useEffect` is adding some listeners when `vim` mode gets toggled on. We're using refs there currently, but we could split that effect into two parts - one just to init `vim` mode and another to add listeners and add to the second effect all the dependencies. This way we won't have to use refs, but we'll have to use 2 useEffects (cause the first one is calling some extra functions to configure vim).

src/Infiltration/ui/SlashGame.tsx
```diff
@@ -25,48 +25,52 @@ const difficulties: {
...
-      props.onSuccess();
+      onSuccess();
     }
   }
-  const hasAugment = Player.hasAugmentation(AugmentationName.MightOfAres, true);
-  const guardingTime = Math.random() * 3250 + 1500 - (250 + difficulty.window);
-  const preparingTime = difficulty.window;
-  const attackingTime = 250;
+
+  const guardingTimeRef = useRef(Math.random() * 3250 + 1500 - (250 + difficulty.window));
```
bezrodnov, line 44: this is one of the places where you can barely get rid of `useRef` - here variable is supposed to be generated only once, during the first render iteration. And then it is passed down both to child component and to `useEffect` (so we can't move it inside useEffect).
An alternative could be React state, but to me it sounds even worse as state is something you expect to change at some point. If you used React class-based component, you would declare a field in this class and initialize it in the constructor (cause you want this value to be set only once). That's effectively the same as using ref in function components.

The only problem I see here is that if the same game could be used twice in a row - this ref won't be recalculated leading to the same guard time value... But this PR does not change this part, runtime result should be the same as before.

src/Infiltration/ui/WireCuttingGame.tsx
```diff
@@ -51,49 +52,50 @@ interface Question {
...
+  const wiresRef = useRef(generateWires(difficulty));
+  const questionsRef = useRef(generateQuestion(wiresRef.current, difficulty));
 
-  function checkWire(wireNum: number): boolean {
-    return questions.some((q) => q.shouldCut(wires[wireNum - 1], wireNum - 1));
-  }
+  const [cutWires, setCutWires] = useState(new Array(wiresRef.current.length).fill(false));
+  const hasAugment = Player.hasAugmentation(AugmentationName.KnowledgeOfApollo, true);
 
   useEffect(() => {
```
bezrodnov, line 71: this whole effect could be rewritten to a regular callback.
E.g. this code could be moved to the `press` function which sets the `cutWires` state. If it also checked if the player won, we would not have to use this effect.

src/Programs/ui/ProgramsRoot.tsx
```diff
@@ -33,8 +33,9 @@ export function ProgramsRoot(): React.ReactElement {
       return (a.create?.level ?? 0) - (b.create?.level ?? 0);
     });
 
+  const programsRef = useRef(programs);
   useEffect(() => {
-    programs.forEach((p) => {
+    programsRef.current.forEach((p) => {
       if (ProgramsSeen.includes(p.name)) return;
       ProgramsSeen.push(p.name);
```
bezrodnov: I've just taken a closer look at this effect... what is it doing and why?
`ProgramsSeen` is only used inside of this `useEffect` - it does not make any sense to me.
I'll drop it.

bezrodnov: Ah, it's exported... nvm

src/Infiltration/ui/SlashGame.tsx
d0sboots: > Also in some cases you just have to use refs.

Yeah, there are a lot of new refs that I'm totally fine with. Honestly, due to the way our components work there's some that should be using less state and more refs, even though that's backwards for React.

What I was questioning was spots where a ref was introduced and never changed from its default value. That's just misuse IMO, created to appease the lint. (Likely the useEffect it's satisfying is also bad/weird.)

Snarling: > The only problem I see here is that if the same game could be used twice in a row - this ref won't be recalculated leading to the same guard time value... But this PR does not change this part, runtime result should be the same as before.

On that point, infiltration alternates between a game and a countdown timer, so the old game has already been totally unmounted by the time the next game is rendered. Even if that gets changed at some point, the game remembers the last game played and doesn't do the same game twice in a row. So each new game is always a new component being rendered, not just a component from last game getting updated.

> What I was questioning was spots where a ref was introduced and never changed from its default value.

The question is, what should be used in those cases? Previously in these cases we were using state, and just never using the setState function. Both state and ref in these cases doesn't really seem right. But what to use for something that is calculated within the component on first render, never changes, and can't be contained within a useEffect?

Edit: useMemo sounds like it would be good for this actually.

d0sboots: Or just compute the value outside and capture it.
If it depends on props, then it's a bit sus (and useMemo probably calls that out the best)

PR #644, opened 2023-06-23
TYPESAFETY: FactionName

Snarling:
* Factions only initialized once, with no init function needed
* Enforce FactionName enum type on most instances of faction names
* Fully safe loading of faction data from player save (see Factions.ts changes)


src/Faction/Factions.ts
```diff
@@ -2,63 +2,47 @@
...
+      faction.favor = loadedFaction.favor;
     }
+    if (
+      "playerReputation" in loadedFaction &&
+      typeof loadedFaction.playerReputation === "number" &&
+      loadedFaction.playerReputation > 0
+    ) {
+      faction.playerReputation = loadedFaction.playerReputation;
+    }
+    if ("alreadyInvited" in loadedFaction && loadedFaction.alreadyInvited) faction.alreadyInvited = true;
```
d0sboots: This entire set of checks would be much less clunky if you did a cast earlier. All the "in" tests are just there to satisfy TS; we're perfectly fine doing a Boolean test against undefined.
I think you want to cast it to a `Record<string, unknown>` instead.

d0sboots (commented):
Nice

src/Faction/Factions.ts
Snarling: It's also clunky finding a good place to do that casting. I can't do it in the for statement because at that point, it could also be `undefined` or `null` (well really just `null` because `undefined` wouldn't be returned by parsing JSON, but ts doesn't know that), and checking a property on something nullish throws an error.

I think I'll make an assertion function that takes in an object and just asserts that it's a better type that's still safe (either Record<string, unknown> or maybe an Unknownify<DesiredType> so that we get autocomplete on keys). Having a reusable function will be very helpful especially if we're going to implement more fully-typesafe loading. (added in 5c76447)

Snarling:
Probably going to add a new test in a separate PR before merging this one.

Edit: added a savegame continuity test in #645. Updated the test in this PR and updated the snapshot to show the changes to faction save.

src/Faction/Factions.ts
d0sboots: > I can't do it in the for statement because at that point, it could also be undefined or null

Huh? By that point, you'd already checked that `loadedFaction` was truthy, so it couldn't be undefined or null. You'd also checked that it was an object. A cast would have been totally safe.

If you want typesafe loading, we'll need a very different structure than what we currently have.

Snarling: The for loop is when loadedFaction is first declared, I have not checked anything about loadedFaction at that point.

I've checked that loadedFactions (the parent object, which is expected to be in the structure of a Record<FactionName, Faction>) is an object, because it makes sense to do that before attempting to iterating through its entries. But those entries have totally unknown values so I cannot assign a type other than unknown inside the for statement.

d0sboots: I feel like I'm taking crazy pills.

https://github.com/bitburner-official/bitburner-src/pull/644/commits/0552f73ee05af90e45c96752ebf71e1c0bc8d9c8#diff-796704400cb6ec180a36942fe55e6af741f2d8aa3f8303084324e468028ed4d5

src/Factions/Factions.ts:31
`if (!loadedFaction) continue;`
Line 33:
`if (typeof loadedFaction !== "object") continue;`

The code has changed since then, but in the context of how it was when this thread started, the checks were there already.

Snarling: That is after the declaration in the for statement. I would have to do a reassignment to give it another type (or do an assertion function like I did)

Snarling: I was just saying that it's a pain to do a manual type reassignment at a time other than when you're declaring the variable (in the for statement declaration, when the correct value is unknown). There are basically these options to reassign the type after-the-fact:

1. Do an assertion function (which I ended up doing) to convert the type of the existing variable
2. Not actually reassign the type, but do `as` statements every time you reference the variable. We would have had a bunch of statements like `if ((loadedFaction as Unknownify<Faction>).isMember) faction.isMember = true`.
3. Assign a new variable with the desired type (e.g. `newLoadedFaction = loadedFaction as Unknownify<Faction>;`) and use the new variable that has the desired type.

Snarling: > If you want typesafe loading, we'll need a very different structure than what we currently have.

I think it's reasonable for things like Factions / Companies / StockMarket where the majority of the data is static and an exact structure can already be generated by the game without needing to load it. In those cases we can just generate the structure statically and then load in any valid info contained in the save.

But yeah pretty much anything else, not going to be so easy, and I think in those other cases we would want to stop the game from loading when we find an error. We could still use a typesafe approach to loading in order to not load invalid data and instead tell the player what is wrong with their save (e.g. a server had a non-string name, or stuff like that), instead of now where we are just loading the invalid data and things may end up with incorrect types and that can cause some random error later on.

d0sboots: > There are basically these options to reassign the type after-the-fact:

There is also option #4, which is to reassign the type before-the-fact, and then check after. It's kinda dirty, but it *is* an option. I was actually assuming option 3 in this particular case, but that's because I was talking about a minor change. Using a type assertion function makes more overall sense, especially when it can be reused.

> But yeah pretty much anything else, not going to be so easy, and I think in those other cases we would want to stop the game from loading when we find an error.

This is basically all my comment was about. Right now we're blithely loading the JSON and calling it good. To do it right requires either a lot of copying and checking, or validating it as we deserialize. The former is probably more robust to our format changes. Either way, it's a lot of extra code. Probably worth doing to prevent weird errors (especially given that there's an achievement for explicitly modifying the save file), but still a bunch of work.

PR #645, opened 2023-06-24
TEST: Add savegame continuity test

Snarling:
This should be useful for flagging unintended modifications to save file, and monitoring the impact of other PRs that affect the save file intentionally


Snarling:
This is probably not written using jest best practices (I imagine a real jest pro would be using Mocks and beforeAll instead of just an initial setup function that gets ran before the tests)

PR #648, opened 2023-06-26
Update bitburner.gang.getascensionresult.md

KrunoSaho:
CATEGORY: DOC

This is unclear, but it was made clear to me in Discord and also in NetscriptDefinitions.d.ts. I am updating this to ensure no one spends an unnecessary amount of time dealing with delta uncertainty.

### DIFF
--- a/markdown/bitburner.gang.getascensionresult.md
+++ b/markdown/bitburner.gang.getascensionresult.md
@@ -4,7 +4,7 @@
 
 ## Gang.getAscensionResult() method
 
-Get the result of an ascension without ascending.
+Get the result of an ascension without ascending. Theses multipliers are gains made from the possible ascension; present_stat + ascension_result_stat = new_stat.
 
 **Signature:**
 


Snarling:
1. The .md files don't get edited manually themselves, they are autogenerated from NetscriptDefinition.d.ts
2. The extra text is not at all clear to me and I think is probably inaccurate. There is no addition of stats. Ascending adds ascension points to the member which results in a new, higher multiplier. getAscensionResult tells you how much each stat's ascension multiplier is multiplied by (here is the internal code for what is being returned to you, it is the multiplicative factor by which your ascension multiplier will increase):
https://github.com/bitburner-official/bitburner-src/blob/9a0a843ffc507a9d11d67d97d0010ff608df5c41/src/Gang/GangMember.ts#L239-L251

KrunoSaho:
Thank you. It seems like I have been fairly confused by this.

Could your explanation be used as the documentation for that function?

Snarling:
See #649

PR #650, opened 2023-06-26
TYPESAFETY: CompanyName

Snarling:
* Uses a CompanyName enum instead of string everywhere where CompanyName is used.
* Companies are now statically initialized, and only favor/reputation are loaded in from/saved to the player save.
* If a CompanyWork (for sleeve or player) is in progress and has invalid loading data, it is loaded in as null (see Work/InvalidWork.ts).
* There were two unused Police jobs that also were named as if they were on the same job track as other security jobs. These have been removed, if they get reimplemented in the future as a separate job track, they would need to actually be separate.
* Initial reputation with a company starts at 0 instead of 1, just like faction.


src/Company/Enums.ts
```diff
@@ -0,0 +1,47 @@
+import type { Member } from "../types";
+
+// Imported from Locations and not from the main enums file to avoid circular dependency
+import { LocationName } from "../Locations/Enums";
+
+// CompanyName is essentially a sub-enum of LocationName
+export const CompanyName = {
```
bezrodnov: why not enum?
I think something like this is possible:

```ts
export enum CompanyName {
  ECorp = LocationName.AevumECorp,
  MegaCorp = LocationName.Sector12MegaCorp,
  BachmanAndAssociates = LocationName.AevumBachmanAndAssociates,
  BladeIndustries = LocationName.Sector12BladeIndustries,
  NWO = LocationName.VolhavenNWO,
  ClarkeIncorporated = LocationName.AevumClarkeIncorporated,
  OmniTekIncorporated = LocationName.VolhavenOmniTekIncorporated,
  FourSigma = LocationName.Sector12FourSigma,
  KuaiGongInternational = LocationName.ChongqingKuaiGongInternational,
  FulcrumTechnologies = LocationName.AevumFulcrumTechnologies,
  StormTechnologies = LocationName.IshimaStormTechnologies,
  DefComm = LocationName.NewTokyoDefComm,
  HeliosLabs = LocationName.VolhavenHeliosLabs,
  VitaLife = LocationName.NewTokyoVitaLife,
  IcarusMicrosystems = LocationName.Sector12IcarusMicrosystems,
  UniversalEnergy = LocationName.Sector12UniversalEnergy,
  GalacticCybersystems = LocationName.AevumGalacticCybersystems,
  AeroCorp = LocationName.AevumAeroCorp,
  OmniaCybersystems = LocationName.VolhavenOmniaCybersystems,
  SolarisSpaceSystems = LocationName.ChongqingSolarisSpaceSystems,
  DeltaOne = LocationName.Sector12DeltaOne,
  GlobalPharmaceuticals = LocationName.NewTokyoGlobalPharmaceuticals,
  NovaMedical = LocationName.IshimaNovaMedical,
  CIA = LocationName.Sector12CIA,
  NSA = LocationName.Sector12NSA,
  WatchdogSecurity = LocationName.AevumWatchdogSecurity,
  LexoCorp = LocationName.VolhavenLexoCorp,
  RhoConstruction = LocationName.AevumRhoConstruction,
  AlphaEnterprises = LocationName.Sector12AlphaEnterprises,
  Police = LocationName.AevumPolice,
  SysCoreSecurities = LocationName.VolhavenSysCoreSecurities,
  CompuTek = LocationName.VolhavenCompuTek,
  NetLinkTechnologies = LocationName.AevumNetLinkTechnologies,
  CarmichaelSecurity = LocationName.Sector12CarmichaelSecurity,
  FoodNStuff = LocationName.Sector12FoodNStuff,
  JoesGuns = LocationName.Sector12JoesGuns,
  OmegaSoftware = LocationName.IshimaOmegaSoftware,
  NoodleBar = LocationName.NewTokyoNoodleBar,
}
```

d0sboots: There's technical reasons we're moving away from enums. Snarling might remember the whole rationale.

Snarling: Yeah there was a lot of conversation on "enum bad", but in the end we are sticking with enums and I think they are good (at least the type of enums we usually use, with string values).

I'm not sure why, but I thought we were not able to use computed property values in an enum, but it's just the keys that aren't allowed to be computed values. So yeah I will probably just make it a normal enum. The only advantage of the current method is that ts knows that CompanyNames are also LocationNames, so the company name can also be used as a location name without typechecking/assertions.

Snarling: Changed to enum in 1f224e5

src/Company/Companies.ts
```diff
@@ -1,46 +1,30 @@
...
-}
+export const Companies: Record<CompanyName, Company> = (() => {
+  const metadata = getCompaniesMetadata();
+  return createEnumKeyedRecord(CompanyName, (name) => new Company(metadata[name]));
+})();
 
 // Used to load Companies map from a save
 export function loadCompanies(saveString: string): void {
-  Companies = JSON.parse(saveString, Reviver);
+  const loadedCompanies = JSON.parse(saveString, Reviver) as unknown;
```
bezrodnov, line 16: @Snarling have you considered using [ts-reset](https://github.com/total-typescript/ts-reset)?

Snarling: Yeah most of the changes in that seem to be improvements on the ts defaults. Unfortunately I think there are several places we are currently abusing the default `any` type and would need significant rewrite to avoid needing an `as any`.

Although maybe it would be better to bring in ts-reset (at least a few select rules from it) and just add `as any` in the places where it was needed, which also would essentially flag the areas for us to address later. Given that the entire loading process is based on JSON.parse, we would basically need fully typesafe loading for all parts of the save file (currently I believe it is only Aliases, Factions, and now with this PR Companies that use typesafe loading instead of just loading the JSON and hoping it's the right structure).

bezrodnov: Yeah, I think it's better to add `as any` in places that are not ready yet than finding and marking all vulnerable places `as unknown`.

src/Company/Company.ts
```diff
@@ -53,26 +37,25 @@ export class Company {
...
     this.salaryMultiplier = p.salaryMultiplier;
     this.jobStatReqOffset = p.jobStatReqOffset;
 
-    this.isPlayerEmployed = false;
     this.playerReputation = 1;
     this.favor = 0;
-    this.isMegacorp = false;
-    if (p.isMegacorp) this.isMegacorp = true;
+    this.hasFaction = false;
+    if (p.hasFaction) this.hasFaction = true;
```
bezrodnov:
```suggestion
    this.hasFaction = !!p.hasFaction;
```

src/Company/utils.ts
```diff
@@ -0,0 +1,11 @@
+import type { CompanyName, LocationName } from "@enums";
+
+type LocationNameString = `${LocationName}`;
+type CompanyNameString = `${CompanyName}`;
+type CompanyNamesAreAllLocationNames = CompanyNameString extends LocationNameString ? true : false;
+const __companyNameCheck: CompanyNamesAreAllLocationNames = true;
```
bezrodnov, line 6: lol, someone deserved a medal for this xD

image

Snarling: I initially just had a top-level for loop that was manually checking that all CompanyName members are also LocationName members, but doing the check at a type level made a lot more sense, since it avoids imports and also we don't have to build the game to figure out there's an issue.

It is pretty jank looking though.

src/Company/Company.ts
Snarling: Yeah I hadn't really messed with this code, it just looked like modified code because of a variable rename (isMegaCorp was a bad name, especially with a company/faction actually named Megacorp in the game). In progress commit addresses this.

src/Company/Companies.ts
Snarling: Will look into adding this in a separate PR after this one.

PR #651, opened 2023-06-27
CODEBASE: Follow-up for lint rules PR + address some eslint disables

Snarling:
This is mostly a follow up to #640, there were just a couple of weird instances of useRef in infiltration games that were used to escape a lint rule error, but didn't seem like correct usage. This PR was originally using useMemo to replace them, but that wasn't a great solution either. The two games have been refactored so that these per-game-instance variables are state and are initialized inside an initializer useEffect that triggers on prop changes, instead of being initialized once per mount.

The current implementation is such that each game is unmounted and then a new game is remounted later, and we never have an existing component receiving updated props as a way of initiating the next game. Relying on this behavior is not great design, especially with a lint rule already forcing some of our code (initializer useEffects) to trigger based on props updates.

I also addressed several eslint disable statements, mostly ones that were already existing before #640. Detailed info is in the commit details for the last block of commits.

Original description:
~~Based on my understanding, useMemo seems like a more correct way to store a value that is not expected to change during the lifetime of a component. useRef should be used when value is expected to change, but the change should not trigger a rerender.~~

~~... list of initial changes~~
~~@d0sboots @bezrodnov Thoughts? This is just a one-file test for now.~~

### DIFF
--- a/src/Infiltration/ui/SlashGame.tsx
+++ b/src/Infiltration/ui/SlashGame.tsx
@@ -1,5 +1,5 @@
 import { Box, Paper, Typography } from "@mui/material";
-import React, { useEffect, useRef, useState } from "react";
+import React, { useEffect, useMemo, useState } from "react";
 import { AugmentationName } from "@enums";
 import { Player } from "@player";
 import { KEY } from "../../utils/helpers/keyCodes";
@@ -25,11 +25,34 @@ const difficulties: {
   Impossible: { window: 150 },
 };
 
-export function SlashGame({ difficulty: _difficulty, onSuccess, onFailure }: IMinigameProps): React.ReactElement {
-  const difficulty: Difficulty = { window: 0 };
-  interpolate(difficulties, _difficulty, difficulty);
+export function SlashGame({ difficulty, onSuccess, onFailure }: IMinigameProps): React.ReactElement {
+  const timingWindow = useMemo(() => {
+    const out: Difficulty = { window: 0 };
+    interpolate(difficulties, difficulty, out);
+    return out.window;
+  }, [difficulty]);
+
+  const guardingTime = useMemo(() => Math.random() * 3250 + 1500 - (250 + timingWindow), [timingWindow]);
+  const hasAugment = useMemo(() => Player.hasAugmentation(AugmentationName.MightOfAres, true), []);
 
   const [phase, setPhase] = useState(0);
+
+  useEffect(() => {
+    const preparingTime = timingWindow;
+    const attackingTime = 250;
+
+    let id = setTimeout(() => {
+      setPhase(1);
+      id = setTimeout(() => {
+        setPhase(2);
+        id = setTimeout(() => onFailure(), attackingTime);
+      }, preparingTime);
+    }, guardingTime);
+
+    return () => {
+      clearInterval(id);
+    };
+  }, [timingWindow, onFailure, guardingTime]);
 
   function press(this: Document, event: KeyboardEvent): void {
     event.preventDefault();
@@ -41,26 +64,6 @@ export function SlashGame({ difficulty: _difficulty, onSuccess, onFailure }: IMi
     }
   }
 
-  const guardingTimeRef = useRef(Math.random() * 3250 + 1500 - (250 + difficulty.window));
-
-  useEffect(() => {
-    const preparingTime = difficulty.window;
-    const attackingTime = 250;
-
-    let id = window.setTimeout(() => {
-      setPhase(1);
-      id = window.setTimeout(() => {
-        setPhase(2);
-        id = window.setTimeout(() => onFailure(), attackingTime);
-      }, preparingTime);
-    }, guardingTimeRef.current);
-
-    return () => {
-      clearInterval(id);
-    };
-  }, [difficulty.window, onFailure]);
-
-  const hasAugment = Player.hasAugmentation(AugmentationName.MightOfAres, true);
   return (
     <>
       <GameTimer millis={5000} onExpire={onFailure} />
@@ -70,7 +73,7 @@ export function SlashGame({ difficulty: _difficulty, onSuccess, onFailure }: IMi
         {hasAugment ? (
           <Box sx={{ my: 1 }}>
             <Typography variant="h5">Guard will drop in...</Typography>
-            <GameTimer millis={guardingTimeRef.current} onExpire={() => null} ignoreAugment_WKSharmonizer noPaper />
+            <GameTimer millis={guardingTime} onExpire={() => null} ignoreAugment_WKSharmonizer noPaper />
           </Box>
         ) : (
           <></>
--- a/src/Infiltration/ui/SlashGame.tsx
+++ b/src/Infiltration/ui/SlashGame.tsx
@@ -31,14 +31,19 @@ export function SlashGame({ difficulty, onSuccess, onFailure }: IMinigameProps):
   const [guardingTime, setGuardingTime] = useState(0);
 
   useEffect(() => {
+    // Determine timeframes for game phase changes
     const newDifficulty: Difficulty = { window: 0 };
     interpolate(difficulties, difficulty, newDifficulty);
     const timePreparing = newDifficulty.window;
     const timeAttacking = 250;
     const timeGuarding = Math.random() * 3250 + 1500 - (timeAttacking + timePreparing);
+
+    // Set initial game state
+    setPhase(0);
     setGuardingTime(timeGuarding);
     setHasAugment(Player.hasAugmentation(AugmentationName.MightOfAres, true));
 
+    // Setup timer for game phases
     let id = setTimeout(() => {
       setPhase(1);
       id = setTimeout(() => {
@@ -65,7 +70,6 @@ export function SlashGame({ difficulty, onSuccess, onFailure }: IMinigameProps):
       <GameTimer millis={5000} onExpire={onFailure} />
       <Paper sx={{ display: "grid", justifyItems: "center" }}>
         <Typography variant="h4">Attack when his guard is down!</Typography>
-
         {hasAugment && (
           <Box sx={{ my: 1 }}>
             <Typography variant="h5">Guard will drop in...</Typography>
--- a/src/Infiltration/ui/SlashGame.tsx
+++ b/src/Infiltration/ui/SlashGame.tsx
@@ -1,5 +1,5 @@
 import { Box, Paper, Typography } from "@mui/material";
-import React, { useEffect, useMemo, useState } from "react";
+import React, { useEffect, useState } from "react";
 import { AugmentationName } from "@enums";
 import { Player } from "@player";
 import { KEY } from "../../utils/helpers/keyCodes";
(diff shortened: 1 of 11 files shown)


bezrodnov:
> useMemo seems like a more correct way to store a value that is not expected to change during the lifetime of a component

I don't think it's actually true - useMemo semantically is more like an optimization technic where you want to calculate a value but not on every re-render, only when dependencies in the useMemo change. And usually, at least in my practice, useMemo has every "mutable" value listed in the dependency list. O/w you again have to ignore/disable eslint rule and probably are doing something wrong.

As per official docs that's when useMemo should be used:
- Skipping expensive recalculations
- Skipping re-rendering of components
- Memoizing a dependency of another Hook
- Memoizing a function
In my opinion, none of this really falls under our use case (the third one is somewhat close, but again, not a perfect use imo).

bezrodnov:
A good read on this topic: https://blog.logrocket.com/when-not-to-use-usememo-react-hook
An example from this article:

> I agree with Dan Abramov [on this one](https://github.com/facebook/create-react-app/issues/6880#issuecomment-485912528). Suppressing the eslint-warnings from the plugin will likely come back to bite you someday in the future.

> That’s a rather brash message, but the linter is simply doing its job. The easy solution is to use a eslint-disable comment, but this isn’t always the best solution because you could introduce bugs within the same useEffect call in the future:
```tsx
useEffect(() => {
  impressionTracker(propA, propB, propC)
  // eslint-disable-next-line react-hooks/exhaustive-deps
}, [])
```
> My suggestion solution is to use the useRef Hook to keep a reference to the initial prop values you don’t need updated:
```tsx
function Example({impressionTracker, propA, propB, propC}) {
  // keep reference to the initial values         
  const initialTrackingValues = useRef({
      tracker: impressionTracker, 
      params: {
        propA, 
        propB, 
        propC, 
    }
  })

  // track impression 
  useEffect(() => {
    const { tracker, params } = initialTrackingValues.current;
    tracker(params)
  }, []) // you get NO eslint warnings for tracker or params

  return <BeautifulComponent propA={propA} propB={propB} propC={propC} />   
}
```
> In all my tests, the linter only respects useRef for such cases. With useRef, the linter understands that the referenced values won’t change and so you don’t get any warnings! Not even useMemo prevents these warnings.
```

**Comment by bezrodnov** [comment]
`useMemo` is even mentioned in the performance section: https://react.dev/reference/react#performance-hooks .
This is not the case for us.

And from the same documentation: `Refs let a component hold some information that isn’t used for rendering`

File: src/Infiltration/ui/SlashGame.tsx [hunk 1]
```diff
@@ -25,42 +25,45 @@ const difficulties: {
   Impossible: { window: 150 },
 };
 
-export function SlashGame({ difficulty: _difficulty, onSuccess, onFailure }: IMinigameProps): React.ReactElement {
-  const difficulty: Difficulty = { window: 0 };
-  interpolate(difficulties, _difficulty, difficulty);
+export function SlashGame({ difficulty, onSuccess, onFailure }: IMinigameProps): React.ReactElement {
+  const timingWindow = useMemo(() => {
```
bezrodnov: here `useMemo` totally makes sense to me 👍 
Especially if underlying operation is expensive. But if it's something simple, it's usually not recommended to use memo (unless it's an object-like structure whose new reference is passed to another hook as a dependency every render iteration).

src/Infiltration/ui/SlashGame.tsx
```diff
@@ -70,7 +73,7 @@ export function SlashGame({ difficulty: _difficulty, onSuccess, onFailure }: IMi
         {hasAugment ? (
```
bezrodnov: nit: this could be `&&` instead of `?` (and without `: <></>` afterwards)

src/Infiltration/ui/SlashGame.tsx
bezrodnov: Thinking more about it - `useMemo` feels a bit unnecessary here: I checked how `interpolate` works and it's pretty simple.

bezrodnov:
There's even a section `useMemo vs. useRef` in the article I shared above 😀
And it basically tells the same as I mentioned before - it's semantically wrong to use memo when the only thing you want is to store a value that never changes between re-renders.

d0sboots:
I think I've identified what bugs me about useRef in situations like this. It probably *is* the right answer if you really need to do it (more than useMemo or disabling the rule for the effect), but most of the time it's still a smell.

Take this code, for instance. We're creating a ref to pass the time in to an effect. Why can't the time be local to the effect, since the effect will only run once? Well, because we also need the time to display on a UI timer.
But the UI timer has a function for when the time finishes, which is being ignored! Using that would allow is to eliminate the effect, entirely, as well as the need for a ref.

bezrodnov:
> But the UI timer has a function for when the time finishes, which is being ignored! Using that would allow is to eliminate the effect, entirely, as well as the need for a ref.

First of all, this timer is rendered conditionally, so we can't rely on it's `onExpire` callback.
Also this `useEffect` creates 3 timeouts, which would be hard to achieve with a GameTimer component.

Of course, we could use an approach with a hidden timer, pass down `onExpire` handler depending on the current `phase`... but I'm not sure if it look better and will be easy to understand. A hidden component is also a bad idea/smell - it should be a hook in this case instead.

bezrodnov:
it could look like this (I did not test it though):
```tsx
// can be moved to ui/hooks out of this file
/**
 * Executes provided `callback` function after the specified `ms` timeout in milliseconds.
 * Re-starts timeout whenever `callback` or `ms` changes.
 */
const useTimeout = (callback: () => void, ms: number) => {
  const timeoutRef = useRef<NodeJS.Timeout>();

  useEffect(() => {
    timeoutRef.current = setTimeout(() => {
      callback();
      timeoutRef.current = undefined;
    }, ms);

    return () => timeoutRef.current && clearTimeout(timeoutRef.current);
  }, [callback, ms]);
};

const calculateTimeWindow = (difficulty: number) => {
  const out: Difficulty = { window: 0 };
  interpolate(difficulties, difficulty, out);
  return out.window;
};

export function SlashGame({ difficulty, onSuccess, onFailure }: IMinigameProps): React.ReactElement {
  const timeWindowRef = useRef(calculateTimeWindow(difficulty));

  const [gameStage, setGameStage] = useState(() => ({
    phase: 0,
    duration: Math.random() * 3250 + 1500 - (250 + timeWindowRef.current),
  }));

  const onPhaseOver = useCallback(() => {
    if (gameStage.phase === 2) {
      onFailure();
    }

    setGameStage((prev) => {
      const attackingTime = 250;
      const preparingTime = timeWindowRef.current;
      const duration = prev.phase === 0 ? attackingTime : preparingTime;

      return { phase: prev.phase + 1, duration };
    });
  }, [gameStage, onFailure]);

  useTimeout(onPhaseOver, gameStage.duration);

  function press(this: Document, event: KeyboardEvent): void {
    event.preventDefault();
    if (event.key !== KEY.SPACE) return;
    if (gameStage.phase !== 1) {
      onFailure();
    } else {
      onSuccess();
    }
  }

  const hasAugment = Player.hasAugmentation(AugmentationName.MightOfAres, true);
  return (
    <>
      <GameTimer millis={5000} onExpire={onFailure} />
      <Paper sx={{ display: "grid", justifyItems: "center" }}>
        <Typography variant="h4">Attack when his guard is down!</Typography>

        {hasAugment && gameStage.phase === 0 && (
          <Box sx={{ my: 1 }}>
            <Typography variant="h5">Guard will drop in...</Typography>
            <GameTimer millis={gameStage.duration} onExpire={() => null} ignoreAugment_WKSharmonizer noPaper />
          </Box>
        )}

        {gameStage.phase === 0 && <Typography variant="h4">Guarding ...</Typography>}
        {gameStage.phase === 1 && <Typography variant="h4">Preparing?</Typography>}
        {gameStage.phase === 2 && <Typography variant="h4">ATTACKING!</Typography>}
        <KeyHandler onKeyDown={press} onFailure={onFailure} />
      </Paper>
    </>
  );
}
```

bezrodnov:
is it really better, cleaner, more readable? I don't know

bezrodnov:
BTW, `KeyHandler` should really be rewritten to hook - it's not a UI component but a re-usable effect-like function/hook.

d0sboots:
K, maybe in this case a ref really is best then. I think you get my point, though: this whole component feels like it's designed a little wrong, and if it were more "right" this issue might go away.
I'm not looking deeply enough at it to get a sense of what "right" would look like, though - It's just pinging my "something is off here" senses.

src/Infiltration/ui/SlashGame.tsx
```diff
@@ -70,7 +73,7 @@ export function SlashGame({ difficulty: _difficulty, onSuccess, onFailure }: IMi
...
           <Box sx={{ my: 1 }}>
             <Typography variant="h5">Guard will drop in...</Typography>
-            <GameTimer millis={guardingTimeRef.current} onExpire={() => null} ignoreAugment_WKSharmonizer noPaper />
+            <GameTimer millis={guardingTime} onExpire={() => null} ignoreAugment_WKSharmonizer noPaper />
```
Snarling: Not convinced that ref is the answer here.

Both guardingTime and hasAugment are:
1. Used during rendering
2. Should not be recalculated multiple times for the same instance of the game (not a big deal for hasAugment)

They should not be useRef because of point 1. From the react documentation: `useRef is a React Hook that lets you reference a value that’s not needed for rendering.`

There is also a pitfall section, stating: `Do not write or read ref.current during rendering.` We are reading the values of both of these during rendering. Since the current value of both of these is needed during rendering and not just in effects and event listeners, they should not be refs.

Since they should not be recalculated every render, they should not be plain variables inside the function body, they need to be kept stable between renders in some manner. Since refs are off the table, the remaining options are useMemo and useState. State might make more sense here, with the game initializer useEffect performing the state update. timingWindow can be entirely contained within the useEffect. I'll post an updated commit.

src/Infiltration/ui/SlashGame.tsx
```diff
@@ -25,11 +25,30 @@ const difficulties: {
...
+    let id = setTimeout(() => {
+      setPhase(1);
+      id = setTimeout(() => {
+        setPhase(2);
+        id = setTimeout(() => onFailure(), timeAttacking);
+      }, timePreparing);
+    }, timeGuarding);
+
+    return () => clearTimeout(id);
+  }, [difficulty, onSuccess, onFailure]);
```
Snarling: So when we added variables to the dependency array of the useEffect that initializes the game timer, we are basically saying that changing props on the component after it's already mounted is a reinitialization of the game.

However, while we made the changes in the useEffect so that prop changes will reinitialize the game timer, other calculated variables were still set up so that they are only initialized once per mount. So the game timer initializes once per props update, but the relevant calculated values only recalculate on unmount/remount, the component is really not able to fully reinitialize the game on props update.

So what I've done here is move the initializers for these calculated values into the useEffect, so they will all be recalculated on a props change. This means that, even though infiltration is currently written in a way that it will always unmount/remount for each game because of the timer displayed between games, this SlashGame component is robust enough that it can actually handle being displayed twice in a row without needing a fresh mount to initialize a new game.

(Actually I need to setPhase back to 0 during the useEffect as well, will add another commit for that)

d0sboots:
This feels much more like the correct direction to me.

Snarling:
Made similar changes in the wire game, plus a little more of a rewrite to address an old TODO/fix some inefficiency. I think those were really the only places with questionable refs from the lint rule PR

PR #664, opened 2023-07-08
FIX: Added checking export statements for renames

G4mingJon4s:
# Changes
The ram evaluation now includes any renames.

Before, exports like these were not included in the ram evaluation, leading to a wrong static ram count:

```ts
function expensiveFunction(ns) {
  ns.corporation.bribe();
}

export { expensiveFunction as myExportedFunction }
```

Now, the acorn walker checks if the export is an explicit export and adds the correct reference to the dependencies.

# Actions
- [X] lint
- [X] format
- [X] test


Snarling:
Very nice!

PR #665, opened 2023-07-09
QUICKFIX: Fixed ram evaluation to include more edge-cases

G4mingJon4s:
# Changes

I did some more testing regarding named exports. My last implementation had some big issues.
The code now no longer adds the entire imported file to the ram count, but only the actually imported function.
The code also accounts for situations, where the export statement comes before the actual declaration.

# Examples
```ts
import { expensiveFunction } from "expensive";

  const pos = expensiveFunction(ns);
}

// ---------------------
// expensive.js

export { a as expensiveFunction }
// before, because this comes before the declaration,
// this would reference undefined, instead of the actual dependencies

function a(ns) {
  return ns.stock.getPosition("ECP");
}

function b(ns) { // this function was included before
  ns.corporation.bribe();
}
```

# Actions
- [X] lint
- [X] format
- [X] test

### DIFF
--- a/src/Script/RamCalculations.ts
+++ b/src/Script/RamCalculations.ts
@@ -270,13 +270,13 @@ function parseOnlyCalculateDeps(code: string, currentModule: string): ParseDepsR
 
   // References get added pessimistically. They are added for thisModule.name, name, and for
   // any aliases.
-  function addRef(key: string, name: string): void {
+  function addRef(key: string, name: string, module = currentModule): void {
     const s = dependencyMap[key] || (dependencyMap[key] = new Set());
     const external = internalToExternal[name];
     if (external !== undefined) {
       s.add(external);
     }
-    s.add(currentModule + "." + name);
+    s.add(module + "." + name);
     s.add(name); // For builtins like hack.
   }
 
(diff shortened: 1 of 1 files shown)


Snarling:
Honestly not sure I understand what's going on here, but it's probably just my unfamiliarity with acorn/parsing. This area of the code has always looked unintelligible to me.

Will probably just do a small amount of testing and then merge, it sounds like you already tested most cases already.

Snarling:
Not a new issue created by this PR, but do you have any idea if this could be extended to also work when exporting something that is not defined in the same file but is itself imported? There's an existing bug that prevents using barrel files because the static ramcost isn't detected.

e.g.

```js
// h.js
export async function h(ns){
  await ns.hack("n00dles");
}

// --- SEPARATE FILE --- //
// barrel.js
export {h} from "h";
// alternately, the commented way below also does not work
// import {h} from "h";
// export {h};

// --- SEPARATE FILE --- //
// test.js
import {h} from "barrel";
export let main=ns=>h(ns);
```

Currently, h.js is the only one paying the ramcost for ns.hack, and running test.js leads to a dynamic ram error due to this.

G4mingJon4s:
So, I implemented the re-exports and refactored a bit.
Though, I'm still not confident that I covered all possible cases.
Would it be a good idea to set up some testcases for this? I haven't looked into how the tests work...

Snarling:
Yeah it looks like it still does not work in some cases, but it does work in the normal barrel file usage and I don't think this adds any new issues, so I'm just merging this in and we can try working through the edge cases in a different PR. Thanks!

Issue #624, opened 2023-06-17
Integrate documentation into the game


Hoekstraa:
hydroflame on Discord, today:
Would be cool to add a markdown renderer in a new tab under Help and move most of readthedoc in-game. 

It might complicate the build of the game somewhat, but it could have a few benefits (_**all depending on implementation!**_):
- Documentation would be linked to the version, and so accurate to that version.
- Documentation would be available to players without internet access

Downsides:
- In-game documentation can't be changed without publishing a new version of the game

lossycrypt:
I too would love this as I play offline alot and think it would fit the theme! When I first started I was infact a bit suprised/disappointed that I couldn't really play the game without first googling the basics out-of-game.

I've tried to use wget/httrack to get it working offline with mediocre success. The "basic" (readthedocs) doc I got working with wget (the built in zip-download has a bug where you can't click the chapter text to expand to subchapters). But basic doc has no info aboutextra apis (singularity, etc). I got the github markdown pages working with `wget --mirror --page-requisites --convert-links` but they do not render correctly no matter what I try (looks like the css doesn't load).

Hoekstraa:
ReadTheDocs documentation has been integrated in #667, so closing this one.
Secondly, I opened #707 for specifically integrating API docs.

PR #667, opened 2023-07-10
DOC: Death to readthedocs

hydroflame:
- Implement markdown with MUI.
- Move and convert all rst documentation to md
- Rework documentation a little bit.
- Rename Tutorial tab to Documentation tab
image


src/Tutorial/ui/root.ts
```diff
@@ -0,0 +1,27 @@
+import { Player } from "@nsdefs";
```
TheMas3212: should be import type since its only used for typings

TheMas3212 (approved):
looks good to me

src/Tutorial/ui/DocumentationRoot.tsx
```diff
@@ -0,0 +1,42 @@
+import React from "react";
+
+import Button from "@mui/material/Button";
+import { MD } from "../../ui/MD/MD";
+
+import { getPage } from "./root";
+import { Navigator, useHistory } from "../../ui/React/Documentation";
+
+const resolveRelativePath = (folder: string, relative: string): string => {
```
d0sboots: Pretty sure we already have utility functions for these.

src/Tutorial/ui/doc/advanced/bladeburners.md
```diff
@@ -0,0 +1,3 @@
+# Bladeburners
+
+PLACEHOLDER
```
d0sboots: There's a lot of files like this still. This is a *lot* of docs to bite off in one chunk. IMO it would make more sense to delete each RST as it is converted, and do them in batches, but if you want to do them all in one go I'm not gonna stop you.

src/Tutorial/ui/doc/notfound.md
```diff
@@ -0,0 +1 @@
+### Page not found
```
d0sboots: Do we need this one?

src/ui/MD/MD.tsx
```diff
@@ -0,0 +1,37 @@
+import React from "react";
```
d0sboots, line 1: I think this should be `md/md.tsx` instead. All-caps file names, and especially directory names are weird and not with our style (to the extent that we have one).

src/ui/MD/md_test.ts
```diff
@@ -0,0 +1,58 @@
+// This file contains almost all features that markdown ever uses. It is split by section to be better able to test different features.
```
d0sboots: This shouldn't be called `md_test.ts` IMO, because that implies it's an actualy test. (Like a unittest.) Maybe `md_sample`?

Also, it appears to be dead code?

d0sboots (commented):
Code generally looks good. I didn't try running it, though.

src/ui/MD/a.tsx
```diff
@@ -0,0 +1,21 @@
...
+  const ref = props.href ?? "";
+  if (ref.startsWith("http"))
+    return (
+      <Link rel="noopener noreferrer" href={props.href} target="_blank">
+        {props.children}
+      </Link>
+    );
+
+  if (isSpoiler(ref)) return <CorruptableText content="UNKNOWN" />;
+  return <Link onClick={() => navigator.navigate(ref)}>{props.children}</Link>;
```
Snarling:
```suggestion
  return (
    <Link component={"button"} variant={"body2"} onClick={() => navigator.navigate(ref)}>
      {props.children}
    </Link>
  );
```
Links with no href are not accessible. They don't have the right cursor, are not focusable, and even if we add href=# to make them focusable, they don't actually run the onClick when they are activated with keyboard. With functionality on an onClick, we need to use a button element, change is based on guidance from mui: https://mui.com/material-ui/react-link/#accessibility

Snarling (commented):
Editing existing comment since the original points were dealt with (oversized headers, scroll to top on navigate, home button) and it was a huge comment.

src/ui/MD/a.tsx
d0sboots: I agree that something should change here, but I disagree that we should be changing to buttons. The action here *is* navigation, therefore they ought to be links.

Instead, I think these should have real hrefs, and we should be making our router a bit smarter so that it works with them. Anything that has a history should potentially use the browser's history.

This page has a good overview of the concepts: <https://blog.pshrmn.com/how-single-page-applications-work/>

d0sboots: Oh, another thing on this thread: There are some people (like me) who instinctively ctrl-click to open all documentation links like this in new tabs. Having these be actual links is good because it supports this... theoretically. However, since it would be opening the game to render the docs, you'd rapidly end up with 10 copies of the game running the same save, which is... not ideal.

Probably the best would be if the doc links opened to an external md renderer (such as github) if opened externally. Actually, the more I think about it, the more important that is: I do *not* want to be looking at these in-game. (It's just how I'm wired, sorry.) I understand that other people do, which is why this is a great PR, but we need to keep the option of a link to docs out-of-game.

src/Tutorial/ui/DocumentationRoot.tsx
Snarling: We do but there is a little finessing needed to get the existing resolveFilePath to work here, since it defaults to using an absolute path unless the path actually starts with `./` or `../`

I posted a suggestion but this can easily be dealt with later once everything is working correctly, even in a followup PR.

src/Tutorial/ui/DocumentationRoot.tsx
```diff
@@ -0,0 +1,54 @@
...
+  return prefix + suffix;
+};
+
+const resolvePath = (currentPath: string, newPath: string): string => {
+  const lastIndex = currentPath.lastIndexOf("/");
+  const folder = lastIndex === -1 ? "" : currentPath.slice(0, lastIndex + 1);
+  if (!newPath.startsWith("../")) return folder + newPath;
+
+  return resolveRelativePath(folder, newPath);
+};
```
Snarling: Adding imports and removing duplicate functionality at DocumentationRoot
```suggestion
import { isFilePath, resolveFilePath } from "../../Paths/FilePath";
```

src/Tutorial/ui/DocumentationRoot.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+export function DocumentationRoot(): React.ReactElement {
+  const history = useHistory();
+  const page = getPage(history.page);
+  const navigator = {
+    navigate(relPath: string, external: boolean) {
+      const newPath = resolvePath(history.page, relPath);
```
Snarling: Use existing pathing helpers, and add a console.error for debugging if somehow we get an unresolvable path (e.g. ../ above top level or an invalid character)
```suggestion
      // resolveFilePath uses absolute path unless relative part starts with ./ or ../
      relPath = "./" + relPath;
      const newPath = isFilePath(relPath) && resolveFilePath(relPath, history.page);
      if (!newPath) {
        console.error(`${relPath} was not a valid filepath while navigating documentation from ${history.page}`);
        return;
      }
```

src/ui/React/Documentation.tsx
```diff
@@ -0,0 +1,71 @@
+import React, { useContext, useState } from "react";
+
+interface Navigator {
+  navigate: (s: string, external: boolean) => void;
+}
```
Snarling, line 5: Add import.
```suggestion
import React, { useContext, useState } from "react";
import { FilePath, asFilePath } from "../../Paths/FilePath";

interface Navigator {
  navigate: (s: string, external: boolean) => void;
}
```

src/ui/React/Documentation.tsx
```diff
@@ -0,0 +1,71 @@
...
+  page: string;
+  push(p: string): void;
+  pop(): void;
+  home(): void;
+}
+
+const defaultPage = "index.md";
+
+const HistoryContext = React.createContext<History>({
+  page: "",
```
Snarling, line 22: Use stricter types for History
```suggestion
interface History {
  pages: FilePath[];
  page: FilePath;
  push(p: FilePath): void;
  pop(): void;
  home(): void;
}

const defaultPage = asFilePath("index.md");

const HistoryContext = React.createContext<History>({
  page: defaultPage,
```

src/ui/React/Documentation.tsx
```diff
@@ -0,0 +1,71 @@
...
+  pages: [],
+  push: () => undefined,
+  pop: () => undefined,
+  home: () => undefined,
+});
+
+export const Provider = HistoryContext.Provider;
+export const useHistory = (): History => useContext(HistoryContext);
+
+const onPush = (h: History, p: string): History => {
```
Snarling, line 32: Stricter type for onPush
```suggestion
const onPush = (h: History, p: FilePath): History => {
```

src/ui/React/Documentation.tsx
```diff
@@ -0,0 +1,71 @@
...
+    page: defaultPage,
+    pages: [],
+  };
+};
+
+export const HistoryProvider = (props: React.PropsWithChildren<object>): React.ReactElement => {
+  const [history, setHistory] = useState<History>({
+    page: defaultPage,
+    pages: [],
+    push(p: string) {
```
Snarling, line 60: Stricter type for push implementation
```suggestion
    push(p: FilePath) {
```

src/Tutorial/ui/DocumentationRoot.tsx
```diff
@@ -0,0 +1,54 @@
...
+      }
+      history.push(newPath);
+
+      // Reset scroll to the top of the page.
+      window.scrollTo(0, 0);
+    },
+  };
+
+  // Reset scroll.
+  useEffect(() => {});
```
Snarling: Not related to FilePath but this can just be removed since it was handled above
```suggestion
```

Snarling (commented):
Hiding this because it's not actually important. I'm fine with just dealing with this later in a followup PR, most important thing is addressing functionality concerns so we can merge this.

These are all the changes needed to use the existing typesafe filepaths (tested locally after getting game to build correctly again).

Note that this would make the documentation filenames need to obey the same naming restrictions as ingame files:

https://github.com/bitburner-official/bitburner-src/blob/e4d3a9020e528974c8b58c1919e57accae5c48c9/src/Paths/Directory.ts#L18-L28

Snarling:
@hydroflame I see what you mean with renaming the directory not working right, it works fine locally but the rename doesn't actually get reflected in the commit, so the committed version still fails to build.

I'm renaming that folder to Markdown, that way the change is not strictly capitalization and it should go through. Sorry for any inconvenience, not trying to step on your toes just trying to get this in a state where people can continue to test it without needing to apply local fixes to make the game build.

Snarling (commented):
Button links don't get the right font right now:

image

Locally I was able to fix this by getting rid of all the useStyles/makeStyles stuff in `a.tsx` and adding to props: `variant="body1" fontSize="inherit"`:

image

hydroflame:
I think addressed everything. 

@Snarling No need to feel bad for fixing my stuff. I appreciate it. I didn't know it prevented other people from building though.

Now I'm going to fill in all the placeholders.

hydroflame:
I had to rename the folder `Markdown` to `MD` again because prettier would ignore the files under it.

hydroflame:
There are still some placeholder files but it's too mind numbing to write these things. I hope a user will get annoyed enough to write it for us.

hydroflame:
Fix #624

d0sboots:
> There are still some placeholder files but it's too mind numbing to write these things. I hope a user will get annoyed enough to write it for us.

Noooooo why not `md` ;_;

Edit: Replied to the wrong comment, but this comment is important too. This is *exactly* what I was worried about when you started by deleting all the .rst. Writing documentation *is* mind-numbing; you can't count on anyone else doing it. I don't think we should be deleting the unconverted docs until they have converted replacements.

PR #671, opened 2023-07-12
DOCS: Add autogenerated api docs to new ingame doc viewer

Snarling:
* Made a new combined .sh for generating api docs and performing the ingame generation afterwards. This sh file also moves the autogenerated api docs to the right location to be included in the ingame doc viewer.
* Added a link to the api docs from doc index, but see issues section below.
* Directory reorganization:
  * moved a lot of the Documentation stuff out of the ui folder, only the React file is still in there.
  * Renamed the src/Documentation/ui/doc folder to src/Documentation/markdown.
  * Existing root markdown folder for generating API docs has been moved to src/Documentation/markdown/netscriptDefinitions.

Issues noted so far:
* The autogenerated api docs utilize html tags inside the markdown, and these are not displayed properly. Need to look into either ways to prevent the api extractor from including html tags, or try and get the html to parse correctly. Some of these appear to be due to the html tags being used directly in NetscriptDefinition.d.ts, but others are autogenerated.
* The autogenerated API docs are very unintuitive to navigate.
* ~~The initial link to the the api docs works, but further links from there are not working, I think it is because the filepaths all start with `./`.~~ Fixed by using FilePath

~~Other issues noted not related to this PR (just forgot about this before merging other PR):~~
* ~~Tools and Resources page is missing~~ was just a bad link


Snarling:
Closing this PR. Eventually this will be implemented, but the doc generator needs to be customized (see https://api-extractor.com/pages/setup/custom_docs/) to produce output that is usable ingame. The current autogenerated API docs just don't work ingame.

PR #673, opened 2023-07-13
NETSCRIPT: Rebalance the cost of sing.getOwnedSourceFiles

d0sboots:
This function (rightly, imo) does not actually require SF4. However, since it currently has the SF4 cost multipliers applied to it, it is largely moot since using it without SF4 requires 80GB.

This changes the RAM cost to be a constant 5GB. Ideally this wouldn't live in the sing namespace at all (nothing that just returns information should), but for compatibility that's how it rolls.

### DIFF
--- a/src/Netscript/RamCostGenerator.ts
+++ b/src/Netscript/RamCostGenerator.ts
@@ -187,7 +187,7 @@ const singularity = {
   getCrimeChance: SF4Cost(RamCostConstants.SingularityFn3),
   getCrimeStats: SF4Cost(RamCostConstants.SingularityFn3),
   getOwnedAugmentations: SF4Cost(RamCostConstants.SingularityFn3),
-  getOwnedSourceFiles: SF4Cost(RamCostConstants.SingularityFn3),
+  getOwnedSourceFiles: RamCostConstants.SingularityFn3, // Intentionally not SF4Cost, since it doesn't require SF4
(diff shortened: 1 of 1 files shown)


Snarling:
Alternate solution:

1. Deprecate existing function
2. Add a new function that returns a data type that isn't gross (`{n: number, lvl: number}[]` is gross) and that isn't in the wrong namespace.

Something like getSourceFileLevel(n: number) that just returns the level as a number would make more sense I think. And with it having a much lower ram cost, people are actually incentivized to move away from the deprecated one.

paul-civitas:
Why can't this information just be rolled into getResetInfo ?

We're trying to push getResetInfo and it already has the spoiler information inside it, so why not put that here too.

Snarling:
> Why can't this information just be rolled into getResetInfo ?
> 
> We're trying to push getResetInfo and it already has the spoiler information inside it, so why not put that here too.

Not a bad idea, it's still data that only changes during resets that is fully cacheable, so I think it fits there just fine. In that case it should probably be returned as a Map (`new Map(Player.sourceFiles)` should work). Plain objects aren't great for a record keyed by number, since all the utility functions like Object.entries or Object.keys will convert the keys to string.

Snarling:
Effectively replaced by `ns.getResetInfo().ownedSF` which also is cheaper.

PR #676, opened 2023-07-19
ENUM: Added FactionName, Enum to NetscriptDefinitions

BladeRavinger:
Added a list of faction names to `ns.enums.FactionName`

very little is currently done with factions in scripts and this could lay some ground work to adding more to gang/factions. 

Reason: i had to dig around in source files to find lists of factions, what factions could have gangs, what type of gang each faction was, some of this information should be available ingame for players, id like to get us started down the road towards making this data available so one can get relevant information for automating faction and gang management in new ways


Snarling:
If this data is added to the api, it will most likely not be through the always-available ns.enums, but instead as a function only available through a lategame api

BladeRavinger:
i had thought of adding a function to ns.gang, that way its part of BN2 unlock, basically where it becomes relevant. but someone suggested to go ahead and add the Enum first

d0sboots:
> If this data is added to the api, it will most likely not be through the always-available ns.enums, but instead as a function only available through a lategame api

So, how would we square that with the current requirement that the entire enum needs to be declared in the .d file?

BladeRavinger:
> 

I suppose make what's useful available in a constants file just like in corporations

PR #688, opened 2023-07-31
STOCKMARKET: Move Constants

Caldwell-74:
Moved the Stock Constants from the general Constants and from different places into one Constants.ts under StockMarket/data/

just some prep work to make following refactoring easier
i will either add onto this over the next days or if its acceptable merge it in smaller steps how ever you like

### DIFF
--- a/src/StockMarket/StockMarket.tsx
+++ b/src/StockMarket/StockMarket.tsx
@@ -206,7 +206,7 @@ export function processStockPrices(numCycles = 1): void {
     return;
   }
 
-  // We can process the update every 1 seconds as long as there are enough
+  // We can process the update every 4 seconds as long as there are enough
   // stored cycles. This lets us account for offline time
   const timeNow = new Date().getTime();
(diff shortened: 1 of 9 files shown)


src/StockMarket/StockMarket.tsx
```diff
@@ -208,17 +206,17 @@ export function processStockPrices(numCycles = 1): void {
     return;
   }
 
-  // We can process the update every 4 seconds as long as there are enough
+  // We can process the update every 1 seconds as long as there are enough
```
d0sboots: This is still 4 seconds.

d0sboots (approved):
Good refactor. I'd submit this separately from (presumably) a future change to tweak the constants.

src/StockMarket/StockMarket.tsx
Snarling:
```suggestion
  // We can process the update every 4 seconds as long as there are enough
```

PR #697, opened 2023-08-01
STOCKMARKET: More Constants moved

Caldwell-74:
found these stray in the files
they are only used in the files they where exported from
wich is why the question would be: should they be exposed to the player or not

I have no real opinion on it, did it just for completionist sake 
and I have no problem if this gets closed :3

edit.: okay they are used in the test aswell


Snarling:
Only reason I didn't merge this yet is the names for those variables are not great, and once we add them to the API it becomes significantly harder to rename them. Not actually a problem with the PR, just a legacy naming problem.

Caldwell-74:
oh yeah i was wondering about that but concluded it wasnt too important and nothing to stress you about

ill see with what i can come up when i had some coffee

PR #706, opened 2023-08-03
API: wrote singularity.getAugmentationFactions

### LINKED ISSUE #705: Write  singularity.getAugmentationFactions()
singularity.getAugmentationFactions() method based on discord chat with hydroflame

based on getAugmentationFactions(augName: string): string[]

I have src branched on VS code, currently learning how the .ts is written

myCatsName:
closes #705

/*
.js code for demo
*/

/** @param {NS} ns */
    ns.clearLog()
    ns.tail()
    const testAugmentation = "Xanipher"
    const testAugArray = ["INFRARET Enhancement","Combat Rib I","NeuroFlux Governor",]
  
    //YES 
    ns.print(ns.singularity.getAugmentationFactions("Xanipher"));// ["NWO"]
    ns.print(`hosts for ${testAugmentation}: ${ns.singularity.getAugmentationFactions(testAugmentation)}`);
    ns.print(ns.singularity.getAugmentationFactions("Power Recirculation Core"));
    ns.print(ns.singularity.getAugmentationFactions("NeuroFlux Governor"));//["Illuminati","Daedalus","The Covenant","ECorp","MegaCorp","Bachman & Associates","Blade Industries","NWO","Clarke Incorporated","OmniTek Incorporated","Four Sigma","KuaiGong International","Fulcrum Secret Technologies","BitRunners","The Black Hand","NiteSec","Aevum","Chongqing","Ishima","New Tokyo","Sector-12","Volhaven","Speakers for the Dead","The Dark Army","The Syndicate","Silhouette","Tetrads","Slum Snakes","Netburners","Tian Di Hui","CyberSec"]
    ns.print(ns.singularity.getAugmentationFactions("Augmented Targeting I"));
    ns.print(ns.singularity.getAugmentationFactions("Combat Rib I"));
    ns.print(ns.singularity.getAugmentationFactions("TITN-41 Gene-Modification Injection"));
  
  for (let i=0; i<testAugArray.length;++i){
    ns.print(`hosts for ${testAugArray[i]}: ${ns.singularity.getAugmentationFactions(testAugArray[i])}\n`)
    //ns.tprint(ns.singularity.getAugmentationFactions(testAugArray[i]));
  }
  
    //NO
    //...("TITN41Injection")
    // ie. it works with the player-facing name, not the @enum
    //[AugmentationName.TITN41Injection] 
  
    //("NeuroFlux Governor - Level 1")
    // Must not include level!
    // May be confusing to players?
  
  
    //ns.print(ns.singularity.getAugmentationBasePrice("ZOE"));
    //ns.print(ns.singularity.getAugmentationFactions("ZOE"));
    //These throw same error, seems appropriate.
  
  
    //ns.print(ns.singularity.getAugmentationFactions("someBS"));
    /*singularity.getAugmentationFactions: Argument augmentationName should be a AugmentationName enum member.
    Provided value: "someBS".
    See the developer console for allowable values.
    Script crashed due to an error.*/
  
  
  }


Hoekstraa:
Implements https://github.com/bitburner-official/bitburner-src/issues/705

src/NetscriptFunctions/Singularity.ts
```diff
@@ -95,6 +95,12 @@ export function NetscriptSingularity(): InternalAPI<ISingularity> {
     getOwnedSourceFiles: () => () => {
       return [...Player.sourceFiles].map(([n, lvl]) => ({ n, lvl }));
     },
+    getAugmentationFactions: (ctx) => (_augName) => {
+      helpers.checkSingularityAccess(ctx);
+      const augName = getEnumHelper("AugmentationName").nsGetMember(ctx, _augName);
+      const aug = Augmentations[augName];
+      return aug.factions;
```
Snarling: Don't return the real array directly to the player. When we pass game information to the player in the form of an object/array/etc, we need to give them a copy of the real thing so any modifications the player makes to the return value do not impact the game.

```suggestion
      return aug.factions.slice();
```

Snarling:
@d0sboots I understand the concerns on issue #705 but currently we don't have a better place to put this.

Also personally I think it does need to be locked behind something lategame, and SF4 seems like a good thing to lock it behind since scripts don't interact with Factions and Augmentations outside of that. Open to suggestions on how to organize it better to keep the action-automation away from the info-gathering, but since we're basically doing semver now for versioning, an actual full changeover will need to wait until we're ready with a bunch of other API improvements to do all at once. But in the meantime we can add the new functionality at a different namespace, and soft deprecate the singularity-info-gatherers until that API break.

d0sboots:
> Also personally I think it does need to be locked behind something lategame

If you think this should be locked behind something, then I don't think there's a better place than sing, certainly not as things currently stand.

My personal feeling is that information which is static (I.e. it doesn't change from run), like this or other enums, should be freely available. However, I can see the argument that knowing which factions give which augments is a spoiler.

PR #712, opened 2023-08-04
CCT: accept IP solutions with matched single or double quotes, or none

myCatsName:
From a discord issue https://discord.com/channels/415207508303544321/923282733332054126/1137052241388392530

"For the IP address contracts, I found that entering an IP list like ['127.0.0.1'] doesn't work, but ["127.0.0.1"] does. Is that intentional?"

more information : "how did your answer get formatted that way? did you write your answer by hand?"

[New player] : "Yes, I did."

# Bug fix

Currently the answer input is sanitized by replacing (only) a double quote " from the start and end of each IP with an empty string, for ex. "12.59.159.35" becomes 12.59.159.35

My suggested solution is based on **regex demo** here : https://regex101.com/r/0Ooa2M/1

name the first quote, which can be either single or double, ie ['"]; capture digits and periods; match the end character only if it is  same as the named capture group; then return the IP group.

Rewards solutions like `1.111.11` or `'1.1.1.1'` or `"1.1.1.1"` so any matching quotes or none. In array, **each IP can be wrapped differently**.

Fails solutions like `'1.1.1.1"` or `""1.1.1.1""`

For testing I've attached an IP cracking script that when run will prompt to input a single IP CCT data. I tested by manually changing quotation style in solutions of various length and submitting. I have not tested this PR with an automated cracking script, but I don't think it would fail.

[IP cracker.txt](https://github.com/bitburner-official/bitburner-src/files/12263819/IP.cracker.txt)

### DIFF
--- a/src/data/codingcontracttypes.ts
+++ b/src/data/codingcontracttypes.ts
@@ -537,7 +537,7 @@ export const codingContractTypesMetadata: ICodingContractTypeMetadata[] = [
       }
 
       const sanitizedAns: string = removeBracketsFromArrayString(ans).replace(/\s/g, "");
-      const ansArr: string[] = sanitizedAns.split(",").map((ip) => ip.replace(/^"|"$/g, ""));
+      const ansArr: string[] = sanitizedAns.split(",").map((ip) => ip.replace(/^['"]|['"]$/g, ""));
       if (ansArr.length !== ret.length) {
         return false;
       }
--- a/src/data/codingcontracttypes.ts
+++ b/src/data/codingcontracttypes.ts
@@ -537,7 +537,9 @@ export const codingContractTypesMetadata: ICodingContractTypeMetadata[] = [
       }
 
       const sanitizedAns: string = removeBracketsFromArrayString(ans).replace(/\s/g, "");
-      const ansArr: string[] = sanitizedAns.split(",").map((ip) => ip.replace(/^['"]|['"]$/g, ""));
+      const ansArr: string[] = sanitizedAns
+        .split(",")
+        .map((ip) => ip.replace(/^(?<quote>['"])([\d.]*)\k<quote>$/g, "$2"));
       if (ansArr.length !== ret.length) {
         return false;
       }


myCatsName:
(btw, I am not trying to make 2 commits, I just don't know github and VS code. I made 2 regex, and the one described above seems better to me; I'm not sure why the first version is listed as a commit, sorry).

d0sboots:
I think named capture groups are slightly overcomplicated for this, but it's fine.

myCatsName:
> I think named capture groups are slightly overcomplicated for this, but it's fine.

It seemed kinda gross in practice to allow `"1.1.1'` format answers, which was my simpler solution. The options here should mimic strings.

Named capture groups was the first back reference syntax that worked in the editor for me, and can be helpful to readers, but if there's a different suggestion I will try to implement a change. I noticed there are maybe a few related issues with other CCTs https://github.com/bitburner-official/bitburner-src/issues/64 , and thought a similar approach might apply, but haven't looked at those in detail. In any case, if there are better ideas, or style preferences, please do LMK, I don't have any problem taking notes. Here to help and learn.

d0sboots:
> Named capture groups was the first back reference syntax that worked in the editor for me, and can be helpful to readers, but if there's a different suggestion I will try to implement a change.

`/^(['"])([\d.]*)\1$/g` should work, and is easier to read (for me), since the group here is so simple.

Either way, something I find useful when dealing with *any* regex is a comment nearby explaining what it is supposed to match, with an example. Like:
`// Matches numerical IP-like addresses with surrounding quotes. Example: "999.0.20.2023"`

myCatsName:
I don't want to mess things up by committing another branch, but yes 
`const ansArr: string[] = sanitizedAns.split(",").map((ip) => ip.replace(/^(['"])([\d.]*)\1$/g, "$2"));` should be functionally the same, and works in limited testing, they're the same, just different backreference syntax. 

It's interesting what we consider easier to read - from my PoV, there's a small chance someone reading \1 would connect it with the numbers around it \d* or $2; named group avoids that. But it does seem like style debate at that point, as you said being in a simple regex with a lot of context to understand a purpose. Thankfully we're not talking about functional issues, I really would not want to break a CCT (or anything at all) with bad code.

Snarling:
JSON.parse also does not allow single-quoted strings.

For contracts in general, my preferred long term change to contracts (which would need to wait until the next major API break) is:

API: Only allow player to answer with the actual answer, not a string version of it.
UI: Only allow valid JSON representation of the actual answer, which would then just get parsed by JSON.parse and used as the submitted answer.

Those changes would mean getting rid of any custom parsing of the player input, and when certain things are not accepted (for instance, single quoted strings when submitting an answer via the UI), we'd have a reason to point to (it's not valid JSON) other than we just forgot to support that.

Making input more permissible seems nice, but if there's a reason certain input is right and certain input is wrong, it is better to just communicate to the player why the bad input is not right instead of making the input more permissible.

myCatsName:
> For contracts in general, my preferred long term change to contracts (which would need to wait until the next major API break...

I would be interested in working or collaborating on that. I noticed a lot of helper and sanitizers and makes sense there could be a way to simplify and standardize as you're suggesting.


> Making input more permissible seems nice, but if there's a reason certain input is right and certain input is wrong, it is better to just communicate to the player why the bad input is not right instead of making the input more permissible.

Currently, description and example set player expectation that double and single quotes are possible formats to submit - 
image
Both are used as examples of answers you could submit (players are not told single quote is not accepted).  

My feeling is favor human error here (and be permissive), in the case of CCT, at least gut feeling. There is basically no feedback when an answer is wrong, which can lead to frustration or confusion. And it seems like attempts to clarify a standard with text / visual examples haven't been necessarily effective. Most descriptions probably would need an extra sentence each, and CCTs do not seem like a good place for extra "rules load" on the player or whatever you want to call it. 

>> ...JSON.parse also does not allow single-quoted strings.

My hope at present is match the general style of the in game editor, allowing ' or ", ignoring a leading white space, while not being sloppy. Encourage players to use feature, not fight it.

Snarling:
> My hope at present is match the general style of the in game editor, allowing ' or ", ignoring a leading white space, while not being sloppy. Encourage players to use feature, not fight it.

Yeah that's fair, expecting exact JSON is probably too strict and a bit arbitrary for something input through a UI. We could even use an eval-based approach to parse string answers (for contract types that are not expecting a string), and then the player could use any valid js representation of the object. So it's the same "rules" as if they were describing that same object in a script.

The eventual standardization should not accept bad js like `[192.168.0.1]` with no quotes, because not having quotes of any sort there is just incorrect.

PR #713, opened 2023-08-05
BLADEBURNER: Added BB Training to sleeve actions

Zelow79:
Training seemed to be the only general action sleeves where missing from BB, this aims to correct that. The Training action appeared to already be written with the player or sleeve performing the task in mind.

image
image

This does add a free alternative to the Gym, but I think there is a few pros and cons to consider. The gym may cost money and only do one stat at a time, but it is much faster (per second iirc), while BB training (not only requires BB), but also takes 30 seconds to complete.

### DIFF
--- a/src/PersonObjects/Sleeve/Sleeve.ts
+++ b/src/PersonObjects/Sleeve/Sleeve.ts
@@ -394,6 +394,9 @@ export class Sleeve extends Person implements SleevePerson {
   bladeburner(action: string, contract: string): boolean {
     if (!Player.bladeburner) return false;
     switch (action) {
+      case "Training":
+        this.startWork(new SleeveBladeburnerWork({ type: "General", name: "Training" }));
+        return true;
       case "Field analysis":
       case "Field Analysis":
         this.startWork(new SleeveBladeburnerWork({ type: "General", name: "Field Analysis" }));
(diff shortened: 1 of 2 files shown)


Snarling:
Just haven't had time to look into how this impacts the overall balance of training stats (e.g. how does the free bladeburner training compare in total xp/s compared to paying for powerhouse gym)

Zelow79:
> Just haven't had time to look into how this impacts the overall balance of training stats (e.g. how does the free bladeburner training compare in total xp/s compared to paying for powerhouse gym)

I am not 100% sure of the specifics, but it isn't lightning fast or anything, especially if the sleeve has no augments. Another thing to take into consideration that I forgot to mention is that training costs stamina. So really early on it can be a little risky/difficult, but I imagine it would pull ahead of sec-12 gym in the late game. The stamina gain feels pretty small too and iirc the stamina gains fall off on install. That is about all I can think of between the two, even with bonus cycles it didn't seem overpowered to me, but I respect whatever you decide. This wasn't that hard to throw together and if need be, somethings could be tweaked for sleeves.

Snarling:
Sorry for not following up on this one for so long.

So Powerhouse Gym is 10/s to a single stat, whereas Blade training is 30/30s to 4 stats, or the equivalent of 4/s in a single stat. Also it looks like Training does not scale at all with Hyperdrive which is probably the main thing I was worried about.

The stamina cost thing is low, and I think the permanent stamina gain is worth more.

Since it's a general action, it doesn't scale with Hyperdrive so that's not an issue.

Overall I think it's pretty balanced, especially because you can't just do it right from the start, you have to actually get into Bladeburner first.

Zelow79:
No worries, I was in no hurry. I am glad to see it made it in though.

PR #716, opened 2023-08-05
UI: "Contract in Progress" window can no longer get lost

### LINKED ISSUE #28: Coding contract window lost due to singularity function
[macdjord]
I had a coding contract window open when a script I had running in the background called the exercise-at-gym singularity function. This caused the game to switch to the gym training screen and the contract window to vanish. When I went back to the terminal and tried to reopen the contract by running it again, I got an error message saying "There's already a Coding Contract in Progress". I had to reload the game to get the contract open again.

Version: Bitburner v1.3.0 (9fdc3ac5)

[SagePtr]
> Also contract window disappears when you return your focus to faction job (hacking contracts for instance) when coding contract window is opened. You cannot programmatically solve contracts after this (the game thinks contract is running) until you reload the game.

[nulflux]
> This also happens when triggering a DOM click() on a menu option like City as the contract window is open. The coding contract is forever in progress and there is no way to return to it. There are two potential solutions for this: 1) instead of an error the run command could default to cancel and re-open the contract or 2) make available a command to cancel the current contract. Solution 1 solves the problem universally; solution 2 allows the users to solve the problem themselves. Since there is no special mechanic preventing players pre-singularity from running a coding contract programmatically I don't believe it should be a singularity function but rather a function available by default.

[phyzical]
> idk if this should be fixed, by changing focus is it not expected to close windows?
> 
> or is it more of a the window should be there when you go back to the terminal?

[macdjord]
> @phyzical The problem isn't that the window closes. The problem is you can't reopen it because the game thinks its still open.

[BrianLDev]
> This bug happened to me today too in the way @SagePtr described.
> 
> If you have a contract open and change focus to Faction work, the coding contract UI disappears but the game incorrectly thinks that it's still open.  Doesn't look like there's a way to get the UI to pop back up again, and trying to run the contract again gives an error message. `There's already a Coding Contract in Progress`.  The only way to fix it is to close and reopen the game.
> 
> image

[BrianLDev]
> So I'm guessing the fix would be in the code that handles focus.  It should check for any open coding contracts and simply close them before changing focus.

[macdjord]
> @BrianLDev That seems like treating the symptom rather than the problem to me. A better solution would be to change the 'is a coding contract in progress' check so it actually checks whether there is a contract window open - as in, examine the DOM - rather than depend on finding every possible way that the contract window could get closed and ensuring each one marks the window as closed first.

[holgero]
> Why is there a guard against opening the contract again, anyway? Perhaps just delete that?

[rayanth]
> @holgero it was likely intended in order to prevent two _different_ coding contracts from being opened at the same time. I agree with @macdjord that the better fix would be in the check for a coding contract in progress, but rather than simply check if a window is open, check if a _different_ contract's window is open. If the same coding contract's window is thought to be open, just display it again.

[YoshiRulz]
> Just hit this on 2.0.2 via Steam.

myCatsName:
#28

# Bug fix

Added a clean up useEffect which runs the .onClose function if modal is unexpectedly unmounted, resolving the CCT promise as "cancelled" (which prints in terminal).

Some [testing details](https://github.com/bitburner-official/bitburner-src/issues/28#issuecomment-1666578711) and a [script used for testing](https://github.com/bitburner-official/bitburner-src/issues/28#issuecomment-1666128871) are included in issue #28 , I would encourage additional testing since at some points the code I used had significant unwanted behaviors (not allowing solution input, for example).

There is an additional / related issue, also mentioned in issue #28 about a React console error. I think that can be corrected by adding unsubscribing in the cleanup - the error points to line 25, the CC Event props subscription useEffect.   https://www.rockyourcode.com/avoid-memory-leak-with-react-setstate-on-an-unmounted-component/

But it may be a non-issue.

image
 - https://stackoverflow.com/questions/71258604/how-to-fix-react-warning-cant-perform-a-react-state-update-on-an-unmounted-co

### DIFF
--- a/src/ui/React/CodingContractModal.tsx
+++ b/src/ui/React/CodingContractModal.tsx
@@ -24,6 +24,12 @@ export function CodingContractModal(): React.ReactElement {
   useEffect(() => {
     CodingContractEvent.subscribe((props) => setContract(props));
   });
+  useEffect(() => {
+    return () => {
+      contract?.onClose();
+    };
+  }, [contract]);
+
   if (contract === null) return <></>;
 
   function onChange(event: React.ChangeEvent<HTMLInputElement>): void {


d0sboots:
I think what is happening here with the event subscriber is fundamentally bad. The proposed solution does not look good to me either, because it's using effects in an very strange way. I think this is because the more fundamental issue hasn't been addressed.

Edit: Unfortunately, I haven't looked at this code enough to have a better suggestion.

myCatsName:
I don't understand why this failed a test.

"it's using effects in an very strange way."

No offense, but can you be more clear please? The solution seems to work as written without unexpected problems. 1 useEffect must run constantly, and the other only on unmounting, they need 2 different dependency arguments and so are written in 2 functions. Feel free to test in practice.

Snarling:
> I don't understand why this failed a test.

It doesn't like the CAPITALIZED: portion of the title being more than one word.

d0sboots:
> "it's using effects in an very strange way."
> 
> No offense, but can you be more clear please? The solution seems to work as written without unexpected problems. 1 useEffect must run constantly, and the other only on unmounting, they need 2 different dependency arguments and so are written in 2 functions. Feel free to test in practice.

Yeah, I get that this was not very actionable feedback. A previous version of the fix that you tried felt more in-line with how effects are "supposed" to be used - the version that used the return value to unregister the contract. But that version didn't (fully) work, because of the weirdness of the event subscriber stuff we're doing. To me, that suggests that we should be fixing that, as opposed to having different effects with different dependencies.

myCatsName:
> It doesn't like the CAPITALIZED: portion of the title being more than one word.

If possible you are more than welcome to change the title, and I'll keep that in mind.

I'm sorry to seem defensive, but working code is working code and I believe I have working code to fix a more than year old issue, that forces players to restart with no explanation or instruction. I hope I fixed that. If I found a way to work with "weirdness", OK. 

>"as opposed to having different effects with different dependencies."

https://legacy.reactjs.org/docs/hooks-effect.html#tip-use-multiple-effects-to-separate-concerns

d0sboots:
> https://legacy.reactjs.org/docs/hooks-effect.html#tip-use-multiple-effects-to-separate-concerns

I think this link highlights what is making me feel uncomfortable. Using multiple effects to separate *unrelated* concerns is good style. This is using multiple effects to separate two parts of the *same* concern (registering/unregistering the contract).

> I'm sorry to seem defensive, but working code is working code

Sometimes working code can still be a written in a way that's hard to understand, or a trap for future bugs. That said, it's not like I've jumped up to fix this either. I have a vague sense that there's a better solution, but the perfect can be the enemy of the good. It's up to Snarling to decide.

Snarling:
I'm making a note to look into this deeper in the future when I have some more time, but since this is a working fix and we don't have a concrete better alternative already suggested or ready to go, I'm going to merge this. Sometimes weird antipatterns like this are needed unless we do a much larger refactor.

Snarling:
Oh @myCatsName just for future reference if you put the text "closes #28" or several other variations in your main PR description, it will auto-close that issue when we merge the PR. Thanks for contributing!

myCatsName:
you are welcome, I appreciate the project and opportunity to contribute

d0sboots:
Yes, tyvm for contributing! Sorry if I came off as harsh.

myCatsName:
> Yes, tyvm for contributing! Sorry if I came off as harsh.

TBF @d0sboots , it's a project open to amateurs, which I am - you have every reason to doubt and question, and you have experience to tell you when something smells off, which it does. I went to sleep thinking there's an obvious way to fix the React console error, but could only find another dozen ways to not improve the situation.

But I've exhausted my knowledge and resources, and hopefully solved a problem along the way. Will be keen to know if / when the better way is found.

myCatsName:
> I'm making a note to look into this deeper in the future...

@Snarling Not sure if you'll see this, but I came around to the same issue and got a bit deeper this time, with a bit more understanding than before. I still have no fix, but can point to src\utils\EventEmitter.ts as the BB implementation of the Redux `store.subscribe` methods.

The React  source for store.subscribe is here - https://github.com/reduxjs/redux/blob/master/src/createStore.ts#L225
Specific React subscribe/unsubscribe useEffect discussion here - https://stackoverflow.com/questions/73997221/how-does-unsubscribe-in-the-useeffect-cleanup-function-actually-work
and - https://stackoverflow.com/questions/59034923/how-to-execute-store-unsubscribe-from-useeffect-using-react-hooks-and-redux/59035126#59035126

PR #718, opened 2023-08-06
BUGFIX: fix NFG buying current Level instead of next Level

Caldwell-74:
added aug.getNextLevel to make it clear where the current level and where the +1 level is needed
fixing the bug that the nfg bough is current level + queued and not +1  
wich leads to one wasted buy per install when nfgs are bought


myCatsName:
The way NFG is treated as a special separate case in some of the aug related files looked a bit scary at first, good job finding a solution. Could I ask though, if `getNextLevel()` is only used in 1 case, and only to add 1, couldn't it be added in place?
`     if (aug.name == AugmentationName.NeuroFluxGovernor) {`
`      queuedAugmentation.level = aug.getLevel() +1;`

Caldwell-74:
> The way NFG is treated as a special separate case in some of the aug related files looked a bit scary at first, good job finding a solution. Could I ask though, if `getNextLevel()` is only used in 1 case, and only to add 1, couldn't it be added in place? ` if (aug.name == AugmentationName.NeuroFluxGovernor) {` ` queuedAugmentation.level = aug.getLevel() +1;`

yes currently its only in that one place and it could be done inplace
but like you mentioned its already rather hard to get a good grasp what happens where 
a new function for it seemed to be the most clear way to implement it and could also be used for other augs if there ever is another one with a level introduced

edit.: that this is in a tsx also screams for a refactor imho

Snarling:
> edit.: that this is in a tsx also screams for a refactor imho

Actually Augmentation.tsx can just be renamed to .ts since it doesn't use React. That's my bad, I refactored it recently to get rid of a bunch of React stuff, but forgot to rename the file. Wanna just rename it before I merge this?

edit: AugmentationHelpers.tsx can also be renamed

PR #720, opened 2023-08-06
API: Fix removed functions

Snarling:
* Removed functions are no longer enumerated via Object.keys / Object.entries
* Removed functions work again (a change in #297 made them rely on getRamCost not throwing, and a change in #526 made getRamCost throw for functions with no defined ram cost which broke the removedFunctions)

Fixes #717

PR #721, opened 2023-08-06
Minor grammar fix

HiEv:
A minor grammar fix in the error message at the end of the code.


d0sboots:
I think you're targeting the wrong branch with your PR

Snarling:
Oh actually this is just something that should have been a comment on the still-open PR #720

Snarling:
Addressed in #720

PR #731, opened 2023-08-10
ENUMS: Fixed CompanyName

Zelow79:
CompanyName seemed to be mostly setup already, I just applied the needed changes to NetscriptFunctions.ts and NetscriptDefinitions.d.ts to allow them to be used in the editor.

image


Zelow79:
Updated some missed types for a few functions. So that the definitions show companyName as type CompanyName

image

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -1759,7 +1759,7 @@ export interface Singularity {
    * @param positionName - Name of position to get the requirements for. Must be an exact match.
    * @returns CompanyPositionInfo object.
    */
-  getCompanyPositionInfo(companyName: string, positionName: JobName): CompanyPositionInfo;
+  getCompanyPositionInfo(companyName: CompanyName, positionName: JobName): CompanyPositionInfo;
```
Snarling: For functions we do not want to use just the enum type, because then:

1. There is no autocompletion when people are typing a valid string literal
2. People who write in ts will get type errors if they pass in a valid string literal.

Same comment all the places where CompanyName is taken as an argument type. See other functions that take in an enum member as an input

Zelow79: I actually thought about this as a possibility as I pushed the update, but figured it should be easily excludable for you if you just wanna take the first PR I submitted on here. I just noticed that is how LocationName was setup and tried to push the update to have an option that kept it in line with how other Enums seemed to be setup.

Zelow79: Even in your screenshot positionName was taking a JobName type value.

PR #736, opened 2023-08-18
UI: Support for terminal history search with arrow keys

ficocelliguy:
Adds support for using the up and down arrow keys to search through your terminal history based on current text, similar to how the [feature works on zsh](https://coderwall.com/p/jpj_6q/zsh-better-history-searching-with-arrow-keys)

For example, typing "ru" and then hitting up arrow will cycle through previous commands you have entered starting with "ru" such as `run n00dles.js` and `run Nuke.exe` until you find the one you want, and hitting "tab" will then fill in that command from the history.  Hitting "enter" will execute the suggestion as a command.

Up and down arrow on an empty terminal is unchanged, and maintains the current functionality of cycling through recent terminal history sorted by time entered.

This feature can be disabled under the "misc" settings if desired by the user.

In the below example, `connect foodnstuff` is the first search result. Even though there are other commands that were entered after "connect foodnstuff"  (such as "home") , they are skipped because the user was searching for commands starting with "co" 

Capture3

### DIFF
--- a/src/Settings/Settings.ts
+++ b/src/Settings/Settings.ts
@@ -27,7 +27,7 @@ export const Settings = {
   /** Whether to enable bash hotkeys */
   EnableBashHotkeys: false,
   /** Whether to enable terminal history search */
-  EnableHistorySearch: true,
+  EnableHistorySearch: false,
   /** Timestamps format string */
   TimestampsFormat: "",
   /** Locale used for display numbers. */
--- a/src/Settings/Settings.ts
+++ b/src/Settings/Settings.ts
@@ -26,6 +26,8 @@ export const Settings = {
   DisableOverviewProgressBars: false,
   /** Whether to enable bash hotkeys */
   EnableBashHotkeys: false,
+  /** Whether to enable terminal history search */
+  EnableHistorySearch: true,
   /** Timestamps format string */
   TimestampsFormat: "",
   /** Locale used for display numbers. */
--- a/src/Terminal/ui/TerminalInput.tsx
+++ b/src/Terminal/ui/TerminalInput.tsx
@@ -1,4 +1,4 @@
-import React, {useState, useEffect, useRef} from "react";
+import React, { useState, useEffect, useRef } from "react";
 import Typography from "@mui/material/Typography";
 import { Theme } from "@mui/material/styles";
 import makeStyles from "@mui/styles/makeStyles";
@@ -44,10 +44,7 @@ export function TerminalInput(): React.ReactElement {
   const [value, setValue] = useState(command);
   const [postUpdateValue, setPostUpdateValue] = useState<{ postUpdate: () => void } | null>();
   const [possibilities, setPossibilities] = useState<string[]>([]);
-  const [searchTerm, setSearchTerm] = useState<string>("");
-  const [searchResultsIndex, setSearchResultsIndex] = useState<number>(-1);
   const [searchResults, setSearchResults] = useState<string[]>([]);
-  const [currentSearchPossibility, setCurrentSearchPossibility] = useState<string>("");
   const [autofilledValue, setAutofilledValue] = useState<boolean>(false);
   const classes = useStyles();
 
@@ -78,13 +75,11 @@ export function TerminalInput(): React.ReactElement {
   function handleValueChange(event: React.ChangeEvent<HTMLInputElement>): void {
     saveValue(event.target.value);
     setPossibilities([]);
-    resetSearch();
+    setSearchResults([]);
+    setAutofilledValue(false);
   }
 
   function resetSearch(isAutofilled = false) {
-    setSearchTerm("");
-    setCurrentSearchPossibility("");
-    setSearchResultsIndex(-1);
     setSearchResults([]);
     setAutofilledValue(isAutofilled);
   }
@@ -222,14 +217,15 @@ export function TerminalInput(): React.ReactElement {
     // Autocomplete
     if (event.key === KEY.TAB) {
       event.preventDefault();
-      if (searchResultsIndex !== -1) {
-        saveValue(currentSearchPossibility);
+      if (searchResults.length) {
+        saveValue(searchResults[0]);
         resetSearch(true);
         return;
       }
       const possibilities = await getTabCompletionPossibilities(value, Terminal.cwd());
       if (possibilities.length === 0) return;
-      resetSearch(true);
+
+      setSearchResults([]);
       if (possibilities.length === 1) {
         saveValue(value.replace(/[^ ]*$/, possibilities[0]) + " ");
         return;
@@ -258,25 +254,24 @@ export function TerminalInput(): React.ReactElement {
         return;
       }
 
-      // If there is a partial command in the terminal, hitting up will filter the history
-      if(value && !autofilledValue && Settings.EnableHistorySearch) {
-        let results = searchResults;
-
-        let index;
-        if (searchTerm !== value) {
-          results = Terminal.commandHistory.filter(item => item?.startsWith(value));
-
+      // If there is a partial command in the terminal, hitting "up" will filter the history
+      if (value && !autofilledValue && Settings.EnableHistorySearch) {
+        if (searchResults.length > 1) {
+          const results = [...searchResults];
+          results.push(results.shift() ?? "");
           setSearchResults(results);
-          setSearchTerm(value);
-          setSearchResultsIndex(index = 0);
-        } else {
-          setSearchResultsIndex(index = (searchResultsIndex + 1) % results.length);
+          return;
         }
+        const newResults = [...new Set(Terminal.commandHistory.filter((item) => item?.startsWith(value)).reverse())];
 
-        const result = results[index];
-        setCurrentSearchPossibility(result)
-
-        return;
+        // If there is only one result, simply make that the value in the terminal
+        if (newResults.length === 1) {
+          saveValue(newResults[0]);
+          return;
+        } else if (newResults.length) {
+          setSearchResults(newResults);
+          return;
+        }
       }
 
       if (i < 0 || i > len) {
@@ -288,7 +283,7 @@ export function TerminalInput(): React.ReactElement {
       }
       const prevCommand = Terminal.commandHistory[Terminal.commandHistoryIndex];
       saveValue(prevCommand);
-      resetSearch( true);
+      resetSearch(true);
       if (ref) {
         setTimeout(function () {
           ref.selectionStart = ref.selectionEnd = 10000;
@@ -301,10 +296,10 @@ export function TerminalInput(): React.ReactElement {
       if (Settings.EnableBashHotkeys) {
         event.preventDefault();
       }
-      if (searchResultsIndex !== -1) {
-        const index = Math.max(searchResultsIndex - 1, 0);
-        setSearchResultsIndex(index);
-        setCurrentSearchPossibility(searchResults[index]);
+      if (searchResults.length > 1) {
+        const results = [...searchResults];
+        results.unshift(results.pop() ?? "");
+        setSearchResults(results);
         return;
       }
 
@@ -419,7 +414,10 @@ export function TerminalInput(): React.ReactElement {
             </Typography>
           ),
           spellCheck: false,
-          onBlur: () => setPossibilities([]),
+          onBlur: () => {
+            setPossibilities([]);
+            resetSearch();
+          },
           onKeyDown: onKeyDown,
         }}
       ></TextField>
@@ -439,7 +437,7 @@ export function TerminalInput(): React.ReactElement {
         </Paper>
       </Popper>
       <Popper
-        open={searchResultsIndex >= 0 && possibilities.length === 0}
+        open={!!searchResults.length && possibilities.length === 0}
         anchorEl={terminalInput.current}
         placement={"top"}
         sx={{ maxWidth: "75%" }}
@@ -449,7 +447,7 @@ export function TerminalInput(): React.ReactElement {
             Search result:
           </Typography>
           <Typography classes={{ root: classes.preformatted }} color={"primary"} paragraph={false}>
-            {`>  ${currentSearchPossibility}`}
+            {`>  ${searchResults[0]}`}
           </Typography>
         </Paper>
       </Popper>
--- a/src/Terminal/ui/TerminalInput.tsx
+++ b/src/Terminal/ui/TerminalInput.tsx
@@ -1,4 +1,4 @@
-import React, { useState, useEffect, useRef } from "react";
+import React, {useState, useEffect, useRef} from "react";
 import Typography from "@mui/material/Typography";
 import { Theme } from "@mui/material/styles";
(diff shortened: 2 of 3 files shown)


src/Terminal/ui/TerminalInput.tsx
```diff
@@ -237,6 +257,28 @@ export function TerminalInput(): React.ReactElement {
...
         return;
       }
+
+      // If there is a partial command in the terminal, hitting up will filter the history
+      if(value && !autofilledValue && Settings.EnableHistorySearch) {
+        let results = searchResults;
+
+        let index;
+        if (searchTerm !== value) {
+          results = Terminal.commandHistory.filter(item => item?.startsWith(value));
```
ficocelliguy: When testing with an artificial history of 10 million items, and with the desktop game open and many other chrome tabs, this search took ~140ms +- 20ms. This is only run once, when the user starts a new search, and is maintained as they navigate through the results.

src/Terminal/ui/TerminalInput.tsx
```diff
@@ -44,6 +44,11 @@ export function TerminalInput(): React.ReactElement {
   const [value, setValue] = useState(command);
   const [postUpdateValue, setPostUpdateValue] = useState<{ postUpdate: () => void } | null>();
   const [possibilities, setPossibilities] = useState<string[]>([]);
+  const [searchTerm, setSearchTerm] = useState<string>("");
+  const [searchResultsIndex, setSearchResultsIndex] = useState<number>(-1);
+  const [searchResults, setSearchResults] = useState<string[]>([]);
+  const [currentSearchPossibility, setCurrentSearchPossibility] = useState<string>("");
+  const [autofilledValue, setAutofilledValue] = useState<boolean>(false);
```
G4mingJon4s: To me, this seems like a lot of unnecessary values are stored...
What I would do is something like this:
- Remove searchTerm, just use value instead
- Remove currentSearchPossibility, use the first value of searchResults as the current value
- Remove searchResultsIndex, shift the searchResults instead
- Make searchResults be a computed value

The idea is to make all of the information you need be the array. The first value is the currently proposed command. Pressing up again shifts the array left, pushing the value from the front to the back.

Snarling:
* Most terminals don't use up/down as a way to search for a string as part of a command, they just use up/down to go between the most recent history entries. Having this be an option does fix that, but it should be a default off option because it's not a normal terminal behavior.
* The tab autocompletion popup thing is weird in this case. I get that it would be a lot more difficult to implement something that works like a real terminal though. Also like regular tab autocompletion, it should probably just immediately go to the one option if there's only one option, instead of showing a popup with one option and making you hit tab to select it.
* ~~Does not seem to work correctly, not all options are shown at least in this case (typed test and then hit up arrow):~~ 
image

Edit: Oh, I see it is only showing one suggestion at a time. It seems to show the oldest suggestion first, which does not seem correct to me.

Also hitting up at the end of the list loops the list, hitting down does not. Weird.

G4mingJon4s (changes requested):
Some minor things about the useStates used bother me a bit.

ficocelliguy:
Changes per feedback:
- Removed most of the state variables for simplicity
- Changed results array to use shift() and unshift() to navigate
- Changed both up and down arrow to loop through history
- Fixed suggestion order to start with most recent
- Added feature to autocomplete if there is only one suggestion that matches search history
- Changed feature flag to default to false

ficocelliguy:
Per suggestions from the discord ( https://discord.com/channels/415207508303544321/459097632896188436/1142187892303736863 ), changed the UI to act like ZSH, allowing for enter key selection and displaying current search result in terminal itself, instead of a pop-up:

Capture3

src/Terminal/ui/TerminalInput.tsx
Hoekstraa: @G4mingJon4s Be aware that shifting has a complexity of O(n), compared to indexing being O(1)

src/Terminal/ui/TerminalInput.tsx
```diff
@@ -237,6 +265,27 @@ export function TerminalInput(): React.ReactElement {
...
+
+      // If there is a partial command in the terminal, hitting "up" will filter the history
+      if (value && !autofilledValue && Settings.EnableHistorySearch) {
+        if (searchResults.length > 1) {
+          const results = [...searchResults];
+          results.push(results.shift() ?? "");
+          setSearchResults(results);
+          return;
+        }
+        const newResults = [...new Set(Terminal.commandHistory.filter((item) => item?.startsWith(value)).reverse())];
```
ncharris93, line 283: is this done to get unique values?

src/Terminal/ui/TerminalInput.tsx
```diff
@@ -301,10 +296,10 @@ export function TerminalInput(): React.ReactElement {
...
         event.preventDefault();
       }
-      if (searchResultsIndex !== -1) {
-        const index = Math.max(searchResultsIndex - 1, 0);
-        setSearchResultsIndex(index);
-        setCurrentSearchPossibility(searchResults[index]);
+      if (searchResults.length > 1) {
+        const results = [...searchResults];
+        results.unshift(results.pop() ?? "");
+        setSearchResults(results);
```
ncharris93, line 314: do you need a shallow copy of the searchResults? 

I might suggest using the setState's callback fn. 
```suggestion
        setSearchResults((results) => {
            results.unshift(results.pop() ?? "");
            return results;
        });
```





technically you could also do 
```suggestion
        searchResults.unshift(searchResults.pop() ?? "");
        setSearchResults(searchResults);
```
though mutating state is usually considered tabboo

src/Terminal/ui/TerminalInput.tsx
```diff
@@ -44,6 +53,8 @@ export function TerminalInput(): React.ReactElement {
   const [value, setValue] = useState(command);
   const [postUpdateValue, setPostUpdateValue] = useState<{ postUpdate: () => void } | null>();
   const [possibilities, setPossibilities] = useState<string[]>([]);
+  const [searchResults, setSearchResults] = useState<string[]>([]);
+  const [autofilledValue, setAutofilledValue] = useState<boolean>(false);
```
ncharris93: Just as a note: it's optional  to type `useState` hooks if they have a default value that can be inferred

src/Settings/Settings.ts
```diff
@@ -26,6 +26,8 @@ export const Settings = {
   DisableOverviewProgressBars: false,
   /** Whether to enable bash hotkeys */
   EnableBashHotkeys: false,
+  /** Whether to enable terminal history search */
```
myCatsName, line 29: @ncharris93 the word is correctly spelled `whether`

src/Terminal/ui/TerminalInput.tsx
ficocelliguy: Yes, to clean up duplicates in the history.

src/Terminal/ui/TerminalInput.tsx
ficocelliguy: Yeah, that is why I originally held an index instead. In addition it seems like updating the index triggers a re-render, unlike shifting an array, and having to allocate a new array each time seems expensive; I think I'll re-implement the index instead, after doing this, unless there is a good reason not to?

Snarling:
* I found it pretty hard to see the ghost text at least on the default theme:
image
* Hitting enter with the autocompleted entry selected leads to odd terminal output (actually runs the selected command from history, but shows on the display as if you only performed the partial entry): 
image
* I don't use zsh so I can't verify this myself, but I think there need to be other keybindings that have special interaction if the user is currently cycling through partially-completed history:
  * Right or left arrow should commit the current search result to the terminal input, and just move the cursor from its equivalent current position
  * Escape should probably abort the history search (currently can be done by typing a character and then deleting it)

Snarling (commented):
* I found it pretty hard to see the ghost text at least on the default theme:
image
* Hitting enter with the autocompleted entry selected leads to odd terminal output (actually runs the selected command from history, but shows on the display as if you only performed the partial entry): 
image
* I don't use zsh so I can't verify this myself, but I think there need to be other keybindings that have special interaction if the user is currently cycling through partially-completed history:
  * Right or left arrow should commit the current search result to the terminal input, and just move the cursor from its equivalent current position
  * Escape should probably abort the history search (currently can be done by typing a character and then deleting it)

d0sboots:
> I don't use zsh so I can't verify this myself, but I think there need to be other keybindings that have special interaction if the user is currently cycling through partially-completed history

I agree, with the caveat that what you suggested isn't how it works in zsh (by default, you can of course customize it however you want), and I think we should stick to how it works in a real shell.

d0sboots:
I use zsh, but with a heavily customized setup that includes using vi keybindings adjusted for Dvorak. So, I did some research on the defaults.

The default keymap is `emacs`, and here's what the manual has to say about the relevant default keybinds:
```
down-line-or-history (CTRL-n DownArrow)
    Move down a line in the buffer, or if already at the bottom line, move to the next event in the history list.

up-line-or-history (CTRL-p UpArrow)
    Move up a line in the buffer, or if already at the top line, move to the previous event in the history list.

down-line-or-search (unbound)
    Move down a line in the buffer, or if already at the bottom line, search forward in the history for a line
    beginning with the first word in the buffer.

up-line-or-search (unbound)
    Move up a line in the buffer, or if already at the top line, search backward in the history for a line beginning
    with the first word in the buffer.

history-incremental-search-backward (Alt-R Alt-Xr)
    Search backward incrementally for a specified string.  The search is case-insensitive  if  the  search  string
    does  not  have  uppercase letters and no numeric argument was given.  The string may begin with `^' to anchor
    the search to the beginning of the line.

    A restricted set of editing functions is available in the mini-buffer.  Keys are  looked  up  in  the  special
    isearch  keymap, and if not found there in the main keymap (note that by default the isearch keymap is empty).
    An interrupt signal, as defined by the stty setting, will stop the search and go back to  the  original  line.
    An  undefined  key will have the same effect.

    (After this, there's a list of how various keybinds interact with this mode, you can man zshzle for details)

history-incremental-search-forward (Alt-S Alt-Xs)
    Search forward incrementally for a specified string.  The search is case-insensitive if the search string does
    not  have  uppercase  letters  and no numeric argument was given.  The string may begin with `^' to anchor the
    search  to  the  beginning  of  the  line.  The functions available in the mini-buffer are  the  same  as  for
    history-incremental-search-backward.

history-search-backward (Alt-P Alt-p)
    Search backward in the history for a line beginning with the first word in the buffer.

history-search-forward (Alt-N Alt-n)
    Search forward in the history for a line beginning with the first word in the buffer.
```

Of particular note is that the behavior that the PR author is used to seems to be `up-line-or-search`/`down-line-or-search`, which is unbound by default but easily swapped for the default keybinds.

Other information: In zsh, while editing commands you can freely edit *any* command in the history. The "new" command is merely an entry that starts out empty, at the end of the history, and which you start editing by default. But until you "accept" a line by pressing enter, *all* changes are remembered, and this is reflected when searching through the history. For example, if the history started like this:

```
ps
ls
```

And then I went up a line, changed the "ls" to "foo", and went down a line (to the blank "new" entry) and typed "f" and then `Alt-p` (for `history-search-backward`), it would jump back an entry to the history entry that (still) lists "foo". Backward searching for "p" would not do anything.

Once a command is "accepted", it is added to the end of the history (possibly being collapsed if it is the same as a previous entry), and any other changes are lost, because the history when beginning to edit the next line will reflect the global history once again.

I'm not suggesting that we attempt to implement the full complexity of all this, just documenting what occurs.

ficocelliguy:
I have fixed the visual bug where the partial command/search term was printed to the terminal's scrolling history instead of the actually executed command.

I have also darkened the search suggestions to make them easier to read. (This involved slightly re-tooling how they are placed, to avoid covering up the user entry)

I added ESC key handling to clear out search suggestions, if the user doesn't want them; it's not a feature I've seen before but it feels intuitive to use.

I had already added support to the TAB key to fill out the suggested text into the value filed in the way that you recommended the arrow keys would work; if you decide we should also update the value and/or clear search when the arrow keys are pressed I can add that as well.


I don't see any support for React + jest component tests in the project currently, am I missing where it lives? I'm hesitant to change the component structure to expose the onKeyDown() method, just for the sake of unit tests; or to install dependencies and make decisions about how component tests should be structured without consulting you all.

Darker suggestion UI:
image

Snarling:
Looks good to me. Just a couple more questions since I am not familiar with the common setup for this type of feature on a real shell like zsh:

Up arrow when the current text does not match any history results: This causes the current text to be destroyed and replaced with a non-matching history entry. Not sure if this is intended behavior, since this is the only time hitting up will destroy the written text with the option enabled.

Down arrow after typing a search string: This doesn't initiate a history search, it just clears the terminal input. Again I don't know if this is expected behavior.

ficocelliguy:
In the zsh config I am used to (more or less how Oh My Zsh comes out-of-the-box, which has been the standard for my last few development clients), if you hit the up arrow with a search string that does not match anything, it does nothing (leaving the text in place).

If you are at the bottom of the history already, it does nothing and beeps at you if you hit the down arrow (both if you have a 'search string' or not)

Because this functionality seems like it is not what many users are used to, I preserved the existing up and down arrow behavior in cases where there isn't any matches to the search string. Up arrow goes to the previous history entry (not filtered); down arrow clearing the terminal is reproducible in the steam version of the game.

If you feel it would be better, I can expand the description of the config setting, and make it so that if there is any search string, the up and down arrows ONLY navigate through search results and not through history when the setting is enabled. Note that without the beep feedback, the feel would be different than zsh , I suspect, and the game doesn't seem to have sound currently. Perhaps a text indicator could be used to tell the user that no results are found or that they are at the bottom of the history, perhaps re-using the "ghost text" element I have added to show search suggestions? We could also simply do nothing in the up and down arrow case with no matches, to be as close to the IRL terminal behavior as possible.

(Currently I am also allowing the user to rotate through the history if there are matches and wrap around from newest to oldest, based on feedback from earlier; I could change this to be closer to zsh if desired and disallow wrapping around)

Snarling:
> In the zsh config I am used to (more or less how Oh My Zsh comes out-of-the-box, which has been the standard for my last few development clients), if you hit the up arrow with a search string that does not match anything, it does nothing (leaving the text in place).

Yeah I think up arrow with no valid search results -> do nothing is important behavior when this option is enabled. Looks like that's how it works in the real shell, and it also is just a consistency thing (hitting up arrow with this option enabled never destroys text the user has written). Snice this is an option, it's okay for it not to allow normal history navigation when a search fails (the user can just start with an empty string to get that).

I'm not as concerned with the down arrow behavior, as initiating a search with down arrow is not expected usage in zsh or here. I'm fine with keeping it how it is, or with changing it to also not do anything (with or without adding an indicator), or with making it initiate a search but wrap around to the oldest match

ficocelliguy:
Up arrow changed to only move through search results when the feature is enabled; nothing happens if there is no results. Down arrow left unchanged from prior commits (clears terminal if there is no results, navigates downward through results if the user is looking through search results).

PR #739, opened 2023-08-20
UI: Suppress confusing reputation-earned dialog on augment install

ficocelliguy:
The game encourages the player to keep a job or faction work running in the background most of the time. Because of this, the player is often gaining reputation / money / experience in the background while "focusing on something else simultaneously".

Installing augmentations cancels the ongoing work, because the player is no longer part of the faction or no longer qualified for the position, etc. 

This `finishWork()` call causes a dialog box to appear informing the player that, for example, they have a bunch of rep with a faction. However, despite what the game apparently told you, when you check with that faction (once you can re-join it), the listed reputation is zero (because it was lost / turned to favor upon augmentation).

In addition to being misleading, this dialog is one of multiple that the user has to wade through after installing augmentations, which does not improve the player experience.

This pull request is a proof-of-concept of a resolution for this. It provides an optional flag to suppress the dialog upon interrupting work, and uses that flag in the augmentation install `finishWork()` call. (Note that I left the grafting interrupted message as-is, as that is still relevant info to the player if it got interrupted upon augment install). This implementation includes a config setting, so players can opt-in to the feature if they want it.

I am looking for feedback on this. I have introduced two people to the game thus far, and both had the same reaction as I did upon getting hit with this pop-up upon augmentation: brief confusion followed by minor irritation as they work to close panels get back to the terminal window. However, it has apparently been in the game for a while, so it may be valuable in some ways, or otherwise not as much of an interruption as it feels like to the small group I have talked to about it.

Example dialog immediately after installing augments:
Screenshot 2023-08-19 113804

No such reputation is present:
Screenshot 2023-08-19 113552

### DIFF
--- a/src/GameOptions/ui/GameplayPage.tsx
+++ b/src/GameOptions/ui/GameplayPage.tsx
@@ -47,6 +47,14 @@ export const GameplayPage = (): React.ReactElement => {
         tooltip={<>If this is set, the confirmation message before buying augmentation will not show up.</>}
       />
       <OptionSwitch
+        checked={Settings.SuppressAugWorkFinishedPopup}
+        onChange={(newValue) => (Settings.SuppressTIXPopup = newValue)}
+        text="Suppress work finished messages after installing augmentations"
+        tooltip={
+          <>If this is set, the "You finished working for..." dialog will not pop up after installing augmentations.</>
+        }
+      />
+      <OptionSwitch
         checked={Settings.SuppressTIXPopup}
         onChange={(newValue) => (Settings.SuppressTIXPopup = newValue)}
         text="Suppress TIX messages"
--- a/src/GameOptions/ui/GameplayPage.tsx
+++ b/src/GameOptions/ui/GameplayPage.tsx
@@ -47,14 +47,6 @@ export const GameplayPage = (): React.ReactElement => {
         tooltip={<>If this is set, the confirmation message before buying augmentation will not show up.</>}
       />
       <OptionSwitch
-        checked={Settings.SuppressAugWorkFinishedPopup}
-        onChange={(newValue) => (Settings.SuppressTIXPopup = newValue)}
-        text="Suppress work finished messages after installing augmentations"
-        tooltip={
-          <>If this is set, the "You finished working for..." dialog will not pop up after installing augmentations.</>
-        }
-      />
-      <OptionSwitch
         checked={Settings.SuppressTIXPopup}
         onChange={(newValue) => (Settings.SuppressTIXPopup = newValue)}
         text="Suppress TIX messages"
--- a/src/Settings/Settings.ts
+++ b/src/Settings/Settings.ts
@@ -42,6 +42,8 @@ export const Settings = {
   RemoteFileApiPort: 0,
   /** Whether to save the game when the player saves any file. */
   SaveGameOnFileSave: true,
+  /** Whether to hide dialogs for stock work competed upon augmentation install. */
+  SuppressAugWorkFinishedPopup: false,
   /** Whether to hide the confirmation dialog for augmentation purchases. */
   SuppressBuyAugmentationConfirmation: false,
   /** Whether to hide the dialog showing new faction invites. */
--- a/src/Settings/Settings.ts
+++ b/src/Settings/Settings.ts
@@ -42,8 +42,6 @@ export const Settings = {
   RemoteFileApiPort: 0,
   /** Whether to save the game when the player saves any file. */
   SaveGameOnFileSave: true,
-  /** Whether to hide dialogs for stock work competed upon augmentation install. */
-  SuppressAugWorkFinishedPopup: false,
   /** Whether to hide the confirmation dialog for augmentation purchases. */
   SuppressBuyAugmentationConfirmation: false,
   /** Whether to hide the dialog showing new faction invites. */
--- a/src/Work/ClassWork.tsx
+++ b/src/Work/ClassWork.tsx
@@ -12,7 +12,6 @@ import { Work, WorkType } from "./Work";
 import { applyWorkStats, newWorkStats, sumWorkStats, WorkStats } from "./WorkStats";
 import { findEnumMember } from "../utils/helpers/enum";
 import { isMember } from "../utils/EnumHelper";
-import { Settings } from "../Settings/Settings";
 
 export interface Class {
   type: ClassType;
--- a/src/Work/ClassWork.tsx
+++ b/src/Work/ClassWork.tsx
@@ -113,7 +113,7 @@ export class ClassWork extends Work {
   }
 
   finish(cancelled: boolean, suppressDialog?: boolean): void {
-    if (!this.singularity && !suppressDialog && Settings.SuppressAugWorkFinishedPopup) {
+    if (!this.singularity && !suppressDialog) {
       dialogBoxCreate(
         <>
           After {this.getClass().youAreCurrently} for{" "}
--- a/src/Work/ClassWork.tsx
+++ b/src/Work/ClassWork.tsx
@@ -111,8 +111,8 @@ export class ClassWork extends Work {
     return false;
   }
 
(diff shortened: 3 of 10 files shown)


myCatsName:
`brief confusion followed by minor irritation`
Bitburner? minorly irritating and briefly confusing?? /joking 

The game is clear that you lose most progress at augment, so if a message suggests you "have" reputation after resetting to 0 that does seem like a problem to fix. However, I think providing no message creates a new problem, in that players now have to notice "work" or whatever is no longer progressing in the overview window, which needs to be maximized. Adjusting the message in a case like this might be better communication, if necessary. 

As an optional feature it might be a fit for this screen 
image

ficocelliguy:
Some thoughts, in no particular order:

The player already has to notice that their scripts have stopped running and they have to start up their things; a pop-up that says "your scripts have stopped! make sure to run them again!" would not add to the player experience, right? Is this different?

In most cases the thing they were working on before augmentation isn't valid or even really desirable to work on yet.

I would really love for it to be a toast, or even a log in the terminal or something, rather than yet another interrupt. maybe it could be added as some small dynamic text to the interrupt dialog that lets you know you just woke up from augmentations and you feel different. Or maybe the little section in the overview window that normally has the "focus" button could have some hint text after installing augs like "you are currently idle..." or sommat ?

Making it an optional setting seems like a safer idea, for sure.

myCatsName:
Prefacing the actual suggestions before getting into the weeds below.
-  I think you're right all the "focus task finished" pop ups should be suppressed when installing , not only the rep-earn ones.
-  Adding an option to suppress all focus finished pop ups could be a nice feature.

weeds:
> Is this different?

Sure, in a way. For one, while the dialogue might be a problem in this case, the "finished working" pop up itself is a consistent signal to the player that a focus task stopped.

> or otherwise not as much of an interruption as it feels

Could also be true. According to Steam less than 1 in 5 accounts install their first augmentation. And it seems to require exactly one extra click - Install augs, "you feel... different" a player will generally click past the message, and underneath is the work popup which disappears with another click. But the experience would be better without.

One omission with the PR, and part of the underlying iceberg is this sort of message is not only with reputation earned, for example university courses will also produce a pop up in the same situation. And in fact the language and tense varies depending on the pop up - "You spent/ You earned" probably aren't misleading, but sometimes unwanted. "You have/ They now have" are sometimes misleading, and sometimes unwanted.

image

image

My feeling is there are many possible UI designs choices throughout BB, I might personally prefer work finished toasts, but a design element in BB ,IMO, is to create some level of friction for the player, incentivizing exploration and solution making (why we're both here presumably, hello). 

Related to install information, and your feeling of "brief confusion followed by minor irritation as they work to close panels get back to the terminal window" - Maybe both parts could be improved with UI updates,  for example the balance of 
image

Tooltip tells players nothing, but is a flavor text (steam achievement hint).

image

info button has a lot to communicate, and says nothing until/unless player mouses over. Personally I would break it up and give basic information somewhere maybe as much space as the NFG text, which includes even a link for more information, which could be considered for augs too. Possibly doing away with the need for this long boy here 
image

Readability and perceived value of the info in the after-install pop up might improve if text block is broken up with color (less a thing to click past, more a helper and achievement moment). Or it may be an appropriate place to remind players about the reset. Maybe install should route players to the stats page instead of Terminal, or player's choice, or just refresh augs page, who knows.


image



 Sometimes it's hard to know what is preference, what is flavor or author's choice, what works well enough, etc. There's no style guide, AFAIK. In this case I think you found something worth changing.

ficocelliguy:
I added a configuration to suppress the pop-up in question upon augment installation, which makes this current PR the minimum viable solution. It sounds like there is a larger UX discussion to be had to flesh out other options and improvements, which will affect a lot of the menus once a plan is hammered out for it. @myCatsName what is the best way to capture the other aspects of the discussion here for the future? (Also, how are UX design decisions typically done?)

myCatsName:
Keep in mind I'm no more an authority here than you. @Snarling is the active maintainer, and I assume the dialogue and UI elements we're discussing are primarily from Hydro (dev who is transitioning to be less active) and a lesser extent the original designer. Snarling might have some idea the best way to capture and collaborate on ideas if there's motivation, for example using the dev discord channel or a separate one.

My general sense is BB is pretty open and designed for modification by its collective community. I don't think hydro left hoping his work would be untouched, but rather offered a base that we all can work with and collab on. Sanity checking each other. I haven't seen ideas being put down or aside, and the most limiting factor seems to be what we implement and push individually. Agree that a larger scope might benefit from a different approach. I'm not sure there is a "community" roadmap broadly speaking.

I'll copy my notes somewhere local. I'd like to audit the new player experience in this way, but this thread obviously isn't the place lol. you seem like a fish in water knocking out these issues, gg.

src/Work/ClassWork.tsx
```diff
@@ -111,8 +112,8 @@ export class ClassWork extends Work {
     return false;
   }
 
-  finish(): void {
-    if (!this.singularity) {
+  finish(cancelled: boolean, suppressDialog?: boolean): void {
+    if (!this.singularity && !suppressDialog && Settings.SuppressAugWorkFinishedPopup) {
       dialogBoxCreate(
```
Snarling: Here Settings.SupressAugWorkFinishedPopup being `true` is a requirement for actually creating the dialog box, this seems backwards.

src/GameOptions/ui/GameplayPage.tsx
```diff
@@ -46,6 +46,14 @@ export const GameplayPage = (): React.ReactElement => {
         text="Suppress augmentations confirmation"
         tooltip={<>If this is set, the confirmation message before buying augmentation will not show up.</>}
       />
+      <OptionSwitch
+        checked={Settings.SuppressAugWorkFinishedPopup}
+        onChange={(newValue) => (Settings.SuppressTIXPopup = newValue)}
```
Snarling: onChange is altering the wrong setting here

src/Work/CompanyWork.tsx
```diff
@@ -53,8 +54,8 @@ export class CompanyWork extends Work {
     influenceStockThroughCompanyWork(company, gains.reputation, cycles);
     return false;
   }
-  finish(): void {
-    if (!this.singularity) {
+  finish(cancelled: boolean, suppressDialog?: boolean): void {
+    if (!this.singularity && !suppressDialog && Settings.SuppressAugWorkFinishedPopup) {
       dialogBoxCreate(
```
Snarling: Same as above, Settings.SupressAugWorkFinishedPopup is being used for the opposite condition (if it's true, it's allowing creation of the dialog box instead of suppressing)

Same comment the other work types as well. But given the suggestion to remove the option altogether, addressing this will just be removing the extra condition instead of fixing the logic of it.

Snarling (commented):
Only just now got a chance to look at this. I don't think it even needs to be an option, the "you finished working and now you have X reputation" dialog is just wrong when the player just performed a reset. The only info that is important in the dialog is that your active work task was cleared.

In order to perform a reset, the player already has to click past a dialog that warns the player about what will be lost upon installing augs. We can just add to that existing dialog that any active work task will be reset as well. Then there would be no need to show an extra dialog after performing the install, and no need for an extra option.

ficocelliguy:
Config option removed, and references to the config option stripped from the code. In addition I added a line item to the aug confirmation dialog that the current work activity will be reset.

src/Settings/Settings.ts
```diff
@@ -42,6 +42,8 @@ export const Settings = {
   RemoteFileApiPort: 0,
   /** Whether to save the game when the player saves any file. */
   SaveGameOnFileSave: true,
+  /** Whether to hide dialogs for stock work competed upon augmentation install. */
+  SuppressAugWorkFinishedPopup: false,
```
Snarling: Unused setting can be removed from Settings.ts

Other than that seems good.

ficocelliguy: Updated, thank you

PR #741, opened 2023-08-21
UI: Prevent recursive aliases from being resolved. Fixes bitburner-official#630

### LINKED ISSUE #630: Change alias to not substitute the same alias recursively

ficocelliguy:
Refactor alias processing: when resolving an alias, look at the tracked list of previously-processed aliases that led to the current one, and do not resolve any ~~aliaii~~ aliases already present in it. This ensures there are no loops in the alias reference tree in a given command, and replaces the current only-one-alias-lookback solution.

closes bitburner-official#630

### DIFF
--- a/src/Alias.ts
+++ b/src/Alias.ts
@@ -66,10 +66,9 @@ export function removeAlias(name: string): boolean {
  * Returns the original string with any aliases substituted in.
  * Aliases are only applied to "whole words", one level deep
  * @param origCommand the original command string
- * @param maxDepth the maximum depth that alias chains are permitted (default 10)
  */
-export function substituteAliases(origCommand: string, maxDepth = 10): string {
-  return applyAliases(origCommand, maxDepth);
+export function substituteAliases(origCommand: string): string {
+  return applyAliases(origCommand);
 }
 
 /**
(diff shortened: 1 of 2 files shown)


Snarling:
Why the max depth check? I think the origin of that was just to prevent recursive replacements, but if you're already preventing recursive replacements then aliases should just continue to replace until there are no more non-recursive replacements to make. e.g. if a=b, b=c, c=d [...] y=z, then `a` should be replaced with `z`, it shouldn't randomly stop replacing aliases 10 levels deep.

ficocelliguy:
I had left in the max depth as a bit of a failsafe (and in case it was an existing feature). I have removed it and updated the unit tests accordingly.

PR #742, opened 2023-08-21
UI: Re-implement limit on Icarus messages after a bitnode is completed

ficocelliguy:
( Apologies for the multiple PRs in a short period, I had some time to get thoughts down in code )

Currently, the icarus.msg pop-up appears every few minutes for any player who is finishing the endgame of a bitnode. While it can be a helpful hint on the very first BN, this can result in many, many messages if the player is working on a bitnode with other long-running objectives, such as BN 10.

This was already fixed in the past at one point, to ensure the message only appears once after the player completes a bitnode:
https://github.com/danielyxie/bitburner/issues/140#issuecomment-332353004

However, the feature seems to have been lost during a later refactor:
https://github.com/bitburner-official/bitburner-src/commit/d1f4f8fd36553ca62746bbd68d0e5172f551dc05#diff-2108e76478b987bae84b72b039959f9d88c78280ed12959798a081b8af264133L90

This PR re-implements the feature. The message will only be sent once per bitnode to players who have completed prior bitnodes.


Snarling:
Adding a test for something like this is not really wanted. This exhaustive approach to testing would result in us needing an insane number of tests overall if we used this approach throughout the codebase. This methodology makes it much harder to contribute - not only would small changes (like this one) be expected to have a test written for it (you literally have a 60 line test for an easy to understand condition change PR), but having tests written at this level would also result in a much higher likelihood that a refactor ends up breaking tests as well, so a contributor has an additional burden of extra code they need to maintain as part of a refactor, on top of the game code they're already refactoring.

There are definitely projects where this sort of testing makes sense, I just don't think it is sensible here

d0sboots:
I disagree, but then again we clearly disagree about the value of testing.

Yes, adding tests means there is a higher chance that you will have to change the tests when refactoring. This is the price you pay for catching mistakes with tests, instead of with your users or via manual testing.

I do not think we should require tests like this for all contributions, but I *definitely* don't think we should be turning then away.

ficocelliguy:
Supposing that the codebase had comprehensive test coverage, that WOULD add another small hurdle to would-be new contributors; they need to be able to understand the existing tests that need to be updated as part of their change. That being said, it is an important skill to have and cultivate in the development community, and I suspect there are few people who know enough about development to add a feature without knowing how to debug a unit test. It is a core paradigm of clean coding.

The codebase does not have good testing, though, which introduces another issue: It is difficult to be confident that you have not broken existing features, because there is not adequate automated coverage to catch regressions, and the QE team is volunteers who may or may not do full feature passes. (This is another potential hurdle that new contributors need to overcome: learn a new codebase, carefully touch it without breaking things, and then just hope all of the various important features they may not be familiar with work OK, because there are few tests). Refactoring is so much less stressful when there is an existing suite of unit tests to reference; even if you have to change them to work with the new structure, you have all of the feature's use cases delineated for you already.


I made an attempt to add documentation to the unit tests I added here so they are easy to understand, and made them flexible if this area is changed in the future. However, I will remove them if that is what you want, especially given how minor this feature is.

Snarling:
Fair points. You're both right that the existing testing leaves much to be desired. I think the main thing is, I wanted to make sure you understand that including tests like this for simple changes is not required or expected. We can probably add some other Message-related tests later to make the inclusion of a Message test file more worthwhile, so let's keep it around.

PR #743, opened 2023-08-22
DOCS: ReactNode edits for clarity

HiEv:
Minor documentation improvement.

### DIFF
--- a/markdown/bitburner.reactnode.md
+++ b/markdown/bitburner.reactnode.md
@@ -4,9 +4,9 @@
 
 ## ReactNode type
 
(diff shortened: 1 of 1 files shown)


Snarling:
These files do not get edited manually, they are autogenerated from ScriptEditor/NetscriptDefinitions.d.ts. Make the edits there, then run `npm run doc` to generated the .md again

Snarling:
Closing since this was not the correct way to make a documentation change for API documentation. As mentioned in previous comment that should have been done via NetscriptDefinitions.d.ts, and then the .md files can be autogenerated from that.

PR #744, opened 2023-08-23
UI: added autorerender time to Buy Augment page

### LINKED ISSUE #695: Faction augmentatons reputation not updating correctly
as the title says when I enter the UI to buy augments from any faction, the reputation doesn't doesn't update in the UI accordingly, if you are doing a task for the faction to increase reputation faster it will show up on the Overview interface but on the "faction augmentations - faction" reputation keeps the same value and on the section where it's showing the money/reputation price the values will stay in red instead to change to white when you have the required repuration to buy an augment as previous version, if you change page and return it will update all values accordingly, also the buy button will update if you have the required money and reputation

- enter the UI to buy an augment from a faction,  leave and return will update
- Version of the game, e.g. Bitburner v2.4.0 (fc86895c3)

Caldwell-74:
Buy Augment page didnt auto update wich for the most part is fine because there isnt much that changes without player interaction (wich would trigger a rerender) 
but the current reputation with that faction doesnt update so i added a 400ms auto rerender to it
  
 edit.: closes  #695

### DIFF
--- a/src/Faction/ui/AugmentationsPage.tsx
+++ b/src/Faction/ui/AugmentationsPage.tsx
@@ -19,7 +19,7 @@ import { useRerender } from "../../ui/React/hooks";
 
 /** Root React Component for displaying a faction's "Purchase Augmentations" page */
 export function AugmentationsPage({ faction }: { faction: Faction }): React.ReactElement {
-  const rerender = useRerender();
+  const rerender = useRerender(400);
 
   function getAugs(): AugmentationName[] {
     return getFactionAugmentationsFiltered(faction);


myCatsName:
Did you notice if your PR also renders the change to rep in this part of the page? (the red column)
image

in the current steam version the "**buy**" color changes when the aug can be bought, but the checkbox and "**rep**" value do not. (in the screenshot the player has 1000 rep, the NFG rep and checkbox should update).

I'm just wondering if it's also updated with this PR, or might need similar sometime.

Caldwell-74:
it actually does
the full AugmentationsPage element rerenders

myCatsName:
Looks like this will close #695 and uses the same change suggested there by @rayanth

PR #748, opened 2023-08-25
UI: fix ActiveScripts Page map scripts with same arguments correctly

zornlemma:
# Bug fix

- For active script, currently could run multiple scripts with the same arguments simultaneously, thus this React `key` setting is not correct.
- For previous code, you could reproduce the bug by running multiple scripts with the same arguments, and click kill script in the `ActiveScripts` Page, it will render more script in the list as it should be. 
- fix tested locally.


d0sboots:
Looks like I missed this spot.

PR #753, opened 2023-08-27
DOCUMENT: ns.getScriptLog update to use "FilenameOrPID" and related ns.tail change

### LINKED ISSUE #714: ns.getScriptLogs() -- call with pid
Hello,

Is it possible to allow ns.getScriptLogs() to accept pid as an argument? Currently to get logs by pid requires calling ns.getRunningScript() at a RAM cost of 0.3 vs 0 for ns.getScriptLogs().

myCatsName:
closes #714

# Bug fix

- Updates ns.getScriptLog description to mention PID
- "" fn argument type from "string" to "filenameOrPID"
- updates ns.tail() to use the same language in description, i.e.  `the PID or filename` replaces `the fn`

script for testing attached. Save as  .js and run. nukes n00dles, writes and exes a script, fetches the PID and demonstrates use of either PID or filename and host to get the same results with both ns.getScriptLog and ns.tail()

[getScriptLogs PID update test.txt](https://github.com/bitburner-official/bitburner-src/files/12448850/getScriptLogs.PID.update.test.txt)

### DIFF
--- a/markdown/bitburner.ns.getscriptlogs.md
+++ b/markdown/bitburner.ns.getscriptlogs.md
@@ -9,14 +9,14 @@ Get all the logs of a script.
 **Signature:**
 
 ```typescript
-getScriptLogs(fn?: string, host?: string, ...args: (string | number | boolean)[]): string[];
+getScriptLogs(fn?: FilenameOrPID, host?: string, ...args: (string | number | boolean)[]): string[];
 ```
 
 ## Parameters
 
 |  Parameter | Type | Description |
 |  --- | --- | --- |
-|  fn | string | _(Optional)_ Optional. Filename of script to get logs from. |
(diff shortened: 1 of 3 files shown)


Snarling:
Another way to fix this would be how ns.kill handles it using two distinct function signatures instead of one function signature that can be used two different ways: https://github.com/bitburner-official/bitburner-src/blob/dev/markdown/bitburner.ns.kill.md which links to a second ns.kill documentation page.

I favor that two-signatures approach, but this is also a correct improvement so I'll just merge this as is.

myCatsName:
Oh, for stronger typing? ex. when someone formats the PID as a string it produces an error.... I hadn't thought of that, and it does produce an error
image


[`FilenameOrPID`](https://github.com/bitburner-official/bitburner-src/blob/dev/markdown/bitburner.filenameorpid.md) is invoked in a few places in case it should be changed. Though if it hasn't caused player confusion in the ns.tail context, it's probably intuitive (or at least not frustrating) to use, if not best practice, which we seem to agree.
image

PR #756, opened 2023-08-28
UI: Basic Map and Set support in ns.print/tprint

ficocelliguy:
Added basic support for printing maps and sets via print and tprint.
Note that the formatting I chose for maps clearly indicates they are not an object with object keys, but rather are entries with a key => value relationship. I am open to suggestions for different formatting styles, as well, since there isn't exactly a standard for the string representation AFAIK.

image

### DIFF
--- a/src/Netscript/NetscriptHelpers.tsx
+++ b/src/Netscript/NetscriptHelpers.tsx
@@ -220,15 +220,20 @@ function argsToString(args: unknown[]): string {
     const nativeArg = toNative(arg);
 
     // Handle Map formatting, since it does not JSON stringify or toString in a helpful way
-    // output is  "< key1 => value1, key2 => value2, >"
+    // output is  "< Map: key1 => value1; key2 => value2 >"
     if (nativeArg instanceof Map && [...nativeArg].length) {
-      out += "<";
-      nativeArg.forEach((key, value) => (out += ` ${key} => ${value},`));
-      return (out += " >");
+      const formattedMap = [...nativeArg]
+        .map((m) => {
+          const key = argsToString([m[0]]);
+          const value = argsToString([m[1]]);
+          return `${key} => ${value}`;
+        })
+        .join("; ");
+      return (out += `< Map: ${formattedMap} >`);
     }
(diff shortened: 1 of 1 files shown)


Snarling:
Yeah I'm not sure the exact format that would work best. Right now some issues I have with the format pictured:

1. Unnecessary closing comma at the end of the map list
2. There's no distinction between how a set is printing and how an array would print, so a Player might think a Set they are printing is actually an array and get confused when they can't do array things with it
3. Already a problem with 2d arrays, but if a Map or Set includes arrays, it will be hard to understand the structure based on what is printed.

What about using semicolons for separation so Arrays don't cause confusion, and including the type of object being printed to minimize the chance the player gets confused about the type of object they are working with?
e.g. the code:
```js
// Map with a number => array and an array => number
const map = new Map([
  [1, [1,2,3]],
  [[1,2,3], 1],
]);
ns.tprint(map)

// Set with a number and an array
const set = new Set([
  1,
  [1,2,3],
]);
ns.tprint("Map: ", map);
ns.tprint("Set: ", set);
```

Might print something like:
```
Map: < Map: 1 => 1,2,3; 1,2,3 => 1 >
Set: < Set: 1; 1,2,3 >
```

Also interested to see other suggestions

ficocelliguy:
Updated with the suggestions. Note that I use the current formatter to format the key & value of the map, which can be various data types:

image

Snarling:
Yeah I do like the recursive display of data in the map version, but I imagine it would error out if there's a circular reference to the same map, e.g.:

```js
const map = new Map();
map.set("self", map);
ns.tprint(map);
```

Right now the Set version is just directly showing the string'ed version of each member, instead of recursively geting the Terminal-appropriate representation of that member. Whichever way we go, Map and Set should be handled consistently to each other

ficocelliguy:
I looked into passing a recursive list of parent references to avoid the circular reference issue you mentioned, but it got really messy really fast. I decided to simply do the string representation of each key and value, as I expect that will take care of most user's basic use cases (and let them make their own, better, formatter if needed), and it also keeps the code simpler.

PR #758, opened 2023-08-29
DOCS: adjusted documentation for host parameter

VictorS:
modified description of host parameter, detailing the behavior when parameter is omitted.
can't run npm run docs, as I do not have npm installed.


VictorS:
installing npm for md generation

PR #760, opened 2023-08-30
UI: bladeburner.getactionrepgain() signature includes more type and default details

### LINKED ISSUE #65: Bladeburner.getActionRepGain() docs mismatch functionality
[smmalis37]
https://github.com/danielyxie/bitburner/blob/dev/markdown/bitburner.bladeburner.getactionrepgain.md specifies that the `level` argument is supposed to be optional, although It doesn't specify what the result will be in the case that this argument is omitted. However the current implementation requires a level value to be passed in, and will throw an exception if one is not.

[phyzical]
> hmm source wise it should be optional, and will default to the current actions level if nothing is provided,
> 
> what is the exception that is thrown?

[smmalis37]
> Taking a quick glance, I believe the exceptions that I'm seeing are all coming from the [helper.number call](https://github.com/danielyxie/bitburner/blob/371de216ea4bb07f83baa142f00f1acf5b4d3520/src/NetscriptFunctions/Bladeburner.ts#L162), as I believe it doesn't let null/NaN through to hit the later if.

myCatsName:
closes #65

# Bug fix

#65 said "the current implementation requires a level value to be passed in, and will throw an exception if one is not." and more specifically "I believe it doesn't let null/NaN through"

image

I tested some values and yes `null` and `NaN` throw exception, but a level value is not required. For my understanding of null, it seems like the behavior is correct (not acting as a number), so I don't suggest that change.
image

I did add a minor update to signature so level type and default might be more clear.

image

### DIFF
--- a/markdown/bitburner.bladeburner.getactionrepgain.md
+++ b/markdown/bitburner.bladeburner.getactionrepgain.md
@@ -18,7 +18,7 @@ getActionRepGain(type: string, name: string, level: number): number;
 |  --- | --- | --- |
 |  type | string | Type of action. |
 |  name | string | Name of action. Must be an exact match. |
-|  level | number | Optional action level at which to calculate the gain |
+|  level | number | Optional number. Action level at which to calculate the gain. Will be the action's current level if not given. |
 
 **Returns:**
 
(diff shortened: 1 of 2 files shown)


Snarling:
Not really "UI" category... But I can just change the title of the squashed merge commit.

PR #761, opened 2023-08-30
BLADEBURNER: fix NaN in success Chance

Caldwell-74:
fix 0/0 NaN in Bladeburner success chance calculation

let me know if you want to check/handle those differently :3

### DIFF
--- a/src/Bladeburner/Action.tsx
+++ b/src/Bladeburner/Action.tsx
@@ -223,7 +223,8 @@ export class Action {
     let low = real - diff;
     let high = real + diff;
     const city = inst.getCurrentCity();
-    const r = city.pop / city.popEst;
+    let r = city.pop / city.popEst;
+    if (Number.isNaN(r)) r = 0;
     if (r < 1) low *= r;
     else high *= r;
     return [clamp(low), clamp(high)];


myCatsName:
Do you have a way to test? And do you think real population === 0 is the cause of this problem?

I'm trying to get population to 0 , if you have a way to do that with dev tools, LMK.

I'm concerned maybe it's not the right fix.

city.pop and city.popEst should always be numbers (they have errors for NaN), so `const r = city.pop / city.popEst` should always be a number as well - `0/0` is typeOf number. Do you know why `high` would be NaN? (btw this should not use `let`, I think, the formula is const.)

I also assume `r=0` won't fix `0%-NaN%` because looking the rest of the function if (r < 1), I think only the lower number is affected.

And lastly, most BB tasks still can succeed with 0 population, right? Only Raid says "there must be an existing Synthoid community in your current city in order for this Operation to be successful." That case is handled differently in the same file : 
`if (this.name == "Raid" && inst.getCurrentCity().comms <= 0) {
      return 0; 
    }` for the successChance.

Caldwell-74:
I did a quick test with the savegame Bob provided in Bug-Reports and the UI was fixed through this
i didnt test the api but can do that now where iam home (did the pr in school)

population is the cause but thats a player made mistake
0/0 = NaN and typeof NaN is number
NaN is not smaller than 1 so it goes into the else and number *= NaN results to NaN
0 is not equal to NaN so it results to in the UI 0%-NaN

this is also all just a player facing/visual/ui NaN it doesnt affect the real success chance
that one is lowered to 0 by `competence *= this.getChaosCompetencePenalty(inst, params);`
wich ends up beeing
 ```
getChaosCompetencePenalty(inst: Bladeburner, params: ISuccessChanceParams): number {
//other code not relevant here
 return Math.pow(city.pop / BladeburnerConstants.PopulationThreshold, BladeburnerConstants.PopulationExponent);
```
wich returns Math.pow(0/1e9,0.7) = 0

so `return Math.min(1, competence / difficulty);`
returns Math.min(1, 0/something) wich is 0


for the let/const it works like this but if Snarling doesnt like the style, there are other ways iam happy to change it to how ever he wants to handle/check those


edit.: 
before ns.bladeburner.getActionEstimatedSuccessChance("Operations", "Investigation") returns

[0,null] 
and after 
[0,0]
 UI also looks fixed
image

zerbosh:
oh and for the let/const
thats the question how Snarling likes to handle NaN we could do a new variable or do a new function for it or iam sure there are many other ways

myCatsName:
>0/0 = NaN and typeof NaN is number
NaN is not smaller that 1 so it goes into the else and number *= NaN results in NaN
0 is not equal to NaN so it results in 0%-NaN

Ah, that's the part I did not know. Thanks for explaining. And I agree you fixed the visual bug, `0%-NaN%` now shows `0%`

To the second point, the competencePenalty means that if population===0 no tasks can ever succeed? That "could" make sense in "game logic" but it's not documented for the player I think, and it makes the Raid description misleading : "there must be an existing Synthoid community in your current city in order for this Operation to be successful." is true, but even having Synthoid communities with 0 population => 0% success. It's not your issue, but maybe needs documentation or formula update.
image

>for the let/const it works like this 

lol `var` works there too, but I bet you don't want that. I'm just teasing you though, it's NBD.

PR #773, opened 2023-09-01
GANG: Updates to Docs, UI, API

### LINKED ISSUE #38: Gangs feature needs documentation
[macdjord]
There is no out-of-game (and almost no in-game) documentation for how gangs are supposed to work. I managed to get through BN2 using [this](https://www.reddit.com/r/Bitburner/comments/e4z6ju/wip_crime_gang_bladeburner_guide/) two-year-old guide on Reddit, plus occasionally reading the source code. Now I'm on another BN, and I can't even figure out how to *start* a gang - I've joined the Slum Snakes faction, but the 'Form gang' button isn't showing up. The guide suggests that you need different amounts of negative karma to form a gang on different bitnodes, but we have (supposedly) complete documentation for the bitnode multipliers and 'karma required for gang' is not among them.

Version: a3c599e6

[zeddrak]
> As far as I know, you need approximately -54k karma to start a gang, This has been the same for all BN's I've played in.
> It's not documented anywhere that I could find either, nor explained well. I just used the API to constantly try to create a gang every few seconds.

[macdjord]
> Note that the question of 'Why can't I start a gang?' is only one part of what needs to be documented. The whole process and mechanics of gangs needs to be explained.

[phyzical]
> yeah part of the issue with this one is i thiiink its meant to be a motivation for the user to find out how/why in the source code, motivate them to learn about opensource
> 
> (encouraging users to understand why things don't work how they expect)
> 
> there is probably a consideration on spoilers also

### LINKED ISSUE #72: Minor UI Bug: Gang Ascension - conflict between UI "Are you sure?" and script ascension
[Penumbra69]
Bitburner v1.7.0 (b71418b0)

Gang ascension - UI and Scripts seem to fight with each other with the member Ascend "Are you sure?" dialog.

If you are using a script to ascend your gang members, but - like me - still enjoy watching numbers tick - you'll find an odd behavior when managing/viewing your gang.  If you have a member who is nearing your scripted ascension threshold, and you are in the UI on the gang screen, which means the member has achieved enough to ascend according to the game UI so you then click the "Ascend" button on the member to watch the factors tick and your threshold get reached; the "Are you sure" ascend dialog window (appropriately) vanishes when the script triggers the ascension and the gang member is no longer available to ascend.  Ok cool!

The issue comes in if you stay on the gang screen, and that same member again reaches a point where the game turns the "Ascend" button back on - at this point the game also re-displays (unhides) the Ascension "Are you sure?" dialog for the member without the user hitting the "Ascend" UI button a second time.

It seems the 'Ascend' "Are you Sure" dialog is improperly tracking show/hide state along with the `Ascend` UI button element.

### LINKED ISSUE #701: Write gang.md
Write information about the mechanic

myCatsName:
# Linked issues

closes #701
closes #38
fixes #72

# Documentation
A fairly wide update to gangs, mostly documentation and UI. In general trying to update facts, and trim small amounts of text to improve display and readability. I will try to list changes then demonstrate with before and after screenshots. 

- Add in game documentation for Gangs
- Add ns.gang.renameMember() method: - Players can rename gang members with API.
- Add ns.gang.getRecruitsAvailable() method: - Return how many recruits are available, used in player API and behind the scenes for UI formatting.
- FactionsRoot: - Gang faction is split into own box shown above other factions
- Bitverse BN-2 description: - Gangs do no offer all augments - TRP is available in BN-2 only
- Create new Gang modal : -Add cancel button - Use reactive faction name - bitflume/ abandoning the node also resets gang faction.
- Management Subpage: -Add "engage territory warfare" switch directly to management subpage -displays RecruitsAvailable or max message if "rep needed for next recruit" is not displayed -Update conditional display of Ethical Hacking and Vigilante in description  - Edit texts
- GangMemberCardContent description: - Ascension bonus is shown before confirmation
- AscensionModal: - Add cancel button - Add line breaks for visibility - "You ascended {member}" => "{member} ascended"
- RecruitModal: -spellCheck="false" - recruiting without a name is disabled, no longer gives "you already have a gang member with that name" error
- TerritoryInfoModal: - Mentions .gang API
- Territorial Warfare Task: - more clear description
- ns.gang.recruitMember(): - Improve error message if no name or name is already used

# Bug fix
Also includes a fix for issue #72 , the Ascension Modal re-opening when closed with API ascension. My fix is a clean-up useEffect in the modal, simply running props.onClose() method. However, React doesn't like this, and the linter makes a complaint that `props` should be in the dependency array or not used. Adding it to the array changes the behavior, and is not a fix. Tried other fixes. I tried destructuring with something like `const { onClose } = props` (https://stackoverflow.com/a/68907940) and OK got  the same behavior as using `props.onClose()` but then the linter wanted onClose in the dependency and so on lol. I found a lot of discussion on the same issue, and commented a link. Ultimately I couldn't solve the bug while avoiding the linter complaint and for that reason [despite really wanting to avoid it] added a comment for the linter to ignore that line, and profuse comments to explain. 

# Demo

NEW :
Separate box for Your Gang on factions page
NEW :
image

OLD:
start a gang old
NEW:
create gang modal new

OLD:
gang management subpage OLD

NEW:
manage last

NEW:
maximummmessage

OLD:
ascend modal OLD

NEW:
ascend modal new

OLD:
territory warfare task OLD

NEW:
war task new


src/NetscriptFunctions/Gang.ts
```diff
@@ -134,17 +167,28 @@ export function NetscriptGang(): InternalAPI<IGang> {
...
+    getRecruitsAvailable: (ctx) => () => {
+      const gang = getGang(ctx);
+      return gang.getRecruitsAvailable();
+    },
     recruitMember: (ctx) => (_memberName) => {
       const memberName = helpers.string(ctx, "memberName", _memberName);
       const gang = getGang(ctx);
       const recruited = gang.recruitMember(memberName);
-      if (recruited) {
+      // TODO : ns.gang.recruitMember() crashes with error, can use top return instead?
```
myCatsName: This TODO comment isn't relevant really, it's the same behavior as any bad call missing args. Thought I had removed this line.

Snarling:
Nice! Will probably take a bit to review, but based on the description, the changes all sound good.

myCatsName:
Awesome. It was a bit of work, so I appreciate you taking the time to review as needed. Happy to help adjust or explain as needed.

src/Gang/ui/RecruitButton.tsx
```diff
@@ -15,24 +14,29 @@ interface IProps {
...
     return (
       <Box display="flex" alignItems="center" sx={{ mx: 1 }}>
         <Button disabled>Recruit Gang Member</Button>
-        <Typography sx={{ ml: 1 }}>{formatRespect(respect)} respect needed to recruit next member</Typography>
+        {gang.getRecruitsAvailable() === -1 && (
+          <Typography sx={{ ml: 1 }}>Maximum gang members already recruited</Typography>
+        )}
+        {gang.getRecruitsAvailable() !== -1 && (
+          <Typography sx={{ ml: 1 }}>{formatRespect(respect)} respect needed to recruit next member</Typography>
+        )}
```
Snarling: If changing getRecruitsAvailable to return 0 when you're at max members and getRespectNeededToRecruitMember changed to return Infinity, then respect needs to be used as the condition here.
Also a ternary works better here than a condition + anticondition (formatting on suggestion might be wrong, typed on github and not in an IDE)

```suggestion
        {respect === Infinity ? (
          <Typography sx={{ ml: 1 }}>Maximum gang members already recruited</Typography>
        ) : (
          <Typography sx={{ ml: 1 }}>{formatRespect(respect)} respect needed to recruit next member</Typography>
        )}
```

src/Gang/ui/RecruitButton.tsx
```diff
@@ -15,24 +14,29 @@ interface IProps {
...
     );
   }
 
   return (
     <>
-      <Box sx={{ mx: 1 }}>
+      <Box display="flex" alignItems="center" sx={{ mx: 1 }}>
         <Button onClick={() => setOpen(true)}>Recruit Gang Member</Button>
+        <Typography sx={{ ml: 1 }}>
+          Can recruit {gang.getRecruitsAvailable()} more gang member{gang.getRecruitsAvailable() === 1 ? "" : "s"}
```
Snarling: Storing recruitsAvailable in a variable once per render and reusing it makes more sense than calling the function multiple times (although with the change above it's only 2 calls now, so not a huge deal)

myCatsName:
Heads up that when I went back to dev build after testing and using this branch, I seemingly had a compile error about  renameMember() not being in an API file (because it's not on dev, obviously). I had to go dev menu > stop gang , and I think that is what fixed the issue for me. Just a note in case that could be a sign of something more critical I missed.

Snarling:
> Heads up that when I went back to dev build after testing and using this branch, I seemingly had a compile error about renameMember() not being in an API file (because it's not on dev, obviously). I had to go dev menu > stop gang , and I think that is what fixed the issue for me. Just a note in case that could be a sign of something more critical I missed.

Was that just in the console for your ongoing `npm run start:dev`? It does seem to run into issues sometimes when switching branches after your dev server is already up. I don't know exactly what triggers it, I thought it was usually related to the two branches having a discrepancy on whether an entire file exists, but that wouldn't be the case on this one. Probably not an issue though. Whenever it happens I just close out of the dev server and restart it, it shouldn't have any impact on anything else.

Snarling (changes requested):
So for API functions, it's most important for us to get them right on initial implementation since there's a lot more friction involved in making changes further down the line. So most of my feedback is on that:

* getRecruitsAvailable - I think this should return 0 (correct numerical value) when you're at max members, instead of a "special flag" return of -1.
* To support the above, I think getRespectNeededToRecruitMember should be exposed in the player facing API, and should also return Infinity (correct numerical value) when player is already at max members. Also it can be renamed to something a bit more friendly if you can think of a shorter/better name, prior to adding it to the API is really the last time we can decide to rename it.
* In general, ns API functions should throw when the input is clearly wrong, e.g. attempting to rename a member to empty string. For input that is sometimes-wrong (like giving the wrong current member name when trying to rename), it could go either way, but personally I lean towards throwing when the function is used incorrectly instead of e.g. return false, as it forces players to write their scripts in a way to ensure proper usage (e.g. get the member names and use that to determine what renames are needed before trying to rename them). Would like to hear some other opinions on throw vs return false though.
* I actually think that eslint disable on the useEffect dependency is not needed. Does it not work correctly if you use `props` or a destructured `onClose` as part of the dependency list?

myCatsName:
Didn't realize the revisions would push to the PR immediately lol that happened a bit premature.

>getRecruitsAvailable - I think this should return 0 (correct numerical value) when you're at max

The `-1` case expanded the scope unnecessarily; I was caught up in my own use case. Thanks.

- Added ns.gang.respectToNextRecruit() , it fits next to a few other .respect... methods. It's also returned as part of ns.gang.getGangInformation()
- conditional rendering "Your gang lost/will lose {} respect." lines on the ascension confirm and summary modals.
>ns API functions should throw when the input is clearly wrong... For input that is sometimes-wrong...

I know it would be work to build and maintain, but so often I think "style guide."  A bit of an aside though.
>Would like to hear some other opinions on throw vs return false though.

I have felt OK with the returns suggested. Though on the other hand I do agree with you and questioned both the existing and updated behaviors. As suggested, I think the behaviors can be described as my interpretation of what you described (before you described it). In these, player gets feedback if they haven't provided all parameters, and less obtrusive feedback if what they supplied won't result in success.  Maybe I under or over thought the details. 

>return Infinity, then respect needs to be used as the condition here

Done. I had also made a version checking gang size vs limit constant, in case it would be better to avoid relying on that function always returning Infinity.

>I actually think that eslint disable on the useEffect dependency is not needed. Does it not work correctly if you use props or a destructured onClose as part of the dependency list?

\>.<
tl; dr No. I'll edit in details.

There *may* be a way to avoid it, but I suspect outside the experience available, for my part at least. It's not a simple detail though, and it made me lol how frequently replies suggested or resorted to various ways of disabling. This is what is linked in the code- https://stackoverflow.com/questions/55840294/how-to-fix-missing-dependency-warning-when-using-useeffect-react-hook
I tried to implement most suggestions. Here's a relevant blog from a React ~dev~ original author - https://overreacted.io/a-complete-guide-to-useeffect/#dont-lie-to-react-about-dependencies and it is a dense section lol. According to him this is called "lying about dependencies."
To naively summarize the issue - `props.onClose()` is not guaranteed to be available to useEffect, in which case useEffect wouldn't know what to do. In our case, it practically is guaranteed to be there, it's written into our modal. And given an empty [] dependency array, the useEffect cleanup runs 1 time when the component unmounts (ie I don't believe it can be a leak). 
`[props]` dependency instead tells the useEffect to render every time there is a change to props, which makes a problem. In general changing from [] changes the behavior of a clean up useEffect.
I think there are ways to avoid -- I am not trying to get a reputation for React code held together with duct tape and disabled linters XD But genuinely outside of what I can do or learn presently. Searching up the exact error there are a lot of topics, here are a couple highlights-
image

image

Caldwell-74:
> Added ns.gang.respectToNextRecruit() , it fits next to a few other .respect... methods. It's also returned as part of ns.gang.getGangInformation()

what about gang.getRecruitmentCost()?
i dont know why but either the "To" or the length (4 words) set me off

myCatsName:
@zerbosh Thanks. 
It's not a cost, just a threshold, a number you pass; but it's not reputation taken from you, which I think about "cost."

>either the "To" or the length (4 words) set me off

The "To" part for me, it takes a little getting used to we might say :P  `.respectToNextRecruit()` Having autocomplete suggestions with .respect should help it be discovered by players. It felt good enough to suggest, and I will think about it again tomorrow. 😴

myCatsName:
Empty dependency example from the React author 
image

From a [React git issue](https://github.com/facebook/react/issues/16265) 
>image
image
image
image

I did try destructuring (among other fixes), as I said up top, functioned but unable to avoid the linter.

myCatsName:
confronting useEffect

Snarling:
This still applies when using props directly, but it's wonkier so for discussion I will assume we are destructuring and using onClose directly.

> To naively summarize the issue - props.onClose() is not guaranteed to be available to useEffect, in which case useEffect wouldn't know what to do. In our case, it practically is guaranteed to be there, it's written into our modal. And given an empty [] dependency array, the useEffect cleanup runs 1 time when the component unmounts (ie I don't believe it can be a leak).

The issue is not whether onClose will exist (that's guaranteed by types), it's whether it's still an up-to-date version of the onClose function.

Imagine that the onClose function could have different values. As written now, this modal component sets up its useEffect using the first onClose that is sent in, and then it never updates. If onClose changes, and then later the modal is unmounted, it will call the original onClose and not the current one.

> [props] dependency instead tells the useEffect to render every time there is a change to props, which makes a problem.
 In general changing from [] changes the behavior of a clean up useEffect.

This is a symptom of a different problem. If we're expecting onClose to never change, and adding onClose to the dependency list of this useEffect causes a change in behavior, then that means onClose is changing (otherwise it would be functionally identical to empty dependency list). Even if onClose is remaining functionally identical, it is actually changing value, causing the useEffect to trigger when we don't want it to (~~and causing an entire rerender of the modal, due to a props change, that is not desired~~ Edit: the rerender happens anyway because parents cause their children to rerender).

The correct solution here is to use useCallback when originally defining onClose in the parent component. If onClose actually does not change, this should be possible using an empty dependency array in the useCallback. Then, with an onClose that is not changing value, we can add it to the dependency list on the modal's useEffect with no change in behavior from empty dependency array.

See https://github.com/facebook/react/issues/16265#issuecomment-517518539 from the same issue you linked, Also discussed on #640 although most of that is probably in resolved comment chains, also some of it might have been on discord. In those conversations I initially favored empty dependency arrays as well.

d0sboots:
+1 to useCallback. That was what I was forgetting in the other PR. If changing the effect dependency causes a behavior change, then you have a real bug, not just a lint error.

myCatsName:
> +1 to useCallback. That was what I was forgetting in the other PR. If changing the effect dependency causes a behavior change, then you have a real bug, not just a lint error.

Snarling spent some time in DMs to explain and offered a hand in refactoring. It went beyond what I would have been comfortable or familiar doing; and clearly wasn't simple or taken lightly when the linter changes and custom hooks were implemented in issue 640 to provide the methods currently in place. 

As to the former CCT modal PR, I was of course curious to see that again in hindsight and I think I'm in the clear :P 

image

Snarling (approved):
Made some changes after discussing on discord.
937dda2: Implemented the React / useCallback changes mentioned in previous comment.
c7a025e: Did a few things:
* Renamed respectToNextRecruit to respectForNextRecruit
* Renamed internal function to match.
* Moved some magic numbers in the internal respectForNextRecruit over to GangConstants.
* Tweaked the renameGangMember function. Behavior change is that now it throws if the player passes in empty string or same-names, since those are always invalid. Other input will still return false on failure, since they might fail due to a script not having up to date info (e.g. if player is performing actions in the ui that conflict with their gang management script)

Edit: e8723bb: Fixed an error from previous commit where I missed renaming in ramCostGenerator. Also renamed the gang.renameGangMember function to just gang.renameMember

myCatsName:
>ns.gang.renameMember()

@Snarling 
FWIW I really tried to avoid using "member" alone because ... you know. It's another term for a penis XD

PR #774, opened 2023-09-02
CORPORATION: add issueNewSharesCooldown and makesMaterial

### LINKED ISSUE #506: getCorporation() lacks issueNewSharesCooldown property
Hi there,

Steam, BB version 2.2.2

here is my script:

```
const corp = ns.corporation;

const getCorpInfo = (verbose = false) => {
		const corpData = corp.getCorporation();
		const cConst = corp.getConstants();
		const cData = {
			divisions: corpData.divisions,
			ownedShares: corpData.numShares,
			isPublic: corpData.public,
			sharePrice: corpData.sharePrice,
			shareSaleCooldown: corpData.shareSaleCooldown,
			totalShares: corpData.totalShares,
			dividendRate: corpData.dividendRate,
			issueNewSharesCooldown: cConst.issueNewSharesCooldown,
			upgradeNames: cConst.upgradeNames,
			unlockNames: cConst.unlockNames, // one-time unlock upgrades
			researchNamesBase: cConst.researchNamesBase,
			researchNamesProductOnly: cConst.researchNamesProductOnly,
			materialNames : cConst.materialNames,
		}
		if (verbose) {
			ns.tprint(`Corporation info: \n\t sharePrice\t\t = ${ns.formatNumber(cData.sharePrice, 2)}` +
				`\n\t shareSaleCooldown\t = ${ns.tFormat(cData.shareSaleCooldown * 200)}` +
				`\n\t issueNewSharesCooldown\t = ${ns.tFormat(cData.issueNewSharesCooldown * 200)}` +
				`\n\t totalShares\t\t = ${ns.formatNumber(cData.totalShares, 3)}` +
				`\n\t divisions\t\t = ${cData.divisions}` +
				`\n\n\t materialNames\t\t = ${cData.materialNames}` +
				`\n\n\t Common researches\t = ${cData.researchNamesBase}` +
				`\n\n\t researchNamesProductOnly\t = ${cData.researchNamesProductOnly}` +
				`\n\n\t One-time upgrades\t = ${cData.unlockNames}` +
				`\n\t upgradeNames\t\t = ${cData.upgradeNames}\n`

			);
		}
		return cData;
	}
```

All other data changes, but issueNewSharesCooldown allways says 12 hours :)

I'm not a javascript programmer (yet), so please, be kind :)

Caldwell-74:
adds issueNewSharesCooldown to getCorporation and makesMaterials to getDivision

makesMaterials should be added to getIndustryData but makesProducts is on getDivision already so it wouldnt makes sense to split it 
moving both would be an API break i didnt want to make for such a detail but i would put it on the list for the next  bigger Update

closes #506


d0sboots:
What about adding both makesMaterials *and* makes Products to getIndustryData now? Then they're both in the right place, and together, and the only downside is a little duplication.

myCatsName:
“Duplication may be the root of all evil in software.” - Robert C. Martin , Clean Code

Snarling:
Yeah I agree with @d0sboots here, they make more sense on getIndustryData since they are applicable to all divisions within an industry. getDivision should just have the info that is specific to that one division within the industry.

I think for now duplicating it is fine, but we should remove the old out of place entry on getDivision in the future, I guess at the next API break.

Caldwell-74:
with them on getIndustryData i dont see a reason to add makesMaterials to getDivision at all

d0sboots:
> with them on getIndustryData i dont see a reason to add makesMaterials to getDivision at all

Yeah, that was the idea.

d0sboots:
> “Duplication may be the root of all evil in software.” - Robert C. Martin , Clean Code

It's much better in this case than making a permanently suboptimal API choice because of a prior bad API choice, or forcing a breaking API change during a minor version bump.

Evolving the BB API has different challenges than other things because it's a push model, not a pull model. (Technically, I suppose we *could* keep every version available in Steam, but I'm not sure how many it allows, and it's really not designed for that.)

Snarling:
Just pushed a quick commit to remove the duplicate makesMaterials from Division, which was not used anywhere.

PR #777, opened 2023-09-02
CORPORATION: fix production limitation

Caldwell-74:
needs testing

this aims on fixing the bug that only the first material that a division produces limit gets used and others get ignored
this also tries to limit the actual production on the highest limit 
if a div produces 2 materials a and b 
and a is limited to 100 and b to 1000 but it could produce 2000 
then it should only buy stuff for 1000 and 900 a would be voided

### DIFF
--- a/src/Corporation/Division.ts
+++ b/src/Corporation/Division.ts
@@ -16,7 +16,6 @@ import { JSONMap, JSONSet } from "../Types/Jsonable";
 import { PartialRecord, getRecordEntries, getRecordKeys, getRecordValues } from "../Types/Record";
 import { Material } from "./Material";
 import { getKeyList } from "../utils/helpers/getKeyList";
-import { isDefined } from "../types";
 
 interface DivisionParams {
   name: string;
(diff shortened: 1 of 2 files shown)


d0sboots:
The logic in your description seems weird to me. A limit should respect the *lowest* of all the bounds - if we assume that `a` and `b` are produced at the same rate, but `a` has a limit of 100, that should limit the whole process to 100.

I don't disagree that throwing away the excess might be more *friendly*, but nothing else in Corps works that way, and managing warehouse space is a key part of the challenge. If you *were* going to get rid of the excess, I think it should be via auto-sale (but that has its own complications, and really should be under user control).

Caldwell-74:
the problem is we dont know anything about the relation of input to output except something goes in and something comes out "is A and B used to produce C and D is a byproduct?"
so there is def alot of room to argue about how it should behave my thinking was:
on a Farm, "food" gets produced but "plant" matter is a waste/byproduct
you can store up to a given amount(limit) and sell it to others and the rest lands on a compost(voided) 
but you can still produce max amounts of "food" 
 sure that doesnt narratively fit into each division but i wont write up a different logic for each type xD

d0sboots:
I agree with the convenience of it for Agri, but it's mixing two mechanics with the same setting so it has a lot of potential confusion.

On one hand, it acts as an overall production limiter, *scaling down* input materials to reduce the output. On the other hand, it acts as an output limiter, *throwing away* produced materials to reduce the output.

These are two different mechanisms jammed together, and either can work alone. You could make the whole thing operate on throwing away output materials, and that would be viable (but wasteful). You can also make it entirely based on scaling down, which is what I propose.

The Agri case can be handled by selling your excess plants (which I think is what we should encourage).

Caldwell-74:
i can take the first section out and just let it keep maxProd with alot of waste
with how cheap everything is in a good corp this shouldnt be a super big deal 
but if we take the lower limit approach then we dont need individual limits at all
one input field/setting per city would be the only thing that makes sense then
this would be code wise alot easier and also an unused feature or for the most part


edit.: another point to consider would be products with a city wide limit and how they are affected in general
the only two use cases for limits on materials are 
to void materials you cant really sell 
or to limit warehouse used so products can be produced

PR #778, opened 2023-09-02
CORPORATION: improve State description

Caldwell-74:
just a tiny change to make it more clear that when getCorporation returns "PURCHASE" than this state hasnt happened just yet but will be the next that gets processed 
removing a bit "immersion" for better clarity

### DIFF
--- a/src/Corporation/ui/DivisionWarehouse.tsx
+++ b/src/Corporation/ui/DivisionWarehouse.tsx
@@ -58,7 +58,7 @@ function WarehouseRoot(props: WarehouseProps): React.ReactElement {
     props.rerender();
   }
 
-  // Next state wich will be processed:
+  // Next state which will be processed:
   let stateText;
   switch (division.state) {
     case "START":
--- a/src/Corporation/ui/DivisionWarehouse.tsx
+++ b/src/Corporation/ui/DivisionWarehouse.tsx
@@ -58,23 +58,23 @@ function WarehouseRoot(props: WarehouseProps): React.ReactElement {
     props.rerender();
   }
 
-  // Current State:
+  // Next state wich will be processed:
   let stateText;
   switch (division.state) {
     case "START":
-      stateText = "Current state: Preparing...";
+      stateText = "Next state: Preparing...";
       break;
     case "PURCHASE":
-      stateText = "Current state: Purchasing materials...";
+      stateText = "Next state: Purchasing materials...";
       break;
     case "PRODUCTION":
-      stateText = "Current state: Producing materials and/or products...";
+      stateText = "Next state: Producing materials and/or products...";
       break;
     case "SALE":
-      stateText = "Current state: Selling materials and/or products...";
+      stateText = "Next state: Selling materials and/or products...";
       break;
     case "EXPORT":
-      stateText = "Current state: Exporting materials and/or products...";
+      stateText = "Next state: Exporting materials and/or products...";
       break;
     default:
       console.error(`Invalid state: ${division.state}`);
--- a/src/Corporation/ui/DivisionWarehouse.tsx
+++ b/src/Corporation/ui/DivisionWarehouse.tsx
@@ -62,19 +62,19 @@ function WarehouseRoot(props: WarehouseProps): React.ReactElement {
   let stateText;
   switch (division.state) {
     case "START":
-      stateText = "Next state: Preparing...";
+      stateText = "Next state: Preparing";
       break;
     case "PURCHASE":
-      stateText = "Next state: Purchasing materials...";
+      stateText = "Next state: Purchasing materials";
       break;
     case "PRODUCTION":
-      stateText = "Next state: Producing materials and/or products...";
+      stateText = "Next state: Producing materials and/or products";
       break;
     case "SALE":
-      stateText = "Next state: Selling materials and/or products...";
+      stateText = "Next state: Selling materials and/or products";
       break;
     case "EXPORT":
-      stateText = "Next state: Exporting materials and/or products...";
+      stateText = "Next state: Exporting materials and/or products";
       break;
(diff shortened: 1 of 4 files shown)


myCatsName:
No offense but I don't think that makes much sense. And the effect in action (as state cycles on corp division page) IMO is visually worse.

>when getCorporation returns "PURCHASE" than this state hasnt happened just yet

`"Current state: Purchasing materials..."` tells us Purchase hasn't happened just yet. It seems clear.

calling ns.corporation.getCorporation() return includes `"state":"SALE"` which I see in the source means

`case "SALE":`
`      stateText = "Current state: Selling materials and/or products...";`

That's as clear as can be. `"Next state: Selling materials and/or products"` would not be....

Caldwell-74:
we had quite a few people getting confused by "its saying SALE but it didnt actually sell anything yet"
having (without bonus time active) 2 seconds where it says "PURCHASING: Purchasing materials" can be very confusing
"did it buy any materials yet, will it buy any materials next game tick?" 
the actual processing of PURCHASE happens at the start of the first tick where PRUCHASE isnt shown anymore the full time of showing PURCHASE is all just the time leading up to it

Caldwell-74:
but if you have any suggestion how it can be worded so people can clearly understand what effect their action will have at that moment then sure iam always open for such

myCatsName:
Change the state variables returned by corp.getCorp() to also use the gerund form [the "ing" ending on a verb tells us something is in action, making it clear the action has not resolved]. ex "SALE" => "SELLING"

Caldwell-74:
but that would not make it clear that nothing of the state has happend yet
and that a change can still be made and take affect this corp cycle

myCatsName:
>a change can still be made and take affect this corp cycle

That's a detail for documentation, not a state variable description. 

>that would not make it clear that nothing of the state has happend yet

The mechanic should be explained in documentation; variables and such only support and remind. A gerund is about as clear a way as English provides to tell us something is happening, but unresolved. If that's clearer than the current state variables, cool. 

I think this can lead to change, but I don't think this PR should be a change. My opinion with full respect to you.

myCatsName:
For example it could be more like Production multiplier.
image
image

- tooltip on hover of the text
- Helper icon opens a modal with as much information as needed 

The "Current state" line doesn't use either of those (I put an icon in just to demo). Lots of room for details and clarity.

Caldwell-74:
The UI uses your gerund?
the only change there is that its specifies thats its not doing it currently but its coming up

and changing state names is a hard API break

src/Corporation/ui/DivisionWarehouse.tsx
```diff
@@ -58,23 +58,23 @@ function WarehouseRoot(props: WarehouseProps): React.ReactElement {
     props.rerender();
   }
 
-  // Current State:
+  // Next state wich will be processed:
```
d0sboots: which

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -7484,7 +7484,7 @@ interface CorporationInfo {
   dividendTax: number;
   /** Your earnings as a shareholder per second this cycle */
   dividendEarnings: number;
-  /** State of the corporation. Possible states are START, PURCHASE, PRODUCTION, EXPORT, SALE. */
+  /** The next state to be processed. Possible states are START, PURCHASE, PRODUCTION, EXPORT, SALE. */
```
d0sboots: Since this is a comment, we have more room for details. I'd add a sentence like: "I.e. when the state is PURCHASE, it means purchasing will occur during the next state transition."

d0sboots (approved):
Nice, this clears up a long standing confusion.

I agree about not changing the enums, that would break every Corp script for little gain.

myCatsName:
IDK what to tell you lol, I would not change the state names or the case descriptions, I am not among those who are confused by what is currently used. 

But the PR language does confuse me, and (seemingly) changes what is being reported rather than clarifying it.

While you have a better understanding of the system, I think I identified a better way to get information to the player with what I shared.

myCatsName:
@ dos, Approving PRs while there is conflict in comments does not feel collaborative.
[Edit to be less derailing.]

src/ScriptEditor/NetscriptDefinitions.d.ts
Snarling: Agreed, also I think this is the main place to make the clarification, I'm not sure the edits in the UI were necessary (were people confused by the UI description, or mostly just by the state in scripts?)

Also `npm run doc` will be needed to generate the new .md from this doc change

myCatsName: I really think clarity here needs to be supported by more comprehensive information update, and the changes being discussed are at best trading one confusion for another. 
"Waiting to purchase" is state, "next state selling" is [e.g.] nextState. The property used in game is `.state`. At least it still strikes me as confusing and less descriptive, for what it's worth.
> I think this is the main place to make the clarification

In the description for ns.corporation.getCorporation().state? :P IMO, if it's not clear from the Divisions subpage, players aren't finding it.
image
currently one of the few pages with italics and cool formula. cool. So why not add a Gantt chart or something that makes it super clear the order of operations and player's options during each state and the effects of each transition; or a process flow chart [presumably as a helper modal, as implied previously]. If it's causing confusion and has such fine detail, an extra sentence buried in a property is likely insufficient.

Just trying to avoid kicking a problem down the road and imagining alternatives.

Caldwell-74: > Since this is a comment, we have more room for details. I'd add a sentence like: "I.e. when the state is PURCHASE, it means purchasing will occur during the next state transition."

image

PR #782, opened 2023-09-05
CORPORATION: Rework share price calculation

### LINKED ISSUE #727: Corporation Stock Price one billion times higher than it should be (Bug?)
Version: Bitburner 2.4:

The game sets the stock price to high by 9 orders of magnitude.

Steps: 
1. Start corporation paying full150B.
2. Create any division that turns a profit (profit probably optional)
3. IPO
4. Buyback **ALL** shares.

Result: Stock price will increase endlessly for hours.

Why? 
getTargetSharePrice() is coded as follows: 
`   this.valuation / (2 * (this.totalShares - this.numShares) + 1);`

Owning all of the shares in your own corporation turns this into valuation / (2 * (1B - 1B) + 1) i.e. valuation / 1;

So, if the valuation for your corporation is 10B, then the stock price will trend toward 10B per share (!!!) 

If I had to guess I think this code is supposed to be like so:
`   this.valuation / (2 *  this.totalShares - this.numShares);
`
If the player doesn't own any shares, both result in the same result of valuation / 2B. 
If the player owns all the shares, the existing code results in valuation, this version results in valuation/1B (a billion times less). 

I do think it's a bug, and I think it's kind of important. I was able to turn 150B into 150000Q in less than 24 hours on a bitnode with a 50% penalty to corporate valuations.  Even a player is bad at corporations they would be able to use this method to make absurd money in a day or two, making the a lot of the game irrelevant since you have virtually infinite money and faction reputation. 

Also: Note that when you IPO the game calculates the share price like so:
`const initialSharePrice = corp.valuation / corp.totalShares;`

So if your company is public and you own all the shares, its worth one billion times more than if your not public and you own all the shares. This is another indicator to me that the publicly traded share price when you own all the shares is unintended.

jjclark1982:
Fixes #727 - an exploit where corporation share price can trend to infinity when the CEO owns 100% of shares.

This branch is available for testing at https://jjclark1982.github.io/bitburner-src/

### Summary of Changes

- Corporation share price uses a new formula that is always proportional to valuation.
- Corporation UI shows expected effects on share price when issuing, buying, or selling shares.
- Buying back shares uses the same market pricing calculation as selling shares.
- Issuing new shares uses the average share price in the range, instead of a fixed 20% penalty.
- Cooldown for issuing new shares reduced from 12 hours to 4 hours (but grows longer as total shares increase).
- Cannot start a new Corporation while the current one has "Sell Shares" on cooldown.

### Implementation

This PR adds a new `Corporation` property `investorShares` tracking the amount of shares held by seed investors and round investors.  

New unit tests are added to confirm that every action that manipulates shares preserves the equality

$$
{\tt totalShares} = {\tt numShares} + {\tt issuedShares} + {\tt investorShares}
$$

This allows us to ensure that share price and "market capitalization" are always proportional to corporation valuation.

#### Current Share Price Formula
Currently, the price of one share is equal to the entire valuation when the CEO owns 100% of shares. This means the market capitalization is 1 billion times higher than the valuation. With repeated share issuing, this can grow to infinity in a few days, even when the corp is producing nothing.

$$
{\tt sharePrice} = \frac{{\tt valuation}}{2 \times (\textrm{num shares not owned by CEO}) + 1}
$$

#### Proposed Share Price Formula
I propose a formula that scales with CEO ownership while remaining proportional to total valuation.

$$
{\tt marketCap} = {\tt valuation} \times \left(\frac{1}{2} + \sqrt{\frac{\textrm{num shares owned by CEO}}{\tt totalShares}}\right)
$$

$$
{\tt sharePrice} = \frac{\tt marketCap}{\tt totalShares}
$$

Stock Price

#### Rationale for this formula

- Market cap is always within 50% of valuation, so there is no more limitless growth.
- Share price obeys appropriate laws of supply and demand for a single-player simulation, resulting in natural penalties to sale price for owning too little, and natural penalties to buyback price for owning too much.
- Selling and buying back stock still works as a "piggy bank" between augmentations, but this is limited to the corp's market cap, so there is incentive to raise the valuation.

### UI Changes

The breakdown of share categories on the overview now includes `investorShares`:

total-shares-breakdown

Cooldowns are visible:

cooldown

Stock action UIs have updated wording and show expected results:

01-create-corp
02-find-investors
03-go-public
04-issue-shares
05-sell-shares
06-buyback-shares


myCatsName:
Awesome job to make those changes yourself. I hope it will be accepted as a positive balance change and fixing an exploit.

jjclark1982:
I have removed valuation changes from this PR, to focus on stock price changes only. Here are my notes about possible valuation re-balancing.

Valuation is based on
- funds
- projected profits
- synergy multiplier (1.1 ^ number of divisions)
- bitnode CorporationValuation multiplier

#### Current Valuation Formula (pre-IPO)
```js
valuation = bitnode multiplier * synergy * (10e9 + funds/3 + (profit * 17.5 hours))
```

#### Current Valuation Formula (post-IPO)
```js
valuation = bitnode multiplier * synergy * (funds + (profit * 4.72 hours))
```

#### Proposed Valuation Formula
```js
valuation = bitnode multiplier * (funds + (synergy * (profit * 24 hours)))
```

Snarling:
The change to the formula is definitely good.

There is a lot of unrelated stuff to review here and I'm just now looking it over, so it might be a little bit. So far what I'm seeing looks like solid changes though.

jjclark1982:
Thanks for reviewing! Here's a summary of things that have moved around:

- `getInvestmentOffer()`, `acceptInvestmentOffer()` and `goPublic()` moved from NetscriptFunctions (and their respective UIs) to Corporation/Actions. Now there is only one implementation of each, and it is unit-tested.
- Error checking in `SellShares()` and `BuyBackShares()` moved from Corporation/Actions (and their respective UIs) to Corporation object. Now there is only one implementation of each.
- `calculateShareSale()` can now handle negative numbers. It is still an iterative calculation that gets slower for large numbers of stock, so I have kept the `1e14` limit in place and made sure to reset the number of shares to 0 in closed modals to avoid continuously running this calculation in the background.
- In addition to the main changes to 8 modals shown above in screenshots, a few unrelated Corp UIs now use the `<Money>` or `<MoneyCost>` element where applicable.

Snarling (changes requested):
Sorry that it took so long to get a review out for this. I do see some issues, not with the main point of the PR but with some of the other auxiliary changes that were made.

1. Improper usage of "Actions": Actions functions, including some of the new ones, intentionally throw errors when they fail to do the thing they are trying to do. However, when you are using these functions in the UI, you are not catching those errors, these are just unhandled exceptions.
2. Some of the newly implemented "Actions" don't follow that design paradigm above - instead of throwing errors, they intentionally return false or return some specific thing to indicate failure. I see there were already a couple of Actions that fail silently (which was news to me), but making these functions act consistently makes it easier to use them, since we can expect a certain behavior from them (e.g. Actions can be called directly from a netscript API function, and will automatically generate a human readable error on failure). Functions that don't function this way can be implemented as Corporation class methods.
3. Not a big issue and I wouldn't kick back a PR for this, but the new functions that return tuples are kind of gross since the return items are just arbitrarily indexed values on the returned array. e.g. canBuybackShares or calculateShareBuyback. There are some existing nearby corp functions that already return something like this, so I get why your added functions here did that as well. If keeping tuple returns, you can label the items in the tuple to make it somewhat less gross: image
Although I prefer to use an object return when multiple data points need to be returned, since keys can be used to directly name the data points.
4. The weird setTimeout thing for interacting with the game via the UI is weird and should not be needed. Every other UI interaction within the game is synchronous.

Since it took me so long to review this, I don't mind making some of the adjustments myself, just let me know if you'd like me to do that (or in a couple days I'll just go ahead if I haven't heard back).

jjclark1982:
@Snarling Thanks! I do want to implement proper typing and exception handling for those functions. The trouble I'm running into is:

Stock transactions depend on the stock price right before the transaction, so exposing this with live updates in the modal is very helpful to users. But it is ugly and confusing to see these modals update with post-transaction information while they are closing. Even if they do not call `rerender()` themselves, the corporation-wide rerender interval can occur during the closing animation. I can think of a few ways to handle this:

1. Delaying effects until midway through the closing animation (currently implemented). This makes it harder to handle exceptions that should prevent the closing, and simply moving the error checking into a separate function is no guarantee that it will keep working as intended if something changes later.
2. Using `useMemo()` to only re-render dynamic values in response to input changes (or global rerender events before the transaction.) This is difficult because of linter rules requiring `useMemo()` to list all simple dependencies.
3. Replacing dynamic values with blank placeholders during certain conditions. It is difficult to express these conditions in terms of the UI props and state, and to make sure the placeholders do not cause resizing. But I think this might be the best approach. I can give it a try.

jjclark1982:
Ok, now all the Actions functions throw Errors, which are handled in their respective Netscript and UI callers.

I've found a way to make the stock transaction expected effect UIs look good without any weird closing process. By defining a component that only updates on changes to the `shares` input field, but including `corp` in its rendering function as a closure from the outside scope. This makes it practical to blank out the parts which should not be updated based on simple conditions like `shares == 0` or `!props.open`.

Snarling:
When removing the canXXX functions, you also duplicated the error checking in multiple locations, which is a bigger problem than a function having a weird return type.

5a59b28 is more-or-less a reimplementation of a canBuybackShares-like function just with a string return.

The same sort of thing should be done other places where this code duplication exists.

src/Corporation/ui/modals/BuybackSharesModal.tsx
```diff
@@ -23,70 +23,85 @@ export function BuybackSharesModal(props: IProps): React.ReactElement {
...
       return (
-        <>
-          Purchase {shares} shares for a total of {formatMoney(shares * buybackPrice)}
-        </>
+        <Typography>
+          <b>{corp.name}</b>'s stock price will rise to <Money money={sharePrice} /> per share.
+        </Typography>
       );
     }
   }
```
Snarling: A react component should never be defined inside of another react function component, they should only be defined at top level scope within a module.

But I'm also not sure why this needs to be a separate component. I believe it would function identically if you just rendered that content directly within BuyBackSharesModal.

Same comment applies for other modals where this same sort of thing was implemented.

myCatsName:
just another 👍 for @jjclark1982 

Some of us are limited to debating a single comma for eternity, meanwhile someone's making big changes on a system that according to some rando on reddit, "almost made me want to stop playing this game." 

One suggestion I could offer is re: this modal
image

It seems like it wants a "Decline" button, even with the updated text. (decline, reject, no deal, whatever). 
Also wonder if the button is positioned where you intended? From the changed code snippet I would've assumed button was below the typography, so am just confused.

jjclark1982:
Sorry, it is difficult to explain the closing-animation problem that I am trying to address. If you pause this recording during the change, you can see the ugly resizing and potential leak of future information.

https://github.com/bitburner-official/bitburner-src/assets/208776/bbef3f56-4f3e-4ca3-82cd-603150021854

Another possible solution is to simply blank out the entire modal when it is closed. This looks like the modal disappears instantly instead of using the standard transition, but it is not noticeable because a dialog is opening at the same time.

https://github.com/bitburner-official/bitburner-src/assets/208776/3e6be928-f35f-4b3d-a179-25a5547fe0b7

This is probably the best for code maintainability as it does not use any weird caching tricks.

jjclark1982:
> 
> One suggestion I could offer is re: this modal image
> 
> It seems like it wants a "Decline" button, even with the updated text. (decline, reject, no deal, whatever). Also wonder if the button is positioned where you intended? 

That's a good point. Some other stock actions are more questions of amount or timing, but this one is a yes/no question, so the game mechanic should be more learnable with a written "no" button.

Screenshot 2023-09-19 at 12 49 07 PM 1

myCatsName:
> If you pause this recording during the change, you can see the ugly resizing

image
slightly cursed lol. I think the second transition you demonstrated looks good.

Snarling:
> Sorry, it is difficult to explain the closing-animation problem that I am trying to address. If you pause this recording during the change, you can see the ugly resizing and potential leak of future information. [...]

This isn't a problem just with your modals and it shouldn't be addressed individually inside every affected modal, it should be addressed at the modal definition itself.

I made a PR #817 that should address this in a more fitting location, let me know if you see any issues with it.

Snarling:
Just did one more commit 20c3863 for the same error unification thing but with issuing shares.

I'm good with merging this, were there any other changes you were going to make here?

jjclark1982:
Looks good all around! Thanks for working through the weird modal behavior with me.

Snarling (approved):

Snarling:
Cool, I will see if there's anything else I can merge tonight and push these updates out to the web and steam deployments.

PR #783, opened 2023-09-06
UI: Add filter to faction augmentation purchase page

ficocelliguy:
Currently, the only way to identify which augmentations are helpful for your current goal (e.g. gain hacking level) is to slowly mouse over each one to see its stats and description. This can become VERY tedious, especially once you are dealing with a gang with dozens and dozens of augments on offer, and when you only need certain stats in the current bitnode.

This change adds a filter box to the augmentation purchase screen for factions, allowing users to narrow down the displayed augmentations to only the ones they are looking for. This feature works with the sort options; you can sort your filtered result by cost, rep, etc. The filter is cleared upon exiting the faction aug purchase screen, preventing it from getting "stuck" and hiding important augs from the player if they forgot about it or accidentally added some filter text.

Note that this allows the users to also filter for text that is only present in the description, in addition to text in the name and stats; in some cases this may clutter the results, but I wanted the feature to be as flexible as possible.

For example, this filters the very large list down to a manageable few that help with crime success rates:
Screenshot 2023-09-06 143403

Placeholder text gives the user a hint as to what it does when empty:
image

### DIFF
--- a/src/Faction/ui/AugmentationsPage.tsx
+++ b/src/Faction/ui/AugmentationsPage.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from "react";
+import React, { useState, useMemo } from "react";
 import { Box, Button, Tooltip, Typography, Paper, Container, TextField } from "@mui/material";
 import SearchIcon from "@mui/icons-material/Search";
 
@@ -23,19 +23,23 @@ export function AugmentationsPage({ faction }: { faction: Faction }): React.Reac
   const rerender = useRerender(400);
   const [filterText, setFilterText] = useState("");
 
-  function getAugs(): AugmentationName[] {
-    return getFactionAugmentationsFiltered(faction).filter(
-      (aug: AugmentationName) =>
-        !filterText ||
-        matches(Augmentations[aug].name, filterText) ||
-        matches(Augmentations[aug].info, filterText) ||
-        matches(Augmentations[aug].stats, filterText),
-    );
-  }
+  const matches = (s1: string, s2: string) => s1.toLowerCase().includes(s2.toLowerCase());
+  const factionAugs = useMemo(() => getFactionAugmentationsFiltered(faction), [faction]);
+  const filteredFactionAugs = useMemo(
+    () =>
+      factionAugs.filter(
+        (aug: AugmentationName) =>
+          !filterText ||
+          matches(Augmentations[aug].name, filterText) ||
+          matches(Augmentations[aug].info, filterText) ||
+          matches(Augmentations[aug].stats, filterText),
+      ),
+    [filterText, factionAugs],
+  );
 
-  const matches = (s1: string, s2: string) => {
-    return s1.toLowerCase().includes(s2.toLowerCase());
-  };
+  function getAugs(): AugmentationName[] {
+    return filteredFactionAugs;
+  }
 
   function getAugsSorted(): AugmentationName[] {
     switch (Settings.PurchaseAugmentationsOrder) {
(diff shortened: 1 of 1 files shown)


src/Faction/ui/AugmentationsPage.tsx
```diff
@@ -202,6 +218,19 @@ export function AugmentationsPage({ faction }: { faction: Faction }): React.Reac
...
+            onChange={handleFilterChange}
+            autoFocus
+            placeholder="Filter augmentations"
+            InputProps={{
+              startAdornment: <SearchIcon />,
+              spellCheck: false,
+            }}
+            style={{
+              paddingTop: "8px",
+            }}
```
Snarling:
```suggestion
            sx={{ pt: 1 }}
```
Since this is a MUI component, this is preferred to directly adding padding.

src/Faction/ui/AugmentationsPage.tsx
```diff
@@ -20,11 +21,22 @@ import { useRerender } from "../../ui/React/hooks";
...
+      (aug: AugmentationName) =>
+        !filterText ||
+        matches(Augmentations[aug].name, filterText) ||
+        matches(Augmentations[aug].info, filterText) ||
+        matches(Augmentations[aug].stats, filterText),
+    );
   }
 
+  const matches = (s1: string, s2: string) => {
+    return s1.toLowerCase().indexOf(s2.toLowerCase()) !== -1;
```
Snarling: `.indexOf !== -1` is more confusing than just using `.includes`

Snarling (commented):
A lot of the expensive operations here (getting the augs, sorting them, filtering them) are currently performed every render cycle instead of just when they need to be. There's room to refactor this with some useEffects to make the expensive operations only trigger when they need to. This PR does make it more important for us to think about optimization here, because typing into the filter field is now a way to trigger rapid rerenders of the component.

But that's not something that needs to be addressed in this PR, just an observation. Other than the two very minor comments below, this seems good and I'll probably merge this in about a day or two to give others the chance to give any feedback.

myCatsName:
Literally to stressed to read the details, but I should've known you'd make an upgrade for this! It's another practical idea, gg.

FWIW, I had been thinking about making a "Calculator" or more specifically something like a "Checkout" helper page, with the idea to package it as an "Augmentation Helper" Program. The idea being eg. list all augs available to you, grey out ones you can't purchase, allow adding and removing from cart with panels to show costs and the changes to your stats if you buy your cart. You can filter by traits, etc. Automatically sort for efficient purchase order, etc. Wishlist.

If that's something you can use, or want to collab on ok.

ficocelliguy:
> Literally to stressed to read the details, but I should've known you'd make an upgrade for this! It's another practical idea, gg.
> 
> adding and removing from cart with panels to show costs and the changes to your stats if you buy your cart. You can filter by traits, etc. Automatically sort for efficient purchase order, etc. Wishlist.

I love the idea of putting together a cart or favorite list for a future enhancement. It's so annoying to switch between factions and find that one expensive aug, then switch back...

Though as I am typing this, it seems like it might be better to give good hooks to the player so they can make their own 'optimal' aug purchase order script? Maybe that already exists in the singularity API?

myCatsName:
There's enough in .singularity to help a player find base prices and make and install purchases. Enough to make a script.

So the idea is to really focus mostly on a first time player. Encourage them to explore augments, since to me I don't like players feeling scared to install and not understanding that this is very normal, expected and necessary. And to show that it's not even possible to eg save up and buy 40 augs in 1.1 Checkout program gives you the information that oh prices get really high really fast, I need to reset that that multiplier to 0 by installing. I want to tell them this will happen a number of times it's safe to install and it is part of the process. 

So I imagine it is something you buy for a fairly cheap price on darkweb or low build time, and essentially expect 90% of the users are going to be in BN 1.1. It's not needed for more comfortable and familiar players, and totally useless for people who script augments.

d0sboots:
I like that idea of a calculator as something in Darkweb that you could buy, and much like the scan upgrades it's useless to experienced players but can provide QoL to players who are just getting their feet wet.

PR #784, opened 2023-09-06
CORPORATION: remove annoying String from TA's

Caldwell-74:
i was wrong that the strings where gone appearently i just didnt test correctly
but to fix that here the PR to just remove that code

removes the "default" string the TA's add
if that causes an issue we'll find a better way to deal with it

oh and i hope i can rely on ts here that the props will always be string| number if so then we dont need the ||'s or ??'s


Caldwell-74:
another/next PR should be removing the TA button/modal and integrating it into the sell modal
ill try myself at it cant guarantee success

Snarling:
> oh and i hope i can rely on ts here that the props will always be string| number if so then we dont need the ||'s or ??'s

I think the only edge case here where it might matter is for the number 0. String(0) converts to "0" instead of "", which may cause a difference in display, and if that resulting string conversion is (for some reason) used as a boolean, then "0" will be truthy where "" was falsy.

Practically here, I think it means if the price for a product is currently 0 (the default), the text box will start with the text "0" instead of being empty, which might be annoying.

Caldwell-74:
the question there would be how to differentiate between an initial value 0 and a player set 0
because the player set 0 should be shown as "0" in the field (would be wierd to set a price and not see it)


initial to null and then price ?? "" would be my way even if i dont like to introduce nulls

Caldwell-74:
should i change the api (getProduct and getMaterial) to also return null when its not set?

Snarling:
So changing it to null has possible other effects.

For instance, if the player types "", should that set it back to null? If "" is going to behave the same as null, why not just change the default to ""?

Caldwell-74:
i have done goofed will redo it

Caldwell-74:
ok no clue how that still has a conflict 
i reset -- hard the branch to upstream/dev
made the changes again and push origin --force it 
there shouldnt be any conflicts?

d0sboots:
Fun with git
Make that mistake 5 more times, and you'll figure out your favorite way of dealing with it XD

PR #789, opened 2023-09-07
CORPORATION: Rework valuation calculation

yichizhng:
### Overview

This change introduces accounting for the total assets of a corporation, and uses deltas in the total assets to compute valuation, instead of the revenue and expenses. The intention of this change is to remove the ability to inflate valuation by buying and selling a material without actually doing any production (aka "real estate fraud").

A demo for this is available at https://yichizhng.github.io/bitburner-src/

Note that this PR will have a merge conflict with https://github.com/bitburner-official/bitburner-src/pull/782, another deserved Corporation nerf.

### API Changes

Two new properties, `totalAssets` and `previousTotalAssets` are added to the Corporation object; they are only used internally. A new method `addNonIncomeFunds` is also added for cases where the company gains funds which should not be treated as income; these are investor rounds, stock issuance, and the Hashnet upgrade.

### UI Changes

The current value of `Corporation.totalAssets` is added to the Corporation overview, below current funds.

### Balance

The total assets of a company is calculated as the sum of the following:
- The corporation's funds
- The creation cost of each division
- The sum of the purchase cost of the materials and products in each warehouse (imports and productions are accounted for at market price)

The intention of this calculation is that actions other than production should not increase the valuation of a company. However, arbitrage still does (but only at sale time, so that valuation does not fluctuate while simply holding materials) - this would take more complicated accounting to address, and the effect is not significant.


jjclark1982:
Since you are checking actions for whether they are *income* or *non-income*, could you also check the 15-or-so places that do some variant of
```js
corporation.funds = corporation.funds - cost;
corp.funds = corp.funds - cost;
corp.funds -= cost;
corp.funds -= price;
```
and rewrite them to use the relevant `addFunds()` method if applicable?

src/Corporation/Corporation.ts
```diff
@@ -181,7 +181,7 @@ export class Corporation {
 
   determineCycleValuation(): number {
     let val,
-      assetDelta = this.totalAssets - this.previousTotalAssets;
+      assetDelta = (this.totalAssets - this.previousTotalAssets)/10;
```
jjclark1982: Where does this factor of 10 come from? I think it's based on `corpConstants.gameCyclesPerCorpStateCycle`, so better to use that directly.

jjclark1982: oh you already did 😆

Caldwell-74:
mughur would be proud of this anti fraud pr :3
but i like!

src/Corporation/ui/Overview.tsx
```diff
@@ -62,6 +62,7 @@ export function Overview({ rerender }: IProps): React.ReactElement {
       <StatsTable
         rows={[
           ["Total Funds:", <Money key="funds" money={corp.funds} />],
+          ["Total Assets:", <Money key="assets" money={corp.totalAssets} />],
           ["Total Revenue:", <MoneyRate key="revenue" money={corp.revenue} />],
           ["Total Expenses:", <MoneyRate key="expenses" money={corp.expenses} />],
           ["Total Profit:", <MoneyRate key="profit" money={corp.revenue - corp.expenses} />],
```
jjclark1982, line 68: should this display `assetDelta` instead of raw profit?

jjclark1982:
`Corporation.Actions.IssueNewShares()` should also use `addNonIncomeFunds()`

src/Corporation/Corporation.ts
```diff
@@ -196,9 +209,27 @@ export class Corporation {
     if (this.valuationsList.length > corpConstants.valuationLength) this.valuationsList.shift();
     let val = this.valuationsList.reduce((a, b) => a + b); //Calculate valuations sum
     val /= corpConstants.valuationLength; //Calculate the average
```
jjclark1982: this should be
```js
val /= this.valuationsList.length;
```
(only relevant within 10 cycles after creating a corp, but good to fix it now)

yichizhng: Done

Snarling:
Sorry, haven't gotten around to really reviewing this one yet. Just from the first cursory look, I don't see any major issues.

yichizhng:
There's one more outstanding "issue" which is that upgrades should really count as CapEx instead of OpEx

I'll get around to it later today probably

yichizhng:
that change is finally in... although addNonIncomeFunds is now a bad name, since it gets passed negative values

src/Corporation/Corporation.ts
```diff
@@ -170,35 +183,55 @@ export class Corporation {
 
   determineCycleValuation(): number {
     let val,
-      profit = this.revenue - this.expenses;
+      assetDelta = (this.totalAssets - this.previousTotalAssets) / corpConstants.secondsPerMarketCycle;
+    // Handle pre-totalAssets saves
+    assetDelta ??= this.revenue - this.expenses;
```
Snarling, line 188: Since totalAssets and previousTotalAssets are members of the Corporation class which have default values of 150e9, I believe a loaded save that is missing these values will load in with 150e9 for each of these and the nullish conditional value isn't doing anything.

Even if totalAssets and previousTotalAssets were something like undefined, assetDelta would not be undefined but would instead be NaN, and the nullish assignment would do nothing.

Snarling (commented):
Right now I believe any loaded corp save will be affected as though the assets of the loaded corp are only 150e9 (even if e.g. a lategame corp is added).

~~In SaveObject.ts we can add a basic valuation approximator for these old corp saves so they don't just get evaluated down to 150e9. I don't mind implementing that, let me know if you'd like assistance with it.~~ I had an overall misunderstanding about how we're tracking assets, disregard. Saves loaded with a corp will still use 150e9, but it will get overwritten with valid assets shortly thereafter.

Also while it works to do `addNonIncomeFunds(-amount)` for expenses, it would be easier to look at and understand code if a separate function was added for spending vs earning (similar to player.loseMoney and gainMoney)

Snarling:
Okay so I had a large misunderstanding on how this was working. The existence of addNonIncomeFunds made me think that the core of how we're tracking assets was by thoroughly tracking all transactions and accurately tracking every change in assets.

However, now I see that we're directly recalculating the assets, by iterating all the owned divisions/materials/products, every time we actually need to know the assets, not relying on our tracked-as-we-go assets info at all.

Given this approach, I'm not sure how useful the addNonIncomeFunds function is. Looks like all it really impacts is the corp overview page, where the displayed assets might get updated mid-cycle (e.g. purchasing an upgrade), and addNonIncomeFunds allows the display of an updated asset value without waiting for the next cycle.

I think in the future it will be possible to use a fuller track-as-we-go approach instead of recalculating assets every time, and there will be some issues to deal with before we can make that change, but with the current method I think this looks good and will go ahead and merge this. Thanks!

PR #790, opened 2023-09-08
UI: Rename gang territory task to avoid confusion with other similarly-named mechanics

ficocelliguy:
The game incentivizes players to make sure their gang members are always engaged (not idle). Currently, one of the things that individual members can be engaged in, is "Territory Warfare". Despite being so engaged, this state is very is different than the "engaged in territory warfare" references in the documents and territory tab. (This caused me quite a lot of Googling on my first gang to try and make sure the greenhorns weren't in danger until I was ready for it!)

To make it more clear that the gang task is distinct (but related) to the state of war the gang can be in, I am proposing to rename the task "Territory Patrol" instead, or some other distinct but clear name. (I welcome input on other wording options!). This gives some association with the territory tab (it is the only task that references "territory") but is also clearly distinct from the clash enabled state.

I have also added some more detail to the task description, to explicitly describe what the task does and how it interacts with warfare.

To make sure this change does not break existing player automation, the gang task API accepts the legacy task name when assigning gang members to this task, to make sure this change does not have detrimental side effects.

image

### DIFF
--- a/src/Documentation/doc/advanced/gang.md
+++ b/src/Documentation/doc/advanced/gang.md
@@ -6,7 +6,7 @@ Seen by most of the population as nihilistic, murderous and vile, occassional ru
 
 ## Starting and Recruiting
 
-Outside of [BitNode-2](bitnodes.md) gangs require much more crime and heartbreak to create, but can still be a great help. Creating a Gang in other [BitNodes](bitnodes.md) will offer more [Augmentations](../basic/augmentations.md) than other [Factions](factions.md), but they will not be a way to destroy the [BitNode](bitnodes.md) alone.
+Outside of [BitNode-2](bitnodes.md) gangs require much more crime and heartbreak to create, but can still be a great help. Creating a Gang in other [BitNodes](bitnodes.md) will offer more [Augmentations](../basic/augmentations.md) than other [Factions](../basic/factions.md), but they will not be a way to destroy the [BitNode](bitnodes.md) alone.
 
 After creating a gang, you will be able to start recruiting, adding members to your gang as you gain Respect. While in a BitNode, your gang and gang member stats will not reset if you install augmentations.
 
@@ -30,6 +30,6 @@ Active gang members earn stats, respect and money based on their current stats, 
 
 Your gang's "Wanted Level" can make tasks much less productive, and is affected by the tasks assigned to gang members. "Ethical Hacking" or "Vigilante Justice" tasks can lower Wanted Level.
 
-"Territory Warfare" is a special task that builds Power for your gang. If "Territory Battles" is enabled [see the Territory subpage of your Gang page], members have a chance to win or lose territory by clashing with other gangs. The % of Territory you control affects most aspects of your gang productivity.
+"Territory Warfare" is a special task that builds Power for your gang. If "Territory Clashes" is enabled [see the Territory subpage of your Gang page], members have a chance to win or lose territory by clashing with other gangs. The % of Territory you control affects most aspects of your gang productivity.
 
 Note that gang members can die during clashes, even if your gang wins.
--- a/src/Documentation/doc/advanced/gang.md
+++ b/src/Documentation/doc/advanced/gang.md
@@ -30,6 +30,6 @@ Active gang members earn stats, respect and money based on their current stats, 
 
 Your gang's "Wanted Level" can make tasks much less productive, and is affected by the tasks assigned to gang members. "Ethical Hacking" or "Vigilante Justice" tasks can lower Wanted Level.
 
-"Territory Warfare" is a special task that builds Power for your gang. If "Territory Warfare" is enabled [see the Territory subpage of your Gang page], members have a chance to win or lose territory by clashing with other gangs. The % of Territory you control affects most aspects of your gang productivity.
+"Territory Warfare" is a special task that builds Power for your gang. If "Territory Battles" is enabled [see the Territory subpage of your Gang page], members have a chance to win or lose territory by clashing with other gangs. The % of Territory you control affects most aspects of your gang productivity.
 
 Note that gang members can die during clashes, even if your gang wins.
--- a/src/Documentation/doc/advanced/gang.md
+++ b/src/Documentation/doc/advanced/gang.md
@@ -30,6 +30,6 @@ Active gang members earn stats, respect and money based on their current stats, 
 
 Your gang's "Wanted Level" can make tasks much less productive, and is affected by the tasks assigned to gang members. "Ethical Hacking" or "Vigilante Justice" tasks can lower Wanted Level.
 
-"Territory Warfare" is a special task that builds Power for your gang. If "Territory Clashes" is enabled [see the Territory subpage of your Gang page], members have a chance to win or lose territory by clashing with other gangs. The % of Territory you control affects most aspects of your gang productivity.
+"Territory Warfare" is a special task that builds Power for your gang. If "Territory Clashes" are enabled [see the Territory subpage of your Gang page], members have a chance to win or lose territory by clashing with other gangs. The % of Territory you control affects most aspects of your gang productivity.
 
 Note that gang members can die during clashes, even if your gang wins.
--- a/src/Gang/data/tasks.ts
+++ b/src/Gang/data/tasks.ts
@@ -379,7 +379,7 @@ export const gangMemberTasksMetadata: IGangMemberTaskMetadata[] = [
     params: { chaWeight: 100, difficulty: 8 },
   },
   {
-    desc: "Members assigned to this task increase your gang's power. They will also fight for territory if 'Territory Clashes' is engaged.<br><br>Note that gang members can be killed while assigned this task during clashes.",
+    desc: "Members assigned to this task increase your gang's power. They will also fight for territory if 'Territory Clashes' are enabled.<br /><br />Gang members performing this task can be killed during clashes.",
     isCombat: true,
     isHacking: true,
     name: "Territory Warfare",
--- a/src/Gang/data/tasks.ts
+++ b/src/Gang/data/tasks.ts
@@ -379,7 +379,7 @@ export const gangMemberTasksMetadata: IGangMemberTaskMetadata[] = [
     params: { chaWeight: 100, difficulty: 8 },
   },
   {
-    desc: "Members assigned to this task increase your gang's power and will fight for territory if 'Territory Warfare' is engaged. Note that gang members can be killed while assigned this task.",
+    desc: "Members assigned to this task increase your gang's power, and will fight for territory if 'Territory Battles' is engaged.<br><br>Note that gang members can be killed while assigned this task during clashes.",
     isCombat: true,
     isHacking: true,
     name: "Territory Warfare",
--- a/src/Gang/data/tasks.ts
+++ b/src/Gang/data/tasks.ts
@@ -379,7 +379,7 @@ export const gangMemberTasksMetadata: IGangMemberTaskMetadata[] = [
     params: { chaWeight: 100, difficulty: 8 },
   },
   {
-    desc: "Members assigned to this task increase your gang's power, and will fight for territory if 'Territory Battles' is engaged.<br><br>Note that gang members can be killed while assigned this task during clashes.",
+    desc: "Members assigned to this task increase your gang's power. They will also fight for territory if 'Territory Clashes' is engaged.<br><br>Note that gang members can be killed while assigned this task during clashes.",
     isCombat: true,
     isHacking: true,
     name: "Territory Warfare",
--- a/src/ScriptEditor/NetscriptDefinitions.d.ts
+++ b/src/ScriptEditor/NetscriptDefinitions.d.ts
@@ -783,7 +783,7 @@ interface GangGenInfo {
   wantedLevel: number;
   /** Wanted level gained/lost per game cycle (negative for losses) */
   wantedLevelGainRate: number;
-  /** Indicating if territory battles is enabled */
+  /** Indicating if territory clashes is enabled */
   territoryWarfareEngaged: boolean;
   /** Number indicating the current wanted penalty */
   wantedPenalty: number;
@@ -3626,13 +3626,13 @@ export interface Gang {
   getAscensionResult(memberName: string): GangMemberAscension | undefined;
 
   /**
-   * Enable/Disable territory battles.
+   * Enable/Disable territory clashes.
    * @remarks
    * RAM cost: 2 GB
    *
-   * Set whether or not the gang should engage in territory battles
+   * Set whether or not the gang should engage in territory clashes
    *
-   * @param engage - Whether or not to engage in territory battles.
+   * @param engage - Whether or not to engage in territory clashes.
    */
   setTerritoryWarfare(engage: boolean): void;
 
--- a/src/ScriptEditor/NetscriptDefinitions.d.ts
+++ b/src/ScriptEditor/NetscriptDefinitions.d.ts
@@ -783,7 +783,7 @@ interface GangGenInfo {
   wantedLevel: number;
   /** Wanted level gained/lost per game cycle (negative for losses) */
   wantedLevelGainRate: number;
-  /** Indicating if territory clashes is enabled */
+  /** Indicating if territory clashes are enabled */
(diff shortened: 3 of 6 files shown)


myCatsName:
I respect you and since you are walking through the same field of decisions and making different choices, I'll share.

Gang member card is already visually noisy, lots of quickly updating info 1 side, menu and text other side, small space.

I worked to minimize the text, and you have chosen to maximize it. The task description has gone from 30 words to 45 and visually doubled to fill the complete space. If this were adding all relevant detail, great. But in fact you have returned some of the exact wordiness that was cut and added more.
OLD
>"Members assigned to this task increase your gang's power and will fight for territory if 'Territory Warfare' is engaged. Note that gang members can be killed while assigned this task.",

New
>"Assign this gang member to maintain the gang's territory. Members assigned to this task will help increase your gang's power. If the gang is engaged in territory warfare, this member will clash with other gangs and attempt to increase the territory your gang controls.",

For example a gang member doesn't need to "_help_ increase your gang's power" they can just "increase" it. For example " If _the gang is_ engaged in territory warfare" vs "If territory warfare is engaged". Both of which are unable to explain the 20 second clash intervals or other specifics, thus making the extra words seem irrelevant, or confusing. Extra player work to keep details and build a mental model.

The player is no longer informed about death as a risk and effect.

The misleading, unnecessary concept of "maintain" territory is re-introduced. Maintain has a literal definition "keep (something) at the same level or rate," which does not apply. Territory % is "maintained" by not activating Territory Warfare. It's fought over with TW engaged.

Changing the name of the task. I would not choose "Patrol" which risks simply adding another new concept to the system. Prepare for War, Waiting to Clash. There are others. Patrol is fine too, name isn't enough to explain the mechanic whatever it may be.

But if you have your reason, don't let me stop you, I only share because of personal respect to you.

ficocelliguy:
Good callout on the description wordiness, it definitely can be improved from what I have. I like your suggestions. I was trying to incorporate some flavor but it's definitely harder to parse then it needs to be.

`Prepare for War` is certainly interesting. Does it still make sense when you are AT war? 

I'll spend some time working on the wording here. Thank you for the feedback!

myCatsName:
gang power, territory warfare, gang member death is a decent amount of flavor in a short space, but I understand.

>Prepare for War is certainly interesting. Does it still make sense when you are AT war?

Not so much. Probably tolerable but not accurate.

If you again trim or revert so there is space under a short paragraph, there should be enough space to add for example 
`Status: Preparing to Clash` and useState to display instead `Status: Clashing` for a short time, or a result.
with space for stats - display the clash or success chance
Those ideas might make Territory Warfare more tolerable as a task name, idk. The interface change, bringing up additional stat and status in that part of the card might be enough to signal to user that TW is something like a sub-system (deserving specific attention).

Snarling:
Not actually a huge fan of "Territory Patrol" as a name. When you do have clashes enabled and are actively trying to increase the size of your gang's territory, patrol kind of misses the mark on that since it makes it seem purely defensive.

Not that the existing name is great either, I'm just not feeling "Patrol", 

Here are some other thoughts:

* Territory Warfare the gang state could also have a slightly changed name - e.g. "Territory Clashes" to resolve the duplicate naming, and that might be an easier change since the player doesn't ever have to specifically name the gang state in their scripts like they do for the task (although it's probably baked into a function name).
* One of the major reasons the duplicate name doesn't work right now is that the TW gang state and member task are only loosely related. You do the member task to gain power, but you don't need to actually be doing clashes. In fact one meta strat is just to do the TW member task with 0% clash chance until you have enough power to handily beat all the other gangs, and then switch to safe tasks before enabling clashes, never putting your gang members at risk of being killed. For players who do this, there's not even gameplay overlap between the TW member task and gang state, you just need to do one before the other.

ficocelliguy:
Hmm, changing the name of the clash state to remove the collision WOULD be a lot easier, and avoid adding 'patrol' etc. to the job list. I'll work on that idea

ficocelliguy:
I have updated the switch labels and documentation to refer to the enabled clashing state as 'Territory Battles' in order to not collide with the 'Territory Warfare' gang member task name.

Note that this change was applied to both switches that engage in clashing: the one on the gang landing/member overview page, as well as the one on the gang territory page.

Note also that this does not change the API, which still uses `territoryWarfareEngaged` and `setTerritoryWarfareEngaged` . I may be able to alias the method to something that matches the new naming scheme, but aliasing the boolean seems more brittle.

Let me know if this phrasing works better, or if you would like any other changes about the way this has been handled.

src/Gang/data/tasks.ts
```diff
@@ -379,7 +379,7 @@ export const gangMemberTasksMetadata: IGangMemberTaskMetadata[] = [
     params: { chaWeight: 100, difficulty: 8 },
   },
   {
-    desc: "Members assigned to this task increase your gang's power and will fight for territory if 'Territory Warfare' is engaged. Note that gang members can be killed while assigned this task.",
+    desc: "Members assigned to this task increase your gang's power, and will fight for territory if 'Territory Battles' is engaged.<br><br>Note that gang members can be killed while assigned this task during clashes.",
```
myCatsName: image

myCatsName:
>refer to the enabled clashing state as 'Territory Battles'

What about being very literal? Specifically "Allow Clashes" ? With sort of the fiction that as gang leader you dictate if members are allowed to clash, and maybe word spreads slowly when you try to call it off. 

>Engaging in Territory Battles sets your clash chance to 100%. Disengaging will cause your clash chance...

=> 

>Allowing Clashes sets your clash chance to 100%. Disengaging will cause your clash chance...

.
>Then, enable "Engage in Territory Battles" to start fighting over territory.

=>

>Then, "Allow Clashes" to start fighting over territory.

.
>/** Indicating if territory battles is enabled */

=>

>/** Indicating if clashes are allowed */

ficocelliguy:
Hmm, that does get right to the heart of the theming. Would it cause confusion if clashes are still happening for a bit after you 'disallow' them?

src/Gang/data/tasks.ts
ficocelliguy: Read the description aloud, and only breathe once for each comma and twice for each period. It's uncomfortable to run straight through to the end with no air. You can feel where the pauses are, and (more importantly), it makes it easier to parse at a glance in this case as well.

Ideally this would be broken down into more sentences with simpler structure, but we have a small space to work with here, so making it easy to glance at is probably the best we can do

myCatsName:
For me no, like I said I can imagine you give the order to stop, but maybe other gangs retaliate, maybe your members are still heated up or active. And that detail could be re-enforced with documentation if necessary.

Would "Allow Clashes" avoid your new player confusion and googling?

src/Gang/data/tasks.ts
myCatsName: Shrug. It's a fine example of what was quoted (a coordinating conjunction). 2 list items, related and of equal importance joined with "and."
- Members assigned to this task increase your gang's power.
- Members will fight for territory if 'Territory Warfare' is engaged.

 "Commas where you breathe" is sort of a broad guideline, and not one that always fits proper style.

myCatsName: From the AP style guide on Coordinating Conjunctions
> ...no comma when the subject of the two clauses ["Members"] is the same and is not repeated in the second: We are visiting Washington and plan to see the White House.

From TextBroker
>If your subject is only stated once, then what you have is a compound predicate. It is incorrect to split a compound predicate.

d0sboots: It's not a simple case of coordinating conjunctions. The conditional phrase "if 'Territory Battles' is engaged" only applies to the 2nd clause. If there were a comma directly before the "if", it would be clear that the conditional applies to *both* clauses. With the comma as written, it's clear that it only applies to the 2nd clause. No comma would be ambiguous.

myCatsName: when the _subject_ of the 2 clauses is the same, which it is. The grammatical subject in this case is "Members", stated once and shared between the clauses.

But citing 3 references is enough (or AP alone, since that really is the authority), if someone wants to add a comma it's not worth trouble. Style is just style.

Edit : in the AP example "We are visiting Washington and plan to see the White House." Add a condition "We are visiting Washington and plan to see the White House if it's open." We would not add a comma before "and". In context the different verb forms "increase power" and "will fight if" functionally separate the clauses - only the latter uses a conditional form. 
Double edit: I enjoy language and grammar, but this is a big case of missing forest for tree.

ficocelliguy:
Its certainly better than what we have now, for sure. it avoids adding new concepts/terms

src/Gang/data/tasks.ts
ficocelliguy: I'm not proposing that this fits style guides; that isn't the main goal of documentation or user interface, especially when you have limited space to convey a lot of info. I'm pointing out that it makes the sentence easier to parse and understand the first time through, which is more valuable to the user as far as I can tell. 

"...increase your gang's power. They will also fight for territory..." is another option I considered. It is bit wordier and doesn't wrap as well, but it does make things very easy to scan.

myCatsName: I mean, I noticed because it _didn't_ make it easier to parse. Improper punctuation is like that, it stands out and breaks flow.

>especially when you have limited space to convey a lot of info

A constraint we both share...
thus a stylistically appropriate way to combine two clauses into one sentence.

Better to break it in 2 than add contradictory commas. If it's difficult without a comma, it's difficult enough with one to just simplify further.

myCatsName: PS - "Members assigned to this task increase your gang's power and will fight for territory if 'Territory Warfare' is engaged. Note that gang members can be killed while assigned this task."

Some online tools put this around 7th or 8th grade reading level. Sometimes material should be aimed lower than that, but in the context of a text based programming game, 8th is a pretty good target IMO.

d0sboots: As someone who also enjoys language, I can't quite drop this.

> in the AP example "We are visiting Washington and plan to see the White House." Add a condition "We are visiting Washington and plan to see the White House if it's open."

This is not an identical example, because it's not ambiguous. Why? Well, because we can see that the 2nd form doesn't work: "We are visiting Washington and plan to see the White House, if it's open." Attempting to rewrite the sentence this way makes the referent of "it" more ambiguous, and the overall implication (that you will cancel your entire trip to Washington if the White House is closed) is awkward enough that it can't really be written in this form.

A better example would be "We are visiting Washington and plan to see the White House if we are well." I think this shows the *same* ambiguity, and thus the same need for a comma, as the base sentence: Either after the "and", to make it clear that the White House visit will be skipped, or before the "if", to make it clear that the entire trip will be called off.

myCatsName: There are 2 realms - The important one is practical. We need to explain a detail to an audience. Any text that fails that or makes it difficult can be considered for change.

Online tools put the original sentence around grade level 7-9. That's a fine general target for a text-based programming game. But I fully respect there are times to simplify. Many general purpose texts should target about a 4th grade level; it's no insult to suggest 8th grade is sometimes too advanced. So in the practical realm, just break it up.

The entirely unimportant-by-comparison is the realm of grammar and theory. There is enough already said to point in a direction of self-ed.

myCatsName: Found an example of text graded closer to 4-7, just for comparison
image

src/Documentation/doc/advanced/gang.md
```diff
@@ -6,7 +6,7 @@ Seen by most of the population as nihilistic, murderous and vile, occassional ru
 
 ## Starting and Recruiting
 
-Outside of [BitNode-2](bitnodes.md) gangs require much more crime and heartbreak to create, but can still be a great help. Creating a Gang in other [BitNodes](bitnodes.md) will offer more [Augmentations](../basic/augmentations.md) than other [Factions](factions.md), but they will not be a way to destroy the [BitNode](bitnodes.md) alone.
+Outside of [BitNode-2](bitnodes.md) gangs require much more crime and heartbreak to create, but can still be a great help. Creating a Gang in other [BitNodes](bitnodes.md) will offer more [Augmentations](../basic/augmentations.md) than other [Factions](../basic/factions.md), but they will not be a way to destroy the [BitNode](bitnodes.md) alone.
```
myCatsName, line 9: thanks for catching that

myCatsName:
If everyone is happy with "Territory Clashes" ok, and I'm glad the fatal comma has been resolved.

Since we've already discussed "Allow Clashes" which avoids this issue, I'll note that the most recent update still keeps the IMO messy "engage" and "enable" language. In case of moving ahead with `Territory Clashes`, would it be good to pick one of those 2 as a standard? 
image

myCatsName:
@ficocelliguy are any of these accurate?

1 - you feel this is complete as is, and `Engage in Territory Clashes` is the right solution for now. Maybe "Allow Clashes" has a problem.
2 - you feel "Allow Clashes" is equally good or better, but would prefer that someone else make that change.
3 - you are still considering options, the process needs more time.
4 - you're waiting for Snarling to take over and make some decision.

ficocelliguy:
The changes in this PR with "engage in territory clashes" solve the initial problem I set out to resolve (clearly differentiate warfare task vs clash state).

@Snarling had asked that the name be "slightly changed" so I tried to keep it similar to the current phrasing re: "engage in".  

My interest is just to clear up the naming collision, though; "Allow" would work just as well to accomplish the initial goal, if the change is something the team wants/likes.

src/Gang/data/tasks.ts
```diff
@@ -379,7 +379,7 @@ export const gangMemberTasksMetadata: IGangMemberTaskMetadata[] = [
     params: { chaWeight: 100, difficulty: 8 },
   },
   {
-    desc: "Members assigned to this task increase your gang's power and will fight for territory if 'Territory Warfare' is engaged. Note that gang members can be killed while assigned this task.",
+    desc: "Members assigned to this task increase your gang's power. They will also fight for territory if 'Territory Clashes' is engaged.<br><br>Note that gang members can be killed while assigned this task during clashes.",
```
Snarling: Slight grammar / wording change + make br tags self closing to match jsx usage
```suggestion
    desc: "Members assigned to this task increase your gang's power. They will also fight for territory if 'Territory Clashes' are enabled.<br /><br />Gang members performing this task can be killed during clashes.",
```

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -783,7 +783,7 @@ interface GangGenInfo {
   wantedLevel: number;
   /** Wanted level gained/lost per game cycle (negative for losses) */
   wantedLevelGainRate: number;
-  /** Indicating if territory warfare is enabled */
+  /** Indicating if territory clashes is enabled */
```
Snarling:
```suggestion
  /** Indicating if territory clashes are enabled */
```

src/Documentation/doc/advanced/gang.md
```diff
@@ -30,6 +30,6 @@ Active gang members earn stats, respect and money based on their current stats,
 
 Your gang's "Wanted Level" can make tasks much less productive, and is affected by the tasks assigned to gang members. "Ethical Hacking" or "Vigilante Justice" tasks can lower Wanted Level.
 
-"Territory Warfare" is a special task that builds Power for your gang. If "Territory Warfare" is enabled [see the Territory subpage of your Gang page], members have a chance to win or lose territory by clashing with other gangs. The % of Territory you control affects most aspects of your gang productivity.
+"Territory Warfare" is a special task that builds Power for your gang. If "Territory Clashes" is enabled [see the Territory subpage of your Gang page], members have a chance to win or lose territory by clashing with other gangs. The % of Territory you control affects most aspects of your gang productivity.
```
Snarling:
```suggestion
"Territory Warfare" is a special task that builds Power for your gang. If "Territory Clashes" are enabled [see the Territory subpage of your Gang page], members have a chance to win or lose territory by clashing with other gangs. The % of Territory you control affects most aspects of your gang productivity.
```

Snarling (commented):
A couple areas where clashes being plural causes some grammar mishaps.

Also the existing statement "Note that gang members can be killed while assigned this task during clashes" was a little bulky and hard to understand so I changed it to "Gang members performing this task can be killed during clashes."

PR #796, opened 2023-09-10
CORPORATION: remove TA modals and integrate into sell modal

Caldwell-74:
as stated removes the TA modals and integrate the TA toggles into the sell modal

keeping them seperated just means alot more clicks to open and close the modals
combining them just makes sense to have one modal where the selling can be interacted with

for the text i sticked with the usefull pieces from the old, suggestions or later prs to change it up and improve it are appreciated


src/Corporation/ui/modals/SellMaterialModal.tsx
```diff
@@ -3,11 +3,11 @@ import { dialogBoxCreate } from "../../../ui/React/DialogBox";
 import { Material } from "../../Material";
 import { SellMaterial } from "../../Actions";
 import { Modal } from "../../../ui/React/Modal";
-import Typography from "@mui/material/Typography";
 import TextField from "@mui/material/TextField";
 import Button from "@mui/material/Button";
 import { KEY } from "../../../utils/helpers/keyCodes";
-
+import { Typography, FormControlLabel, Switch, Tooltip } from "@mui/material";
+import { Division } from "src/Corporation/Division";
```
Snarling, line 10: So this doesn't end up causing any issues because it's being imported just as a type, but we actually need to use relative paths for imports. If Division was being imported for runtime use then you would get a build error due to this being an absolute path.

Also when importing a class just to reference its type, it's best practice to do `import type` and put it above all the regular imports, this makes it clearer what the import dependencies are as far as figuring out circular dependencies. Types can have circular imports and they're fine with that, but runtime imports can sometimes cause issues when there is a circular dependency.

src/Corporation/ui/modals/SellProductModal.tsx
```diff
@@ -9,8 +9,10 @@ import TextField from "@mui/material/TextField";
 import Button from "@mui/material/Button";
 import FormControlLabel from "@mui/material/FormControlLabel";
 import Switch from "@mui/material/Switch";
+import Tooltip from "@mui/material/Tooltip";
 import { KEY } from "../../../utils/helpers/keyCodes";
 import { CityName } from "@enums";
+import { Division } from "src/Corporation/Division";
```
Snarling, line 15: Same comment about import statement

Snarling (commented):
Seems good. Just a minor comment about some import statements

src/Corporation/ui/modals/SellMaterialModal.tsx
Snarling: If you use VSCode you can have it automatically do relative paths for things it autosuggests, I didn't know about this for the longest time and was manually converting to relative paths:

File -> Preferences -> Settings -> User or Workspace tab (I have it set for User but Workspace is more correct) -> Filter for "import" -> Select Extensions/Typescript -> set "Import module specifier" for JS and TS to "relative"

image

Snarling:
See the comments for future reference, I am just merging this and I'll make some related import changes afterwards.

src/Corporation/ui/modals/SellMaterialModal.tsx
Caldwell-74: ill keep that in mind thanks!

PR #807, opened 2023-09-16
MISC: Added runDelay parameter to ns.spawn, allowing user defined delay

muesli4brekkies:
By suggestion on the discord.

Tested with 0, -n and 0.n and Infinity - It spawns the script immediately as expected. Passing a string or null/undefined also throws the correct error.

There may be some spooky game-breaking tricks that might be possible with this, so it may not be a desirable feature. :-)


myCatsName:
In general maybe this should be an optional parameter with a default value (10 seconds)?

muesli4brekkies:
> In general maybe this should be an optional parameter with a default value (10 seconds)?

Good idea. This and other methods default to one thread if left blank, for instance, so it'd make sense to follow that behaviour. 

Standby!

Snarling (commented):
This would be an API break, as anyone passing an arg to a script with ns.spawn would now be sending an invalid runDelay (or if their first arg is numeric, it would be consumed and used as a runDelay instead).

muesli4brekkies:
> This would be an API break, as anyone passing an arg to a script with ns.spawn would now be sending an invalid runDelay (or if their first arg is numeric, it would be consumed and used as a runDelay instead).

Fair. That's reason enough to close this PR in my opinion, if you agree? I don't think an API break is necessary for such a minor feature tweak.

myCatsName:
>That's reason enough to close this PR in my opinion, if you agree?

*Wave*
 *disagrees*
making spawn delay user-provided makes sense. And I think (unfortunately) adding as an optional parameter in existing ns.spawn() might be better than creating a separate method or signature.
image
image

If @Snarling has a different approach, OK, but the feature seems worth continued consideration. It might not be worth the change given probably a small % ever use ns.spawn(), but frankly the 10s delay might be a factor today.

Snarling:
There's a way we can get this done using an altered RunOptions

Yeah the 10s delay is probably a huge reason why people don't use spawn more often. I think it was originally to prevent ram cheesing by "passing the baton" between different scripts that each have a low ram cost but together have a larger cost.

But there are other ram minimisation strats that are just as effective, I don't think ns.spawn needs to remain chained.

Is there ever a use case where you want to set the delay to something other than "as quickly as the game will let me"? Maybe we only need to reduce the 10s time to something small and fixed like 200ms, instead of making it configurable. Zero delay is probably bad ~~because it could invite a new cross-script form of infinite loop that can hang the game~~. (Edit, actually it probably wouldn't hang due to the setTimeout, but I still think the delay should exist even if it's something small like 50ms)

myCatsName:
>using an altered RunOptions

RunOptions is used in run, spawn, and exec - might be worth considering user-defined delay for all since you mention it.

>Is there ever a use case where you want to set the delay to something other than "as quickly as the game will let me"?

My gut says maybe not many _efficient_ use cases? But also that ns.spawn() is a feature more likely to interest new players and option to self define would increase usability. A not efficient or necessary example is I did make a "slowAscend.js" basically just to allow me to start a script in terminal and navigate a few UI pages before it ran. It's dumb and obvs can be accomplished in other ways, but does suggest someone wanting to mess with a delay factor. Random delays, that's gotta be fun for someone.

runOptions currently looks like
image

d0sboots:
`RunOptions` is definitely the place to add this, if it needs to be configurable. It would be a little unfortunate that it would only affect `spawn`, but so be it. (I *don't* think we should extend it to `exec` and `run`, because the difference in semantics between synchronous vs. delayed is **very** large, and it would also make `spawn` completely obsolete.)

Snarling:
Yeah the way to delay an exec or run call would be to just sleep/asleep before it. Also it would be an api break to let those allow a delay "in-function", since they immediately return the pid of the executed script (which will not be available until after the execution is launched)

I was thinking to avoid an issue where one of the RunOptions doesn't have an effect on most of the functions that take that type, actually extending RunOptions to a new type, e.g. `type SpawnOptions = RunOptions & {spawnDelay?: number}` or something like that. But then the autogenerated documentation for it might end up looking wonky.

myCatsName:
If runDelay is only a ns.spawn() feature, it might make sense to keep it in the main parameters, for visibility and adoption. Kind of janky either way.

>I don't think we should extend it to exec and run

probably wise, it would be a significant change and probably broken. But you said it would be unfortunate if limited to `spawn`, was there more to that?

myCatsName:
This issue seems to be stalling.

@muesli4brekkies Are you still interested in this PR? 
If not, I will try to pick it up, make the runDelay an optional parameter with 10000ms default, polish and push so it's available in that state.

muesli4brekkies:
> This issue seems to be stalling.
> 
> @muesli4brekkies Are you still interested in this PR? If not, I will try to pick it up, make the runDelay an optional parameter with 10000ms default, polish and push so it's available in that state.

I am interested but I've been dead busy of late, mea culpa. Feel free to pick this up if I don't get to it before Monday say? Or sooner if you have more time that me right now. I'm not protective over the PR :)

That said though, did we reach a conclusion on where to put it? I'm happy for it to be a optional param, but as mentioned that would be an API break, so should it now go into runOptions?

myCatsName:
Please don't make 1 design choice with the expectation it will be done a different way in the future, it makes the adoption and learning process just difficult. If spawn() is the only instance of a special parameter, that param should be visible.

Undocumented guidance about API breaks might be a source of stalling we need to address. 

> Feel free to pick this up if I don't get to it before Monday say?

I'm in no rush and don't mean to put that on you either. I just want to understand the status and prevent stalling and accumulation in general. If it's still in process, that's fine.

muesli4brekkies:
> Please don't make 1 design choice with the expectation it will be done a different way in the future, it makes the adoption and learning process just difficult. If spawn() is the only instance of a special parameter, that param should be visible.
> 
> Undocumented guidance about API breaks might be a source of stalling we need to address.

Sorry, I'm slightly confused by this comment. I didn't mean to offend, or make any design choices, and if I did, they weren't intended. I only mean to follow your leads because I'm not nearly as involved, invested or knowledgable about the game, its goals and features as the rest of the dev team. 

Hence my hesitation and question about where the param should go :)

> > Feel free to pick this up if I don't get to it before Monday say?
> 
> I'm in no rush and don't mean to put that on you either. I just want to understand the status and prevent stalling and accumulation in general. If it's still in process, that's fine.

I'll see if I can get around to it this evening :)

d0sboots:
Unfortunately, making the parameter optional doesn't change the fact that adding a new parameter in that position is an API break. Anyone who was calling `spawn` with arguments will have the first argument treated as a `runDelay`, whether or not that argument was numeric.

This is why the options parameter `RunOptions` exists, so that we can safely expand the arguments to these functions.

Edit: If we want to create a new type `SpawnOptions` for greater visibility, that works too, although it would need to inherit from `RunOptions` and thus the docs would become less clear IMO.

myCatsName:
>API break. 

=>
>Undocumented guidance about API breaks might be a source of stalling we need to address.

(not @ anyone, just rhetorical)  How many people run ns.spawn(), and how long should those players expect a stable function? Are breaking changes to ns.hack() pushed on the same timeline as ns.spawn()? 

These are basic guideline questions that can be documented and applied. Setting expectations for users and devs.

I assume most ns.spawn() users will prefer and adopt an optional runDelay, wherever it fits, and we generally seem to agree. So feature delay balances against that.

myCatsName:
> I'll see if I can get around to it this evening :)

gg muesli. ~~I'll try to pick it up tomorrow and create a version with something like the `spawnOptions` suggestion from dos, for comparison and options.~~

muesli4brekkies:
I've had another bash and added a spawnOptions (et al) runOpts facsimile. All seems to be working correctly.

I tried to fork this with the idea of making a separate PR referenced here, so the best option can be selected, but alas command of git eludes me and I pushed to my main fork.

ps - I'm sure there's a better way to do this than the copy-paste grep-variable job I've done here. :thinking:

myCatsName:
> ps - I'm sure there's a better way to do this than the copy-paste grep-variable job I've done here. 🤔

More than good enough to demonstrate `SpawnOptions` can have good visibility. 🥇

d0sboots:
> Undocumented guidance about API breaks might be a source of stalling we need to address.

There are *lots* of things that are undocumented; right now core parts of the game are still undocumented. All of that is gated on developer time and (especially) interest. If you want to add something to `CONTRIBUTING.md` that discusses what API breaks are and the trade-offs around them, have at it.

src/NetscriptFunctions.ts
```diff
@@ -744,21 +744,20 @@ export const ns: InternalAPI<NSFull> = {
     },
   spawn:
     (ctx) =>
-    (_scriptname, _thread_or_opt = 1, ..._args) => {
+    (_scriptname, _thread_or_opt = 1, _spawndelay, ..._args) => {
```
d0sboots, line 747: This still has the wrong signature.

d0sboots: You forgot to revert this part. I'm surprised TypeScript didn't yell at you... I guess because of the varargs `...args` that comes after?

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -281,6 +281,37 @@ interface RunOptions {
...
+   */
+  ramOverride?: number;
+  /**
+   * Should we fail to run if another instance is running with the exact same arguments?
+   * This used to be the default behavior, now defaults to false.
+   */
+  preventDuplicates?: boolean;
+  /**
+   * Number of milliseconds to delay before spawning script, defaults to 10000 */
+  spawnDelay?: number;
```
d0sboots, line 287: Optional: Much like `additionalMsec` in HGWOptions, I would call this `spawnDelayMs` to make it clear in the name that it's milliseconds. Time units that aren't strongly typed are *very* easy to get mixed up.

Edit: Oops, should be spelled "Msec" for consistency.

src/Netscript/NetscriptHelpers.tsx
```diff
@@ -207,6 +216,45 @@ function runOptions(ctx: NetscriptContext, threadOrOption: unknown): CompleteRun
   return result;
 }
 
+function spawnOptions(ctx: NetscriptContext, threadOrOption: unknown): SpawnRunOptions {
```
d0sboots: So, duplicating the options in `SpawnOptions` has some documentation benefits, but it also is duplication. I'm a bit leery of duplication, but I can see the argument there.

The amount of duplication here is going too far IMO. Probably you should be calling `threadOptions`, then doing additional casting/validation to get a `SpawnOptions`. Casting from a `ThreadOptions` to a `SpawnOptions` is gross though, and only needed because `SpawnOptions` isn't inheriting from `ThreadOptions` like it logically actually does.

Lets see what Snarling thinks.

d0sboots (commented):
Something you missed, and a larger discussion point.

myCatsName:
>There are lots of things that are undocumented; right now core parts of the game are still undocumented. All of that is gated on developer time and (especially) interest. If you want to add something to CONTRIBUTING.md that discusses what API breaks are and the trade-offs around them, have at it.

You _are_ trying to be dismissive, right? I just can't imagine saying something so obvious and unhelpful without that intent.

src/NetscriptFunctions.ts
muesli4brekkies: I'm surprised as well lol. I've sorted this. :+1:

src/ScriptEditor/NetscriptDefinitions.d.ts
muesli4brekkies: Very good :)

myCatsName:
>There are lots of things that are undocumented; right now core parts of the game are still undocumented. All of that is gated on developer time and (especially) interest. If you want to add something to CONTRIBUTING.md that discusses what API breaks are and the trade-offs around them, have at it.

David, if you're not interested in an idea, or don't feel like participating, ok. This reply seems so obvious and unhelpful I have to assume you just want to send a dismissive message. It doesn't feel good. IDK why you put yourself in a position where you think you're giving permission and deciding what is worth being interested in as a developer, or who qualifies as such.

Snarling:
@muesli4brekkies see cc469d5, I refactored the spawnOptions helper so that it calls the runOptions helper instead of duplicating all the code. I think this is what @d0sboots was getting at in the previous review.

Personally I prefer the property name spawnDelay instead of spawnDelayMsec, but I didn't want to just change that unilaterally. What do others prefer? @myCatsName @d0sboots

myCatsName:
> I prefer the property name spawnDelay instead of spawnDelayMsec,

Yeah. It's defined well enough in the definition. Abbreviating milliseconds in a variable name is it's own debate, which can just be avoided if the usage here is as simple as it should be - figuring out the unit probably won't be the most difficult part of using this method.

Good feature, muesli, glad you saw it through.

d0sboots:
I gave my reasoning earlier. I've found that it's particularly useful to have the unit in the name of time APIs, extra much so when there are APIs dealing with (for instance) both seconds *and* milliseconds, like we have in BB. This also mirrors the guidance of others in some style guides (but that's not universal, of course.) If you don't find it adds enough value, it's your call to make.

Snarling:
Oh I must have missed that previous comment, I didn't know the addition of Msec was based on review.

I can see either way. It's probably true that the current API has some locations where a time is actually seconds and there's room for confusion, but AFAIK very few existing API variable names include the unit, so while I can see that being a standard for a freshly made API ("always include units for numeric values" would be a good rule, e.g. ramOverrideGB), I think it would be a departure from the norm for our existing API. I think consistency and simplicity here win out over the improved clarity from adding the unit to the property name.

The documentation stating that it's a delay in ms should be good enough for the player to avoid misuse.

muesli4brekkies:
Thanks for the polite and encouraging guidance everyone. :)

PR #808, opened 2023-09-16
BITNODE: BN3 starts with 1tb home

Caldwell-74:
the other PR's will fix the current "exploit" to get quick access to enough Ram to run corp scripts
i think its time to just give the player the ability to run those without exploits

### DIFF
--- a/src/BitNode/BitNode.tsx
+++ b/src/BitNode/BitNode.tsx
@@ -108,8 +108,9 @@ export function initBitNodes() {
         own community, and evict their neighbors from houses in their desperate bid to increase their wealth.
         <br />
         <br />
-        In this BitNode you can create and manage your own corporation. Running a successful corporation has the
-        potential of generating massive profits.
+        In this BitNode you can create and manage your own corporation, you have full access to the Corporation API and
+        Home starts with enough ram to use it. Running a successful corporation has the potential of generating massive
+        profits.
         <br />
         <br />
         Destroying this BitNode will give you Source-File 3, or if you already have this Source-File it will upgrade its
--- a/src/Prestige.ts
+++ b/src/Prestige.ts
@@ -226,6 +226,8 @@ export function prestigeSourceFile(isFlume: boolean): void {
       "You received a copy of the Corporation Management Handbook on your home computer. " +
         "Read it if you need help getting started with Corporations!",
     );
+    //start with enough ram to use corporation functions
+    homeComp.setMaxRam(1024);
   }
 
   // BitNode 6: Bladeburners and BitNode 7: Bladeburners 2079


myCatsName:
Costs $321m to buy a 1T p server in BN3. Don't need exploits for that. Something I missed?

Isn't it just part of the "power ladder" or the "bildungsroman" story arc in BN3 that you must find a way to get that money before automating with API?

And as a logical teaching device? Interact with the UI until you can afford API access, by which time hopefully you understand corporations a bit and have ideas to automate.

Caldwell-74:
even less when you reroll for a high fulcrum server (can be 1tb or 2tb) and buy port openers
but hacking is really nerfed in the node and with crimes alone it takes ages
cct's can be an option but shouldnt be a requirement

we are bouncing back and forth between leting exploits exist that make it possible to script right away and fixing them 
yeah corps are op but they wont be less op just cuz you have to grind 3x the early to get enough ram to run a script
and early scripting lets people learn corps step by step, automating one part (like party/tea, or product cycles) while learning other more complex stuff manually

and this is just another of those "talked about 1000x times but no one ever made a pr" so i just made it if snarling doesnt like it iam okay with that but atleast i tried :3

myCatsName:
Just for the data point - if a BN3 player does only crime for money and buys a $321m p server, they can have API access in less than 24 hours. 

If I think from a story perspective, it would be cool if creating a corporation gave you access to a corporate computer. At home, you're small, you had to get a loan from the city, so it makes sense you have small RAM. Maybe creating the corp adds a 1T Corporate server at the same time.

Caldwell-74:
maybe starting a corp upgrades home to 1tb if its below it?

edit.: but that would prevent getting bn1+ achievement when corps are used

d0sboots:
I have a different suggestion. Let's look at the base of the issue: we don't want people in BN3 to be unable to use the API, because we think it makes for a much better experience if you can start scripting when/where you want to. OK, I agree. What's stopping that? The fact that you need 1TB RAM to use the corp API. Well, why is it so expensive?

Seriously, what is the justification? I can't think of any, and it seems to be only hurting gameplay. You could make it lower only in BN3, but that isn't really needed - the cost to start a Corp is much greater than the cost to get 1TB, so it is *only* a roadblock for BN3 in the first place.

I think we should just lower the corp API to the same cost as regular functions, and call it good.

myCatsName:
[Outdated but relevant](https://steamcommunity.com/app/1812820/discussions/0/5219148331315434851/)

> what is the justification (for 1TB API)? I can't think of any

- Robbing a few k from n00dles vs. transacting corporate bribes, expanding divisions and moving quintillions of dollars or bulk materials. It's easy to believe corporate API requirement is on a different level. 
- "Power ladder" : early player feels powerless => grows to realize money is cheap and corps are strong. "How can I do that" => "I did that."

Not to argue, just one answer. 

>we think it makes for a much better experience if you can start scripting when/where you want to

BN6 (no API) vs. BN7 (API). SF-4 16x => 1x ns.singularity multiplier. These suggest BB is not a game that always let's you do what you want when/where you want, but at the same time puts doing exactly that within your reach - through exploit or progress.

I'm not sure what Caldwell is feeling about this, but if the effect on balance and time is OK and it improves the issue, I think corporate server is a solution worth working on. A way to effectively "start BN3 with 1tb" with in-game context.
 
 - server access can have requirements (not only hack level - for example, having negative money the corp might lock you out)
 - can host corp documentation or .lit "flavor" texts

Snarling:
I imagine after the upcoming PRs that address the incorrect share price and valuation calculations, corp should be in a much healthier spot compared to other mechanics and shouldn't need the harsh 1TB cost to allow API interaction with it.

I think if the premise is that corp will no longer be strong enough to need the 1TB ramcost as a gate for automation, the solution isn't really to give the player a free 1TB server in BN3, it's to reduce the ramcost of the API to something more in line with the power level.

Reducing the API cost to something like 62.4GB per function would allow a savvy player to run their corporation using a 64GB server, or for less savvy players they would still pay a much larger ram cost since the cost is per function.

Right now corp API is in a weird spot, where it's expensive but the player also is not encouraged to make the script ram-efficient because there is a hard cap of 1TB thread cost for the entire script.

myCatsName:
Ask the internet "are idle games just about big numbers?" and no, they aren't but also yes, idle games are big number games. 
image

It's not like they _have to_, but if the balance, timing and availability of a 1TB API and a 64GB API is not very different, see above.

>corp should be in a much healthier spot compared to other mechanics and shouldn't need the harsh 1TB cost

Makes sense, "corps are strong and needed a barrier" is another answer to the question of why API cost 1TB. But so is offering an unexpected experience to the player, in the node where they are likely to experience greater scale and complexity than before. 

"64gb" also implies a BN3 player can't access API immediately, just with a lower requirement than 1TB. And if we agree "by the time you can get a corp, you can easily get 1TB"... - it seems like OK, the API doesn't need to cost 1TB and could be lowered; but also can cost 1TB and issues can be solved in a way that adds to the mechanic, and in itself can be considered flavor. 

"Corporate stuff is complicated, so you now have access to a new server and all the tools you need. Good luck... - IT."

d0sboots:
I don't think 62.4GB is a good cost. There are no starting servers with that much RAM, so it still requires hacking to either unlock additional servers and/or get money to buy pservers. And then it *requires* RAM-dodging to use the RAM you have, which many people view as cheating. Fundamentally, it preserves the same issue we have currently: In BN3, the node where you should be learning about the mechanics, you instead have to jump through extra hoops to use the API. But in other BNs, you trivially will have enough money to buy a 1TB server in order to afford starting a corp, so using the API is trivial. This is backwards from how things should be.

There are 60 methods across the full corp API. Of those, roughly ~20 are critical to running a corp. The RAM you have access to on start is 32 (home) + 4 (n00dles) + 16 (foodnstuff) + 16 (sigma-cosmetics) + 16 (joesguns) + 16 (nectar-net) + 16 (hong-fang-tea) + 16 (harakiri-sushi) = 4GB * 33. I feel like 3.5GB would be a decent cost as a result; it's enough to run all the functions you need if you spread them across all the servers you have initial access to, or you can do RAM-dodging, or create special-purpose corp subscripts, or other options. It's big enough to still enforce being careful with RAM (you can still only fit 8 on home) without prescribing a single solution.

Snarling:
Going with the #818 instead of this. The player will still need to farm for money first if they want to run a single central corp script, but there's no reason that needs to be provided for free right from the getgo. The player can manually interact with the mechanic or use smaller single purpose automation scripts until they have enough ram to handle the API costs.

PR #809, opened 2023-09-16
UI: Issue 53 augs suppressed affects soft reset menu button

### LINKED ISSUE #53: Option "Suppress augmentation confirmation" out of sync with the button "Soft Reset"
[OriginalRobotWizard]
Steam beta development version a0361c9a has the following issues with the option "Suppress augmentation confirmation" and the button "Soft Reset":
- Changing the option "Suppress augmentation confirmation" affects the confirmation dialog for the button "Soft Reset"
- But it only applies after reloading the game or soft resetting (At least for the button)

So the following things should be changed:
- Confirmation dialog for "Soft Reset" should always appear, no matter if the option is enabled or disabled
- Changing the option should apply without having to reload or soft reset

As for the immediate effect of changing the option, I were only able to test it with the button, but not with buying augmentations or installing them, as of course, after the accidental soft reset, well, everything was reset.

[Evidence.zip](https://github.com/danielyxie/bitburner/files/8401113/Evidence.zip)

myCatsName:
closes #53

# Bug fix

"When Suppress augmentation confirmation is enabled, Soft Reset won't give a confirmation [modal]" => removed that feature. Soft Reset relates to augmentations abstractly, but a confirmation modal would be appropriate and shouldn't tie to suppressing augmentation messages.

The previous tooltip was completely inaccurate, given that Soft Reset _does_ install augmentations if any are purchased.
OLD:
SR toolt

 Augmentation install is similar and fine as a "short hand," but now more accurate
NEW:
SR tool NEW

Modal has room to be verbose, so the Soft Reset process is now distinctly de-coupled from augmentation language

OLD:
sr modal

NEW:
SR NEW

# Additional changes

- Travel agency, Delete Game and Restart Tutorial modals add a cancel button
- Arrival message can be suppressed 
- Removed Augmentations page arbitrary advice to preferentially buy "everything from any faction of your choosing."
- Tutorial modal text edit

### DIFF
--- a/src/Augmentation/ui/AugmentationsRoot.tsx
+++ b/src/Augmentation/ui/AugmentationsRoot.tsx
@@ -163,8 +163,7 @@ export function AugmentationsRoot(props: IProps): React.ReactElement {
                 <br />- home ram and cores
                 <br />
                 <br />
-                It is recommended to install several Augmentations at once. Preferably everything from any faction of
-                your choosing.
+                It is recommended to install several Augmentations at once.
               </>
             }
           />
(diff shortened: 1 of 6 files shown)


Snarling:
Very nice!

PR #810, opened 2023-09-17
UI: Create "Credits" button and modal for options menu

myCatsName:
# NEW

credits fingerprint
=>
Credits modal

image

- (Also adds Tooltip for Bug Report)

### DIFF
--- a/src/GameOptions/ui/CreditsModal.tsx
+++ b/src/GameOptions/ui/CreditsModal.tsx
@@ -20,34 +20,26 @@ Olivier Gagnon
 or to participate yourself`;
 
   return (
-    <Modal open={props.open} onClose={props.onClose}>
-      <Typography align="center" alignContent={"start"} variant="h3">
+    <Modal open={props.open} onClose={props.onClose} sx={{ textAlign: "center" }}>
+      <Typography alignContent={"start"} variant="h3">
         Bitburner
       </Typography>
-      <Typography align="center" sx={{ textDecoration: "underline" }}>
-        Original Code and Concept
-      </Typography>
-      <Typography align="center">danielyxie</Typography>
+      <Typography sx={{ textDecoration: "underline" }}>Original Code and Concept</Typography>
+      <Typography>danielyxie</Typography>
       <br />
-      <Typography align="center" sx={{ textDecoration: "underline" }}>
-        Lead Developers:
-      </Typography>
-      <Typography align="center" style={{ whiteSpace: "pre-wrap" }}>
-        {leadDevs}
-      </Typography>
+      <Typography sx={{ textDecoration: "underline" }}>Lead Developers:</Typography>
+      <Typography style={{ whiteSpace: "pre-wrap" }}>{leadDevs}</Typography>
       <br />
-      <Typography align="center" sx={{ textDecoration: "underline" }}>
-        Current Maintainer
-      </Typography>
-      <Typography align="center">{currentMaintainer}</Typography>
+      <Typography sx={{ textDecoration: "underline" }}>Current Maintainer</Typography>
+      <Typography>{currentMaintainer}</Typography>
       <br />
-      <Typography align="center" style={{ whiteSpace: "pre-wrap" }}>
+      <Typography style={{ whiteSpace: "pre-wrap" }}>
         <Link href={contributorsURL} target="_blank">
           {contributorsMessage}
         </Link>
       </Typography>
       <br />
-      <Typography align="center" fontSize={"large"}>
+      <Typography fontSize={"large"}>
         <Button onClick={props.onClose} size="large">
           Thanks for Playing!
         </Button>
--- a/src/GameOptions/ui/CreditsModal.tsx
+++ b/src/GameOptions/ui/CreditsModal.tsx
@@ -1,0 +1,57 @@
+import React from "react";
+import { Modal } from "../../ui/React/Modal";
+import { Typography, Link, Button } from "@mui/material";
+
+interface IProps {
+  open: boolean;
+  onClose: () => void;
+}
+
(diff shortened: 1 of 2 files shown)


src/GameOptions/ui/CreditsModal.tsx
```diff
@@ -0,0 +1,57 @@
...
+`;
+
+  const currentMaintainer = `@Snarling`;
+
+  const contributorsURL = `https://github.com/bitburner-official/bitburner-src/graphs/contributors`;
+  const contributorsMessage = `Visit GitHub to see all contributors
+or to participate yourself`;
+
+  return (
+    <Modal open={props.open} onClose={props.onClose}>
```
myCatsName: If possible, probably better to avoid duplicating `align="center"` in the following lines - I wasn't able to do that with sx , style, etc. in line 23.

Snarling:
@hydroflame Would you prefer to appear on the credits page as your username?

Zelow79:
Is there any way to dynamically include all contributors? I think it could serve as a little more incentive to get people to want to contribute. Maybe another button to pop out a list of contributors? I love this though, great idea.

myCatsName:
@Zelow79 I thought about the same idea, I guess - My plan was "Recent Patch contributors" listing a set of all the @ names in the most recent patch, pulling info from changelog. (Leaving the full list in the github link as shown)

I didn't follow up on that, but can try at least for testing or comparison.

Zelow79:
> @Zelow79 I thought about the same idea, I guess - My plan was "Recent Patch contributors" listing a set of all the @ names in the most recent patch, pulling info from changelog. (Leaving the full list in the github link as shown)
> 
> I didn't follow up on that, but can try at least for testing or comparison.

That would be pretty cool too. Might not even be too hard to make a contributors.ts that can be updated manually as new contributors come around. If it's too tedious for anyone else I would be willing to do the copy paste work on making one. Could probably have it return as an array so your ui change only needs to iterate it. Just an idea.

Zelow79:
Might not be a bad idea to make an official discord channel for new contributors to request the role in, this would double as a clear way for everyone to see when a new contributor needs to be added to the list and make a proper place for people to request the role.

Hoekstraa:
> Might not even be too hard to make a contributors.ts that can be updated manually as new contributors come around.

My spidey senses start tingling when you introduce manual actions, especially ones for which you should already have a listing (through Git and/or Github). It'd make more sense to automate this if you wanted to do this.

Secondly, having a bunch of Discord channels for discrete purposes is something you'd want to limit.
Don't want to overload the people with a bunch of slow moving channels. You can always ask for the role in #development, right where all things related to dev work go. I generally try to be ahead of the curve a bit by keeping track of new contributors, so some never even get to ask before they have the contributor role in Discord. Would I have to be less pro-active to ensure people take a manual action? What if that then means some never request or receive it?

myCatsName:
>Don't want to overload the people with a bunch of slow moving channels.

Adding a single inactive channel for a specific admin purpose isn't exactly that, though. Sticky a message at the top that explains how to contribute and the discord flair details, and delete incoming comments as they are verified and flaired.

Was not my suggestion, and I haven't seen a "need" given the slow rate of new contributors, and I respect and defer to the person who does the task. If I was the one, I think I'd add a channel and try it out, but it could be dead space. If it brings more minds to contribute, cool. Discord channels come and go.

>It'd make more sense to automate this if you wanted to do this.

I'm a ((sim|com)plex?) person, I see opportunity for regex, I regex.

Zelow79:
> > Might not even be too hard to make a contributors.ts that can be updated manually as new contributors come around.
> 
> My spidey senses start tingling when you introduce manual actions, especially ones for which you should already have a listing (through Git and/or Github). It'd make more sense to automate this if you wanted to do this.

I wasn't trying to imply that was the best solution, rather just one we could do if automation didn't seem possible or too complex for now. Obviously a dynamic approach is preferable.

> Is there any way to dynamically include all contributors? I think it could serve as a little more incentive to get people to want to contribute. Maybe another button to pop out a list of contributors? I love this though, great idea.

I did originally suggest a dynamic solution.

myCatsName:
@Zelow79 

something like this as a starting point
Edit: Data from CONSTANTS.LatestUpdate, adds about 25 lines and some regex. A bit fragile because it depends on some formatting staying consistent when updating LatestUpdate. The exact code I'm trying may not be viable outside .dev

image

src/GameOptions/ui/CreditsModal.tsx
```diff
@@ -0,0 +1,49 @@
+import React from "react";
+import { Modal } from "../../ui/React/Modal";
+import { Typography, Link, Button } from "@mui/material";
+
+interface IProps {
```
Snarling, line 5: There are still several similarly-named interfaces in older code, but naming an interface with an I at the start and naming it something nonspecific like just "Props" are both bad style in my opinion.

I would name this interface "CreditsModalProps"

Snarling (commented):
I think programmatically searching the latest update text for contributors names is a good idea. Contributor names should be the only thing parenthesized and with an `@`.

I am seeing one edge case in the current patch notes, where multiple contributors are credited for the same changelog entry (within the same parentheses).

If you still plan on implementing that, I will hold off on merging this until then.

src/GameOptions/ui/CreditsModal.tsx
myCatsName: copypasted, will update

myCatsName:
@Snarling 
> I think programmatically searching the latest update text for contributors names is a good idea

pushed as #836 . They've been inactive a week, but these should both be stable and from the same base; you can merge and close or test whatever is best for you.

Snarling:
Merged #836

PR #814, opened 2023-09-18
MISC: Add configuration to eslintrc to fix the warning thrown when linting

ficocelliguy:
This fixes the `Warning: React version not specified` warning thrown every time lint is run.


myCatsName:
I'm going to miss that warning :')

PR #815, opened 2023-09-18
API: Add ns.bladeburner.getNextBlackOp()

### LINKED ISSUE #66: API Request: Bladeburner.getNextBlackOpName
[smmalis37]
Currently the only way to get information about the available Black Ops is to start by calling `Bladeburner.getBlackOpNames()`. However most of the time users will not care about black ops that are already completed, or that can not yet be attempted. Really only the currently available Black Op is the one that matters. It'd be nice to have a function that returns a single name of just the black op that's relevant at that point in time.

[LeviPesin]
> Yeah, that'd be nice. Currently the following function can be used instead:
> 
> ```
> function getNextBlackOp(ns) {
> 	for (let op of ns.bladeburner.getBlackOpNames()) {
> 		if (ns.bladeburner.getActionCountRemaining('Black Operation', op) > 0)
> 			return op;
> 	}
> }
> ```

myCatsName:
# Linked issues

closes #66
"It'd be nice to have a function that returns a single name of just the black op that's relevant at that point in time."

# Documentation
image

# Feedback

- user-facing title `ns.bladeburner.getBlackOp()` seems useable - avoid hardcoding "next" / "current" / "available". 
- `.getBlackOpNames()` already exists as a way to list _all_ BlackOps names, so `getBlackOpName()` seems like conflict.
- the internal method _is_ called `getNextBlackOp()`
-  The return is `Name of current BlackOp as a string, or "All BlackOps Completed."` , which is maybe not ideal, not sure what else though. Also considered it could return Name and Rank requirement as an object, but `getBlackOpRank()` exists.


Zelow79:
Wouldn't it make more sense to expect the player to solve this? Since Black Ops only have a max count of 1. It's not hard to solve. getBlackOpNames returns them in order, so they just need to check for the first Black Op with a count greater than 0.

myCatsName:
> Wouldn't it make more sense to expect the player to solve this? Since Black Ops only have a max count of 1. It's not hard to solve. getBlackOpNames returns them in order, so they just need to check for the first Black Op with a count greater than 0.

Sounds like the same suggestion from the linked issue. And I do agree, it's practically easy and less than half the lines for a player to DIY. I worked on this just to move the issue along, but at the same time I'm not sure why not to offer it to API users.

Zelow79:
> > Wouldn't it make more sense to expect the player to solve this? Since Black Ops only have a max count of 1. It's not hard to solve. getBlackOpNames returns them in order, so they just need to check for the first Black Op with a count greater than 0.
> 
> Sounds like the same suggestion from the linked issue. And I do agree, it's practically easy and less than half the lines for a player to DIY. I worked on this just to move the issue along, but at the same time I'm not sure why not to offer it to API users.

The only real reason I have to say exclude it, is to avoid api bloat and potential name conflicts. I forgot the details, but Hydro explained it to me once, even if it's in a different name space 2 functions can't have the same name or something like that. getBlackOp() might be a safe name to use, so the 2nd part is a minor point. However having a ton of functions in one name space can feel a little crowded quickly. Honestly, I'm not against it, but I feel like it's kind of cutting out programming from the programming game.

Snarling:
So I disagree with the above feedback that it's removing an important programmatic challenge.

There is info that the player should need to use a programmatic approach to retrieve from the API, but the name of the one available blackop does not fit this category IMO. The current requirement to iterate a list of all blackops and do manual checking to determine which one is the next one you can do is just unnecessary clunkiness. The info is provided very directly to the player in the UI, so it also makes sense to provide it in a straightforward way in the API.

The current clunkiness is a bit like if you didn't have direct access to the name of your gang's faction, but instead were given an array of all gangFactions and had to iterate over that to see which one has a playerGang flag set.

That said I have some other feedback:

* I don't like the name getBlackOp, it is too general and does not communicate what it is going to do. Especially with the existence of existing functions like getBlackOpRank, I would assume from the name of it that the player would put in a blackop name and would get out some kind of blackop object. I think the name getNextBlackOp is a lot better and gives the player a much better idea of what the function is going to do without needing to look at the documentation.

* next vs current vs available: "available" is the most technically accurate but it just sounds bad, at least in the function name. Current I think is not accurate, and I think the documentation wording "the current black op" needs to be changed. It's currently available, but it's not a current black operation unless it is in progress. "Next" is the best word to use IMO, and the documentation should refer to "the next black op that can be completed" which also makes the word make sense in the case that the player is in progress on the black op.

* Returning a special magic string is something to avoid. In order to use this function for logic, a player would now have to hardcode this special magic string in their scripts. Instead, when an empty string is semantically accurate, that is what should be used as the special return for a string. In this case, empty string is accurate in the special case, because there is no next blackop. This allows for easy use in logic without any hardcoding, e.g:
```js
const nextBlackOp = ns.bladeburner.getNextBlackOp();
if (!nextBlackOp) {/* Logic to do if no black ops are left */}
```

Zelow79:
I never used the word challenge, but it would be removing something that should be very simple for anyone at this stage of the game to implement imo. You're the current maintainer so whatever you decide is for you to decide I guess, but I find it weird that so many things are getting simplified lately and yet, you were the one so against things like adding buffs to the game or QoL changes like additionalMS to hacking options. I guess I just don't understand what your visions are for this game and I don't think I ever will.

> * next vs current vs available: "available" is the most technically accurate but it just sounds bad, at least in the function name. Current I think is not accurate, and I think the documentation wording "the current black op" needs to be changed. It's currently available, but it's not a current black operation unless it is in progress. "Next" is the best word to use IMO, and the documentation should refer to "the next black op that can be completed" which also makes the word make sense in the case that the player is in progress on the black op.

I actually think `nextBlackOP()` and `getNextBlackOp()` both sound great, but for constructive feedback about better naming how about any of these:
* `blackOpAtHand()`
* `accessibleBlackOP()`
* `primedBlackOP()`
* `readyBlackOP()`
* `priorityBlackOP()`
* `paramountBlackOP()`

> * Returning a special magic string is something to avoid. In order to use this function for logic, a player would now have to hardcode this special magic string in their scripts. Instead, when an empty string is semantically accurate, that is what should be used as the special return for a string. In this case, empty string is accurate in the special case, because there is no next blackop. This allows for easy use in logic without any hardcoding, e.g:

I completely agree here and totally missed that. IMO, that should return `null`/`undefined` on empty.

myCatsName:
>I think the name getNextBlackOp is a lot better and gives the player a much better idea of what the function is going to do

What about just rolling rank and name into a BlackOp object? Cut `getBlackOpRank()` entirely. I wish we had data for how methods are adopted, but I think we'd all say getting rank required for a BOp name parameter is limited use. `getBlackOps()` could similarly update to return all {name, rank} objects. or `getAllBlackOps()`.

As you said, without some specificity, anything `getBlackOp` suggests an object. 

Otherwise, sure getNextBlackOp. I equally like Zelow's readyBlackOp (though that suggests some "ready" action vs. "get") and accesibleBlackOp (though if you don't have the rank, is it accessible? Similar to getAvailable) . They all fail in their own way.

> IMO, that should return null/undefined on empty.

I also thought `null` seems technically correct here (and allows for a falsy return?). `null` was my original plan but I know `string | null` could also be "less than ideal". Returning an empty string intuitively seems to lack feedback, but maybe it's all the same once described in docs and signature.

> I guess I just don't understand what your visions are for this game

It's a bit pointed, but yeah, seems like a side effect of guiding design principles not being shared or visible.

> Info provided directly to the player in the UI also makes sense to provide in API.

Something like this sentence can stand as a design principle. A project like this is in a position where documenting principles and values would be appropriate. Document principles, offer interpretation, refine. It's awkward when we can point to outside principles that reflect on situations arising in our context, but find those principles clash with whatever internal ideas are guiding decisions. 

Not that anyone needs an example, but - https://github.com/airbnb/javascript 

For Bitburner it would be some technical, some stylistic , some UI, some mechanical, some abstract, some human-focused. 

ETA: PS > "(Someone) is the current maintainer so whatever they decide is for them to decide I guess"
I think it's time we evolve a bit past this kind of decision making. It's not a personal dig, but we don't need a King Under the Mountain here, and the role of maintainer can be internally defined in a way that clarifies this.

myCatsName:
>I would assume from the name of it that the player would put in a blackop name and would get out some kind of blackop object

I've suggested an update to `.getBlackOp()` that does simply return an object, or `null`.  It is also renamed `.getNextBlackOp()`.
name and operation BlackOp object
return null

The empty string return just doesn't seem right when printed, IMO. If you say empty string is still the correct or consistent choice for cases like this, I respect it and will make the change.

>documentation should refer to "the next black op that can be completed"

Documentation should be improved now.
image


And following my logic, newest commit also includes a new `.getAllBlackOps()` to return all black op objects. This makes an iterable return that someone can use to get either `.name` or `.rank` properties of any BOp, or use the above option to simply get the available Bop.
getAllBlackOps() definition
image

Together these would mean `.getBlackOpRank(name)` and `.getBlackOpNames()` can both be deprecated, but avoids making a direct conflict with either.

PR #818, opened 2023-09-20
CORPORATION: Set per-function RAM costs

jjclark1982:
One potential way to approach the Corp API RAM cost situation is to set per-function RAM costs that are achievable but still limiting.

If we use 10GB per each read-only function and 20GB per each action function:

- Minimal script with one getter and one action: 31.6 GB
- Typical script like [this corp.js](https://github.com/Sordani/Bitburner-script-building/blob/main/corp.js): 782.1 GB
- Maximal script that calls every Corp API function: 951.6 GB

With this approach, there would be some incentive to split scripts into smaller parts that can be deployed on early-game servers in parallel. For example:

- 32GB server: a script that only calls `getInvestmentOffer()` and `acceptInvestmentOffer()`
- 64GB server: a "pseudo-market-TA" script that calls `getMaterial()`, `getMaterialData()`, `buyMaterial()`, `sellMaterial()`


d0sboots:
I like this approach of setting "real" costs that are asymmetric a lot. I think they're still too high; for instance it's impossible to write a smart supply script in the natural way, because it would have to look at both material and product data, and that won't fit on any starting server. Also, you don't start with *any* 64 GB servers.

That said, it's a big improvement on where we are.

jjclark1982:
It might also make sense to do something like singularity functions, and have lower RAM costs (1-2GB each) within Bitnode 3, and higher costs (10-20GB each) in other Bitnodes.

Ideally a player should be able to progress from manual play to partial automation to full automation. So it would be nice to fit a single-purpose smart supply script into an affordable server on Bitnode 3.1 or 3.2.

d0sboots:
As I've said before, you can already make the RAM trivial *outside* BN3 by just buying a large server. I don't think it's worth adding extra complications when they're easy to work around; better to focus on a solid foundation for BN3.

d0sboots:
Also wanted to say: Thanks for setting up that poll on the discord! Unsurprisingly, opinions seems to be pretty evenly split all over the place XD

PR #820, opened 2023-09-21
BUGFIX: Patch a minor hole in RAM calculation

d0sboots:
Using the `ramOverride` option, a worker script could be set to 1.73266GB (1.75/1.01) and then 1% more workers fit into available RAM.

Did some basic smoke tests that launching scripts works when it should and is denied when it should be.

### DIFF
--- a/src/Netscript/NetscriptHelpers.tsx
+++ b/src/Netscript/NetscriptHelpers.tsx
@@ -385,7 +385,9 @@ function updateDynamicRam(ctx: NetscriptContext, ramCost: number): void {
   ws.dynamicLoadedFns[fnName] = true;
 
   ws.dynamicRamUsage = Math.min(ws.dynamicRamUsage + ramCost, RamCostConstants.Max);
-  if (ws.dynamicRamUsage > 1.01 * ws.scriptRef.ramUsage) {
+  // This constant is just a handful of ULPs, and gives protection against
+  // rounding issues without exposing rounding exploits in ramUsage.
+  if (ws.dynamicRamUsage > 1.00000000000001 * ws.scriptRef.ramUsage) {
     log(ctx, () => "Insufficient static ram available.");
     const err = makeRuntimeErrorMsg(
       ctx,


src/Netscript/NetscriptHelpers.tsx
```diff
@@ -385,7 +385,9 @@ function updateDynamicRam(ctx: NetscriptContext, ramCost: number): void {
   ws.dynamicLoadedFns[fnName] = true;
 
   ws.dynamicRamUsage = Math.min(ws.dynamicRamUsage + ramCost, RamCostConstants.Max);
-  if (ws.dynamicRamUsage > 1.01 * ws.scriptRef.ramUsage) {
+  // This constant is just a handful of ULPs, and gives protection against
+  // rounding issues without exposing rounding exploits in ramUsage.
+  if (ws.dynamicRamUsage > 1.00000000000001 * ws.scriptRef.ramUsage) {
```
jjclark1982, line 390: Why is multiplication needed? Wouldn't all rounding errors be caught by `ws.scriptRef.ramUsage + 0.001`?

d0sboots: Addition is not sensitive to the magnitude of what you're adding to, so you need to add a larger quantity to avoid the term being rounded away completely. Like in this example: using .001 instead of (1.)00000000000001. That allows the exploit to creep back in, since now you could take .001GB free/script. (It's not as bad as what it was before, but it's still not as watertight as the best solution.)

Snarling:
Another option would be just requiring ramOverride to be a direct multiple of 0.01 or even 0.05 since that should be the smallest legitimate ram increment. Using nonstandard amounts of ram also causes most ram displays to be imprecise, since they usually just show 2 decimals.

But I think this is a good solution. Since we were doing a multiplication already here, we might as well multiply by something that is more correct.

d0sboots:
> Another option would be just requiring ramOverride to be a direct multiple of 0.01 or even 0.05 since that should be the smallest legitimate ram increment.

The fundamental issue is that "a direct multiple of 0.01 or even 0.05" doesn't *exist* as a concept in our standard computer arithmetic. 0.01 is 1/100 which is not a representable binary number; it can only be approximated. Same with 0.05 (which is 1/20). If we were dealing with integers (or other exactly representable numbers) we wouldn't need any of this, and could just use the simple/obvious comparisons.

Incidentally, I did first try to make a change to use MB instead of GB as the memory size (internally only, all the player-visible interfaces would still be in GB for compatibility). That would fix this and several other places where we have to do rounding to maintain consistency. Unfortunately, it foundered on the fact that we use the `Server` interface both internally *and* externally, so there was no way to make the change transparently.

Snarling:
Yeah makes sense

PR #821, opened 2023-09-21
MISC: spelling and consistency fixes in the NetscriptDefinitions.ts file

ficocelliguy:
Minor doc tweaks to reduce some of the code analysis warnings on these files.

### DIFF
--- a/markdown/bitburner.bladeburner.upgradeskill.md
+++ b/markdown/bitburner.bladeburner.upgradeskill.md
@@ -16,7 +16,7 @@ upgradeSkill(skillName: string, count?: number): boolean;
 
 |  Parameter | Type | Description |
 |  --- | --- | --- |
-|  skillName | string | Name of skill to be upgraded. Case-sensitive and must be an exact match |
+|  skillName | string | Name of skill to be upgraded. Case-sensitive and must be an exact match. |
 |  count | number | _(Optional)_ Number of times to upgrade the skill. Defaults to 1 if not specified. |
 
 **Returns:**
--- a/markdown/bitburner.bladeburner.upgradeskill.md
+++ b/markdown/bitburner.bladeburner.upgradeskill.md
@@ -9,14 +9,14 @@ Upgrade skill.
 **Signature:**
 
 ```typescript
-upgradeSkill(name: string, count?: number): boolean;
+upgradeSkill(skillName: string, count?: number): boolean;
 ```
 
 ## Parameters
 
 |  Parameter | Type | Description |
 |  --- | --- | --- |
-|  name | string |  |
+|  skillName | string | Name of skill to be upgraded. Case-sensitive and must be an exact match |
 |  count | number | _(Optional)_ Number of times to upgrade the skill. Defaults to 1 if not specified. |
 
 **Returns:**
--- a/markdown/bitburner.stanek.canplacefragment.md
+++ b/markdown/bitburner.stanek.canplacefragment.md
@@ -16,10 +16,10 @@ canPlaceFragment(rootX: number, rootY: number, rotation: number, fragmentId: num
 
 |  Parameter | Type | Description |
 |  --- | --- | --- |
-|  rootX | number | rootX Root X against which to align the top left of the fragment. |
-|  rootY | number | rootY Root Y against which to align the top left of the fragment. |
-|  rotation | number | rotation A number from 0 to 3, the amount of 90-degree turns to take. |
-|  fragmentId | number | fragmentId ID of the fragment to place. |
+|  rootX | number | Root X against which to align the top left of the fragment. |
+|  rootY | number | Root Y against which to align the top left of the fragment. |
+|  rotation | number | A number from 0 to 3, the amount of 90-degree turns to take. |
+|  fragmentId | number | ID of the fragment to place. |
 
 **Returns:**
 
--- a/markdown/bitburner.stanek.canplacefragment.md
+++ b/markdown/bitburner.stanek.canplacefragment.md
@@ -18,7 +18,7 @@ canPlaceFragment(rootX: number, rootY: number, rotation: number, fragmentId: num
 |  --- | --- | --- |
 |  rootX | number | rootX Root X against which to align the top left of the fragment. |
 |  rootY | number | rootY Root Y against which to align the top left of the fragment. |
-|  rotation | number | rotation A number from 0 to 3, the mount of 90 degree turn to take. |
+|  rotation | number | rotation A number from 0 to 3, the amount of 90-degree turns to take. |
 |  fragmentId | number | fragmentId ID of the fragment to place. |
 
 **Returns:**
--- a/src/ScriptEditor/NetscriptDefinitions.d.ts
+++ b/src/ScriptEditor/NetscriptDefinitions.d.ts
@@ -632,7 +632,7 @@ interface BitNodeMultipliers {
   CompanyWorkExpGain: number;
   /** Influences how much money the player earns when completing working their job. */
   CompanyWorkMoney: number;
-  /** Influences the amount of divisions a corporation can have have at the same time*/
+  /** Influences the amount of divisions a corporation can have at the same time*/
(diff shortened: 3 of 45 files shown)


src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -1631,7 +1631,7 @@ export interface Singularity {
    * purchasing a TOR router using this function is the same as if you were to
    * manually purchase one.
    *
-   * @returns True if actions is successful or you already own TOR router, false otherwise.
+   * @returns True if actions is successful or if you already own TOR router, false otherwise.
```
myCatsName: >True if actions is successful

=> True if action...
?

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -2506,7 +2506,7 @@ export interface Hacknet {
    * Returns false otherwise.
    *
    * @param index - Index/Identifier of Hacknet Node.
-   * @param n - Number of levels to purchase. Must be positive. Rounded to nearest integer. Defaults to 1 if not specified.
+   * @param n - Number of levels to purchase. Must be positive. Rounded to the nearest integer. Defaults to 1 if not specified.
```
myCatsName: image
Can something be rewritten for clarity? [Edit: or use `Will be rounded to the nearest integer.` as a standard from earlier in the file?]

As an example - given a hacknet-server-1 at Level 1, what outcome should be expected from `ns.hacknet.upgradeLevel(1,20.5)`
Error because 20.5 isn't rounded? upgrade to level 21? to level 22?

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -5868,7 +5869,7 @@ export interface NS {
    * RAM cost: 0.1 GB
    *
    * Returns a boolean indicating whether the specified file exists on the target server.
-   * The filename for programs is case insensitive, other file types are case sensitive.
+   * The filename for programs is case-insensitive, other file types are case-sensitive.
```
myCatsName, line 5872: predicate adjectives (describes what the subject "is") aren't hyphenated. In general there is no ambiguity in the meaning, so no reason to hyphenate "case-sensitive" in other cases either, but those are more style choices.

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -6162,6 +6163,7 @@ export interface NS {
    *
    * Write data to the given Netscript port.
    * @param portNumber - Port to write to. Must be a positive integer.
+   * @param data - the data to write to the port
```
myCatsName: No caps, no period? Another line uses `Data to write.`, if that's not an OK standard, should probably decide one.

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -7106,7 +7108,7 @@ export interface OfficeAPI {
    * @param size - Amount of positions to open
    * @returns Cost of upgrading the office
    */
-  getOfficeSizeUpgradeCost(divisionName: string, city: CityName | `${CityName}`, asize: number): number;
+  getOfficeSizeUpgradeCost(divisionName: string, city: CityName | `${CityName}`, size: number): number;
```
myCatsName, line 7111: need to `npm run doc` to get this updated in the relevant docs.

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -1844,7 +1844,7 @@ export interface Singularity {
    *
    * This function will return true if you successfully get a job/promotion,
    * and false otherwise. Note that if you are trying to use this function to
-   * apply for a promotion and you don’t get one, it will return false.
+   * apply for a promotion, and you don’t get one, it will return false.
```
myCatsName: maybe

>If you are using this function to apply for a promotion and don’t get one, the function will return false.

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -5272,6 +5272,7 @@ export interface NS {
    * It is possible to pass any React Node instead of a string.
    * See {@link ReactElement} and {@link ReactNode} types for additional info.
    *
+   * @param title - The new title for the tail window
```
myCatsName: Match style of nearby content if there is no special reason not to. Add period to `The new title for the tail window.`

Signature definitions in general might not need periods, but that's not a guideline we currently have.

src/ScriptEditor/NetscriptDefinitions.d.ts
myCatsName: In general - sure, "nearest integer" refers to a specific, countable noun, making "Rounded to **the** nearest integer" grammatically correct. 

But in the context, it's not the stylistic choice I would make - it's 1 of 4 sentences in a very small space and "nearest integer" is plenty understandable; it's the specific of what is rounded, how and by whom that is unclear.

src/ScriptEditor/NetscriptDefinitions.d.ts
ficocelliguy: I'm a bit hesitant to change this from a one-file PR to a 25-file PR. I think that doc command is part of cutting a release, but if the maintainers want I can run that for this PR.

myCatsName: I think it's the intended process.

> If your PR includes ns API function changes, do not manually modify markdown files.
> Instead, you can run `npm run doc` to autogenerate new markdown files that reflect your submitted API changes

d0sboots: +1 on it being part of the intended process. Really, those files should be autogenerated as opposed to checked-in, but there are issues with doing that at this point.

markdown/bitburner.stanek.canplacefragment.md
```diff
@@ -18,7 +18,7 @@ canPlaceFragment(rootX: number, rootY: number, rotation: number, fragmentId: num
 |  --- | --- | --- |
 |  rootX | number | rootX Root X against which to align the top left of the fragment. |
 |  rootY | number | rootY Root Y against which to align the top left of the fragment. |
-|  rotation | number | rotation A number from 0 to 3, the mount of 90 degree turn to take. |
+|  rotation | number | rotation A number from 0 to 3, the amount of 90-degree turns to take. |
```
myCatsName: This method signature (at least) is duplicating param names that should be trimmed.

image

markdown/bitburner.bladeburner.upgradeskill.md
```diff
@@ -9,14 +9,14 @@ Upgrade skill.
...
-upgradeSkill(name: string, count?: number): boolean;
+upgradeSkill(skillName: string, count?: number): boolean;
 ```
 
 ## Parameters
 
 |  Parameter | Type | Description |
 |  --- | --- | --- |
-|  name | string |  |
+|  skillName | string | Name of skill to be upgraded. Case-sensitive and must be an exact match |
```
myCatsName: >must be an exact match [PERIOD]

2 sentences - the first uses a period, so should the second.

myCatsName (changes requested):
If I can figure out how to contribute to your PR, I will be happy to help.

markdown/bitburner.stanek.canplacefragment.md
ficocelliguy: I'm not sure I understand what you are saying here. Can you clarify?

d0sboots: I think they're talking about how stuff like "rootX", "rotation" etc. is duplicated needlessly. It should be specified as the argument to param, but doesn't need to be duplicated in the description.

At first I thought it was just the rendering, but you can see it's present in the source definition file.

ncharris93 (approved):
LGTM!

PR #822, opened 2023-09-21
DRAFT: Add achievement if all nodes complete

myCatsName:
# Linked issue

#108

>"we could perhaps have an achievement for completing all the bit nodes? That would continue to be something to aim for when more bit nodes are added in the future."

# Documentation
Just a draft for feedback on name and design, and general interest. Is this an achievement we want?
active: 
ALLBN active
inactive:
BN inactive

AFAIK there are a 2-3 spots available.
The relevant detail deciding if BB is limited by Steam at 100 achievements or more:
steam achieve limit

# Bug
Some achievement `svg` are being re-positioned left. Affects some current icons. Draft SVG is centered like this outside of BB.
inkscape assassin x centered

### DIFF
--- a/assets/Steam/achievements/real/ALLNODES.svg
+++ b/assets/Steam/achievements/real/ALLNODES.svg
@@ -13,7 +13,7 @@
    version="1.1"
    id="svg5"
    inkscape:version="1.0.2-2 (e86c870879, 2021-01-15)"
-   sodipodi:docname="bitnode burner toast ff.xaml">
+   sodipodi:docname="bitnode assassin X.svg">
   <metadata
      id="metadata27">
     <rdf:RDF>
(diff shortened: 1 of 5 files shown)


Snarling:
I believe there is an issue with Steam where we actually cannot add more steam achievements because of the game's status (something like, because we're not actually an official steam partner / not verified in some manner?)

The dist changes error can be ignored, it's expected in this PR.

Snarling:
Actually it looks like we're only at 98 achievements, so we have room for at least 2 more. So possibly this can be added. In addition to this PR I would also have to add the achievement in the steam backend so that steam knows about it.

d0sboots:
I thought we were saving those for a new BN or 2?

Edit: This is a different PR than what I thought, it directly relates to BNs already. Although I *do* think that was the logic behind why those slots were open, don't interpret that as me being against this PR.

Snarling:
Yeah I don't know if the extra achievement slots are already spoken for. I guess with both BN14 and BN16 being in the works, those would kind of make sense as the last two slots.

It's possible that we won't always be limited to 100. I'm actually not seeing anything explicitly warning me that we're limited to 100 on the steam partner page for adding/editing achievements, so it's even possible that the limit already doesn't apply to us anymore.

d0sboots:
Time to add 3 copies of this achievement to test XD

myCatsName:
>  it's even possible that the limit already doesn't apply to us anymore.

https://partner.steamgames.com/doc/marketing/profile I still don't see a simple link to check, but you only need to discover if BB has limited profile features or not. May need to contact support if it's not obvious from a dev page.

PR #825, opened 2023-09-22
BUG: Minor fix to devmenu and .getOwnedSourceFiles()

myCatsName:
There was some possible confusion or conflict in reporting Source Files owned. Addressed with 2 changes

1 - `ns.singularity.getOwnedSourceFiles()` did not include "SF -1," the exploits "SF". Since I would expect it in a list of Owned SFs, and was confused by the `.length` without it, I updated that method. I would also recommend updating at least the single letter `".n"` property name, but didn't change it.
fixed count but bad param

For example even `"SF"` seems improved.
report sf -1 and update property name

2 - Added "Exploits" as a property to compare in the "Import Save" modal. Also tooltips, so it is more clear why "Source Files" can exceed the total nodes in game.
source files

- documentation suggestions on the exploits file.

# Bug fixes
- icon in dev menu caused console error. Fixed and looks same, tooltip still active. 
image

- fixed console error due to list element nesting affecting SF-5 on the augment page
add spans after

# note!
What's included here should be stable for testing.
I'm working on a further update, including refactoring "SFminus1" . But that requires more time and testing. It also leads to re-writing the getOwnedSF() method and parts of the save import details included in this PR. 
image
image


myCatsName:
I'm also a bit confused that `getOwnedSourceFiles()` was moved out of singularity 
image

and later un-done.  43b8cfec5c9756a9096f6ccdec2e9867750ec362
image

Is either better?

Snarling:
getOwnedSourceFiles is supposed to be soft deprecated, the expected location to get this info is from getResetInfo.

Probably this is not documented well enough (or maybe at all)

Snarling:
I think the "Source Files" field for the import modal could be renamed to "Source File Levels" and it would be a lot clearer without being too verbose. But the tooltip should be fine too if you prefer that.

So looking at the import modal, I think adding exploits as a new line item, instead of bundling them with the regular source file levels, is the right call. SF-1 aren't like actual source files in the ways that matter. Actual source files take a long time to earn, and if you accidentally load a game that was missing one, it could mean having to re-do a bunch of progression.

But this separation in how we're communicating them to the player also to me reinforces that they should *not* be provided on getOwnedSourceFiles or on the newer getResetInfo().ownedSF. We also can't just change what is provided there freely, as this is an API break, and also I think it's good for us to provide info to the player that mirrors the internal data when possible (and internally, exploits are actually separate from sourceFiles)

myCatsName:
>"Source Files" field for the import modal could be renamed to "Source File Levels"

Sure. 
> I think it's good for us to provide info to the player that mirrors the internal data when possible (and internally, exploits are actually separate from sourceFiles)

In this case, it's hard to follow your logic. Player expectation is set clearly in this example -
image

Someone made a choice to call exploits SFminus1 and from that point on, I think it's reasonable "SF -1" is counted in "OwnedSourceFiles". The logic to keep them separate in the file comparison is sound, and the tooltip can share that detail.

>this is an API break

With all the undocumented details API breaks have. Maybe consider using the Milestones feature? "This is an API break" comes up all the time - when are expected API breaks? What happens to features until then? Is this the right situation for a features branch? 

>getResetInfo().ownedSF

can update to include SF -1 as well

d0sboots:
An API break is any change to the API that could cause user scripts to break. This is not a new unique concept to BitBurner; all libraries have the same concept, and those that use semantic versioning won't have any API breaks in a patch or minor release.

BB is in a tougher spot because, unlike most libraries, users are *forced* (via Steam or the live website) to update to the latest version. It is *possible* to continue playing on an older version, but this requires a fair amount of effort and know-how.

As a result, we try to be very conservative with API breaks, and when they can't be avoided, to lump them together in a major release. There are also some things that are minor enough that we let them through anyway; for instance, I would not consider adding "-1" to the keys of `ownedSF` to be a significant break (even though it could affect scripts that were depending on `.empty()`).

Snarling:
> for instance, I would not consider adding "-1" to the keys of ownedSF to be a significant break (even though it could affect scripts that were depending on .empty()).

It could also break scripts that e.g. iterate through ownedSF entries to determine if any nodes need to be revisited to reach level 3. Currently the player can always consider the keys of ownedSF to be valid bitnode numbers, but that's not the case if we add SF-1 to it.

SF-1 isn't like other nodes where a level actually means anything useful to a player. SF1 level 2 means the player should do one more iteration of bitnode 1 to max that out. SF-1 level 2 means that any two exploits are owned. It doesn't say which two are owned, it doesn't say which ones are missing. Knowing that you have 2 levels of SF-1 does not provide any useful information. To provide useful info to the player about exploits, a new function would be needed (And it should probably be in NSExtra so it remains undocumented).

I agree that consistency with the UI is not there and that's a bad thing. But adding consistency with the UI means adding inconsistency with the internals, adding info in a way that is not actually useful to a player, and also causing an API break, all of which are also bad.

Also of note, actual source files owned are "reset info" in that you can define an individual node-level reset based on the specific source files owned, or on the combined levels of source files. SF-1 on the other hand have nothing to do with resets, as they can be accrued at any time with no impact on the rest of the game. Anyone actually using the provided sourceFiles as reset info would suddenly not be able to rely on that anymore, or would need to code in special exceptions to ignore levels of SF-1

myCatsName:
>To provide useful info to the player about exploits, a new function would be needed (And it should probably be in NSExtra so it remains undocumented).

~~I will work on that, and~~ you can reject the singularity.ts change, which seems to be the conflict here.

d0sboots:
> Also of note, actual source files owned are "reset info" in that you can define an individual node-level reset based on the specific source files owned, or on the combined levels of source files. SF-1 on the other hand have nothing to do with resets, as they can be accrued at any time with no impact on the rest of the game. Anyone actually using the provided sourceFiles as reset info would suddenly not be able to rely on that anymore, or would need to code in special exceptions to ignore levels of SF-1

That's a good point. In general, I'm not super for inclusion of SF -1 in resetInfo either; I was just using it as an example of something I consider (still) to be a minor API break.

PR #827, opened 2023-09-23
MISC: Fix check for "True Recursion" achievement

jjclark1982:
The `bitburner-legacy` iframe uses `postMessage()` to report achievement status to the main game in its parent window. It sends a message with `{data: false}` when the achievement is not yet unlocked, and `{data: true}` when the achievement is unlocked. `BBCabinet.tsx` receives this message.

When a player has React-Devtools running in their browser, it also uses `postMessage()` to send things like `{data: { source: "react-devtools-bridge", payload: {…} }}` between these frames. This value is truthy and comes from an appropriate origin, so the achievement can be unlocked accidentally for players who took no action.

This PR changes the check to `event.data === true`, which can still be crafted by a player looking for a shortcut, but is unlikely to come from a third-party source accidentally.

### DIFF
--- a/src/Arcade/ui/BBCabinet.tsx
+++ b/src/Arcade/ui/BBCabinet.tsx
@@ -14,7 +14,7 @@ const style = {
 export function BBCabinetRoot(): React.ReactElement {
   useEffect(() => {
     window.addEventListener("message", function (this: Window, ev: MessageEvent<boolean>) {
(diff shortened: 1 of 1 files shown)


d0sboots:
What's the actual failure case here? Perhaps you need a description. XD

jjclark1982:
I have confirmed that the problem exists with React DevTools on https://bitburner-official.github.io/ . The original reporter @stillhart does not have this browser extension, but it is easy to see how other extensions could cause the same problem.

PR #830, opened 2023-09-25
UI: Make Source-Files devmenu page reactive

myCatsName:
# Documentation

Source-Files devmenu page now indicates owned SF levels, and updates reactively.

Old:
image

=> New:
image
image
image
image
image
image

### DIFF
--- a/src/DevMenu/ui/SourceFilesDev.tsx
+++ b/src/DevMenu/ui/SourceFilesDev.tsx
@@ -1,4 +1,5 @@
 import React from "react";
+import { useState } from "react";
 
 import Accordion from "@mui/material/Accordion";
 import AccordionSummary from "@mui/material/AccordionSummary";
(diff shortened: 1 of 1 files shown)


myCatsName:
Also available as an extension of this PR -  adds feature to increment SF-12 by 10 or 100

https://github.com/bitburner-official/bitburner-src/compare/dev...myCatsName:bitburner-src:add-lot-of-sf12-button


image

Snarling (commented):
I'm going to make some tweaks after merging this, so just jotting down some stuff that will be addressed.

1. sfData state variable isn't being used for anything except triggering a rerender, so it doesn't need to be complex data. If a state variable is needed only for triggering a rerender, use the useRerender custom hook to get a rerender function

2. When overriding styles, doing it through static colors on a style object is not the right way. With MUI components, the better way to change a component's color is with the color prop, and even when we're using direct style overrides we should still be referencing the theme colors instead of using a hardcoded color like black or green.

I also thought the color change from this PR didn't make the usability of the buttons obvious. On first glance I actually thought the Set All buttons (and other buttons with black backgrounds) were all disabled since they didn't look like normal buttons anymore. Messing around with it a bit, I'm not really finding a way of doing the colors that doesn't look bad and also communicates what the status of everything is.

Here's an example using the Mui color prop, but it's still confusing:
image

On that attempt I colored them "error" (clicking button will destroy SF levels), "warning" (clicking button effectively does nothing), and "success" (clicking button adds SF levels). However when the game's default theme also uses green for regular text ("success" and "primary" are the same or nearly the same), it then leaves the question of why are all the "Set All" buttons also green?

I think I will remove the alternate styling and just provide a text indicator of the current SF level for each one. If there's a better way to use color for differentiation I'm open to that, but it would need to use the player's theme, and should still create buttons that still look like the rest of the buttons from the game.

PR #833, opened 2023-09-27
MISC: Readthedocs configuration that redirects to this repo

### LINKED ISSUE #824: Documentation Website Needs To Be Updated
# Documentation Website Needs To Be Updated

If you search for bitburner documentation, this website is usually at the top of the list:  https://bitburner.readthedocs.io/en/latest/

But it is using this archived and deprecated repository: https://github.com/danielyxie/bitburner/tree/master

Can you please change the website to point to this (the new) repository? So users can get the latest documentation easily and without getting confused.

Right now, the website has a lot of broken links and missing data which can be dissuading for a new player.

Hoekstraa:
This PR reintroduces a ReadTheDocs configuration, that in combination with some tweaks on the ReadTheDocs side (which I will make once this merges, but have tested with my own fork) redirects people from ReadTheDocs to the documentation available at https://github.com/bitburner-official/bitburner-src/blob/dev/src/Documentation/doc/index.md.

It:
- closes #824
- stops the e-mails the ReadTheDocs admins have been getting (finally!!)
- redirects people to up-to-date documentation, what more do you want?

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,0 +1,16 @@
+# .readthedocs.yaml
+# Read the Docs configuration file
+# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+
+# Required
+version: 2
+
+# Set the version of Python and other tools you might need
+build:
+  os: ubuntu-22.04
+  tools:
+    python: "3.11"
+
+# Build documentation in the docs/ directory with Sphinx
+sphinx:
+  configuration: doc/conf.py
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -1,0 +1,34 @@
+# Configuration file for the Sphinx documentation builder.
+#
+# This file only contains a selection of the most common options. For a full
+# list see the documentation:
+# https://www.sphinx-doc.org/en/master/usage/configuration.html
+
+# -- Path setup --------------------------------------------------------------
+
+# If extensions (or modules to document with autodoc) are in another directory,
+# add these directories to sys.path here. If the directory is relative to the
+# documentation root, use os.path.abspath to make it absolute, like shown here.
+#
+import os
+import sys
+
+# -- Project information -----------------------------------------------------
+
+project = 'Bitburner'
+
+# -- General configuration ---------------------------------------------------
+
+# Add any Sphinx extension module names here, as strings. They can be
+# extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
+# ones.
+extensions = [
+              ]
+
+# List of patterns, relative to source directory, that match files and
+# directories to ignore when looking for source files.
+# This pattern also affects html_static_path and html_extra_path.
+exclude_patterns = ['*']
+exclude_patterns = ['doc/index.rst']
+
+# -- Options for HTML output -------------------------------------------------
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -1,0 +1,5 @@
+.. meta::
+   :http-equiv=Refresh: 0; url='https://github.com/bitburner-official/bitburner-src/blob/dev/src/Documentation/doc/index.md'
+
+This link is outdated as documentation for Bitburner has been migrated to an in-game menu, this page should have redirected you to the new location.
+You can also click `here to go to the game's documentation <https://github.com/bitburner-official/bitburner-src/blob/dev/src/Documentation/doc/index.md/>`_.
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -1,5 +1,5 @@
 .. meta::
-   :http-equiv=Refresh: 0; url='https://github.com/bitburner-official/bitburner-src/blob/dev/src/Documentation/doc/index.md'
+   :http-equiv=Refresh: 0; url='https://github.com/bitburner-official/bitburner-src/blob/stable/src/Documentation/doc/index.md'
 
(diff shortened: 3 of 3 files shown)


d0sboots:
Is it possible to not hardcode the repo + branch name? This is not entirely theoretical; we used to publish different readthedocs from different branches, and there's still the dev/beta/release split in terms of versions.

Snarling:
Yeah there should be the ability to point to stable instead of dev.

If we point to dev all the time, newly added api functionality from the current dev cycle will be shown to players who are playing stable branch. Additionally, when we are doing the 3.0 dev cycle and adding in all the major API breaks we want to do piece by piece, the existing documentation for the stable version essentially "disappears."

But with RTD being deprecated to just be a redirect, I guess it's not a big deal. But if it's not simple to make it capable of being dynamic, I think it should just statically target the stable branch instead of dev.

Snarling:
Actually the more I think about it, just always redirecting all the RTD to stable branch is probably the way to go. If players want dev branch documentation, they can use the links from ingame in the dev version.

Snarling (changes requested):
I would make the change myself but I don't know if it requires changes on the RTD side as well. I think we should link RTD's to the stable branch docs instead of dev, to prevent normal players from being redirected to docs that aren't for their current version.

Hoekstraa:
Will change to stable when I have the time. I was checking some things on the RTD side and there's some issues there still.
I'm going to have to do some more work, because RTD seems to have removed some features over time that we could've used for this. E.g. it seems to me they (actively?) try to prevent projects from doing these types of redirects.

Plan now is to test if we can remove everything and only have a 404 page that redirects to the docs.
I'll make sure that one will link to the stable version. ;)

d0sboots:
> E.g. it seems to me they (actively?) try to prevent projects from doing these types of redirects.

I could have sworn I read some docs (on the RTD side) that talked about how you could do exactly this. Have they broken that?

Hoekstraa:
Yeah, I've sent some docs about it in the Discord, those exact pieces of
advice and even the official documentation refer to an option that just
isn't available (anymore) it seems. Specifically the 'force redirect'
option if I remember it correctly off the top of my head. I'm not sure, but
it might be a ReadTheDocs premium/gold/enterprise option, it has been
removed, or it never got shipped, but it's not where I expect it to be.

This essentially breaks redirecting to a specific link.
This is important as users might not arrive at the root of the
documentation, but say search for gang info and land at the gang info. In
every case I've found it adds the subcontents of the url after the
redirect. For example:

bla.readthedocs.blegh/en/latest/randompageaboutgangs

Will turn into:

github.com/bitburner-official/bla-src/bla/bla/src/Documentation/index.md/randompageaboutgangs

Regardless of type of redirect. Trust me, I've most assuredly tried more
than 60 variations of them by now.
Overall the redirection is a big ol' mess. Hence my next idea would be to
strip the readthedocs of anything and use a custom 404 with a meta
redirect, so that we can avoid having to deal with readthedocs' internal
router.

Sorry for the crappy explanation as I'm on my phone as I'm writing this
through e-mail.
Let me know either here or on Discord if anything's unclear.

Snarling:
Have you tried including a useless html query at the end of the URL? Then you'd end up with a url like:
https://github.com/bitburner-official/bitburner-src/blob/stable/src/Documentation/doc/index.md?none/of/this/does/anything.md

Which still points to the main documentation page.

Snarling:
@Hoekstraa Did you see my previous comment about possibly using a query to throw away extra path info and always land at the same page? Not sure if it will work but it may be worth a shot.

Hoekstraa:
@Snarling Thank you for the suggestion. Sorry for the delay, I had some important work to do last week so didn't have the time to dig into it. I'm actually looking at it right this moment, and the query suggestion seems to help in the case I described. It doesn't seem to work 100% of the time as there's some other edge cases, so we still need to pull some code into the repo.
I'm hoping to get a fix out this evening.

Hoekstraa:
Changes on the RTD side have been made for bitburner-official.readthedocs.io.
We just need the changes made here for it all to line up and redirect all requests to that webpage to the Github docs.

Hydro is going to try and get access for the other pages (bitburner, bitburner-beta) next week he said. So we'll get those aligned if and when we can.

Snarling:
Sorry, just now checking back on this. Are the changes here still needed? I'm seeing all of these pages now redirect to the documentation, so it looks like you got it working without any changes needed here?

https://bitburner-src.readthedocs.io/
https://bitburner-src.readthedocs.io/en/
https://bitburner-src.readthedocs.io/en/latest/
https://bitburner-src.readthedocs.io/en/latest/advancedgameplay/

Hoekstraa:
Oh! Sorry for the confusion. Bitburner-src is actually my testing environment for this (based on my branch).
It's bitburner-official leading to this repository, and bitburner and bitburner-beta leading to the older repositories.
Once we get ownership of bitburner and bitburner-beta we'll redirect them to here too.

The thing is that search engines will point to either bitburner-official, but most likely bitburner and bitburner-beta.
So if we want to redirect them, we need the changes made in this PR.

I'm glad though that https://bitburner-src.readthedocs.io/ works for you, because that means my demo/PR works correctly ;)

Hoekstraa:
So once we merge this PR, https://bitburner-official.readthedocs.io/ should behave the same, and when we get control of the others, they can be converted to use this PR too. So this is the only change necessary.

Snarling:
Oh okay, I misunderstood lol. I was wondering why searching was never bringing that one up.

PR #835, opened 2023-09-27
UI: add hashes and update general devmenu page

myCatsName:
old
image

=> new
image

- create RAM button group
- add hash feature
- ui update
- refactor
- fix typos and nesting error on unrelated pages

### DIFF
--- a/src/DevMenu/ui/General.tsx
+++ b/src/DevMenu/ui/General.tsx
@@ -122,7 +122,7 @@ export function General(): React.ReactElement {
       </AccordionSummary>
       <AccordionDetails>
         {moneyValues.map((value) => (
-          <Button onClick={addMoney(value)}>
+          <Button key={`add money ${value}`} onClick={addMoney(value)}>
             <pre>
               + <Money money={value} />
             </pre>
--- a/src/DevMenu/ui/General.tsx
+++ b/src/DevMenu/ui/General.tsx
@@ -8,6 +8,7 @@ import {
   Select,
   SelectChangeEvent,
   TextField,
+  Tooltip,
   Typography,
 } from "@mui/material";
 import ExpandMoreIcon from "@mui/icons-material/ExpandMore";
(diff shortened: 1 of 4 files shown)


src/DevMenu/ui/General.tsx
```diff
@@ -28,14 +29,25 @@ export function General(): React.ReactElement {
   const [corporationName, setCorporationName] = useState("");
   const [gangFaction, setGangFaction] = useState(FactionName.SlumSnakes);
   const [devMoney, setDevMoney] = useState(0);
+  const [hash, setHash] = useState(Player.hashManager.hashes);
+  const [homeRam, setHomeRam] = useState(Player.getHomeComputer().maxRam);
```
Snarling, line 33: Similar to a remark from #830 , if a state variable isn't being used for anything except the act of triggering a state update (to trigger a rerender), then it's preferable to just use the useRerender hook to get a rerender function. homeRam/setHomeRam are only being used to tigger a rerender.

src/DevMenu/ui/General.tsx
```diff
@@ -91,89 +103,152 @@ export function General(): React.ReactElement {
...
+  const largeButtonStyle = { width: "20rem" };
+  const noArrowsNumberField = {
+    "& input::-webkit-outer-spin-button, & input::-webkit-inner-spin-button": {
+      display: "none",
+    },
+    "& input[type=number]": {
+      MozAppearance: "textfield",
+    },
+  };
+
```
Snarling, line 117: Not a fan of the way these overrides are being done, but all uses of makeStyles need to get changed eventually anyway so it's not a huge deal.

However, I don't think the static width buttons look good, with widely varying amounts of effective padding.

There is also no need to make a special definition for a number input just for DevMenu/General. There is an existing standardized NumberInput component, which also handles parsing player input using bitburner's numeric suffixes.

So I don't think these overrides are necessary anyway.

Snarling (changes requested):

Snarling:
I will just make followup changes in a separate PR

PR #840, opened 2023-10-02
MISC: Fixed typo in Interactive Tutorial

Hexy32:
# MISC: Fixed typo in Interactive Tutorial

MISC: There was a small typo in the tutorial so I fixed it :D

### DIFF
--- a/src/ui/InteractiveTutorial/InteractiveTutorialRoot.tsx
+++ b/src/ui/InteractiveTutorial/InteractiveTutorialRoot.tsx
@@ -258,7 +258,7 @@ export function InteractiveTutorialRoot(): React.ReactElement {
 
           <Typography>
             {" "}
-            shows that there do not need to be any open ports on this machine for the NUKE virus to work, so go ahead
+            shows that there does not need to be any open ports on this machine for the NUKE virus to work, so go ahead
             and run the virus using{" "}
           </Typography>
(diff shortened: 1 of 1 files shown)


Snarling:
This is not correcting anything, I believe it's actually taking a more correct sentence and making it less correct.

The phrase "there does not need to be any ports" doesn't get flagged directly as a grammar mistake by online grammar checkers I've found (even though it sounds terrible to me), but it also doesn't get flagged when keeping it as "there do not seem to be any ports."

However, with the same sentence structure but different word choice, I am getting corrected for trying to use your "does not" wording with a plural:

image

"Ports" is plural like "issues" above, so we should keep "do not" instead of switching to "does not."

d0sboots:
I went round and round on this - at first I thought the correction was right. So I thought I'd share:

As Snarling identified, the key sub-phrase is "there does/do not need to be any open ports." To me, both versions sound "rightish." This is because the verb (to do) has to agree in number with "there," but "there" is actually an expletive(!) in this usage, and really just refers to the *delayed subject* which comes later.

As a result, our brains (and apparently grammar checkers) are used to "there" having either type, and can be fooled if there's a lot of stuff in between. But if we rearrange the sentence to eliminate "there, " it becomes clear:

"Any open ports do not need to be" vs "Any open ports does not need to be"

Hexy32:
@d0sboots you are totally right, my brain was saying it was wrong until I read these comments. Brains are weird, now "do" sounds better. Thanks for your time!

PR #838, opened 2023-09-30
CORPORATION: UI to display production and sales multipliers

jjclark1982:
This PR collects production and sales multipliers in tooltips on the office stats screen. These values were already available to players but difficult to work with due to being in different formats on different pages.

### Summary of changes

- Tooltip for material/product buy/sell/prod/export amounts: re-ordered according to game cycles, positive/negative sign of values changed to always be relative to local warehouse space
- Tooltip for Office material/product production amount: reviewed wording and added breakdown of factors
- Tooltip for Office sales multiplier: reviewed wording and added breakdown of factors
- Tooltip for Division advertising factors - added formula, updated table formatting
- Created a new formatter function for Corporation multipliers, to look good right-aligned with formatted Corp stats but include "x", applied it throughout Corp UIs

material breakdownproduct breakdown
sales breakdown
Screenshot 2023-10-02 at 10 44 24 AM

---

DarkTechnomancer:
> IMO the biggest help for the lowest effort would be exposing the mechanics in-game. Make VeChain free for a start, but then the somewhat more complicated fix would be moving the multipliers to the places where they apply rather than just showing them in a vacuum. Seeing that you have a x14 Business multiplier and a x0.74 advertising multiplier isn't really useful if you don't know what those mean.
> 
> For example, I think it would be better if you could hover over the number on material production and get a tooltip that breaks it down like 1.4 (Employee Production) * 128 (Warehouse Production Bonus) * 1.1 (Research) * 1 (Morale) = 197.12

PR #841, opened 2023-10-02
UI: Add slight left padding to non-first cells in StatsTable rows

Snarling:
* Reorganize/simplify imports
* Rename variables
* Add classes for first cell in a row vs. other cells
* Add left padding to cells other than first cell

### DIFF
--- a/src/ui/React/StatsTable.tsx
+++ b/src/ui/React/StatsTable.tsx
@@ -1,29 +1,33 @@
-import React from "react";
+import React, { ReactNode, ReactElement } from "react";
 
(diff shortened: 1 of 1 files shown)


jjclark1982:
This looks like a good way to do it. I have confirmed that the tables display with more space between columns, but still line up properly with elements above and below them.

Here is a branch merging this PR with #838 so that there are more tables to look at: https://jjclark1982.github.io/bitburner-src/

Snarling:
@jjclark1982 Oh I only opened up a PR because I didn't see anything that seemed related. I've cherry picked the commit over to that PR

PR #845, opened 2023-10-03
API: Add `nextUpdate()` promise for systems with bonus time

jjclark1982:
This PR adds `nextUpdate()` Netscript API functions which sleep until the next update of their respective system has been processed, similar to `port.nextWrite()`.

Promises resolve to information about the update that has just been processed. For example:

```js
const prevCorpState = await ns.corporation.nextUpdate();
ns.print(`I'm done with ${prevCorpState}, next will be ${ns.corporation.getCorporation().state}`);

const prevBladeDuration = await ns.bladeburner.nextUpdate();
ns.print(`I'm done with ${ns.tFormat(prevBladeDuration)} of Bladeburner actions.`);

const prevGangDuration = await ns.gang.nextUpdate();
ns.print(`I'm done with ${ns.tFormat(prevGangDuration)} of Gang activity.`);

await ns.stock.nextUpdate();
let stockBonusTime = ns.stock.getBonusTime();
```

These new APIs process multiple calls in FIFO order (unlike `port.NextWrite` which still processes in LIFO order, to remain compatible with existing scripts).


src/Netscript/RamCostGenerator.ts
```diff
@@ -355,6 +359,8 @@ const grafting = {
 const corporation = {
   hasCorporation: 0, // This one is free
   getConstants: 0,
+  getBonusTime: 0,
+  nextCycle: RamCostConstants.CycleTiming,
```
d0sboots: I think this is the wrong cost, since the API it would replace (getCorporation.getState()) has a cost of CorporationInfo.

(I think that will be too expensive, but I think *all* the corp costs are too expensive, so at least this is consistent.)

src/Corporation/Corporation.ts
```diff
@@ -109,6 +111,11 @@ export class Corporation {
       const gameCycles = marketCycles * corpConstants.gameCyclesPerCorpStateCycle;
       this.storedCycles -= gameCycles;
 
+      // Handle "nextCycle" resolvers at the start of this cycle
+      while (this.resolvers.length > 0) {
+        this.resolvers.shift()?.(state);
+      }
```
d0sboots: Same comment for all these loops:
This is needlessly O(n^2), since it's O(n) for each shift.
```suggestion
      for (const res of this.resolvers) {
        res(state);
      }
      this.resolvers.length = 0;
```

d0sboots (approved):
At first, I was worried because things like this have been suggested before, but there's always the risk of assuming that one callback = one cycle, when in fact the game can "skip" and do multiple cycles, especially if there is lag.

However, the API is well designed to give the user tools for that, and at least in the corp case, the game itself will never actually advance more than one state at a time.

src/Corporation/Corporation.ts
jjclark1982: presumably the best performance would be 

```js
for (const resolver of resolvers) {
  resolver();
}
resolvers.length = 0;
```

but `pop()` seems more sane

d0sboots: The access of `resolvers` inside `this` will essentially be hoisted out by the JIT compiler... (not exactly, but essentially), so there's no need to do that explicitly unless this is the tightest of loops and profiling shows it matters.

Optional: However, there's a subtle issue with `pop()`: It changes the behavior from being queue-like to stack-like, i.e. the first awaiter on `nextCycle` will become the last to be resolved. This is the opposite of the typical behavior of the microtask queue (that's why it's called a queue), and thus (IMO) against the expectations of users.

Although there's countless ways to do this in JS, I think the `for..of` loop is the clearest and most correct way for this situation.

jjclark1982: Ah, it looks like the `length = 0` pattern is already in this code base a lot.

I am running into some trouble with saving and loading data with the `for ... of` form, that I think `shift()?()` and `pop()?()` were avoiding. Is there a good way to block these resolvers from being included in save data so that they don't have to be checked at every access?

src/NetscriptFunctions/Bladeburner.ts
```diff
@@ -325,9 +325,8 @@ export function NetscriptBladeburner(): InternalAPI<INetscriptBladeburner> {
       const bladeburner = getBladeburner(ctx);
       return Math.round(bladeburner.storedCycles / 5) * 1000;
     },
-    nextCycle: (ctx) => async () => {
-      const bladeburner = getBladeburner(ctx);
-      return new Promise<number>((res) => bladeburner.resolvers.push(res));
+    nextCycle: () => async () => {
```
d0sboots: Optional: A minor thing, these don't need to be async, they're just regular functions that return a promise. I think having them be `async` is technically observable in terms of exactly how the resulting promise is wrapped, but not in a way anyone should be depending on.

d0sboots (commented):
Ah, putting the resolvers outside the class is a slightly nicer solution.

src/Netscript/RamCostGenerator.ts
jjclark1982: I reckon this approach is reasonable as it is similar to the situation with `ns.getServer()` for 2 GB vs the various `ns.getServer___()` functions for 0.1 GB each. It adds some new tradeoffs to consider when designing small single-purpose corp scripts, but it does not break any existing APIs.

Caldwell-74:
what about sleeves maybe even player tasks?
could maybe use a await taskCompletion ?

also not a blocking issue but something to consider is
that time sensitive strategys like gang flash TW are getting alot easier to manage with this

edit.: what about nextUpdate instead of nextCycle, makes the functionality more clear?

jjclark1982:
> what about nextUpdate instead of nextCycle, makes the functionality more clear?

Yes, that sounds like more appropriate language across various systems.


> what about sleeves maybe even player tasks?
> could maybe use a await taskCompletion ?

That seems like a bigger balance shift. I wonder whether it would be better to have a single person-wide taskCompletion promise, or turn individual active tasks into await-able functions like ns.hack?

jjclark1982:
As this is implemented now, promises resolve during an update and give information about that update to the calling script, which does not process its `await` or `.then()` until the update has finished.

So the caller is operating on out-of-date information, for example:

```js
let corpState = await ns.corporation.nextCycle();
ns.print(`${corpState}, ${ns.corporation.getCorporation().state}`);

// prints: EXPORT, SALE
```

So the promise resolutions should be moved to the end of each system update function, or their handlers should be given a tick to run before continuing with the update.

I would prefer not to change `engine.updateGame()` into an async function.

Caldwell-74:
does it need to return the state?

edit.: or just change the description to something along the line of "returns the state that just processed"

d0sboots:
If it doesn't return the proper state, this function will become a n00b trap due to the awkward API. Same for the other versions: they ought to return whatever makes the most sense to view as being "processed".

We only really get one chance to make new APIs correct, so we should nail that first and then deal with the code implications.

> I would prefer not to change `engine.updateGame()` into an async function.

Under what circumstances would you need to? Triggering promises doesn't require `async`, you only need that keyword when you will be `await`ing in the function.

jjclark1982:
> Under what circumstances would you need to? Triggering promises doesn't require async, you only need that keyword when you will be awaiting in the function.

If we want to give user scripts a chance to run right before an update, those user scripts would be running with `await` or `.then()`, so the engine would need to give them a tick to run. I think it's better to let user scripts run right after an update instead of trying to squeeze them in right before an update.

d0sboots:
> I think it's better to let user scripts run right after an update instead of trying to squeeze them in right before an update.

Ah, now I see what you were getting at. And I agree, running after update seems most natural (for a number of reasons)

jjclark1982:
Commit [487a773](https://github.com/bitburner-official/bitburner-src/pull/845/commits/487a773a075b6f7a42e160147ed8cf3431df79d4) changes the `prevState` output of `corporation.nextUpdate()` to the number of milliseconds processed, to be consistent with other `nextUpdate()` functions and avoid the redundancy with `getCorporation().state`.

This means the total RAM cost for responding to the current state will be higher, and tracking the previous state will be harder. It would make sense to remove this information here if PR #861 could also add an API to read `getCorporation().prevState`.

@Caldwell-74 what do you think about this approach?

Caldwell-74:
if snarling and d0s are okay with it then i would add prevState and nextState and mark state as beeing deprecated

d0sboots:
I think the Corp version should still return a state of some form. The number of milliseconds processed is never useful information for Corp, whereas the state is vital. IMO the proper way to think about consistency for these functions is "they should all return the most useful information."

Edit: Number of milliseconds is not particularly useful for most of the other mechanics, either. It's just that they don't have an obvious state or other clear data that's better to return.

jjclark1982:
I agree that all these promises should return the most useful information possible, which generally means the type or magnitude of the update that just happened.

I would also like to make usage as clear and consistent as possible, to enable usage like this:

```js
while (await ns.gang.nextUpdate()) {
  // Manage the Gang
}
while (await ns.bladeburner.nextUpdate()) {
  // Manage the Bladeburner division
}
while (await ns.stock.nextUpdate()) {
  // Manage your stock portfolio
}
```

As these 3 only vary in magnitude, having them all return the number of milliseconds is clear even when it is not actionable. And having them return a truthy value supports this while-loop format. (Note that astute users can convert the number of Gang milliseconds to the number of cycles, so this makes Gang micro-optimization easier without making it trivial.)

But Corporation varies in type instead of magnitude, so it would be most clear if it continues to return nextState:

```js
 while (true) {
   const prevState = await ns.corporation.nextUpdate();
   switch (prevState) {
     // Manage each state-specific case
   }
   // Manage things that happen every Corporation update
 }
```

Snarling:
Sorry it took so long for me to look at this.

I thought there would be an issue with the nextCycle functions all having the same name and having a non-zero ramcost, but it looks like the dynamic ram checker is only tracking the function name, not the actual specific function you are allocating ram for.

Currently, the ram cost is 1GB, but you can run all 4 of the nextCycle functions off that 1GB which is kind of weird. It does mean every nextCycle function must have the same cost, and that it's effectively a buy one, get all.


About the example code posted above, I don't think this is how people would want to use these functions together in the same script.

>I would also like to make usage as clear and consistent as possible, to enable usage like this:
```js
while (await ns.gang.nextUpdate()) {
  // Manage the Gang
}
while (await ns.bladeburner.nextUpdate()) {
  // Manage the Bladeburner division
}
while (await ns.stock.nextUpdate()) {
  // Manage your stock portfolio
}
```

A couple issues with the above as a usage examples. First, it doesn't actually use the value that these promises are resolving to, which might end up being normal for gang/blade/stock since the return value doesn't matter much, but it won't be normal usage for corporation.

Second, the way that is written it would never get to the second while loop, because the first one would always return a truthy value and it would just stay in the first loop.

It also doesn't work well to just do the updates within the same loop, e.g.:
```js
/** @param {NS} ns */
  while (true) {
    await ns.bladeburner.nextUpdate().then(bladeUpdateHandler);
    await ns.corporation.nextUpdate().then(corpUpdateHandler);
    await ns.gang.nextUpdate().then(gangUpdateHandler);
    await ns.stock.nextUpdate().then(stockUpdateHandler);
  }
}
```

Because then, we're not actually catching every update like we probably wanted to. For instance multiple blade cycles will pass while it waits for all the other promises. Instead, the "right" way to use it would be to launch a separate async function (or similar functionality with promise chains) for each of the types of update. There are a lot of ways to do it, but here's an example:

```js
/** @param {NS} ns */
  async function repeatedlyHandleUpdates(nextUpdateFn, updateHandler) {
    while (true) await nextUpdateFn().then(updateHandler)
  }
  repeatedlyHandleUpdates(ns.bladeburner.nextUpdate, bladeUpdateHandler)
  repeatedlyHandleUpdates(ns.corporation.nextUpdate, corpUpdateHandler)
  repeatedlyHandleUpdates(ns.gang.nextUpdate, gangUpdateHandler)
  repeatedlyHandleUpdates(ns.stock.nextUpdate, stockUpdateHandler)
  // Just to prevent the main function from ending
  while(true) await ns.asleep(60000);
}
```

I would just provide the player with good usage for one function and let them figure out how to work it together into the same script though. But the single-function usage example should show how to use the resolved promise value.

d0sboots:
I think they already have a good example of how to use the returned value for Corp:
```js
 while (true) {
   const prevState = await ns.corporation.nextUpdate();
   switch (prevState) {
     // Manage each state-specific case
   }
   // Manage things that happen every Corporation update
 }
```
You also don't need any promise-chaining to deal with this. For instance, your function can be written as:
```js
async function repeatedlyHandleUpdates(nextUpdateFn, updateHandler) {
  while (await nextUpdateFn()) updateHandler();
}
```
That's really just throwing an async wrapper around the existing while loop. I suspect it won't be needed very often here though, because these particular mechanics (gang, stock, bladeburner, corp) are rarely used together.

jjclark1982:
Regarding the combined RAM costs, it seems like it would be possible to have a single "master control program" wait on various timers and `exec` only one other script at a time, as long as it could wait for that script to complete without missing any other timers. That would make a good puzzle to optimize RAM by writing individual scripts that can terminate immediately and load/save persistent state through other channels.

(This also suggests that a promise-based interface to `ns.isRunning()` would be useful, similar to bash's `wait` command.)

I'll write the docs to encourage one script per mechanic, with no hints about RAM-dodging or flash-mobbing.

---

Here is a detailed usage example:
```js
  let lastUpdate = performance.now();
  while (true) {
    const duration = await ns.gang.nextUpdate();
    const now = performance.now();
    ns.print(`${ns.gang.getGangInformation().faction} completed ${ns.tFormat(duration)} of activity in ${ns.tFormat(now-lastUpdate, true)}.`);
    ns.print(`Gang bonus time remaining: ${ns.tFormat(ns.gang.getBonusTime())}`);
    lastUpdate = now;
    // Manage the Gang
  }
}
```
Output:
```
Slum Snakes completed 5 seconds of activity in 0.220 seconds.
Gang bonus time remaining: 14 days 19 hours 59 minutes 13 seconds
Slum Snakes completed 5 seconds of activity in 0.182 seconds.
Gang bonus time remaining: 14 days 19 hours 59 minutes 8 seconds
Slum Snakes completed 5 seconds of activity in 0.197 seconds.
Gang bonus time remaining: 14 days 19 hours 59 minutes 3 seconds
```

Updated usage examples in documentation:
```js
while (true) {
  const duration = await ns.bladeburner.nextUpdate();
  ns.print(`Bladeburner Division completed ${ns.tFormat(duration)} of actions.`);
  ns.print(`Bonus time remaining: ${ns.tFormat(ns.bladeburner.getBonusTime())}`);
  // Manage the Bladeburner division
}

while (true) {
  const duration = await ns.gang.nextUpdate();
  ns.print(`Gang completed ${ns.tFormat(duration)} of activity.`);
  ns.print(`Bonus time remaining: ${ns.tFormat(ns.gang.getBonusTime())}`);
  // Manage the Gang
}

while (true) {
  const prevState = await ns.corporation.nextUpdate();
  const nextState = ns.corporation.getCorporation().state;
  ns.print(`Corporation finished with ${prevState}, next will be ${nextState}.`);
  // Manage the Corporation
}

while (true) {
  await ns.stock.nextUpdate();
  // Manage your stock portfolio
}
```

I also added the expected range of real time and mechanic time to the documentation.

jjclark1982:
Example of using `nextUpdate()` to control sleeve action timing:

```js
export async function completeSingleBladeAction(sleeveID, action) {
  let actionTimeRemaining = ns.bladeburner.getActionTime(action);
  ns.sleeve.setToBladeburnerAction(sleeveID, action);
  while (actionTimeRemaining > 0) {
    const elapsedTime = await ns.bladeburner.nextUpdate();
    actionTimeRemaining -= elapsedTime;
  }
  ns.sleeve.setToIdle(sleeveID);
}
```

PR #848, opened 2023-10-04
MISC: Update Electron to 26.2.4

jjclark1982:
The version of Electron should be updated to address these vulnerabilities:

- [CVE-2023-4863](https://nvd.nist.gov/vuln/detail/CVE-2023-4863)
- [CVE-2023-5217](https://nvd.nist.gov/vuln/detail/CVE-2023-5217)

It is unlikely for a Bitburner player to try to display .webp or .vp8 media, but it is technically possible.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -72,7 +72,7 @@
         "babel-jest": "^29.3.1",
         "babel-loader": "^9.1.2",
         "css-loader": "^6.7.3",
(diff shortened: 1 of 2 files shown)


Snarling:
This was done on #850, just upgrading electron on its own like this led to a runtime error that needed some other changes

PR #852, opened 2023-10-05
DRAFT: dependency cleanup

Caldwell-74:
not sure if needed or wanted or if i forgot anything

did audit fix - dedupe - update - dedupe
to clean up the dependencies and bring them up to date
only error was with Popper needing two more props not sure here if they could be used better i set them to undefined

tried run start:dev and build/electron on win64 and worked so far


Snarling:
Seems good, but I have no idea why Popper is now requiring two extra props to be specified (but still accepting the value `undefined` for those props). The v4 mui docs don't even show those props being used in the demos https://v4.mui.com/components/popper/ (neither does the v5 mui popper)

Snarling:
Alright the problem was due to an outdated version of @mui/material (we were on 5.10.15, upgraded to 5.14.12) and some differences between types in different versions of React. See https://github.com/mui/material-ui/issues/35287 for more info.

Caldwell-74:
was the version from mui locked and thats why it didnt update with npm update?
also not sure about an appropriate title for this so feel free :3

Snarling:
Yeah because the mui entry was with a ~ instead of ^, it wasn't willing to upgrade any higher than 5.10.x or anything higher. But it doesn't seem to cause any issues upgrading it.

PR #855, opened 2023-10-06
BUGFIX: Update NetscriptHelpers.tsx to verify servers are on network when getServer is called

### LINKED ISSUE #854: ns.singularity.destroyW0r1dD43m0n allow completetion before installing The Red Pill
when calling `ns.singularity.destroyW0r1dD43m0n` the code in 'Singularity.ts' that checks requirements for that function is below

```
 const hackingRequirements = () => {
        if (Player.skills.hacking < wd.requiredHackingSkill) return false;
        if (!wd.hasAdminRights) return false;
        return true;
      };
```
      
It is only checking to see if the player has enough hacking skill to backdoor the server and if the server has been nuked.  The Red Pill is not needed to nuke a server, so all the player needs is a high enough hacking skill to destroy the bitnode.  The admin rights check should be changed to checking if the player has The Red Pill installed.

TheAimMan:
CATEGORY: Stops World Daemon from having ports opened before install The Red Pill augmentation

CATEGORY BUGFIX
Port openers now check to make sure the server you're targeting exists on the network

# Linked issues

fixes https://github.com/bitburner-official/bitburner-src/issues/854

# Bug fix

Tested on local build.  Verified old functionality worked and that new functionally stop WD from having ports opened before installing TRP

Left image is attempt to open ports on WD.  Right image is attempt to open ports after installing TRP
image


d0sboots:
I think you applied that last commit to the wrong branch...

TheAimMan:
Oh man, I'm sorry.  This is my first time doing any of this.  I thought when I pushed my code change to my forked repository it wouldn't have effected my already opened and existing PR.  I was going to make a new one once this one went through as a learning process.

I'm attempting to revert the commit but it doesn't appear to be removing it from this PR.  I apologize for the inconvenience.  Not sure how to get the changes to Corporation/Actions.ts off this PR.

TheAimMan:
Figured it out, sorry about that.

d0sboots:
Yeah, git can be tricky like that. The way that's recommended to avoid this issue (in the future) is to do PR development on separate branches, and keep `dev` as a clean mirror of upstream.

TheAimMan:
Yeah I didn't realize branches were a thing.  I discovered them while I was trying to undo my commit.  I will now use them going forward.  Thanks for the feedback!

Snarling (changes requested):
Sorry for getting back to you late here.

Checking individually on 5 different functions to make them fail in this particular case isn't really what we should be doing here.

This should apply to every ns function that interacts with servers - before this server is on the network, it should be treated in all ways as if it does not exist. Right now, passing in any other invalid hostname causes these functions to throw an error, so just returning false is also inconsistent with other invalid hostnames.

The getServer helper function in NetscriptHelpers.ts would be a better place for this fix.

TheAimMan:
Ok, I modified the getServer function in NetscriptHelpers.tsx to check to make sure the server it's about to return in on the network.  Tested in my dev environment and received an 'Invalid Hostname' when trying to `ns.getServer("w0r1d_d43m0n") and I get the results back when TRP is installed.

I believe I've reverted the changes from NetscriptFunctions.ts... hopefully I did it properly.

Snarling:
The change looks good, but if you check the files changed tab, it looks like 3 weird files got added and picked up somehow. Any idea where these came from? Might be something I have to add to .gitignore to prevent it happening in the future.

A brief search says maybe it's related to a Visual Studio bug that may have been fixed in version 17.7.0 Preview 1 of Visual Studio, based on this thread: https://developercommunity.visualstudio.com/t/173-Preview-Indexing-process-fills-disk/10098622

TheAimMan:
Sorry about that, 100% new to this.  I've told GitHub desktop to ignore those files from now on.  Looks like they're not longer in the PR.  Hopefully everything else is squared away now.

Snarling (approved):

TheAimMan:
I'm wondering if the failure is due to the test.  
`FAIL test/jest/Netscript/RunScript.test.ts (10.187 s)`

There are no other servers on the network so my check for 'server.serversOnNetwork.length==0' would always be 0, and it would take the failure path on getServer().

I can make the test pass on my computer if I change my code from
`server.serversOnNetwork.length == 0`
to
`(server.serversOnNetwork.length == 0 && server.hostname !="home")`
which essentially proves that serversOnNetwork.length is 0 or non-existent in the current test.

d0sboots:
I think it's going to be a little more complicated than this. Personal servers and hashnet servers aren't on the network, but you definitely need to be able to access them. You should probably be able to call GetServer on DarkWeb as well, at least when you own it.

Edit: Wait, I guess personal servers and hashnet are on the network. DarkWeb is still a question, though.

TheAimMan:
It looks like Darkweb is also on network.  At least it shows in my scanning script.  
So I think we either change the test (eww) or I change the code to bypass my "on network" check for "home".

Or we go back to the original suggestion of just stopping the programs from nuking servers not on the network.
Open to suggestions

d0sboots:
In that case, your proposed change seems fine. The issue is with "stub" networks that only contain home; these are valid, so the real code should handle them.

TheAimMan:
Alright, I pushed my change with "&& server.hostname !="home""
I verified on my dev build that darkweb will fail to resolve until its purchased.  Once purchased, you can get it/connect/etc

Snarling:
Sorry for the delay. I saw there was a formatting issue so I went ahead and pushed a fix for that so I can just merge this.

For future reference, when you're making changes to a lot of files, `npm run prettier` will run through the entire project and fix formatting issues.

If you're just working in one file like on this PR, if you're using something like vscode with the Prettier extension, you should be able to do Shift + Alt + F to fix the formatting just in that one file

Also, after this PR is merged, you'll no longer be a "first time contributor" so those tests will automatically run after you commit to future PRs, instead of someone like me needing to manually tell the tests to run.

d0sboots:
> Also, after this PR is merged, you'll no longer be a "first time contributor"

Also, we can give you the "Contributor" badge on discord (if you're there and you want it)

PR #857, opened 2023-10-06
CORPORATION: fix division state display

Caldwell-74:
remove this.state from division class
changing the calls to it to be based on corporation.state.getState to make sure the correct state variable is used
fixing a bug that the division ui shows the past state instead of the next state 
and bring it in line with how the ui text is worded and what the api returns (getCorporation().state)

### DIFF
--- a/src/Corporation/Corporation.ts
+++ b/src/Corporation/Corporation.ts
@@ -115,7 +115,7 @@ export class Corporation {
         ind.resetImports(state);
       }
       for (const ind of this.divisions.values()) {
-        ind.process(marketCycles, state, this);
+        ind.process(marketCycles, this);
       }
 
       // Process cooldowns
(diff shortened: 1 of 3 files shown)


d0sboots:
LGTM

Caldwell-74:
image
did some groundwork to prep for an api break regarding getCorp().state -> geCorp().nextState
now we remember the prev state and can provide it to the player in getCorp() as .prevState 
(i didnt add it yet might cause confusion with state)
also modified the ui state display making it a progress bar surrounded by prev state and next state

edit.: if there is a better way to force Typography to not trim the leading whitespace paddings let me know

Caldwell-74:
reverted to just ui bugfix
will do another pr for the other changes (based on this)

PR #859, opened 2023-10-07
API: Change singularity.applyToCompany() to use an enum for job field

alutman:
# API: Change singularity.applyToCompany() to use an enum for job field

## Summary
`singularity.applyToCompany()` requires a string input to specify what field type the job is. The values of this field are hardcoded and not available through the API. This change adds in an enum called `JobField` to replace the string parameter, availble from the API via `ns.enums.JobField` or via a result from `getCompanyPositionInfo()`.

This allows for player scripts to automatically promote current job positions without sticking to a single job field or having to hardcode a relationship between job name and field.

## Changes
* Adds a new enum alongside `JobName` called `JobField`, exposed to `ns.enums`
* `CompanyPosition`, `CompanyPositionInfo` and `CompanyPositionCtorParams` all have a new property `field` with `JobField` type
* `getCompanyPositionMetadata()` has been updated to specify what `JobField` each job is
* `singularity.applyToCompany()`  now accepts a `JobField` enum for its `_field` parameter and uses the enum values in the switch to choose the apply method. It's intened for use with `singularity.getCompanyPositionInfo()` to extract the `JobField` from it to use here.
* Added some tests around the new enum, mainly validating that all previously valid string inputs to `applyToCompany` still work to ensure backwards compatibility.

## Notes
* `singularity.applyToCompany` now throws an exception on an invalid job field since it's using the enum validator. This is different behaviour to before as an invalid field would just log an error, returning false. Happy to revert if this is too breaking.
* `npm run doc` picked up some rename with gangs from `warfare` to `clashes`. I dropped them from the commit.
* The new property and enum both use the word `field` mainly because `singularity.applyToCompany()` calls the parameter `_field`. Easy enough to change to a different word.
* There's a whole bunch of logic around translating jobs to types (`isSoftwareJob()`, `applyForSoftwareJob()`, `hasSoftwarePositions()` etc). I've left these untouched but since `CompanyPosition` now holds the `JobField` in it, this logic could be consolidated to use the field to determine what job needs what function call.


src/NetscriptFunctions/Singularity.ts
```diff
@@ -736,48 +738,48 @@ export function NetscriptSingularity(): InternalAPI<ISingularity> {
     applyToCompany: (ctx) => (_companyName, _field) => {
       helpers.checkSingularityAccess(ctx);
       const companyName = getEnumHelper("CompanyName").nsGetMember(ctx, _companyName);
-      const field = helpers.string(ctx, "field", _field);
+      const field = getEnumHelper("JobField").nsGetMember(ctx, _field, "field");
```
alutman: `singularity.applyToCompany` now throws an exception on an invalid job field since it's using the enum validator. This is different behaviour to before as an invalid field would just log an error, returning false. Happy to revert if this is too breaking.

Snarling (commented):
Unfortunately the jobs are not actually in separate, unique job tracks. For instance the it3 job promotes into software5. So the software5 job is both an IT job and a software job (among others). This is why these weren't already handled by an enum.

If we're adding an enum, the enum's capitalization should match how that job track is displayed in the UI, not lowercased. And then the enum should be used in the UI as well.

Taking a script that was trying to apply to a field that didn't exist, and making that script throw an error instead of just returning false isn't a big deal - it already was a broken script.

The big breaking change here is that the field input on applyToCompany would now be case-sensitive, and would be needed to put in using the exact capitalization of whatever the enum uses. So that's what might take a currently-working script and make it no longer work.

alutman:
> Unfortunately the jobs are not actually in separate, unique job tracks...

I saw that but I didn't consider it an issue. it3 would report JobField.it and software5 would report JobField.software. To promote it3 you still promote under the IT track and just get moved to software. The promotions logic just looks for whatever `nextPosition` available past the one you have. The UI reflects this and you can't become a software0 if you've promoted through it0 to software5.


> ...applyToCompany would now be case-sensitive...

Ah good catch there. I've pushed a change to accommodate by lower casing the user's input before turning it to an enum. A little odd but will preserve whatever case people were using.


> ...enum's capitalization should match how that job track is displayed in the UI...

This is pretty awkward with the `toLowerCase()` in `applyToCompany()`. I've pushed a second commit that keeps the UI's casing and set the UI to use the enums. Like the case-sensitive fix, `applyToCompany()` applies casing to the user input to match the enum. It's pretty weird now but does give cased Enums, hooked to the UI and preserves functionality of any existing script. I'd imagine if the enum was used in a new API the future, you'd avoid any autocasing to force people to use the enums directly, just have `applyToCompany()` have the case-insensitive enum.

Snarling:
Going to merge this but I'm going to make some followup changes, I will tag you on those.

PR #861, opened 2023-10-08
CORPORATION: adding prevState and changed state display in the warehouse UI

Caldwell-74:
based on #857

Corp State now has getters to access nextState and prevState
prep work to include prevState in getCorporation() didnt include that yet 
because an API break is needed to change .state to .nextState to not cause confusion when .prevState is added

also changed how states are displayed in the warehouse ui to clear up the state system

image


src/Corporation/CorporationState.ts
```diff
@@ -9,20 +9,21 @@ export class CorporationState {
   // Get the name of the current state
   // NOTE: This does NOT return the number stored in the 'state' property,
   // which is just an index for the array of all possible Corporation States.
-  getState(): CorpStateName {
+  get nextState(): CorpStateName {
```
jjclark1982: maybe call these `state.nextName` and `state.prevName`?

Caldwell-74: what do you think about 
.nextStateName
.prevStateName ?

Caldwell-74: oh nvm the state is already in there yeah .nextName .prevName sound good

Caldwell-74:
there was the idea from myhf to use a different progress bar like
image
it would be a bit more complicated because the state name would need to move to "wrap" around correctly but if its wanted i can try to implement it

Snarling:
Looks good! I just had to do a quick commit 0524112 to fix a conflict from #845

Caldwell-74:
just for the future i run "npm run doc" by default
if i catch doc changes outside of the scope of my pr do you want me to make a new pr for those or just put them into that pr like with this?

Snarling:
If you already have doc changes in yours, just include all the fixes. If you don't have your own doc changes, any of the three here are acceptable:

1. Include the doc updates in your existing pr
2. Don't include them, make a separate pr
3. Don't include them, no separate pr

Ideally if any got missed, they'll be caught on someone's PR or on the next changelog update.

PR #868, opened 2023-10-17
ENUMS: Followup for #859

Snarling:
Streamlining and tweaks following up on #859 @alutman fyi

091567c: Tweak to the nsGetMember enum helper function to allow fuzzy matching, and use that instead of having a custom fuzzy matching function just for applyToCompany. This is the main part that's actually followup, the rest is just streamlining after changing nsGetMember to accept fuzzy matching.

d49ca89: Some changes to the enum helper, basically getting rid of fuzzyGetMember and rolling that functionality into getMember and nsGetMember via the new GetMemberOptions interface.

b70bad4: Just changed the capitalization from "part-time Waiter" to "Part-time Waiter" (or employee), which matches the existing capitalization in JobName. This is a small UI change as now this will be capitalized on the button for applying for job (it should be capitalized anyway). Doesn't make a difference for the API as it's already case-insensitive for this function.

16055e1: Ran prettier and fixed the capitalization in NetscriptDefinitions.d.ts which I had initially forgotten.

1414a06: Removed the singularity NetscriptFunctions test. No need for the specific tests for these two api functions as the tested stuff is either already guaranteed by ts, or is guaranteed as long as nsGetMember functions correctly.


Snarling:
Still might want to make some more tweaks. Right now the API could potentially cause confusion, as it's possible for a company to have a job available where the associated field is not something that company has available.

e.g. imagine a company with only the IT jobs. When checking the jobs this company provides, the player might find that some are in the software field (since those two fields converge at the high end of software), and then the player might try applying to the software field, which would fail even though the company has some jobs which report that they are in that field.

PR #870, opened 2023-10-17
CODE: Enable es2023 features

### LINKED ISSUE #853: missing signature help / parameter hints for new Electron methods
with the update of Electron came new JS methods like 
array.toSorted()
array.toReversed()
currently these just show "any" in the popup
image

Snarling:
Fixes #853 

For some reason ES2023 was not accepted as a valid target, even though documentation online says it should be. But selecting ESNext works. Trying to get es2023 to work is why I updated some packages, I don't think that was necessary otherwise.

Not a huge fan of keeping jsconfig.json in the distributed folder with all the electron js files, although package.json is already included so it's not a new issue with this file.

### DIFF
--- a/electron/jsconfig.json
+++ b/electron/jsconfig.json
@@ -1,0 +1,3 @@
+{
+  "compilerOptions": { "target": "ESNext" }
+}
--- a/package-lock.json
+++ b/package-lock.json
@@ -74,7 +74,7 @@
         "css-loader": "^6.7.3",
         "electron": "^26.2.4",
         "electron-packager": "^17.1.1",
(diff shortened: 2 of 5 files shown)


d0sboots:
Looks like typescript got more script with the upgrade?

Snarling:
Yeah for some reason it thought a numerical comparison in Validator.ts was being done before we verified that the values being compared were numeric.

But we did verify the values were numeric before the comparison, ts just wasn't smart enough to realize it. I rewrote it in a way that is still functionally identical but now ts doesn't complain about it.

PR #874, opened 2023-10-18
INFILTRATION: Change hermes AUG to instead show the next characters

alutman:
This is a _suggested_ change to how the Shadows of Anarchy Hermes augmentation (for the Cheat Code infiltration) works.

I've found the Hermes aug to be the worst of all the SoA augs and barely helpful. Most SoA augs are at least desirable. This changes the aug to instead show what characters are going to come next. The opposite character functionality is removed.

Through play-testing this, I found you needed to also show the previous characters for the augmentation to not simply be more confusing. It's difficulty with the augmentation is now about on par with the Ares/Slash minigame; you'll probably succeed but if you slack off a bit you'll lose.

The flavour text could possibly be updated to reflect the new behavour but if you mentally squint, it can still makes sense.

### Two Arrows
Here's a gif with two arrows before and two arrows after the "real" arrow. This is what is implemented in the PR.
arrows_new

### DIFF
--- a/src/Augmentation/Augmentations.ts
+++ b/src/Augmentation/Augmentations.ts
@@ -1781,7 +1781,7 @@ export const Augmentations: Record<AugmentationName, Augmentation> = (() => {
       repCost: 1e4,
       moneyCost: 1e6,
       info: "Penta-dynamo-neurovascular-valve inserted in the carpal ligament, enhances dexterity.",
-      stats: "This augmentation makes the Cheat Code minigame easier by allowing the opposite character.",
+      stats: "This augmentation makes the Cheat Code minigame easier by showing what character will come next.",
       isSpecial: true,
       factions: [FactionName.ShadowsOfAnarchy],
     },
--- a/src/Infiltration/ui/CheatCodeGame.tsx
+++ b/src/Infiltration/ui/CheatCodeGame.tsx
@@ -1,16 +1,9 @@
 import { Paper, Typography } from "@mui/material";
 import React, { useState } from "react";
 import { AugmentationName } from "@enums";
+import { Settings } from "../../Settings/Settings";
 import { Player } from "@player";
-import {
-  downArrowSymbol,
-  getArrow,
-  getInverseArrow,
-  leftArrowSymbol,
-  random,
-  rightArrowSymbol,
-  upArrowSymbol,
-} from "../utils";
+import { downArrowSymbol, getArrow, leftArrowSymbol, random, rightArrowSymbol, upArrowSymbol } from "../utils";
 import { interpolate } from "./Difficulty";
 import { GameTimer } from "./GameTimer";
 import { IMinigameProps } from "./IMinigameProps";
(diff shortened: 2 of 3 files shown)


d0sboots:
Visually, I'm finding this harder to track than the non-upgraded version. Maybe it would be different if I were playing with it, I'm not sure.

One definite improvement would be to change the non-current arrows to a different color, for better contrast.

alutman:
> Visually, I'm finding this harder to track than the non-upgraded version. Maybe it would be different if I were playing with it, I'm not sure.
> 
> One definite improvement would be to change the non-current arrows to a different color, for better contrast.

It uses the `disabled` color of the theme which SoA augs utilize for highlighting new information. I didn't want to break convention here. `secondary` can work too but it's very theme dependent. Some themes have` secondary` as almost the same color as `primary`. 

I've instead used dashed arrows to show the non-current arrows which should avoid any theming issues.

Snarling:
I also am finding this more visually difficult to follow.

alutman:
I've widened the gap between the arrows to help it a bit. If you can, give it a go in-game, might be a bit easier to follow when you're controlling.

Here's a quick diff to use make infils always arrows
```diff
diff --git a/src/Infiltration/ui/Game.tsx b/src/Infiltration/ui/Game.tsx
index 0d028a9e5..02af3c51a 100644
--- a/src/Infiltration/ui/Game.tsx
+++ b/src/Infiltration/ui/Game.tsx
@@ -46,7 +46,7 @@ export function Game(props: GameProps): React.ReactElement {
   const [results, setResults] = useState("");
   const [gameIds, setGameIds] = useState({
     lastGames: [-1, -1],
-    id: Math.floor(Math.random() * minigames.length),
+    id: 4,//Math.floor(Math.random() * minigames.length),
   });

   const setupNextGame = useCallback(() => {
@@ -56,7 +56,7 @@ export function Game(props: GameProps): React.ReactElement {
       while (ids.includes(id)) {
         id = Math.floor(Math.random() * minigames.length);
       }
-      return id;
+      return 4;//id;
     };

     setGameIds({
```

d0sboots:
> If you can, give it a go in-game,

Since Snarling especially is pressed for time these days, it would be very helpful if you could post another moving gif of the adjusted behavior.

At the same time, I do understand that stuff like this is different when you're controlling it.

alutman:
> ...another moving gif of the adjusted behavior.

Already updated the primary post with the adjusted behavior, I should've mentioned that.

d0sboots:
Ah yes, that is at minimum no worse than without the aug. I can't speak to how it feels without trying it, but the arrows are definitely more distinguishable now.

Snarling:
Sorry for not following up on this one for so long. I'm still not happy with this for a few reasons:
1. Still visually jarring to have the code teleport as the player makes entries. The player can look ahead at future arrows, but then they teleport when the player enters the current arrow. I think an approach where the code is shown in a static location on the screen works better.
2. Using the highlighted color as "non-active" arrows is opposite of how most of the infiltration games are coded - usually the highlighted entry is the entry of interest (even though the color is the "disabled" theme color). I think opacity ends up working better and more consistently across different themes as an active/inactive indicator.

I used this PR as a basis for a different rework at #994 , @alutman can you take a look and see if you like those changes?

alutman:
>. ...
> I used this PR as a basis for a different rework at #994 , @alutman can you take a look and see if you like those changes?

Ah yours is much better. I remember not wanting to edit the un-augmented gameplay, but it's much better when both are updated. Thanks for having a pass at this! 

Closed due to https://github.com/bitburner-official/bitburner-src/pull/994.

PR #878, opened 2023-10-19
UI: grammar/typo fixes

Squirlll:
Fixed typos and grammar mistakes in:
-Augmentation descriptions
-The dialogue box that appears after you purchase an augmentation
-The Offline Scripts and Bonus Time tab in Advanced Mechanics documentation


hydroflame:
FWIW this user edited all of this directly from the website so it's worth downloading and making sure it builds before merging.

d0sboots:
Well, it passes typescript at least. I don't think a local build will reveal anything different than the autobuild, and the fact that it's definitely only strings changes makes it a lot safer.

PR #881, opened 2023-10-20
BUGFIX: Update SleeveCrimeWork.ts to work through all cyclesWorked if its greater than cyclesNeeded

### LINKED ISSUE #876: Sleeves performing certain task in overclock duration result in cyclesWorked overflow
### Steps to reproduce
1. Manage a sleeve to make it have overclock cycles stored.
2. Assign certain task, namely Shoplift under Commit Crime to the sleeve.
3. Check the Sleeve tab HUD OR trace the property cyclesWorked of that sleeve.

### Result
A number which goes far beyond cyclesNeeded is revealed in HUD:
image

in object properties:
image

### Game version
Bitburner v2.5.0 (b87b8b4be)

TheAimMan:
CATEGORY: BUGFIX

Work through all stored cycles possible.
1. This will actually use all of the stored cycles for Shoplift
2. This will fix the UI progress indication noted in the submitted issue

# Linked issues

fixes https://github.com/bitburner-official/bitburner-src/issues/876

# Bug fix

Reproduced issue Dev environment
Issue is that the cycles needed for Shoplift are less than the cycles worked in an Overclocked Sleeve
So when the process hits
`this.cyclesWorked -= this.cyclesNeeded()`

There are still cyclesWorked left over that are actually greater than the number of cycles that would be need to complete the task again.  The progress bar is just a display of cycles worked.  Since every overclock cycle adds more cyclesWorked than removed by the Shoplift task, it grows forever.

Throwing in a loop to work through all remaining cycles until it's actually less than the needed cycles allows Sleeves to achieve their full potential for overclocked shoplifting and it fixes the UI issue.

### DIFF
--- a/src/PersonObjects/Sleeve/Work/SleeveCrimeWork.ts
+++ b/src/PersonObjects/Sleeve/Work/SleeveCrimeWork.ts
@@ -38,7 +38,7 @@ export class SleeveCrimeWork extends SleeveWorkClass {
     this.cyclesWorked += cycles;
     if (this.cyclesWorked < this.cyclesNeeded()) return;
 
-     while (this.cyclesWorked > this.cyclesNeeded()) {
+    while (this.cyclesWorked > this.cyclesNeeded()) {
       const crime = this.getCrime();
       const gains = this.getExp(sleeve);
       const success = Math.random() < crime.successRate(sleeve);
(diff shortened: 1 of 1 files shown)


TheAimMan:
TIL that you have to run `npm run format` multiple times :(
I still had my CMD window open from last night when I ran it before pushing.  I used the same window to run the same command again...

image
I then ran it a 3rd time to make sure it would pass :)

d0sboots:
That's very strange, it should only need to run once. Is it possible that some git or VS action undid the changes in-between?

TheAimMan:
Nah, I stopped using VS.  My original While loop was janky.  This only changed a single space.  I can only surmise I must have accidentally put an extra space and saved it before committing.  Checks out now so not going to sweat it but I will double run before every commit now :)

PR #882, opened 2023-10-20
MISC: Nerf Noodle Bar

jjclark1982:
Recent Corporation anti-fraud accounting did not account for money laundering through Noodle Bar.

It was possible to exploit previous fraud strategies by eating noodles 10 times per second.

### DIFF
--- a/src/Locations/ui/SpecialLocation.tsx
+++ b/src/Locations/ui/SpecialLocation.tsx
@@ -116,7 +116,7 @@ export function SpecialLocation(props: SpecialLocationProps): React.ReactElement
       }
 
       if (Player.corporation) {
-        Player.corporation.funds += Player.corporation.revenue * 0.01;
+        Player.corporation.addNonIncomeFunds(Player.corporation.revenue * 0.01);
       }
     }
 


d0sboots:
One thing you could do is change Corporation.funds to a getter and have no setter, so that it can't accidentally be adjusted wrongly.

Snarling:
@jjclark1982 This is marked as draft, intentional?

Caldwell-74:
since n00dles is meant to be "exploity" why not instead of killing the exploit turn down the numbers to make it close to nothing just like the others do?

Snarling:
This PR is actually not nerfing the overall effect of the noodles, which is to give your corporation more money. It's just making those added funds not look like legitimate income, so they don't have an outsized impact on valuation.

Basically, for cycles where you're eating noodles, you currently get 85000x the impact on valuation for the noodles you ate during that cycle, e.g. if you gained $1m from noodles, you would essentially have an $85b higher valuation as long as you kept up that rate of noodle eating. With this PR you keep the added valuation from just having more money (so $1m gained from noodles would cause a lasting $1m valuation increase), but the noodles don't look like "income" so the noodle money doesn't get the temporary 85000x multiplier on valuation during cycles where you are eating lots of noodles.

Caldwell-74:
yeah iam aware but it would still fit into the "theme" of n00dles to be the one legitimate "fraud" mechanic
it just needs a ton more 0 to be a "meme" amount added to valuation imho

jjclark1982:
> @jjclark1982 This is marked as draft, intentional?

Yeah, it's still not clear what the intended balance should be. Using Noodle Bar to activate fraud-like accounting rewards a deep understanding of the source code, so maybe this should just be left as-is.

You make a good point about changing funds to a getter to prevent this kind of bug from popping up in the future. Maybe that is the root issue here.

PR #884, opened 2023-10-23
BUGFIX: Check Corp Research pre-reqs in research api call

TheAimMan:
BUGFIX: Current code does not check to make sure the previous research requirements are met.  It allows any research to be gain as long as the player doesn't currently have it, and has enough research points.

# Bug fix

Reproduce original issue.
Coded and tested fix.  Verified pre-req research was required.  If not, an error is now thrown

image

### DIFF
--- a/src/Corporation/Actions.ts
+++ b/src/Corporation/Actions.ts
@@ -463,12 +463,14 @@ export function Research(researchingDivision: Division, researchName: CorpResear
   const researchTree = IndustryResearchTrees[researchingDivision.type];
   if (researchTree === undefined) throw new Error(`No research tree for industry '${researchingDivision.type}'`);
   const research = ResearchMap[researchName];
-  const child = researchTree.findNode(researchName);
+  const researchNode = researchTree.findNode(researchName);
+  const researchPreReq = researchNode?.parent?.researchName;
   //Check to see if the research request has any pre-reqs that need to be researched first.
-  if (child !== null && child.parent !== null && child.parent.researchName !== null) {
-    const parent = child.parent.researchName;
-    if (!researchingDivision.researched.has(parent)) {
-      throw new Error(`You need ${parent} before you can research ${research.name}`);
+  if (researchPreReq) {
+    if (!researchingDivision.researched?.has(researchPreReq)) {
+      throw new Error(
+        `Division ${researchingDivision.name} requires ${researchPreReq} before researching ${research.name}`,
+      );
     }
   }
   if (researchingDivision.researched.has(researchName)) return;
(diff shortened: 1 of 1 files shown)


src/Corporation/Actions.ts
```diff
@@ -463,7 +463,14 @@ export function Research(researchingDivision: Division, researchName: CorpResear
   const researchTree = IndustryResearchTrees[researchingDivision.type];
   if (researchTree === undefined) throw new Error(`No research tree for industry '${researchingDivision.type}'`);
   const research = ResearchMap[researchName];
-
+  const child = researchTree.findNode(researchName);
+  //Check to see if the research request has any pre-reqs that need to be researched first.
+  if (child !== null && child.parent !== null && child.parent.researchName !== null) {
+    const parent = child.parent.researchName;
```
Snarling: This can be simplified with "optional chaining", and the name of the `parent` variable can be improved as well and be defined before the conditional:
```ts
  const parentResearchName = researchNode?.parent?.researchName;
  if (parentResearchName) {
```

src/Corporation/Actions.ts
```diff
@@ -463,7 +463,14 @@ export function Research(researchingDivision: Division, researchName: CorpResear
   const researchTree = IndustryResearchTrees[researchingDivision.type];
   if (researchTree === undefined) throw new Error(`No research tree for industry '${researchingDivision.type}'`);
   const research = ResearchMap[researchName];
-
+  const child = researchTree.findNode(researchName);
```
Snarling: `child` is not a great variable name here, it's a research tree node associated with the "research" variable, so I would name it `researchNode`.

src/Corporation/Actions.ts
```diff
@@ -463,7 +463,14 @@ export function Research(researchingDivision: Division, researchName: CorpResear
...
+  //Check to see if the research request has any pre-reqs that need to be researched first.
+  if (child !== null && child.parent !== null && child.parent.researchName !== null) {
+    const parent = child.parent.researchName;
+    if (!researchingDivision.researched.has(parent)) {
+      throw new Error(`You need ${parent} before you can research ${research.name}`);
```
Snarling: It's not "you" or the player who needs the parent research, it's the researchingDivision who needs it.

Some of the existing error messages also make this mistake.

Snarling (changes requested):
Functionality looks good, just suggesting some changes to variable names (parent and child) and an error message that can be improved.

Snarling (approved):
Looks good! Will merge this in the next batch of merges.

PR #887, opened 2023-10-23
CORPORATION: move product productionCost into cityData

Caldwell-74:
> When products are sold, their production cost is updated every time (https://github.com/bitburner-official/bitburner-src/blob/dev/src/Corporation/Division.ts#L826), it's the cost of all input materials multiplied by 5. Since material price changes, product.productionCost also changes. This updated value is what's shown in the corp UI and returned by ns.corporation.getProduct().productionCost. However, it is not city-specific, but material prices are different in different cities. So as a result, that number is only accurate for the city that was processed last.
https://discord.com/channels/415207508303544321/415213413745164318/1165999223989342280

Moved Product productionCost into cityData to save the individual productionCost
this might make it easier to "fraud" by moving items from a city with low production Cost into a city with high production cost to sell there
but i think good information are more important than preventing that

### DIFF
--- a/src/Corporation/Corporation.ts
+++ b/src/Corporation/Corporation.ts
@@ -232,7 +232,7 @@ export class Corporation {
           assets += mat.stored * mat.averagePrice;
         }
         for (const prod of ind.products.values()) {
-          assets += prod.cityData[warehouse.city].stored * prod.productionCost;
+          assets += prod.cityData[warehouse.city].stored * prod.cityData[warehouse.city].productionCost;
         }
       }
     });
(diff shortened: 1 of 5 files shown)


Snarling:
> this might make it easier to "fraud" by moving items from a city with low production Cost into a city with high production cost to sell there, but i think good information are more important than preventing that

Agreed. Also it's already how selling materials works anyway.

Caldwell-74:
yeah it essentially just removes the need to sum the cost themself but anyone that tries to attempt such a complex exporting structure should be capable to do that summing easily

Snarling (approved):
Looks good to me, will merge in the next batch of merges.

One area I'd probably make some changes is in ProductElem.tsx. There are a whole lot of instances of `product.cityData[city]` in the file, I would probably shove that double lookup into its own variable since the component is just for the product in one city.

Caldwell-74:
something like that?

PR #891, opened 2023-10-24
BUGFIX: Add effect definition to fragmentObject

### LINKED ISSUE #463: Stanek's gift: FragmentType is missing from the API?
I'm ... assuming it isn't a game design decision that `FragmentType`, from https://github.com/bitburner-official/bitburner-src/blob/19e8946d9aed90a26be35b8f8999238cdd70d5e5/src/CotMG/FragmentType.ts at the moment, isn't included in the API?

I certainly didn't notice any lore about Stanek's Gift involving poorly documented numeric identifiers, which is what `Fragment`'s `type` field contains at the moment.

(If it *is* intended, such lore should be added :-)

TheAimMan:
For continuity I reused the definition's that already exist and are displayed in the UI.  There no Booster type definition, and it looks like the UI was hard-coded to pull the power value of the booster fragment, which I can't do on fragment creation.  Open to just changing the "1.1x" to "boosts" in case something is added in the future that actually changes a booster fragments power.

fixes https://github.com/bitburner-official/bitburner-src/issues/463

# Documentation

I ran `npm run doc` and it doesn't seem to have changed the md files, even though I've added a 6th attribute to the fragment object

# Bug fix

Effect will now be listed
Made sure I could still place blocks and charge.

image


TheAimMan:
I had started with just changing type to the actual name, and not the array number like it is now, but I didn't want to break any existing scripts that used that.  Maybe add typeName as a property so certain shards can be filtered out from the result?

Snarling:
> I ran npm run doc and it doesn't seem to have changed the md files, even though I've added a 6th attribute to the fragment object

You would have to also add the property to the type definition in NetscriptDefinitions.d.ts, see 4b64a2c for future reference (just adding the missing property to ns definitions and regenerating docs)

Issue #896, opened 2023-10-25
Bladeburner Populations should have a base growth number along with percentage growth


TheAimMan:
Current random events in BladeBurner only grow a cities population by a percentage of its current population.  This creates the possibility for a player to completely eradicate a city, causing the growth to never increase again.

While in fits the nature of the game and the objective behind the entire Blade Burner lore, its not an enjoyable game mechanic to kill all the synths and be stuck with no way to progress.

Sphyxis:
Proposed fix in my PR #897

PR #897, opened 2023-10-26
SLEEVE API UI UPDATE && FIX ISSUE #896 BB NEEDS BASE GROWTH - Added setToBladeburnerSingleAction and the ability to set sleeves to IDLE from UI and a base growth to BB

Sphyxis:
This updates sleeves with the following abilities:
In the sleeve UI, if you select ------ and set task, your sleeve will go into IDLE mode

New API function call:  ns.sleeve.setToBladeburnerSingleAction
-Identical to it's counterpart, except that your sleeve will stop what they are doing automatically after their action has completed

Tested by running sleeves side by side, half with runonce commands.  All tested ok.

-update: ------ now renamed to Idle
-update: Added a fix for issue #896 
-Tested by killing all the population in every city.  Waited, and my base population was added at the next event.

### DIFF
--- a/src/Bladeburner/Bladeburner.tsx
+++ b/src/Bladeburner/Bladeburner.tsx
@@ -894,6 +894,9 @@ export class Bladeburner {
     const count = Math.round(sourceCity.pop * percentage);
     sourceCity.pop -= count;
     destCity.pop += count;
+    if (destCity.pop < BladeburnerConstants.PopGrowthCeiling) {
+      destCity.pop += BladeburnerConstants.BasePopGrowth;
+    }
   }
 
   triggerPotentialMigration(sourceCityName: CityName, chance: number): void {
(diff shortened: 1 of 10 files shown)


Sphyxis:
I may be missing an opportunity to collect bonus sleeve time.  Right before it stops the action, there is a taskWorked - taskNeeded.  Perhaps taskWorked (after the - ) could be added to the sleeve action bank?

Caldwell-74:
if "-----" is idle why not rename it to idle ?

Sphyxis:
> if "-----" is idle why not rename it to idle ?

Done and done.

Caldwell-74:
this should be 2 pr's for sure

Sphyxis:
> this should be 2 pr's for sure

Yah, my first time with GH.  Let me redo it into two

PR #898, opened 2023-10-27
FORMULAS: Removing the possible exception when gymGains/universityGains is called

cigarmemr:
For locations without a corresponding server, calling `ns.formulas.work.gymGains()` or `ns.formulas.work.universityGains()` in NetScript will raise the "reading properties of null" exception.

As GetServer() in line 87 returns `null` for these locations, trying to access its `backdoorInstalled` causes the issue above. Change the operator to `?.` and deal with possible `undefined` should fix it.

This makes the two forementioned NetScript functions treat all locations uniformly regardless of they have a backdoorable server or not.


d0sboots:
The `??` isn't needed... null is falsey, so it will work fine in the ternary.

cigarmemr:
> The `??` isn't needed... null is falsey, so it will work fine in the ternary.

I guess you were meant to say `undefined` is falsey but yes you're right.

PR #902, opened 2023-10-29
SLEEVE: BladeburnerSingleAction mechanic

Sphyxis:
Ran format, lint.

Introduced the ability for Sleeves to run a single Bladeburner action and then stop.

### DIFF
--- a/src/Netscript/RamCostGenerator.ts
+++ b/src/Netscript/RamCostGenerator.ts
@@ -319,6 +319,7 @@ const sleeve = {
   getSleevePurchasableAugs: RamCostConstants.SleeveBase,
   purchaseSleeveAug: RamCostConstants.SleeveBase,
   setToBladeburnerAction: RamCostConstants.SleeveBase,
+  setToBladeburnerSingleAction: RamCostConstants.SleeveBase,
   getSleeveAugmentationPrice: RamCostConstants.SleeveBase,
   getSleeveAugmentationRepReq: RamCostConstants.SleeveBase,
 } as const;
--- a/src/NetscriptFunctions/Sleeve.ts
+++ b/src/NetscriptFunctions/Sleeve.ts
@@ -261,6 +261,36 @@ export function NetscriptSleeve(): InternalAPI<NetscriptSleeve> {
 
       return Player.sleeves[sleeveNumber].bladeburner(action, contract);
     },
+    setToBladeburnerSingleAction: (ctx) => (_sleeveNumber, _action, _contract?) => {
+      const sleeveNumber = helpers.number(ctx, "sleeveNumber", _sleeveNumber);
+      const action = helpers.string(ctx, "action", _action);
+      let contract: string;
+      if (typeof _contract === "undefined") {
+        contract = "------";
+      } else {
+        contract = helpers.string(ctx, "contract", _contract);
+      }
+      checkSleeveAPIAccess(ctx);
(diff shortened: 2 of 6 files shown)


Snarling:
I think adding an entirely new action type doesn't make sense here, especially with that action type being inaccessible via the ui and only available via a new api function. But the functionality you're looking for as a player makes sense.

What are your thoughts on #916 as a different way to provide players the ability to perform single actions?

Sphyxis:
> I think adding an entirely new action type doesn't make sense here, especially with that action type being inaccessible via the ui and only available via a new api function. But the functionality you're looking for as a player makes sense.
> 
> What are your thoughts on #916 as a different way to provide players the ability to perform single actions?

That unfortunately won't work for what I was looking for.  The ability to set the action to stop after 1 rotation without needing to await anything is what I was looking for.  I could see if I can add a checkbox or something similar to the UI to allow a single action to be set if that will work better.

Snarling:
You don't need to await anything, that's just what I did in the example code. You can use .then instead, and continue doing other things in your current script. The only limitation is that your script does still need to be alive when the promise gets resolved, otherwise it will not be able to set the sleeve back to idle (or to whatever other task you want to do).


Here's an example of doing a single task:

https://github.com/bitburner-official/bitburner-src/assets/84951833/89653171-0780-4743-b0e2-339f2ae3cec7

```js
/** @param {NS} ns */
  ns.tail();
  ns.disableLog("asleep")
  let task = ns.sleeve.getTask(0);
  if (task.type !== "BLADEBURNER") throw new Error("Task somehow was not correct type.");
  let stillWaiting = true;
  task.nextCompletion.then(() => {
    ns.sleeve.setToIdle(0);
    ns.print("Set sleeve 0 to idle");
    stillWaiting = false;
  });

  while (stillWaiting) {
    // This loop is just to keep this script alive, you can do whatever you want in the script during this time.
    ns.print("Doing other stuff while waiting on task completion.");
    await ns.asleep(1000);
  }
  ns.print("Single task was finished, doing something else.")
}
```

Sphyxis:
> You don't need to await anything, that's just what I did in the example code. You can use .then instead, and continue doing other things in your current script. The only limitation is that your script does still need to be alive when the promise gets resolved, otherwise it will not be able to set the sleeve back to idle (or to whatever other task you want to do).
> 
> Here's an example of doing a single task:
> 
>  2023-11-05.01-21-55.mp4 
> ```js
> /** @param {NS} ns */
> export async function main(ns) {
>   ns.tail();
>   ns.disableLog("asleep")
>   ns.sleeve.setToBladeburnerAction(0, "Take on contracts", "Tracking");
>   let task = ns.sleeve.getTask(0);
>   if (task.type !== "BLADEBURNER") throw new Error("Task somehow was not correct type.");
>   let stillWaiting = true;
>   task.nextCompletion.then(() => {
>     ns.sleeve.setToIdle(0);
>     ns.print("Set sleeve 0 to idle");
>     stillWaiting = false;
>   });
> 
>   while (stillWaiting) {
>     // This loop is just to keep this script alive, you can do whatever you want in the script during this time.
>     ns.print("Doing other stuff while waiting on task completion.");
>     await ns.asleep(1000);
>   }
>   ns.print("Single task was finished, doing something else.")
> }
> ```

It will take some thought but I think this will work.

Snarling:
Merging alternative #916 instead

PR #904, opened 2023-10-29
DOC: Create factions-list documentation page

jjclark1982:
Convert old "List of Factions and their Requirements" documentation page to markdown format and place it in the "advanced" section

v2.1.0 readthedocs:
https://bitburner.readthedocs.io/en/latest/basicgameplay/factions.html

v2.3.1 source: 
https://github.com/bitburner-official/bitburner-src/blob/v2.3.1/doc/source/basicgameplay/factions.rst

this branch:
https://github.com/jjclark1982/bitburner-src/blob/faction-list/src/Documentation/doc/advanced/faction_list.md


Snarling:
Sorry for the late merge, I was going back and forth on whether this should be included in official docs. For now I think it's fine, although it is something that requires active maintenance to keep it in sync with the actual ingame requirements as new factions are added or as old ones possibly change.

PR #905, opened 2023-10-31
SINGULARITY: Added getOwnedNFGAugmentations

Sphyxis:
Add the ability to get installed/owned/queued NFG augmentation totals.
New command:  ns.singularity.getOwnedNFGAugmentations(?true/false)

Imported my game.  Tested outputs, showed correct for Total Installed and Queued installs.  Tested OK after install as well.


Snarling:
Sorry for being super duper late on providing feedback here.

NFG is currently the only augmentation where level matters, but that is not guaranteed to always be the case and I think it's out of place to have a function just for one specific augmentation.

Something like `getAugmentationLevel(augName, includeQueued=false)` would make more sense as an API function. It would return 0 if the augmentation is not owned, and the actual augmentation level otherwise. It's still useful for other augs, and would also be extensible if other levelable augs are introduced in the future.

Sphyxis:
Once I’m able to I will change this up.

PR #906, opened 2023-10-31
BLADEBURNER: Updated descriptions

Sphyxis:
Updated descriptions of startAction and getActionEstimatedSuccessChance

### DIFF
--- a/src/ScriptEditor/NetscriptDefinitions.d.ts
+++ b/src/ScriptEditor/NetscriptDefinitions.d.ts
@@ -2885,6 +2885,7 @@ export interface Bladeburner {
    *
    * Attempts to start the specified Bladeburner action.
    * Returns true if the action was started successfully, and false otherwise.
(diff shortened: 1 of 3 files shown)


src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -2885,6 +2885,7 @@ export interface Bladeburner {
    *
    * Attempts to start the specified Bladeburner action.
    * Returns true if the action was started successfully, and false otherwise.
+   * Example: ns.bladeburner.startAction("Contracts", "Tracking")
```
Snarling: Examples should use the `@example` jsdoc syntax

Sphyxis: Is this correct now?

Snarling: Looks good, thanks. I will try to get you feedback on the other PRs tonight

PR #908, opened 2023-10-31
DEVMENU: Updated Sleeves

Sphyxis:
Added the ability to add/remove sleeves from the dev menu.

Tested by adding/removing sleeves.  Confirmed that they came and went and that the sleeves sidebar was removed when you had no sleeves left.

### DIFF
--- a/src/DevMenu.tsx
+++ b/src/DevMenu.tsx
@@ -27,19 +27,23 @@ import { AchievementsDev } from "./DevMenu/ui/AchievementsDev";
 import { EntropyDev } from "./DevMenu/ui/EntropyDev";
 
 import { Exploit } from "./Exploits/Exploit";
+import { useRerender } from "./ui/React/hooks";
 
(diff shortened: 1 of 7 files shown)


Snarling:
Having this within the sleeves menu makes it possible to lock yourself out of the sleeves menu while you're within it. Also because the sleeves are just added directly, any excess sleeves added will be removed the next time a soft reset is performed, which is probably not desirable.

Snarling:
Since this PR is from October (my bad for getting to it so late), I went ahead and made some changes myself since I didn't hear back.

Refactored and moved this to the SF page, which now also immediately adds/removes sleeves as appropriate when SF10 level is changed.

image

Sphyxis:
Thanks so much!  I was starting to wrap up on Sunday when I got the message.  Been thinking about it and was going to change it tonight when I had some time.  I’ll be changing the -all pr in about 5-6 hours.

PR #911, opened 2023-11-02
STOCKMARKET: randomize initial ticksUntilCycle value

Caldwell-74:
initialising the Stockmarket with a set ticksUntilCycle value encourages the player to count the game ticks to easily find the flip timing of the stockmarket
this trivializes the main challenge of the stock mechanic

this can be prevented by initialising it with a random value 

and doc changes i missed on my last pr :/


d0sboots:
Could you maybe reword the description? I'm not following it at all. (Maybe it just needs some punctuation.)

jjclark1982:
Hmm, does `ns.stock.nextUpdate()` also make it too easy to identify flip events?

d0sboots:
> Hmm, does `ns.stock.nextUpdate()` also make it too easy to identify flip events?

I don't think so? It makes it a bit easier, since it's easier to react to a tick (that was the whole point), but you still have to count cycles (or now, look for state changes) to actually see the flip.

Krate8:
In my humble opinion. I don't think identifying flips is "the main challenge of the stock mechanic". Simply developing your own forecasting strategy is the main challenge. That's more that enough, honestly looking for flips is not necessary; just gravy for those looking to add a little extra to their algorithm.  

But, taking this idea further, should stock flips always be 75 ticks? Should that be so easy and consistent a number? maybe it should be random between 70 and 80? maybe each stock should have a small chance to flip every tick instead of all the stocks having a significant chance to flip every so often.

Maybe each stock could have a "tick offset" number between 0 and 75 so the stocks don't all flip at once. 

I'm not saying any of that should happen, but I think there's definitely a question of whether counting ticks for flips is a good mechanic or not, how easy should it be, 

Just some thoughts.

Caldwell-74:
the usual steps for solving stock market is post 4S forecast (+ maybe flip) -> pre 4S forecast -> pre 4S flip
sure most people dont got over pre 4S forecast and that is most of the time enough to make some money over long enough time but its not reliable and comes often with huge setbacks (investing into a stock right before it flips, not detecting this for X0 cycles)
flip timing is what gives confidence in the data accumulated, what makes profit estimation possible and the script overall reliable

i dont want to make it impossible or super hard to detect the flip and thats what it would be if each stock had its own flip timer or when the flip length is random

i want to prevent imho toxic strategies that rely on monitoring every single tick since install or similiar

Snarling:
@Caldwell-74 I know it's really late feedback, but this seems kind of pointless - even with this change, once you successfully detect one flip, you can just use that as your timing baseline and know exactly when all the other flip opportunities will be. Doesn't seem any less toxic IMO. Am I missing something?

However, there's really no issue with this so I'll just merge it.

d0sboots:
I understand and agree with Caldwell's motivation here: The objective is not to make flip detection a substantially more interesting problem, but rather to eliminate a strategy that *requires* you to start a script in the first tick of the BN. If you miss that window, you have to soft-reset to get it back.

Basically, it's a strat that seems easier at the beginning but has some far-reaching hidden downsides, that's why Caldwell labeled it as toxic.

PR #913, opened 2023-11-04
IMPROVEMENT: Apply core bonus for share and stanek

TheAimMan:
Currently cores don't add a bonus for running share or charging stanek fragments.  I'm not sure why not, and it doesn't seem game breaking (share core bonus only gives ~1% increase) so I decided to add it in.

Tested locally, verified share/charge work, and used dev tools to make sure the thread count was increased.

### DIFF
--- a/src/NetscriptFunctions.ts
+++ b/src/NetscriptFunctions.ts
@@ -35,6 +35,7 @@ import {
   numCycleForGrowthCorrected,
   processSingleServerGrowth,
   safelyCreateUniqueServer,
+  getCoreBonus,
 } from "./Server/ServerHelpers";
 import {
   getPurchasedServerUpgradeCost,
@@ -397,7 +398,8 @@ export const ns: InternalAPI<NSFull> = {
           helpers.log(ctx, () => "Server is null, did it die?");
           return Promise.resolve(0);
         }
-        const coreBonus = 1 + (host.cpuCores - 1) / 16;
+        const cores = helpers.getServer(ctx, ctx.workerScript.hostname).cpuCores;
+        const coreBonus = getCoreBonus(cores);
         const weakenAmt = CONSTANTS.ServerWeakenAmount * threads * coreBonus;
         server.weaken(weakenAmt);
         ctx.workerScript.scriptRef.recordWeaken(server.hostname, threads);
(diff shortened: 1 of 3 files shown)


Snarling:
Instead of being duplicated in so many functions, the core bonus should be moved to a helper function.

The ? and conditional for cpuCores is unnecessary. helpers.getServer always returns a BaseServer, and server.cpuCores will always be a number.

TheAimMan:
I can look into creating a helper function.

The ? is there because the compiler complained that it could be null.

image

d0sboots:
That's because you're using the wrong `getServer` (we have too many). The one in helpers, which you should be using, does validation and never returns null.

TheAimMan:
Well that's not confusing at all! :)
Yes, if I change it to use the helpers.getServer and not the getServer function that exists within NetScriptsfunctions.ts the compiler doesn't complain.
I'll keep that in mind when writing the coreBonus helper function.

Snarling:
So there's a reason for having two different getServer functions.

The stuff in `helpers` is mostly for validating unknown input from the player via the API, and always returns the type it's supposed to return (throwing an error if the type is wrong). For the netscript API, this is usually the desired result: on invalid input, throw an error message saying why it was invalid input, but on valid input if we make it past the validation function without throwing, then the type is validated.

helpers.positiveInteger -> throws if the number isn't a positive integer, always returns a validated positive integer if it doesn't throw
helpers.getServer -> takes in a basic string, throws if it's not the name of an actual server on the network, always returns the actual validated server object if it doesn't throw

When we're dealing with game internals, throwing is usually not desirable. The "GetServer" function takes in a string and returns a validated server object, but if the hostname provided didn't belong to a server, it returns null to indicate that. Instead of the failure path being determined by throwing a standard error, each usage has specific handling for the null case to define how what that particular code should do when the provided hostname was not associated with a server object.

d0sboots:
As usual, you're correct with the longer explanation. And this isn't technically a user param, since it's the hostname the script is running on. But if server lookup fails for that, IMO throwing the same sort of exception makes sense.

TheAimMan:
Created a function called getCoreBonus in ServerHelpers (since it had to do with the servers cores) and replaced the hardcoded calculation in 5 places with the new function.

Tested with debugger to verify I was getting the correct core bonus value back for share, stanek, and weaken and grow.

PR #914, opened 2023-11-04
TERMINAL: Updated Alias/Unalias

Sphyxis:
Reserved the -all from Alias as a remove command.
Added -all to unalias to remove all aliases set.
Updated help files and descriptions when alias/unalias are entered with nothing else

Tested by trying to set an alias to -all, showed error message that it's reserved.
Tested with unalias.  Shows all aliases removed.  Only works if it's the first command, so unalias -g *x* will still work for specifics global aliases.

### DIFF
--- a/src/Terminal/commands/alias.ts
+++ b/src/Terminal/commands/alias.ts
@@ -6,8 +6,8 @@ export function alias(args: (string | number | boolean)[]): void {
     printAliases();
     return;
   }
-  if (args[0] === "-all") {
-    Terminal.print(`-all is reserved for removal`);
+  if (args[0] === "--all") {
+    Terminal.print(`--all is reserved for removal`);
     return;
   }
   if (args.length === 1) {
--- a/src/Terminal/commands/alias.ts
+++ b/src/Terminal/commands/alias.ts
@@ -7,7 +7,7 @@ export function alias(args: (string | number | boolean)[]): void {
     return;
   }
   if (args[0] === "--all") {
-    Terminal.print(`--all is reserved for removal`);
+    Terminal.error(`--all is reserved for removal`);
     return;
   }
   if (args.length === 1) {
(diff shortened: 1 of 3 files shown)


jjclark1982:
Bitburner commands generally use `libarg` style CLI flags such as `--all` (with two `-`s as it is a multi-letter flag name), or the special argument `*`.

Sphyxis:
I based it off of what's being used for the -g command to set global aliases.

Sphyxis:
> Bitburner commands generally use `libarg` style CLI flags such as `--all` (with two `-`s as it is a multi-letter flag name), or the special argument `*`.

I based it off of the -g command that's already in alias/unalias.  Does this mean the -g global option should be changed as well or is it alright?

d0sboots:
Short options (one letter) have a single dash, long options have two.

Sphyxis:
> Short options (one letter) have a single dash, long options have two.

Thank you for the clarification.  Redid it to be --all instead.

src/Terminal/commands/alias.ts
```diff
@@ -6,6 +6,10 @@ export function alias(args: (string | number | boolean)[]): void {
     printAliases();
     return;
   }
+  if (args[0] === "--all") {
+    Terminal.print(`--all is reserved for removal`);
```
Snarling: This should be displayed w/ Terminal.error, since alias is failing to do what the player asked it to do and the message is telling the player why it failed.

Sphyxis:
Thank you for the note on the double check.  I see it’s already been added – is there anything else you need for it?

src/Terminal/commands/alias.ts
Snarling:
```suggestion
    Terminal.error(`--all is reserved for removal`);
```

PR #916, opened 2023-11-05
SLEEVE: Add a nextCompletion promise to SleeveBladeburnerWork

Snarling:
Allows waiting for the next completion of a bladeburner action, in case the player wants to track individual action completions as a new place to make a decision during a script.

Tested with the following example script to show how a player might use this to trigger code execution after each individual action completion (see main function, the rest is just some React stuff).

```js
/** @param {number | undefined} ms */
export function useRerender(ms) {
  const [__, setRerender] = React.useState(false);
  const rerender = React.useCallback(() => setRerender((old) => !old), []);
  React.useEffect(() => {
    if (!ms) return;
    const intervalID = setInterval(rerender, ms);
    return () => clearInterval(intervalID);
  }, [rerender, ms]);
  return rerender;
}

/** @param {{ms?: number, children: ReactNode}} */
export function RerenderingComponent({ ms, children }) {
  useRerender(ms ?? 200);
  return children;
}

/** @param {NS} ns */
  ns.tail();
  let task = ns.sleeve.getTask(0);
  while (task?.type === "BLADEBURNER") {
    const rawArray = ["Waiting for next completion..."];
    ns.printRaw(React.createElement(RerenderingComponent, { ms: 10 }, rawArray));
    await task.nextCompletion;
    rawArray.push("Done");
    task = ns.sleeve.getTask(0);
  }
  ns.print("No longer performing bladeburner work.")
}
```


jjclark1982:
FYI, here's my script used to detect action completion with `nextUpdate`. Unlike this `task.nextCompletion` interface it requires knowledge of when the task started.

```js
export async function completeSingleBladeAction(sleeveID, action) {
  let actionTimeRemaining = ns.bladeburner.getActionTime(action);
  ns.sleeve.setToBladeburnerAction(sleeveID, action);
  while (actionTimeRemaining > 0) {
    const elapsedTime = await ns.bladeburner.nextUpdate();
    actionTimeRemaining -= elapsedTime;
  }
  ns.sleeve.setToIdle(sleeveID);
}
```

Snarling:
@jjclark1982 While I was comparing your method, I found that there was a logic error in the SleeveBladeburnerWork processing, where if you have exactly the right number of cycles worked, it would still wait a whole extra second to process the work. Fixed in cab49e6

But as far as different methods of achieving the same sort of functionality, the bladeburner.nextUpdate method has a lot of pitfalls.
Contracts take a different amount of time based on stats/augs and bladeburner.getActionTime assumes the player is the one to perform the action. Also sleeve work tasks do not necessarily process on the same schedule as the bladeburner mechanic, so even with player+sleeve stats the same, only sometimes would you get accurate results with that method. Bonus time also throws a wrench in things, since bladeburner bonus time and sleeve bonus time are completely separate.

Another method would be to just check how many cycles are needed, and then wait that number of game cycles using a regular timer.

All three methods result in pretty close results, but importantly the bladeburner.nextUpdate method sometimes results in thinking you're done with the task before you actually are (even after the timing fix above). 

e.g. here's an instance where these sleeve works are being processed about 400ms away from when bladeburner is processed, so switching based on the bladeburner method would result in switching too early / not completing the task:

image

Other times, the updates for sleeve work and bladeburner updates do sync up (again, if player and sleeve stats are kept close enough in sync):

Here's the test script I used for comparing the different methods:

```js
/** @param {NS} ns */
export let main = ns => new Promise(async r => {
  ns.atExit(r);

  async function testSleeve(i, contract) {
    ns.sleeve.setToIdle(i);
    while (true) {
      ns.sleeve.setToBladeburnerAction(i, "Take on contracts", contract);
      const start = Date.now();
      let promises = [
        // Method 1: bladeburner.nextUpdate
        (async () => {
          let actionTimeRemaining = ns.bladeburner.getActionTime("Contracts", contract);
          while (actionTimeRemaining > 0) {
            const elapsedTime = await ns.bladeburner.nextUpdate();
            actionTimeRemaining -= elapsedTime;
          }
          return Date.now() - start;
        })(),
        // Method 2: getTask(i).nextCompletion
        (async () => {
          await ns.sleeve.getTask(i).nextCompletion;
          return Date.now() - start;
        })(),
        // Method 3: getTask(i).cyclesNeeded
        (async () => {
          const timeLeft = ns.sleeve.getTask(i).cyclesNeeded * 200;
          await ns.asleep(timeLeft);
          return Date.now() - start;
        })(),
      ];
      await Promise.all(promises);
      ns.tprint(`Times for sleeve ${i} performing ${contract} were:
bladeburner.nextUpdate: ${await promises[0]}
sleeve.getTask(i).nextCompletion: ${await promises[1]}
sleeve.getTask(i).cyclesNeeded: ${await promises[2]}\n\n`)
      ns.sleeve.setToIdle(i);
    }
  }

  // Can't do the same contract with more than one sleeve
  const contracts = ["Tracking", "Bounty Hunter", "Retirement"];
  // Launch test for all sleeves
  for (let i = 0; i < 3; i++) testSleeve(i, contracts[i]);
});
```

PR #922, opened 2023-11-08
SLEEVE: Add crime and BB chance functions for sleeves

Kelenius:
My personal pet peeve.

изображение
изображение


d0sboots:
I was going to say you could take "Sleeve" out of the name to make them shorter, since all these are in the sleeve namespace to begin with, but I see that all the other functions are named similarly so it's better to be consistent.

Caldwell-74:
sleeves can already use the formulas crimeSuccessChance function
not sure if a separate one is needed

Kelenius:
Interesting, I didn't know that. I still think it's a good addition, though, I don't think you should need formulas to get a number you can see in the UI.

Caldwell-74:
not saying it should but then it would make sense to move the player crime chance out of formulas aswell or not?

Kelenius:
[`ns.singularity.getCrimeChance()`](https://github.com/bitburner-official/bitburner-src/blob/dev/markdown/bitburner.singularity.getcrimechance.md) already exists.

cigarmemr:
closes #1009

Snarling:
Sorry I never left feedback on this PR from a while ago.

The function names are ridiculously long. I don't think functions specific to sleeve really make sense here when this is not really sleeve-specific functionality. To fit into the existing API, this functionality makes sense at the fomulas level.

Alternately, if a sleeve is already performing a work task, then maybe the success chance can be added to the API return of that work task. This would mean the method for checking the success chance would be to initiate the work task and then ask the API for details about the current work task (which is the same way this information is gained through the UI - only displayed when the work task is active)

Kelenius:
> The function names are ridiculously long.

I can shorten them a bit but I have to keep them unique, or else it confuses the RAM counter.

> I don't think functions specific to sleeve really make sense here when this is not really sleeve-specific functionality.

It's a cross between sleeve and BB. I could move them to BB space, but I don't see how that's any better.

> To fit into the existing API, this functionality makes sense at the fomulas level.

The problem with putting this into formulas is that it depends on a whole lot of BB stats (city population, chaos, skills, off the top of my head), which means that either the function would need to accept a massive number of parameters, or I need to create a bladeburner object that contains all of that info, a function for bladeburner that returns it, and then make the formula read that object.

I wanted to put it into formulas at first, but I ran into the above issue and decided that this is much easier.

> Alternately, if a sleeve is already performing a work task, then maybe the success chance can be added to the API return of that work task. This would mean the method for checking the success chance would be to initiate the work task and then ask the API for details about the current work task (which is the same way this information is gained through the UI - only displayed when the work task is active)

This is a very bad solution for the problem I'm trying to solve, which is determining when it's worth sending a sleeve to do BB stuff. It would involve setting the sleeve to the task (which interrupts the current action), then maybe cancelling it.

d0sboots:
I just saw another similar issue, but for gangs. I agree, formulas is probably the right place for this, but there is groundwork to be done because the API is so sparse (since we've been shoving everything directly into BB).

Zelow79:
Sleeve crime chance imo should be in formulas for sure. I think there is already a formula for crime chance and if it doesn't except a sleeve object already, it should be just made to accept one. The other BB stuff I love here, I agree dropping the getSleeve part should be removed it is already in the sleeve namespace, but maybe it can be further shortened to something like `BladeburnerEstChance()` I would imagine this is fine with the RAM counter.

Kelenius:
I think #1428 does this better.

PR #929, opened 2023-11-14
STANEK: Remove inter-fragment borders in Stanek's gift

Kelenius:
Old: 
image
New:
image-1


Snarling:
Looks way better, especially with those same-colored adjacent fragments. Thanks!

PR #930, opened 2023-11-14
SLEEVE: Add task counter to Crimes and BladeBurner

TheAimMan:
Added a counter on sleeve tasks.  This is for when a sleeve is in Overclock mode.  The progress bar doesn't reflect the actual completions, and on some tasks will just appear to be stuck, even though the Sleeve is doing work.

I just added an attribute to the SleeveWork object, and increment it by one when a task is finished.

SleeveTasksCounter

### DIFF
--- a/src/PersonObjects/Sleeve/Work/SleeveBladeburnerWork.ts
+++ b/src/PersonObjects/Sleeve/Work/SleeveBladeburnerWork.ts
@@ -17,6 +17,7 @@ export const isSleeveBladeburnerWork = (w: SleeveWorkClass | null): w is SleeveB
 
 export class SleeveBladeburnerWork extends SleeveWorkClass {
   type: SleeveWorkType.BLADEBURNER = SleeveWorkType.BLADEBURNER;
+  tasksCompleted = 0;
   cyclesWorked = 0;
   actionType: "General" | "Contracts";
   actionName: string;
@@ -69,6 +70,7 @@ export class SleeveBladeburnerWork extends SleeveWorkClass {
       if (this.actionType === "Contracts") {
         applySleeveGains(sleeve, scaleWorkStats(retValue, sleeve.shockBonus(), false));
       }
+      this.tasksCompleted++;
       this.cyclesWorked -= this.cyclesNeeded(sleeve);
       // Resolve and reset nextCompletion promise
       const resolver = this.signalCompletion;
@@ -82,6 +84,7 @@ export class SleeveBladeburnerWork extends SleeveWorkClass {
       type: SleeveWorkType.BLADEBURNER as "BLADEBURNER",
       actionType: this.actionType,
       actionName: this.actionName,
+      tasksCompleted: this.tasksCompleted,
       cyclesWorked: this.cyclesWorked,
       cyclesNeeded: this.cyclesNeeded(sleeve),
       nextCompletion: this.nextCompletion,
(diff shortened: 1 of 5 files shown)


src/PersonObjects/Sleeve/ui/SleeveElem.tsx
```diff
@@ -31,15 +31,17 @@ function getWorkDescription(sleeve: Sleeve, progress: number): string {
...
+          work.tasksCompleted,
+        )}\n \n` + `Progress: ${formatPercent(progress)}`
       );
     case SleeveWorkType.CRIME: {
       const crime = work.getCrime();
       return (
         `This sleeve is currently attempting ${crime.workName} (Success Rate: ${formatPercent(
           crime.successRate(sleeve),
-        )}).\n\n` + `Progress: ${formatPercent(progress)}`
+        )}).\n\nTasks Completed: ${formatNumber(work.tasksCompleted)}
```
Snarling: Given that this is an integer, it is probably not useful to include decimal places. There is a `formatInt` function that provides sane defaults for displaying integers (no decimals until suffix is shown, suffix begins at 1000, so the display would show `999`, and then the next value would be `1.000k`)

src/PersonObjects/Sleeve/Work/SleeveBladeburnerWork.ts
```diff
@@ -82,6 +84,7 @@ export class SleeveBladeburnerWork extends SleeveWorkClass {
       type: SleeveWorkType.BLADEBURNER as "BLADEBURNER",
       actionType: this.actionType,
       actionName: this.actionName,
+      tasksCompleted: this.tasksCompleted,
```
Snarling, line 87: Since this property is being added to the API copy for Sleeve Blade and Crime work, it should also be added in ScriptEditor/NetscriptDefinitions.d.ts definitions for SleeveCrimeTask and SleeveBladeburnerTask

Snarling (changes requested):

src/PersonObjects/Sleeve/Work/SleeveBladeburnerWork.ts
TheAimMan: pretty sure I just submitted what you asked for

Snarling: Yep! Will still need to re-run `npm run doc` so the generated docs get updated too

TheAimMan: I ran it, but it doesn't appear to have done anything to any files.  GitHub desktop isn't showing any changes.

image

Snarling (approved):

Snarling:
Oh shoot, I forgot about the `npm run doc` that was still needed and I merged this. Did a one-off commit a3c26ca afterwards to get those doc changes in.

src/PersonObjects/Sleeve/Work/SleeveBladeburnerWork.ts
d0sboots: It's like that. You won't see any result from the command-like output or from `git diff`, because it adds the changes as staged. You'll see them if you do `git status` or other commands like `git commit `.

TheAimMan: thanks.  I wanted to make sure I didn't have something mis-configured.  The messages about the 'fail to fine node.js install directory' made me think something wasn't working properly.

PR #933, opened 2023-11-17
UI: Fixed padding issue with server page slector and made script page selector static

TheAimMan:
The server pagination selector is set to only be displayed when there are > 10 servers running scripts.  Whenever it was displayed, it would push the whole table of scripts down.  When it would hide, the table would move up.  I added padding to stop its appearance and disappearance from causing the whole table to shift.

The script pagination selector causes a much greater jump in the UI anytime its displayed/hide which gets worse as certain script execution times get lower.

JumpingScripts

I changed the behavior to always show the page selector as I couldn't think of a better solution, aside from maybe making it an option to hide or always show.

### DIFF
--- a/src/ui/ActiveScripts/ServerAccordionContent.tsx
+++ b/src/ui/ActiveScripts/ServerAccordionContent.tsx
@@ -25,7 +25,7 @@ export function ServerAccordionContent(props: IProps): React.ReactElement {
 
   return (
     <>
-      {props.workerScripts.length > 10 ? (
+      {props.workerScripts.length > 0 ? (
         <TablePagination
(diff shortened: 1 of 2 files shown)


Snarling:
It's not a new problem from this PR, but I'm not a huge fan of the weird vertical spacing that's created on the left side by introducing the pagination at the top. Especially when there are only a couple scripts running that extra space on the left looks really out of place.

Might try some quick style modifications to get the pagination to go on the same line as the server name if it can fit, only dropping to the next line if there's not enough room.

Snarling:
Replaced by #1083

PR #943, opened 2023-11-25
REMOTE: Added possibility to change target IP of remote API

Specker:
#  Description 
Added possibility to change target of the Remote API to allow development of scripts on other device than where game is running.

It is a niche situation but with Steam Deck gaining popularity and being a good device for idlers, I think giving this possibility for the more savvy would be great.

## Possible usage
- Game running on different device(Steam Deck, laptop, vm)
- File synchronization utility([Viteburner](https://github.com/Tanimodori/viteburner-template),[bitburner-filesync](https://github.com/bitburner-official/bitburner-filesync)) on main device(main pc, work pc)

# Features
- possibility to change target IP of the Remote API

image
image

### DIFF
--- a/src/GameOptions/ui/RemoteAPIPage.tsx
+++ b/src/GameOptions/ui/RemoteAPIPage.tsx
@@ -37,7 +37,7 @@ export const RemoteAPIPage = (): React.ReactElement => {
       <Tooltip
         title={
           <Typography>
-            This address is used to connect to a Remote API port, please ensure that it matches with your Remote API
+            This address is used to connect to a Remote API, please ensure that it matches with your Remote API
             address. Default localhost.
           </Typography>
         }
@@ -50,13 +50,14 @@ export const RemoteAPIPage = (): React.ReactElement => {
           value={remoteFileApiAddress}
           onChange={handleRemoteFileApiAddressChange}
           placeholder="localhost"
+          size={"medium"}
         />
       </Tooltip>
       <br />
       <Tooltip
         title={
           <Typography>
-            This port number is used to connect to a Remote API port, please ensure that it matches with your Remote API
+            This port number is used to connect to a Remote API, please ensure that it matches with your Remote API
             server port. Set to 0 to disable the feature.
           </Typography>
         }
@@ -73,6 +74,7 @@ export const RemoteAPIPage = (): React.ReactElement => {
           value={remoteFileApiPort}
           onChange={handleRemoteFileApiPortChange}
           placeholder="12525"
+          size={"medium"}
         />
       </Tooltip>
       <br />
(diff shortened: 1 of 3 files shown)


Hoekstraa:
Note: This wasn't a feature up until now due to potential security risks. If this gets accepted without a security assessment, I won't take responsibility for anything that happens due to this change.

src/GameOptions/ui/RemoteAPIPage.tsx
```diff
@@ -28,6 +34,25 @@ export const RemoteAPIPage = (): React.ReactElement => {
         </Link>
       </Typography>
       <ConnectionBauble isConnected={isRemoteFileApiConnectionLive} />
+      <Tooltip
+        title={
+          <Typography>
+            This address is used to connect to a Remote API port, please ensure that it matches with your Remote API
```
d0sboots: This was copied from port, should be updated for address.

src/RemoteFileAPI/RemoteFileAPI.ts
```diff
@@ -5,8 +5,9 @@ let server: Remote;
 
 export function newRemoteFileApiConnection(): void {
   if (server) server.stopConnection();
-  if (Settings.RemoteFileApiPort === 0) return;
-  server = new Remote("localhost", Settings.RemoteFileApiPort);
+  if (Settings.RemoteFileApiAddress === "") Settings.RemoteFileApiAddress = "localhost";
```
d0sboots: Don't change the value in settings, that's (potentially) permanent. Set a local instead.

d0sboots (commented):
Although it's wise to be cautious, I don't see any security issues here. It's an outbound connection; any script can make those at any time through various built-in JS APIs. And the configuration is player controlled; a sufficiently exploity script could change the address, but it's possible that a sufficiently exploity script could have done that *anyway*.

src/RemoteFileAPI/RemoteFileAPI.ts
Specker: I was thinking about this line as more of return to default value in settings and connect. 
Same way the port field defaults to 0 when you empty the field. But string field would return to localhost when you remove 'l'.

Specker: Switched logic to display "localhost" placeholder instead switching it in Settings. 
Connection with empty string to localhost is now done with Ternary.

d0sboots (commented):
Looks good, but @Snarling has the final say.

Snarling:
Sorry for the late merge @Specker , I was conflicted when I first saw this. In the end I agree with @d0sboots analysis above.

PR #944, opened 2023-11-26
UI: 'Disable Text Effects' setting applies to "corrupted text"

### LINKED ISSUE #892: Accessibility: Support disabling animation of "corrupted text" effect
The game "Disable text effects" setting and the system "prefers reduced motion" setting don't do anything to make `<CorruptedText>` less distracting.

draughtyan:
Adds "spoiler" boolean property to CorruptableText; text will be corrupted and unreadable if true, else text will have zero effects
All changes only apply while 'Disable Text Effects' is toggled ON in settings, otherwise everything appears as usual.

Closes #892

Old:
old_corruptabletext
New:
new_corruptabletext

### DIFF
--- a/src/Locations/ui/GenericLocation.tsx
+++ b/src/Locations/ui/GenericLocation.tsx
@@ -21,8 +21,6 @@ import { CasinoLocation } from "./CasinoLocation";
 import { Location } from "../Location";
 import { LocationType } from "@enums";
 
-import { Settings } from "../../Settings/Settings";
-
 import { isBackdoorInstalled } from "../../Server/ServerHelpers";
 import { GetServer } from "../../Server/AllServers";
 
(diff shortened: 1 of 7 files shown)


src/Locations/ui/GenericLocation.tsx
```diff
@@ -97,10 +95,10 @@ export function GenericLocation({ loc }: IProps): React.ReactElement {
     <>
       <Button onClick={() => Router.toPage(Page.City)}>Return to World</Button>
       <Typography variant="h4" sx={{ mt: 1 }}>
-        {backdoorInstalled && !Settings.DisableTextEffects ? (
+        {backdoorInstalled ? (
           <Tooltip title={`Backdoor installed on ${loc.name}.`}>
```
jjclark1982: Only loosely related, but this should be:

```suggestion
          <Tooltip title={`Backdoor installed on ${serverMeta.hostname}.`}>
```

PR #949, opened 2023-11-29
CORPORATION: Use accounting methods for all funds transactions

jjclark1982:
This PR adds a "source" argument to all Corporation funds transactions. It is similar to the Player MoneySourceTracker but without persistence. The source is used to categorize each transaction as "operating expenses" (included in earnings projections) or "capital expenditure" (not included in earnings projections).

Fixes an exploit reported by @Azirale — 11/24/2023 3:35 AM
> it seems if you start up a new corp, build out a division to get a tick of profit, and just go ahead and sell the division for ~$27B, on the next cycle you'll get ~$10T from private investors. This appears to be enough money to immediately jump into robotics with a full build out and a 150-person office to make the first product.

cc @yichizhng

### DIFF
--- a/src/Corporation/Corporation.ts
+++ b/src/Corporation/Corporation.ts
@@ -78,7 +78,7 @@ export class Corporation {
 
   gainFunds(amt: number, source: FundsSource): void {
     if (!isFinite(amt)) {
-      console.error("Trying to add invalid amount of funds. Please report this bug.");
+      console.error("Trying to add invalid amount of funds. Please report to game developer.");
       return;
     }
     if ((FundsSourceCapEx as readonly FundsSource[]).includes(source)) {
@@ -139,9 +139,6 @@ export class Corporation {
           this.revenue = this.revenue + ind.lastCycleRevenue;
           this.expenses = this.expenses + ind.lastCycleExpenses;
         });
-        const cycleRevenue = this.revenue * (marketCycles * corpConstants.secondsPerMarketCycle);
-        const cycleExpenses = this.expenses * (marketCycles * corpConstants.secondsPerMarketCycle);
-        const cycleProfit = cycleRevenue - cycleExpenses;
         if (isNaN(this.funds) || this.funds === Infinity || this.funds === -Infinity) {
           dialogBoxCreate(
             "There was an error calculating your Corporations funds and they got reset to 0. " +
@@ -150,8 +147,11 @@ export class Corporation {
           );
           this.funds = 150e9;
         }
-        this.loseFunds(cycleExpenses, "operating expenses");
+        const cycleRevenue = this.revenue * (marketCycles * corpConstants.secondsPerMarketCycle);
+        const cycleExpenses = this.expenses * (marketCycles * corpConstants.secondsPerMarketCycle);
+        const cycleProfit = cycleRevenue - cycleExpenses;
         this.gainFunds(cycleRevenue, "operating revenue");
+        this.loseFunds(cycleExpenses, "operating expenses");
         if (this.dividendRate > 0 && cycleProfit > 0) {
           // Validate input again, just to be safe
           if (isNaN(this.dividendRate) || this.dividendRate < 0 || this.dividendRate > corpConstants.dividendMaxRate) {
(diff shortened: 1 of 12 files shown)


yichizhng:
rip noodles fraud T_T

petition to categorize noodles as opex

src/Corporation/Corporation.ts
```diff
@@ -75,23 +76,19 @@ export class Corporation {
...
   }
 
-  addFunds(amt: number): void {
+  gainFunds(amt: number, source: FundsSource): void {
     if (!isFinite(amt)) {
-      console.error("Trying to add invalid amount of funds. Report to a developer.");
+      console.error("Trying to add invalid amount of funds. Please report to game developer.");
       return;
     }
+    if ((FundsSourceCapEx as readonly FundsSource[]).includes(source)) {
```
Snarling: Array.includes relies on iterating the entire array (e.g. if the expense is opex, it has to iterate the entire capex array to figure that out). Also the type definitions are requiring a type assertion here to allow array.includes to be used.

### Suggestion 1: Enums

One way to address this would be by changing the FundsSources to be enums (add to the Corporation/Enums.ts file), and then use the enum helper to check if the source is in the requested enum. e.g. this line would be

```ts
if (getEnumHelper("FundsSourceCapEx").isMember(source)) {
```

Which instead of iterating an array, becomes a Set.has operation which should be more efficient for checking if an item exists within a list. Moving to an enum would also mean changing each usage of the gainFunds / loseFunds so that the source needs to be referenced using Enum.key, instead of as a raw string at each location. Which would also help in cases where it's not immediately clear to everyone whether that particular item fits in as opex or capex. e.g.

```ts
corp.addFunds(1e9, "private equity");

// would turn into:

corp.addFunds(1e9, FundsSourceCapEx.privateEquity);
```

Which would make it clear to anyone reading it that it's considered to be capexm, without needing to refer to the objects to determine which sources are under which categories. Although moving to an enum helper would also mean some additional upfront memory usage for the two enums and their helpers.

----

### Suggestion 2: CapEx Set

Another option would be just making the capex runtime object into a set directly, and then just having a type defined that includes all possible sources, and only using some of them for the capex set. Opex doesn't even need a runtime object since it is unused:

e.g.

```js
// FundsSource.ts, alternate suggestion without changing to enums
// CapEx sources at the top
export type FundsSource = "product development" |
  "division" |
  "office" |
  "warehouse" |
  "upgrades" |
  "bribery" |
  "public equity" |
  "private equity" |
  "hacknet" |
  "force majeure" |
  // OpEx
  "operating expenses" |
  "operating revenue" |
  "dividends" |
  "tea" |
  "parties" |
  "advert" |
  "materials" |
  "glitch in reality";

// Set of all CapEx sources, for checking at runtime.
// "CapEx" sources only affect valuation through a change in funds, and are not seen as profit/loss for earnings projections.
export const CapExSources = new Set<FundsSource>([
  "product development",
  "division",
  "office",
  "warehouse",
  "upgrades",
  "bribery",
  "public equity",
  "private equity",
  "hacknet",
  "force majeure",
])
``` 

Do either of those sound like good changes to you?

Not super thrilled with the names CapEx and OpEx, since the "Ex" in for expenditures, but it's not just for tracking expenditures but revenue as well. But I don't have a better suggestion so I'm fine with the naming how it is, just a very minor gripe I figured I would mention as well.

jjclark1982: I agree that CapEx/OpEx is not a very clear phrasing. The main distinction is one-time/repeatable or long-term/short-term, so I can change it to a name that better reflects that.

I'd like to keep the style of `gainFunds()` similar to `Player.gainMoney()`, using a string instead of an Enum. But I also don't like repeating the values in multiple places which could get out of sync.

What do you think of this style with a Set built from a const array? I can't figure out a way to define a type as a key of a const Set without also including the array.

src/Corporation/data/FundsSource.ts
```diff
@@ -12,8 +13,9 @@ export const FundsSourceCapEx = [
   "force majeure",
 ] as const;
 
-// "Operating Expense" categories which should be included in earnings projections for valuation
-export const FundsSourceOpEx = [
+// Funds transactions which should be included in earnings projections for valuation.
+// This includes all automatic or indefinetly-repeatable income and operating expenses.
+const FundsSourceShortTerm = [
```
Snarling: Yeah I think this is good as far as not defining the magic strings in multiple places.

Super minor, but since FundsSourceShortTerm isn't used for any runtime purpose (only for extracting type information), FundsSourceShortTerm can be a `declare const` to make it not create a runtime object, and ShortTermFundsSources can just be removed since it is unused. Just having the long term set is enough to allow us to discern whether a FundsSource is long term or short term.

PR #950, opened 2023-11-30
FUNCTION: Allow user to disable the disable log message

TheAimMan:
calling the disableLog function traces out in the log that the function was disabled.  When calling this function from within a function, will still get displayed.  Added an option to suppress this log message.

Example of the previous behavior for disabling 'sleep' on top, and the new parameter being set to 'true' on the bottom

image

### DIFF
--- a/src/NetscriptFunctions.ts
+++ b/src/NetscriptFunctions.ts
@@ -505,20 +505,25 @@ export const ns: InternalAPI<NSFull> = {
   clearLog: (ctx) => () => {
     ctx.workerScript.scriptRef.clearLog();
   },
-  disableLog: (ctx) => (_fn) => {
-    const fn = helpers.string(ctx, "fn", _fn);
-    if (fn === "ALL") {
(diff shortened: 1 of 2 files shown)


Snarling:
1. It seems like a usage error to constantly be re-disabling a type of log over and over.
2. Is there a reason not to just do ns.disableLog("disableLog") if that's what you want to do?
image

TheAimMan:
Well I want back to test what I thought I had already tested and I can't reproduce my original issue.  Go figure.
I had disabled getServerMaxRam outside of the main scripts 'while' loop but I kept seeing traces for getServerMaxRam within the logging.  That made me believe that it was because the function was the one that needed the tracing disabled.  

While this allows for more granular approach (maybe you want to see that you disabled other traces, but not all traces) its not necessary. Apologize for the inconvenience.

Snarling:
No need to apologize! Are we good to close this, or is there something that still looks like it needs a change?

TheAimMan:
We're good to close. Thanks!

PR #952, opened 2023-11-30
UPDATE: 2.5.1 Release

Snarling:
This is a main version release for 2.5.1.

Aside from the changes that have existed on the dev branch, this also removes the newly-added getFactionInviteRequirements function, which will likely be modified in the 2.5.2 cycle (see details on 01f1af0)

@jjclark1982 FYI, sorry for not providing that feedback earlier


jjclark1982:
Thanks for the feedback. I reckon `getFactionInviteRequirements()` would work better with a structured format similar to `getCompanyPositionInfo()`.

PR #927, opened 2023-11-12
COMPANY: Clarify job listings and requirements

jjclark1982:
### Original Problem

The "Apply to Job" buttons are an ongoing source of confusion. The displayed job fields do not match the actual jobs the company offers or the actual promotion tracks that they follow. The job you see the requirements for is sometimes different from the one you effectively apply for. There is no indication when you have reached the end of a job track.

Fixes #20 (except for removing the "Quit" button)

Fixes  #708

Fixes #856

Fixes this problem reported on discord:

> derpygamer2142:
> not sure if this is a bug but there's a little empty box here if you get to the max level
> apply-to-job

### PR Description

This PR adds tooltips to "Apply for Job" buttons showing a checklist of requirements.

Instead of showing "Apply for [field] Job" buttons for each of several hard-coded fields, this shows the specific jobs available at the company.

### Balance changes:
- The secret rep bonus to faction requirements for backdooring a company server now also applies to job requirements.
- Part-time jobs have removed the penalty for doing something else simultaneously.
- Previously hidden jobs like Network Engineer and Business Consultant are now available through the UI.

### API changes:
- `ns.singularity.applyToCompany()` now returns the name of the job received (or null if none). Compatible with the previous boolean return value.

### Screenshots of new UIs:
No requirements
Screenshot 2023-11-29 at 12 31 20 PM

Overqualified
Screenshot 2023-11-29 at 12 28 34 PM

Qualified
Screenshot 2023-11-29 at 12 28 41 PM

Underqualified
Screenshot 2023-11-29 at 12 28 50 PM

Listing at a small company
Screenshot 2023-11-29 at 7 31 29 PM

Listing at a medium-size company
Screenshot 2023-11-29 at 12 32 43 PM

Listing at a large company
Screenshot 2023-11-29 at 12 29 12 PM

https://github.com/bitburner-official/bitburner-src/assets/208776/2e1533d5-4ae8-4252-85c3-6ad3d57c5611

https://github.com/bitburner-official/bitburner-src/assets/208776/cbfccdcf-80c7-49c0-969d-7695943badfa

PR #953, opened 2023-11-30
API: Programming-friendly interface to getFactionInviteRequirements

jjclark1982:
This PR defines a programming-friendly interface to `ns.singularity.getFactionInviteRequirements()`

A requirement is an object with a declared JSON-like interface, including recursive references to sub-requirements. For example:

```js
ns.singularity.getFactionInviteRequirements("The Syndicate")
[
  { "type": "someCondition", "conditions": [
      { "type": "city", "city": "Aevum" },
      { "type": "city", "city": "Sector-12" }
    ]
  },
  { "type": "not", "condition": {
      "type": "employedBy", "company": "Central Intelligence Agency"
    }
  },
  { "type": "not", "condition": {
      "type": "employedBy", "company": "National Security Agency"
    }
  },
  { "type": "money", "money": 10000000 },
  { "type": "skills", "skills": { "hacking": 200 } },
  { "type": "skills", "skills": { "strength": 200, "defense": 200, "dexterity": 200, "agility": 200 } },
  { "type": "karma", "karma": -90 }
]
```
- For fields with numerical value > 0, the player must have at least this value.
- For fields with numerical value <= 0, the player must have at most this value.
- For "not", the sub-condition must be failed instead of passed.
- For "someCondition", at least one sub-condition must be passed.


Snarling:
Referencing game types in nsdefs will be a reference error when you `npm run doc`. nsdefs need to use types that exist in the API, not internal game types.

**Edit: see number 4 at the bottom, the API seems usable without totally refactoring it like I was suggesting in most of this comment. If you're good with the API as written, I'm good with it too. Leaving the rest of the comment intact but you can ignore most of it.**

The way this is set up still makes it very difficult to actually use in a player script, particularly one written in ts. Either that or I am just having a hard time seeing it. If you think this is a good format for the API, can you show me an example script of how the player might use the API?

Personally I would do it similar to SleeveTask, here's an example of how I would expect interaction with the API to go as a player, with a few of the different requirement types handled, if this was handled in a similar way to SleeveTask:

```js
// assume ns is in scope for all this, as if it's all inside main
// waitForSkills is an existing player function that levels up the needed skills and returns a promise
//    that resolves when the player has achieved the required skills.
// Other similarly named functions for individual requirements work the same as above, working to meet
//    the requirement and resolving a promise when it is met.


// Returns a promise that resolves when one requirement is met
/** @param {Requirement} requirement */
function waitForRequirement(requirement){
  switch (requirement.type){
    case "money": // This hardcoded string has autocompletion for all valid options
      // because we've checked that requirement.type is "money", ts knows that
      //   requirement is a MoneyRequirement, so it knows requirement.money is a number.
      return waitForMoney(requirement.money);
    case "skills":
      return waitForSkills(requirement.skills);
    case "companyRep":
      return waitForCompanyRep(requirement.company, requirement.rep);
    case "someCondition":
      return Promise.any(requirement.conditions.map(waitForRequirement));
    // etc to handle all cases
  }
}

// Returns a promise that resolves when all conditions to join a faction are met
function waitForFactionReqs(factionName){
  const requirements = ns.singularity.getFactionInviteRequirements(factionName)
  return Promise.all(requirements.map(waitForRequirement));
}
```

I'm struggling to see a way to get similar levels of usability out of the existing API (but again I might just be missing something).

I think realistically we're doing one of the following with the current API:

1. Something like `const [requirementType, data] = Object.entries(requirement)[0];`
    The problem here is that ts is basically useless here. We basically have to do hardcoded comparisons against different strings, with no autocomplete for the possible options, and getting the specific value of requirementType still gives us no info about data.

2. Using "in" to see if a property exists on the object, e.g. checking `"money" in requirement`. Again ts doesn't actually know that when the key exists, the value is no longer possibly undefined, so you then need to do another check to make sure it's defined.

3. Just boolean checking every possibility. This fails because a condition can have a numeric value of 0 which is falsey, but it's still a real requirement.\

4. Checking every property as !== undefined would probably be the cleanest way with the current API (**and actually it seems like a decent way to use the API, so I added an edit above**):

```js
/** @param {Requirement} requirement */
function waitForRequirement(requirement){
  if (requirement.money !== undefined) return waitForMoney(requirement.money);
  if (requirement.skills !== undefined) return waitForSkills(requirement.skills);
  if (requirement.companyRep !== undefined){
    const [company, rep] = requirement.companyRep;
    return waitForCompanyRep(company, rep);
  }
  if (requirement.someCondition !== undefined) return Promise.any(requirement.someCondition.map(waitForRequirement));
  // etc
}
```

Snarling:
See the edit on my previous comment, the structure of the API is probably fine as is. Let me know when you consider this ready to merge (need to run `npm run doc`, and fix the errors that cause that to fail)

jjclark1982:
Thanks for the review. I still have a lot to learn about TypeScript practices.

I have defined explicit sub-types with unique "type" fields, and gotten them to work with the doc generator and the in-game type annotations:

Screenshot 2023-12-03 at 1 42 18 PM

This has helped clarify some descriptions and edge cases and interactions with  #927.

jjclark1982:
@Snarling I reckon this is ready now. I have tested it with some simple user scripts using the "type" interface.

I still have a few concerns about code organization, but they can be revisited later without changing the API:

- Use of the terms "Requirement" and "Condition" is a bit ambiguous
- Class names refer specifically to FactionJoinConditions, but many of these apply to Company jobs, Milestones, and other types of checklists
- No great way for checklist UIs to flatten a nested list of "AND" conditions. For example:
    - Hacknet stats being split into 3 different types makes for a good-looking checklist, but makes the data structures more complex
    - "All combat stats" being combined into a single type matches the original documentation pages, but makes the checklists less responsive

jjclark1982:
Ok, I thought of a way to handle flattening nested condition lists that supports shorthand like "all combat stats" (for a concise docs page or a single item in the Netscript API) while also supporting separate checklist items.

Screenshot 2023-12-03 at 10 59 24 PM
Screenshot 2023-12-03 at 11 26 50 PM

This also simplifies the main internal API with a recursive top-level `isSatisfied` method:

```js
-    if (inviteReqs.every((req) => req.isSatisfied(this))) invitedFactions.push(faction);
+    if (inviteReqs.isSatisfied(this)) invitedFactions.push(faction);
```

but does not force a fully-recursive solution on users of the Netscript API.

PR #954, opened 2023-12-02
UI: For active scripts page, home always on top, then purchased servers, then the rest

TheAimMan:
Took the list of servers, which is derived from the workerscript objects based on pid value, and filtered it.

Placed home first, then any player purchased server, then any remaining.

Purchased servers are alphabetical

OrderedServers

Tested locally.  The array that I'm filtering is just the current servers running scripts, and not the scripts themselves.  Max size shouldn't exceed 100.  Didn't notice any performance hit when running 3000 scripts on 16 servers.

### DIFF
--- a/src/ui/ActiveScripts/ServerAccordions.tsx
+++ b/src/ui/ActiveScripts/ServerAccordions.tsx
@@ -78,7 +78,7 @@ export function ServerAccordions(props: IProps): React.ReactElement {
 
   // Match filter in the scriptname part of the key
   const pattern = matchScriptPathUnanchored(lodash.escapeRegExp(filter));
-  const filtered = Object.values(serverToScriptMap).filter((data) => {
+  const filtered = Object.values(serverToScriptMap).filter((data): data is IServerData => {
     if (!data) return false;
     if (data.server.hostname.includes(filter)) return true;
     for (const k of data.server.runningScriptMap.keys()) {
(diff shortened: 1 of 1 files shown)


Kelenius:
Placing home first makes sense but I don't see why purchased servers should be prioritized. There's nothing special to them.

TheAimMan:
My original intent was to just put home at the top but I received feedback that it would be nice to see the more important (purchased) servers on top.   Then I decided it would be really nice to order it by the name as well, and I am about to submit a PR to change the auto-naming for purchased servers and hacknet to append a 0 for single digit server creation.

Below is what my active scripts page looks like with my continious batcher without the changes

BeforeOrdered

And once the ordering is put in place, and the server names are consistent in naming with XX you get a page that looks like this

AfterOrdered

The active server page before the changes in my situation is basically useless.  The servers never stop scrolling.  If I actually wanted to select a server on that list it would be random chance.  And even if I did it would immediately be bounced up or down the list 400ms later.

TheAimMan:
Since it was decided that changing the name of hacknet servers would be a breaking change, I use that same logic (that hacknet won't change) and am now order hacknet servers by both name and the length of the name.  Server 0-9 are 1 shorter than 10-19, so I can find them, order them, and then add them to the array.

ServersStayInPlace

Snarling:
Sorry I totally didn't take this PR into account when making the changes on #1083. Ended up totally removing the conflicting ServerAccordions.tsx file (everything that was handled there is now handled one step up at the ActiveScriptsPage)

Snarling:
Reimplemented as #1102 since it would have been too complex to rebase this one, mind taking a look?

Issue #956, opened 2023-12-02
`formatNumber` returns NaN on Zero as input


steven-r:
- Release: Bitburner 2.5.1 (05875e)

The function `ns.formatNumber` doesn't return a proper result for `n = 0`, `fractionalDigits = 0` and `minDecimalPlaces = 0`. The expected result would be `0`, but the response is `NaNundefined ` .

Example-Code:
```js
ns.tprint("Format 1:       ", ns.formatNumber(1, 0, 0))
ns.tprint("Format 0:       ", ns.formatNumber(0, 0, 0))
ns.tprint("Format 0 (def): ", ns.formatNumber(0))
```

Expected output:
```
test_fm.js: Format 1:       1
test_fm.js: Format 0:       0
test_fm.js: Format 0 (def): 0.000
```

Current Output:
```
test_fm.js: Format 1:       1
test_fm.js: Format 0:       NaNundefined
test_fm.js: Format 0 (def): 0.000
```

steven-r:
As mentioned by TheMainMan (https://discord.com/channels/415207508303544321/415213413745164318/1180514580749885450):

looks like its due to this piece of the number formatter code

```js
if (nAbs < suffixStart) {
    if (isInteger) return basicFormatter.format(n);
    return getFormatter(fractionalDigits).format(n);
  }
```
issue goes away if I change it to 

```js
if (nAbs <= suffixStart) {
    if (isInteger) return basicFormatter.format(n);
    return getFormatter(fractionalDigits).format(n);
  }
```

New output:

```
Running script with 1 thread(s), pid 1 and args: [].
1.js: Format 1:       1
1.js: Format 0:       0
```

LJNeon:
This has been fixed and can be closed (see #957).

PR #957, opened 2023-12-02
BUGFIX: Update formatNumber.ts to allow formating 0 with 0 suffix

TheAimMan:
From the discord Bug Report room.  Player noticed they would get NaN when attempting to call ns.formatNumber

```
ns.tprint("Format 1:       ", ns.formatNumber(1, 0, 0))
ns.tprint("Format 0:       ", ns.formatNumber(0, 0, 0))
ns.tprint("Format 0 (def): ", ns.formatNumber(0))

```
```
test_fm.js: Format 1:       1
test_fm.js: Format 0:       NaNundefined
test_fm.js: Format 0 (def): 0.000
```

It looks like the code to check for non-suffixed numbers doesn't account for 0.  Adding in a check just for 0 allows the correct output in this case.

```
1.js: Format 1:       1
1.js: Format 0:       0
1.js: Format 0 (def): 0.000
```

### DIFF
--- a/src/ui/formatNumber.ts
+++ b/src/ui/formatNumber.ts
@@ -114,8 +114,8 @@ export function formatNumber(n: number, fractionalDigits = 3, suffixStart = 1000
   // Special handling for Infinities
   if (nAbs === Infinity) return n < 0 ? "-∞" : "∞";
 
-  // Early return for non-suffix
-  if (nAbs < suffixStart) {
+  // Early return for non-suffix or if number and suffix are 0
+  if (nAbs < suffixStart || (nAbs == 0 && suffixStart == 0)) {
     if (isInteger) return basicFormatter.format(n);
     return getFormatter(fractionalDigits).format(n);
   }
--- a/src/ui/formatNumber.ts
+++ b/src/ui/formatNumber.ts
@@ -113,9 +113,12 @@ export function formatNumber(n: number, fractionalDigits = 3, suffixStart = 1000
 
   // Special handling for Infinities
   if (nAbs === Infinity) return n < 0 ? "-∞" : "∞";
+  if (suffixStart < 1000) {
+    throw new Error("suffixStart must be greater than or equal to 1000");
+  }
 
   // Early return for non-suffix or if number and suffix are 0
-  if (nAbs < suffixStart || (nAbs == 0 && suffixStart == 0)) {
+  if (nAbs < suffixStart) {
     if (isInteger) return basicFormatter.format(n);
     return getFormatter(fractionalDigits).format(n);
   }


steven-r:
This fixes #956

Snarling:
This is a usage error, setting suffixStart to less than 1000. A more correct fix to avoid this misuse would be setting a minimum for suffixStart as 1000. While it's true that the number 0 should never have to go through a suffix calculation, that's also true for e.g. 654 or 1.

Leaving suffixStart at the default leads to the expected results for 0 digits:
```js
ns.tprint("Format 1:       ", ns.formatNumber(1, 0))
ns.tprint("Format 0:       ", ns.formatNumber(0, 0))
ns.tprint("Format 0 (def): ", ns.formatNumber(0))
```
image

steven-r:
I agree, but the bahaviour is inconsistent (looking at `1` and `0`). 

And if suffixStart < 1000 doesn't make sense (I agree on this), an exception should be thrown.

TheAimMan:
Its completely possible the player may never want to suffix their number but may want fractional digits.

```js
ns.tprint(ns.formatNumber(0,3,0))
ns.tprint(ns.formatNumber(0,3,1))
ns.tprint(ns.formatNumber(1,3,0))
```
image

It seems odd to me that having a suffix of 0 only breaks... 0.

d0sboots:
From an API perspective, using 0 to indicate "I never want a suffix" seems quite natural. The other alternative is requiring people to specify `1/0`, which seems... clunkier.

Snarling:
> Its completely possible the player may never want to suffix their number but may want fractional digits.
> ...
> It seems odd to me that having a suffix of 0 only breaks... 0.

suffixStart of 0 does not mean never use a suffix, it means start using a suffix as soon as the number is at least 0. Since there is no suffix for numbers until 1000, setting suffixStart to lower than 1000 does not mean anything. Infinity would be the correct suffixStart to say that you never want a suffix applied.

image

0 breaks when you use suffixStart of 0 because it tries to do a suffix calculation (since you are at or above the suffixStart), and a suffix calculation fails for 0 due to log math. But other than making 0 format as NaN, setting suffixStart to 0 does not do anything.

Suffixes are really the only bitburner-specific part of number formatting. If you just want to format a number without using bitburner suffixes, there's no reason to rely on the game's API for that (but you still can, by telling it to never use suffixes as shown above).

d0sboots:
It sounds like you're set on 0 being an error, in which case I agree with Steven that an exception should be thrown (possibly pointing to infinity as the correct value)

TheAimMan:
So it sounds like any suffix below 1000 should result in an error being thrown, and the documentation for the command should be updated to reflect that.

Snarling:
I'd also be okay with just using helpers.positiveInteger to validate suffixStart (throws for 0 which is the one case that really causes an issue), but a custom error for any input less than 1000 would be able to more clearly explain to the player why numbers lower than 1000 don't make sense (can't start using suffixes before suffixes exist).

TheAimMan:
Just changed it to throw an error if suffixStart is less than 1000.  Error will now be thrown stating as such.

image

PR #958, opened 2023-12-03
IMPROVEMENT: Pad purchased and hacknet server numbers

TheAimMan:
Hacknet and purchased servers automatically get incremented numerically, but it's not a padded number.  If you want to sort by server name, 10 will come before 2.

Padded the functions to add a leading 0 for digits 0-9. so the servers would be ordered properly.

Change is mainly to coincide with https://github.com/bitburner-official/bitburner-src/pull/954#issuecomment-1837259330

### DIFF
--- a/src/PersonObjects/Player/PlayerObjectServerMethods.ts
+++ b/src/PersonObjects/Player/PlayerObjectServerMethods.ts
@@ -45,7 +45,7 @@ export function getUpgradeHomeCoresCost(this: PlayerObject): number {
 
 export function createHacknetServer(this: PlayerObject): HacknetServer {
   const numOwned = this.hacknetNodes.length;
-  const name = hasHacknetServers() ? `hacknet-server-${numOwned}` : `hacknet-node-${numOwned}`;
+  const name = hasHacknetServers() ? `hacknet-server-${String(numOwned).padStart(2,'0')}` : `hacknet-node-${String(numOwned).padStart(2,'0')}`;
   const server = new HacknetServer({
     adminRights: true,
     hostname: name,
--- a/src/PersonObjects/Player/PlayerObjectServerMethods.ts
+++ b/src/PersonObjects/Player/PlayerObjectServerMethods.ts
@@ -45,7 +45,9 @@ export function getUpgradeHomeCoresCost(this: PlayerObject): number {
 
 export function createHacknetServer(this: PlayerObject): HacknetServer {
   const numOwned = this.hacknetNodes.length;
-  const name = hasHacknetServers() ? `hacknet-server-${String(numOwned).padStart(2,'0')}` : `hacknet-node-${String(numOwned).padStart(2,'0')}`;
+  const name = hasHacknetServers()
+    ? `hacknet-server-${String(numOwned).padStart(2, "0")}`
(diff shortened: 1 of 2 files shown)


Snarling:
This is a minor API break in that it changes the actual names of the one-digit hacknet servers (and purchased servers if the player is relying on name collision to automatically increment the server name). Any player scripts that rely on the existing names would need to be updated to also add in manual padding of the name. So I'm not sure how I feel about this one.

steven-r:
Two points here:

* For consistency, the first server at all, which currently results into `hostname`, without any appendix, should be `-00`
* The second server should be named `-01`, instead of `-0`

Maybe this can be done with an optional parameter (e.g. `alignNames`)

TheAimMan:
Yeah, renaming the hacknet servers broke my stuff... but that's what I get for hardcoding a potential dynamic property instead of scanning the possible servers on the network :).  It's mainly to get a consist result from the linked PR to it doesn't drive me crazy seeing 10,11,12, etc before 2.

As for the purchased servers, players name their servers.  The number's aren't appended to a purchased server unless the name already exists.  We can't know if the player is going to name their servers all the same when they purchase their first server, hence why there is no number attached to the end.

d0sboots:
It puts more work on your other PR, but I think the better solution would be to sort properly using the existing scheme. Yeah, I know it's hard :/

TheAimMan:
I can't think of a scenario where a solution isn't a janked together amalgamation of ugliness.
The current PR I'm doing this for is probably already questionable to actual coders.  It's an attempt to implement a feature that would really best be served earlier on the in process.

The best scenario I could think of would be a new server object literally called order that would assign home 1, any pservers 10-100 (just in case that ever changes). Hacknet servers 200-300, and the remaining servers 300+.  Then we could just order by that attribute.

But then, due to current saves whose servers wouldn't have that object, I'd have to check to make sure it exists first or else the sorting function breaks for saves loaded prior to the server object change.

I'm struggling to come up with a solution that has to take a list of string values, divy them into 3-4 different lists, and order them logically, then smash them all together again.

I'd rather just not attempt it, let server 10 be before server 2 in the list, and not let perfect become the enemy of good.

TheAimMan:
Bah, it's always fitting after a bit of a rant you find a partial solution.  It appears hacknet servers have an 'onlineTime' server property so I could probably just order by that desc.  

If the hacknet server names aren't changed, is this PR as breaking if just the pserver numerator was changed from 1 to 01?

Snarling:
> If the hacknet server names aren't changed, is this PR as breaking if just the pserver numerator was changed from 1 to 01?

As you mentioned in a previous comment, purchased servers are normally numbered by what the player decides to name them. It's not expected that the player is just going to keep trying to name servers the same thing and rely on name collision to increment the name automatically for them. The player is expected to increment the names themselves however they want.

With that in mind, it's probably fine to make a change to how the name collision handles automatically renaming the server (especially since purchaseServer returns the name given to the server), but I also don't really see the point and it would also not match the non-padded numbering of hashnet.

TheAimMan:
Good point.  Since the player has control over the naming, they can control if they want their servers named -1 or 01, or whatever.  
I'll go ahead and close this PR then as without the hacknet padding changes, there is no reason to do the purchased server padding changes.

PR #963, opened 2023-12-06
MISC: Add a potential of more than 1 core for initial servers

TheAimMan:
Thought it would be fun to add more than 1 core to the existing servers.
Core count is based on network layer.  The first few layers will have 1 core, then randomly it can go up to 7.  
If a server rolls its minimum RAM calculation, it gets an extra core which means the max core count for existing servers is 8.

Since existing servers are early game, this shouldn't a crazy change.  Current scripts that don't factor in core count can continue to ignore it. It just gives an incentive to look at cores on servers other than home and hacknet servers.


jjclark1982:
This seems like a good gameplay improvement for hack/grow/weaken optimization.

Currently the only servers with multiple cores are `home` and BN9 servers. (BN9 servers generally contribute more to hacking profitability by running no scripts.) A memory manager can ignore cores completely and still deliver RAM to a batcher with nearly-optimal performance.

The basic memory management strategy at various points in the game is:

- Most RAM is available on initial servers: run scripts on any available server.
- Most RAM is available on purchased servers: run most scripts on pservers. The extra RAM outweighs any extra cores, but a smart memory manager can increase grow efficiency.
- Most RAM is available on `home` server: run most scripts on `home`. The extra RAM outweighs any extra cores, but a smart memory manager can increase grow efficiency.

So this PR adds more benefits to smart memory management in the early game, but does not require players to pay attention to it before solving other problems.

src/Server/AllServers.ts
```diff
@@ -144,16 +149,27 @@ export function initForeignServers(homeComputer: Server): void {
...
     if (metadata.requiredHackingSkill) serverParams.requiredHackingSkill = toNumber(metadata.requiredHackingSkill);
     if (metadata.serverGrowth) serverParams.serverGrowth = toNumber(metadata.serverGrowth);
 
     const server = new Server(serverParams);
+
+    if (metadata.networkLayer) {
+      const maxCount = Math.floor(toNumber(metadata.networkLayer) / 2);
+      const minCount = Math.ceil(toNumber(metadata.networkLayer) / 4);
+      const coreCount = Math.floor(Math.random() * (maxCount - minCount + 1)) + minCount;
+      server.cpuCores = toNumber(coreCount);
```
Snarling: no need to use `toNumber` on coreCount since we know it's already a number. toNumber is for parameters that might be a number but might also be a minmax range.

src/Server/AllServers.ts
```diff
@@ -144,16 +149,27 @@ export function initForeignServers(homeComputer: Server): void {
...
     if (metadata.moneyAvailable) serverParams.moneyAvailable = toNumber(metadata.moneyAvailable);
     if (metadata.requiredHackingSkill) serverParams.requiredHackingSkill = toNumber(metadata.requiredHackingSkill);
     if (metadata.serverGrowth) serverParams.serverGrowth = toNumber(metadata.serverGrowth);
 
     const server = new Server(serverParams);
+
+    if (metadata.networkLayer) {
+      const maxCount = Math.floor(toNumber(metadata.networkLayer) / 2);
+      const minCount = Math.ceil(toNumber(metadata.networkLayer) / 4);
+      const coreCount = Math.floor(Math.random() * (maxCount - minCount + 1)) + minCount;
```
Snarling: Not that important, but the existing getRandomInt helper function can be used here instead of writing out the calculation again.

src/Server/AllServers.ts
```diff
@@ -144,16 +149,27 @@ export function initForeignServers(homeComputer: Server): void {
...
     if (metadata.hackDifficulty) serverParams.hackDifficulty = toNumber(metadata.hackDifficulty);
     if (metadata.moneyAvailable) serverParams.moneyAvailable = toNumber(metadata.moneyAvailable);
     if (metadata.requiredHackingSkill) serverParams.requiredHackingSkill = toNumber(metadata.requiredHackingSkill);
     if (metadata.serverGrowth) serverParams.serverGrowth = toNumber(metadata.serverGrowth);
 
     const server = new Server(serverParams);
+
+    if (metadata.networkLayer) {
+      const maxCount = Math.floor(toNumber(metadata.networkLayer) / 2);
+      const minCount = Math.ceil(toNumber(metadata.networkLayer) / 4);
```
Snarling: For network layer 1 the "maxCount" is 0 since it's floor(1/2) but the minCount is 1 since it's ceil(1/4). maxCount should not be lower than minCount.

Snarling (changes requested):
Sorry for the late review, just have some minor comments. Other than the minor gripes below I have no issues with this being merged.

TheAimMan:
So its been a long time since I wrote this, and you commented on it right around the time I stepped away for a bit.  I basically ripped everything out and simplified the calculation for cores to use already existing utilities.

src/Server/AllServers.ts
```diff
@@ -144,16 +144,21 @@ export function initForeignServers(homeComputer: Server): void {
...
 
     const server = new Server(serverParams);
+
+    if (metadata.networkLayer) {
+      const layer = toNumber(metadata.networkLayer);
+      server.cpuCores = getRandomIntInclusive(Math.ceil(layer / 2), layer);
+    }
+    if (metadata.maxRamExponent !== undefined) {
+      serverParams.maxRam = Math.pow(2, toNumber(metadata.maxRamExponent));
+    }
```
Snarling: This should be moved back to somewhere before line 152. 152 is creating the server based on server params, so all creation params need to be set before that.

It looks like it should be fine where it was originally

Snarling:
Sorry @TheAimMan for the late re-review

PR #965, opened 2023-12-06
MISC: Switch port.nextWrite() from LIFO to FIFO

LJNeon:
This has been discussed on Discord already and the consensus is that FIFO behavior makes more sense and is worth changing to. This is a breaking change but there's no evidence of anyone actually using multiple nextWrites on the same port, and those who are considering it would need FIFO behavior for what they want to do.

This should be trivial to test but I can't do so on mobile.

### DIFF
--- a/src/NetscriptPort.ts
+++ b/src/NetscriptPort.ts
@@ -42,7 +42,7 @@ export function writePort(n: PortNumber, value: unknown): PortData | null {
       `port.write: Tried to write type ${typeof value}. Only string and number types may be written to ports.`,
     );
   }
-  let { data, resolvers } = getPort(n);
+  const { data, resolvers } = getPort(n);
   data.push(value);
   for (const res of resolvers.splice(0, resolvers.length)) res();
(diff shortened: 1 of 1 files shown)


src/NetscriptPort.ts
```diff
@@ -42,9 +42,10 @@ export function writePort(n: PortNumber, value: unknown): PortData | null {
       `port.write: Tried to write type ${typeof value}. Only string and number types may be written to ports.`,
     );
   }
-  const { data, resolvers } = getPort(n);
+  let { data, resolvers } = getPort(n);
   data.push(value);
-  while (resolvers.length > 0) resolvers.pop()?.();
+  for (const res of resolvers) res();
+  resolvers = [];
```
Snarling: As written this does not clear out old resolvers from the port. The new empty array is assigned to the `resolvers` variable from this function, but this does not eliminate any resolvers stored on the port.

LJNeon: oh I see, I'm modifying a local variable instead of the original array

LJNeon: Should be fixed now.

PR #969, opened 2023-12-07
BUGFIX: Properly apply entropy to hacknet multiplers

TheAimMan:
Entropy is supposed to be a negative effect applied to the player when grafting augments.  Current implementation of entropy actually reduces the costs of hacknet instead of increasing the cost.

Changed the calculation to divide by entropy instead of multiply

199 Entropy before:
image

199 Entropy after:
image

### DIFF
--- a/src/PersonObjects/Grafting/EntropyAccumulation.ts
+++ b/src/PersonObjects/Grafting/EntropyAccumulation.ts
@@ -31,7 +31,7 @@ export const calculateEntropy = (stacks = 1): Multipliers => {
     crime_money: Player.mults.crime_money * nerf,
     crime_success: Player.mults.crime_success * nerf,
 
-    hacknet_node_money: Player.mults.hacknet_node_money / nerf,
+    hacknet_node_money: Player.mults.hacknet_node_money * nerf,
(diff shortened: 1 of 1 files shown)


src/PersonObjects/Grafting/EntropyAccumulation.ts
```diff
@@ -31,11 +31,11 @@ export const calculateEntropy = (stacks = 1): Multipliers => {
     crime_money: Player.mults.crime_money * nerf,
     crime_success: Player.mults.crime_success * nerf,
 
-    hacknet_node_money: Player.mults.hacknet_node_money * nerf,
-    hacknet_node_purchase_cost: Player.mults.hacknet_node_purchase_cost * nerf,
-    hacknet_node_ram_cost: Player.mults.hacknet_node_ram_cost * nerf,
-    hacknet_node_core_cost: Player.mults.hacknet_node_core_cost * nerf,
-    hacknet_node_level_cost: Player.mults.hacknet_node_level_cost * nerf,
+    hacknet_node_money: Player.mults.hacknet_node_money / nerf,
```
jjclark1982: I think this should still be `*` as it is a beneficial value

TheAimMan: Updated

Issue #416, opened 2023-03-08
Corps - Quality of life issues (10 key enter fails, and previous sell not offered as default)


OneOfMany07:
Not new issues since around 2.1 when I started recently, but I keep hitting them so thought I'd report them together for anyone working on corporations.

1. Click Black Op's "Set Team Size" button.  Type lots of 9's to have it shift to using all available recruits.  Press enter on the 10 key... I expect the dialog to act like I clicked save/OK.  It currently does nothing.  I have to use my mouse to click or I think maybe tab manually to the confirm button (I rarely try that in the game...sadly I need my mouse for lots).  That enter key does work sometimes, which is part of why I keep triggering it.

2. Create a new corp, new industry, set sales price for first city (MAX, MP).  Now expand to new city and attempt the same.  The default values will be blank instead of keeping the old values like sometimes happens (and speeds manually setting up all cities).  If you first expand into all locations (and get warehouse?), then set a price at one, it seems to remember the old value for the next.  I might be missing a detail here, but this should be close..

Thanks in advance for sharing everyone's great work here!

Version: Steam v2.2.2 (6eb5b5ab)

Snarling:
Fixed the first issue via #984

hexagonrecursion:
> Keeping the old values like sometimes happens

Keeping the old values is actually unintentional. When you open the modal, the sell amount and sell price should always match the ones currently configured for a given city.

Here is what happens:
1. When you select a city in the corp UI, several instances of `SellMaterialModal` are mounted (one for each input, output or boost material) _if they do not already exist_
2. When a `SellMaterialModal` is mounted, it reads desiredSellAmount and desiredSellPrice for the corresponding material and stores it in its own private state variables
3. When you switch to a different city or a different division, if the city has a warehouse, the instances of `SellMaterialModal` are reused and _their state is not updated_
4. When you switch to a UI that does not contain a `SellMaterialModal` (e.g to expand to a new city), `SellMaterialModal`s are unmounted. Subsequently fresh `SellMaterialModal`s will be mounted when needed

Other affected modals:
1. `PurchaseMaterialModal`
2. The division selector, city selector and export amount field in the `ExportModal`
3. `SellProductModal` is affected even if "Set for all cities" is turned off

Unaffected:
1. The list of exports in `ExportModal` is not affected because it is directly bound to the corresponding list in corporation state
2. `LimitMaterialProductionModal` because it explicitly resets internal state when it is closed
   ```ts
[...]

PR #975, opened 2023-12-13
BLADEBURNER: support Enter Key in team size modal

jjclark1982:
This PR fixes a UI bug reported in #416


Snarling:
Same functionality added via #984

PR #976, opened 2023-12-13
FACTIONS: Support favor gain over 34,000

jjclark1982:
Reported on Discord by @Azirale — 12/11/23, 12:00 PM
> Might want a Number.MAX_VALUE guard on this, or something, because after a reset the first tick of working for the faction will give you literally infinite money. It also breaks the favor on the company, because it goes NaN after the next install.
It seems to be once you get past some septillion or octillion rep, something in the favour calculation hits infinite, and it just explodes.

In order to support favor values between 34,867 and Infinity, we can perform calculations in a lower base. Using a base of 1.00002 supports favor values up to about 3.4e7.

It might be better gameplay to support stacking up to the max value each prestige by growing arithmetically instead of logarithmically above level. I will try to graph the difference between these formulas...

### DIFF
--- a/src/Company/Company.ts
+++ b/src/Company/Company.ts
@@ -104,11 +104,17 @@ export class Company {
   }
 
   getFavorGain(): number {
-    if (this.favor == null) this.favor = 0;
-    const storedRep = Math.max(0, favorToRep(this.favor));
+    const storedFavor = Math.max(0, this.favor || 0);
+    const storedRep = Math.max(0, favorToRep(storedFavor));
     const totalRep = storedRep + this.playerReputation;
-    const newFavor = repToFavor(totalRep);
-    return newFavor - this.favor;
+    if (totalRep <= Number.MAX_VALUE) {
+      const newFavor = repToFavor(totalRep);
+      return newFavor - storedFavor;
+    }
+    else {
+      // Above effectively-infinite favor levels, support gaining up to the max amount each prestige.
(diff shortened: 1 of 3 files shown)


d0sboots:
Why not not just do like the suggestion, and cap rep at MAX_VALUE? (Or some other large, non-infinite number.) I don't get the point of the extra bases or trying to change the formula.

jjclark1982:
If all we do is cap the favorToRep formula, then it can still overflow when adding and subtracting reps in getFavorGain. A limit also needs to be applied to either total favor, or favor gain per augmentation.

Changing to a lower base allows the formula to stay the same with no hard cap on total favor, but looking at the values this produces I think that the diminishing returns are too extreme with a single continuous formula, and might as well be rounded down to 0.

Changing the formula above the soft cap allows players to make measurable gains above 34,000, even though there is no game content at those levels.

I will probably back one of these out after a bit more playtesting.

cigarmemr:
Just encounted the same thing. Once the reputation of a company reaches certain value, the favor will become infinity on next soft reset:
image

A following soft reset will render the favor to become NaN:
image

Finally another soft reset, the favor will be trapped in NaN and cannot go back to some sensible value again:
image

Hope this will be fixed ASAP

gmcew:
#1321 handles the overflow issue that started this, any future change to that favor cap is probably best as a new PR.

PR #977, opened 2023-12-15
CORPORATION: Account for CapEx and Cap Gains differently

jjclark1982:
This PR handles some edge cases in #949 reported by jeek

- Count capital expenditures as positive assets during the cycle they are purchased.  
(In the next cycle they will depreciate to their recoupable value.)
- Count capital gains as virtual positive assets during the cycle they are realized.  
(In the next cycle they will remain as cash and not effect projected earnings.)

This fixes an exploit where some capital expenditures were counted as capital gains instead, resulting in absurd valuations.

### DIFF
--- a/src/Corporation/Actions.ts
+++ b/src/Corporation/Actions.ts
@@ -74,7 +74,7 @@ export function purchaseOffice(corporation: Corporation, division: Division, cit
   if (division.offices[city]) {
     throw new Error(`You have already expanded into ${city} for ${division.name}`);
   }
-  corporation.loseFunds(corpConstants.officeInitialCost, "office");
+  corporation.loseFunds(corpConstants.officeInitialCost, "division");
   division.offices[city] = new OfficeSpace({
     city: city,
     size: corpConstants.officeInitialSize,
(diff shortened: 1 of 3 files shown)


jjclark1982:
Script by @CatLover  demonstrating the exploit:

```js
const cities = ["Aevum", "Sector-12", "Chongqing", "Ishima", "New Tokyo", "Volhaven"];

/** @param {NS} ns */
  ns.corporation.createCorporation("Corp", false);
  for (let i = 0; i < 3; i++) {
      const divisionName = Math.random().toFixed(5).toString();
      ns.corporation.expandIndustry("Restaurant", divisionName);
      const division = ns.corporation.getDivision(divisionName);
      for (const cityName of cities) {
          if (ns.corporation.getCorporation().funds < 4e9) {
              break;
          }
          if (!division.cities.includes(cityName)) {
              ns.corporation.expandCity(divisionName, cityName);
          }
          if (!ns.corporation.hasWarehouse(divisionName, cityName)) {
              ns.corporation.purchaseWarehouse(divisionName, cityName);
          }
      }
  }
  ns.tail();
  while (true) {
      ns.print("state:", ns.corporation.getCorporation().prevState);
      ns.print("offer:", ns.formatNumber(ns.corporation.getInvestmentOffer().funds));
      await ns.corporation.nextUpdate();
  }
}
```

PR #978, opened 2023-12-15
MISC: Revert default HGW threads from 1 to number of script threads

DJMatch3000:
A recent update changed the default behavior of hack(), grow(), and weaken() to use one thread instead of the total script threads. This reverts that change.


d0sboots:
Oops, that was my oversight wasn't it. Thanks for the patch!

PR #984, opened 2023-12-18
BLADEBURNER: Different solution for Enter key support in team member modal

Snarling:
Alternative to #975 , Using a form to add keyboard support for submitting a modal that is essentially a form (instead of sending custom onKeyDown handler on the text field that manually checks for the enter key).

@jjclark1982 Any concerns with using this approach instead?


jjclark1982:
I like this approach using the form submit event.

When I have some time I would like to apply this same method to the 20 other modals that use a `<TextField>` and check for a `KEY.ENTER` event.

PR #989, opened 2023-12-24
API: Updating typing for ns.singularity.getCurrentWork()

Semanual:
The typing of `ns.singularity.getCurrentWork()` was `any`, so it was not at all helpful. I created new types based on the `type` variable. Now, when calling the function you have autocomplete for all different types of work.
Example:
imagem
For each `task`, it has their own type, and their own autocomplete.
Testing code:
```js
/** @param {NS} ns */
  const task = ns.singularity.getCurrentWork();
  switch (task.type) {
    case "CLASS":
      task; // hover the cursor
      break;
    case "COMPANY":
      task; // hover the cursor
      break;
    case "CREATE_PROGRAM":
      task; // hover the cursor
      break;
    case "CRIME":
      task; // hover the cursor
      break;
    case "FACTION":
      task; // hover the cursor
      break;
    case "GRAFTING":
      task; // hover the cursor
      break;
  }
}
```


Snarling:
Looks good, nice work! There was some redundancy in type declarations, so I got rid of the duplicates to made it work a bit more like the SleeveWork, see the changes in 627bafe

Semanual:
That redundancy was because of enum vs primitive strings declarations, some of them are just primitive strings when we have an enum (because otherwise we would show all factions names' for example for the user). But okay, if not using the enums in code is enough, that's perfect.

PR #1007, opened 2024-01-01
NETSCRIPT: Add .script deprecation notice & migration guide

LJNeon:
I don't feel the need to justify or explain this change as it has already been agreed to unanimously on Discord.

I can't test this PR, both the migration guide and the terminal warning need to be tested before merge.

### DIFF
--- a/src/Documentation/doc/migrations/ns2.md
+++ b/src/Documentation/doc/migrations/ns2.md
@@ -2,7 +2,7 @@
 
 The game allows two script formats:
 
-- `.script` (also sometimes called "Netscript 1" or "NS1") files are ran through an interpreter that is based on a version of Javascript from 2009 (ES5). These files are no longer actively supported and should be converted to the newer `.js` format. 
+- `.script` (also sometimes called "Netscript 1" or "NS1") files are ran through an interpreter that is based on a version of Javascript from 2009 (ES5). These files are no longer actively supported and should be converted to the newer `.js` format.
 
 - `.js` (also sometimes called "Netscript 2" or "NS2") files are native javascript that is ran directly by the web browser. Modern features of javascript that are supported by your web browser are supported in `.js` files, because they are run as native js.
 
--- a/src/Documentation/doc/migrations/ns2.md
+++ b/src/Documentation/doc/migrations/ns2.md
@@ -1,0 +1,57 @@
+# Netscript 2 Migration Guide
+
+With the deprecation of Netscript 1 there are required changes to migrate over.
+
+## Basic Steps
+1. You need to wrap the entire script inside of an exported main function, like so:
+```js
+/** @param {NS} ns */
+export async function main(ns) {
+  // your code here
+}
+```
+2. Add `ns.` as a prefix to all game functions (such as `ns.hack()` or `ns.nuke()`).
+3. If a function returns a `Promise`, you need to put the `await` keyword before it (With the JSDoc comment you can hover over the function to see the return type).
+4. Long-running loops (like `while(true)`) need to `await` a function (usually `ns.sleep()`) at least once or they will crash the game.
+
+## A demonstration of migration
+
+Original (`early-hacking-template.script`):
+```js
+var target = "n00dles";
+var moneyThresh = getServerMaxMoney(target) * 0.9;
+var securityThresh = getServerMinSecurityLevel(target) + 5;
+
+while(true) {
+  if (getServerSecurityLevel(target) > securityThresh) {
+    weaken(target);
+  } else if (getServerMoneyAvailable(target) < moneyThresh) {
+    grow(target);
+  } else {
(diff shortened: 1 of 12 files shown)


LJNeon:
Also I am unable to run prettier, hopefully that can be done via maintainer edits which should be enabled.

src/Documentation/doc/migrations/ns2.md
```diff
@@ -0,0 +1,57 @@
+# Netscript 2 Migration Guide
+
+With the deprecation of Netscript 1 there are required changes to migrate over.
```
Alpheus: Add a hint to whom this migration doc is for, ie.
"... if you have any .script files." If all your scripts are .js you do not need to follow this guide.

src/Documentation/doc/migrations/ns2.md
```diff
@@ -0,0 +1,57 @@
...
+    } else {
+      await ns.hack(target);
+    }
+  }
+}
+```
+
+## Additional problems or edge cases
+
+To get additional help with the migration join the [official Discord server](https://discord.gg/TFc3hKD).
```
Alpheus, line 78: Great shoutout for discord!

src/Terminal/commands/common/editor.ts
```diff
@@ -47,10 +49,17 @@ export function commonEditor(
...
+    if (path.endsWith(".script")) hasNs1 = true;
     const file = server.getContentFile(path);
     const content = file ? file.content : isNs2(path) ? newNs2Template : "";
     files.set(path, content);
     if (content === newNs2Template) CursorPositions.saveCursor(path, { row: 3, column: 5 });
   }
+  if (hasNs1) {
+    Terminal.warn(
+      "NOTICE: NS1 (.script) scripts are deprecated and will be removed in a future update." +
+      " Migrate to NS2 (.js) scripts instead."
```
Alpheus: Please add a comment to seek the migration guide in the documentation, ie.
" Migrate to NS2 (.js) scripts following the migration guide in the documentation."

src/Terminal/commands/runScript.ts
```diff
@@ -52,6 +52,12 @@ export function runScript(path: ScriptFilePath, commandArgs: (string | number |
   const success = startWorkerScript(runningScript, server);
   if (!success) return Terminal.error(`Failed to start script`);
 
+  if (path.endsWith(".script")) {
```
Alpheus, line 56: We'll have to hunt these migration notices down to remove them once deprecation is final. 

Extract to a migration ts file or leave a comment so the search for them can be semi-automated.

LJNeon: They are only in two places, the commonEditor function and the runScript function. Also it's extremely easy to find them by searching "migrate", I just checked.

src/Documentation/doc/migrations/ns2.md
LJNeon: Good idea, done.

src/Terminal/commands/runScript.ts
d0sboots: I agree with idea of extraction, for different reasons: we might want to adjust the message, or the criteria (maybe only the first time you edit?) and that should be kept in sync.

LJNeon: I've moved it out into it's own file, that should make modifications trivial.

Snarling (commented):
The terms "ns1" and "ns2" are not really accurate and should not be used. .script are a legacy script format that is no longer supported, and .js are javascript files.

The original game documentation pretended that "Netscript" was the language used by game scripts. It's not it's own language, and it's helpful for players if they know the scripts they are writing are literally javascript. `.script` files were an interpreted version of javascript based on es5, whereas `.js` files are actual javascript modules running natively within the player's browser. We intentionally removed references to "ns1" and "ns2" a long time ago.

Netscript is still the name of the game's API, which is passed in as the first argument for player script `main` functions with the default parameter name `ns`. But there is no "ns1" or "ns2", as both script formats can access the same API (although some parts of it don't work in .script files)

LJNeon:
So you want to go by just file extensions instead? Or do you want that specific "legacy script format" and "javascript files" distinction?

Alpheus:
> So you want to go by just file extensions instead? Or do you want that specific "legacy script format" and "javascript files" distinction?

It seems the distinction is in the "execution format" being deprecated, not anything Netscript-related. Netscript is the API and it stays. In that context the message applies to: "deprecating the ability to run '.script' files to access the Netscript API".

LJNeon:
@Snarling is that wording better?

Snarling:
@LJNeon I made some additional changes and added some more content, mind looking this over before I merge it? Sorry for the delay in following up on this.

LJNeon:
Looks good to me!

PR #1013, opened 2024-01-02
API: Fix `serverExists()` return true for not revealed servers

### LINKED ISSUE #980: serverExists() returns `true` for not revealed servers
### Description
As of version 2.5.1, some NS functions that work on hostnames will raise `Invalid hostname` error when the server is not revealed yet. However `serverExists()` still returns true for these servers.
This will lead some scripts to fail especially if they pass script arguments directly to those functions, as `data.servers` in autocomplete() still contains the unreachable servers.
In this case returning `false` will let the scripts to skip the unreachable servers more easily by calling `serverExists()` and make a judgement.

### Steps to reproduce
Execute a script with statement `ns.serverExists('w0r1d_d43m0n')` BEFORE the player takes Red Pill.
OR
(Not tested) Execute a script with statement `ns.serverExists('darkweb')` BEFORE the player buys tor router.

### Result
The function returns `true` as if the server exists.

### Version
Bitburner v2.5.1 (0c2997d32)

cigarmemr:
closes #980

copied condition from [here](https://github.com/bitburner-official/bitburner-src/blob/dev/src/Netscript/NetscriptHelpers.tsx#L456)

### DIFF
--- a/src/NetscriptFunctions.ts
+++ b/src/NetscriptFunctions.ts
@@ -1116,7 +1116,8 @@ export const ns: InternalAPI<NSFull> = {
   },
   serverExists: (ctx) => (_hostname) => {
     const hostname = helpers.string(ctx, "hostname", _hostname);
-    return GetServer(hostname) !== null;
+    const server = GetServer(hostname);
+    return server !== null && server.serversOnNetwork.length > 0 || server.hostname === "home";
   },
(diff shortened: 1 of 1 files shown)


ghost:
Recently ran into this issue, glad it's getting fixed.

PR #1016, opened 2024-01-04
IPVGO: Remove testing flag [Merge once dev cycle is almost complete]

ficocelliguy:
Currently, IPvGO always appears as a shortcut in the left-hand menu (to make it easy to test). This PR makes it only appear if the user is in BN14, or if the user has an active IPvGO subnet ongoing.

(The subnet is accessible via the UI from DefCon in new Tokyo or CIA in Sector-12, or via the API)


Snarling:
Marked as draft just so I can more easily ignore this when looking through PRs, will unmark as draft before merging

ficocelliguy:
When you make the patch notes, please mention the entrance locations for the mechanic ( DefCon in New Tokyo or CIA in Sector-12 ) for people to try it out. New players will stumble across it just exploring, but after your first few BNs I imagine people stop clicking around in random locations like that.

PR #1017, opened 2024-01-05
EDITOR: Add file URI's to monaco to improve intellisense

shyguy1412:
This change adds file URI's to the monaco models, allowing monaco to resolve imports to the corresponding scripts. This means that imports can now be typed with JSDoc and have autocomplete.

### DIFF
--- a/src/ScriptEditor/ui/OpenScript.ts
+++ b/src/ScriptEditor/ui/OpenScript.ts
@@ -24,6 +24,13 @@ export class OpenScript {
   }
 
   regenerateModel(): void {
-    this.model = editor.createModel(this.code, this.isTxt ? "plaintext" : "javascript");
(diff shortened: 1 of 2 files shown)


d0sboots:
I'm not quite sure what this is supposed to be doing from the description. Could you include a screenshot or something?

shyguy1412:
This allows allows monaco to actually resolve import paths to the correct scripts for their type definitions, given that the imported script is also currently open

firefox_C4kJWhZ07r
firefox_RWo44wFISA

Without this the types wouldnt be available

firefox_kefsGZD9kD

d0sboots:
I'm surprised it was that easy, given our psuedo FS. Nice job!

Snarling:
That is awesome! I will merge this later today before pushing the next dev version update.

PR #1019, opened 2024-01-05
PORTS: added rejection to nextWrite

Caldwell-74:
one common problem with nextWrite are the lingering resolvers on the port when the script is already dead

we would now push a closure instead of just the resolvers
containing a reference to the ctx and based on the env.stopFlag reject or resolve the promise

a rejection is catched with a small error message in the tail
image

### DIFF
--- a/src/NetscriptPort.ts
+++ b/src/NetscriptPort.ts
@@ -2,6 +2,8 @@ import { Settings } from "./Settings/Settings";
 import { NetscriptPort } from "@nsdefs";
 import { NetscriptPorts } from "./NetscriptWorker";
 import { PositiveInteger } from "./types";
+import { NetscriptContext } from "./Netscript/APIWrapper";
+import { helpers } from "./Netscript/NetscriptHelpers";
 
 type PortData = string | number;
 type Resolver = () => void;
@@ -23,13 +25,13 @@ export class Port {
   data: PortData[] = [];
   resolvers: Resolver[] = [];
 }
-export function portHandle(n: PortNumber): NetscriptPort {
+export function portHandle(n: PortNumber, ctx: NetscriptContext): NetscriptPort {
   return {
     write: (value: unknown) => writePort(n, value),
     tryWrite: (value: unknown) => tryWritePort(n, value),
     read: () => readPort(n),
     peek: () => peekPort(n),
-    nextWrite: () => nextWritePort(n),
+    nextWrite: () => nextWritePort(n, ctx),
     full: () => isFullPort(n),
     empty: () => isEmptyPort(n),
     clear: () => clearPort(n),
@@ -76,9 +78,11 @@ export function peekPort(n: PortNumber): PortData {
   return port.data[0];
 }
 
-function nextWritePort(n: PortNumber) {
+function nextWritePort(n: PortNumber, ctx: NetscriptContext) {
   const { resolvers } = getPort(n);
-  return new Promise<void>((res) => resolvers.push(res as Resolver));
+  return new Promise<void>((res, rej) => resolvers.push(() => (ctx.workerScript.env.stopFlag ? rej() : res()))).catch(
+    () => helpers.log(ctx, () => `nextWrite was rejected on port ${n}`),
+  );
 }
 
 function isFullPort(n: PortNumber) {
--- a/src/NetscriptPort.ts
+++ b/src/NetscriptPort.ts
@@ -4,6 +4,7 @@ import { NetscriptPorts } from "./NetscriptWorker";
 import { PositiveInteger } from "./types";
 import { NetscriptContext } from "./Netscript/APIWrapper";
 import { helpers } from "./Netscript/NetscriptHelpers";
+import { ScriptDeath } from "./Netscript/ScriptDeath";
 
 type PortData = string | number;
 type Resolver = () => void;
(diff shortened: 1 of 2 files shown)


Caldwell-74:
maybe the resolver could be renamed to reflect the new structure but i didnt come up with a good name and saw it as low prio

src/NetscriptPort.ts
```diff
@@ -76,9 +79,14 @@ export function peekPort(n: PortNumber): PortData {
...
+function nextWritePort(n: PortNumber, ctx: NetscriptContext) {
   const { resolvers } = getPort(n);
-  return new Promise<void>((res) => resolvers.push(res as Resolver));
+  return new Promise<void>((res, rej) => resolvers.push(() => (ctx.workerScript.env.stopFlag ? rej() : res()))).catch(
+    () => {
+      helpers.log(ctx, () => `nextWrite was rejected on port ${n}`);
+      throw new ScriptDeath(ctx.workerScript);
+    },
+  );
 }
```
ZachHaber, line 90: This could also be done with an `AbortSignal` to completely clear the reference and abort the promise early, so that it doesn't wait until the next time the port is written to. The other benefit of allowing the user to pass in a signal, is that they can control the lifetime of when they want to work with the event.

```ts
function nextWritePort(n: PortNumber, signal: AbortSignal) {
  const { resolvers } = getPort(n);
  return new Promise<void>((res, rej) => {
    resolvers.push(res as Resolver);
    signal.addEventListener(
      "abort",
      () => {
        rej(signal.reason);
        const resolverIndex = resolvers.indexOf(res);
        if (resolverIndex !== -1) resolvers.splice(resolverIndex, 1);
      },
      { once: true }
    );
  });
}
```

It could be expanded using the context and an AbortSignal on the context to reject the promise as well so that it would reject on the script ending. - It would help with stopping other async functionalities on script end as well.

d0sboots:
I don't think this is the right solution to the issue.

You can solve this completely without adding anything to the game: you can clear pending nextWrites in atExit, or you can check if the script is alive before executing your promise chain. Since there are cases where you want listeners to persist *after* script death, I don't think we should disallow that totally.

Something like ZachHuber's approach could be useful to help players cancel listeners, but I would be very careful with the implementation because we do not want to bloat memory usage of nextWrite for the common case. (This is also an issue with your impl.)

We could add an additional array of promises, storing the returned promise, and then a `cancel NextWrite(n, promise)` would cancel that specific promise. Although I really don't think such find-gained control is commonly needed, instead of `clearPort`.

ZachHaber:
```ts
function nextWritePort(n: PortNumber, signal?: AbortSignal) {
  const { resolvers } = getPort(n);
  return new Promise<void>((res, rej) => {
    resolvers.push(res as Resolver);
    signal?.addEventListener(
      "abort",
      () => {
        rej(signal.reason);
        const resolverIndex = resolvers.indexOf(res);
        if (resolverIndex !== -1) resolvers.splice(resolverIndex, 1);
      },
      { once: true }
    );
  });
}
```
I had meant to make it optional - which should keep things the same for memory usage unless you pass in the signal manually.

Though there are probably a couple other things that would be needed in order to fully do it - like stopping the listener when it resolves as well.

d0sboots:
Using a full-blown event listener seems like a bit of overkill when all you want is a simple callback, but I guess it makes the API simple. I don't think I'd reject the promise, though - I see this more as cancellation rather than rejection, and rejection means adding a reject handler to the promise chain that you otherwise don't need

Caldwell-74:
from a quick test in node with 400k nextWrites (instance cap) created
currently 87.66 MB/ 69.009ms
mine 207.05 MB/ 102.363ms
eventlistener 476.26 MB/ 1.313s

iam not sure what usecases that would be where we would want to keep nextWrites alive after a script is dead 
even more because for me that would fall way more into the realm of "player has to figure that out/handle on their own"

and i dont think i have to tell you how many bugs this "side effect" caused so far in player scripts...

edit.: exact mem usage values from test

d0sboots:
Right now, ports can *entirely* be used without a ns context, once you have the port handle itself. This can be useful in UI scripts (for instance) which do react stuff and communicate while being "dead".

I'm all for improving the common case, but I still don't understand why the solution to this is more complicated than putting `clearPort` in `atExit`.

Edit: Part of my resistance comes from the fact that this is a general problem. Running callbacks after they are no longer relevant happens *all the time*, in every corner of CS. Learning to deal with that is a fundamental skill.

Caldwell-74:
clearPort explicity doesnt clear resolvers it even checks if resolver.length >0 and then doesnt delete the port object

sure that was another behaviour i wanted to adress maybe with an optional param to pass in
but what happens with the awaited promises when their resolver are gc'd? 
are they gc'd aswell?

my plan was to then with that param to iterate resolvers and reject them all

Caldwell-74:
https://gist.github.com/Caldwell-74/69ed8c93c356d98eec15befbc72a79b6

the scripts used for the test

d0sboots:
OK, that's an issue then - there *should* be an easy way to clear all listeners, *without* running them.

If a promise has no references it will be gc'd, same as every other object. You just need to be careful when accounting for references - the res/rej functions keep a reference to the promise, for instance.

LJNeon:
I have to completely agree with d0sboots here, both of the mentioned solutions seem to create more complexity and problems than what it's trying to solve. I also really like the idea of listeners being wiped on clearPort, that seems like a much better way to go about this.

Caldwell-74:
what problems would be created here?
just clearing on portClear only solves the zombie nextWrite if the player uses portClear which would decrease the needed functions for that by half sure but it wouldnt fix the bugs for people that dont use the atExit at all
and introduces the possibility of scripts that await nextWrite, are alive but without a way to resolve their promise thats not good
to prevent that we would have to actually handle the promises

edit.: even just storing the reject in the resolver array with the resolvers increases the test to 134.18 MB / 80.704ms

PR #1021, opened 2024-01-07
AUG: Silhouette-faction augmentation for CEO-based compensation.

jeek:
New augmentation that increases pay for any work based on how many active CEO titles the player has.

### DIFF
--- a/src/Work/Formulas.ts
+++ b/src/Work/Formulas.ts
@@ -3,7 +3,7 @@ import { Crime } from "../Crime/Crime";
 import { newWorkStats, scaleWorkStats, WorkStats, multWorkStats } from "./WorkStats";
 import { Person as IPerson } from "@nsdefs";
 import { CONSTANTS } from "../Constants";
-import { ClassType, FactionWorkType, LocationName } from "@enums";
+import { ClassType, FactionWorkType, LocationName, AugmentationName } from "@enums";
 import {
   getFactionFieldWorkRepGain,
   getFactionSecurityWorkRepGain,
@@ -19,6 +19,7 @@ import { serverMetadata } from "../Server/data/servers";
 import { Company } from "../Company/Company";
 import { CompanyPosition } from "../Company/CompanyPosition";
 import { isMember } from "../utils/EnumHelper";
+import { PlayerOwnedAugmentation } from "../Augmentation/PlayerOwnedAugmentation";
 
 const gameCPS = 1000 / CONSTANTS.MilliPerCycle; // 5 cycles per second
 export const FactionWorkStats: Record<FactionWorkType, WorkStats> = {
@@ -119,7 +120,7 @@ export const calculateCompanyWorkStats = (
   const gains = scaleWorkStats(
     multWorkStats(
       {
-        money: companyPosition.baseSalary * company.salaryMultiplier * bn11Mult * currentNodeMults.CompanyWorkMoney,
+        money: companyPosition.baseSalary * company.salaryMultiplier * bn11Mult * currentNodeMults.CompanyWorkMoney * ((Player.augmentations.map((aug:PlayerOwnedAugmentation)=>{return aug.name}).filter(aug=>aug==AugmentationName.EmpathySuppressor).length==1) ? Math.pow(10, Object.values(Player.jobs).filter((job:string)=>job=="Chief Executive Officer").length) : 1),
         hackExp: companyPosition.hackingExpGain,
         strExp: companyPosition.strengthExpGain,
         defExp: companyPosition.defenseExpGain,
--- a/src/Work/Formulas.ts
+++ b/src/Work/Formulas.ts
@@ -120,7 +120,18 @@ export const calculateCompanyWorkStats = (
   const gains = scaleWorkStats(
     multWorkStats(
       {
-        money: companyPosition.baseSalary * company.salaryMultiplier * bn11Mult * currentNodeMults.CompanyWorkMoney * ((Player.augmentations.map((aug:PlayerOwnedAugmentation)=>{return aug.name}).filter(aug=>aug==AugmentationName.EmpathySuppressor).length==1) ? Math.pow(10, Object.values(Player.jobs).filter((job:string)=>job=="Chief Executive Officer").length) : 1),
+        money:
+          companyPosition.baseSalary *
+          company.salaryMultiplier *
+          bn11Mult *
+          currentNodeMults.CompanyWorkMoney *
+          (Player.augmentations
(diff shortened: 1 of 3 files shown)


src/Work/Formulas.ts
```diff
@@ -119,7 +120,18 @@ export const calculateCompanyWorkStats = (
...
+          companyPosition.baseSalary *
+          company.salaryMultiplier *
+          bn11Mult *
+          currentNodeMults.CompanyWorkMoney *
+          (Player.augmentations
+            .map((aug: PlayerOwnedAugmentation) => {
+              return aug.name;
+            })
+            .filter((aug) => aug == AugmentationName.EmpathySuppressor).length == 1
+            ? Math.pow(10, Object.values(Player.jobs).filter((job: string) => job == "Chief Executive Officer").length)
```
Snarling, line 133: Am I reading this right that it's a 10x boost per CEO position, e.g. 3 CEO positions = 1000x salary?

My initial reaction is that this is probably overtuned. Certainly open to other feedback on this.

src/Work/Formulas.ts
```diff
@@ -119,7 +120,18 @@ export const calculateCompanyWorkStats = (
...
+        money:
+          companyPosition.baseSalary *
+          company.salaryMultiplier *
+          bn11Mult *
+          currentNodeMults.CompanyWorkMoney *
+          (Player.augmentations
+            .map((aug: PlayerOwnedAugmentation) => {
+              return aug.name;
+            })
+            .filter((aug) => aug == AugmentationName.EmpathySuppressor).length == 1
```
Snarling, line 132: This all can be replaced with `Player.hasAugmentation(AugmentationName.EmpathySuppressor, true)`

Your current code iterates the entire augmentations array 2 times over, where hasAugmentation or using your own .find or .some will only iterate through it once partially (stopping as soon as a match is found).

Snarling (changes requested):

src/Work/Formulas.ts
jeek: You have to become a CEO somewhere, join Silhouette, buy the Aug, install, and then become a CEO again before you ever see any use from it. With the time investment involved and what you have to not be doing in order to get those CEO spots, the balancing seems about right, and it brings back company work as something worth doing, *eventually*. But then you still have to gain rep to go about getting your augs and whatnot afterwards. I suspect that it may actually be too weak, but have been wracking my brain for months for *something* to give Silhouette, which is a pain in the butt to join, and has one weak unique aug and the rest are shared with other factions.

I initially thought even the existence of this aug would be gated behind a BN prize, but hydro seemed to think it'd be fine as just a regularly available augment.

jeek: BN idea: Thoughts on reducing hack to 0 like in BN8, and then having a new aug in Silhouette that multiplies job salary by 10^(number of CEO positions you have) ?
hydroflame: I think that could be a regular job tbh.
jeek: ?
hydroflame: regular aug *

Snarling: 10^number of CEO positions isn't just "makes company work worth doing eventually," it is just a total balance destroyer. If company work is ever worthwhile whatsoever, then being able to possibly get e.g. millions of times the return from it is just not in line with what augments do. There's no other augment that affects something with that kind of multiple, the highest is like a 5x total multiplier for a parameter. Even with just one CEO spot, this aug would be doubling the highest existing multiplier, and it scales exponentially from there (2->100x, 3->1000x).

A lategame aug that scales really well should not be out of the question, but to me 10^x just seems so far from how anything else in the game is balanced. And with it being locked behind a lategame setup, it does nothing to make the underlying mechanic fundamentally less weak for the normal portion of gameplay (if the problem is that company work is weak).

But I'd love to get feedback from some other folks, maybe I'm missing something here.

d0sboots: I agree, it feels like a better solution would be to make Silhouette less awful to get into, and then it wouldn't feel weak to have a softer buff. Incredibly strong buffs are incredibly hard to balance, they're always on a knife's edge of either being useless or game breaking.

Dodn: Exponential in num CEO positions seems a bit too much, yeah. Maybe the boost could be something like calcMult(charisma) * numCEOpositions, where calcMult() is similar to calculateIntelligenceMult(). Would scale decently (which can be balanced in calcMult) and create some more use for charisma as a stat.

Alpheus:
So I've been looking at this and playing it locally. It's not super fun during normal gameplay.
1) it takes way too long to acquire the aug (assuming grafting is not intended)
2) once you have the aug, the CEO positions take ages to farm
3) once you have a lot of CEO positions it becomes ridiculously OP but it feels like an unnecessary cliff, or even risk to the game, rather than a fun loop

Here's my suggestion:
- [ ] Rather than an augmentation, let's make this the "power" of BN11, awarded to the player permanently after Bn11.3. Big economic crash, everyone seeks to make money hacking and plain old jobs.
- [ ] Merge the BN11 bitnode bonuses into this power, growing for each BN11 completion up to 11.3
- [ ] Rather than just a CEO title, each title held at a company should gradually increase all other profits. We should add a UI screen somewhere to aggregate the gains visually. 
- [ ] Split the bonuses to horizontal and vertical. Horizontal for each additional company where you hold a title. Vertical for highest title for each org. Horizontal bonuses should award an incremental bonus for all rep gain in companies (not factions) to make promotion easier. Vertical bonuses should increase salary across all jobs (exponentially, multiplicatively or otherwise).
- [ ] (optional) As Silhouette is a crime megacorp, we could increase the bonus for the player's Karma. This may also make more sense in BN11
- [ ] (optional) Add a multi-tier aug similar to Stanek's where the impact of the bonus grows.
- [ ] (optional) Rather than awarded by default after BN11, could setup an "acceptance" similar to stanek somewhere on the map (Node 31 in Aevum?) that requires negative karma and a non-trivial title at a job.

The reward should be to infiltrate and corrupt multiple corporations gradually, rather than holding CEO titles that you acquire slowly one at a time.

Thoughts? @jeek @d0sboots

d0sboots:
I was just bounced back here after Jeek and I talked about it in Discord some. I'm amused to see that my off-the-cuff impression there exactly matches the comment I left here over a year ago.

To @Alpheus: Your playtesting matches what my instincts were telling me. Improving things like Silhouette is a good goal, but this is not the way.

Building out BN11 to be more interesting also has appeal. Ofc, it would be a similar amount of work as building a new BN, with the additional overhead of changing something that exists (as opposed to building something new). I think a necessary part of that, if it were to be undertaken, would be a way to make working for corporations more interesting - right now it is just a time grind and there's not much going on.

PR #1022, opened 2024-01-07
PORTS: One promise, one resolver (test)

Snarling:
Store and reuse the nextWrite promise when it is created. One promise only needs one resolver.

Will not merge if there are behavioral changes. Testing to see if there are measurable performance/memory improvements.


d0sboots:
Promises chained off of the new singular promise should resolve in FIFO order, so this should have equivalent semantics.

PR #1023, opened 2024-01-07
API: Modify the port opening functions to return a bool instead of throwing an error

Hydrogeniouss:
Modified the opening functions and nuke function to return booleans based on their success instead of simply throwing an error and crashing the script, should make the whole procress of writing a nuking script simpler.

### DIFF
--- a/src/NetscriptFunctions.ts
+++ b/src/NetscriptFunctions.ts
@@ -606,10 +606,12 @@ export const ns: InternalAPI<NSFull> = {
       return true;
     }
     if (!Player.hasProgram(CompletedProgramName.nuke)) {
-      throw helpers.makeRuntimeErrorMsg(ctx, "You do not have the NUKE.exe virus!");
+      helpers.log(ctx, () => "You do not have the NUKE.exe virus!");
+      return false;
     }
     if (server.openPortCount < server.numOpenPortsRequired) {
-      throw helpers.makeRuntimeErrorMsg(ctx, "Not enough ports opened to use NUKE.exe virus.");
+      helpers.log(ctx, () => "Not enough ports opened to use NUKE.exe virus.");
+      return false;
     }
     server.hasAdminRights = true;
     helpers.log(ctx, () => `Executed NUKE.exe virus on '${server.hostname}' to gain root access.`);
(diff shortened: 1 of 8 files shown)


ZachHaber:
I currently have my nuking script using try/catch for determining which files I have. Running the port opener functions without the file doesn't trigger the script death, just throws an error.

```js
  const portOpeners = [
    { file: 'BruteSSH.exe', fn: ns.brutessh },
    { file: 'FTPCrack.exe', fn: ns.ftpcrack},
    { file: 'relaySMTP.exe', fn: ns.relaysmtp },
    { file: 'HTTPWorm.exe', fn: ns.httpworm },
    { file: 'SQLInject.exe', fn: ns.sqlinject },
  ].filter((opener) => {
    try {
      opener.fn(homeName);
      return true;
    } catch (error) {
      return false;
    }
  });
```

Hydrogeniouss:
I am aware that people often use try { } catch { } to avoid the error thrown by the port opening functions, my own nuking function does it too.
```js
export function openPorts(ns, target) {
  [ns.brutessh, ns.ftpcrack, ns.relaysmtp, ns.httpworm, ns.sqlinject, ns.nuke].map(x => { try { x(target) } catch { } })
}
```
But leaving the catch block empty is generally seen as bad practise

d0sboots:
I'm torn on this: on one hand, I feel like these functions are probably better returning bool, since trying when it's not possible is not an "exceptional" situation. But on the other hand, this is absolutely a breaking change (although it probably wouldn't break most people). But there *is* a major version coming up...

steven-r:
> [...] But there _is_ a major version coming up...

This will have an impact on hundreds or even thousands 😁 of scripts. But they can be fixed in a way that both (exception and bool return) can be supported.

src/NetscriptFunctions.ts
```diff
@@ -593,123 +593,136 @@ export const ns: InternalAPI<NSFull> = {
...
+      }
+      if (!Player.hasProgram(CompletedProgramName.bruteSsh)) {
+        helpers.log(ctx,() => "You do not have the BruteSSH.exe program!");
+        return false;
+      }
+      if (!server.sshPortOpen) {
+        helpers.log(ctx, () => `Executed BruteSSH.exe on '${server.hostname}' to open SSH port (22).`);
+        server.sshPortOpen = true;
+        ++server.openPortCount;
+        return true;
```
steven-r: Is there a reason that you moved the `return true;` inside the if/else block? From visibility point of view I liked the `return true;` as the 'closer' of the function

ZachHaber:
I was just looking at converting my script to support either boolean or errors. I get a Typescript error due to checking `void` for truthiness. This would require using a ts-ignore, ts-expect-error, or a typecast `as unknown as boolean` to work, which is less than ideal.

```ts
// TS-Error: An expression of type 'void' cannot be tested for truthiness.ts(1345)
return ns.sqlinject.fn('home') ?? true;
```

Snarling:
This change is probably a good candidate for something to change at 3.0. I'll try to put together a planned 3.0 changes document soon so we can start building a list of intended 3.0 API changes.

PR #1026, opened 2024-01-09
EDITOR: fix disposing models

Caldwell-74:
the unmount of the editor didnt dispose the models correctly 
could be that this was already the case before the URI changes and only got visible after

added a few more checks that if a model already exist that its loaded instead of creating a new one

and removed an imho annoying toTerminal call when closing and saving a script
it will still go to the terminal when its the last script that gets closed

the extra dispose and checks might be hard to verify that they work but i disabled the dispose on unmount and tested each of them alone

### DIFF
--- a/src/ScriptEditor/ui/Editor.tsx
+++ b/src/ScriptEditor/ui/Editor.tsx
@@ -42,7 +42,7 @@ export function Editor({ onMount, onChange }: EditorProps) {
     // Unmounting
     return () => {
       subscription.current?.dispose();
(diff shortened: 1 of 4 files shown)


Caldwell-74:
extracted the model logic 
that way if we ever want to support more file extenstion we have one place to change the logic

Caldwell-74:
we could keep the models of openScripts around so we dont have to recreate them not sure how heavy that computational wise is

Snarling:
Nice work! I will get this built on web/steam shortly

Issue #1012, opened 2024-01-02
getAugmentationGraftTime returns incorrect value


fasmat:
I've been experimenting with grafting and I am working on a script that automates it for me.

It seems like `getAugmentationGraftTime` returns incorrect values. If I sleep for the value returned and then start a new graft it will result in the current graft being cancelled at 98 or 99 percent. There seems to also be no way to check if there is already a graft in progress via scripting which would be a way around the issue. I could probably fix this myself with access to the singularity API (that I haven't unlocked yet) but I feel like this should be possible without?

Alternatively would it make sense to have a promise that can be awaited for the graft to be finished? In case it shouldn't be too easy without access to singularity.

cigarmemr:
Say T is the time returned by `getAugmentationGraftTime()`. In case you:
- have NOT installed the augment "Neuroreceptor Management Implant"
AND 
- are NOT focused on the grafting process (clicked "Do something else", or set `focus` to `false` in `graftAugmentation()`)

Then the actual time required for grafting will be amplified by about 25% (to 125%*T) , as the "speed of grafting" is reduced by 20% when the two forementioned conditions are met. This is usually why script-held grafting processes are frequently interrupted by each other.

With singularity, you can easily go through this by either:
[...]

PR #1027, opened 2024-01-09
GRAFTING: Add Check for Grafting

TheAimMan:
Potential fix for https://github.com/bitburner-official/bitburner-src/issues/1012

Current commands for grafting only return how long it will take to graft with perfect focus.  Any deviation from that will cause the sleep to fire off too soon.  If the player does not have access to SF4 they can not tell if they're still grafting or not.

Figured an easy solution would just be to add a command to let the player know if they are grafting or not.  Then they can wait until that condition is false and act on that information, instead of attempting to sleep for a very specific amount of time.

*NOTE: I ran 'npm run doc' and GitHub doesn't show that I can commit anything new :(.  I seem to always fail at this part of the PR.

### DIFF
--- a/src/NetscriptFunctions/Grafting.ts
+++ b/src/NetscriptFunctions/Grafting.ts
@@ -52,6 +52,18 @@ export function NetscriptGrafting(): InternalAPI<IGrafting> {
       return graftableAugs;
     },
 
+    isGrafting: (ctx) => () => {
+      checkGraftingAPIAccess(ctx);
+      let grafting = false;
+      const work = Player.currentWork;
+      if (work) {
+        if (work.type == "GRAFTING") {
+          grafting = true;
+        }
+      }
+      return grafting;
+    },
+
     graftAugmentation:
       (ctx) =>
       (_augName, _focus = true) => {
--- a/src/NetscriptFunctions/Grafting.ts
+++ b/src/NetscriptFunctions/Grafting.ts
@@ -54,14 +54,8 @@ export function NetscriptGrafting(): InternalAPI<IGrafting> {
 
     isGrafting: (ctx) => () => {
       checkGraftingAPIAccess(ctx);
-      let grafting = false;
-      const work = Player.currentWork;
-      if (work) {
-        if (work.type == "GRAFTING") {
-          grafting = true;
-        }
-      }
-      return grafting;
+      const work = Player.currentWork ?? false;
+      return work && work.type === "GRAFTING";
     },
 
     graftAugmentation:
--- a/src/Netscript/RamCostGenerator.ts
+++ b/src/Netscript/RamCostGenerator.ts
@@ -380,6 +380,7 @@ const grafting = {
   getAugmentationGraftTime: 3.75,
   getGraftableAugmentations: 5,
   graftAugmentation: 7.5,
+  isGrafting: 3.75,
 } as const;
 
 const corporation = {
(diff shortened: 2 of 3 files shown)


src/NetscriptFunctions/Grafting.ts
```diff
@@ -52,6 +52,18 @@ export function NetscriptGrafting(): InternalAPI<IGrafting> {
       return graftableAugs;
     },
 
+    isGrafting: (ctx) => () => {
+      checkGraftingAPIAccess(ctx);
+      let grafting = false;
+      const work = Player.currentWork;
```
daniel-stickerapp: You can simplify and consolidate the conditional logic like this.

```suggestion
    const work = Player.currentWork;
    return work && work.type === "GRAFTING";
```

daniel-stickerapp (changes requested):

src/NetscriptFunctions/Grafting.ts
TheAimMan: While that simplifies it, it does give more information away to the player.  They could use it to tell if the player is just busy, and not just grafting (something only available in the singularity API)
isGrafting is null when not doing any work
isGrafting is false when doing work that isn't grafting 
isGrafting is true when Grafting
The compiler also didn't like the type of boolean, as boolean can't be null.  Had to change it to an 'any' type.  No idea if that's a bad thing.

TheAimMan: I've changed the work declaration line to return false if null
`const work = Player.currentWork ?? false;`
Thanks for the suggestion

d0sboots:
IMO it would be simpler (and more consistent) to just remove the SF4 check from `getCurrentWork`. Paying 8GB when you don't have sing doesn't seem unreasonable.

> *NOTE: I ran 'npm run doc' and GitHub doesn't show that I can commit anything new :(. I seem to always fail at this part of the PR.

`npm run doc` adds files as staged to-be-committed, so they won't show up with `git diff` (with default options). They will show up with `git status`.

TheAimMan:
Ha!  I just created a PR to lock it behind the singularity functions specifically because its in the singularity API.  :)  It already had the check bypassed, but that seems like a poor fix as the only way any player would know they could use it would be to look at the code. Otherwise they will assume (rightfully so) that they don't have access to it.  

https://github.com/bitburner-official/bitburner-src/pull/989

The best fix would be to pull it out of singularity completely but then that's a breaking change.

d0sboots:
Did you link the right PR? #989 changed types but not the check.

If we change the docs, then people wouldn't need to look at the code, but I agree that it's still harder to find. We could also just copy (alias) the function, until it is removed from sing (if we want to go that way).

TheAimMan:
Nope, totally not the right PR. This is the one.

https://github.com/bitburner-official/bitburner-src/pull/971

I'm fine with basically replicating the sing function as being a base function.  I just figured adding this function would rock the boat less.  I also figured duplicating a function would be frowned upon :)

Zelow79:
I think adding more grafting specific options is a great idea. Maybe something like how `await ns.getPortHandle().nextWrite()` works, where sleeping is built into it.

Zelow79:
Oh, another idea, but what about making `graftAugmentation` return a promise? Perhaps not ideal for everyone and likely API breaking, but imo would make a lot of sense.

d0sboots:
I like the idea of having more promise-based APIs. Making `graftAugmentation` return a promise would be a bit too much of a breaking change (synchronous vs. asynchronous workflows, plus either you lose out on immediate signaling of errors, or it gets weird), but if the current PR were changed from `isGrafting` to `grafting.sleepUntilDone` (name needs some work) and returns a promise, that would be enough different from the sing API to be worth adding IMO.

Zelow79:
Yeah, or maybe even an optional parameter that accepts an object like how hack has it with additionalms.

d0sboots:
I'm gonna close this, because #1435 essentially implements this idea, but with a promise-based API.

Issue #955, opened 2023-12-02
Last 3 sleeves are less effective


hollowleviathan:
Sleeves 6, 7 , and 8 are slightly less effective than the other sleeves. You can easily reproduce this by starting a new node on an account with 8 sleeves and watching the Shock go down slower for them, or assign all sleeves to the same crime* and the success chance will be lower for the last three sleeves despite identical stats.

```
export function main(ns) {
  for (var iter = 0; iter < ns.sleeve.getNumSleeves(); ++iter) {
    ns.sleeve.setToCommitCrime(iter, "Mug");
  }
}
```

This is on v2.5.0 (b87b8b4be)

Kelenius:
Could it be because you bought them later and so they have less intelligence?

hollowleviathan:
Sleeves have a hidden Intelligence stat as well?

Yes, I see the intelligence stat via ns.sleeve.getSleeve(). Well, that explains it.

hollowleviathan:
Perhaps Sleeves should display their intelligence stat if the player has Source-File 5?

Snarling:
> Perhaps Sleeves should display their intelligence stat if the player has Source-File 5?

@hollowleviathan Added via #1030

PR #1030, opened 2024-01-10
UI: show sleeves int stat if appropriate

Caldwell-74:
currently the sleeve int stat is only available on the Sleeve object from the getSleeve function
this change adds it to the sleeve overview and the more stats overview when in bn5 or if SF5 Level >0
grafik
grafik

there are probably cleaner solutions if so let me know!


cigarmemr:
A solution to #955

PR #1031, opened 2024-01-10
SERVERS: upgradable cores for pserver

Caldwell-74:
first of i would like to reuse the existing pserver functions to handle core upgrades aswell instead of making new once
i would like to not make a breaking change here, that would affect way too many scripts/players

so i want to default the cores params to the current cores of the servers but
currently there are 3 layers
1. player facing just some type checking
2. first "internal" here the values are checked against caps and current values
3. cost calculation

if i want to do a default on layer 1 i would need to drag the server stuff from layer 2. to 1.

i could default it to 1 on layer 1 and then just max  it later on layer 2 but that would mean no error message when the player input a value that is lower than current

or pass in the cores untested into layer 2 and do all the validation there
or make new functions 

would like opinions on it

edit.:
i choose the path of defaulting to 1 and then maxing it with the current cores
so people wont get an error when they input a valid ram but a lower core count
thats not perfect but i think the best of the options we have with the 4 different player facing functions that have 3 different validations (ram + cores, new hostname + ram + cores, existing hostname + ram +cores)

i also changed the input to positiveInteger that wont be catched and will throw errors if 0 not an number is inputed for ram and cores

another option would have been to extract all the input validation into a new helper function but it got so convoluted with testing all the three different variants that i didnt think that would be a good solution


src/Server/ServerPurchases.ts
```diff
@@ -17,46 +17,50 @@ import { workerScripts } from "../Netscript/WorkerScripts";
...
   if (isNaN(sanitizedRam) || !isPowerOfTwo(sanitizedRam) || !(Math.sign(sanitizedRam) === 1)) {
     return Infinity;
   }
-
   if (sanitizedRam > getPurchaseServerMaxRam()) {
     return Infinity;
   }
 
   const upg = Math.max(0, Math.log(sanitizedRam) / Math.log(2) - 6);
-
```
jjclark1982, line 34: This would be a good place to verify that `cores` is a positive integer.

Caldwell-74: oh thats actually what i meant with "layer 3" 
its only called once from outside the layer chain i described and that are the tech stores where i set it to 1
layer 1 and 2 should do the validation and what ever gets passed into that should be valid imo

jjclark1982: Ah, in that case the layer 1 checks could use `helpers.positiveInteger` instead of `helpers.number`

TheAimMan:
What is the reasoning for not adding 2 new functions
getPurchaseServerCoreUpgrade
upgradePurchaseServerCores
Not to say that your solution isn't more elegant, but it would follow with existing convention.  For all the other server actions (hacknet/home) they each have individual options to upgrades cores/ram/level/cache.  I'm not sure mixing in a different format would be the best option for newer players to get their head around.  Home and Hacknet are upgraded this way, Pservers are upgraded a different way.

Caldwell-74:
mainly because i think it would be wierd to not be able to set higher cores on purchaseServer
and adding it there as optional but then making a new function for upgrading also didnt seem right for me

Zelow79:
What's the progress on this PR? I am very intrigued. This is a great idea.

PR #1035, opened 2024-01-13
MISC: HGWOptions accepts non integer

Caldwell-74:
currently the HGW Option only accepts positive Integers which makes this option close to useless
because instead of using it to decrease threads used you could always just respawn the script with fewer threads

with this non integers are accepted and used allowing for a more fine tuned outcome of hacks in more advanced hacking scripts
the player would still pay the full ram cost

also added a few more types/type guards
if they are ok i would also switch the PositiveInteger type of threads to PositiveSafeInteger later


Snarling:
One potential issue I see is that the security increase from `grow` has a floor that is the smallest whole number of threads needed to get the amount of growth that actually occurred (numCycleForGrowthCorrected). To actually allow running a grow as 0.2 threads but still do a security increase as if it was 1 thread is weird, especially since it makes the 3 functions inconsistent WRT how they handle security for noninteger threads.

Also just not sure if I'm onboard with the change in general, but I'm interested to see other people's feedback on that.

Caldwell-74:
isnt the numCycleForGrowthCorrected a ceil of how much security can be increased by?

but yeah i (wrongly) assumed that it would only be used for hack because there is no benefit when lowering strength of grow or weaken to be made

for feedback of others i did ask for some here and there on the discord but havent pinged the usual for it yet
maybe i should

d0sboots:
I think this could be fun and useful, although it only affects edge cases. In other words, a good nerd snipe for us optimizers. XD

I think this would be useful for `grow` too, but only if `grow` could actually make use of the fractional threads properly.

Caldwell-74:
image
fixed the floor in grow seems to work accurate now

tested with this script
https://gist.github.com/Caldwell-74/d317ee307d8584ac9159ab7e9256e9e4
started it with 3 threads

Caldwell-74:
question would be if this can move forward if formulas should also accept fractionals  and/or return fractionals
or if the ns functions to determine outcome that accept threads should

d0sboots:
So, it's more complicated than just that. I gotta eat, but I'll followup in another post.

d0sboots:
Here's the state of things, as I see it:
1. `weaken` can use fractional threads just fine. This is not helpful, since there is no use for partial weakens.
2. `hack` can use fractional threads just fine. This was the primary motivation for the PR, and can provide a slightly more optimal thread layout in cases where an almost-full hack thread would push grow and/or weaken over an integer limit.
3. `grow` can *mostly* use fractional threads. The growth effect is properly applied, and in most cases the weaken amount will use fractional threads. The relevant expression is `usedCycles = Math.min(Math.max(0, Math.ceil(usedCycles)), threads)` in `ServerHelpers.ts`. `usedCycles` is the result of `numCycleForGrowthCorrected()` and that calculation is **fundamentally** limited to integer precision currently. The `Math.min` means that in the cases where `threads` are being used to limit growth, we will have fraction fortify. However, when you over-grow, the fortify-reduction will be on an integer-threads basis. I think this is probably OK.

Other observations:
Lots of handling between hack/grow/weaken is already inconsistent. For instance, over-grow reduces the amount of server security added to only the amount of growth actually applied. However, over-hack does not - it is only limited to the maximum amount of hacking that would be needed to reduce from full to 0.
Similarly, grow gives full XP in all circumstances, while hack must drain some money to give XP.

For these reasons, I think the small inconsistency is probably fine. We could also increase the precision of `numCycleForGrowthCorrected()` (potentially up to machine precision), but I would need to rewrite it probably to make it fast enough.

Caldwell-74:
but this inconsistency of grow is already present in the current code this pr would not change it
if a 3 thread grow gets capped to 2 or a 3 thread grow, set to 2.x gets capped to 2 it stays the same

i dont really see a benefit of lowering grow overall because a cap exists 
but i might be overlooking something

d0sboots:
Assume it takes 2.6 threads to grow to full. It's the difference between a 3 thread grow, set to 2.6 and thus having a 2.6 security effect vs. a 3 thread grow capped to 2.6 which ceils to 3 and has a 3 security effect.

Edit: Like I said above, I think this is probably fine. But it's what I think you were overlooking.

LJNeon:
This seems like it does way more harm than good, it creates the possibility for a lot of obnoxious bugs and enables a complicated micro optimization that doesn't make a significant difference in exchange.
For example, floating point errors that would be caught immediately now will instead silently alter the strength of a job. Overall a whole class of calculation mistakes that lead to non-integer results will no longer error, making them much more difficult to debug.

d0sboots:
> Overall a whole class of calculation mistakes that lead to non-integer results will no longer error, making them much more difficult to debug.

I don't see how this follows. Normally, people don't use the threads part of `HGWOptions` at all, so it is currently catching no errors. People pass integer threads to **exec**, but that won't be changing.

LJNeon:
I guess I forgot how uncommon permanent workers are outside of BN12 now that performance improvements have reduced their benefits. With those it's pretty common to utilize that argument heavily.

LJNeon:
I guess I forgot how uncommon permanent workers are outside of BN12 now that performance improvements have reduced their benefits. With those it's pretty common to utilize that argument heavily.
That does reduce the amount of players who would be negatively affected by this quite a lot considering how few people grind BN12.

d0sboots:
Huh, I never considered using the threads option with permanent workers. Any time you're reducing threads you're leaving RAM unused, so it's always suboptimal (without the new non-integer options, which opens up new possibilities).

Ofc, this is why I generally haven't considered permanent workers much in general: It adds a lot of complication to get the same level of benefit.

LJNeon:
At least in BN12 it's not worth the time it takes to respawn workers, you literally make more money by using the thread argument (to an extent, eventually it does become worth a respawn).

Caldwell-74:
permanent workers doesnt mean they are using the HGWOptions currently 
and from what i read on the discord is that atleast AimMan who is the highest bn12 that we know of has to reduce its hack strength by weakening its stanek boost to be able to hack less than 100% with one thread

existing scripts would not behave differently with this pr because if they use that Option they already floor/ceil/round w/e the input 
and for high lvl players/high bn12 and the likes this would actually enable better gameplay

edit.: i would actually rename it to strength or smt instead of threads to make it clearer that they are different but that would actually be a breaking change

LJNeon:
Yea I realized the amount of people affected negatively by this is much smaller than I originally thought, and yes not all BN12 players will even use the option (although not sure why you focused only on hack, the one least likely to need it as you found out). I think you didn't understand what I said initially though based on your second paragraph, the whole point is that it creates room for a lot of bugs that aren't possible currently not that it will go wrong when done correctly - obviously it won't.

And thinking about it more I still really don't like this change even ignoring the BN12 issue. It's just such a small optimization that adds so much complexity that I don't see as necessary or worth it. Is someone really going to make significantly more money by utilizing this? I'm also not sure how many people will actually take advantage of this, it can probably be counted on two hands.

Caldwell-74:
hack is the task that profits the most from this
fractional hacks in a batch that allow you to not have to increase the grow threads so you can actually use more optimal thread ratios would be the most common one mainly found in high end batchers
but reducing hacks strength is also something people want to do with scripts like eht where the same thread count for hack grow and weaken can cause issues, finer adjustmens here help people
even more because they currently have to floor/ceil the threads and then have to recalculate the actual money that will be stolen
this would make the calculation easier because you can actually use the fractional you get when calculating from hackPercent/hackAnalyze
and for bn12 its hack thats a big problem because when you need more than 1 grow thread to compensate you add hundreds of weaken threads to each batch
currently there is no way then deactivating stanek to keep your hack in check and thats rather bad imo

i dont see the bugs you are talking about maybe you could name a few?

LJNeon:
> For example, floating point errors that would be caught immediately now will instead silently alter the strength of a job. Overall a whole class of calculation mistakes that lead to non-integer results will no longer error, making them much more difficult to debug.

To quote my original message. And can I see an example of how much this can actually affect profit? It looks like it would be a micro optimization that only provides a slight improvement, but is complicated to implement at the same time so few people will end up using it.

Caldwell-74:
floating points are only a problem when they use fractional stuff
currently everyone has to round it in some way and when they want to use fractional with this than they have to deal with floating points thats just the nature of it
its not complicated to implement its just passing in the unrounded value they get from calculating hack effects already
and when i look at the responce from the bn12 guys its very needed there
i have no data on how much profit would increase but that isnt needed at all here
the discussion isnt to buff hacking gains or nerf it but to give players more controll over the outcome of hack

LJNeon:
To clarify, I'm talking about bugs. Expecting everyone to always write code correctly is very unrealistic.

And the impression I was given from reading the conversation was that this was intended as an optimization for batching. No one seemed to think of BN12 until I mentioned it. I was trying to understand the original reasoning for this PR and instead got a new reason I was already aware of. It doesn't actually change anything I said, it's a slight difference that not many people would actually use. It sounds like they already have solutions in BN12, like modifying stanek.

Snarling:
> 3\. [...] The `Math.min` means that in the cases where `threads` are being used to limit growth, we will have fraction fortify

I totally missed that on my initial look. So with e.g. 2.6 threads needed:
* less than 3 threads specified, only pay for the threads specified for security increase. 
* more than 3 threads specified, pay for 3 threads of security increase.

I don't really have an issue with it working this way. On my initial look, I thought in the first case, it would still make the player pay for 3 threads of ram usage even though they specified and needed fewer threads than that.

My understanding is now:

1. Anyone who is not using HGWOptions to override threads will not be impacted.
2. Anyone who is using HGWOptions to override threads with an integer value will not be impacted.
3. The change just allows a slight optimization by allowing players to specify fractional threads if they want to, with all 3 functions scaling appropriately for the fractional threads specified.

The possible cons I see:

* It slightly raises the optimization ceiling in some cases, so a script that doesn't buy in to using fractional threads is not optimized, which soft-forces players to adopt fractional threads. This is both a pro and a con depending on how you look at it.
* It makes it simpler to target exact percentages for hack%, instead of being forced to search for an efficient number of hack threads. You can just use the exact fractional threads needed for e.g. 5% hacking, and then use nearly exact fractional threads needed for growth back to 100%. With integers, you end up needing to analyze different pairings of thread counts because the rounding makes some pairings more efficient (especially noticeable at low thread counts) - now this is totally unneeded. Again this could be a pro or a con depending on viewpoint. Edit: It's not quite as simple as I made it out to be here, since you still pay the ram cost for the entire thread even if you're only using a fraction of it.

Caldwell-74:
in BN12 the issue is very noticeable in batchers
when some extra stength of hack means you have to add a second grow thread and then 700 more weaken threads yeah thats an issue
and i originally said "more advanced hacking scripts" which yeah people in high bn12 tend to use? 

"bugs" but you havent actually said anything about a real bug yet
the only point is floating points which only occures when people want to use fractional but thats not a bug rather the nature of it

that someone who doesnt want to use fractional, inputs an integer as thread on script startup and then a non integer value into the least used option in the game and then puts the blame on the game when the documentation of that option clearly states that it accepts non ints is such a bs szenario to bring up that i get the impression that this more a personal thing for you and not a real discussion about a pr

d0sboots:
Please don't make this personal, we all have opinions and like to argue and that's fine. :D

LJNeon:
> Anyone who is using HGWOptions to override threads with an integer value will not be impacted.

I disagree with this, because currently passing a non integer value by mistake throws an error, but after this change it will silently alter behavior instead. This can lead to obnoxious to discover bugs that affect people who don't intend to use fractional thread counts. (Although the amount of people affected by this is very small, so it's debatable how much this actually matters)

> It makes it simpler to target exact percentages for hack%, instead of being forced to search for an efficient number of hack threads.

As you seem to have realized in your edit, it's not that simple because the ram usage remains unchanged and only the effects of HGW changes.

Caldwell-74:
> Anyone who is not using HGWOptions to override threads will not be impacted.

that goes for the people that are currently using it because their logic would fail if not
for new people its the question how likely it is that someone who doesnt want to use fractional threads inputs a non rounded value for it when they used a rounded value in the exec/run
but even then if they expect the hack to only be 2 threads strong  but input 2.4 is it really the game fault is that really something we should encourage? floatings points in themselfs arent a problem a rounded value wont suddenly get a floating point issue, the issue could be people not inputing what they want it to be-> their bad

> It slightly raises the optimization ceiling in some cases, so a script that doesn't buy in to using fractional threads is not optimized, which soft-forces players to adopt fractional threads. This is both a pro and a con depending on how you look at it.

i dont really see the force argument here because already most people dont go the full way to the most optimal batching method but just use the alot more simple shotgun method

> The change just allows a slight optimization by allowing players to specify fractional threads if they want to, with all 3 functions scaling appropriately for the fractional threads specified.

slight outside bn12, inside bn12 it can easily mean hundreds of weaken threads per batch differences

LJNeon:
Also after catching up on the conversation in endgame-general I can see this PR would be a nice QoL change for BN12 players. I still don't see any noteworthy benefit elsewhere, if I'm missing something I would appreciate an explanation.

If this change does go through I would prefer renaming it to something else as caldwell suggested at one point, "threads" could be misleading since the actual thread count of the script is unchanged.

d0sboots:
I could see renaming it to something else, but that would make it a breaking change. :(

Edit: Also, the same logic applies to it in its current integer version (the actual thread count is unchanged, it only affects the power of the operation), and no one seems to have been confused.

TheAimMan:
Since I'm a named defendant in the PR I figure I'll chime in.  I actually didn't know this PR existed and after loading my game up in BN1 and seeing that I'd need 4.8 trillion grow threads for a very boosted 'ecorp' server I coded up my own fix so I could actually batch in that BN.

https://github.com/bitburner-official/bitburner-src/compare/dev...TheAimMan:bitburner-src:hackLimiter

Mine is much more targeted as it would only apply to hack and wouldn't change the current thread behavior.  

I'm fine with either approach, and I always welcome the potential for other optimizations (this PR) but if the concern is about grow threads/weaken threads being affected by the PR the approach could be more targeted.  Regardless, I think the player should be able to limit their hack strength if needed.

PR #1036, opened 2024-01-14
PORTS: Add ns.nextPortWrite() as a safer option

LJNeon:
As discussed on discord, this fills a hole in the API and also provides a safer to use option that checks for concurrent function calls and script death.

This needs to be tested by someone but that should be very easy to do.

### DIFF
--- a/src/Netscript/RamCostGenerator.ts
+++ b/src/Netscript/RamCostGenerator.ts
@@ -544,6 +544,7 @@ export const RamCosts: RamCostTree<NSFull> = {
   peek: 0,
   clear: 0,
   writePort: 0,
+  nextPortWrite: 0,
   readPort: 0,
   getPortHandle: 0,
   rm: RamCostConstants.ReadWrite,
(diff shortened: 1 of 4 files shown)


Snarling:
Looks good to me, but some people might want it to behave differently.

Right now it will still resolve the promise even if the script died sometime between calling nextPortWrite and the next write. I know some people were saying they thought it should reject the promise in that case, which would require adding a wrapper promise in ns.nextPortWrite.

I'm fine with no wrapper promise though, if people are using the ns versions of port manipulation functions, then they will not be able to read/write from the port on a dead script anyway, so there's no real harm in allowing this promise to resolve. Also not having a wrapper should make this much more performant, especially for people who await a port write from a huge number of scripts, e.g. all workers in a batcher.

I'll wait to merge this until tomorrow in case there is some other feedback I'm missing.

LJNeon:
IMO the reject on script death is an unrelated issue that should be addressed in a separate PR. I believe this should be added regardless of how that discussion goes, and that PR can change behavior as needed.

PR #1040, opened 2024-01-17
UI: Fix location color on light background.

### LINKED ISSUE #1039: UI: locations in city render in white text regardless of background color
In the city map view, the clickable locations are all rendered in white (or whatever the theme calls "white") regardless of the background color.  For light backgrounds, this makes the locations hard or impossible to see.

### Dark background:
image
### Light background:
image

adeilt:
# UI: Fix location color on light background.

# NOTE: See PR #1079 for a better solution to issue #1039.

The locations on the city map and the cities on the world map always use the "white" color configured in the theme.  This PR uses the theme-configured "white" color when the theme's background color is dark and the "black" color when the background color is light.

Specifically, it converts the `theme.backgroundprimary` color to HSL and sets the locations to render in `theme.colors.white` when `L<50` and `theme.colors.black` when `L>=50`.

### Dark background:
image
### Light background:
image

### Dark background:
image
### Light background:
image

# Linked issues
fixes #1039
see also the followup PR #1041

### DIFF
--- a/src/Locations/ui/City.tsx
+++ b/src/Locations/ui/City.tsx
@@ -29,7 +29,8 @@ interface IProps {
 const useStyles = makeStyles((theme: Theme) =>
   createStyles({
     location: {
-      color: rgbToHsl(hexStringToRgb(Settings.theme.backgroundprimary))[2] < 50 ? theme.colors.white : theme.colors.black,
+      color:
+        rgbToHsl(hexStringToRgb(Settings.theme.backgroundprimary))[2] < 50 ? theme.colors.white : theme.colors.black,
       whiteSpace: "nowrap",
       margin: "0px",
       padding: "0px",
--- a/src/Locations/ui/City.tsx
+++ b/src/Locations/ui/City.tsx
@@ -20,6 +20,7 @@ import Button from "@mui/material/Button";
 import { Theme } from "@mui/material/styles";
 import makeStyles from "@mui/styles/makeStyles";
 import createStyles from "@mui/styles/createStyles";
+import { hexStringToRgb, rgbToHsl } from "../../utils/helpers/colorTools";
 
 interface IProps {
   city: City;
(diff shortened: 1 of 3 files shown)


adeilt:
I wasn't sure where to put the color conversion functions so I picked what seemed like a reasonable location.  If they belong elsewhere, please let me know and I'm happy to move them.

adeilt:
Found the Prettier issue.  I was pretty sure I ran it, but I think I must have run it on the wrong branch.  Fixing.

adeilt:
That should do it for the formatting.

adeilt:
Just realized that it's a very small and similar change to correct the cities on the world map as well.  I've added a commit to do that and updated the PR description.  Testing of the world map change was manual and it seems to be working as expected.

Snarling:
Should the fix for light themes not be to change the "white" and "black" theme colors to be opposite? Or are there other cases where the theme's "white" value actually needs to be a light color even with an inverted/light theme?

If we're using the configured theme to decide color for something, then we should let the theme handle making the colors have proper contrast, I don't think it makes sense to do a manual color comparison.

adeilt:
I'm using this playthrough of BB as a way to learn Typescript (and get better at software engineering in general), so I appreciate the feedback.  I'm totally willing to ruthlessly abandon this in favor of something better.

Handling the dark/light split at the theme level certainly feels more correct to me, and I'll post some discussion on #1039.  In the meantime, I'll set this to a draft and close it in at most a week if I don't see any future for it.  (Or feel free to close it if that's more appropriate.)

adeilt:
The remaining value I see here would be in the context of #1079.  The possibility I see is that the color functions could be used to have a little tag appear if the color contrast would be low on a configured theme color.

e.g. If the background color is light yellow and a text color such as `primarylight` is configured as white, then the lightness measurement I used in this PR could detect the small difference and put a little icon on the color's configuration box in the Theme Editor.  A tooltip could explain that the contrast would likely be low to prompt the user to adjust.

adeilt:
Actually, these functions could also be used to set the light/dark variant colors from the main color selected.  For example, a checkbox "automatically set light/dark variants" could be provided and, when checked, would automatically change `primarydark` and `primarylight` when `primary` is changed.

adeilt:
No more reason to keep this open; closing.

PR #1041, opened 2024-01-17
UI: Correct black color value in light theme.

adeilt:
# UI: Correct black color value in light theme.

# Documentation

The color "black" in the light theme is set the same as "white": #F7F7F7.  Instead, black should be: #000000 (or at the very least not == white).  This PR fixes that minor issue.

This is a followup to pull request #1040 and issue #1039 by making the Light theme that is selectable in the Theme browser show cities on the city map.

I poked around a bit and didn't spot any places where the color change made anything look weird or wrong.

I didn't create an Issue on Github for this because it seemed too minor; please let me know if it's preferred to have an Issue for every PR.

### DIFF
--- a/src/Themes/data/light/index.ts
+++ b/src/Themes/data/light/index.ts
@@ -29,7 +29,7 @@ export const Theme: IPredefinedTheme = {
     welllight: "#f9f9f9",
     well: "#eaeaea",
     white: "#F7F7F7",
-    black: "#F7F7F7",
+    black: "#000000",
     hp: "#BF5C41",
     money: "#E1B121",
     hack: "#47BC38",


Snarling:
This will also be covered by #1079 once that is merged in. Without the previous PR #1040 being merged in, the correct fix here isn't to make black into a black color for this theme, the fix would be to configure the theme to have a dark color for "white" (because the configured color for white is supposed to have a high contrast against the background). A light theme should not have a white-ish color assigned as the value for white, because that's not going to have proper contrast anywhere where the game uses white.

adeilt:
Agreed.  I just now realized that this one was obsolete and - in fact - misguided from the start.  I came looking for it to close it only to realize that you got to it before me.  Thanks!

PR #1042, opened 2024-01-17
UI: Kill all running scripts when entering the bitverse

### LINKED ISSUE #510: [Suggestion] stop all scripts upon backdooring worlddaemon
As the title says, I would like to suggest that backdooring the worlddaemon kills all other scripts.

I am playing bitburner on a tablet. The system frequently hangs a little bit, especially with batching algorithms. When I backdoor worlddaemon without killing all my skripts first, the bitnode screen loads extremely slow. Slow like 1 new character per 5 seconds...

I reloaded a save-game shortly prior to taking down worlddaemon, killed all scripts, backdoored WD and everything worked fine.

Obviously, in the first situation all my skripts were still running in the background, probably generating exp and money, which are useless in the bitnode-selection screen.
This should be prevented, because it introduces "exploit" options, especially for intelligence grinding. The player is usually pretty powerful  when taking down WD. Especially running intelligence-grinding scripts in the "background" of the bitnode selection screen should not be rewarded.

LJNeon:
While on the bitverse page, running scripts can cause lag or even navigate you out of it in some cases.

fixes #510

### DIFF
--- a/src/ui/GameRoot.tsx
+++ b/src/ui/GameRoot.tsx
@@ -13,6 +13,7 @@ import { InteractiveTutorialRoot } from "./InteractiveTutorial/InteractiveTutori
 import { ITutorialEvents } from "./InteractiveTutorial/ITutorialEvents";
 
 import { prestigeAugmentation } from "../Prestige";
+import { prestigeWorkerScripts } from "../NetscriptWorker";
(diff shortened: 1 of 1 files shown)


TheAimMan:
Tested the code change.  Works for b1t_flume and manually backdooring a server.

PR #1044, opened 2024-01-19
BUGFIX: Make grow formulas faster and more accurate.

d0sboots:
This also adds a test for accuracy of the formulas. It is notable that the test does *not* pass on the previous version of the code. In particular, `numCyclesForGrowthCorrected` failed occasionally but catastrophically once values got above 1.44e20.

This also gets rid of a lot of annoying duplicate code, which is good because these functions can be relied upon to return identical answers now.

### DIFF
--- a/src/Constants.ts
+++ b/src/Constants.ts
@@ -125,7 +125,7 @@ export const CONSTANTS: {
 
   // Server-related constants
   HomeComputerMaxRam: 1073741824, // 2 ^ 30
-  ServerBaseGrowthIncr: 0.03, // Unadjusted growth increment (growth rate is 1 + this * adjustment)
+  ServerBaseGrowthIncr: 0.03, // Unadjusted growth increment (growth rate is this * adjustment + 1)
   ServerMaxGrowthLog: 0.00349388925425578, // Maximum possible growth rate accounting for server security, precomputed as log1p(.0035)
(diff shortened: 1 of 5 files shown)


src/Constants.ts
```diff
@@ -125,8 +125,8 @@ export const CONSTANTS: {
 
   // Server-related constants
   HomeComputerMaxRam: 1073741824, // 2 ^ 30
-  ServerBaseGrowthRate: 1.03, // Unadjusted Growth rate
-  ServerMaxGrowthRate: 1.0035, // Maximum possible growth rate (max rate accounting for server security)
+  ServerBaseGrowthIncr: 0.03, // Unadjusted growth increment (growth rate is 1 + this * adjustment)
```
steven-r: I would suggest writing `(1 + this * adjustment)`

d0sboots: I'm not quite sure what you mean... would the whole thing would become "(growth rate is (1 + this * adjustment))", with two parens?

steven-r: I just looked at the comment --> 1 + this * adjustment is different to (1 + this) * adjustment

d0sboots: Now I'm even more confused... Are you saying you read it with the wrong order of operations initially?

Edit: Does this make it more clear?

PR #1055, opened 2024-01-23
TERMINAL: Add --ram-override flag to the run command

LJNeon:
When executing scripts programmatically there's a ram override option that can be quite useful. This pull request adds that option to the run command as a flag.

~~This still needs to be tested.~~


Semanual:
Tested, it works as intended
test.js:
```js
/** @param {NS} ns */
  while (true) {
    await ns.hack("n00dles");
  }
}
```
Script real cost: 1.7
`run test.js` runs as expected, using 1.7 GiB RAM;
`run test.js --ram-override 4` runs as expected, using 4 GiB RAM;
`run test.js --ram-override 1.2` does not run, because it's below 1.6;
`run test.js --ram-override 1.6` does run, but then dynamic checker catches it up and kills the script.
imagem
imagem

PR #960, opened 2023-12-04
REMOTE: add 'getAllServers' method to RFA

shyguy1412:
# Description

In the Discord server it was discussed that it would be useful if the RFA had access to a full list of servers + their root access status
this change adds that method to the RFA

PR #1058, opened 2024-01-24
REMOTE: add 'getAllServers' method to RFA

shyguy1412:
this is the same as PR #960 I just moved branch


LJNeon:
Seems like the check failed due to a really strange error?

d0sboots:
I tried rerunning it, but it's persistent.

LJNeon:
Wait is there a leading space in the title? That might be causing it.

shyguy1412:
Good catch! I totally didn't see that, removed now

LJNeon:
@d0sboots can you rerun it one more time?

PR #1059, opened 2024-01-24
API: make ns.atExit add the callback to an array instead of setting it

shyguy1412:
The documentation for ns.atExit says 'Add callback function when the script dies'. This has lead to confusion and unexpected behaviour for me and others as it was expected to be able to add multiple listeners.
This change adds every callback passed to atExit to an array and calls them each before the script dies to be more in line of how the documentation is worded.

### DIFF
--- a/src/Netscript/WorkerScript.ts
+++ b/src/Netscript/WorkerScript.ts
@@ -79,7 +79,7 @@ export class WorkerScript {
   hostname: string;
 
   /**Map of functions called when the script ends. */
-  atExit: Map<string, (() => void) | undefined> = new Map();
+  atExit: Map<string, () => void> = new Map();
 
   constructor(runningScriptObj: RunningScript, pid: number, nsFuncsGenerator?: (ws: WorkerScript) => NSFull) {
     this.name = runningScriptObj.filename;
--- a/src/Netscript/WorkerScript.ts
+++ b/src/Netscript/WorkerScript.ts
@@ -79,7 +79,7 @@ export class WorkerScript {
   hostname: string;
 
   /**Map of functions called when the script ends. */
-  atExit: Record<string, () => void> = {};
+  atExit: Map<string, (() => void) | undefined> = new Map();
 
   constructor(runningScriptObj: RunningScript, pid: number, nsFuncsGenerator?: (ws: WorkerScript) => NSFull) {
     this.name = runningScriptObj.filename;
--- a/src/Netscript/WorkerScript.ts
+++ b/src/Netscript/WorkerScript.ts
@@ -78,8 +78,8 @@ export class WorkerScript {
   /** hostname on which this script is running */
   hostname: string;
 
-  /** Function called when the script ends. */
-  atExit: (() => void) | undefined = undefined;
+  /**Array of functions called when the script ends. */
+  atExit: (() => void)[] = [];
 
   constructor(runningScriptObj: RunningScript, pid: number, nsFuncsGenerator?: (ws: WorkerScript) => NSFull) {
     this.name = runningScriptObj.filename;
--- a/src/Netscript/WorkerScript.ts
+++ b/src/Netscript/WorkerScript.ts
@@ -78,8 +78,8 @@ export class WorkerScript {
   /** hostname on which this script is running */
   hostname: string;
 
-  /**Array of functions called when the script ends. */
-  atExit: (() => void)[] = [];
+  /**Map of functions called when the script ends. */
+  atExit: Record<string, () => void> = {};
 
   constructor(runningScriptObj: RunningScript, pid: number, nsFuncsGenerator?: (ws: WorkerScript) => NSFull) {
     this.name = runningScriptObj.filename;
--- a/src/Netscript/killWorkerScript.ts
+++ b/src/Netscript/killWorkerScript.ts
@@ -43,9 +43,9 @@ function stopAndCleanUpWorkerScript(ws: WorkerScript): void {
   ws.delayReject?.(new ScriptDeath(ws));
   ws.env.runningFn = "";
 
-  ws.atExit.forEach((atExit) => {
+  for (const key in ws.atExit) {
     try {
-      atExit();
+      ws.atExit[key]();
     } catch (e: unknown) {
       handleUnknownError(e, ws, "Error running atExit function.\n\n");
     }
@@ -53,7 +53,7 @@ function stopAndCleanUpWorkerScript(ws: WorkerScript): void {
       // If atExit() kills the script, we'll already be stopped, don't stop again.
       return;
     }
-  });
+  }
 
   //ws.atExit was previously set to undefined after being called
   //so empty to queue to stay consistent
--- a/src/Netscript/killWorkerScript.ts
+++ b/src/Netscript/killWorkerScript.ts
@@ -43,11 +43,13 @@ function stopAndCleanUpWorkerScript(ws: WorkerScript): void {
   ws.delayReject?.(new ScriptDeath(ws));
   ws.env.runningFn = "";
 
-  for (const key in ws.atExit) {
+  for (const key of ws.atExit.keys()) {
     try {
-      const atExit = ws.atExit[key];
-      delete ws.atExit[key];
-      atExit();
+      const atExit = ws.atExit.get(key);
+      //Calling ns.exit inside ns.atExit can lead to recursion
+      //so the handler must be removed immediatly
+      ws.atExit.set(key, undefined);
+      if (typeof atExit == "function") atExit();
     } catch (e: unknown) {
       handleUnknownError(e, ws, "Error running atExit function.\n\n");
     }
--- a/src/Netscript/killWorkerScript.ts
+++ b/src/Netscript/killWorkerScript.ts
@@ -42,23 +42,25 @@ function stopAndCleanUpWorkerScript(ws: WorkerScript): void {
   if (ws.delay) clearTimeout(ws.delay);
   ws.delayReject?.(new ScriptDeath(ws));
   ws.env.runningFn = "";
+  const atExit = ws.atExit;
+  //Calling ns.exit inside ns.atExit can lead to recursion
+  //so the map must be cleared before looping
+  ws.atExit = new Map();
 
-  for (const key of ws.atExit.keys()) {
+  for (const key of atExit.keys()) {
     try {
-      const atExit = ws.atExit.get(key);
-      //Calling ns.exit inside ns.atExit can lead to recursion
-      //so the handler must be removed immediatly
-      ws.atExit.set(key, undefined);
-      if (typeof atExit == "function") atExit();
+      const callback = atExit.get(key);
+      if (typeof callback == "function") callback();
     } catch (e: unknown) {
       handleUnknownError(e, ws, "Error running atExit function.\n\n");
     }
-    if (ws.env.stopFlag) {
-      // If atExit() kills the script, we'll already be stopped, don't stop again.
-      return;
-    }
   }
-
+  
+  if (ws.env.stopFlag) {
+    // If atExit() kills the script, we'll already be stopped, don't stop again.
+    return;
+  }
+  
   ws.env.stopFlag = true;
   removeWorkerScript(ws);
 }
--- a/src/Netscript/killWorkerScript.ts
+++ b/src/Netscript/killWorkerScript.ts
@@ -56,8 +56,8 @@ function stopAndCleanUpWorkerScript(ws: WorkerScript): void {
   }
 
   //ws.atExit was previously set to undefined after being called
-  //so empty to queue to stay consistent
-  ws.atExit = [];
+  //so empty the map to stay consistent
+  ws.atExit = {};
 
   ws.env.stopFlag = true;
   removeWorkerScript(ws);
--- a/src/Netscript/killWorkerScript.ts
+++ b/src/Netscript/killWorkerScript.ts
@@ -55,12 +55,12 @@ function stopAndCleanUpWorkerScript(ws: WorkerScript): void {
       handleUnknownError(e, ws, "Error running atExit function.\n\n");
     }
   }
-  
+
   if (ws.env.stopFlag) {
     // If atExit() kills the script, we'll already be stopped, don't stop again.
     return;
   }
-  
+
   ws.env.stopFlag = true;
   removeWorkerScript(ws);
 }
--- a/src/Netscript/killWorkerScript.ts
+++ b/src/Netscript/killWorkerScript.ts
@@ -43,10 +43,8 @@ function stopAndCleanUpWorkerScript(ws: WorkerScript): void {
   ws.delayReject?.(new ScriptDeath(ws));
   ws.env.runningFn = "";
 
-  if (typeof ws.atExit === "function") {
+  ws.atExit.forEach((atExit) => {
     try {
-      const atExit = ws.atExit;
-      ws.atExit = undefined;
       atExit();
     } catch (e: unknown) {
       handleUnknownError(e, ws, "Error running atExit function.\n\n");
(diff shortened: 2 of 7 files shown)


Caldwell-74:
would it make sense to give the player the ability to remove/update a specific atExit?


edit.:
Maybe in form of ns.atExit returning the id of the set function and accepting a second optional param to specify which id/function to override

LJNeon:
Personally I would prefer updating the documentation instead to make it more clear that there's only one callback. This introduces no new functionality that can't already be implemented currently.

But if this change is made I would support a `setTimeout()` type API where `ns.atExit()` returns a unique identifier (a symbol?) that can be used in a new `ns.clearAtExit()` (I don't like that name, but it gets the idea across) function.

shyguy1412:
Right now there is no way to remove atExit once the function has been set, so I haven't considered adding it in now because I wanted to keep the change minimal. This does however open the door to a JS like change where we keep atExit the way it is and add addAtExit and removeAtExit. remove can either work by function or ID, i am impartial to that. This way the existing behaviour doesnt change tho, there are only new methods

LJNeon:
I would strongly prefer modifying the current function instead of adding bloat for the sake of backwards compatibility.

shyguy1412:
I agree. another option could be implementing a real event system. I actually kinda wanted something like that too, where you can just do ns.addEventListener, ns.removeEventListener and ns.dispatchEvent (aka make ns an event target). That way atExit can be just one of potentionally many events? Just throwing some ideas out there

Caldwell-74:
> Right now there is no way to remove atExit once the function has been set

removing currently would be just setting a null function atleast that (updating of an existing one) should be preserved imho

LJNeon:
That just seems like pointlessly overcomplicating it to me. There's zero benefit compared to the much simpler `setTimeout()` style API where `ns.atExit()` returns a symbol ID that can be used in `ns.removeAtExit()`.

shyguy1412:
Yeah, I just really like events and event targets x.x
I agree that the setTimeout style would probably be the best for the game tho
Before I add that in I would like some feedback from snarling or d0sboots, just so I know I am not wasting my time here wanting a change that will never be added

d0sboots:
I don't think this is great as-is because it's a breaking change: Everyone currently using arExit will have to update their code, and that's likely quite a lot of people. This is compounded by the fact that it's easy to miss a failing atExit handler.

However, making the type polymorphic so that it takes either a function or an array is also a bit gross. And once you can set multiple, it does start asking for the ability to remove individual functions...

I'm with LJ here that I think the solution is to fix the docs instead. Unlike setTimeout, atExit is typically set once: we don't need fancy handling for multiple callbacks. You can call it with an empty callback to remove the handler.

Caldwell-74:
giving it an optional param would make it possible to preserve the current way but still add more options
when its not passed than the function replaces the old/main atExit but you could pass in an id that allows you to store a second one /more 
internally just an object literal with a "main" prop for the "unnamed" atExit and then on exit iterating it

shyguy1412:
I think that could also be a very good solution. 
However I dont agree with just changing the docs. The way atExit works right now makes it impossible to make isolated, self contained functions and classes that clean up after themselves. There is always a need for either global state or passing all the cleanup functions to a single location so that they can all be invoked. Its incredibly tedious and annoying imo.

d0sboots:
> The way atExit works right now makes it impossible to make isolated, self contained functions and classes that clean up after themselves.

It's not particularly hard to do that, you'd just do what you're doing in this PR, but in your code instead.

Actually, I don't think this PR helps with your goal: having a single function that takes an array doesn't make it any easier to have isolated cleanup functions. You're still going to need to write your own `addAtExit` function, and it will look basically the same as without this PR.

shyguy1412:
I think you misundertood what I want to do. I dont want ns.atExit to take an array, i want successive calls to ns.atExit be able to add more than one handler. To implement a system like that right now would require some sort of global state that every function that needs cleanup has access too. 
I really think Caldwells solution is the best of both worlds. I wont change anything with how it currently works but still allows adding more than one function.
I would imagine it look something like this 

ns.atExit(callback, id="default");

It would behave exactly the same as it already does if you dont pass the second param, but if you do you can still set more listeners if you need to.

LJNeon:
I'd still consider a breaking change to be very worth it for a better implementation if we don't go the documentation route. Needing to supply your own IDs doesn't seem ideal when it can be effortlessly avoided.
Also, not everyone who uses `ns.atExit()` would be affected, only players who use it multiple times (I don't know of anyone who does that).

d0sboots:
> I think you misundertood what I want to do. I dont want ns.atExit to take an array, i want successive calls to ns.atExit be able to add more than one handler.

Yeah, I misread the code initially.

> To implement a system like that right now would require some sort of global state that every function that needs cleanup has access too.

Although you're correct, it's not like it's complicated.

```js
class GlobalState {
  /** @type {NS} */
  ns;
  exiters = [];  // Mutate this, don't replace it

  /** @param {NS} ns */
  constructor(ns) {
    this.ns = ns;
    const boundExiters = this.exiters;
    ns.atExit(() => {for (const f of boundExiters) f()});
  }
}

/** @param {NS} ns */
  const state = new GlobalState(ns);
  state.ns.tprint("In main");
  state.exiters.push(() => state.ns.tprint("Leaving main 1"));
  state.exiters.push(() => state.ns.tprint("Leaving main 2"));
}
```

`ns` is *already* global state that everyone needs access to, and there are other things in a script that are similar to it, so packaging them together in a class is natural. Then, you pass the instance of your state around instead of `ns`. It's a very small amount of code to add (especially if you already have a global state class, in which case it's literally just 2 more lines), and more importantly *everyone has the freedom to implement it the way they want*.

I'd feel differently if this were something that was complicated to implement, or that everyone was needing, but neither of those are the case.

shyguy1412:
I really think that only allowing one atExit handler is an unnecissary inconvinience that shouldn't need to be fixed by players.
I made a change with the idea Caldwell suggested, this way ns.atExit behaves exactly the same way if you dont pass an ID but you have the option too do so if you please.
This wont break anything and only add some QOL

```js
ns.atExit(() => cleanupA()) //still works exactly the same
ns.atExit(() => cleanupB(), 'cleanupB') //but you can add more if you pass id
```

src/Netscript/killWorkerScript.ts
```diff
@@ -43,11 +43,9 @@ function stopAndCleanUpWorkerScript(ws: WorkerScript): void {
   ws.delayReject?.(new ScriptDeath(ws));
   ws.env.runningFn = "";
 
-  if (typeof ws.atExit === "function") {
+  for (const key in ws.atExit) {
     try {
-      const atExit = ws.atExit;
-      ws.atExit = undefined;
-      atExit();
+      ws.atExit[key]();
```
d0sboots: You need to keep the same semantics here (empty the map immediately) to avoid infinite recursion.

I would do the same thing of setting it to undefined and having an explicit test.

shyguy1412: I dont quite see how this causes recursion, would you mind explaining?

regardless of that, instead of setting the key to undefined I think it would be better to straight up delete it then

```js
const atExit = ws.atExit[key];
delete ws.atExit[key];
atExit();
```

shyguy1412: nevermind, I think I just got it. If a script calls ns.exit inside ns.atExit that would cause the recursion right?

shyguy1412: the reason I want to delete instead of setting to undefined is that atExit is of type Record<string, () => void>, so typescript wouldnt catch that a key is undefined leading to the possibility of trying to call 'undefined' as function

Snarling:
I think for those that want handling for multiple callbacks, it's a great place for the player to add their own custom handling. If you want it to work exactly like setTimeout, a player can make their custom wrapper work exactly that way. Or if you want it to work exactly like addEventListener, you can make it work that way instead.

The string-keyed object isn't really similar to any existing / standard ways of managing multiple callbacks or handlers.

That said, the current changes in this PR don't break anything and should only introduce a very small amount of extra overhead to scripts, so I'm fine with merging this if others want these changes.

LJNeon:
So far this seems highly contested, some people want it and some don't. Although I'm very strongly opposed to the current behavior of this PR, I would be much less dissatisfied with a better implementation even if it breaks a very very small amount of scripts (might just not exist).

Caldwell-74:
i dont see a downside to this
sure a "proper" handler would be nice but that could also mean more overhead
and i think its always good to expand what the game allows/offers on possibilities for code structures especially when  it doesnt come with a downside

src/Netscript/killWorkerScript.ts
d0sboots: This still isn't going to be sufficient. Let me quote from <https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in>:

> In general, it is best not to add, modify, or remove properties from the object during iteration, other than the property currently being visited. The spec explicitly allows the implementation to not follow the algorithm above in one of the following cases:
>
> ...
> - A property is deleted from the object or its prototype chain during iteration.

There are still tricky recursive cases where things can go haywire if you delete things piecemeal like this. Also, it's slower. The only truly safe approach is to prevent *all* future iteration, immediately.

In addition, you should use a `Map` here, because iteration of Objects is very slow. The Typescript type you want is `Map<string, () => void> | undefined`.

shyguy1412: Dang, recursion really can get tricky!
Also yeah, I didn't think about performance cause I usually only work on projects where it really doesn't matter 😅

Ill make the changes you suggested and then test around a bit to make sure nothing is broken. Thanks for being so patient with this!

shyguy1412: I implemented the changes you suggested and also ran some ingame tests with the follwing script 
```js
/**
 * 
 * @param {NS} ns 
 */
  ns.atExit(() => {
    ns.tprint('AT EXIT PRE DEFAULT');
    ns.exit();
  }, 'PRE DEFAULT');

  ns.atExit(() => {
    ns.tprint('AT EXIT');
    ns.exit();
  });

  ns.atExit(() => {
    ns.tprint('AT EXIT POST DEFAULT');
    ns.exit();
  }, 'POST DEFAULT');

  ns.atExit(() => {
    ns.tprint('AT EXIT OVERRIDE TEST');
  }, 'OVERRIDE TEST');

  ns.atExit(() => {
    ns.tprint('OVERRIDDEN');
  }, 'OVERRIDE TEST');
}
```

This is the result 
firefox_lSpHzy7KAO

LJNeon:
There's no difference in the amount of overhead, one (setTimeout style) just provides IDs for you while the other requires manually chosen IDs and uses a non standard API. There's not a relevant difference in terms of code breakage either, as I don't know of anyone who actually takes advantage of how atExit overwrites the callback.

src/Netscript/WorkerScript.ts
```diff
@@ -78,8 +78,8 @@ export class WorkerScript {
   /** hostname on which this script is running */
   hostname: string;
 
-  /** Function called when the script ends. */
-  atExit: (() => void) | undefined = undefined;
+  /**Map of functions called when the script ends. */
+  atExit: Map<string, (() => void) | undefined> = new Map();
```
Snarling: Why is undefined an allowable value for this Map?

src/Netscript/killWorkerScript.ts
```diff
@@ -43,11 +43,13 @@ function stopAndCleanUpWorkerScript(ws: WorkerScript): void {
...
+  for (const key of ws.atExit.keys()) {
     try {
-      const atExit = ws.atExit;
-      ws.atExit = undefined;
-      atExit();
+      const atExit = ws.atExit.get(key);
+      //Calling ns.exit inside ns.atExit can lead to recursion
+      //so the handler must be removed immediatly
+      ws.atExit.set(key, undefined);
+      if (typeof atExit == "function") atExit();
```
Snarling: Setting individual keys to undefined one at a time does not make sense to me. Is there a reason this is being done instead of just clearing the Map once after storing the relevant contents?

e.g.

```ts
const atExits = [...ws.atExit.values()];
ws.atExit.clear();
for (const atExit of atExits) {
  try {
    atExit();
  }
  catch (e: unknown) {
    handleUnknownError(e, ws, "Error running atExit function.\n\n");
  }
}
```

Snarling (commented):
Just have some questions about the Map.

src/Netscript/killWorkerScript.ts
shyguy1412: I made the change as @d0sboots suggested. However, your version iterates the atExit values twice. That might not be optimal if you can just do it in one go.

Snarling: Even if keeping it to the current one-iteration approach, is there a reason to explicitly set the value to "undefined" instead of just deleting that key? That would also allow making it a Map of functions, instead of a Map of function|undefined.

Iterating only once is good if it can have the correct behavior. In my suggestion code I captured the entire array of atExits before running any of them, so that the atExits are "locked" at the time the script is exited. Iterating the Map directly while running player code in between iterations leaves the door open for the player to modify the atExits while the script is already in the exiting phase.

Here's an example player script:

```js
/** @param {NS} ns */
  function setupExitHandler(n) {
    if (n > 10) return; // This is just to prevent an infinite loop
    ns.tprint(`Setting up exit handler ${n}`)
    ns.atExit(() => {
      ns.tprint(`Running exit handler ${n}`);
      setupExitHandler(n + 1);
    }, n.toString());
  }
  setupExitHandler(0);
  ns.tprint("Exiting script");
  ns.exit();
}
```

And here is the actual output in your current code:
```
test.js: Setting up exit handler 0
test.js: Exiting script
test.js: Running exit handler 0
test.js: Setting up exit handler 1
test.js: Running exit handler 1
test.js: Setting up exit handler 2
test.js: Running exit handler 2
test.js: Setting up exit handler 3
test.js: Running exit handler 3
test.js: Setting up exit handler 4
test.js: Running exit handler 4
test.js: Setting up exit handler 5
test.js: Running exit handler 5
```

Whereas I would expect the output to just be:
```
test.js: Setting up exit handler 0
test.js: Exiting script
test.js: Running exit handler 0
test.js: Setting up exit handler 1
```

During exit handler 0, the player attempts to set up exit handler 1, but exit handler 1 does not get run because it wasn't set up prior to script death.

So on the iterate once vs. iterate twice issue, it depends on whether you think the first behavior is correct (where the player can continue adding more exit handlers to the pile while the script is already dying).

d0sboots: Sorry I wasn't paying attention to this earlier. Just wanted to chime in that I agree with Snarling 💯 and thus is what I was trying to get at earlier, but I didn't express it properly and ended up being confusing.

src/NetscriptFunctions.ts
```diff
@@ -1716,14 +1716,21 @@ export const ns: InternalAPI<NSFull> = {
...
-    }; // Wrap the user function to prevent WorkerScript leaking as 'this'
-  },
+  atExit:
+    (ctx) =>
+    (f, id = "default") => {
+      if (typeof f !== "function") {
+        throw helpers.makeRuntimeErrorMsg(ctx, "argument should be function");
+      }
+
+      if (typeof id !== "string") {
```
d0sboots, line 1718: Optional: helpers.strimg here?

d0sboots:
The code looks solid to me now. The only remaining question is if we actually want this change. I don't see it as being super helpful (you can put this same map in your own atExit function very easily), but it's also backward compatible and as long as it's not drastically hurting script start/stop performance, I'm ambivalent.

shyguy1412:
If this change doesnt get accepted I will try again with a setTimeout style change for v3.
As it has been discussed plenty, this change wont enable you to do anything that wasnt possible before but it would be a massive QOL change to me. It just allows for library code to be far cleaner imo. I wont give up on this until I am explicitly told that it is a no go and unwanted x.x

d0sboots:
Yup, I was just stating my position. It's up to Snarling to make the final call for something controversial like this.

Snarling:
Looks good overall. Will merge this shortly, Just rebased and making a few minor edits (will just make a followup PR for this). Just trying to merge this in before releasing 2.6.0

I also mostly agree with d0s' opinion above - I'd personally prefer implementing custom handling in my own player scripts instead of the game handing me a ready made solution. But I see the appeal for some people, and it doesn't come with much of a performance burden.

PR #1062, opened 2024-01-25
BUGFIX: Attach hostname to script links in ls command

### LINKED ISSUE #1051: Weird way to copy scripts between servers
I don't know if it's supposed to work that way, I probably know the reason why it's happening. You can copy scripts between servers in an odd way following those steps:

## Steps to reproduce
1. Create script on server-1
2. Save it
3. Run command `ls` on server-1
4. Connect with a server-2
5. Click on a script from `ls` command
6. It will open file in editor but on the server-2
7. You can save file now and it's copied

## Possible reason
Clicking a file probably runs `nano %filename%` so it runs on currently connected server.

## Game version
2.5.2 (https://github.com/bitburner-official/bitburner-src/commit/c6141f2adfe1e8b12015b8c19db21fb3d3db5a6c) Steam edition

LJNeon:
~~This needs to be tested.~~

fixes #1051

### DIFF
--- a/src/Terminal/commands/ls.tsx
+++ b/src/Terminal/commands/ls.tsx
@@ -136,8 +136,9 @@ export function ls(args: (string | number | boolean)[], server: BaseServer): voi
     function onClick() {
       const code = server.scripts.get(fullPath)?.content ?? "";
       const files = new Map<ContentFilePath, string>();
+      const options = { hostname: server.hostname };
       files.set(fullPath, code);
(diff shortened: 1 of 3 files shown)


Snarling:
Not able to test this right now, waiting to merge.

Semanual:
Just tested, works as intended.
Upon clicking the link:
imagem
imagem

And old nano behavior was kept:
imagem
imagem

PR #1063, opened 2024-01-26
TERMINAL: Fix inconsistent / janky scrolling behavior

Snarling:
Scrolling on the terminal used to cause the terminal input to slightly change vertical position: fixed, now it actually stays at a fixed position onscreen while scrolling.

Scrolling on the terminal was previously possible when there were not enough contents to require scrolling: fixed

@shyguy1412 FYI

### DIFF
--- a/src/Terminal/ui/TerminalInput.tsx
+++ b/src/Terminal/ui/TerminalInput.tsx
@@ -12,9 +12,6 @@ import { longestCommonStart } from "../../utils/StringHelperFunctions";
 
 const useStyles = makeStyles((theme: Theme) =>
   createStyles({
-    textfield: {
-      margin: theme.spacing(0),
-    },
     input: {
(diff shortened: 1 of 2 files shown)


Snarling:
On live, displayed `scan-analyze 2` to overfill the terminal and showed the scrolling behavior (terminal input wobbles up and down). Then cleared the terminal and showed that it still scrolls:


https://github.com/bitburner-official/bitburner-src/assets/84951833/eae07cdd-3d79-41c4-a4b7-69c1bf667b9f


Performed the same tests on this PR, showing that the terminal input holds steady and it is no longer possible to scroll the terminal when there aren't enough contents to scroll:


https://github.com/bitburner-official/bitburner-src/assets/84951833/c6a284ec-3a83-486e-ad63-166ad71a2e1c

PR #1064, opened 2024-01-26
BUGFIX: Properly deep copy array data for coding contracts 

### LINKED ISSUE #1010: Wrong output example for "Find All Valid Math Expressions" contract
I think it would be clearer in the output example to explicit the expected result is an array of strings. Currently, it's an invalid JS expression

Actual: Output: [1+2+3, 1*2*3]
Expected:  Output: ["1+2+3", "1*2*3"]

```
["4592",97]

You are given the following string which contains only digits between 0 and 9:

 4592

 You are also given a target number of 97. Return all possible ways you can add the +(add), -(subtract), and *(multiply) operators to the string such that it evaluates to the target number. (Normal order of operations applies.)

 The provided answer should be an array of strings containing the valid expressions. The data provided by this problem is an array with two elements. The first element is the string of digits, while the second element is the target number:

 ["4592", 97]

 NOTE: The order of evaluation expects script operator precedence NOTE: Numbers in the expression cannot have leading 0's. In other words, "1+01" is not a valid expression Examples:

 Input: digits = "123", target = 6
 Output: [1+2+3, 1*2*3]

 Input: digits = "105", target = 5
 Output: [1*0+5, 10-5]
```

### LINKED ISSUE #1038: Data is mutable in coding contract "Proper 2-Coloring of a Graph"
Description:
data[1][n] in coding contract "Proper 2-Coloring of a Graph" is mutable from user script. It should be protected as it makes the coding contract unsolvable.

Version: Bitburner v2.5.2 (launched by steam on windows11)

Reproducer:

```javascript
/** @param {NS} ns */
  // Find a contract "Proper 2-Coloring of a Graph" and set its hostname and filename.
  let host = "fulcrumassets"
  let filename = "contract-881531.cct"

  let data = ns.codingcontract.getData(filename, host);
  let description = ns.codingcontract.getDescription(filename, host);

  ns.tprintf("=================================");
  ns.tprintf("description:\n%s", description);

  for (let i of data[1]) {
    i.splice(0, i.length)    // break the original data
    i.push("broken!")
    i.push("broken!")
    i.push("broken!")
  }
  data = ns.codingcontract.getData(filename, host);
  description = ns.codingcontract.getDescription(filename, host);
  ns.tprintf("=================================");
  ns.tprintf("description:\n%s", description);
  
  for (let i of data[1]) {
    i.splice(0, i.length)    // break the original data
  }
}
```

This reproducer removes all edge information from the contract. After running this script several times, the following output can be obtained. The original data is corrupted so the problem description is also broken.

Result:

```
=================================
description:
You are given the following data, representing a graph:
 [6,[[],[],[]]]
 Note that "graph", as used here, refers to the field of graph theory, and has no relation to statistics or plotting. The first element of the data represents the number of vertices in the graph. Each vertex is a unique number between 0 and 5. The next element of the data represents the edges of the graph. Two vertices u,v in a graph are said to be adjacent if there exists an edge [u,v]. Note that an edge [u,v] is the same as an edge [v,u], as order does not matter. You must construct a 2-coloring of the graph, meaning that you have to assign each vertex in the graph a "color", either 0 or 1, such that no two adjacent vertices have the same color. Submit your answer in the form of an array, where element i represents the color of vertex i. If it is impossible to construct a 2-coloring of the given graph, instead submit an empty array.

 Examples:

 Input: [4, [[0, 2], [0, 3], [1, 2], [1, 3]]]
 Output: [0, 0, 1, 1]

 Input: [3, [[0, 1], [0, 2], [1, 2]]]
 Output: []
=================================
description:
You are given the following data, representing a graph:
 [6,[["broken!","broken!","broken!"],["broken!","broken!","broken!"],["broken!","broken!","broken!"]]]
 Note that "graph", as used here, refers to the field of graph theory, and has no relation to statistics or plotting. The first element of the data represents the number of vertices in the graph. Each vertex is a unique number between 0 and 5. The next element of the data represents the edges of the graph. Two vertices u,v in a graph are said to be adjacent if there exists an edge [u,v]. Note that an edge [u,v] is the same as an edge [v,u], as order does not matter. You must construct a 2-coloring of the graph, meaning that you have to assign each vertex in the graph a "color", either 0 or 1, such that no two adjacent vertices have the same color. Submit your answer in the form of an array, where element i represents the color of vertex i. If it is impossible to construct a 2-coloring of the given graph, instead submit an empty array.

 Examples:

 Input: [4, [[0, 2], [0, 3], [1, 2], [1, 3]]]
 Output: [0, 0, 1, 1]

 Input: [3, [[0, 1], [0, 2], [1, 2]]]
 Output: []
```

LJNeon:
fixes #1010
fixes #1038

### DIFF
--- a/src/NetscriptFunctions/CodingContract.ts
+++ b/src/NetscriptFunctions/CodingContract.ts
@@ -67,17 +67,8 @@ export function NetscriptCodingContract(): InternalAPI<ICodingContract> {
       const contract = getCodingContract(ctx, hostname, filename);
       const data = contract.getData();
       if (Array.isArray(data)) {
-        // For two dimensional arrays, we have to copy the internal arrays using
-        // slice() as well. As of right now, no contract has arrays that have
-        // more than two dimensions
-        const copy = data.slice();
-        for (let i = 0; i < copy.length; ++i) {
-          if (data[i].constructor === Array) {
-            copy[i] = data[i].slice();
-          }
-        }
-
-        return copy;
+        // For two dimensional arrays, we have to copy the internal arrays as well
+        return JSON.parse(JSON.stringify(data));
       } else return data;
     },
     getDescription: (ctx) => (_filename, _hostname?) => {
--- a/src/NetscriptFunctions/CodingContract.ts
+++ b/src/NetscriptFunctions/CodingContract.ts
@@ -67,7 +67,7 @@ export function NetscriptCodingContract(): InternalAPI<ICodingContract> {
       const contract = getCodingContract(ctx, hostname, filename);
       const data = contract.getData();
       if (Array.isArray(data)) {
-        // For two dimensional arrays, we have to copy the internal arrays as well
+        // For multi-dimensional arrays, we have to copy the internal arrays as well
         return JSON.parse(JSON.stringify(data));
       } else return data;
     },
--- a/src/data/codingcontracttypes.ts
+++ b/src/data/codingcontracttypes.ts
@@ -1198,9 +1198,9 @@ export const codingContractTypesMetadata: ICodingContractTypeMetadata[] = [
         `"1+01" is not a valid expression`,
         "Examples:\n\n",
         `Input: digits = "123", target = 6\n`,
-        `Output: [1+2+3, 1*2*3]\n\n`,
+        `Output: ["1+2+3", "1*2*3"]\n\n`,
         `Input: digits = "105", target = 5\n`,
-        `Output: [1*0+5, 10-5]`,
+        `Output: ["1*0+5", "10-5"]`,
       ].join(" ");
     },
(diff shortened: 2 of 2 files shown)


Snarling:
Forgot about this when merging yesterday, but cloneDeep (from lodash) would probably have been a better solution here, as it preserves type info. The JSON functions abuse `any` types so it's easy to get in trouble with those, but it shouldn't cause any issues in this case since our internal code isn't using the result for anything, it's just being passed to the player.

LJNeon:
Actually turns out we have plans to start using `structuredClone()` in other places (see development channel in discord) and it would probably be better to use for this as well as anywhere we use lodash's deep clone or the typical stringify/parse.

PR #1065, opened 2024-01-26
UI: Changing color of stocks going up to success instead of primary

Semanual:
The stock market color of the stocks going up was your primary color, instead of success. Even though with the default theme it makes no difference (because primary and success are exactly equal to each other), it leads to some unexpected behaviors like this, when the primary color is changed in the theme editor:
imagem

I changed the code so that it uses now the success color. After the change, with primary color set to purple:
imagem


Semanual:
By the way: one having 4s and the other not is because the first one is in my actual save file (which I don't have 4s yet) and the second in my local branch; it has nothing to do with the PR.

PR #1069, opened 2024-01-27
MISC: change all nextUpdates to use one Promise

Caldwell-74:
partly a duplicate to #1068 but didnt want to fish that one change out of it since i had them all allready done


Caldwell-74:
i changed the sleeves bladeburner task to only create promises when needed instead of for each task by default

Snarling:
Due to duplicate types being made, I just made a generic PromisePair type for storing a promise and its resolver. Rebase was just a conflict with corporation imports.

Caldwell-74:
i did think about that but wasnt comfy with generics in general but thanks and good job!

PR #1070, opened 2024-01-27
UI: Make Hacknet node grid column count dynamic

shyguy1412:
Right now the Hacknet node grid is fixed to 3 columns and overflows when the viewport is too small.

https://github.com/bitburner-official/bitburner-src/assets/48560522/540501f6-5457-44fc-8dd5-0b20702ad9b1

I changed this so the column count is dynamic and adjusts to the available space

https://github.com/bitburner-official/bitburner-src/assets/48560522/dead1fb3-496e-4f57-9f42-e0e07b721cfd


Snarling:
A little concerned with the exact column width being specified in pixels, but from what I recall grid doesn't like it if you use something dynamic (like min-content or max-content) for automatic sizing.

PR #1075, opened 2024-01-31
MISC: move server constants into their own constant

Caldwell-74:
pretty much just the title


d0sboots:
IMO the weaken changes should be it's own PR, that way this larger refactor would be much more obviously safe and the weaken change would be easier to verify. But that's a minor thing.

Caldwell-74:
yeah just stumbled over it when i renamed everything and remembered how long it took until it was common knowledge that that factor even exists
i can extract that tomorrow back down again

Caldwell-74:
maybe its better to then also change the other stuff and make it more uniform (some function directly modify the server objects others use a method on it to change stuff)

PR #1076, opened 2024-02-01
MISC: refactor weaken effect calculation

Caldwell-74:
so far we calculate the effect of weaken in three  +1 places
ns.weaken
ns.weakenAnalyze
terminal weaken

and server.weaken where the bn mult is applied

i extracted the logic into a new netscript helper function getWeakenEffect
this gives us one place if we want to change the formula

a side effect i added the server.cpuCores to the terminal weaken to future proof it if the npc server core pr (#963) is merged


Caldwell-74:
oh ill move getWeakenEffect into ServerHelpers it fits way better there

d0sboots:
You missed a usage in Script/ScriptHelpers.ts

Caldwell-74:
> Script/ScriptHelpers.ts

done but i added the runningscript.threads to it 
currently it just calculated without threads /as if it was run with one thread 
to offset the "nerf" that the bn mult is now applied aswell

d0sboots:
Sorry I didn't mention this earlier, and maybe you already did this, but for peace of mind can you smoke test a few weakens/the other affected functions and make sure they still work right?

Edit: You don't need to check the exact numbers on offline weakening, just that it still works at all. What you did looks like a bugfix to me.

Caldwell-74:
oh yeah for sure i can 
if iam lucky then i still have the test script around 
i tested them when they where part of that other pr and afterwards only did build tests
so its fair to retest them now

Caldwell-74:
not sure if it should show  "0"  or "0.05" when the server is already at min on a terminal weaken
```
bn1
terminal:
`Security decreased on 'n00dles' by 0.050 from 1.000 to 1.000 (min: 1.000) and Gained 3.300 hacking exp.`


1 core
Target: joesguns
Before: 15
After: 14.95
Effect: 0.05
Correct? true   //returned value from weaken is equal to observed change
Expected? true  // weaken effect is equal to weakenAnalyze 


2 cores
Target: joesguns
Before: 14.95
After: 14.893749999999999
Effect: 0.05625
Correct? true
Expected? true 
```

```
bn12.4
terminal:
`Security decreased on 'joesguns' by 0.046 from 22.500 to 22.454 (min: 8.000) and Gained 13.556 hacking exp.`

1 core
Target: joesguns
Before: 22.453807728698674
After: 22.40761545739735
Effect: 0.046192271301325714
Correct? true
Expected? true 

2 cores
Target: joesguns
Before: 22.40761545739735
After: 22.35853616913969
Effect: 0.049079288257658574
Correct? true
Expected? true 
```

PR #1077, opened 2024-02-03
MISC: Use structuredClone() for deep cloning

LJNeon:
Rather than sketchy or custom solutions like `JSON.parse(JSON.stringify())` or `_.cloneDeep()` it's better to use the native solution. `structuredClone()` is supported by [>93%](https://caniuse.com/mdn-api_structuredclone) of users and is polyfilled on electron.

It might be ideal to merge this sooner rather than later because it modifies a varied list of files and is thus susceptible to merge conflicts.

### DIFF
--- a/.browserslistrc
+++ b/.browserslistrc
@@ -1,4 +1,3 @@
 last 8 versions
-node 16
 not dead
 not ie <= 11
--- a/.browserslistrc
+++ b/.browserslistrc
@@ -1,3 +1,4 @@
 last 8 versions
+node 16
 not dead
 not ie <= 11
--- a/FixJSDOMEnvironment.ts
+++ b/FixJSDOMEnvironment.ts
@@ -1,10 +1,13 @@
 import JSDOMEnvironment from "jest-environment-jsdom";
+import { cloneDeep } from "lodash";
 
 // https://github.com/facebook/jest/blob/v29.4.3/website/versioned_docs/version-29.4/Configuration.md#testenvironment-string
 export default class FixJSDOMEnvironment extends JSDOMEnvironment {
   constructor(...args: ConstructorParameters<typeof JSDOMEnvironment>) {
     super(...args);
 
+    // TODO Tests aren't polyfilled.
+    this.global.structuredClone = cloneDeep;
     // FIXME https://github.com/nodejs/node/issues/35889
     // Add missing importActual() function to mirror requireActual(),
     // which lets us work around the ESM bug.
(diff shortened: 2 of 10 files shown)


LJNeon:
It appears that `structuredClone()` was added in node.js v17, the tests use v16 and don't appear to polyfill it. Is this something I should fix in the PR by updating the node.js version used in the Test workflow?

d0sboots:
Yeah the test pipeline is it's own thing, it doesn't go through the regular build steps including TS(!), which means our tests are probably broken type-wise (but okay in terms of functionality).

If you update the node version that should probably be a standalone change, the one used in our tests should probably stay the same as what we use elsewhere?

LJNeon:
Would it be viable to add the regular build steps to the Test workflow instead? I think we can avoid duplicate work if we make it dependent on the Build workflow. That would solve this while also improving the quality of the tests significantly.

d0sboots:
The main issue is it makes the tests a *lot* slower overall.

LJNeon:
I guess I might have to give up on this PR then unfortunately, I recall people trying to bump node.js versions and running into several complications before giving up. I think there were some abandoned dependencies that didn't support newer versions? Either way not something I could try to tackle on mobile.

LJNeon:
Although out of curiosity what harm is there if the tests are in a different nodejs version? Any build issues will get caught anyways by the Build job, and I don't think one version bump would cause significant differences that could affect the Test job?

d0sboots:
Something a lot simpler would be to set up/replace `structuredClone` in the test envelope with an impl based on cloneDeep. We already have a custom environment for dealing with a different test issue.

LJNeon:
~~As I mentioned on discord I may have found a much better and simpler solution. There is a `babel-polyfill` package that works with jest alongside `babel-jest`. We're missing that package for some reason.~~ Nevermind, jest just doesn't work as intended so a workaround is the simplest alternative until someone wants to rework the tests.

LJNeon:
@Snarling This should be ready to be merged now, if you could look it over.

Snarling:
@LJNeon @d0sboots I'm still not sure why we're moving to structuredClone instead of just continuing to use cloneDeep, can you provide me with a brief high level reason this should be preferred?

My understanding is that cloneDeep doesn't support Classes and certain other exotic objects, but in return cloneDeep is faster. Since cloneDeep is usually used as a bridge between internal code and something we're providing to the player through the API, it's not actually a drawback that cloneDeep doesn't support classes because we wouldn't want to accidentally leak internal classes to the player anyway.

I'm still merging it though, I know there has been a lot of discussion about it on the discord but I have not really kept up on it since I've been pretty busy.

d0sboots:
It's largely for consistency, cloneDeep would also still be fine. JSON.stringify + parse is the one we'd like to stop using mostly for cloning data structures.

Because of the speed issue, using `{...orig}` is best when you know the object is shallow.

LJNeon:
Also I thought that cloneDeep did support classes partially? It maintains the prototype of the object.

PR #1079, opened 2024-02-07
UI: Improve theme support.

### LINKED ISSUE #1039: UI: locations in city render in white text regardless of background color

adeilt:
# UI: Improve theme support.

Originally this was about better support for light themes, but now it's more about applying themes to more UI elements with a few extra changes that improve things when using light themes.

Where I've changed colors that might be noticeable I've tried to keep the colors close to the original and consistent with the visual style.

Added:
- `mapstatic` for map elements like names and roads
- `maplocation` for map elements that are clickable/visitable
- `currentcity` for the city where the player is currently located
<details><summary>some possibly spoilery elements</summary>

- `bnlvl0`, `bnlvl1`, `bnlvl2`, and `bnlvl3` for coloring the clickable bitnode selectors on the bitnode selection screen (post WD destruction or `b1t_flum3.exe` run)
</details>

# Linked issues

fixes #1039

# Comparison with release version 2.5.2

Here's are some visual comparisons with the version running at https://bitburner-official.github.io/:

| Before  | After |
| ------- | ----- |
| image | image |
| image | image |
| image | image |
| image | image |

## Additions to Theme Editor

<details><summary>The changes to the Theme Editor have potentially spoilery names.</summary>

image</details>

# Review Notes

There are some things that I want to highlight for review and possible revision before merging:

- should I name the spoilery theme color names something more obfuscated?  Something like `secretlvl0` instead of what I have now?  Or are the current names (see the Additions to Theme Editor section above) obscure enough until you reach that point in the game?
- The Theme Editor is getting pretty crowded.
  - Should I attempt to add that to this PR?  It seems to me that it might be better to make a new issue/PR for that kind of work.
  - I'm also less confident that I could make it look good.  I can add to what's there and try to match an existing visual style, but historically I've been *catastrophically* bad at frontend visual design.
- As a light theme user, my eyes are of limited use testing the effect of my changes on darker themes in the sense that poking around looking for stuff that doesn't look right is very likely to miss something because I'm unfamiliar with how the game looks with a dark background.  I could use some help there if someone used to dark themes is willing to donate a few minutes to run the game with these changes and poke around a bit looking for things out of place.
- Finally, I'm sure I've missed stuff.  If you know of things that don't look right with light themes then I'd be delighted to hear about and add them to this PR.  I don't have access to all the parts of the game yet as a player and the note about unfamiliarity above therefore applies.


Snarling:
For static map elements, they were previously using the "primary" color, is there a reason they should use a different color now?

So many new specific colors that are just used in one place is weird and not really how theming should be done IMO. Were there not existing colors already present in the theme that would be suitable for use?

Snarling:
Assigning theme colors for the hardcoded colors (e.g. the existing `#111` for MuiAccordionSummary) is a great change and necessary for making light themes work well in situations where that color is used. I'm just not sure whether we need to add a whole bunch more theme colors, vs. just reusing existing theme colors, and for instances where we're already using a theme color, there's a good chance we don't need to change which of the existing theme colors it's looking at, the theme just needs to be updated to have proper contrast using the existing colors (e.g. the city point of interest issue is due to the "white" color in that theme having no contrast against the background, when that is expected to be a high-contrast color pairing).

Would appreciate any other feedback on this.

src/Themes/ui/Theme.tsx
```diff
@@ -66,6 +72,9 @@ export function refreshTheme(): void {
       successdark: Settings.theme.successdark,
       white: Settings.theme.white,
       black: Settings.theme.black,
+      mapstatic: Settings.theme.currentcity,
```
d0sboots: You messed this up

src/Themes/ui/Theme.tsx
```diff
@@ -251,7 +260,7 @@ export function refreshTheme(): void {
             color: Settings.theme.primary,
             backgroundColor: Settings.theme.well,
             borderRadius: 0,
-            border: "2px solid white",
+            border: "2px solid " + Settings.theme.welllight,
```
d0sboots: IMO this should be theme.white

d0sboots (changes requested):

adeilt:
> For static map elements, they were previously using the "primary" color, is there a reason they should use a different color now?

My preferred `primary` and `maplocation` colors are close together, so it was tough to visually distinguish the clickable cities from the static map characters.  Having the clickable things lighter than the non-clickable things distinguishes them, but then they seem faded and less important.

On the other hand, I could just use color to make the locations stand out instead of lightness, and that would remove the need for the `mapstatic` color.

> So many new specific colors that are just used in one place is weird and not really how theming should be done IMO. Were there not existing colors already present in the theme that would be suitable for use?

The `lvlN` colors were hardcoded for visibility against a dark background and kind of had a `red->yellow->cyanish->blue` progression that makes some sense from the perspective of "traffic-light" colors (in my country, anyway).

We could lose those four new theme colors, but we'd lose the sense of progression that those familiar with that color scheme would be familiar with.  We could go `error->warning->success->info` to more closely match the existing colors, but if green/success will be in there it feels like it should be last.  However, that would conflict with the existing color scheme with blue/info last.

The `mapstatic` theme color makes sense to me to remove, but I'm less certain about the four `lvlN` colors.  At the moment I'm leaning toward removing them and going with a `error->warning->info->success` progression for the leveled colors and just accept the significant change in colors on that screen in favor of putting `success` as the highest level (even though I hate when interfaces change).  However, I don't feel strongly about that, so I'm happy to go with the recommendation of those better at visual/frontend design.  :)

src/Themes/ui/Theme.tsx
adeilt: I spent a bit of time mulling this one over.  I went with `welllight` because it's both consistent with other borders in the game and can be made to work well with light-background themes.

d0sboots:
I both agree and disagree. I agree that some of the new theme settings aren't needed: for instance, mapstatic and currentcity can be merged, and maplocation should probably just be dropped for theme.white.

Some of the bn themes also have overlap (we could use errorlight etc.) but it would be an odd reuse, and bnlvl2/3 still don't have good equivalents. I'm inclined to leave all of them as their own thing.

src/Themes/ui/Theme.tsx
adeilt: Ah, thanks for the review!  Turns out I'm removing `mapstatic` anyway, so this embarrassing mistake will be cast into the void.  Or maybe the null.

d0sboots:
Oh, I'm also of the opinion that in a reversed theme, theme.white and theme.black are likely meant to be reversed as well (by the person editing the theme)

src/Themes/ui/Theme.tsx
adeilt: Although by that logic I probably should have found something else to use above for MuiInput's borderBottomColor.  Hrm.

adeilt:
> I both agree and disagree. I agree that some of the new theme settings aren't needed: for instance, mapstatic and currentcity can be merged, and maplocation should probably just be dropped for theme.white.

The use of `theme.white` was exactly the problem I set out to resolve to begin with (see also #1040).  Per discussion above, I'm inclined to lose `mapstatic`, but I still see value in the current city being visually distinguished from the static map characters.

Alternately, the Travel screen could just say the name of the current city at the top of the screen:

image

or even

image

I'd see no more need for the `currentcity` color with something like that.

> Some of the bn themes also have overlap (we could use errorlight etc.) but it would be an odd reuse, and bnlvl2/3 still don't have good equivalents. I'm inclined to leave all of them as their own thing.

What would you think of renaming them to `secretlvl0` or just `lvl0`?  It's not all that big a spoiler, especially given that most folks probably never open the Theme Editor and probably wouldn't know what the `bn` indicated, but I'm still a little conflicted about the current naming given that it's available from the first moment of the game.

adeilt:
I think I could probably get rid of `currentcity` in favor of `disabled`.

d0sboots:
> The use of theme.white was exactly the problem I set out to resolve to begin with (see also https://github.com/bitburner-official/bitburner-src/pull/1040). Per discussion above, I'm inclined to lose mapstatic, but I still see value in the current city being visually distinguished from the static map characters.

So, I was saying you could merge `mapstatic` and `currentcity` because you had them set to the same value across all the themes. Thus, they were indistinguishable (at the current time) from a theme level.

Similarly, `maplocation` is set to `#ffffff`, which is the same as the (default) value of `theme.white`. In light themes, `white` should actually be dark, which should solve the contrast issue. I'm not sure why you'd need a separate knob just for the map locations.

> What would you think of renaming them to `secretlvl0` or just `lvl0`?

I think `bnlvl0` is completely inscrutable if you don't know what a "bn" is. If you do, it doesn't matter. Using the abbreviation to sufficiently obscure the meaning is a compromise we've used in lots of other places.

adeilt:
> > The use of theme.white was exactly the problem I set out to resolve to begin with (see also #1040). Per discussion above, I'm inclined to lose mapstatic, but I still see value in the current city being visually distinguished from the static map characters.
> 
> So, I was saying you could merge `mapstatic` and `currentcity` because you had them set to the same value across all the themes. Thus, they were indistinguishable (at the current time) from a theme level.
> 
> Similarly, `maplocation` is set to `#ffffff`, which is the same as the (default) value of `theme.white`. In light themes, `white` should actually be dark, which should solve the contrast issue. I'm not sure why you'd need a separate knob just for the map locations.

The matching of existing colors was by design; I wanted to change the dark-background themes - which are the vast majority - as little as possible.  The `light` theme is the only one where I've set `maplocation` and `currentcity` to anything different:

```
    maplocation: "#000000",
    currentcity: "#b0b0b0",
```

I've already reverted `mapstatic` per the earlier discussion.  I still think that an indication of the current city would be good, but at this point I'd prefer putting the name of the current city at the top of the page more than piling on more theme colors.

> > What would you think of renaming them to `secretlvl0` or just `lvl0`?
> 
> I think `bnlvl0` is completely inscrutable if you don't know what a "bn" is. If you do, it doesn't matter. Using the abbreviation to sufficiently obscure the meaning is a compromise we've used in lots of other places.

A fair point.

adeilt:
My current plan:

- lose `currentcity` in favor of one (or both) of:
  - use the `disabled` color for the current city
  - put the name of the current city in place of the word "here" at the top of the Travel map
- use `theme.white` and `theme.black` as roughly contrasting with and matching the background, respectively
  - test the use of `theme.white` set to `#000000` to see where that breaks and fix it
  - find a way to indicate that usage of `theme.black` and `theme.white` in the Theme Editor

adeilt:
I tried organizing the Theme Editor a little differently and despite my earlier comments, it might be reasonable to include here:

image

Please let me know if you think that's a good change and whether it's appropriate to include here.

EDIT: I'm still tweaking this a bit; the screenshot is a draft.

adeilt:
I've stashed the Theme Editor changes for now as I think they should definitely be a different branch/PR.

src/Themes/ui/Theme.tsx
adeilt: Actually, given my now clarified understanding of the intent of `theme.white`, I think leaving `MuiInput` as `theme.white` does make more sense than using a different color.

The only thing unresolved is `theme.welllight` vs `theme.white` for this border.  I went with `welllight` because most of the other border colors in this file seem to be `welllight` or `well`.  That said, `theme.white` is probably fine as well, so I'll change it.

Also, speaking as the newest victim of https://github.com/orgs/community/discussions/10369, thanks for your patience.  :)

adeilt: I'm still unfamiliar with Github's code review tools and I'm unsure why `welllight` still appears above, but the "files changed" tab of this PR shows `white`, so I'm calling this done.

d0sboots (approved):

d0sboots:
> Also, speaking as the newest victim of https://github.com/orgs/community/discussions/10369, thanks for your patience. :)

Oh, your comments were showing up to me, I just didn't have anything to say (it seemed like you were figuring stuff out and I was ultimately fine either way on white vs welllight)

PR #1083, opened 2024-02-09
UI: Active Scripts changes (followup/alternative to #933)

Snarling:
This started off as just making some changes on top of #933 to tweak the UI but it turned into a much larger rewrite of the existing Active Scripts code. Some of it was stuff I had to change organizationally to make the UI changes work, some is just efficiency changes (prevent unnecessary iteration / high cost function calls during iteration), some was just stylistic changes.

@TheAimMan mind taking a look at these changes and provide some feedback?

Reviewing the code is kind of gross because I did it all as one big commit instead of organizing the changes better, my bad there. The main thing I'm concerned with is how you feel about the actual UI changes / if you have suggestions on how it could be improved before I merge this in.


Snarling:
Video showing differences between the existing page and the changes:

https://github.com/bitburner-official/bitburner-src/assets/84951833/04a1e601-79a4-418b-8b59-75a0b496dcc4

The existing page is first, the changes are afterwards (on the tab that says 2.5.1)

TheAimMan:
Since I'm not a super coder I just copied all of the changed files over to a local branch to test it out.  I think its good!  I only have one observation that seemed out of place to me, and its the buttons you can/can't press.  I feel like the current implementation should be swapped.  It's like if you had a hyperlink that's not underlined and blue.  You're not really aware it can be pressed.

Now just clean up my crappy code in https://github.com/bitburner-official/bitburner-src/pull/954 and the Active Scripts page will be in a very good place :D

Snarling:
The blue disabled buttons are in the existing UI as well (the leftside buttons are blue in both screenshots since we're on the first page of servers/first page of scripts)

image

It's just because the default theme uses this blue color as the "disabled" color. There's not really an easy way to override it other than changing the default theme.

But maybe adding the text in between the buttons makes it easier to not notice the clickable buttons on the left. I'll try some rearrangement to see if that helps.

TheAimMan:
Oh wow, I didn't realize it was blue before lol.  I'd agree with you, having the arrows be the same color as the text, and 1-2 spaces away from said text, makes it seem like its just a continuation of the string and not an actionable item.

d0sboots:
I'm with AimMan here, I think the old-style of the next/forward links was a bit better. Having them all together and to the right makes the information easier to scan, and also makes it faster/easier to flip back and forth between pages.

Snarling:
Yeah, agreed. I moved stuff around again, here is what it looks like now.

image

Merging, but I'm not going to be offended if someone thinks this can be improved further and suggests further changes in another PR. Just trying to get stuff closed out.

PR #1060, opened 2024-01-24
API: Changing return value of ns.bladeburner.getSkillUpgradeCost to return Infinity when the skill's level overshoot the maximum level

Semanual:
We have no way to know whether a skill overshooted or not the maximum skill level other than hardcoding that Overclock is at most 90. Following `getPurchasedServerCost`'s logic, I made it return Infinity when it does.

Testing code, with overclock at level 70:
```js
/** @param {NS} ns */
export async function main(ns) {
  ns.tail();
  ns.print(ns.bladeburner.getSkillUpgradeCost("Overclock"))
  ns.print(ns.bladeburner.getSkillUpgradeCost("Overclock", 10))
  ns.print(ns.bladeburner.getSkillUpgradeCost("Overclock", 20))
  ns.print(ns.bladeburner.getSkillUpgradeCost("Overclock", 21))
  ns.print(ns.bladeburner.getSkillUpgradeCost("Overclock", 90))
}
```
imagem

PR #1084, opened 2024-02-10
BUGFIX: Fixed bug that returned Infinity cost for most skills

Semanual:
In [my last PR](https://github.com/bitburner-official/bitburner-src/pull/1060), I didn't notice skills with no maximum were 0, I assumed it was Infinity. Yes, I know, my bad, I'm really sorry...
This introduced a bug, in which whenever your skill level were above 1 in any skill without maximum level, the cost would always return Infinity.
imagem
This PR fixes this bug, keeping the last PR's intended functionality.
```js
/** @param {NS} ns */
  console.log(ns.bladeburner.getSkillUpgradeCost("Cloak")); // With level 1+, this was Infinity
  console.log(ns.bladeburner.getSkillUpgradeCost("Cloak", 90)); // With level 1+, this was Infinity
  console.log(ns.bladeburner.getSkillUpgradeCost("Overclock")); // Overclock is the only one that was working properly at any level
  console.log(ns.bladeburner.getSkillUpgradeCost("Overclock", 100));
}
```
/\ Testing code here.

### DIFF
--- a/src/Bladeburner/Bladeburner.tsx
+++ b/src/Bladeburner/Bladeburner.tsx
@@ -2299,14 +2299,13 @@ export class Bladeburner {
     }
 
     const skill = Skills[skillName];
-    const currentLevel = this.skills[skillName];
-    if (currentLevel == null) {
-      return skill.calculateCost(0, count);
-    } else if (currentLevel + count > skill.maxLvl) {
+    const currentLevel = this.skills[skillName] ?? 0;
+
+    if (skill.maxLvl !== 0 && currentLevel + count > skill.maxLvl) {
       return Infinity;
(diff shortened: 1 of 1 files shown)


Snarling:
I will make sure to merge this in. Still busy for the next few hours so it'll be a little bit.

Infinity or undefined/null make more sense than 0 for indicating that there's no maximum, would it be a lot to just change the skills to use one of those instead of 0 when there's no maximum?

Semanual:
I don't know, I just used what was there already, I may take a look at that if you want me to

Snarling:
I'll use this as is, and we'll look at changing it to make more sense later on.

PR #1085, opened 2024-02-10
FOLLOWUP: Streamlining for #1056

Snarling:
I couldn't push to #1056 so I'm just making a followup PR.

* Removed some extra iteration steps
* Hardcoded 1.008 as the base instead of calculating 1.1^(1/12) every time it was needed. 1.008 is very slightly rounding up (results in 5.08 instead of 5.05 mult at max corp loadout)


Snarling:
Superseded

PR #1086, opened 2024-02-10
UI: Added some VFX for the w0r1d_d43m0n destroying screen

Semanual:
Trying to make a bit of game juice, me and @shyguy1412 made a glitching effect for the w0r1d_d43m0n and b1t_flume.exe screens. It's effect is initially very subtle and slowly goes up based on how many source files you have, and also progressively increase while the text shows up (which means, at the beginning of the screen it is very subtle, then the effect goes up until it's maximum at the end of the screen). It also implements a fade in/fade out screen, which was used for the beginning of BN selection screen.

**VISUAL EPILEPSY WARNING: THE FOLLOWING VIDEOS CONTAIN FLASHING LIGHTS**
Initial effect, without any source files:
https://github.com/bitburner-official/bitburner-src/assets/99666293/ae520549-7249-4f6d-b12c-5ebb1ae4b4e1

Middle effect, when you have 3 source files of each:
https://github.com/bitburner-official/bitburner-src/assets/99666293/dc7a6967-b12c-466e-818c-afead63a584a

Strongest effect, when you have 3 source files of each AND at least SF 12 level 34 (which is the cap for SF 12 contribution to the effect):
https://github.com/bitburner-official/bitburner-src/assets/99666293/19b39afb-95c4-4e2f-9e8a-81bbdea6ebbc
As you can probably see in the video above, the dead pixels effect is masked to form the bitburner logo. It's imperceptible at first, but as you destroy more and more bitnodes, it looks more clear.

To make it safe for people with visual epilepsy, I also made a switch that enables/disables all VFX, which is disabled by default:
imagem

When disabled, the screen behavior is the same as before the PR.
https://github.com/bitburner-official/bitburner-src/assets/99666293/a2fb97da-be98-4680-89b9-f94ee5bc8013

Because they are disabled by default, to prevent it to become a useless feature because nobody knows it's there, I also added some new pages in the tutorial:
imagem
imagem

Lastly, all of the visual effects are easily customizable, if you want me to tweak something, think an effect is too weak/too strong etc, just let me know!


d0sboots:
A suggestion (that will require a little more work): You could have the option default to off for existing saves, but on for new saves. That way it wouldn't "get" anyone who's currently playing (and possibly is already at a high BN), but new players would start out at the lowest intensity of the effect so it should hopefully be OK (while also being more discoverable as a result)?

Semanual:
AFAIK even for first level, it still triggers epilepsy because of the dead pixels flashing... but I don't know, I have little knowledge about epilepsy

LJNeon:
What's holding up this PR from being merged currently? Just the merge conflicts?

d0sboots:
> What's holding up this PR from being merged currently? Just the merge conflicts?

I think my last comment, where I was suggesting that we could default on for new saves. I was looking for a compromise that would allow us to default-on it, because I know how few people go into the options to enable things.

E9cipher:
@d0sboots are you still in favor of this? I could take it and resolve conflicts, if you want.

d0sboots:
Sure. It'll require some git-fu: you'll want to rebase instead of merge to get this up to date. 

There's also the settings work to do: I'd like this to default to on for new saves, but off if you have any levels of BN1 (i.e. are already in endgame). That way no one gets flashbanged, but it's not useless.

E9cipher:
Got it. One question though: this is a spoiler so it should be "spoiled" in the options tab, I guess. Instead of saying "flashing pixels in the bitverse", on BN1.1 we might say "turn this off if you have epilepsy". Idk if this is already done, haven't checked the code yet.

d0sboots:
If you look at the screenshots in the PR description you can see it's already done that way.

E9cipher:
I'll get to this tomorrow or the next day. It's been a busy week (though I never missed a Bitburner update lol)

catloversg:
IIRC, the last time this PR was discussed, the major concern was the potential discomfort that the flashing effect may cause. Personally, I would strongly request that the intensity of the text flashing effect be reduced. I don't have visual epilepsy, but I still feel very uncomfortable when looking at the "Middle effect". The speed of the flashing color pixels also needs to be reduced.

d0sboots:
> IIRC, the last time this PR was discussed, the major concern was the potential discomfort that the flashing effect may cause. Personally, I would strongly request that the intensity of the text flashing effect be reduced. I don't have visual epilepsy, but I still feel very uncomfortable when looking at the "Middle effect". The speed of the flashing color pixels also needs to be reduced.

Interesting, I don't have any particular discomfort looking at those, and in particular the color cycling on the pixels is a complete non-issue. 

What I find particularly interesting about that is that I usually disable effects like CRT distortion, scanlines, and especially chromatic aberration that are appearing in a lot of games, because I find them annoying.

We are only two data points of course, but IMO what we should be shooting for here is an effect that is visually fun *for the people who want that* - there will always be a significant chunk of people (beyond those with epilepsy) who don't want this type of effect at all.

A possible improvement for the text effect would be to increase the situation by a lot? Basically, turning it into separate copies instead of a sort-of blurring effect.

PR #1087, opened 2024-02-10
SLEEVES: add nextCompletion to InfiltrationWork

Caldwell-74:
title and change nextCompletion in Sleeve BladeburnerWork to use the new PromisePair type


Snarling (commented):
Sorry for the late review

I think nextCompletionPair will need to be excluded from the SleeveInfiltrateWork save data similar to how it's removed in SleeveBladeburnerWork.

Otherwise if there is a savegame while the parameters are not null, they will be stringified incorrectly, and then invalid data will be loaded in.

Here's a quick thing in the js console just showing what I mean by it creating invalid data on stringify - the promisePair property remains defined, but it no longer matches the expected shape of the promisePair object. promisePair.promise is truthy but isn't a promise while promisePair.resolve is undefined.
image

The game will load that incorrect format onto the promisePair property with no typechecking. If it's not included in the save, then it will just be initialized as null/null which is desired.

Caldwell-74:
right i missed that!
ill update it r eal quick

PR #1088, opened 2024-02-10
CORPORATION: Improve performance of cycle valuation

catloversg:
This PR is an alternative for PR #1085. It includes these changes:
- Create a new property: `numberOfOfficesAndWarehouses`. It's used for calculating cycle valuation.
- Hardcode 1.0079741404289038 instead of calculating `Math.pow(1.1, 1 / 12)`.

### DIFF
--- a/src/Corporation/Corporation.ts
+++ b/src/Corporation/Corporation.ts
@@ -471,13 +471,13 @@ export class Corporation {
 
   /** Initializes a Corporation object from a JSON save state. */
   static fromJSON(value: IReviverValue): Corporation {
-    // numberOfOfficesAndWarehouses is excluded from JSON save state, so we need to calculate it here.
-    value.data.numberOfOfficesAndWarehouses = 0;
-    for (const division of value.data.divisions.values()) {
-      value.data.numberOfOfficesAndWarehouses += getRecordValues(division.offices).length;
-      value.data.numberOfOfficesAndWarehouses += getRecordValues(division.warehouses).length;
+    const corporation = Generic_fromJSON(Corporation, value.data, Corporation.includedProperties);
+    // numberOfOfficesAndWarehouses is not in the included properties and must be calculated
+    for (const division of corporation.divisions.values()) {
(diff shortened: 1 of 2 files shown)


src/Corporation/Corporation.ts
```diff
@@ -460,13 +461,22 @@ export class Corporation {
...
 
   /** Initializes a Corporation object from a JSON save state. */
   static fromJSON(value: IReviverValue): Corporation {
+    // numberOfOfficesAndWarehouses is excluded from JSON save state, so we need to calculate it here.
+    value.data.numberOfOfficesAndWarehouses = 0;
+    for (const division of value.data.divisions.values()) {
+      value.data.numberOfOfficesAndWarehouses += getRecordValues(division.offices).length;
+      value.data.numberOfOfficesAndWarehouses += getRecordValues(division.warehouses).length;
+    }
     return Generic_fromJSON(Corporation, value.data);
```
Snarling: Operating on value.data directly isn't great for typesafety since value.data is type `any`. I recommend converting the unknown data into a known class object first.

Whether includedProperties should be passed in to the fromJSON is up for debate, I would probably pass it in but not passing it in shouldn't break anything since that excluded property has never been part of a savegame.

```suggestion
    const corp = Generic_fromJSON(Corporation, value.data, Corporation.includedProperties);
    // numberOfOfficesAndWarehouses is not in the included properties and must be calculated
    for (const division of corp.divisions.values()) {
      corp.numberOfOfficesAndWarehouses += getRecordValues(division.offices).length;
      corp.numberOfOfficesAndWarehouses += getRecordValues(division.warehouses).length;
    }
    return corp;
```

Snarling (commented):
Pretty much exactly how it should be done IMO, I would just approach the fromJSON slightly differently (made a suggestion comment)

src/Corporation/Corporation.ts
catloversg: I agree with your suggestion and have committed the changes.

d0sboots:
All comments are addressed so this seems good to go

PR #1089, opened 2024-02-10
PORTS: Support all serializable data.

LJNeon:
A significant portion of players who use ports are passing objects through them. Currently they are required to handle that themselves via JSON serialization. This PR adds better support for passing objects; which is more convenient, extensive, and optimized (probably, more on this one later).

This adds zero overhead to existing (or when passing any primitive types) port usage, and also isn't a breaking change. The questions to debate here are:
1. Should objects be supported in the first place?
2. If so, how exactly do we want to serialize objects?

Based on an extensive discussion in Discord, the overwhelming majority answered "yes" to question one. As for question two, that has been much more hotly contested. There were three main ideas:
1. Just use `structuredClone()`, it's not slow enough to actually matter. (@d0sboots is looking into that)
2. Use JSON serialization with a custom `replacer` and `reviver` to add better error handling and support a few popular types such as `Map`s and `Set`s.
3. Use a simplified deep cloning method, it'll only support basic objects and arrays that contain primitive values. (maybe only one dimensional?)

Since the first idea has been the most popular so far, I opened the PR with it. I can switch to a different serialization option if desired after testing and feedback.

### DIFF
--- a/src/NetscriptPort.ts
+++ b/src/NetscriptPort.ts
@@ -8,6 +8,10 @@ const emptyPortData = "NULL PORT DATA";
 /** The object property is for typechecking and is not present at runtime */
 export type PortNumber = PositiveInteger & { __PortNumber: true };
 
+function isObject(value: unknown): value is Object {
+  return (typeof value === "object" && value !== null) || typeof value === "function";
+}
+
 /** Gets the numbered port, initializing it if it doesn't already exist.
  * Only using for functions that write data/resolvers. Use NetscriptPorts.get(n) for */
 export function getPort(n: PortNumber) {
@@ -22,7 +26,8 @@ export class Port {
   data: any[] = [];
   resolver: Resolver | null = null;
   promise: Promise<void> | null = null;
-  resolve() {
+  add(data: any): any {
+    this.data.push(data);
     if (!this.resolver) return;
     this.resolver();
     this.resolver = null;
@@ -43,28 +48,18 @@ export function portHandle(n: PortNumber): NetscriptPort {
 }
 
 export function writePort(n: PortNumber, value: unknown): any {
-  let data = value;
-  // Primitives don't need to be cloned.
-  if ((typeof value === "object" || typeof value === "function") && value !== null) {
-    data = structuredClone(data);
-  }
   const port = getPort(n);
-  port.data.push(data);
-  port.resolve();
+  // Primitives don't need to be cloned.
+  port.add(isObject(value) ? structuredClone(value) : value);
   if (port.data.length > Settings.MaxPortCapacity) return port.data.shift();
   return null;
 }
 
 export function tryWritePort(n: PortNumber, value: unknown): boolean {
-  let data = value;
-  // Primitves don't need to be cloned.
-  if ((typeof value === "object" || typeof value === "function") && value !== null) {
-    data = structuredClone(data);
-  }
   const port = getPort(n);
   if (port.data.length >= Settings.MaxPortCapacity) return false;
-  port.data.push(data);
-  port.resolve();
+  // Primitives don't need to be cloned.
+  port.add(isObject(value) ? structuredClone(value) : value);
   return true;
 }
 
@@ -79,12 +74,8 @@ export function readPort(n: PortNumber): any {
 export function peekPort(n: PortNumber): any {
   const port = NetscriptPorts.get(n);
   if (!port || !port.data.length) return emptyPortData;
-  const value = port.data[0];
   // Needed to avoid exposing internal objects.
-  if ((typeof value === "object" || typeof value === "function") && value !== null) {
-    return structuredClone(value);
-  }
-  return value;
+  return isObject(port.data[0]) ? structuredClone(port.data[0]) : port.data[0];
 }
 
 export function nextPortWrite(n: PortNumber) {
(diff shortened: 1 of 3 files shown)


Snarling:
I don't have strong feelings on this PR (either of the first two main questions), but if I was a player writing a network of scripts that needed objects passed around constantly, I would just use a shared context (either through globals or through shared imports) to share the data directly, instead of cloning the data. But I know some players consider that to be taboo/exploity.

d0sboots:
Edit: Ignore the bobble in the history, that was my mistake.

After some rather extensive benchmarking + ad-hoc testing, this LGTM

In particular, although `structuredClone` has some heavy start-up costs, making it look particularly bad when comparing it against other options for stuff like cloning `{}`, once the objects start getting stuff in them other hand-rolled options start coming down to its performance level.

If people need more speed, they should write their own serialization to string (which is still fast), which can be done much more quickly with specific knowledge of what is being serialized.

LJNeon:
Just realized as a final touch it might be better to create an `isObject()` function instead of repeating the type check three times, but where should it be put? At the top of the file? In the src/types.ts file? Somewhere else?

EDIT: Put it at the top of the file for now and did some code cleanup, ~~waiting for d0sboots to confirm I didn't break anything.~~

LJNeon:
@Snarling This should be good to merge, it's been tested again after the cleanup changes.

src/NetscriptPort.ts
```diff
@@ -8,6 +8,10 @@ const emptyPortData = "NULL PORT DATA";
 /** The object property is for typechecking and is not present at runtime */
 export type PortNumber = PositiveInteger & { __PortNumber: true };
 
+function isObject(value: unknown): value is object {
```
d0sboots: I'd call this "isNonPrimitive", although it's technically true that functions are objects, it's also confusing and what we really care about here is primitive-ness.

d0sboots:
I'll take Snarling's statement of non-caring as being sufficient to merge, modulo the naming thing

src/NetscriptPort.ts
d0sboots: Oh, and because that's a sort of ugly name, you could (optionally) do "isPrimitive" instead and invert the logic.

LJNeon: I think `isObject` more directly describes what the function is doing. I don't really understand why there would be confusion over functions, both the name (`isObject`) and return type (`value is object`) make it very clear what is going on? I could add a comment link to the mdn typeof docs or just explaining that functions are objects if that's acceptable as well?

LJNeon: Also I very intentionally went with isObject instead of isPrimitive because new primitive types have been added several times and that method would break if another was added. I felt it was much safer to check for an object due to that.

d0sboots: > I don't really understand why there would be confusion over functions, both the name (isObject) and return type (value is object) make it very clear what is going on?

So, when I see a function named `isObject` my first guess would be that it's checking (only) `typeof val === "object"`. Which leads to the secondary question of why even have it as a function. There's enough doubt that I will *never* be able to be certain of the exact purpose of a function named `isObject`.

> I felt it was much safer to check for an object due to that.

I wasn't suggesting that you change the way the logic works, just that you put a `!` in front of it. Or the equivalent after applying DeMorgan's law: `value === null || (typeof value !== "object" && typeof value !== "function")`

Anyway, I don't care super much, one way or another I'll merge once I get the chance to re benchmark

LJNeon: > So, when I see a function named `isObject` my first guess would be that it's checking (only) `typeof val === "object"`. Which leads to the secondary question of why even have it as a function. There's enough doubt that I will *never* be able to be certain of the exact purpose of a function named `isObject`.

If the return value was just `boolean` I could see this being a concern, but it's a type predicate that states precisely what the function is doing. What room for doubt does that leave?
I also don't understand the function confusion because I never thought anything of functions being objects, you can add and modify properties so it's clearly not just an intermediary wrapper like primitives get:
```js
function sum(a, b) {
  return a + b;
}
sum.madeBy = "By: ";
console.info(sum.madeBy); // "By: "
sum.madeBy += "John Doe";
console.info(sum.madeBy); // "By: John Doe"
```

LJNeon: And wait, how does your proposed change solve the problem you have? It's common knowledge that everything is either a primitive or an object, so if people think functions aren't objects they must consider them primitives? At which point they will be equally confused regardless of how the function logic is structured, because it always marks functions as objects?

Snarling: I would call it `isObjectLike` because both object and function types can be considered object-like.

In some senses, js functions are literally objects, yes. But when using `typeof` like we are, functions are classified separately from other objects, so just calling them objects is confusing.

d0sboots: It's like if you had a function called "isFruit" which returned `true` for nuts. Yes, nuts are biologically fruits, but that's not how we colloquially classify them. There are multiple ways to look at the meaning of "fruit," and so *anything* called "isFruit" is ambiguous.

Anyway `isObjectLike` or `isPrimitive` both work. Or even `isObject` if you're really set on it. I'll give you a chance to decide and then run final benchmarks/test.

PR #1090, opened 2024-02-11
NETSCRIPT: add formulas.hacking.growAmount()

d0sboots:
People still are getting hung up on the complexities of grow, and we don't actually have an inverse to `growThreads`.

This also improves docs for our grow-related formulas functions, it was very much needed for `growPercent`.


Alpheus:
Is the console.warn intentional in the calculateGrowMoney assertion?

d0sboots:
> Is the console.warn intentional in the calculateGrowMoney assertion?

Yes, it's just moved from the other file. I suppose it is more possible to trigger now (you can pass negative threads or use other out-of-bounds values for server params), but I think that's ok.

Alpheus:
> > Is the console.warn intentional in the calculateGrowMoney assertion?
> 
> Yes, it's just moved from the other file. I suppose it is more possible to trigger now (you can pass negative threads or use other out-of-bounds values for server params), but I think that's ok.

Hrm... should we guard against those on the parameter usage (for Server and HGWOptions mostly)?

d0sboots:
For `Server`, you are always capable of setting the fields to anything you want. You could make them negative, or NaN, or objects. If you ask the API a sufficiently stupid question, you will get a dumb answer.

Formulas functions don't take `HGWOptions`, but those are sanitized to a much tighter standard since hack/grow/weaken actually affect the state.

Alpheus:
> For `Server`, you are always capable of setting the fields to anything you want. You could make them negative, or NaN, or objects. If you ask the API a sufficiently stupid question, you will get a dumb answer.
> 
> Formulas functions don't take `HGWOptions`, but those are sanitized to a much tighter standard since hack/grow/weaken actually affect the state.

Fair, I like that distinction. If I understand you correctly, since formulas are just a "read API" the query in question can be as unsanitary as the user wishes as long as they deal with it—it won't have undue effect on the game state.

Snarling (approved):
Looks good to me!

d0sboots:
After an embarrassingly long time, I'm rebasing and merging this.

PR #1092, opened 2024-02-11
DEV: improve "git blame" for contributors

### LINKED ISSUE #1091: DEV: "git blame" could be better for contributors
_Note: If you don't work on Bitburner itself, then this probably isn't relevant to you and can be safely ignored._

`git blame` - and vscode's Gitlens extension - currently show a lot of commits that could be ignored for the sake of blame.  In particular, there are multiple commits with the description "prettier" that are just reformatting done by the Prettier tool.  (The main one I keep seeing is [this commit](https://github.com/bitburner-official/bitburner-src/commit/a18bdd6afc77752bded2f0794bf35ad9d6455fa1).)

Adding these to a file (e.g. `.git_blame_ignore_revs`) tracked in the repository would allow contributors to ignore these commits for the purposes of blame with this command line:

```
git config --local blame.ignoreRevsFile .git_blame_ignore_revs
```

adeilt:
# DEV: improve "git blame" for contributors

This adds a file containing the commit hashes of commits that are formatting-only (mostly).  It also adds self-contained documentation ~~to `CONTRIBUTING.md`~~ in that file on how to have git use the file to stop assigning blame to the authors of such commits.

The only deliberate exception I made was [this commit](https://github.com/bitburner-official/bitburner-src/commit/8f13363466c2e825bbfb7e4b66d853c4805eda61).  Even though it had a few minor non-trivial changes, the tens of thousands of lines of trivial changes made it worth including in the file in my not-very-strong opinion.  Feel free to change my mind.

# Why

I was working on some other PRs for Bitburner and was trying to determine what the intent was behind some existing lines of code that I wanted to work with and modify.  When inspecting an individual line, I find it helpful to look at the other code from the same commit to get a bigger picture.  However, this is much less effective when the line was most recently modified by one or more large-scale Prettier or lint reformatting commits.

# Doesn't affect anything unless enabled

This is entirely optional.  If you don't enable it explicitly then it won't do anything outside of add a small file, add that file to Prettier's ignore file, and document how to use it if desired.  This PR adds less than 6k to the BB repo.

# Linked issues

fixes #1091

# In context of discussion on #1091

## Arguments for this change

- Blame is more likely to show significant changes *by default*.
- This change offers the option to improve the blame layer for contributors using tools that don't provide an easy way to skip backward through historical blame.
- It doesn't do anything unless enabled.
- Low maintenance will be required as prettier-only commits are no longer being made.

## Argument against this change

- It adds another file to the repo that - in all likelihood - will be used by only a tiny number of people.
- At least one of the commits ignored contains non-prettier changes - a compromise with the fact that tens of thousands of prettier changes were in the same commit.
- I'm pretty sure that everyone reading this has access to the GitHub version of the "walk back to previous blame" feature.


adeilt:
Argh.  Just realized that I should have also searched for "lint".  "Prettier" had around 36 results.  "lint" has over a hundred.  Sigh.

adeilt:
And almost immediately realized upon reviewing the first few commits that lint fixes are much more likely to be actual code changes (like import removal and such).  I'm less inclined to include them here in bulk.  Maybe if there were a few really big ones.

PR #1093, opened 2024-02-12
BUGFIX: few bandaids for Bladeburner

Caldwell-74:
bandaids for 3 bugs in bladeburner
this really needs proper fixes and a alot of refactoring!

1.
the manual action start didnt start tasks the right way, modifying an existing action object instead of creating a new one
therefore the current action wasnt shown on the stats overview

2.
the api start action didnt check for the BladesSimulacrum Aug and didnt stop current Player tasks 
so the next time Bladeburner proccessed it stopped the bladeburner tasks again 
when the player was doing something else like crimes

3.
sometimes blops had an action.count of 0 even when they wherent done in that bladeburner instance yet

this happends because the BlackOps class instances are only initialized on game load 
and then later on BlackOps completion manipulated
this change doesnt reset on a bitnode change or when bladeburner is deleted through the dev Menu
as a quick fix i added a new resetBlackOps function that always runs when Bladeburner processes
this isnt the best solution but any proper fix i came up with requires a refactor that i couldnt do at this moment

credits to @TheAimMan for finding the clue that the count is the problem not the rank!

edit,;
added a 4th bandaid to avoid NaN Stamina Penalty when stamina is infinite

### DIFF
--- a/src/Bladeburner/Bladeburner.tsx
+++ b/src/Bladeburner/Bladeburner.tsx
@@ -111,10 +111,8 @@ export class Bladeburner {
   }
   /*
     just a quick fix for the broken implementation
-    the BlackOperations are only initialized on game load with a count of 1
-    and not reset on BitNode change or dev menu reset
-    this disables the start button for manual play
-    if the game wasnt reloaded since last completion of the Blop
+    BlackOperations are only initialized on game load with a count of 1
+    and are not reset on BitNode change or dev menu reset of bladeburner
   */
   resetBlackOps(): void {
     for (const [blackopName, blackop] of Object.entries(BlackOperations)) {
@@ -127,6 +125,7 @@ export class Bladeburner {
   }
 
   calculateStaminaPenalty(): number {
+    if (this.stamina === this.maxStamina) return 1;
     return Math.min(1, this.stamina / (0.5 * this.maxStamina));
   }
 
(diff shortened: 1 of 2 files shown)


Semanual:
Shouldn't it fix NaN stamina too? At file `src/Bladeburner/Bladeburner.tsx`:
```js
calculateStaminaPenalty(): number {
  return Math.min(1, this.stamina / (0.5 * this.maxStamina));
}
```
To handle infinities (preventing infinity / infinity = NaN):
```js
calculateStaminaPenalty(): number {
  if (this.stamina === this.maxStamina) return 1;
  return Math.min(1, this.stamina / (0.5 * this.maxStamina));
}
```
This was @TheAimMan's actual major breaking problem

Caldwell-74:
oh i thought you would do something for that so i didnt
dont like to just "take" other peoples pr

Semanual:
I would, but then I noticed it would be useless as infinities shouldn't be a thing, so I gave up on that PR. You can still do it if you want to, I don't mind.

TheAimMan:
Fixes https://github.com/bitburner-official/bitburner-src/issues/1096

src/Bladeburner/Bladeburner.tsx
```diff
@@ -109,12 +109,23 @@ export class Bladeburner {
...
+      blackop.count = Number(!this.blackops[blackopName]);
+    }
+  }
 
   getCurrentCity(): City {
     return this.cities[this.city];
   }
 
   calculateStaminaPenalty(): number {
+    if (this.stamina === this.maxStamina) return 1;
```
d0sboots, line 128: Should add a comment here explaining why this is needed (0, Infinity, etc.)

d0sboots (approved):

PR #1094, opened 2024-02-13
BUGFIX: Make hacknet node a bit bigger and change unit to em

shyguy1412:
PR #1070 caused some overflow when having hacknet servers.

pic from @Semanual 
grafik

I made the nodes a little bigger and also changed the unit from px to em.


Semanual:
This should probably be marked as BUGFIX instead of UI.

shyguy1412:
Youre right, did that now

d0sboots:
I'm just going to blindly trust/hope that this is really big enough now for all combinations of browser/font/UI scaling XD

PR #1098, opened 2024-02-15
Documentation added to PersonObjects Directory

rpgroves:
Added documentation to, Person.ts, and PersonMethods.ts describing the properties and methods contained within the file.

Updated the ReadMe to reflect the documentation added in the above files, outlining everything a Person contains and is capable of doing through its methods.

CATEGORY: DOC

(First time submitting a pull request to an open source project! I hope I was helpful!)

### DIFF
--- a/src/PersonObjects/Person.ts
+++ b/src/PersonObjects/Person.ts
@@ -1,17 +1,19 @@
 import type { Skills } from "./Skills";
 import type { HP } from "./HP";
-import type { Person as IPerson } from "@nsdefs";
+import type { Person as IPerson } from "@nsdefs"; // Importing Person interface from namespace @nsdefs
 
-import * as personMethods from "./PersonMethods";
-import { CityName } from "@enums";
-import { PlayerOwnedAugmentation } from "../Augmentation/PlayerOwnedAugmentation";
-import { calculateSkill } from "./formulas/skill";
-import { defaultMultipliers } from "./Multipliers";
-import { IReviverValue } from "../utils/JSONReviver";
+import * as personMethods from "./PersonMethods"; // Importing methods for managing person-related actions
+import { CityName } from "@enums"; // Importing CityName enum
+import { PlayerOwnedAugmentation } from "../Augmentation/PlayerOwnedAugmentation"; // Importing PlayerOwnedAugmentation class
+import { calculateSkill } from "./formulas/skill"; // Importing function for calculating skill levels
+import { defaultMultipliers } from "./Multipliers"; // Importing default multipliers
+import { IReviverValue } from "../utils/JSONReviver"; // Importing interface for JSON reviver values
 
 // Base class representing a person-like object
 export abstract class Person implements IPerson {
-  hp: HP = { current: 10, max: 10 };
+  hp: HP = { current: 10, max: 10 }; // Current and maximum hit points of the person.
+
+  // Skills possessed by the person by level.
   skills: Skills = {
     hacking: 1,
     strength: 1,
@@ -21,6 +23,8 @@ export abstract class Person implements IPerson {
     charisma: 1,
     intelligence: 0,
   };
+
+  // Experience gained towards each skill.
   exp: Skills = {
     hacking: 0,
     strength: 0,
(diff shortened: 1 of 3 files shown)


src/PersonObjects/Person.ts
```diff
@@ -1,17 +1,19 @@
 import type { Skills } from "./Skills";
 import type { HP } from "./HP";
-import type { Person as IPerson } from "@nsdefs";
+import type { Person as IPerson } from "@nsdefs"; // Importing Person interface from namespace @nsdefs
```
d0sboots, line 3: For these imports, they're all mechanical and (depending on your editor) managed semi-automatically, so I don't think they need to be documented. I've certainly never seen an import/include line documented anywhere else, except in some very extreme/odd cases.

(Applies to other files as well)

src/PersonObjects/Person.ts
```diff
@@ -1,17 +1,19 @@
...
+import { CityName } from "@enums"; // Importing CityName enum
+import { PlayerOwnedAugmentation } from "../Augmentation/PlayerOwnedAugmentation"; // Importing PlayerOwnedAugmentation class
+import { calculateSkill } from "./formulas/skill"; // Importing function for calculating skill levels
+import { defaultMultipliers } from "./Multipliers"; // Importing default multipliers
+import { IReviverValue } from "../utils/JSONReviver"; // Importing interface for JSON reviver values
 
 // Base class representing a person-like object
 export abstract class Person implements IPerson {
-  hp: HP = { current: 10, max: 10 };
+  hp: HP = { current: 10, max: 10 }; // Current and maximum hit points of the person.
```
d0sboots, line 14: Sorry if this comes across as a little harsh, but this is IMO a classic example of "just restating the code."

I.e. the comment is not saying anything here that the code does not already say; the variable and field names already say that these are `current` and `max` HP (a very well-understood abbreviation for hit points, honestly the abbreviation is better known than the full phrase).

Most of the time, you want the code to be self-explanatory like this, so that comments aren't needed. Where comments *should* be added are where something tricky is going on, and the code itself is not enough. Even then, the comment should not *restate* what the code does, but rather explain the *motivation* for the code, which is usually the context that is missing.

(Applies to other files as well)

src/PersonObjects/Person.ts
```diff
@@ -31,35 +35,44 @@ export abstract class Person implements IPerson {
...
-  /** Augmentations */
-  augmentations: PlayerOwnedAugmentation[] = [];
-  queuedAugmentations: PlayerOwnedAugmentation[] = [];
+// Multipliers affecting various actions and attributes, represents skills gained and new speeds / rates because of these skills.
+mults = defaultMultipliers();
 
-  /** City that the person is in */
-  city: CityName = CityName.Sector12;
+/** Augmentations */
+augmentations: PlayerOwnedAugmentation[] = []; // List of augmentations owned by the person
```
d0sboots, line 42: These two are solid explanatory comments for these fields. 👍They explain the difference between the two in a way that the variable name alone does not convey. Keep these, but I feel like everything else up until here (and after here, until the next comment) should be reverted as already self-explanatory.

src/PersonObjects/Person.ts
```diff
@@ -31,35 +35,44 @@ export abstract class Person implements IPerson {
...
 
-  abstract takeDamage(amt: number): boolean;
-  abstract whoAmI(): string;
-  abstract toJSON(): IReviverValue;
+/** Reset all multipliers to 1 */
+resetMultipliers() {
+  this.mults = defaultMultipliers();
 }
+
+// Abstract method for taking damage amount amt. Returns true if person is alive after damage and false otherwise.
```
d0sboots, line 68: For methods, you should use JSDoc syntax if you're going to document them. https://jsdoc.app/about-getting-started
Don't feel the need to go overboard or document every parameter.

You don't need to mention "abstract", that's already part of the signature.

This function is a good use of comment: The return value is non-obvious.

src/PersonObjects/Person.ts
```diff
@@ -31,35 +35,44 @@ export abstract class Person implements IPerson {
...
+abstract takeDamage(amt: number): boolean;
+
+// Abstract method for identifying the person.
+// Returns A string representing the person's identity.
+abstract whoAmI(): string;
+
+// Abstract method for serializing the person to JSON format.
+// Returns The serialized representation of the person.
+abstract toJSON(): IReviverValue;
+}
```
d0sboots, line 78: Your editor lost the newline at the end-of-file

src/PersonObjects/PersonMethods.ts
```diff
@@ -104,6 +108,9 @@ export function gainIntelligenceExp(this: Person, exp: number): void {
     this.skills.intelligence = Math.floor(this.calculateSkill(this.exp.intelligence, 1));
   }
 }
+
+// Increases all skills' experience based on the given WorkStats and updates the corresponding skill levels.
```
d0sboots, line 112: Similar comment about using JSDoc (which is mostly just a matter of changing the comment type)

src/PersonObjects/PersonMethods.ts
```diff
@@ -193,4 +203,4 @@ export function hasAugmentation(this: Person, augName: string, ignoreQueued = fa
     return true;
   }
   return false;
-}
+}
```
d0sboots, line 206: newline at end-of-file

src/PersonObjects/README.md
```diff
@@ -1,2 +1,16 @@
 Implementation of all Person-type objects, including but not limited to
 the "PlayerObject" and Sleeves.
+
+Person.ts:
+Person.ts serves as a foundational piece for modeling characters within the system, 
+providing a structure for defining their attributes, skills, and behavior.
+
+A Person contains the following main attibutes:
```
d0sboots, line 8: I would cut everything from here to the end out. You're just re-iterating stuff that exists in the code, and what you say here will inevitably become out-of-date.

What you added before this is a good start (or you can leave it at just that) for what to add to the README: an overall sense of design/purpose. Here is for the high-level overview, the details are for the individual files.

d0sboots (changes requested):
Thanks for contributing!

Since you're doing this for a class, I'm going to be a bit more strict (and leave a lot more comments) than I typically would (especially for a first-time contributor), so apologies in advance. XD

Snarling (commented):
Generally agree with most of the feedback d0s provided here, but I would say about 90% of the comments added here are not needed / are documenting code that is already easily understood.

Code should generally be self-documenting, by just using variable names that are straightforward and descriptive and doing things in a straightforward way. In such cases, comments don't really add anything.

Comments should instead be used when something is not clear, for instance:

1. When the code is not clear as to what is occurring, e.g. complex math stuff
2. When some part of how a function works is not clear from its name and function signature, e.g. it returns a number but it's not clear from the function name what that returned number represents
3. If an odd approach is intentionally being used for something, a comment can explain why it was done that way

The above is not an exhaustive list, but the general idea is that code should be written to be easily understood, and if the code is easily understood then there is no need for separate commenting to explain what the code is doing.

If code is not clear because the variable names are bad, then if possible the better fix is to change the variable names so the code is easily understood. An example of this would be the parameter `retValue` in the `gainStats` function. `retValue` does not convey what that parameter actually is, and the odd naming just serves to confuse the reader. If it was just renamed to a more sensible name like `workStats`, `gains`, or `expGains`, the function is easily understood without need for a comment to explain what that function does, since it's clear from the function name and signature.

Snarling:
Closing PR, see previous comments

PR #1101, opened 2024-02-16
DOC: Fix outdated example for ns.spawn()

adeilt:
# DOC: Fix outdated example for ns.spawn()

The existing example for `ns.spawn()` is outdated and this PR updates it.  Also mentions the 10s default `spawnDelay` which is not otherwise in the description of `spawn` anymore.

### Background

During the course of #807 which made `spawn`'s now default 10s delay configurable (thanks!), the example was added and was accurate at that time.  Later in the PR, the call signature of `spawn()` changed, but the example was apparently overlooked for update.  This brings the example in line with the current implementation of `spawn`.


adeilt:
Fun fact: I discovered this while "passing the baton".  :)  [see this comment](https://github.com/bitburner-official/bitburner-src/pull/807#issuecomment-1722586622) on #807.

PR #1102, opened 2024-02-16
UI: Reimplement #954 (Servers sorted on active scripts page)

Snarling:
The changes made in https://github.com/bitburner-official/bitburner-src/pull/1083 were not compatible with https://github.com/bitburner-official/bitburner-src/pull/954. Making changes based on https://github.com/bitburner-official/bitburner-src/pull/954 was not practical due to the scope of changes so I reimplemented the functionality here.

The sorting method here was written to minimize iteration of the base array, which is why I took a different approach to the comparison logic.

* Added an isHacknetServer optional attribute to BaseServers
* HacknetServer has a truthy value added for the new attribute to make comparisons easier
* ActiveScriptsPage gets the server objects because they are needed for sorting. ActiveScriptsPage new passes the actual server object to the ServerAccordion, instead of just the hostname.
* Changed a temporary data structure from a regular object to a Map, so it can be keyed by actual server objects instead of hostnames.

image

Sorting: `home` is first, then all hacknet servers in numerical order, then all other purchased servers in alpha order, then all other servers unsorted

### DIFF
--- a/src/Hacknet/HacknetServer.ts
+++ b/src/Hacknet/HacknetServer.ts
@@ -53,6 +53,8 @@ export class HacknetServer extends BaseServer implements IHacknetNode {
   // Flag indicating whether this is a purchased server
   purchasedByPlayer = true;
 
+  isHacknetServer = true;
+
   constructor(params: IConstructorParams = { hostname: "", ip: createRandomIp() }) {
     super(params);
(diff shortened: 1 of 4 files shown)


TheAimMan:
Ship it!

PR #1104, opened 2024-02-16
MISC: clamping numbers

Caldwell-74:
just a small start for the big project of clamping numbers 
more to provide a place for discussions

like visible here my way would be to clamp values in key spots with a helper function


src/utils/helpers/clampNumbers.ts
```diff
@@ -0,0 +1,11 @@
...
+ * caps the value on a lower and an upper bound
+ * @param {number} value value to cap
+ * @param {number} min lower bound
+ * @param {number} max upper bound
+ * @returns {number} capped value
+ */
+function clampNumber(value:number,min=Number.MIN_VALUE,max= Number.MAX_VALUE){
+    return Math.max(Math.min(value,max),min)
+}
+export default clampNumber
```
d0sboots: Missing newline at EOF

d0sboots: I don't think we should use the default import syntax for this. Given that, we shouldn't define it, either, and it would be better to just declare this `export function` directly.

src/utils/helpers/clampNumbers.ts
```diff
@@ -0,0 +1,11 @@
+/**
+ * caps the value on a lower and an upper bound
```
d0sboots: Since you're introducing a new utility function, I'm going to be more picky about comments than I might be otherwise. :D

Please capitalize.

d0sboots: You should document (and also think about) what the function returns when `value` is NaN. Right now the answer is `NaN`, which is consistent but probably not what we want. Minimum and maximum would both be sensible, but I'm not sure which would be generally better.

src/PersonObjects/formulas/skill.ts
```diff
@@ -1,37 +1,33 @@
+import clampNumber from "../../utils/helpers/clampNumbers"
```
d0sboots: Optional: I have a slight preference for `util/helper/clampNumber` (singular) over plural, and I think that fits the style of the existing code better?

d0sboots (commented):
Nice, I like it!

As is usual with my reviews, I tend to leave a lot of comments. Most of them are minor/nitpicks.

src/utils/helpers/clampNumbers.ts
Caldwell-74: i think throwing is the right way for NaNs
getting those out of the system is imo even more important than preventing Infinities

d0sboots:
You still have this marked as a draft, but I think this could be shipped as-is.

Pinging @Snarling for judgement.

Caldwell-74:
yes i think adding more files to this is not needed
ill just open new prs then
will undraft it

Snarling (approved):
Looks good to me. I think throwing for NaN on the dev branch but just returning min otherwise was a good call.

If we run into cases where returning min doesn't make sense as a default return for NaN inputs, we could add a 4th param to clampNumber: `default = min` so that it can be overridden to default to something else. But we can make that adjustment if/when that kind of edge case comes up.

PR #1105, opened 2024-02-17
BUGFIX: fix current HP from being NaN when a stat is infinite

TheAimMan:
Current HP is a ratio of Current/Max.  If Max is infinite that ratio is invalid so current becomes NaN

image

Changing the ratio to return 1 if NaN allows for the calculation to succeed

image

### DIFF
--- a/src/PersonObjects/PersonMethods.ts
+++ b/src/PersonObjects/PersonMethods.ts
@@ -180,7 +180,7 @@ export function updateSkillLevels(this: Person): void {
     Math.floor(this.calculateSkill(this.exp.charisma, this.mults.charisma * currentNodeMults.CharismaLevelMultiplier)),
   );
 
-  const ratio: number = Math.min(this.hp.current / this.hp.max, 1);
+  const ratio: number = Math.min(this.hp.current / this.hp.max, 1) || 1;
   this.hp.max = Math.floor(10 + this.skills.defense / 10);
   this.hp.current = Math.round(this.hp.max * ratio);
 }
--- a/src/PersonObjects/PersonMethods.ts
+++ b/src/PersonObjects/PersonMethods.ts
@@ -181,7 +181,7 @@ export function updateSkillLevels(this: Person): void {
     Math.floor(this.calculateSkill(this.exp.charisma, this.mults.charisma * currentNodeMults.CharismaLevelMultiplier)),
   );
   //In case MaxHP is Infinite, setting ratio to 1
-  const ratio = Number.isFinite(this.hp.current) ? this.hp.current / this.hp.max : 1;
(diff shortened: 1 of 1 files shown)


src/PersonObjects/PersonMethods.ts
```diff
@@ -44,7 +44,8 @@ export function gainDefenseExp(this: Person, exp: number): void {
   }
 
   this.skills.defense = calculateSkill(this.exp.defense, this.mults.defense * currentNodeMults.DefenseLevelMultiplier);
-  const ratio = this.hp.current / this.hp.max;
+  //In case MaxHP is Infinite, setting ratio to 1
+  const ratio = this.hp.current / this.hp.max || 1;
```
d0sboots: This doesn't work, it sets the ratio to 1 when it should be 0.

This is a case of protecting against an Infinity where (IMO) there shouldn't be one (health should not be infinite), *but* I'm OK with defense in depth so do it like this:
```suggestion
  const ratio = Number.isFinite(this.hp.current) ? this.hp.current / this.hp.max : 1;
```
That will also catch NaNs in the current hp.

src/PersonObjects/PersonMethods.ts
```diff
@@ -181,7 +181,7 @@ export function updateSkillLevels(this: Person): void {
     Math.floor(this.calculateSkill(this.exp.charisma, this.mults.charisma * currentNodeMults.CharismaLevelMultiplier)),
   );
   //In case MaxHP is Infinite, setting ratio to 1
-  const ratio: number = Math.min(this.hp.current / this.hp.max, 1) || 1;
+  const ratio = Number.isFinite(this.hp.current) ? this.hp.current / this.hp.max : 1;
```
d0sboots: You lost the Math.min here (looking back, it looks like there are/were edge-cases where HP could go over max and this is needed to deal with that).

d0sboots: *poke* Still interested in this one?

src/PersonObjects/PersonMethods.ts
```diff
@@ -181,7 +181,7 @@ export function updateSkillLevels(this: Person): void {
     Math.floor(this.calculateSkill(this.exp.charisma, this.mults.charisma * currentNodeMults.CharismaLevelMultiplier)),
   );
   //In case MaxHP is Infinite, setting ratio to 1
-  const ratio = Number.isFinite(this.hp.current) ? this.hp.current / this.hp.max : 1;
+  const ratio = Math.min(Number.isFinite(this.hp.current) ? this.hp.current / this.hp.max : 1);
```
d0sboots, line 184: Now you are just taking the min of a single number XD

PR #1107, opened 2024-02-18
SETTINGS: add tailRenderInterval

Caldwell-74:
currently tails are on a static rerender time of 1000ms
there are ways to force a rerender with ns.moveTail / ns.resizeTail 
but i dont know why this shouldnt be a setting set by the player

currently its static set on creation of the tail window but that could be expanded to allow setting it per script 
and maybe through a ns function

image


src/GameOptions/ui/SystemPage.tsx
```diff
@@ -132,6 +138,20 @@ export const SystemPage = (): React.ReactElement => {
           tooltip={<>The time (in seconds) between each autosave. Set to 0 to disable autosave.</>}
           marks
         />
+        <OptionsSlider
+          label="Tail render intervall (ms)"
+          initialValue={tailrenderInterval}
+          callback={handleTailIntervalChange}
+          step={200}
+          min={200}
```
d0sboots: I think we should have min go down to 100, possibly even 50. People might come up with interesting animations or something at high rerenders, and I don't think it should hurt performance too badly (we'll test). But the important part is that it's a setting, so it's users choice if they want to be extreme.

src/GameOptions/ui/SystemPage.tsx
```diff
@@ -132,6 +138,20 @@ export const SystemPage = (): React.ReactElement => {
           tooltip={<>The time (in seconds) between each autosave. Set to 0 to disable autosave.</>}
           marks
         />
+        <OptionsSlider
+          label="Tail render intervall (ms)"
```
d0sboots:
```suggestion
          label="Tail render interval (ms)"
```

src/GameOptions/ui/SystemPage.tsx
```diff
@@ -132,6 +138,20 @@ export const SystemPage = (): React.ReactElement => {
...
+          initialValue={tailrenderInterval}
+          callback={handleTailIntervalChange}
+          step={200}
+          min={200}
+          max={5 * 1000}
+          tooltip={
+            <>
+              The minimum number of milliseconds between two tail rerender. Setting this too low can result in poor
```
d0sboots:
```suggestion
              The minimum number of milliseconds between tail rerenders. Setting this too low can result in poor
```

src/GameOptions/ui/SystemPage.tsx
```diff
@@ -132,6 +138,20 @@ export const SystemPage = (): React.ReactElement => {
...
+              performance if you have many Tail windows open.
```
d0sboots:
```suggestion
              performance if you have many tail windows open.
```

src/Settings/Settings.ts
```diff
@@ -72,6 +72,8 @@ export const Settings = {
   ExcludeRunningScriptsFromSave: false,
   /**  Whether the game's sidebar is opened. */
   IsSidebarOpened: true,
+  /** Tail rendering intervall in ms */
+  TailRenderIntervall: 1000,
```
d0sboots:
```suggestion
  TailRenderInterval: 1000,
```
Also anywhere else I missed, you made this misspelling several times.

d0sboots (commented):
Could you attach a screenshot of the new settings? Just to verify it isn't overflowing and everything looks proper.

Testing that it works at a faster speed is the only non-nitpick I've got.

Change looks good, I look forward to merging and not seeing the port doc changes anymore XD

src/GameOptions/ui/SystemPage.tsx
Caldwell-74: then with a smaller step or is there a way to change the step size depending on the current value? (like 50-200 steps of 10, 200-2000 steps of 200, 2000-5000 steps of 1000 or so)

Caldwell-74:
ive tested it with 200 and 5000 yesterday and it seemed to work as expected with its limitations

src/GameOptions/ui/SystemPage.tsx
d0sboots: I'm pretty sure step size has to be fixed (limitation of the actual DOM slider). Options would be to decrease the step size, or it might work out if you leave step at 200 since initial is 1000, giving steps at 1000, 800, 600, 400, 200, 50?

Caldwell-74: i tried min 50 with 200 steps 
sadly ends with 50-250-450  etc after the first move away from 1000

d0sboots: Dang. Well, go with whatever you think works best then. I'd lightly prefer the ability to have a smaller min refresh, but I'm not dead set on it.

Caldwell-74: i hoped "we" could get a ns method at some point to set it to whatever the player wants to use
(and to reapply a changed setting on already existing tails?)
but thats above my react level

d0sboots: IMO we'd want that to apply per-tail and not affect the global setting, but I'm okay with that as a justification for the global only going to 200.

Caldwell-74: oh yes totaly meant as per tail like the other tail functions

PR #1112, opened 2024-02-20
BUGFIX: Prevent renaming servers to hacknet.

### LINKED ISSUE #1111: Servers can be renamed into the hacknet namespace.
When buying a purchased there's a check to ensure that the server name doesn't fall into the possible hacknet server names. However this check is not present when renaming purchased servers.

LJNeon:
fixes #1111

### DIFF
--- a/src/NetscriptFunctions.ts
+++ b/src/NetscriptFunctions.ts
@@ -1167,6 +1167,10 @@ export const ns: InternalAPI<NSFull> = {
       helpers.log(ctx, () => `Invalid argument: hostname='${hostnameStr}'`);
       return "";
     }
+    if (hostnameStr.startsWith("hacknet-node-") || hostnameStr.startsWith("hacknet-server-")) {
+      helpers.log(ctx, () => `Invalid argument: hostname='${hostnameStr}' is a reserved hostname.`);
+      return "";
+    }
 
     if (Player.purchasedServers.length >= getPurchaseServerLimit()) {
       helpers.log(
--- a/src/Server/ServerPurchases.ts
+++ b/src/Server/ServerPurchases.ts
@@ -125,7 +125,7 @@ export function purchaseServer(hostname: string, ram: number, cost: number): voi
     return;
   }
 
-  if ((hostname.startsWith("hacknet-node-") || hostname.startsWith("hacknet-server-")) {
+  if (hostname.startsWith("hacknet-node-") || hostname.startsWith("hacknet-server-")) {
     dialogBoxCreate(`'${hostname}' is a reserved hostname, please try again.`);
     return;
   }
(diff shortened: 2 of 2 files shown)


yichizhng:
For consistency I would suggest making purchaseServer also throw this error - currently it goes to the server constructor, which replaces the name with a random 10-character string.

LJNeon:
The reason I didn't do that is because it looks like the Server class is used in a few places and some of them definitely need the current check (like purchasing a server manually). Also the code that uses the Server class isn't setup to handle errors like the code that uses the rename function is.

EDIT: Search for `safelyCreateUniqueServer` and `new Server` to see what I mean.

LJNeon:
Nevermind I looked at it again and most of the uses weren't exposed to the player, so it was actually easy to add validation to the two that are and then remove the check from the Server constructor.

d0sboots:
There's a bunch of repeated code here, but that's a project for a larger refactor to share all the error handling logic for these.

LJNeon:
So I don't need to concern myself with it now?

src/Server/Server.ts
```diff
@@ -59,11 +58,6 @@ export class Server extends BaseServer {
   constructor(params: IConstructorParams = { hostname: "", ip: createRandomIp() }) {
     super(params);
 
-    // "hacknet-node-X" hostnames are reserved for Hacknet Servers
```
d0sboots: You can't remove this code. I'm not confident I've tracked down *all* the places the constructor gets called from (it's a lot), but at minimum this can be invoked with old savefiles that have purchased servers named "hacknet-node-X", due to the bug you are fixing or other reasons.

LJNeon: As far as I can tell, the `Server` class is only constructed by two functions: [`safelyCreateUniqueServer()`](https://github.com/bitburner-official/bitburner-src/blob/dev/src/Server/ServerHelpers.ts#L41) and [`initForeignServers()`](https://github.com/bitburner-official/bitburner-src/blob/dev/src/Server/AllServers.ts#L156). The later will never touch a purchased server so it's not a concern here, and the former is only called in [four places](https://github.com/search?q=repo%3Abitburner-official%2Fbitburner-src%20path%3A%2F%5Esrc%5C%2F%2F%20safelyCreateUniqueServer&type=code). That doesn't seem like "a lot" to me?

LJNeon: Also that's a good catch, I'm assuming that it should just create a random hostname like it did before?

d0sboots: Yeah, just revert this part and keep it for safety. I like the more explicit checks you added elsewhere, so this should only be needed as a fallback.

d0sboots: Reverted via force-push.

PR #1113, opened 2024-02-20
BUGFIX: Close deleted scripts on install.

### LINKED ISSUE #1050: Scripts on other servers can be restored after installing augments
It is possible to recover scripts onto servers that you don't have access to after installing augments. I noticed it yesterday. I had a script from `silver-helix` server open in editor. I have installed augments resetting progress. Script was still open in editor, I tried to close it and it asked me if I want to save changes. I clicked "Yes" and played normally. Now after getting to that server again, I see that this script is already here. 

## Steps to reproduce
1. Create a script on one of the servers except `home`
2. Leave script open in editor
3. Install augments resetting your progress (or do a soft reset. It will probably work the same)
4. Go to editor and save the script again. It will save the script to that server even if it's locked for you after resetting and you don't have access to it.

## Possible fix
Close all scripts in editor from servers that are not a home server.

## Game version
2.5.2 (c6141f2ad) Steam edition

LJNeon:
There was some disagreement on if this fix is necessary, see the related issue for details.

fixes #1050

### DIFF
--- a/src/ScriptEditor/ui/ScriptEditorRoot.tsx
+++ b/src/ScriptEditor/ui/ScriptEditorRoot.tsx
@@ -56,13 +56,15 @@ interface IProps {
 const openScripts: OpenScript[] = [];
 let currentScript: OpenScript | null = null;
 
-export function closeDeletedScripts(onlyHome: boolean): void {
-  for (let i = openScripts.length - 1; i >= 0; i--) {
-    const {hostname} = openScripts[i];
-    if (GetServer(hostname) === null || (onlyHome && hostname !== "home")) {
-      openScripts.splice(i, 1);
+export function closeDeletedScripts(allServers: boolean): void {
+  let idx = 0;
+  for (const script of openScripts) {
+    openScripts[idx] = script;
+    if (GetServer(script.hostname) !== null && (allServers || script.hostname === "home")) {
+      ++idx;
     }
   }
+  openScripts.length = idx;
   if (currentScript && GetServer(currentScript.hostname) === null) {
     currentScript = openScripts[0] ?? null;
   }
@@ -183,7 +185,7 @@ function Root(props: IProps): React.ReactElement {
   let decorations: monaco.editor.IEditorDecorationsCollection | undefined;
 
   // Prevent Crash if script is open on deleted server
-  closeDeletedScripts(false);
+  closeDeletedScripts(true);
 
   const save = useCallback(() => {
     if (currentScript === null) {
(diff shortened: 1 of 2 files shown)


src/ScriptEditor/ui/ScriptEditorRoot.tsx
```diff
@@ -40,6 +40,18 @@ interface IProps {
 const openScripts: OpenScript[] = [];
 let currentScript: OpenScript | null = null;
 
+export function closeDeletedScripts(onlyHome: boolean): void {
+  for (let i = openScripts.length - 1; i >= 0; i--) {
+    const {hostname} = openScripts[i];
+    if (GetServer(hostname) === null || (onlyHome && hostname !== "home")) {
+      openScripts.splice(i, 1);
+    }
+  }
```
d0sboots: Optional: It always bugs me a little to write a N^2 delete algorithm when the O(N) algorithm is basically as simple, even in places where the performance doesn't matter.
```suggestion
  let replaceIdx = 0;
  for (const script of openScripts) {
    openScripts[replaceIdx] = script;
    if (GetServer(script.hostname) !== null && (allServers || script.hostname === "home")) {
      replaceIdx++;
    }
  }
  openScripts.length = replaceIdx;
```

d0sboots (commented):
I think this change makes sense, but it'll need good testing because this whole area is a bunch of edge cases.

src/ScriptEditor/ui/ScriptEditorRoot.tsx
LJNeon: Done

LJNeon:
What edge cases are there? This seems like a simple enough function to me, it closes open scripts from deleted servers and updates the currently open script accordingly. The behavior has also not changed at all except for what it does on prestige (install/soft reset/backdooring wd).

EDIT: Not trying to argue this doesn't need to be tested, just not sure if it's as complicated as you said.

d0sboots:
I was referring to the game itself, not so much this code. The bug this is fixing is an edge case, and there are just a variety of cases of "you do/don't have certain windows open at certain times" that should be checked. Off the top of my head, the full matrix of:
* Script is/isn't deleted
* Script is/isn't on home
* Open window is/isn't the currently connected host

looked at across installs (or some other reset, they should all act the same), and also whatever it is that triggers the other codepath that calls your function

LJNeon:
The other trigger ~~was added by caldwell to solve a bug afaik,~~ (nope, was remembering wrong) it runs every time the script editor tab is opened.

d0sboots:
Note to possibly future self: Based on Discord history, this might be on hold pending Monaco shenanigans discovered during testing

PR #1114, opened 2024-02-21
UI: Infiltration rewards default to the faction you're currently working for

LJNeon:
As [requested in the suggestions channel](https://discord.com/channels/415207508303544321/415213435974975508/1209338928423768074) on Discord, this changes the default faction in the drop-down on the victory screen to be the faction the player is currently working for when possible.

This change needs to be tested.

### DIFF
--- a/src/Infiltration/ui/Victory.tsx
+++ b/src/Infiltration/ui/Victory.tsx
@@ -81,12 +81,10 @@ export function Victory(props: IProps): React.ReactElement {
       <Box sx={{ width: "fit-content" }}>
         <Box sx={{ width: "100%" }}>
           <Select value={factionName} onChange={changeDropdown} sx={{ mr: 1 }}>
-            {defaultFaction === "none" ? (
+            {defaultFaction === "none" && (
               <MenuItem key={"none"} value={"none"}>
                 {"none"}
               </MenuItem>
-            ) : (
-              <></>
             )}
             {Player.factions
               .filter((f) => Factions[f].getInfo().offersWork())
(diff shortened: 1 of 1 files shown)


src/Infiltration/ui/Victory.tsx
```diff
@@ -79,9 +81,11 @@ export function Victory(props: IProps): React.ReactElement {
       <Box sx={{ width: "fit-content" }}>
         <Box sx={{ width: "100%" }}>
           <Select value={factionName} onChange={changeDropdown} sx={{ mr: 1 }}>
-            <MenuItem key={"none"} value={"none"}>
-              {"none"}
-            </MenuItem>
+            {defaultFaction === "none" && (
```
d0sboots, line 84: I think we should keep the "none" option unconditionally, just for consistency/maybe someone wants to throw away rep for some mad reason

d0sboots:
Looks good, aside from testing

src/Infiltration/ui/Victory.tsx
LJNeon: I disagree on the need for consistency (why include a none option when it isn't necessary) but I can understand that argument, however I'm not sure why we would enable throwing away rep even if people want to. If they don't care where it goes they can just let it go to the default faction, which is most likely the best one to pick.

d0sboots: Ok, the button doesn't actually function when none is selected, so it's not like a usable (even if silly) option is being removed.

Just waiting on testing

Semanual:
Testing:
Case 1: When doing infiltration without joining any faction, it shows up only "none":
imagem

Case 2: When doing infiltration after joining a faction, it selects "none" as default, and lets you choose another faction:
imagem

Case 3: When doing infiltration while working for a faction (tested with multiple working types), it selects the working faction as default, and does not let you choose none, but lets you choose any other faction:
imagem

Case 4: When doing infiltration while working for a company that has a faction, it selects none as default:
imagem

So, probably working as intended?

LJNeon:
Thank you for testing this!

PR #1115, opened 2024-02-21
IPVGO: ensure the special world demon opponent is handled correctly

ficocelliguy:
Ensure the world demon is reset on ascension, and is a valid opponent for the API, by simplifying its usage and access.

### DIFF
--- a/src/Go/boardAnalysis/goAI.ts
+++ b/src/Go/boardAnalysis/goAI.ts
@@ -4,6 +4,7 @@ import {
   Move,
   MoveOptions,
   opponentDetails,
+  opponentList,
   opponents,
   PlayerColor,
   playerColors,
@@ -825,3 +826,7 @@ export function sleep(ms: number): Promise<void> {
 export function showWorldDemon() {
   return Player.augmentations.some((a) => a.name === AugmentationName.TheRedPill) && Player.sourceFileLvl(1) > 0;
 }
+
+export const getOpponentList = () => {
+  return opponentList.filter((o) => (showWorldDemon() ? true : o !== opponents.w0r1d_d43m0n));
+};
--- a/src/Go/boardAnalysis/goAI.ts
+++ b/src/Go/boardAnalysis/goAI.ts
@@ -823,5 +823,5 @@ export function sleep(ms: number): Promise<void> {
 }
 
 export function showWorldDemon() {
-  return Player.augmentations.some((a) => a.name === AugmentationName.TheRedPill) && Player.sourceFileLvl(1);
+  return Player.augmentations.some((a) => a.name === AugmentationName.TheRedPill) && Player.sourceFileLvl(1) > 0;
 }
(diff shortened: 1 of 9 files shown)


src/Go/ui/GoHistoryPage.tsx
```diff
@@ -21,7 +21,7 @@ export const GoHistoryPage = (): React.ReactElement => {
   const priorBoard = Player.go.previousGameFinalBoardState ?? getNewBoardState(7);
   const score = getScore(priorBoard);
   const opponent = priorBoard.ai;
-  const opponentsToShow = showWorldDemon() ? [...opponentList, opponents.w0r1d_d43m0n] : opponentList;
+  const opponentsToShow = getOpponentList(showWorldDemon());
```
d0sboots: I don't think you ever use `getOpponentList` with a first arg that *isn't* `showWorldDaemon`. You could move the helper to goAI (constants is a strange place for a function anyway) and eliminate the first arg.

src/Go/ui/GoSubnetSearch.tsx
```diff
@@ -24,16 +24,15 @@ export const GoSubnetSearch = ({ open, search, cancel, showInstructions }: IProp
     opponent === opponents.w0r1d_d43m0n ? 19 : Math.min(Player.go.boardState?.board?.[0]?.length ?? 7, 13);
   const [boardSize, setBoardSize] = useState(preselectedBoardSize);
 
-  const opponentFactions = [opponents.none, ...opponentList];
-  if (showWorldDemon()) {
-    opponentFactions.push(opponents.w0r1d_d43m0n);
-  }
+  const opponentFactions = getOpponentList(showWorldDemon(), true);
```
d0sboots: I think this is the only place where you use the 2nd arg? I think it would be slightly clearer to just push the no opponent here.

src/Go/boardAnalysis/goAI.ts
```diff
@@ -823,5 +823,5 @@ export function sleep(ms: number): Promise<void> {
 }
 
 export function showWorldDemon() {
-  return Player.augmentations.some((a) => a.name === AugmentationName.TheRedPill) && Player.sourceFileLvl(1);
+  return Player.augmentations.some((a) => a.name === AugmentationName.TheRedPill) && Player.sourceFileLvl(1) > 0;
```
d0sboots, line 827: Is this fixing a bug? I don't see how this is any different, which has me worried.

d0sboots (commented):
Some minor questions.

src/Go/ui/GoHistoryPage.tsx
ficocelliguy: Good call, updated

src/Go/boardAnalysis/goAI.ts
ficocelliguy: This is just to make the method explicitly return a boolean (instead of the previous `boolean | number` type) to make typescript happy

Snarling:
This will conflict with #1120 and the changes from 1120 already cover the same things in a different way, so I will probably need to just close this without merging. @ficocelliguy running that by you before I close your PR

PR #1119, opened 2024-02-22
GANG: Clarify install behavior & add getInstallResult()

### LINKED ISSUE #1073: Bug: Installing Augmentations or performing a soft reset reduces gang members ascension multipliers
When installing augmentations or performing a soft reset, I've noticed that the gang member multipliers from when they've Ascended are reduced by ~2.53% each time.

I believe it's an issue because of the text on the Gang screen

"Installing Augmentations does NOT reset progress with your Gang."

The image below showing a before and after of performing a soft reset.

image

This is really noticed when installing augmentations in quick succession and can quickly decimate time spent training their stats, as Charisma is particularly slow.

## Version 
Bitburner v2.5.2 (c6141f2ad)

## Steps to reproduce
1. Join a gang
2. Recruit a gang member
3. Train the gang member so that they can gain an ascension multiplier
4. Take note of the multipliers 
5. Perform a soft reset or install augmentations on the player
6. Take note of the multipliers afterwards

LJNeon:
fixes #1073

### DIFF
--- a/src/NetscriptFunctions/Gang.ts
+++ b/src/NetscriptFunctions/Gang.ts
@@ -310,7 +310,7 @@ export function NetscriptGang(): InternalAPI<IGang> {
         def: member.calculateAscensionMult(postInstall.def) / preInstall.def,
         dex: member.calculateAscensionMult(postInstall.dex) / preInstall.dex,
         agi: member.calculateAscensionMult(postInstall.agi) / preInstall.agi,
-        cha: member.calculateAscensionMult(postInstall.cha) / preInstall.cha
+        cha: member.calculateAscensionMult(postInstall.cha) / preInstall.cha,
       };
     },
     setTerritoryWarfare: (ctx) => (_engage) => {
--- a/src/NetscriptFunctions/Gang.ts
+++ b/src/NetscriptFunctions/Gang.ts
@@ -310,7 +310,7 @@ export function NetscriptGang(): InternalAPI<IGang> {
         def: member.calculateAscensionMultiplier(postInstall.def) / preInstall.def,
         dex: member.calculateAscensionMultiplier(postInstall.dex) / preInstall.dex,
         agi: member.calculateAscensionMultiplier(postInstall.agi) / preInstall.agi,
-        cha: member.calculateAscensionMultiplier(postInstall.chs) / preInstall.cha
+        cha: member.calculateAscensionMultiplier(postInstall.cha) / preInstall.cha
       };
     },
     setTerritoryWarfare: (ctx) => (_engage) => {
--- a/src/NetscriptFunctions/Gang.ts
+++ b/src/NetscriptFunctions/Gang.ts
@@ -303,7 +303,7 @@ export function NetscriptGang(): InternalAPI<IGang> {
       const member = getGangMember(ctx, memberName);
       if (!member.canAscend()) return;
       const preInstall = member.getCurrentAscensionMults();
-      const postInstall = member.getInstallResults();
+      const postInstall = member.getPostInstallPoints();
       return {
         hack: member.calculateAscensionMult(postInstall.hack) / preInstall.hack,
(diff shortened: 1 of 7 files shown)


src/NetscriptFunctions/Gang.ts
```diff
@@ -297,6 +297,22 @@ export function NetscriptGang(): InternalAPI<IGang> {
         ...member.getAscensionResults(),
       };
     },
+    getInstallResult: (ctx) => (_memberName) => {
+      const memberName = helpers.string(ctx, "memberName", _memberName);
+      getGang(ctx);
```
d0sboots, line 302: What is the purpose of this, you don't use the result?

src/NetscriptFunctions/Gang.ts
```diff
@@ -297,6 +297,22 @@ export function NetscriptGang(): InternalAPI<IGang> {
         ...member.getAscensionResults(),
       };
     },
+    getInstallResult: (ctx) => (_memberName) => {
```
d0sboots, line 300: Why does the netscript function `getInstallResult` do something different from the gang function `getInstallResults`? It feels like either one of those is misnamed, or misimplemented.

src/NetscriptFunctions/Gang.ts
LJNeon: With `getGang()` it's just doing the same thing all the other functions do, that method throws if the player doesn't have a gang. And same goes for the method names, I just did the exact same thing `getAscensionResult()` does, plural in the gang member function and singular in the player exposed function.

src/NetscriptFunctions/Gang.ts
LJNeon: If you have issues with either of these then I would need to change getAscensionResult and other methods too, because I just matched the behavior and naming style of those.

src/NetscriptFunctions/Gang.ts
d0sboots: OK, in that case I'd throw a comment on it here saying it's called for error-checking only.

src/NetscriptFunctions/Gang.ts
d0sboots: It's not though? `getAscensionResult` is returning literally the result of `member.getAscensionResults`, except with `earnedRespect` tacked on. Here, you're doing this transformation where the gang function returns the post-install result but the netscript function returns a ratio.

LJNeon: Ah. that's just shifted over, look at the function body of getAscensionResults on GangMember. I shifted that over because the install usage needs the raw points while the player exposed method needs the multipliers. getAscensionResult doesn't need to care about raw points.

src/NetscriptFunctions/Gang.ts
LJNeon: Should I add that comment to every function then?

d0sboots: Ah I see, it is actually a common pattern among several of the other functions. I guess nothing needs to change then.

src/NetscriptFunctions/Gang.ts
d0sboots: Yes, that's my entire point. Gang has `getAscensionMultsAfterAscend` for the raw values (not entirely needed, because it's only used in one place, but it's fine) and `getAscensionResults` returns the ratios. And then this matches the NS function.

My argument is that the gang function you're adding ought to match the NS function. (Or if it doesn't, use a different name.)

LJNeon: Honestly I don't like how it looks either but I couldn't think of a better way and it's the closest possible match to existing behavior in the twin function (getAscensionResult).

d0sboots: How about `getPostInstallPoints`?

LJNeon: Done.

PR #1120, opened 2024-02-22
GO: Various changes for 2.6.0

Snarling:
* Moved Go to top level scope object, instead of living on Player
* Custom save/load routine, to avoid huge data structures in save file and allow typesafety during loading. For an example of this see the differences in the FullSave test (which didn't even have board history included, which is the biggest savegame reduction)
* Lots of variable renaming and some reorganization, at least for the stuff that was previously in boardState/goConstants
* Game no longer stores history as full boards, now just saves the one previous board, as a simple board (no analytics).
* Less reassignment of Go.currentGame (previously Player.go.boardState), allowing equality of this variable to be used as a test for whether it is still the same game (e.g. after awaiting during async functions).
* Go.stats (previously Player.go.status) no longer contains all opponent entries from the getgo, now only contains entries when there is relevant data to store for the opponent.

There are some changes outside of go files as well. This is due to some changes I had to make to accommodate Jest, which included moving some netscript error handling functions to a new file, and I also renamed some of those functions at the same time.

This also replaces #1115 since the issues addressed there are also fixed here, and due to the scope of changes here there is conflict between the two.


ficocelliguy:
This looks good! One minor thought regarding your todos: We currently show the final state of the current subnet on the status page and main go board for the player to review if desired. I also show the previous game's final position on the history page. 

Sometimes games can end fairly abruptly, and I feel it is important for new players to be able to look at the final board state to better understand what worked (or to debug/improve their own ipvgo scripts). Having the prior game's endstate especially is helpful when determining if your script is doing well. These states can be kept in the save object in a simpler way, though, it doesn't need all of its analytics baked in like it currently has.

Snarling:
Yeah I messed around a little bit trying to get the boardState to be nullable, but I ran into some issues and might just put that off for now and revisit at a later date. Mostly just didn't like that for a player who hasn't interacted with the mechanic at all during a run, it creates a board for them in their save file.

But on the topic of slimming down the save file, I have a couple thoughts I'll try out but I'm not sure if I will run into any problems here:

boardState.history: Board[] is pretty heavy when we might be talking about each board being possibly 13x13 and each point on the board being an object with a bunch of data itself.  An array of moves would make more sense (x, y, color) as a way to track history. But if there's a bunch of logic that relies on history being a Board[] right now, I might have trouble getting some of it to work with a format change.

boardState.board: I am thinking about not storing the actual PointState[][] but just storing something similar to the simplified board here, just showing ownership of points instead of all the other analytic data (chains, etc). If we store the board in a simplified state, is there anything else we need to store to faithfully recreate the board after loading the save, or can all the stuff be calculated afterwards (chains, liberties, etc)?

Snarling:
dc7c2dc fixes a weird bug with unexpected TS behavior. It looks like when we reference one enum while defining another, TS automatically makes reverse lookup entries which is not wanted. It's assumed for string enums that all values are actually enum members, but this weird TS behavior broke that assumption. Here's GoOpponent before the fix:

image

Snarling:
@ficocelliguy I believe this should be fully functional again now if you want to look it over before I merge it. Otherwise I'll go ahead and merge it and wait a few days / a week before actually pushing the 2.6.0 release.

Also @d0sboots FYI in case you want to take a look. I also changed some stuff outside Go during this PR because it was causing issues after I moved Go to a top level object. Biggest non-go thing is I moved the home location of the helpers.makeRuntimeErrorMsg function: now it's called errorMessage and it's still available on helpers but actually lives in a different file.

test/jest/__snapshots__/FullSave.test.ts.snap
```diff
@@ -548,6 +548,29 @@ exports[`Check Save File Continuity FactionsSave continuity 1`] = `
 }
 `;
 
+exports[`Check Save File Continuity GoSave continuity 1`] = `
```
d0sboots, line 551: Optional: Maybe have something other than a *completely* fresh game to test more of the save/load surface?

d0sboots (commented):
Change *all* the files! XD

Snarling:
After merging #1124, I messed around with some stuff, mostly to get rid of the eslint overrides.

First I tried using destructuring (95466d9), and this worked but it looked really stupid. This was just a fake way of making it seem like these functions were actually relying on e.g. previousPlayer.

I modified a bunch of the function signatures so they indicated whether they needed a whole boardState or just the board. Turns out the relevant functions just needed the board.

I removed the periodic rerender and found that rerenders were only being triggered when a move was completed. But with the periodic rerender missing, netscript interactions with Go would not trigger rerenders, so I added an EventEmitter and made netscript trigger that event when manipulating the board. Now rerendering is only performed when the board is actually updated (which requires running these calculations that were previously memoized), so there was no need for memoization and I removed the useMemos.

Snarling:
Merging this, followup changes can be done via another PR

PR #1124, opened 2024-02-25
IPVGO: add previousPlayer as a dependency to useMemo so it correctly updates when the board is updated

ficocelliguy:
React `useMemo()` only updates if an object dependency is completely re-assigned, not if it is updated. This can be fixed by passing the primitive `boardState.previousPlayer` as a dependency to useMemo(). ESLint doesn't understand why this is needed, so some disable-next-lines have been added, along with a comment explaining it.

### DIFF
--- a/src/Go/ui/GoGameboard.tsx
+++ b/src/Go/ui/GoGameboard.tsx
@@ -22,12 +22,19 @@ export function GoGameboard({ boardState, traditional, clickHandler, hover }: Go
   const currentPlayer =
     boardState.ai !== GoOpponent.none || boardState.previousPlayer === GoColor.white ? GoColor.black : GoColor.white;
 
+  // "boardState.previousPlayer" is added as a useMemo dependency because useMemo only does pointer comparison for
+  // determining when objects change, so a primitive has to be used to correctly update it
   const availablePoints = useMemo(
     () => (hover ? getAllValidMoves(boardState, currentPlayer) : []),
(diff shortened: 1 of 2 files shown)


d0sboots:
The correct way of handling this in React is to use immutable objects, so the object identity *does* change when its fields change. I highly recommend using that approach if possible, instead of disabling the lints.

Snarling:
@d0sboots While that makes sense for React-specific objects, it doesn't make sense for the actual game internals. Go.currentGame shouldn't become a whole new reference every time a move is being made. Prior to #1120  it was doing this because the entire state of the current game was being cloned after every move and then the object reassigned, but that's very inefficient given how bulky a BoardState object is.

I don't like the lint disable either though, I'll see if there's another easy way to get this functionality.

d0sboots:
Yup, that's a downside of how we're directly using the game objects in our React components and the two models don't mesh well.

Snarling:
Removed the useMemo entirely in #1120. The memoization was only needed because of the periodic auto-rerender, which was only needed due to the board needing to respond to netscript updates (playing go through the UI already triggers rerenders at only the appropriate times). I played around with it some more in #1120 and after adding an EventEmitter for netscript to trigger UI updates at appropriate times, I was able to remove the periodic rerender and remove the memozation entirely (the entire rerender is now only triggering during times when the previous memoization would trigger)

PR #1125, opened 2024-02-26
MISC: change ls to accept relative paths

Caldwell-74:
snarling added a good chunk of relative path resolving for functions like fileExists
ns.ls wants to filter based on a substring so it doesnt directly expect a path

i dont think we need to care about "absolute" paths because those should be fine as substring for filter
with this relative paths get resolved to provide the correct base directory

i have tested the edgecases i came up with but a second test is appreciated because i dont fully understand the regexp stuff


d0sboots:
I think this is fine, but
1. This is technically a breaking change (although I doubt it will break any/many people) but
2. Therefore, it should at least be documented in the `ls()` function page.

d0sboots:
Actually, the more I am looking at the code here the more I am having 2nd thoughts.

Previously, it was a simple "includes" check. This means you could do wildcard-type searches by including the directory (or part of the directory), or you could search for part of a filename, etc.

Now, what you have is both less expressive than that, *and* I'm not sure it'll work correctly.

Caldwell-74:
image
i dont actually think thats is a breaking change because isAbsolutePath if i understand the regexp correctly looks for "./" or "../"
for filepaths these have to be at the start to work and the game (tried nano) resolves them correctly

if the ls substring has ./ or ../ not at the start but in it then doesnt work 
if its at the start then it doesnt work currently but the pr would resolve that
if it doesnt have them the substring isnt changed and works like currently

Caldwell-74:
i would curios where this would break anything because it still allows the partial directory/filename/substring filter


edit.: its kinda hard to make testcase with filestructure but if you have a good idea for it let me know and i ll run through it

Snarling:
Actually I don't think this breaks anything in a meaningful way.

isAbsolutePath is just checking whether a generic string passes a regex to detect relative paths and returns the negation of that. So if isAbsolutePath returns false, then the filter matches the relative path regex. That's the only case where behavior is being changed.

The relative path regex matches any time where `./` or `../` are either at the start of a string, or immediately after another `/` somewhere else in the string. It should be impossible for an actual FilePath to match that relative path regex, so if a player was already sending in a relative path as their filter for `ls` then they would have always received an empty array back. Now, instead of receiving an empty array for this relative path filter, it will be treated as a relative path based on the script's directory. So technically in this case it's a breaking change, but no player would have been sending in valid relative paths as the filter before this since they cannot return anything other than `[]`.

The major issue I still see here is that if the player's supplied relative path does not properly resolve, resolveDirectory returns `null`, and  then their filter is thrown out and ls just returns everything instead of nothing. In these cases (player provided what looks like a relative path but it doesn't resolve correctly), we should either return empty array (nothing matches the invalid filter) or throw an error telling them their relative path was bad.

Caldwell-74:
ill modify it to throw when that happens
an empty array might cause confusion

d0sboots:
OK, unfortunately @Snarling is wrong here about relative paths not being able to match.

```
[home /]> nano blar./foo.js
[home /]> run eval.js "ns.ls('home', './foo')"
Running script with 1 thread(s), pid 2299686 and args: ["ns.ls('home', './foo')"].
eval.js: ["blar./foo.js"]
```

(While reading LJ's PR where you mentioned the trickiness with extensions, I realized there are no limitations on dots in directiories.)

Snarling:
Ah yup, I forgot ending a directory with a . was allowed.

d0sboots:
So, taking a step back: what do we want to do here?

On one hand, I feel like being able to support relative paths would be good.

On the other hand, I feel like the existing state of ls() makes this uniquely bad. For usability and quick implementation it went with substring search instead of file glob patterns, which works OK on the commandline and at least is simple/consistent to explain as an API, but doesn't interact with actual path logic well at all.

PR #1126, opened 2024-02-26
TESTS: Prevent some test spam

Snarling:
Just modifying some tests to reduce console spam and make it easier to see actual errors from the test.


Snarling:
On a previous PR, running tests produced 2205 lines of output with no actual test failures:
image

Now completing in just 34 lines of output when there are no test failures
image

PR #1127, opened 2024-02-27
HACKNET: Disallow negative `count` argument for spendHashes

yichizhng:
Tested manually.

This is a very exploitable bug; consider the following script:

```js
  for (let i = 0; i < 1000; ++i) {
    ns.hacknet.spendHashes('Increase Maximum Money', 'joesguns', -1);
    ns.hacknet.spendHashes('Increase Maximum Money', 'ecorp', 1);
  }
}
```

This boosts the max money of ecorp for no cost (it doesn't decrease joesguns max money, in fact).

### DIFF
--- a/src/NetscriptFunctions/Hacknet.ts
+++ b/src/NetscriptFunctions/Hacknet.ts
@@ -203,7 +203,7 @@ export function NetscriptHacknet(): InternalAPI<IHacknet> {
         const upgTarget = helpers.string(ctx, "upgTarget", _upgTarget);
         const count = Math.floor(helpers.number(ctx, "count", _count));
         if (!(count >= 0)) {
-          throw helpers.errorMessage(ctx, 'count may not be negative');
+          throw helpers.errorMessage(ctx, "count may not be negative");
         }
         if (!hasHacknetServers()) {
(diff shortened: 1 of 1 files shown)


Snarling:
Could should actually use helpers.positiveInteger, as non-integer numbers of upgrades also don't make sense. Although that would cause an error message for a count of 0.

yichizhng:
Done (although I kind of don't like the throw on count=0, because I had a script that didn't check...)

Snarling:
Hmmm. Maybe it's better that we hold off on that change until 3.0 API break and just go with what you had initially. There are probably also people with scripts that send in decimal values when spending hashes for money, and just rely on the function to floor the number for them.

I think accepting only positive integers is the _right_ API, but it is an API break for us to no longer accept 0 or decimal values and it's bad to break people's scripts. Then again, the error message from positiveInteger should help the player easily fix their own script if it was sending in something else.

I'm good with either way, which would you prefer @yichizhng ?

yichizhng:
Changed it back to the previous implementation.

d0sboots:
Merging based on severity and the PR seeming settled.

PR #1128, opened 2024-02-28
BUGFIX: ns.getServerPortNumber now returns undefined on hacknet servers instead of 5

muesli4brekkies:
Little fix for a discrepancy pointed out on the discord.

image

pic of the issue

This PR just changes the function to return `undefined` rather than `5`, in line with the warning message the function prints, and property on ns.getServer for hacknet servers. It also tweaks the docs a little to reflect that the function can return `undefined`

### DIFF
--- a/src/ScriptEditor/NetscriptDefinitions.d.ts
+++ b/src/ScriptEditor/NetscriptDefinitions.d.ts
@@ -3969,7 +3969,7 @@ export interface Go {
    * @remarks
    * RAM cost: 4 GB
    *
-   * @returns a promise that contains if your move was valid and successful, the opponent move's x and y coordinates (or pass) in response, or an indication if the game has ended
+   * @returns a promise that contains the opponent move's x and y coordinates (or pass) in response, or an indication if the game has ended
    */
   makeMove(
     x: number,
@@ -3987,7 +3987,7 @@ export interface Go {
    * This can also be used if you pick up the game in a state where the opponent needs to play next. For example: if BitBurner was
    * closed while waiting for the opponent to make a move, you may need to call passTurn() to get them to play their move on game start.
    *
-   * @returns a promise that contains if your move was valid and successful, the opponent move's x and y coordinates (or pass) in response, or an indication if the game has ended
+   * @returns a promise that contains the opponent move's x and y coordinates (or pass) in response, or an indication if the game has ended
    *
    * @remarks
    * RAM cost: 0 GB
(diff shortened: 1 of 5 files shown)


src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -6422,7 +6422,7 @@ export interface NS {
    * @param host - Hostname of target server.
    * @returns The number of open ports required to successfully run NUKE.exe on the specified server.
```
d0sboots: This should mention the circumstances where it returns undefined. Other than that, LGTM

muesli4brekkies: This did cross my mind. Is there a way to make it clear in a non-spoilery way? This is a function new players are likely to come across.

d0sboots: "Returns undefined for servers that don't have openable ports"?

It's a little ambiguous, because arguably the same applies to personal servers and home.

Snarling: If it should return undefined for all player-owned servers, then using the "failOnHacknetServer" function is not great.

Really it's the internal structure that kind of sucks, where BaseServer doesn't have those properties. `Server` is basically a "hackable" server, but home and regular purchased servers are also this type of server even though they're not hackable. Hacknet servers inherit from BaseServer instead of Server, so they end up missing a bunch of properties specific to hackable servers.

Not sure what the best solution is here.

muesli4brekkies: Fair points. 

Is it worth putting this on the back-burner, in the event a more thorough rework of these systems could be planned? 

I doubt it would be much work to make things consistent but I suspect those would be breaking changes.

d0sboots: IMO it's worth making this more consistent, even if it's not perfect. The situation with hashnet vs pservers vs home extends to much more than just open ports, and it's definitely a larger issue.

d0sboots:
Before I merge this, I will throw out the other possibility, which is to go the opposite direction: make things consistent by having getServer return 5 for ports on hacknet servers.

muesli4brekkies:
That would work also. I'm not sure what's better. Over this weekend I'll chuck together that PR as well and see which one you guys like best :D

PR #1129, opened 2024-02-29
CONTRACTS: createDummyContract returns filename

Spartelfant:
# createDummyContract() method now returns filename…

…of generated dummy contract.

This tiny quality-of-life improvement means you don't have to search for the dummy contract you just created, instead its filename is simply returned by the method used to create it.

PR #1130, opened 2024-02-29
createDummyContract() method now returns filename

Spartelfant:
# createDummyContract() method now returns filename…

…of generated dummy contract.

Documentation update for pull request #1129

### DIFF
--- a/src/ScriptEditor/NetscriptDefinitions.d.ts
+++ b/src/ScriptEditor/NetscriptDefinitions.d.ts
@@ -3577,8 +3577,9 @@ export interface CodingContract {
    * Generate a dummy contract on the home computer with no reward. Used to test various algorithms.
    *
    * @param type - Type of contract to generate
+   * @returns Filename of the contract.
    */
-  createDummyContract(type: string): void;
+  createDummyContract(type: string): string;
 
   /**
    * List all contract types.


Snarling:
First, thanks for contributing! But there are a few issues with this:

1. The changes should all be in the same PR, not spread across two different PRs.  You can do this by making another commit on your existing branch `patch-1`, instead of here where you made a new branch `patch-2`. Any commits to `patch-1` will automatically be added to the existing PR #1129 you have open for that branch.
2. You have a title error, that's just because we expect the PR title to be in a specific format. Something like `CATEGORY: Description of change` is expected, with an all-uppercase category with no spaces. At this time there are not strict rules on what categories are acceptable. For instance you could name your PR `CONTRACTS: createDummyContract returns a filename`.
3. You have a build error, this is due to the actual netscript function createDummyContract not returning a string even though your documentation says it is supposed to return a string. This isn't just due to the previous commit not being part of this PR, it's also because you never changed the createDummyContract function in `NetscriptFunctions/CodingContract.ts`, you only changed the internal parent function generateDummyContract.
4. Any time documentation is changed, it's also best practice to run `npm run doc` to regenerate new standalone documentation files. For this one you would definitely need a whole dev environment set up if you don't already.

If you'd like to fix this, I will let you try to figure it out. Let me know if you need any assistance with any of that.

I'll close this PR and post basically the same comment over on your other PR.

PR #1131, opened 2024-03-01
API: Add karma to ns.getPlayer() & document ns.heart.break()

### LINKED ISSUE #931: No Karma information available
Bitburner v2.5.0 (b87b8b4be)

----

The only mention of "karma" is at the tooltip of the "Create Gang" button. It tells me, that I need -54000 karma to unlock. But there is no information about players current karma level anywhere. This should be displayed at Stats page and should be accessible through the Player interface (https://github.com/bitburner-official/bitburner-src/blob/dev/markdown/bitburner.player.md).
And at the Gang documentation should use the term "karma" in the "Starting and Recruiting" section to make the topic more clear.

LJNeon:
Implemented as discussed on Discord.

fixes #931

### DIFF
--- a/src/NetscriptFunctions.ts
+++ b/src/NetscriptFunctions.ts
@@ -1683,6 +1683,7 @@ export const ns: InternalAPI<NSFull> = {
       jobs: structuredClone(Player.jobs),
       factions: Player.factions.slice(),
       entropy: Player.entropy,
+      karma: Player.karma,
     };
     setDeprecatedProperties(data, {
       playtimeSinceLastAug: {
--- a/src/NetscriptFunctions/Formulas.ts
+++ b/src/NetscriptFunctions/Formulas.ts
@@ -108,6 +108,7 @@ export function NetscriptFormulas(): InternalAPI<IFormulas> {
       jobs: {},
       factions: [],
       entropy: 0,
+      karma: 0,
     }),
     mockPerson: () => (): IPerson => ({
       hp: { current: 0, max: 0 },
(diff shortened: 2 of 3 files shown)


d0sboots:
For this to be complete, I think heart.break needs to be moved out of the hidden functions and into the regular ones.

LJNeon:
You want a src/NetscriptFunctions/Heart.js` file for just one function? Also just because it's documented now doesn't mean it's not an easter egg? I thought keeping it as an easter egg was part of why we didn't make a new documented method?

d0sboots:
No, I'm not asking for a whole separate file. But as long as it remains one of the hidden functions, intellisense won't show it and it is nearly undiscoverable. When I said I liked it still being an Easter egg, I was referring to the name.

It's much less pressing now, because with karma as a property of player there's another way for people to find it.

LJNeon:
But it's not hidden and it does show up on intellisense? That's the whole point of adding it to the definitions file?

Snarling:
I was confused by this at first too, with it being given an entry in the NS type definition, but still being in Extra.ts. Extra.ts is really for undocumented functions - where every other NetscriptFunctions file's contents are required to be represented in the NS type definition, that's not true for Extra.ts, which is why we use that file for all the hidden functions.

The function still being in Extra.ts is probably what confused d0s here as well.

I made a followup PR #1133 for just moving it to the main NetscriptFunctions.ts file.

LJNeon:
Ah my bad, at least it was a trivial fix.

PR #1132, opened 2024-03-04
MISC: Correct rm command help text

LJNeon:
The `rm` command can delete all file types except for `.msg`, however the help text incorrectly states that `.msg` files can be deleted and doesn't even mention `.txt` or `.lit` files.


LJNeon:
For reference I tried to match the wording used in the docs for `ns.rm()`.

PR #1135, opened 2024-03-06
API: Minor followup changes for #1059

Snarling:
@shyguy1412 @d0sboots FYI

killWorkerScript:
* Callbacks are already guaranteed to be functions -> removed typecheck
* Loop through map directly instead of looping through keys only and then looking up each callback through the map
* Show id of callback if there is an error while running it

TypeAssertion:
* Added an assertion helper for player-provided functions

NetscriptFunctions:
* Use helper functions for type validation
* Set callback directly instead of wrapping in an anonymous function. No `this` leakage possible with new format.


d0sboots:
Oh cool, I had thought of some of those changes but hadn't said anything out loud.

PR #1138, opened 2024-03-06
EDITOR: general improvement of developing experience

shyguy1412:
This keeps the loaded models in sync with the filesystem while the editor is open and is an improvement of 
This means typings from imported files can be loaded, even if that file is not currently open.
I also added support for ctrl+clicking imports/identifiers to open the source file in a new tab when clicked

firefox_WOitrdTAIF

I also made it so that the editor updates automatically if a script or the RFA writes to a file, given that there are no unsaved changes currently

firefox_wJE3enEZKi

### DIFF
--- a/src/ScriptEditor/ui/Editor.tsx
+++ b/src/ScriptEditor/ui/Editor.tsx
@@ -4,6 +4,8 @@ import * as monaco from "monaco-editor";
 
 import { useScriptEditorContext } from "./ScriptEditorContext";
 import { scriptEditor } from "../ScriptEditor";
+import { GetAllServers } from "../../Server/AllServers";
+import { makeModel } from "../Model";
 
 interface EditorProps {
   /** Function to be ran after mounting editor */
@@ -25,6 +27,12 @@ export function Editor({ onMount, onChange, onUnmount }: EditorProps) {
     if (!containerDiv.current) return;
     // Before initializing monaco editor
     scriptEditor.initialize();
+
+    for (const server of GetAllServers()) {
+      for (const [path, script] of server.scripts.entries()) {
+        makeModel(server.hostname, path, script.code);
+      }
+    }
 
     // Initialize monaco editor
     editorRef.current = monaco.editor.create(containerDiv.current, {
--- a/src/ScriptEditor/ui/ScriptEditorRoot.tsx
+++ b/src/ScriptEditor/ui/ScriptEditorRoot.tsx
@@ -22,7 +22,7 @@ import { PromptEvent } from "../../ui/React/PromptManager";
 
 import { useRerender } from "../../ui/React/hooks";
 
-import { dirty, getServerCode, makeModel } from "./utils";
+import { dirty, getServerCode } from "./utils";
 import { OpenScript } from "./OpenScript";
 import { Tabs } from "./Tabs";
 import { Toolbar } from "./Toolbar";
@@ -33,6 +33,7 @@ import { useCallback } from "react";
 import { type AST, getFileType, parseAST } from "../../utils/ScriptTransformer";
 import { RamCalculationErrorCode } from "../../Script/RamCalculationErrorCodes";
 import { hasScriptExtension, isLegacyScript } from "../../Paths/ScriptFilePath";
+import { makeModel } from "../Model";
 
 interface IProps {
   // Map of filename -> code
@@ -313,8 +314,7 @@ function Root(props: IProps): React.ReactElement {
         },
       });
     }
-    //unmounting the editor will dispose all, doesnt hurt to dispose on close aswell
-    closingScript.model.dispose();
+
     openScripts.splice(index, 1);
     if (openScripts.length === 0) {
       currentScript = null;
--- a/src/Server/BaseServer.ts
+++ b/src/Server/BaseServer.ts
@@ -23,7 +23,10 @@ import { getKeyList } from "../utils/helpers/getKeyList";
 import lodash from "lodash";
 import { Settings } from "../Settings/Settings";
 
+import monaco from "monaco-editor";
+
 import type { ScriptKey } from "../utils/helpers/scriptKey";
+import { getModel, makeModel } from "../ScriptEditor/Model";
 
 interface IConstructorParams {
   adminRights?: boolean;
@@ -180,6 +183,9 @@ export abstract class BaseServer implements IServer {
       const textFile = this.textFiles.get(path);
       if (!textFile) return { res: false, msg: `Text file ${path} not found.` };
       this.textFiles.delete(path);
+      //dispose of editor model
+      const model = getModel(this.hostname, path);
+      if (model) model.dispose();
       return { res: true };
     }
     if (hasScriptExtension(path)) {
@@ -188,6 +194,9 @@ export abstract class BaseServer implements IServer {
       if (this.isRunning(path)) return { res: false, msg: "Cannot delete a script that is currently running!" };
       script.invalidateModule();
       this.scripts.delete(path);
+      //dispose of editor model
+      const model = getModel(this.hostname, path);
+      if (model) model.dispose();
       return { res: true };
     }
     if (hasProgramExtension(path)) {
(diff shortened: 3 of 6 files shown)


src/Server/BaseServer.ts
```diff
@@ -247,45 +256,41 @@ export abstract class BaseServer implements IServer {
   }
 
   /**
-   * Write to a script file
-   * Overwrites existing files. Creates new files if the script does not exist.
+   * Write to a Script or TextFile
+   *    * Overwrites existing files. Creates new files if the script does not exist.
    */
-  writeToScriptFile(filename: ScriptFilePath, code: string): writeResult {
```
d0sboots, line 253: The two functions were separate for type-checking reasons. I'm leery of just mashing them together. I think you can accomplish the same code-sharing goals by having a private impl that takes care of the shared work.

shyguy1412: In that case wouldn't it make sense to add the functions back in but have them call writeContentFile? The actual implemention of these functions was identical beside the different variable names so I dont really see a reason not to combine them.
Both textfile and script file inherit the content file interface which should be the interface used to manage the content of these files.

Snarling: It's just a bit inefficient to combine these. If we already know we're dealing with a TextFilePath, then we can directly use writeToTextFile and we're not having to call hasTextExtension again to decide what type of file to write. Same with if we already know it's a ScriptFilePath, doing an extra hasTextExtension check is not actually needed, but if we just have a combined writeToContentFile function then there's no way to escape performing this extra check.

If the code we're in doesn't already know which type of filepath we're dealing with, that would be when it makes sense to run another hasTextExtension check to determine which type of file to write.

Snarling (commented):
The added functionality here looks great. Looks like it's failing some tests right now, and also some console errors in the showcase images that should be addressed.
Let me know when this is ready for closer review.

shyguy1412:
I did a force push to clean up some messy commits.
Everything should be fine now, the console errors are unrelated to the changes I made. They came from my own broken scripts.

Some further testing from someone with a larger codebase would probably be good to make sure there isnt a substantial perfomance drop when generating the models for all scripts.

Snarling:
What were the bugs you ran into with this?

shyguy1412:
The tabs are very unhappy with the way ctrl+click opens new files. The current tab gets set weirdly and switching back to the tab you came from causes it to get added to the openScripts array again, but in index -1 causing a bunch of weird errors.
The main thing, having a model for each file and updating them when they change and dont have unsaved changes, works perfectly fine according to my tests.

shyguy1412:
I removed the features that @tomprince is currently working on so now this just keeps models for all scripts so they dont need to be open in order to get type info. I tested this with larger codebases and didnt notice a perfomance change

src/Server/BaseServer.ts
d0sboots: This part is still unresolved. I don't think you even need this change anymore for what you're doing?

shyguy1412: Models are also created for text files, so when a text file is open in the editor and written to it should still update. That's why I kept in the unification of those methods, its removes a lot of imo uneeded code duplication

src/Server/BaseServer.ts
```diff
@@ -246,46 +255,55 @@ export abstract class BaseServer implements IServer {
...
+    return this.writeToScriptFile(path, content);
   }
 
-  // Write to a text file
-  // Overwrites existing files. Creates new files if the text file does not exist
-  writeToTextFile(textPath: TextFilePath, txt: string): writeResult {
-    // Check if the text file already exists, and overwrite if it does
-    const existingFile = this.textFiles.get(textPath);
-    // overWrite if already exists
+  #writeContentToFile<T extends ContentFilePath>(
```
d0sboots, line 276: A couple things about this:
1. Rereading this, I agree that it's reached a point where it makes sense to have a helper method. Which I guess is why I suggested it originally. XD
2. It doesn't actually need to be private. The only member you use off `this` that I see is `hostname`, I think it'll work out better to just pass that as well and then have it be a plain-old (genericized) function, as opposed to any kind of method.

src/Server/BaseServer.ts
```diff
@@ -246,46 +255,55 @@ export abstract class BaseServer implements IServer {
...
+    path: T,
+    content: string,
+    files: Map<string, ContentFile>,
+    ContentFile: new (path: T, content: string, host?: string) => ContentFile,
```
d0sboots, line 280: `ContentFile` is a type, this is reusing the same name as a constructor which gets awfully ambiguous (since the constructor is *not* creating a literal ContentFile, but rather a subtype of it.)

I'd call this parameter `ctor` instead.

src/Server/BaseServer.ts
```diff
@@ -246,46 +255,55 @@ export abstract class BaseServer implements IServer {
...
+  ) {
+    const existingFile = files.get(path);
+
+    //Try to get the editor model
+    const model = getModel(this.hostname, path);
+
+    //If model exists and there are no unsaved changes, update
+    if (model && existingFile?.content == model.getValue()) model.setValue(content);
```
d0sboots, line 288:
```suggestion
    if (model && existingFile?.content === model.getValue()) {
      model.setValue(content);
    }
```

src/Server/BaseServer.ts
```diff
@@ -246,46 +255,55 @@ export abstract class BaseServer implements IServer {
...
+    //We know path must be the right path
+    const newFile = new ContentFile(path, content, this.hostname);
+    files.set(path, newFile);
 
-  /** Write to a Script or TextFile */
-  writeToContentFile(path: ContentFilePath, content: string): writeResult {
-    if (hasTextExtension(path)) return this.writeToTextFile(path, content);
-    return this.writeToScriptFile(path, content);
+    //if editor is open, create model for new file
+    if (monaco.editor?.getModels().length && path.__fileType == "Script") {
```
d0sboots, line 302: Once you have one `?.`, you need more or you'll get null dereferences. Or maybe you don't actually need the `?.` at all (if this has never crashed?)

src/Server/BaseServer.ts
```diff
@@ -188,6 +194,9 @@ export abstract class BaseServer implements IServer {
       if (this.isRunning(path)) return { res: false, msg: "Cannot delete a script that is currently running!" };
       script.invalidateModule();
       this.scripts.delete(path);
+      //dispose of editor model
+      const model = getModel(this.hostname, path);
+      if (model) model.dispose();
```
d0sboots, line 199:
```suggestion
      getModel(this.hostname, path)?.dispose();
```

src/Server/BaseServer.ts
```diff
@@ -180,6 +183,9 @@ export abstract class BaseServer implements IServer {
       const textFile = this.textFiles.get(path);
       if (!textFile) return { res: false, msg: `Text file ${path} not found.` };
       this.textFiles.delete(path);
+      //dispose of editor model
+      const model = getModel(this.hostname, path);
+      if (model) model.dispose();
```
d0sboots, line 188: If you want to avoid a multi-line if, I'd do
```suggestion
      getModel(this.hostname, path)?.dispose();
```

src/ScriptEditor/ui/Editor.tsx
```diff
@@ -26,6 +28,12 @@ export function Editor({ onMount, onChange, onUnmount }: EditorProps) {
     // Before initializing monaco editor
     scriptEditor.initialize();
 
+    for (const server of GetAllServers()) {
+      for (const [path, script] of server.scripts.entries()) {
```
d0sboots, line 32: I have some concerns about this. Does it make sense to create models for *every* script across *every* server? Although technically you can be editing code across multiple servers at once, it's a confusing and suboptimal experience. (On the other hand, have scripts on every server, due to automatic copying, is quite common.)

src/ScriptEditor/ui/ScriptEditorRoot.tsx
```diff
@@ -313,8 +314,7 @@ function Root(props: IProps): React.ReactElement {
         },
       });
     }
-    //unmounting the editor will dispose all, doesnt hurt to dispose on close aswell
-    closingScript.model.dispose();
```
d0sboots, line 317: If I'm reading this right, script files are now meant to exist as models permanently. Text files, however, are not: They get models when they are opened. These models aren't ever disposed of, though.

shyguy1412:
closing this cause I dont use the internal editor myself and honestly cant bring up any motivation to work on this anymore x.x

PR #1139, opened 2024-03-06
GO: hide buttons until discovered

Caldwell-74:
iam not sure if GO is supposed to be visible before entering BN14 or having SF14 
but currently the button is shown in the Locations

this adds a small check if it should be shown or not

might want to change the playerHasDiscoveredGo function aswell force check BN14 || SF14 
ill do that if it should be hidden if not then ill just close this


d0sboots:
@ficocelliguy
As I understand it, this is intentional. Whether or not that's best for the game... I dunno? But IIUC it's not a bug.

Caldwell-74:
yeah its a bit wierd to make a new mechanic with a new bn and then make it available from the get go
even more with all those factions unlocked right away

PR #1140, opened 2024-03-06
IPVGO: Fix self-capture move evaluation

ficocelliguy:
Fix for issue breaking player go scripts found here:
https://discord.com/channels/415207508303544321/415213413745164318/1214987456076709898

### DIFF
--- a/src/Go/boardAnalysis/boardAnalysis.ts
+++ b/src/Go/boardAnalysis/boardAnalysis.ts
@@ -485,7 +485,7 @@ export function findAllCapturedChains(chainList: PointState[][], playerWhoMoved:
   }
 
   const friendlyChainsToCapture = findCapturedChainOfColor(chainList, playerWhoMoved);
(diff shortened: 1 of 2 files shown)


Snarling:
Thanks for the quick fix!

Just made a small change to remove the `?.` that wasn't needed (enemyChainsToCapture is always an array, so doesn't need a ?. to access the length property)

Using `?.` when it is not needed can lead to confusion when people are trying to read the code later. TS is pretty good at determining whether something is possibly nullish, so in the majority of cases it will tell you if the value could be undefined and needs a `?.` for property access.

The exception is when a member of an array is accessed directly using its numerical index. e.g. TS would interpret `enemyChainsToCapture[0]` as being a `PointState[]`, but actually it could be `undefined` as well because the original array is not guaranteed to have a member at index 0. So if we needed the length property of the first chain (if present), then we would need to write that as `enemyChainsToCapture[0]?.length` even though TS wouldn't force us to use `?.`.

PR #1141, opened 2024-03-06
API: Add donationForRep() to Formulas

LJNeon:
This was motivated by [a player asking](https://discord.com/channels/415207508303544321/415247422638522395/1215024497661321338) if there was a formulas method to calculate the donation needed to gain an amount of reputation, noting that one existed to calculate the reputation gained from a donation.

With that in mind, these methods simply implement the formulas shown in the faction UI:
```js
ns.formulas.reputation.calculateRepToFavor();
ns.formulas.reputation.repFromDonation();
```
And there's already an inverse method for the first one:
```js
ns.formulas.reputation.calculateFavorToRep();
```
But there was no inverse method for the second one, which is why this PR is adding this:
```js
ns.formulas.reputation.donationForRep();
```

### DIFF
--- a/src/Faction/formulas/donation.ts
+++ b/src/Faction/formulas/donation.ts
@@ -9,6 +9,10 @@ export function repFromDonation(amt: number, person: IPerson): number {
   return (amt / CONSTANTS.DonateMoneyToRepDivisor) * person.mults.faction_rep * currentNodeMults.FactionWorkRepGain;
 }
 
+export function donationForRep(rep: number, person: IPerson): number {
+  return (rep * CONSTANTS.DonateMoneyToRepDivisor) / person.mults.faction_rep / currentNodeMults.FactionWorkRepGain;
+}
+
 export function repNeededToDonate(): number {
   return Math.floor(CONSTANTS.BaseFavorToDonate * currentNodeMults.RepToDonateToFaction);
 }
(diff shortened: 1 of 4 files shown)


src/Faction/formulas/donation.ts
```diff
@@ -9,6 +9,10 @@ export function repFromDonation(amt: number, person: IPerson): number {
   return (amt / CONSTANTS.DonateMoneyToRepDivisor) * person.mults.faction_rep * currentNodeMults.FactionWorkRepGain;
 }
 
+export function donationForRep(rep: number, person: IPerson): number {
+  return (rep * CONSTANTS.DonateMoneyToRepDivisor) / person.mults.faction_rep / currentNodeMults.FactionWorkRepGain;
```
d0sboots, line 13: Optional: ` / (x * y)` is a little clearer than ` / x / y`

d0sboots:
Seems good.

src/Faction/formulas/donation.ts
LJNeon: To be honest that's what I originally had but I realized this way is perfectly symmetrical with the inverse method above it, just with the operators all flipped. But if you still want that change I'll make it.

jjclark1982:
related to #976

LJNeon:
Did some basic testing and the new formulas method appears to work as intended.

PR #1142, opened 2024-03-07
IPVGO: Add support to netscript API for game state, current player, and alternate ways to check/wait on AI turn

ficocelliguy:
Per discussion in Discord https://discord.com/channels/415207508303544321/1171490088702718043/1215308181429948416

* Changed all move, pass, and cheat netscript functions to synchronously throw an error if an invalid move is provided
* Removed all async opponent move logging, unless you specifically call `opponentNextTurn()` and don't disable logging with `opponentNextTurn(false)`
* Changed the `Play` return type to give null instead of -1 for opponent passes or game-over states
* Fixed a bug where the 'new subnet' button in the UI was not applying offline nodes to the subnet
* Tied the UI AI moves into the same code as the netscript AI moves, so that the opponent next move promises still work for manual play
* Updated documentation to include changed return types, opponent move logging, and prevented an error at end-of-game in example script
* Updated documentation to avoid breaking Github's markdown tables with preformatted newlines

* Added API option to get current game state: current player, current score, and last move
* Added API option to await in-progress AI move `opponentNextTurn()`

* Minor formatting and cleanup
* Doc updates


src/Go/Enums.ts
```diff
@@ -12,7 +12,7 @@ export enum GoOpponent {
 export enum GoColor {
   white = "White",
   black = "Black",
-  empty = "Empty",
+  empty = "None",
```
d0sboots: This is an API break now that IPvGO is released in a stable version. Is the name difference really worth enough for that?

src/Go/effects/netscriptGoImplementation.ts
```diff
@@ -161,6 +170,57 @@ export function getControlledEmptyNodes() {
...
+
+/**
+ * Find a move made by the previous player, if present.
+ */
+function getPreviousMove(): [number, number] {
+  if (Go.currentGame.passCount) {
+    return [-1, -1];
+  }
+  return Go.currentGame.board.reduce(
+    (acc, row, index) => {
```
d0sboots: Optional: This feels like it would be a solid step more simple/readable as a simple for-of loop that you could return early from. You're not actually accumulating anything.

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -3967,6 +3969,22 @@ export interface Go {
...
   }>;
 
+  /**
+   *  Returns a promise that resolves with the success or failure state of your last move, and the AI's response, if applicable.
+   *  x:0 y:0 represents the bottom-left corner of the board in the UI.
+   *
+   * @remarks
+   * RAM cost: 0 GB
+   *
+   * @returns a promise that contains if your last was valid and successful, the opponent move's x and y coordinates (or pass) in response, or an indication if the game has ended
```
```suggestion
   * @returns a promise that contains if your last move was valid and successful, the opponent move's x and y coordinates (or pass) in response, or an indication if the game has ended
```

Snarling (changes requested):
For getGameState().previousMove,  if there was no previous move then I think `null` makes more sense than `[-1, -1]`. Not only is it more semantically accurate for "there was no previous move", but it also allows easy conditional checking, e.g.

```js
const state = ns.go.getGameState();
if (!state.previousMove) {
  // handle passes or start-of-game
} else {
  // handle previousMove as valid coordinates on the board
}
```

Likewise currentPlayer can just return `null` when the game is over, instead of changing the GoColor enum values to provide a suitable string within that enum - this also makes it easier to check if the game is still in progress with a simple truthiness check instead of needing to do a direct string comparison.

src/Go/Enums.ts
Snarling: Actually I don't think any of the existing API used GoColor.empty when interacting with player scripts, so this probably wouldn't cause any issues. Also wouldn't cause issues with loading savegames, as boards are stored in a simplified state and don't reference GoColor in the save data except for storing the previous player, which is never GoColor.empty.

However I do still think this change is unnecessary, see my review comment.

src/Go/Go.ts
```diff
@@ -1,5 +1,5 @@
-import type { GoOpponent } from "./Enums";
-import type { BoardState, OpponentStats } from "./Types";
+import type { GoOpponent } from "@enums";
+import { BoardState, OpponentStats, Play } from "./Types";
```
Snarling: Play is also imported as a type here, we can keep this as `import type`

Snarling (commented):
invalidMoveResponse from netscriptGoImplementation is never used
GoPlayType.invalid is never used outside of the above

There is still a weird confusing mishmash of data in the "Play" object, because it's being used as the return to partially indicate info about the player's turn, and partially indicate info about the opponent's turn in response.

For instance, look at the "success" property. This property is true except for the following situations:
* When a new game has been started since the opponent's move was awaited. In this case, it's trying to tell the player that the opponent's move is not a valid thing to respond to, and the player needs to reset any ongoing tracking of that game because that entire game is no longer valid.
* When the player's last turn was a cheat and it failed, leading to the player's turn being skipped. In this case, it just indicates failure of the cheat function.
* If the ai attempts to make an invalid move, which shouldn't happen.

So there's no way to describe what is being returned by the `success` prop because all 3 of these are totally different and these cases also require the player to respond differently. e.g. if the player uses a cheat function and gets success: false, they have no idea whether it's just indicating that their cheat failed, or indicating that a new game has started since waiting for the opponent's turn, and these need totally different responses from the player's script.

It is way easier to make the API make sense if info about the player's move and info about the opponent's move in response are separated, so that functions which make a move only return info about the player's move, and nextOpponentMove returns info about the opponent's move.

ficocelliguy:
I have simplified the Play type/response to only contain info about the opponent's move, since the player will be given instant feedback via an error if their move did not go through. This makes sure all of the data returned is clear and avoids the ambiguity mentioned above.

The only downside to this is that you will need to call getBoardState() to check if a cheat was successful or not (outside of logging), but I think that is probably fine.

Snarling (approved):

PR #1144, opened 2024-03-07
GANG: Correctly display effects of justice tasks on wanted gain rate

### LINKED ISSUE #1015: Gang: Wanted level evolution value off by ca. 0.005/sec
I observed in the gang Management tab that the true wanted level increase is about 0.005/sec less then the displayed evolution

This is mostly noticeable for growth values very close to zero (0.00xxx / sec with xxx < 500), where the evolution of the "Wanted level" displays a positive value (increase) while the wanted level is in fact decreasing. 

Corresponding save (although probably easily reproducible by reaching lower wanted growth values on any save)
[bitburnerSave_1704228885_BN4x1.json](https://github.com/bitburner-official/bitburner-src/files/13814126/bitburnerSave_1704228885_BN4x1.json)

LJNeon:
This needs to be tested.

fixes #1015

### DIFF
--- a/src/Gang/Gang.ts
+++ b/src/Gang/Gang.ts
@@ -152,7 +152,7 @@ export class Gang {
     if (this.wanted !== 1 || wantedLevelGainPerCycle >= 0) {
       const oldWanted = this.wanted;
       const newWanted = oldWanted + wantedLevelGainPerCycle * numCycles;
-      // Ensure that even members with poor stats can effectively lower wanted level
+      // Allows recovery when wanted / respect ratio is too high
       this.wanted = newWanted * (1 - justice * 0.001);
       this.wantedGainRate -= newWanted - this.wanted;
       // Prevent overflow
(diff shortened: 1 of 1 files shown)


src/Gang/Gang.ts
```diff
@@ -149,15 +149,14 @@ export class Gang {
...
+      const newWanted = oldWanted + wantedLevelGainPerCycle * numCycles;
+      // Ensure that even members with poor stats can effectively lower wanted level
+      this.wanted = newWanted * (1 - justice * 0.001);
+      this.wantedGainRate -= newWanted - this.wanted;
       // Prevent overflow
-      if (wantedLevelGainPerCycle <= 0 && newWanted > oldWanted) newWanted = 1;
-
-      this.wanted = newWanted;
-      if (this.wanted < 1) this.wanted = 1;
+      if (this.wanted < 1 || (wantedLevelGainPerCycle <= 0 && this.wanted > oldWanted)) this.wanted = 1;
```
Snarling, line 159: What's the purpose of this extra condition? When would we be gaining negative wanted per cycle, but this.wanted would end up greater than the old wanted level?

LJNeon: I don't know it was already there so I just kept it untouched. It had a "Prevent overflow" comment.

Snarling: Oh I see that now, you just combined the two existing conditions for setting wanted to 1.

I'm not sure which edge case that is accounting for but I guess just leave it in. This looks fine to me.

PR #1147, opened 2024-03-08
BUGFIX: Disallow all whitespace in filenames

LJNeon:
Currently only the space character " " invalidates a filename, all other whitespace characters are allowed. This PR expands the invalid character list to include all whitespace and corrects all filenames with whitespace in existing saves.

This has been tested by @Fireball5939.

### DIFF
--- a/src/Paths/Directory.ts
+++ b/src/Paths/Directory.ts
@@ -24,11 +24,11 @@ export const root = "" as Directory;
  * |: Invalid because it might have a use in the terminal in the future.
  * #: Invalid because it might have a use in the terminal in the future.
  * (quote marks): Invalid to avoid conflict with quote marks used in the terminal.
- * (space): Invalid to avoid confusion with terminal command separator */
-const invalidCharacters = ["/", "*", "?", "[", "]", "!", "\\", "~", "|", "#", '"', "'", " "];
+ * (whitespace): Invalid to avoid confusion with terminal command separator */
+const invalidCharacters = ["/", "*", "?", "[", "]", "!", "\\", "~", "|", "#", '"', "'"];
 
 /** A valid character is any character that is not one of the invalid characters */
-export const oneValidCharacter = `[^${escapeRegExp(invalidCharacters.join(""))}]`;
+export const oneValidCharacter = `[^${escapeRegExp(invalidCharacters.join(""))}\\s]`;
 
 /** Regex string for matching the directory part of a valid filepath */
 export const directoryRegexString = `^(?<directory>(?:${oneValidCharacter}+\\/)*)`;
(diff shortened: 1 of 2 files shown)


src/Paths/Directory.ts
```diff
@@ -24,8 +24,8 @@ export const root = "" as Directory;
  * |: Invalid because it might have a use in the terminal in the future.
  * #: Invalid because it might have a use in the terminal in the future.
  * (quote marks): Invalid to avoid conflict with quote marks used in the terminal.
- * (space): Invalid to avoid confusion with terminal command separator */
-const invalidCharacters = ["/", "*", "?", "[", "]", "!", "\\", "~", "|", "#", '"', "'", " "];
+ * (whitespace): Invalid to avoid confusion with terminal command separator */
+const invalidCharacters = ["/", "*", "?", "[", "]", "!", "\\", "~", "|", "#", '"', "'", "\\s"];
```
d0sboots: I don't think this will work as-is, because of how `oneValidCharacter` is constructed via escapeRegExp in the next line.

d0sboots (commented):
Since this involves a savefile migration, it needs extra careful testing. Also, it should bump the savefile version.

src/Paths/Directory.ts
d0sboots: Oh yeah, it's failing tests.

LJNeon: Would I just add it in the line below? I'm not sure if it's possible to get that working within escapeRegExp?

LJNeon:
How do I bump the version? Just increment 38 to 39? Anything else I need to do?

d0sboots:
Ye, just increment

src/SaveObject.ts
```diff
@@ -719,6 +721,39 @@ Error: ${e}`);
...
+    }
+    if (found) Terminal.error("Filenames with whitespace found and corrected, see console for details.");
+  }
+}
+
+function removeWhitespace(hostname: ServerName, file: ContentFile, files: ContentFileMap): void {
+  let filename = file.filename.replace(/\s+/g, "-") as ContentFilePath;
+  // avoid filename conflicts
+  if (files.has(filename)) {
+    const [path, ext] = filename.split(".");
```
Snarling: Files can have . in them before the extension, this doesn't reliably provide you with the extension

LJNeon: Should I just go by `lastIndexOf` instead then?

Snarling: Yeah that should work, the last . and anything after it is always the extension

LJNeon:
@d0sboots this should be good to merge unless you're not satisfied with [fireball's testing](https://discord.com/channels/415207508303544321/459097632896188436/1338242781524525176). If so please let me know what additional testing you'd like done.

d0sboots:
We have to remember to actual *do* the savegame bump, now.

PR #1148, opened 2024-03-09
SAVEGAME: Reduce savegame size / complexity of save/load

Snarling:
Loading for Companies and Factions already initializes all companies and factions, and then just checks the savedata for valid data points and loads those data points onto the already-initialized companies and factions.

However, since both were previously using their classes + the JSONReviver for save/load, they were saving extra data (the IReviverValue), and they were also initilizing class members when loading, even though these class members would never be used as members of the class, only for loading properties onto the existing class members.

Due to no longer using the IReviverValue, the size and complexity of the savegame data is reduced, which should slightly reduce disk usage and save/load times.


d0sboots:
Why are you changing the savegame JSON format for these?

I understand and agree with dropping everything that has default values, that saves space. And I also agree with changing *how* they're loaded, so we don't have issues with stale data if/when we change a company or faction.

But removing the IReviverValue wrapper saves minimal space, and creates the need for a migration step, which is always more risky (and has increased chance of rotting over time).

Snarling:
The IReviverValue overhead is not negligible when we're talking about numerous 1-layer objects with two simple properties each. The overhead is about half of each Company/Faction when it's included.

For most of game loading, we have to load objects using the constructors for the class that they're going to be loaded as, because the loaded objects directly become the game objects.

That isn't true for Faction or Company, where all actual game objects are created statically upfront, and then the savedata is just used to copy properties to the actual game objects. This allows actually typechecking the savedata, which the reviver method doesn't handle. The load method for Companies was changed in #650 and for Factions was changed in #644. There may have been some further changes since, but those were when the overall loading method was changed from "load in savedata to initialize the game objects" to "initialize game objects statically, use loaded data to copy valid properties over to game objects". This PR can really be seen as a followup on those two to make the save format actually make sense for that loading method.

d0sboots:
> This PR can really be seen as a followup on those two to make the save format actually make sense for that loading method.

I guess the difference in my perspective is, I don't see either of these formats as intrinsically making more or less sense for what you're doing. Rather, I see one as slightly larger but also more backward-compatible. The fact that you can simply have Reviver return the wrapped value as the migration method means it isn't *much* less backward compatible, but conversely it also shows how the wrapped format isn't any less suitable for what you're doing, either.

But since the migration is so simple, I don't have significant concerns.

PR #1151, opened 2024-03-09
GANG: Gang overhaul

Kelenius:
Test version is up at: https://kelenius.github.io/bitburner-src/

Also it's obviously going to be a completely breaking change, but limited to gang.

I've had an earlier idea about making a separate territory clash for hackers (cyberspace) but it quickly became clear that it's just overcomplicated, confusing, and causes a lot of duplicate code, so I scrapped that idea.

Goals:

* Making hacking gangs more relevant and competitive with combat gangs.
* Make creating gang with a high-end faction more rewarding.
* Don't make the gangs significantly weaker or stronger overall.

Features, some finished, some WIP:

* Gangs are no longer split into combat and hacking. Instead, gang members can be either enforcers or hackers. Different gang factions have different maximum amounts of each type.
* * Combat-focused gangs tend to allow more enforcers, hacking-focused gangs tend to allow more hackers, and "lategame" gangs allow high amount of both.
* * Gangs are still limited to 12 maximum gang members.
* Gang faction rebalance:
* * Slum Snakes, Tetrads, and Speakers are combat-focused gangs.
* * NiteSec and Black Hand are hacking-focused gangs.
* * Syndicate and Dark Army are "lategame" gangs.
* * Gangs that can be joined earlier have more restrictive enfocer/hacker maximums.
* Warfare rebalance
* * Previously, gang strength was: 1, 2, 2, 2, 2, 5, 5. Joining one of two strongest gangs made territory warfare more difficult because one strong gang would kill the rest quickly.
* * Now, gang strength is: 1, 2, 3, 3, 4, 4, 4. Joining one of the strongest gangs still leaves two strong gangs that can fight each other.
* The purpose of the previous two changes is to make creating a gang with a faction that's more difficult to join more rewarding. Currently Slum Snakes are easiest to join and also the best.
* New warfare tasks that let gang members build up strength at a slower rate (50%), but without the risk of death.
* * Useful for hackers who have low defense.
* Task rebalance
* * Hacking tasks have better territory scaling, particularly for money
* * Enforcer tasks bring less money overall
* * Hacking and Charisma training buffed (were previously much weaker than combat for no discernible reason)
* Equipment rebalance
* * Previous maximums: (Hack, Str/Def/Dex/Agi, Cha) (2.27, 29.21/15.52/2.22/7.29, 1.46) - too much emphasis on strength among combat stats, and too little hacking for hackers to be viable
* * Current maximums: 8.68, 20.04/14.72/3.40/7.28, 1.83
* * Nerfs to weapon strength
* * Vehicles now give dex
* * Rootkits doubled in effectiveness
* * New category: "Equipment". Contains various tools with emphasis on stats that are currently low.
* * * The idea of the above changes is that enforcers should be good at gaining respect, and hackers should be good at gaining money.
* * Real-name names removed and replaced with fictional or generic
* * * I really really really do not like using real-life names, especially if they're trademarked, in the game. I also think that fictional names or references to sci-fi or cyberpunk media would fit the game much better. Additionally, in some cases it might not be clear what the name means.
* * All equipment is now visible
* New function: ns.gang.retireMember()
* * Has the same effect as being killed in a clash, exists so you can replace them with another type. (Also, why not)


MaWo2:
As you are asking for feedback: I think this might be a good idea, although a really breaking one. The mechanics should be properly explained somewhere. Currently, one only finds tutorials for combat gangs, but nobody explains, why hacking gangs are bad. At least I never found it. I am using a hacking gang and am quite happy with it. For me and my purpose, combat gangs were too difficult and counter-intuitive in a hacking-focussed game.

With regard to the new function `ns.gang.executeMember()`: Please consider renaming it to "retireMember"!

Although this is all a fictional story, is it really necessary to kill someone you do not need any longer? If the member cannot be trained to a new task (resetMember???), I think it's is sufficient to just give it the golden clock and send it home...

EDIT: For future ideas (don't know), it might even be possible to call a retired member back into work, if the gang has demand for it. The member could have been degrading in skills over time, but would still be better trained than a fresh one.

Kelenius:
> The mechanics should be properly explained somewhere. Currently, one only finds tutorials for combat gangs, but nobody explains, why hacking gangs are bad. At least I never found it. I am using a hacking gang and am quite happy with it. For me and my purpose, combat gangs were too difficult and counter-intuitive in a hacking-focussed game.

I think the basics are fairly well-explained, and in the rework there are no hacking and combat gangs anymore. I added some more explanations to the territory screen.

> Please consider renaming it to "retireMember"!

I mean, it's a gang... it's a bit weird to be averse to killing when running one, I think. And you don't have to do this if you don't want to, the only practical purpose it has is replacing them with another type.

Kelenius:
Also, the PR is now ready for testing. I think I'm happy with the big changes and now it's down to fixing smaller things and rebalance.

User670:
> > Please consider renaming it to "retireMember"!
> 
> I mean, it's a gang... it's a bit weird to be averse to killing when running one, I think. And you don't have to do this if you don't want to, the only practical purpose it has is replacing them with another type.

Still prefer a more neutral name. I'm thinking `removeMember`.

Killing a member just because you need a different type of member still feel mildly weird and uncomfortable for me. Unless you are justifying it by saying "if we don't kill him, he'll join the other gang and work against us"

Kelenius:
Well, if people feel strongly about it, I can change it to something more ambiguous, I just find it odd that this is an issue given what the gang and the player are already doing. I called it execute because it seemed fitting, in a "you have outlived your usefulness" kind of way.

Tyasuh:
Just a couple thoughts so far : 
-Manual Recruit = point out the ability to toggle recruitment type? I missed it first go round
-recruitMember("Name",false) //doc reads like true/false to be an enforcer or hacker.   
-getTaskNames() = getEnforcerTasks(), getHackTasks() ? Or optional to split the list between isHacking or not?

Zelow79:
I hate to be that guy, so I'll start by saying this is just my 2 cents and how I feel about it and it's nothing personal, but I think combat gangs as they are now are perfect and a dramatic change to them could effect my overall enjoyment of the game. I know territory warfare seems to be a taboo thing for many, but it was honestly fun to implement solutions with it as it is. I'd love to see an improvement to hacking gangs, but not at the expense of what combat gangs are right now. When it comes to territory warfare timing and the whole every 20 seconds flashmob stuff, I think it's great it's possible, but I don't think everyone should have to feel like they can make their script get the same results. That's like making corps easier in my eyes. Aside from hacking gangs being better, some kind of rework to karma could be nice too. Always felt lame that gangs took so long just to start.

d0sboots:
My two cents: I have no opinion here. I actually haven't done gangs yet, so I don't feel qualified to weigh in, except on technical/code-style matters. My gut feeling is to be conservative, because I haven't seen lots of people complaining about gangs, and a fair number of people like them. Ultimately Snarling has the final call.

Alpheus:
**My 2 cents**
👍 Love that there is no difference in gang types anymore
👍 Warfare rebalance is welcome, though I don't think it's as important as the other changes
👍 Equalizing training helps a lot (also, see suggestion below for more on this avenue)

Now the bad stuff:
👎 New function: ns.gang.executeMember() I see no place for this. I don't think any game-optimal loop that rewards killing off your gang members makes the game fun
👎 Still no practical approach to joining the late-game gangs on purpose.

**Suggestions (no need to follow this, just my ideas)**

1. Set "type of member" based on their dominant stat. Combat stat members get access to combat tasks, hacking members get access to hacking tasks. Under a certain threshold, tasks are limited to the basics (training, ...)

2. Have a scaling factor for stat XP depending on the current layout of enforcer vs. hacker. Scaling factor threshold depends on the faction (ie. Slum Snakes would penalize XP gain of combat stats from having 8 enforcers less than Black Hand).

3. Rather than executeMember give an option to reset stats or transfer the member's ascension bonus to someone else (like re-sleeving). 

4. The way the ascension and training loop works right now means it is punishing to delay creating a gang if you're doing it for performance reasons (money, rep) within the gameplay of the current node. If anything, all gangs should be available merely from _rumored_ state, but have the harder ones require outside funding to prep up faster.

5. The player's hacking or combat skills _could_ contribute in some way to early game xp gain on each member (or respect for that matter). This way a player with NFGs who has foregone starting gangs early could catch up with endgame game. Otherwise Slum Snakes will always have the advantage of having the most effective time horizon

Snarling:
Sorry, just now taking a look at this. I haven't really dived into the code yet, this is just based on what I've seen in the description here:

I was going to make a list of stuff I like and don't like as I read through the major changes, but most of the changes sound pretty agreeable to me. Here was my list of things I like:
* Replacing hacking/combat gang categories with similar gangmember categories - allows more freedom
* Buffing hacking side of gangs
* Reworking the equipment to not be so lopsided towards strength
* Making it more attractive to join lategame gang factions
* Equipment names not being real-world branded things

Kelenius:
This is now ready for a full review and I think is in an acceptable state except for one question: what should we do regarding save conversion? The big two sticking points are that gang members have types now, and that equipment has been changed. I think the best approach would be to just delete a gang if they player is in one, because converting accurately would be pretty hard - but that's pretty punishing for the player.

d0sboots:
Save conversion is always tricky, especially for big overhaul-type changes. I would say "it should convert as faithfully as possible," but before you sink a lot of time into that, I think we should get consensus on whether or not we are going forward with this.

xxxsinx:
xsinx here, I don't see anything about the TW tick in the comments, I did not look at the code changes. I was under the impression whatever next PR was going to tackle this, some have asked for a nerf/removal, I'm more interested in a better mechanic that doesn't involve tracking time. I'm going to guess this PR doesn't touch this?

I don't know about the power numbers change for gangs. We're still making the entry, first available gang the best? Despite it requiring the least effort to join? I'm wondering if literally inverting the power scale might be best? The further up the requirement ladder to get in, you end up getting a more powerful gang (considering a lower number currently yields a greater advantage). Or perhaps power should be a bit like stocks, and fluctuate/flip over time, making every run feel different (as far as the enemy gangs go at the very least, since currently that number only directly affects other gangs).

I'd like to see a change to the 54k level requirement too. Would it not make more sense to be able to create a gang as soon as you get access to it's faction? Perhaps there could be a readjustment of requirements that goes with that. This would make gangs more powerful without any adjustment, but it could be that the starting territory % is lower to counter.

Just some ideas, sorry if it's a bit all over the place.

Snarling:
For save conversion, it will be a bit tricky to do it in a real 1:1 kind of way, I don't think that kind of conversion is really worth it but I'm also not opposed if that's how you want to implement it.

I think as long as we maintain most of the power of the gang, it should be fine. It doesn't seem like any of this should be an issue to preserve: The number of members, their stats, their ascension mults, the gang's power, territory, respect, and wanted level. Existing members should probably be given the type corresponding to their legacy gang. If you were looking for advice on the implementation of the conversion code, I usually prefer to initialize an entirely new object and just load in the values you're trying to preserve from the saved object.

Things I probably wouldn't preserve: member equipment, member tasks, whether warfare is enabled. Along with a dialog telling players about the changes, this prompts the player to go through their gang and set things up how they want with the new system. Since it's a one-time conversion, it's also fine to be generous in the conversion, e.g. load the gang with 1.5x the power and reputation, to make up for the hassle / time the player has to spend re-configuring their gang.

Faenre:
Would the Gang Overhaul PR be a good opportunity to adjust the karma requirements as well? Gangs outside of BN2 seem to take an incredibly long time just to kick off. Grinding the -54k essentially removes work mechanics for ~two days for anyone who wants to make a Gang.

Zelow79:
> Would the Gang Overhaul PR be a good opportunity to adjust the karma requirements as well? Gangs outside of BN2 seem to take an incredibly long time just to kick off. Grinding the -54k essentially removes work mechanics for ~two days for anyone who wants to make a Gang.

Iirc, with sleeves the karma grind is ~18hrs, however I agree. Not that I agree with most of this PR, karma should be addressed. If gang needs to be rebalanced, then why not make karma part of that rebalancing. I think it could be cool if karma was more optional, maybe the less karma you have the better some gang functions scale, rather than some gate keeping you from the mechanic for so long.

PR #1154, opened 2024-03-11
BLADEBURNER: Typesafety / refactoring

Snarling:
### Non-bladeburner changes
* clampNumber defaults minimum value is -Number.MAX_VALUE instead of Number.MIN_VALUE. MIN_VALUE is almost never something we would want to set as a limit.

### BREAKING CHANGES
* Loading a game from before this PR that was in the middle of Bladeburner will reset your BlackOps progress.
* A savegame from this PR that has an active Bladeburner instance will not load correctly in the main branch, due to lots of static info being removed from the Bladeburner savedata. This has no real impact since the Compression PR was also just merged and causes a similar incompatibility.

### FUTURE TODOS:
* Rework completeAction, completeContract, and completeOperation, implement as Action.complete. Would have dealt with this during this PR, but it was already taking a long time and I wanted to get this out. I'll work on this in the coming weeks.
* 3.0: Get rid of fuzzy matching for player input on enum values.

### Commit details (also just see the commit comments)

458421a: No real substantive changes yet, just jotting down enums that should exist (not all populated yet)
ed8d502: Populated all the created enums with the correct values. They are still not actually used anywhere.
1ef61ef: Made BlackOperations object keyed by enum, and made necessary fixes elsewhere to account for key no longer being a magic string (needs typechecking before access). Couple of other minor changes. See commit details for more info.
b916a51: Remove extra BlackOperations file in data directory, moved descriptions to main BlackOperations.ts file and changed them from React to strings.
23b4e24: Fixed an import
e487134: Changed how black op completion is tracked, see commit details.
94478dd: Fix a test
75833d9: Change format of ActionIdentifier to be typesafe, and other changes to support that. See commit details for more info.
32f8c34: Some changes to GeneralActions and SleeveBladeburnerWork
28178c6: Make Action into an abstract class (GeneralAction is now its own class, so no objects actually use the Action class)
f5085cd: Better directory organization, slim down data saved by Contract/Operation, create a LevelableAction abstract class for Actions that have levels, and move level-related stuff to that class.
18afc2d: Some typesafety stuff for Skills
3838a42: Moving some netscript-specific code out of the Bladeburner class
09ef617: All actions now have a member function for determining if that action is available, use that to streamline some more special-case code
...: Stopped individually writing up every commit. At one point I had Contracts and Operations acting as static objects just like GeneralActions and BlackOperations, but I reverted that due to some issues around save/load.
...


Snarling:
~~There's an unfortunate issue with the design I went with on Contracts and Operations, where these classes hold all of the information about the object, and the objects are statically initialized with the dynamic content loaded in when loading a bladeburner object.~~

~~The Bladeburner object can be loaded in instances where we're not actually trying to load the game yet, e.g. when doing the import comparison thing. This would trigger the code that loads the contracts+operations savedata onto the static objects even if the player doesn't actually go through with the import.~~

~~Might have to switch over to something more like how Factions works - where the actual object that interacts with savedata is minimal, and there's a separate ContractsData or something that has all of the static data. I don't really like that idea though as it would require a lot of changes to Actions in general, makes it harder to interact with the data, and just generally is kind of an ugly solution. I will think a little more on a better way to handle this.~~
Addressed in 453ddaf, it was a lot simpler than I thought it would be,

d0sboots:
> Addressed in https://github.com/bitburner-official/bitburner-src/commit/453ddaf3ceaa88fd5fd6d6bac972b33ec658b688, it was a lot simpler than I thought it would be

Nice, because that's the design I'd prefer as well :D

Snarling:
Test failure is just due to removing this.reset() from Bladeburner constructor, so contracts/operations are only randomized in count once instead of twice, and end up with different counts than the snapshots. I'll include a test fix in the next commit.

PR #1155, opened 2024-03-11
UI: Fix go colors for light theme

Snarling:
~White should be a black color in a light theme~
Actually it appears that the white and black theme colors are not currently used for anything that is actually visible, outside of go. So now this does the opposite change: white/black are now the literal colors in the light theme (changed black to a black color).


ficocelliguy (approved):

User670:
White using black and black using white will make IPvGO colors be inversed - when the game says "black" pieces it alpears white on-screen.

User670:
Where are white and black used, other than Go? Should "white" and "black" be literal?

Snarling:
We discussed this and looked into it some more on discord, to me it looks like the game doesn't actually use the "black" or "white" colors anywhere that actually can be seen, other than go.

So I think for now I will change this to be the opposite of the original change (change black to be a black color in light theme, keep white as a white color)

PR #1156, opened 2024-03-11
IPVGO: Ensure that IPvGO styling is not broken by unusual themes

ficocelliguy:
* Ensure that there is always a contrast between light pieces and dark pieces (because some themes use the same color for the 'white' and 'black' theme entries) via luminance calculation of hex color
* Use a softer color if there is not contrast, to ensure it isn't too bright for people with eye sensitivity
* Ensure that light themes have a medium-luminance background, so you can see all pieces easily
* Ensure changing to traditional go style re-renders the board


ficocelliguy:
Closing this, as it is handled by #1155

User670:
> Ensure changing to traditional go style re-renders the board

This bit should still be added despite #1155, unless I misunderstood this sentence.

PR #1159, opened 2024-03-12
MISC: Compress savefiles

### LINKED ISSUE #800: Savegame compression
Currently, our savegames are raw base64-encoded json. This is simple, but can lead to large savegames, especially in some semi-common circumstances.

The purpose of this FR is to scope out a design for adding compression. It might be that this feature is deemed too complicated for the benefit, which is fine - this then will serve as a central place to explain why.

A major trickiness is that JS does not have built-in support for compression *that is exposed as an API*, even though support for deflate is mandatory in various web standards. There are packages that implement deflate in JS, but their error-handling (and performance) is often less than great. Still, for handling working files of reasonable size that is an option, and we already have a dependency on jszip.

The other route is to compile zlib into WASM. I've done this, it works, is high-quality, fast, and (when done properly) can have a very small library size, but means more infrastructure work.

There's also work to be done with versioning. Probably the existing JSON structure would need/want to be kept for compatibility and the quick stats view during loading, but only a stub of info would be kept there, with the bulk of the data being in a compressed JSON payload after.

yichizhng:
Test server available at: https://yichizhng.github.io/bitburner-src/

Backward compatible with old savefiles and with old browsers which don't have CompressionStream, in which case it will fall back to the old save format. (Maybe add a toggle or button to generate the old savefile for someone who switches between browsers which do and don't have it?)

A further improvement would be to store the Uint8Array directly in indexedDB and only do the base64 on export/import, but that would require some refactoring.

closes #800


src/SaveObject.ts
```diff
@@ -719,10 +730,20 @@ Error: ${e}`);
   }
 }
 
-function loadGame(saveString: string): boolean {
+async function loadGame(saveString: string): Promise<boolean> {
   createScamUpdateText();
   if (!saveString) return false;
-  saveString = decodeURIComponent(escape(atob(saveString)));
+  if (saveString.startsWith('{"CompressedSave": ')) {
```
d0sboots, line 736: You're not actually using the fact that the savegame is JSON here - this is a simple magic-string check. Indeed, the save could be valid JSON containing a CompressedSave block but fail this check.

This is why I think it would be better to go with a gzipped version of the existing saveString. You could perform the same magic-string check (but with a different byte sequence), and then most of the logic would end up the same. But it would be much easier for users to look at. (Any outside tool will understand that a `.json.gz` file is just a gzipped `.json` file.)

src/SaveObject.ts
```diff
@@ -864,6 +885,63 @@ function download(filename: string, content: string): void {
...
+async function concatUint8Arrays(uint8arrays: Uint8Array[]): Promise<Uint8Array> {
+  const blob = new Blob(uint8arrays);
+  const buffer = await blob.arrayBuffer();
+  return new Uint8Array(buffer);
+}
+
+// This is not the most efficient implementation;
+// See https://developer.mozilla.org/en-US/docs/Glossary/Base64
+function base64ToBytes(base64: string): Uint8Array {
+  const binString = atob(base64);
```
d0sboots, line 936: You can use TextEncoder here and it's both faster and simpler.

src/SaveObject.ts
```diff
@@ -864,6 +885,63 @@ function download(filename: string, content: string): void {
...
+}
+
+async function concatUint8Arrays(uint8arrays: Uint8Array[]): Promise<Uint8Array> {
+  const blob = new Blob(uint8arrays);
+  const buffer = await blob.arrayBuffer();
+  return new Uint8Array(buffer);
+}
+
+// This is not the most efficient implementation;
+// See https://developer.mozilla.org/en-US/docs/Glossary/Base64
```
d0sboots, line 934: I would actually suggest using the gross but efficient implementations. If not that, then the better versions of these that don't use closures. (The difference is significant.)

package-lock.json
```diff
@@ -46,7 +46,8 @@
         "react-resizable": "^3.0.5",
         "react-syntax-highlighter": "^15.5.0",
         "remark-gfm": "^3.0.1",
-        "sprintf-js": "^1.1.3"
+        "sprintf-js": "^1.1.3",
+        "stream-browserify": "^3.0.0"
```
d0sboots: What is this here for?

src/SaveObject.ts
yichizhng: I mean, I could just not do the magic string check and try JSON.parse directly -  then it would be equivalent to the magic bytes approach, but more in the spirit of high level languages.

To summarize some discussion on the discord, there are several options, ordered in ascending savefile size:
- Write the compressed bytes to file with some magic bytes in front; this is optimal for performance and size
- Base64 the compressed bytes and write it to a JSON file (what this PR does)
- Compress the current savefile; this will result in a worse compression ratio, but make it easy to use external tools to edit the save

Note that the export and save formats can differ - we could write the Uint8Array to indexedDB as a blob and then only do the transformations on export/import, and I think that happens rarely enough that the performance difference is not important.

package-lock.json
yichizhng: oops

src/SaveObject.ts
d0sboots: What I was suggesting was the first option (mostly): gzip files are compressed bytes with a header (and also a minimal footer). An advantage is that the format is natively supported by CompressionStream; it's just a change in the constructor from "deflate-raw" to "gzip". I misspoke slightly when I said "gzipped version of the existing saveString," because what I actually meant was "gzipped version of the existing json" (i.e. before it goes through base64 encoding). IMO this also makes it easiest to use external tools; Windows itself can natively open .gz files.

Also agreed that we can potentially do other things for internal saves. The most obvious one would be skipping compression if it were a performance hit. Avoiding conversion to a string is another.

Edit: Oh, and I think you want to do the check for the magic string first. It matters in terms of error reporting: If you can't manage to load anything valid, generally you are reporting errors on the 2nd method you tried, because it was the catch-all one. In our case, that should be JSON.parse.

d0sboots:
Sorry for the silence on this one. It looks like we're going to go with #1162 instead, mostly because that one adds a lot of useful comments (otherwise, they're largely quite similar).

PR #1161, opened 2024-03-12
MISC: Stanek UI description corrections

### LINKED ISSUE #689: Misleading/incorrect documentation/UI for Stanek's Gift
Just started BN 13 (steam version 2.4.0), and noticed an issue with how charges work vs. how they are explained/displayed.

The help popup says "Note that it is no more effective to charge a fragment many times with few threads than to charge few times with many threads, so there is no need to distribute charging jobs across multiple scripts.", and the "Charge: 10.000" display agrees, showing the total number of threads that have charged it. However, the resulting effect from that fragment depends on how exactly it was charged: charging fewer times but with more threads will end up having a greater effect than charging many times with few threads.

For example, charging a "hacking skill" fragment 1 time with 10 threads will make it "Charge: 10.000" and "Effect: +7.50% hacking skill", while charging 10 times with 1 thread each will also make it "Charge: 10.000" but have "Effect: +2.44% hacking skill"

Separately, some of the fragments don't say all the stats they effect, e.g. the "hacking skill" fragments also increase hacking exp (same for all the physical stats as well), and the "crime money" fragment also increases crime success chance.

gmcew:
Closes #689

Minor description changes to Stanek charging effectiveness UI entry and stat fragment properties.

It's a tiny contribution, but my only coding education has been through this game. If I can update some documentation it's a small thank you at least!

### DIFF
--- a/src/CotMG/ui/StaneksGiftRoot.tsx
+++ b/src/CotMG/ui/StaneksGiftRoot.tsx
@@ -170,7 +170,7 @@ export function StaneksGiftRoot({ staneksGift }: IProps): React.ReactElement {
                   Stat Fragments are charged using the stanek.chargeFragment(rootX, rootY) NetScript API function. The
                   charging process ordinarily takes 1000ms to complete, but only takes 200ms during bonus time. When the
                   function finishes executing, the fragment's charge levels will be raised by an amount corresponding to
-                  the number of threads that were used. Note that it is significantly more effective to charge a fragment
+                  the number of threads that were used. Note that it is generally more effective to charge a fragment
                   a few times with many threads than to charge it many times with few threads, so there is a benefit to
                   distributing charging jobs across multiple scripts. As a Stat Fragment's charge level is increased, its
                   bonuses will increase, but there will be diminishing returns. As such, it is generally most efficient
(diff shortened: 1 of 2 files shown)


src/CotMG/ui/StaneksGiftRoot.tsx
```diff
@@ -170,11 +170,11 @@ export function StaneksGiftRoot({ staneksGift }: IProps): React.ReactElement {
                   Stat Fragments are charged using the stanek.chargeFragment(rootX, rootY) NetScript API function. The
                   charging process ordinarily takes 1000ms to complete, but only takes 200ms during bonus time. When the
                   function finishes executing, the fragment's charge levels will be raised by an amount corresponding to
-                  the number of threads that were used. Note that it is no more effective to charge a fragment many
-                  times with few threads than to charge few times with many threads, so there is no need to distribute
-                  charging jobs across multiple scripts. As a Stat Fragment's charge level is increased, its bonuses
-                  will increase, but there will be diminishing returns. As such, it is generally most efficient to
-                  charge all of the placed fragments equally. The charge level of a fragment will not decrease over
+                  the number of threads that were used. Note that it is generally more effective to charge a fragment
+                  a few times with many threads than to charge it many times with few threads, so there is a benefit to
```
d0sboots: > so there is a benefit to distributing charging jobs across multiple scripts

This conclusion is contradicted by the premises? If charging is best done only a few times with many threads, you want only a few scripts doing it.

gmcew: You're right, I just inverted the sentence without checking it was actually correct. I've updated it to reflect that conclusion.

d0sboots:
You need to run `npm run format`. It probably wants to adjust how long some of the lines are.

PR #1162, opened 2024-03-13
MISC: Support compression of save data

### LINKED ISSUE #800: Savegame compression

catloversg:
This PR adds support for compression of save data.
Closes #800.

This is a recap of changes:
- Use Compression Streams API to compress save data.
- Compression is enabled by default if the browser supports Compression Streams API. If the browser does not support that API, we fall back to the old base64 format. We still use the old extension (.json) for the old base64 format.
- New binary format is gzip. New extension is .json.gz.
- Players can still import save file that uses the old base64 format.
- In Electron app, `Compress Disk Saves (.gz)` option is removed. Save data is compressed by default now.
- Proper support for other tools (DevMenu, pretty-save.js). Full support for DevMenu will be added in a follow-up PR. Check the comments in `src\DevMenu\ui\SaveFileDev.tsx` for details.
- Custom recovery mode UI in 2 cases:
  - Browser does not support Compression Streams API.
  - Cannot decompress save data in IndexDB.
- If there is a `SaveDataError` when decoding save data, the toast shows a relevant error message instead of the generic `Save game is invalid` message.

----
The details below are the original description of this PR. Some of them are outdated. I only keep them for historical context.

----

### PR's goals
- Use Compression Streams API instead of jszip or other libraries.
- Compression of save data is optional. The player can disable it if they want. This is a setting in Options → System.
- Remove usage of base64 in the new binary format.
- Do not convert binary data to string and back. The type of save data is `SaveData`, it's either `string` (old base64 format) or `Uint8Array` (new binary format).
- Proper support for interacting with electron-related code. Electron-related code assumes that save data is in the base64 format.
- Proper support for other tools (`DevMenu`, `pretty-save.js`). Full support for `DevMenu` will be added in a follow-up PR. Check the comments in `src\DevMenu\ui\SaveFileDev.tsx` for details.

### Binary format
Format: `[[Magic byte(s)] BinaryFormatVersionByte [Gzip bytes]]`
- Magic byte(s): `\x00`.
- `BinaryFormatVersionByte`: start at 0. It's only increased when we change the binary format.

#### Path of `saveDataBinaryFormat.js`
Code in this script is used by electron-related code and game logic code. Originally, I put it in the top-level folder, but the electron build failed to include it. I didn't troubleshoot and just put it in the electron folder. I can try to move it to other places if the maintainers want.

#### Use gzip directly instead of the new binary format
I create new binary format for these reasons:
- Custom support for whatever compression algorithm/format.
- Proper versioning in case we want to change the binary format, e.g., change compression algorithm/format.
- Embed other data in the save file, not just gzip of json save string. In the prototype of this PR, I have plan to embed `ImportPlayerData` in the save file, so we can show the `Compare Save` UI without decompressing the entire save file. However, I remove that feature later because it complicates the logic without any practical benefit. Decompression of gzip is really fast.

This comes with a downside: the player cannot edit the save file easily by decompressing it with gzip tools.

@d0sboots I know that you want to just use gzip format, and I partially agree with you. If you have strong opinion on it, I don't mind removing the new binary format and just stick to gzip.


d0sboots:
I haven't looked at the code yet, just commenting on the format.

I have been (and continue to be) a strong proponent of "just gzip the json." However, I understand and sympathize with the desire to have extensibility/versioning. If we *do* want some level of that, I think an architecture like yichizhng's would be better, with a minimal JSON file containing a base64'd gzip'd payload. This is mostly because as a game that is designed to be hacked on by our players, I think it's important that the save format not be super-opaque.

electron/storage.js
```diff
@@ -199,19 +217,31 @@ async function pushGameSaveToSteamCloud(base64save, currentPlayerId) {
...
 async function getSteamCloudSaveString() {
   if (!isCloudEnabled()) return Promise.reject("Steam Cloud is not Enabled");
   log.debug(`Fetching Save in Steam Cloud`);
   const cloudString = await getCloudFile();
-  const gzippedBase64Buffer = Buffer.from(cloudString, "base64");
-  const uncompressedBuffer = await gunzip(gzippedBase64Buffer);
+  // Decode cloudString to get compressedBuffer back.
+  const compressedBuffer = Buffer.from(cloudString, "base64");
+  // uncompressedBuffer is the binary form of the json save string.
+  const uncompressedBuffer = await gunzip(compressedBuffer);
```
d0sboots: Wait, so cloud saves are *already* just the gzipped save string? XD

catloversg: Yes, the save file in Steam Cloud is the gzipped json save string.

catloversg:
In the latest commit, I decide to use gzip format as the binary format.

I don't like the approach of using a minimal json file as the new format:
- When we have 2 formats, we need a good way to distinguish between them. Checking for `{"CompressedData":` works, but it's fragile. It's much easier/safer with the binary format.
- Still need to convert binary data to string.
- The binary format is not opaque at all if we write document about it. If a player wants to edit their save data, reading a few lines of the save file format is a reasonable requirement. `pretty-save.js` is also a good example for them.

Beside the advantages of being a binary format, the gzip format also brings these benefits:
- No need to devise a new format and write the related document.
- Player can use gzip tools to get access to the json save string.
- If we need to add new data in the future, we can "piggyback" the gzip header by setting FLG.FEXTRA flag and using the "extra field" (check RFC 1952 for details). In this case, good documentation is mandatory.

d0sboots:
> If we need to add new data in the future, we can "piggyback" the gzip header by setting FLG.FEXTRA flag and using the "extra field" (check RFC 1952 for details). In this case, good documentation is mandatory.

Although I would guess/recommend that instead we would just extend the JSON inside the compressed gzip, instead of messing around with the gzip header.

electron/menu.js
```diff
@@ -81,7 +81,7 @@ function getMenu(window) {
               defaultPath: defaultPath,
               buttonLabel: "Load",
               filters: [
-                { name: "Game Saves", extensions: ["json", "json.gz", "txt"] },
+                { name: "Game Saves", extensions: ["save", "json", "json.gz", "txt"] },
```
d0sboots: Given that the format is now gzipped json, I think `json.gz` is the proper extension to use, so we don't need a custom opaque `.save` extension.

electron/saveDataBinaryFormat.js
```diff
@@ -0,0 +1,12 @@
+const magicBytes = [0x1f, 0x8b];
```
d0sboots:
```suggestion
// The 2 magic bytes of the gzip header plus the mandatory compression type of DEFLATE
const magicBytes = [0x1f, 0x8b, 0x08];
```

electron/storage.js
```diff
@@ -183,12 +192,31 @@ async function pushGameSaveToSteamCloud(base64save, currentPlayerId) {
...
+    // Let's decode the base64 string so GZIP is more efficient.
+    const buffer = Buffer.from(saveData, "base64");
+    compressedBuffer = await gzip(buffer);
+  }
+
+  /**
+   * When we push save file to Steam Cloud, we use greenworks.saveTextToFile. It seems that this method expects a string
+   * as the file content. That is why compressedBuffer is encoded in base64 and pushed to Steam Cloud as a text file.
+   *
+   * The comment below was added in the original commit which added support for Steam Cloud & local file system
```
d0sboots: There's no need to keep old, confusing comments around. Let's just keep this good explanation of why utf8 is inappropriate here, and remove the old cruft.

electron/storage.js
```diff
@@ -199,19 +227,37 @@ async function pushGameSaveToSteamCloud(base64save, currentPlayerId) {
   }
 }
 
+/**
+ * This function processes the save file in Steam Cloud and returns the save data in the base64 format. I don't change
+ * it to use the binary format, so there is no change in the logic of processing Steam Cloud save file.
```
d0sboots:
```suggestion
 * This function processes the save file in Steam Cloud and returns the save data in the base64 format. It is not
 * changed to use the binary format, so there is no change in the logic of processing Steam Cloud save file.
```

tsconfig.json
```diff
@@ -15,7 +15,8 @@
     "skipLibCheck": true,
     "sourceMap": true,
     "strict": true,
-    "target": "ESNext"
+    "target": "ESNext",
+    "allowJs": true
```
d0sboots: Why did this get added? It's a bit concerning.

tools/pretty-save.js
```diff
@@ -1,19 +1,27 @@
...
-  const save = JSON.parse(decodeURIComponent(escape(atob(data))));
+  const save = JSON.parse(jsonSaveString);
   const saveData = save.data;
   let gameSave = {
     PlayerSave: JSON.parse(saveData.PlayerSave),
     CompaniesSave: JSON.parse(saveData.CompaniesSave),
     FactionsSave: JSON.parse(saveData.FactionsSave),
     AliasesSave: JSON.parse(saveData.AliasesSave),
     GlobalAliasesSave: JSON.parse(saveData.GlobalAliasesSave),
-    MessagesSave: JSON.parse(saveData.MessagesSave),
```
d0sboots, line 16: Was this a bad merge? This feels like a mistake.

src/utils/compressionUtils.ts
```diff
@@ -0,0 +1,9 @@
+export async function compress(dataString: string): Promise<Uint8Array> {
```
d0sboots: Maybe Snarling will have a different opinion, but I don't think we're going to use compression anywhere else, so this can just go in saveDataUtils.

src/utils/saveDataUtils.ts
```diff
@@ -0,0 +1,28 @@
...
+  return "CompressionStream" in globalThis;
+}
+
+export async function encodeJsonSaveString(jsonSaveString: string, useBinaryFormat: boolean): Promise<SaveData> {
+  // Fallback to base64 format if the player doesn't want to compress the save data or their browser does not support
+  // Compression Streams API.
+  if (useBinaryFormat && canUseBinaryFormat()) {
+    return await compress(jsonSaveString);
+  } else {
+    return btoa(unescape(encodeURIComponent(jsonSaveString)));
```
d0sboots:
```suggestion
    // The unescape(encodeURIComponent()) pair does utf8 encoding into a "binary string",
    // despite (at first glance) seeming like a no-op.
    return btoa(unescape(encodeURIComponent(jsonSaveString)));
```

src/Electron.tsx
```diff
@@ -85,7 +84,7 @@ function initWebserver(): void {
     if (!path) return { res: false, msg: "Invalid file path." };
     if (!hasScriptExtension(path)) return { res: false, msg: "Invalid file extension: must be a script" };
 
-    code = Buffer.from(code, "base64").toString();
```
d0sboots, line 88: This seems like a spurious change.

src/GameOptions/ui/SystemPage.tsx
```diff
@@ -154,6 +155,13 @@ export const SystemPage = (): React.ReactElement => {
         />
       </>
 
+      <OptionSwitch
```
d0sboots: I don't think this option is needed. If compression is better (and presumably, we've measured, and it is?), then we should just use it.

As my grand-boss liked to say: "Just make it an option!" https://i.ibb.co/NV3BxqG/6548f64c-6409-4e20-95db-c01c2a7db0ee.jpg

src/SaveObject.ts
```diff
@@ -38,20 +38,27 @@ import { Terminal } from "./Terminal";
...
+import { isBinaryFormat } from "../electron/saveDataBinaryFormat";
 
 /* SaveObject.js
  *  Defines the object used to save/load games
  */
 
-export interface SaveData {
+/**
+ * This interface is only used for transferring game data to electron-related code, so I rename it from SaveData to
+ * ElectronGameData.
```
d0sboots:
```suggestion
 * This interface is only for transferring game data to electron-related code.
```

Code comments should reflect the (current) state of the code, you can use PR comments to discuss/justify changes being made in a particular change.

src/SaveObject.ts
```diff
@@ -110,115 +117,103 @@ class BitburnerSaveObject {
...
+      playerIdentifier: Player.identifier,
+      fileName: this.getSaveFileName(),
+      save: saveData,
+      savedOn,
+    };
+    pushGameSaved(electronGameData);
+
+    if (emitToastEvent) {
+      SnackbarEvents.emit("Game Saved!", ToastVariant.INFO, 2000);
+    }
```
d0sboots, line 143: Technically you should wrap this in a `try catch` that does console.error(err) followed by re-throw, to keep parity.

I'm not sure how important that is, maybe break `save` intentionally and see how good the error reporting is without it?

src/SaveObject.ts
```diff
@@ -110,115 +117,103 @@ class BitburnerSaveObject {
...
     // Save file name is based on current timestamp and BitNode
     const epochTime = Math.round(Date.now() / 1000);
     const bn = Player.bitNodeN;
-    let filename = `bitburnerSave_${epochTime}_BN${bn}x${Player.sourceFileLvl(bn) + 1}.json`;
+    /**
+     * Old behavior: save file uses .json extension. Save data is json save string in base64 format.
+     *
+     * New behavior:
+     * - Enable compression: save file uses .json.gz extension. Save data is json save string in binary format.
+     * - Disable compression: save file uses .save extension. Save data is json save string in base64 format.
```
d0sboots: Even though the `.json` extension is a misnomer, because it's actually base64'd, I don't think we should be changing filenames for compatibility. Especially because, even though it's wrong, .json gives more of a clue about what's in the file than `.save`.

d0sboots:
```suggestion
     * - Compression available: save file uses .json.gz extension. Save data is gzipped json save string in binary format.
     * - Compression unavailable: save file uses .json extension. Save data is json save string in base64 format.
```

Same note about "comments document the current code."

src/Settings/Settings.ts
```diff
@@ -46,6 +47,8 @@ export const Settings = {
   RemoteFileApiAddress: "localhost",
   /** Port the Remote File API client will try to connect to. 0 to disable. */
   RemoteFileApiPort: 0,
+  /** Whether to compress the save data. */
+  CompressSaveData: false,
```
d0sboots: Flagging here as well (per other comment)

src/ui/LoadingScreen.tsx
```diff
@@ -31,27 +31,24 @@ export function LoadingScreen(): React.ReactElement {
...
-          }
-
-          pushGameReady();
-          setLoaded(true);
-        })
-        .catch((reason) => {
-          console.error(reason);
-          Engine.load("");
+    load()
+      .then(async (saveData) => {
```
d0sboots: This is the weirdest mishmash of `async` and promise chains. IMO, let's chop out a layer of this.

If we (reasonably) assume that `pushGameReady` and `setLoaded` won't throw (they're quite simple and I've never seen any issues with them), then the final catch is really just catching the result of `throw err`. So it can all be merged into the upper catch as
```js
console.errror(err);
ActivateRecoveryMode();
await Engine.load("");
setLoaded(true);
```

tools/pretty-save.js
```diff
@@ -1,19 +1,27 @@
 /* eslint-disable @typescript-eslint/no-var-requires */
 const fs = require("fs").promises;
 const path = require("path");
+const { isBinaryFormat } = require("../electron/saveDataBinaryFormat");
 
 async function getSave(file) {
-  const data = await fs.readFile(file, "utf8");
+  const data = await fs.readFile(file);
+
+  let jsonSaveString;
```
d0sboots, line 9: Why is this duplicating code from the utils?

d0sboots (commented):
Obviously this is mutually exclusive with #1159. I'm leaning towards this one, because it adds a lot of useful comments, and heavens knows our codebase is lacking in those. :D @Snarling is going to need to be involved in some decision-making here though, since how we do compression is a big decision with long-lasting consequences.

tools/pretty-save.js
Snarling: Looks like MessagesSave was part of the save file at one point, which was removed in https://github.com/bitburner-official/bitburner-src/commit/dfadfcea0932ace33cd82af654836d3cf088133d in April 2022

I personally have never used this pretty-save.js tool, I imagine not a lot of people do so this discrepancy between the actual save object and the code in pretty-save.js went unnoticed for a while.

src/utils/compressionUtils.ts
Snarling: Either is fine with me.

Actually we do use zip files for the `download` terminal command when downloading multiple files. So potentially we could reuse some utilities between the two (doesn't really matter though because the `download` terminal command doesn't really need to be optimized as it's very rare in usage)

tools/pretty-save.js
catloversg: As Snarling said, `MessagesSave` was removed in a previous commit, so I will mark this as "Resolved conversation".

src/Electron.tsx
catloversg: That line decodes `code`. I change it to `decodeURIComponent(escape(atob(code)))` to remove the usage of Buffer. That's why I can remove `buffer` in `package.json`.
Test code:
```js
const string = "UTF-8 string: ❤️";
const base64String = btoa(unescape(encodeURIComponent(string)));
const string1 = Buffer.from(base64String, "base64").toString();
const string2 = decodeURIComponent(escape(atob(base64String)));
console.log(string1 === string2);
```

tools/pretty-save.js
Snarling: This is a standalone js tool, it's not going to be able to import code from the ts utils files.

catloversg:
> > If we need to add new data in the future, we can "piggyback" the gzip header by setting FLG.FEXTRA flag and using the "extra field" (check RFC 1952 for details). In this case, good documentation is mandatory.
> 
> Although I would guess/recommend that instead we would just extend the JSON inside the compressed gzip, instead of messing around with the gzip header.

Just a small clarification: "add new data" means the fancy feature that embeds `ImportPlayerData` in the save file, so we can show the `Compare Save` UI without decompressing the entire save file. I doubt that we will implement this kind of feature in the near future, so the note about "piggyback" the gzip header is just for documenting the idea in the case that it's needed.

In most cases, I agree that we should extend the JSON data.

tsconfig.json
catloversg: It allows importing JS file (saveDataBinaryFormat.js) without declaration file. I will commit saveDataBinaryFormat.d.ts and remove this setting later.

src/SaveObject.ts
catloversg: Re-throwing the error creates an unhelpful popup:
error-popup
I'll change it to this:
```typescript
try {
  await save(saveData);
} catch (error) {
  console.error(error);
  dialogBoxCreate(`Cannot save game. Error: ${error}.`);
  return;
}
```

d0sboots:
Codewise, this looks solid. I'm gonna need to manually test a lot of the edge cases to make sure nothing weird slipped through, though. If you could document the cases you've already tested, that would be helpful.

catloversg:
@d0sboots My test cases:
- Web version:
  - Try to import base64 format, then test export feature.
  - Try to import new exported save file, then test export feature.
- Electron version:
  - Try to import base64 format, then test export feature.
  - Try to import new exported save file, then test export feature.
  - Test: "Export save".
  - Test: "Load last save".
  - Test: "Load from file".
    - Import base64 format.
    - Import binary format.
  - Check the auto-save generated after quitting electron app:
    - Test if we can decompress it.
    - Test if we can import it.

I did NOT test the Steam Cloud, so please test it for me.

catloversg:
My test cases:
- Web version:
  - Import base64 format, then test export feature.
  - Import new exported save file, then test export feature.
- Electron version:
  - Import base64 format, then test export feature.
  - Import new exported save file, then test export feature.
  - Test: "Export save".
  - Test: "Load last save".
  - Test: "Load from file".
    - Import base64 format.
    - Import binary format.
  - Check the auto-save generated after quitting electron app:
    - Test if we can decompress it.
    - Test if we can import it.

I did NOT test the Steam Cloud, so please test it for me.

d0sboots:
I finally got time to build and test this change myself. It looks like you covered the "standard" cases pretty well, so I was focused on the unusual/error cases. My testing was web-exclusive.
* Loading my previously existing save is fine.
* Exported save is indeed a gzipped-json that can be inspected with standard tools.
* Emulated a non-CompressionStream-enabled browser while my savegame was compressed.
  * This leads to a generic recovery screen. The only hint as to what went wrong was in the console log. **This should probably be fixed.** Maybe an addition/tweak to RecoveryRoot.tsx?
* Tried to import a compressed save in the same situation as above. This leads to a generic "Save game is invalid" toast. A better toast would be good here, too.
* Changed the CRC in the gzip to be invalid. Got `TypeError: Failed to fetch` as the console error. The same was returned for trying to load a file that consists *only* of the 3 bytes that make up the magic header. **I think the use of `Response` is masking any useful results in the error messages.**
* Tried to import a 2MB .gz file that uncompresses to 2GB of zeros. The tab crashed. Since this seems to be happening before anything we have control over, no real issue, I just wanted to know the result.
* Did the same for a 512KB file uncompressing to 512MB. This one fails successfully. It doesn't look like there are any permanent memory leaks.

catloversg:
New changes:
- Custom recovery mode UI in 2 cases:
  - Browser does not support Compression Streams API.
  - Cannot decompress save data in IndexDB.
- Pipe `decompressedReadableStream` through `TextDecoderStream`. Error message is not masked anymore.
- If there is a `SaveDataError` when decoding save data, the toast shows a relevant error message.

src/utils/SaveDataUtils.ts
```diff
@@ -11,7 +22,24 @@ async function compress(dataString: string): Promise<Uint8Array> {
...
+  // Use "done" here to stop Lint from showing error
+  const done = false;
+  try {
+    while (!done) {
+      const readResult = await reader.read();
+      if (readResult.done) {
+        break;
+      }
+      result += readResult.value;
+    }
```
d0sboots: I've done this before, and I think there's a more elegant way of structuring this.

```suggestion
  const reader =  decompressedReadableStream.pipeThrough(new TextDecoderStream("utf-8", {fatal: true})).getReader();
  let result = "";
  try {
    for (let {value, done} = await reader.read(); !done; {value, done} = await reader.read()) {
        result += value;
    }
```

Also, there's no need for the `finally` since the reader immediately goes out of scope anyway (and in general, it is single-use.)

d0sboots:
Let's goooooooo

Fingers crossed XD

d0sboots:
Oh, for posterity maybe edit the base PR description to describe how it ended up (some of the stuff there got changed as the PR evolved)

PR #1167, opened 2024-03-17
MISC: Remove jquery

catloversg:
This PR replaces the usage of `jquery.get` with `fetch`. After that, we can remove the `jquery` library.

One concern that I notice when making this PR is the duplication of logic between many functions in `NetscriptFunctions.ts` and code in `src\Terminal\commands\*.ts`. It may require a big PR to solve that problem, so I skip it in this PR.

### DIFF
--- a/src/NetscriptFunctions.ts
+++ b/src/NetscriptFunctions.ts
@@ -1636,37 +1636,42 @@ export const ns: InternalAPI<NSFull> = {
       });
     });
   },
-  wget: (ctx) => (_url, _target, _hostname) => {
+  wget: (ctx) => async (_url, _target, _hostname) => {
     const url = helpers.string(ctx, "url", _url);
     const target = helpers.filePath(ctx, "target", _target);
     const hostname = _hostname ? helpers.string(ctx, "hostname", _hostname) : ctx.workerScript.hostname;
     const server = helpers.getServer(ctx, hostname);
     if (!target || (!hasTextExtension(target) && !hasScriptExtension(target))) {
       helpers.log(ctx, () => `Invalid target file: '${target}'. Must be a script or text file.`);
-      return Promise.resolve(false);
+      return false;
     }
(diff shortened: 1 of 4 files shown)


d0sboots:
> One concern that I notice when making this PR is the duplication of logic between many functions in NetscriptFunctions.ts and code in src\Terminal\commands\*.ts.

Yes, this is both large and very complicated (the duplication is not simple to resolve), so it should probably be done in small pieces.

src/NetscriptFunctions.ts
```diff
@@ -1647,22 +1646,26 @@ export const ns: InternalAPI<NSFull> = {
       return Promise.resolve(false);
     }
     return new Promise(function (resolve) {
-      $.get(
-        url,
-        function (data) {
-          const res = server.writeToContentFile(target, data);
-          if (res.overwritten) {
+      fetch(url)
```
d0sboots: IMO the flow of all of this would be simpler if you promoted wget itself (inside the ctx wrapper) to async. Then you can just return true/false directly, and catch the error on fetch to do error handling instead of it being buried later.

src/Terminal/commands/wget.ts
```diff
@@ -1,37 +1,34 @@
...
       }
-      return Terminal.print(`wget successfully retrieved content to new file ${target}`);
-    },
-    "text",
-  ).fail(function (e) {
-    return Terminal.error("wget failed: " + JSON.stringify(e));
-  });
+    })
+    .catch((reason) => {
+      Terminal.error(`wget failed: ${JSON.stringify(reason, Object.getOwnPropertyNames(reason))}`);
```
d0sboots, line 33: What's the reason for the filtering? I could see it leaving out info, depending on how error classes are composed.

catloversg: I added comment for the reason in the latest commit.

PR #1175, opened 2024-03-19
IPVGO: Remove current game history from savefile, re-implement superko

ficocelliguy:
* Re-implemented superko (repeated position prevention) since the Go AI loves getting into infinite loops, and optimal play is sometimes to capture even if it continues the loop. (The first player to 'chicken out' of the loop gets a worse position, in some cases)

* Remove prior moves on the current game from save history, to prevent increasing the size of the game save. Short loops will still be halted by the superko rule even if the player saves and loads going into a capture loop.


Snarling:
Seems good. I only did a brief review and I'll just take a closer look before merging this. I'll merge some of the earlier Go PRs so there's time for any possible rebasing that's required.

This does allow the player to get away with repeating a move if they save/reload to cheese the game into forgetting board history. But it's not like there aren't plenty of other ways the player can cheese by abusing save/load, so I think it's fine.

ficocelliguy:
It's true that a player could save and reload to make a move that would be otherwise forbidden by repeating - but in most cases their opponent could capture right back, since THEIR move history was also removed on reload! I don't think it wouldn't gain the player much, if anything, in scenarios I can think of.

PR #1177, opened 2024-03-19
SLEEVES: add nextCompletion to SleeveInfiltrationWork

Caldwell-74:
did some git mess 
just remade it with the fix


d0sboots:
> just remade it with the fix

`git push --force-with-lease` 😀

It's not a big deal, but (for the future) that way we didn't lose the PR history

d0sboots:
Just double checking, sleeves still work right, right?

Caldwell-74:
i can link the history here #1087
for if it works from my quick test yesterday it does

PR #1121, opened 2024-02-22
CORPORATION: Add documentation

catloversg:
This PR adds documentation for Corporation.
Closes #700.
Screenshots:
screenshot1
screenshot2
screenshot3

PR #1179, opened 2024-03-20
CORPORATION: more granular office size upgrades

### LINKED ISSUE #1166: CORPORATION: expanding office size by a non-multiple of 3 has incorrect cost
# CORPORATION: expanding office size by a non-multiple of 3 has incorrect cost

## Summary

`corporation.upgradeOfficeSize` rounds up expansion cost to the nearest multiple of 3, but does not actually increase the size accordingly.  Either the cost should not be rounded up for non-multiples of 3 or the actual increase should match the cost.

## Issue Reproduction

The (extremely quick-and-dirty) demonstration script `test.js` (source below) demonstrates the issue by creating a new division, then calling `upgradeOfficeSize` to increase the max size of the offices in three different cities by 1, 2, and 3 employees respectively:

```
[home /]> run test.js
Running script with 1 thread(s), pid 2 and args: [].
test.js: Expanded AgriTest office size in Sector-12 by 1 at a total cost of: -4.360e9
test.js: Expanded AgriTest office size in Aevum by 2 at a total cost of: -4.360e9
test.js: Expanded AgriTest office size in Ishima by 3 at a total cost of: -4.360e9
```

Note that increasing the Sector-12 office from 3->4 cost 4.36b and increasing Ishima from 3->6 also cost 4.36b.

The relevant functions that compute the costs (`UpgradeOfficeSize` in `Actions.ts` and `corporation.getOfficeSizeUpgradeCost`) haven't changed since at least #503 (2023-05-15).

<details><summary>test.js demonstration script (click to expand)</summary>

```
/** @param {NS} ns */
  const div = `AgriTest${ns.args[0] ?? ""}`;
  ns.corporation.expandIndustry("Agriculture", div);
  ns.corporation.expandCity(div, "Aevum");
  ns.corporation.expandCity(div, "Ishima");
  ns.corporation

  const cities = ["Z", "Sector-12", "Aevum", "Ishima"];

  for (let i=1; i<4;i++) {
    const funds_pre = ns.corporation.getCorporation().funds;
    const size_pre = ns.corporation.getOffice(div, cities[i]).size;
    ns.corporation.upgradeOfficeSize(div, cities[i], i);
    const funds_post = ns.corporation.getCorporation().funds;
    const size_post = ns.corporation.getOffice(div, cities[i]).size;
    ns.tprint(`Expanded ${div} office size in ${cities[i]} by ${size_post - size_pre} at a total cost of: ${ns.formatNumber(funds_post - funds_pre)}`);
  }
}
```
</details>

## Suggested fixes

The easiest would probably be to just use something like
```
office.size += Math.ciel(requested_size_increase / 3) * 3;
```

for the actual increase as the cost computations wouldn't need to change.

Alternately, office size increases could be made more granular with single-employee size increases supported by lower costs.  This would be more work to implement as the existing cost curve would need to be matched and the cost computations rewritten.  However, it's my preferred solution as it rewards scripters over UI corporation users by allowing finer control over office size.

~~I'm happy to work on this, but I'll leave a little while for discussion before I start.~~

I've started on this in #1179.

adeilt:
# CORPORATION: more granular office size upgrades

Allows `corporation.upgradeOfficeSize` to increase the size of a Corporation office by a non-multiple of 3 and also be charged a corresponding amount of corporate funds.  See #1166 for details of current behavior.

### Before

```
[home /]> run test.js
Running script with 1 thread(s), pid 1 and args: [].
test.js: Expanded AgriTest office size in Sector-12 from 3 to 4 at a total cost of: -4.360e9
test.js: Expanded AgriTest office size in Aevum from 3 to 5 at a total cost of: -4.360e9
test.js: Expanded AgriTest office size in Ishima from 3 to 6 at a total cost of: -4.360e9
```

### After

```
[home /]> run test.js 
Running script with 1 thread(s), pid 2 and args: [].
test.js: Expanded AgriTest office size in Sector-12 from 3 to 4 at a total cost of: -1.412e9
test.js: Expanded AgriTest office size in Aevum from 3 to 5 at a total cost of: -2.865e9
test.js: Expanded AgriTest office size in Ishima from 3 to 6 at a total cost of: -4.360e9
```

## Plan [DONE]

1. **[DONE]** Factor out the corporate office size upgrade cost computation into a helper function.
2. **[DONE]** Add tests to make sure the values closely match those in v2.6.0 of the game.
31415926. **[DONE]** Replace computation with one that will allow upgrade cost to change with each employee slot added instead of only in multiples of 3. (Thanks for the actual formula used, @d0sboots!)
5. **[DONE]** Update documentation as required. (Deferring additional documentation changes to @catloversg in #1191.)
9. **[DONE]** Profit!!!!1! (this is, after all, about Corporate stuff)

## Linked issues

closes #1166

# Testing

- Some unit tests have been added for the actual computation.
- UX changes were tested manually to make sure UX behavior had not changed.

### DIFF
--- a/src/Corporation/helpers.ts
+++ b/src/Corporation/helpers.ts
@@ -13,13 +13,13 @@ export function calculateUpgradeCost(corporation: Corporation, upgrade: CorpUpgr
   return cost;
 }
 
-export function calculateOfficeSizeUpgradeCost(currentSize: number, sizeIncrease: number): number {
-  if (sizeIncrease < 0) throw new Error("Invalid value for sizeIncrease argument! Must be at least 0!");
-  if (sizeIncrease == 0) return 0;
-  const baseCostMultiplier = 1.09;
+export function calculateOfficeSizeUpgradeCost(currentSize: number, sizeIncrease: PositiveInteger): number {
+  if (sizeIncrease <= 0) throw new Error("Invalid value for sizeIncrease argument! Must be at least 0!");
+  const baseCostDivisor = 0.09;
+  const baseCostMultiplier = 1 + baseCostDivisor;
   const currentSizeFactor = baseCostMultiplier ** (currentSize / 3);
   const sizeIncreaseFactor = baseCostMultiplier ** (sizeIncrease / 3) - 1;
-  return (corpConstants.officeInitialCost / (baseCostMultiplier - 1)) * currentSizeFactor * sizeIncreaseFactor;
+  return (corpConstants.officeInitialCost / baseCostDivisor) * currentSizeFactor * sizeIncreaseFactor;
 }
 
 export function calculateMaxAffordableUpgrade(corp: Corporation, upgrade: CorpUpgrade): 0 | PositiveInteger {
--- a/src/Corporation/helpers.ts
+++ b/src/Corporation/helpers.ts
@@ -16,14 +16,10 @@ export function calculateUpgradeCost(corporation: Corporation, upgrade: CorpUpgr
 export function calculateOfficeSizeUpgradeCost(currentSize: number, sizeIncrease: number): number {
   if (sizeIncrease < 0) throw new Error("Invalid value for sizeIncrease argument! Must be at least 0!");
   if (sizeIncrease == 0) return 0;
-  const baseSize = corpConstants.officeInitialSize;
-  const initialPriceMult = Math.round(currentSize / baseSize);
-  const costMultiplier = 1.09;
-  let mult = 0;
-  for (let i = 0; i < sizeIncrease / baseSize; ++i) {
-    mult += Math.pow(costMultiplier, initialPriceMult + i);
-  }
-  return corpConstants.officeInitialCost * mult;
+  const baseCostMultiplier = 1.09;
+  const currentSizeFactor = baseCostMultiplier ** (currentSize / 3);
+  const sizeIncreaseFactor = baseCostMultiplier ** (sizeIncrease / 3) - 1;
+  return (corpConstants.officeInitialCost / (baseCostMultiplier - 1)) * currentSizeFactor * sizeIncreaseFactor;
 }
 
 export function calculateMaxAffordableUpgrade(corp: Corporation, upgrade: CorpUpgrade): 0 | PositiveInteger {
--- a/src/Corporation/helpers.ts
+++ b/src/Corporation/helpers.ts
@@ -3,6 +3,7 @@ import { PositiveInteger, isPositiveInteger } from "../types";
 import { formatShares } from "../ui/formatNumber";
 import { Corporation } from "./Corporation";
 import { CorpUpgrade } from "./data/CorporationUpgrades";
+import * as corpConstants from "./data/Constants";
 
 export function calculateUpgradeCost(corporation: Corporation, upgrade: CorpUpgrade, amount: PositiveInteger): number {
   const priceMult = upgrade.priceMult;
@@ -10,6 +11,19 @@ export function calculateUpgradeCost(corporation: Corporation, upgrade: CorpUpgr
   const baseCost = upgrade.basePrice * Math.pow(priceMult, level);
   const cost = (baseCost * (1 - Math.pow(priceMult, amount))) / (1 - priceMult);
   return cost;
+}
+
+export function calculateOfficeSizeUpgradeCost(currentSize: number, sizeIncrease: number): number {
+  if (sizeIncrease < 0) throw new Error("Invalid value for sizeIncrease argument! Must be at least 0!");
+  if (sizeIncrease == 0) return 0;
+  const baseSize = corpConstants.officeInitialSize;
+  const initialPriceMult = Math.round(currentSize / baseSize);
+  const costMultiplier = 1.09;
+  let mult = 0;
+  for (let i = 0; i < sizeIncrease / baseSize; ++i) {
+    mult += Math.pow(costMultiplier, initialPriceMult + i);
+  }
+  return corpConstants.officeInitialCost * mult;
 }
 
 export function calculateMaxAffordableUpgrade(corp: Corporation, upgrade: CorpUpgrade): 0 | PositiveInteger {
--- a/src/Documentation/doc/advanced/corporation/office.md
+++ b/src/Documentation/doc/advanced/corporation/office.md
@@ -55,8 +55,6 @@ Maximum size with a given `MaxCost`:
 
 $$MaxUpgradeLevel = 3\ast\log_{1.09}\left( MaxCost\ast\frac{0.09}{BasePrice} + {1.09}^{\frac{CurrentSize}{3}} \right)$$
 
-When we divide office's size by 3, we need to use `Math.ceil()` to round up the value. This means upgrading from size 3 to size 4 costs the same as upgrading from size 3 to size 6. The most cost-effective size is always a multiple of 3.
-
 ## Energy and morale
 
 They are calculated in START state.
--- a/src/NetscriptFunctions/Corporation.ts
+++ b/src/NetscriptFunctions/Corporation.ts
@@ -63,7 +63,7 @@ import { InternalAPI, NetscriptContext, setRemovedFunctions } from "../Netscript
 import { helpers } from "../Netscript/NetscriptHelpers";
 import { getEnumHelper } from "../utils/EnumHelper";
 import { MaterialInfo } from "../Corporation/MaterialInfo";
-import { calculateUpgradeCost } from "../Corporation/helpers";
+import { calculateOfficeSizeUpgradeCost, calculateUpgradeCost } from "../Corporation/helpers";
 import { PositiveInteger } from "../types";
 import { getRecordKeys } from "../Types/Record";
 
@@ -511,13 +511,15 @@ export function NetscriptCorporation(): InternalAPI<NSCorporation> {
       const size = helpers.number(ctx, "size", _size);
       if (size < 0) throw new Error("Invalid value for size field! Must be numeric and greater than 0");
       const office = getOffice(divisionName, cityName);
-      const initialPriceMult = Math.round(office.size / corpConstants.officeInitialSize);
-      const costMultiplier = 1.09;
-      let mult = 0;
-      for (let i = 0; i < size / corpConstants.officeInitialSize; ++i) {
-        mult += Math.pow(costMultiplier, initialPriceMult + i);
-      }
-      return corpConstants.officeInitialCost * mult;
+      return calculateOfficeSizeUpgradeCost(office.size, size);
+      // const initialPriceMult = Math.round(office.size / corpConstants.officeInitialSize);
+      // const costMultiplier = 1.09;
+      // let mult = 0;
+      // //
+      // for (let i = 0; i < size / corpConstants.officeInitialSize; ++i) {
+      //   mult += Math.pow(costMultiplier, initialPriceMult + i);
+      // }
+      // return corpConstants.officeInitialCost * mult;
     },
     setAutoJobAssignment: (ctx) => (_divisionName, _cityName, _job, _amount) => {
       checkAccess(ctx, CorpUnlockName.OfficeAPI);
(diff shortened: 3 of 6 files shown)


catloversg:
@adeilt I can handle the documentation for you. There are wrong formulas in the documentation, so I'm updating it.

adeilt:
> @adeilt I can handle the documentation for you. There are wrong formulas in the documentation, so I'm updating it.

Thanks!  I'm a big fan of #1121, by the way.

I'm not very familiar with GitHub's PR processes and Bitburner's policies/conventions yet.  Are the documentation changes something that we should (could?) add to this PR or should they be handled separately?  If the latter, then I'll make a final pass over my changes and mark this as ready for review.

catloversg:
You should mark it as ready for review, so our maintainers know that they can review it now. The documentation can be updated in a follow-up PR after all the changes have been finalized. If our maintainers want to update the documentation in this PR, I can write the patch and send it to you.

d0sboots:
> I'm not very familiar with GitHub's PR processes and Bitburner's policies/conventions yet. Are the documentation changes something that we should (could?) add to this PR or should they be handled separately?

Normally I'd say do the doc changes in the same PR, but if cat's going to do the doc changes then separate is fine for easier Git logistics.

src/Corporation/helpers.ts
```diff
@@ -12,6 +13,15 @@ export function calculateUpgradeCost(corporation: Corporation, upgrade: CorpUpgr
   return cost;
 }
 
+export function calculateOfficeSizeUpgradeCost(currentSize: number, sizeIncrease: number): number {
+  if (sizeIncrease < 0) throw new Error("Invalid value for sizeIncrease argument! Must be at least 0!");
+  if (sizeIncrease == 0) return 0;
+  const baseCostMultiplier = 1.09;
+  const currentSizeFactor = baseCostMultiplier ** (currentSize / 3);
+  const sizeIncreaseFactor = baseCostMultiplier ** (sizeIncrease / 3) - 1;
+  return (corpConstants.officeInitialCost / (baseCostMultiplier - 1)) * currentSizeFactor * sizeIncreaseFactor;
```
d0sboots: This might be why your test failed - The expression `(1.09 - 1)` is much less accurate than `0.09`. (This is known as "catastrophic cancellation." https://en.wikipedia.org/wiki/Catastrophic_cancellation#Example:_Radix_conversion)

src/Corporation/helpers.ts
```diff
@@ -12,6 +13,15 @@ export function calculateUpgradeCost(corporation: Corporation, upgrade: CorpUpgr
   return cost;
 }
 
+export function calculateOfficeSizeUpgradeCost(currentSize: number, sizeIncrease: number): number {
+  if (sizeIncrease < 0) throw new Error("Invalid value for sizeIncrease argument! Must be at least 0!");
+  if (sizeIncrease == 0) return 0;
```
d0sboots: Optional: This is unneeded, it will come out to exactly 0 when sizeIncrease is 0.

src/Corporation/helpers.ts
```diff
@@ -12,6 +13,15 @@ export function calculateUpgradeCost(corporation: Corporation, upgrade: CorpUpgr
   return cost;
 }
 
+export function calculateOfficeSizeUpgradeCost(currentSize: number, sizeIncrease: number): number {
```
d0sboots: I feel like `sizeIncrease` should be an `Integer`, from types.ts. This might cause type issues though, so if it gives you too much trouble you don't need to do it.

src/NetscriptFunctions/Corporation.ts
```diff
@@ -511,13 +511,7 @@ export function NetscriptCorporation(): InternalAPI<NSCorporation> {
       const size = helpers.number(ctx, "size", _size);
```
d0sboots: should be helpers.positiveInteger, and then the extra check isn't needed

src/NetscriptFunctions/Corporation.ts
```diff
@@ -559,7 +553,9 @@ export function NetscriptCorporation(): InternalAPI<NSCorporation> {
       const divisionName = helpers.string(ctx, "divisionName", _divisionName);
       const cityName = getEnumHelper("CityName").nsGetMember(ctx, _cityName);
       const size = helpers.number(ctx, "size", _size);
```
d0sboots: same

test/jest/Corporation.test.ts
```diff
@@ -112,4 +116,59 @@ describe("Corporation", () => {
       expectSharesToAddUp(corporation);
     });
   });
+
+  describe("helpers.calculateOfficeSizeUpgradeCost", () => {
+    // These values were pulled from v2.6.0 as reference test values
+    const refCosts = {
+      from3: {
```
d0sboots: Here's the results of a `bc` program (bc is the unix-standard arbitrary-precision calculator):
```bc
define cost(size, increase) {
  auto i, mult, old_scale, incr3, size3;
  old_scale = scale;
  scale = 0;
  incr3 = increase / 3;
  size3 = size / 3;
  scale = old_scale;
  for (; i < incr3; ++i) {
    mult += 1.09 ^ (size3 + i);
  }
  return 4000000000 * mult;
}

scale=40
cost(3, 3)
cost(3, 15)
cost(3, 150)
print "\n"
cost(6, 3)
cost(6, 15)
cost(6, 150)
print "\n"
cost(9, 3)
cost(9, 15)
cost(9, 150)
quit
```
```
4360000000.00
26093338259.6000000000
3553764305895.2490263450780533572274291097124000000000

4752400000.0000
28441738702.964000000000
3873603093425.8214387161350781593778977295868000000000

5180116000.000000
31001495186.23076000000000
4222227371834.1453682005872351937219085252500000000000
```
As you can see, there was some mild inaccuracy in the old method. I'd use these numbers (rounded to double-precision level, i.e. 18 digits) instead.

test/jest/Corporation.test.ts
```diff
@@ -112,4 +116,59 @@ describe("Corporation", () => {
...
+        150: 3873603093425.833,
+      },
+      from9: {
+        3: 5180116000.000002,
+        15: 31001495186.230778,
+        150: 4222227371834.1587,
+      },
+    };
+
+    describe("upgrade office size from size 3", () => {
```
d0sboots: You can use table-driven tests to take all the redundancy out of this; you've already got most of the data structure set up, above.

d0sboots (commented):
Nice! Some comments about subtle numerical things.

src/Corporation/helpers.ts
adeilt: "Catastrophic Cancellation" is a great a name for that concept and it would also be a great name for a band.

src/Corporation/helpers.ts
adeilt: I think I put that in there when I was still getting a feel for the problem.  I saw a `while(testExpression) { do a bunch of iterative stuff }` and thought to short circuit the zero case right up front.  Not nearly as impactful with the explicit formula.

test/jest/Corporation.test.ts
adeilt: It's probably been decades since I last used `bc` and I never knew you could define functions in it.  That's handy to know about, thanks!

src/Corporation/helpers.ts
adeilt: If I'm understanding the problem correctly then this should fix the issue as addition shouldn't have the same trouble.  However, my numerical analysis is much weaker than my discrete math, so I'm prepared to be wrong.

src/Corporation/helpers.ts
adeilt: I think it's worth the effort to be more specific here.

adeilt: I went further and used `PositiveInteger`, but according to my understanding `sizeIncrease: PositiveInteger` doesn't actually validate the value of `sizeIncrease` like `helpers.positiveInteger` does, so I'm leaving this check in place.

test/jest/Corporation.test.ts
adeilt: I ended up just removing the least significant digit of each number until eslint stopped complaining that I'd lost precision at runtime.

test/jest/Corporation.test.ts
adeilt: I was previously unaware of table-driven tests.  Thanks; this is much cleaner now.

adeilt:
Thanks for the detailed review, @d0sboots.  I'm learning quite a bit.  :)

src/Corporation/helpers.ts
d0sboots: The correct thing to do with these types is to (runtime) check them *as* they are TS-asserted as `PositiveInteger` or `Integer`. The type assertion functions in `types.ts` do both - they take a generic `number` and assert that it has the proper form if the function returns true. You are expected to return/throw/otherwise abort flow-control if false, and then TS can know that the asserted-true path is also of the more restrictive type.

As a result, anything that is declared `PositiveInteger` *should have already be runtime checked as a positive integer.* Anything else is a bug.

Edit: That said, the casts you are doing seem OK, because you can statically guarantee that the numbers in question are PositiveIntegers. The only one I might change is the one where you have the loop.

src/Corporation/helpers.ts
d0sboots: Ah, I see what you mean. Addition *can* have the same problem, it's all about the specifics of the numbers involved. (Imagine adding two numbers like 1.01 and -1.02.) But for what you did, there won't be any issues.

test/jest/Corporation.test.ts
```diff
@@ -117,58 +117,24 @@ describe("Corporation", () => {
...
+      { fromSize: 6, increaseBy: 3, expectedCost: 4752400000.0 },
+      { fromSize: 6, increaseBy: 15, expectedCost: 28441738702.964 },
+      { fromSize: 6, increaseBy: 150, expectedCost: 3873603093425.821 },
+      { fromSize: 9, increaseBy: 3, expectedCost: 5180116000.0 },
+      { fromSize: 9, increaseBy: 15, expectedCost: 31001495186.23076 },
+      { fromSize: 9, increaseBy: 150, expectedCost: 4222227371834.145 },
+    ])(
+      "should cost $expectedCost to upgrade office by $increaseBy from size $fromSize",
+      ({ fromSize, increaseBy, expectedCost }) => {
+        expect(calculateOfficeSizeUpgradeCost(fromSize, increaseBy as PositiveInteger)).toBeCloseTo(expectedCost, 1);
```
d0sboots, line 136: Can you take out the `1` now? Although I wouldn't be surprised if not, Jests implementation of `toBeCloseTo` is completely dumb (it's strictly absolute error instead of relative error).

src/Documentation/doc/advanced/corporation/office.md
```diff
@@ -55,8 +55,6 @@ Maximum size with a given `MaxCost`:
 
 $$MaxUpgradeLevel = 3\ast\log_{1.09}\left( MaxCost\ast\frac{0.09}{BasePrice} + {1.09}^{\frac{CurrentSize}{3}} \right)$$
 
-When we divide office's size by 3, we need to use `Math.ceil()` to round up the value. This means upgrading from size 3 to size 4 costs the same as upgrading from size 3 to size 6. The most cost-effective size is always a multiple of 3.
```
catloversg, line 58: You should revert the change here. I'll handle it in #1191.

catloversg:
@adeilt Can you edit your original post and change the "Deferring additional documentation changes" part to #1191?

test/jest/Corporation.test.ts
adeilt: I can't, unfortunately.  I left it out at first when I moved to the new testing style, but the default 2 is too stringent and the +150 tests all failed.

adeilt:
@catloversg, didn't catch that change.  Thanks and done.

src/Documentation/doc/advanced/corporation/office.md
d0sboots: I'm going to go ahead and merge as-is, and you can fix it up more later.

adeilt: Apologies.  I've been fighting my home network and missed this.

src/Corporation/helpers.ts
adeilt: The loop one _should_ be ok because the while condition is the strict inequality `upgradeSizeMax > props.office.size` so `upgradeSizeMax - props.office.size` should always be positive.

d0sboots: But props.office.size could be a non-integer. (It shouldn't be, but the type system doesn't have anything to say about it.) I'm this case that's fine because the formula actually functions fine with non-integer sizeIncrease, even if that were to happen.

PR #1183, opened 2024-03-21
CREDITS: Add @d0sboots as a documented maintainer

Snarling:
Also updated hydro's credit entry to match what he prefers to be credited as.


Snarling:
Was reminded of the credits page by #1178 , checked up on it and realized I never added attribution for d0s as a maintainer. I also changed the entry for hydro to match his preferred attribution.

Credits page itself probably could use some other changes, but I just wanted to get this minimal change in first.

As far as other changes that probably should be made here at some point: It's not clear what the "Lead Developers" section is really talking about as that doesn't have a clear meaning. There probably should still be a permanent section for people who have contributed major things (although determining what is "major" is still fuzzy). Similar to the WIP contributors list over at #1178.

PR #1188, opened 2024-03-22
IPVGO: Nerf overly-difficult handicap, [issue #1169] misc bugfixes

### LINKED ISSUE #1169: IPvGO 5x5 two handicap is extreme
Hello,

In the IPvGo Subnet Illuminati/w0r1d_d43m0n 5x5 board, the handicap is 2. This is too high for reasonable play (this is coming from a ~6k go player).

**Background**
On a 5x5 board, the winners are highly constrained under optimal play, see https://www.researchgate.net/publication/2925531_Solving_Go_On_Small_Boards 

With no handicap, black will always win if they play optimally and start at the center, gaining control of the entire board. If black plays at C2, optimally black will win by 3 points in traditional scoring (not quite the same as the game). Any other position, B2 and the edges, white can win under optimal conditions.

** Problem Statement **
The inclusion of two handicap stones most likely makes it impossible for the non-handicapped player to win under optimal conditions. I recognize that the algorithms used are not optimal conditions, but it makes it really stringent/you need to understand how the algorithms mess up, rather than understanding how the game of Go works.  

Under current setup, to win, you practically have to kill the other group. The board size is 25, the komi is 7.5 (or 9.5) (https://github.com/bitburner-official/bitburner-src/blob/edf8e24046cf07fdffd4d74242a1f88ab919387c/src/Go/Constants.ts#L53), meaning the total number of points is 32.5 (or 34.5). The minimum survivable group size is 8. This means that if white has a group larger than (or equal to) the corner life, white will win.  

** Suggestion for change **
I would suggest changing for the board size of 5 to have a handicap of 1 instead of 2. This could be done by explicitly limiting the maximum handicap size of the 5x5 board to 1.  
https://github.com/bitburner-official/bitburner-src/blob/edf8e24046cf07fdffd4d74242a1f88ab919387c/src/Go/boardState/boardState.ts#L63

If this makes it **too** easy, I would suggest pinning the handicap of the 5x5 board at the center of the board, as that is the optimal position. This change would be in the same file and fairly easy to make as the board and the number of handicaps are already in the function call. I wouldn't be opposed to fixing the position of the handicaps on other boards because that is the standard practice for 9x9 (c3 and center) and 13x13 boards (d4 and center). 
https://github.com/bitburner-official/bitburner-src/blob/edf8e24046cf07fdffd4d74242a1f88ab919387c/src/Go/boardState/boardState.ts#L116

I wouldn't suggest changing the komi, because that would involve more code changes/displace the changes to multiple locations in the code base. 

Thanks, 
MR herb
https://online-go.com/player/558487/

ficocelliguy:
closes #1169 

Currently, the handicap of two stones on the 5x5 board is too strong (see discussion in issue linked above). This reduces the handicap and makes it less random where the stone is placed.

It also fixes a bug where extra offline nodes were applied to the mystery opponent.

Also adds unit tests for boardState.ts

### DIFF
--- a/src/Go/boardState/boardState.ts
+++ b/src/Go/boardState/boardState.ts
@@ -24,6 +24,7 @@ export function getNewBoardState(
   if (ai === GoOpponent.w0r1d_d43m0n) {
     boardToCopy = resetCoordinates(rotate90Degrees(boardFromSimpleBoard(bitverseBoardShape)));
     boardSize = 19;
+    applyObstacles = false;
   }
 
   const newBoardState: BoardState = {
@@ -64,7 +65,13 @@ export function getNewBoardState(
 export function getHandicap(boardSize: number, opponent: GoOpponent) {
   // Illuminati and WD get a few starting routers
   if (opponent === GoOpponent.Illuminati || opponent === GoOpponent.w0r1d_d43m0n) {
-    return ceil(boardSize * 0.35);
+    return {
+      [5]: 1,
+      [7]: 3,
+      [9]: 4,
+      [13]: 5,
+      [19]: 7,
+    }[boardSize];
   }
   return 0;
 }
@@ -123,6 +130,13 @@ export function applyHandicap(board: Board, handicap: number): void {
   const availableMoves = getEmptySpaces(board);
   const handicapMoveOptions = getExpansionMoveArray(board, availableMoves);
   const handicapMoves: Move[] = [];
+
+  // Special handling for 5x5: always have handicap in the center of the board
+  if (availableMoves.length < 26 && board[2][2]) {
+    board[2][2].color = GoColor.white;
+    updateChains(board);
+    return;
+  }
 
   // select random distinct moves from the move options list up to the specified handicap amount
   for (let i = 0; i < handicap && i < handicapMoveOptions.length; i++) {
--- a/src/Go/boardState/boardState.ts
+++ b/src/Go/boardState/boardState.ts
@@ -131,8 +131,8 @@ export function applyHandicap(board: Board, handicap: number): void {
   const handicapMoveOptions = getExpansionMoveArray(board, availableMoves);
   const handicapMoves: Move[] = [];
 
-  // Special handling for 5x5: always have handicap in the center of the board
-  if (availableMoves.length < 26 && board[2][2]) {
+  // Special handling for 5x5: extra weight on handicap piece in the center of the board
+  if (availableMoves.length < 26 && board[2][2] && Math.random() < 0.2) {
     board[2][2].color = GoColor.white;
     updateChains(board);
     return;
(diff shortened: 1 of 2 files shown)


src/Go/boardState/boardState.ts
```diff
@@ -64,7 +65,13 @@ export function getNewBoardState(
 export function getHandicap(boardSize: number, opponent: GoOpponent) {
   // Illuminati and WD get a few starting routers
   if (opponent === GoOpponent.Illuminati || opponent === GoOpponent.w0r1d_d43m0n) {
-    return ceil(boardSize * 0.35);
+    return {
+      [5]: 1,
```
d0sboots, line 69: Just make it an array! :trollface:

pls don't actually tho

src/Go/boardState/boardState.ts
```diff
@@ -124,6 +131,13 @@ export function applyHandicap(board: Board, handicap: number): void {
   const handicapMoveOptions = getExpansionMoveArray(board, availableMoves);
   const handicapMoves: Move[] = [];
 
+  // Special handling for 5x5: always have handicap in the center of the board
```
d0sboots: Per the bug thread, I'm not sure this is a good idea. Although I'm not sure it's a bad idea, either.

d0sboots (commented):
Yay more tests!

src/Go/boardState/boardState.ts
ficocelliguy: I adjusted this to favor the tengen point as a starting position more so than other options, but it won't usually pick it. (something like 22-24% of games will have the handicap point in the center)

markdown/bitburner.goopponent.md
```diff
@@ -8,5 +8,5 @@
 **Signature:**
 
 ```typescript
-type GoOpponent = "Netburners" | "Slum Snakes" | "The Black Hand" | "Tetrads" | "Daedalus" | "Illuminati";
+type GoOpponent = "Netburners" | "Slum Snakes" | "The Black Hand" | "Tetrads" | "Daedalus" | "Illuminati" | "????????????";
```
d0sboots: Didn't you explicitly remove ????? from GoOpponent to fix a bug? Is this not just reintroducing the bug?

Edit: Regardless, this feels completely unrelated to changing 5x5 boards.

ficocelliguy: This is just the visual documentation for the Go API, it is different than the core enum change you are thinking of. netscript definitions can't seem to import things from outside this file, so it's just a local type to the documentation. (I added unit tests for the original bug accessing the secret opponent at the wrong time)

I moved that and the other bugfixes to #1193

d0sboots: This file can't import from outside this file, but the game can (and does) import types from here. In other words, this is the root of the types tree in many cases.

It was not immediately obvious that you weren't doing that here, especially when I was quickly checking on my phone XD

PR #1192, opened 2024-03-23
NETSCRIPT: Add Singularity.getFactionEnemies()

jaguardeer:
This PR adjusts Singularity.getFactionInviteRequirements() to indicate that the player should not join factions' enemies.
The requirement is generated as a not(some(enemies.map(joinedFaction(enemy)) inside getFactionInviteRequirements.
I generated it inside getFactionInviteRequirements so that other code that uses requirements, like the rumors display, is not affected.

I was unable to get the 'npm run doc' command to work.

Here is a test script to see the change:
```javascript
/** @param {NS} ns */
  const factions_without_enemies = ['The Black Hand', 'Church of the Machine God']
  const factions_with_enemies = ['Sector-12', 'Volhaven', 'Chongqing']
  const factions = ns.args.length > 0 ? ns.args : [...factions_with_enemies, ...factions_without_enemies];
  for (const faction of factions) {
    const reqs = ns.singularity.getFactionInviteRequirements(faction)
    ns.tprint(faction, "\n", JSON.stringify(reqs, null, 2));
  }
}
```


jaguardeer:
Added a `declare enum FactionName` entry to NetscriptDefinitions.d.ts. Docs now compile, but faction names will be visible to players in the documentation (I don't believe they were visible previously).
Alternatively I could make the joinedFaction a string instead of an enum, mirroring the structure of BackdoorRequirement.

d0sboots:
The core change looks good. However, I'm bringing in @Snarling because I'm not as certain on the situation with the Faction enums.

src/Faction/FactionJoinCondition.ts
```diff
@@ -111,6 +112,26 @@ export const notEmployedBy = (companyName: CompanyName): PlayerCondition => ({
...
+export const haveJoinedFaction = (factionName: FactionName): PlayerCondition => ({
+  toString(): string {
+    return `Joined ${factionName}`;
+  },
+  toJSON(): JoinedFactionRequirement {
+    return { type: "joinedFaction", joinedFaction: factionName };
+  },
+  isSatisfied(p: PlayerObject): boolean {
+    const factionsJoined = p.factions;
+    return factionsJoined.includes(factionName);
```
Snarling: I would check this using Factions[factionName].isMember instead, much more performant.
But with this being a synthetic requirement that isn't used internally, it doesn't actually matter because this isSatisfied function will never be called.

src/NetscriptFunctions/Singularity.ts
```diff
@@ -766,7 +767,11 @@ export function NetscriptSingularity(): InternalAPI<ISingularity> {
       helpers.checkSingularityAccess(ctx);
       const facName = getEnumHelper("FactionName").nsGetMember(ctx, _facName);
       const fac = Factions[facName];
-      return [...fac.getInfo().inviteReqs].map((condition) => condition.toJSON());
+      const enemyNotJoinedReqs =
```
Snarling: The fact that the API isn't actually mimicking the internal code, it's just pretending faction enemies work the same way as other requirements, is a red flag. Why not instead add a getFactionEnemies function that returns an array of enemies?

markdown/bitburner.factionname.md
```diff
@@ -0,0 +1,53 @@
...
+[Home](./index.md) &gt; [bitburner](./bitburner.md) &gt; [FactionName](./bitburner.factionname.md)
+
+## FactionName enum
+
+Names of all factions
+
+**Signature:**
+
+```typescript
+declare enum FactionName
```
Snarling: FactionName is one of the more spoilery enums, so providing it to the API is possibly something we should avoid. Personally I'm okay with providing it but I'd like to hear some other viewpoints on this.

src/NetscriptFunctions/Singularity.ts
Snarling: Another solution to the mismatch would be to add the actual requirements (probably inside the FactionInfo constructor) internally, but there are already many places where enemies are presented differently to all other requirements (e.g. they have their own section in the UI, the invitation modal calls them out separately, etc), and it would also mean basically getting rid of the isBanned flag for factions. Also we would not actually be able to remove the enemies property, since it's still needed for handling removing invites from enemies when joining a faction.

It's not always a problem to provide info differently to the API than it is handled internally. But the question is, why is there a mismatch?

Internally enemies are separate because they do things other requirements don't - joining a faction removes invites from enemy factions. Is this something a player might want to account for? I think so.

For instance imagine a player has already received an invite from faction X, but they have not yet accepted the invite due to e.g. trying to manipulate where faction rep from coding contracts gets distributed. Their script plans on joining faction X and any other lingering invites sometime later this run, but is just holding off for now. Since the invite has already been obtained, the player script should not be looking at invite requirements anymore (e.g. falling below a money requirement doesn't withdraw the invite), but the script should still have a way of knowing that in order to keep the invite for faction X, they need to make sure not to join their enemy faction Y. So I think to really provide players a correct API, enemies do need to be provided in a different manner (and for basically the same reason why they are handled separated internally).

markdown/bitburner.factionname.md
jaguardeer: I like the idea of adding getFactionEnemies(). What do you think would be a fair RAM cost for it? 3GB?

jaguardeer:
I have reworked this into Singularity.getFactionEnemies in order to match the internal code and in-game UI more closely.

markdown/bitburner.factionname.md
Snarling: Seems good, thanks!

PR #1194, opened 2024-03-26
IPVGO: Tweak cheat success scaling so it is applicable even to endgame players

ficocelliguy:
* Make the 14.3 source-file bonus stronger, but still require significant cheat success chance to get many 100% guaranteed cheat attempts.
* Make the previous-cheat debuff scale faster with the number of attempted cheats in the current game, to still be a factor for deep BN12 accounts


d0sboots:
I've been sitting on this one for a while because I wanted to come up with a better function. However, I haven't made any progress, so let's just go with this for now.

PR #1195, opened 2024-03-28
DOCUMENTATION: Fix wrong identification of inline mathematical expressions

catloversg:
Our in-game markdown viewer renders incorrectly in some cases when it encounters "potential" inline mathematical expressions (`$something$`):

Error:
error

Fixed:
fixed


d0sboots:
I'll take this as-is if you feel strongly, but IMO it would be better for consistency to either escape *all* dollar signs, or at least all of them in cases where there is more than one per line. That'll be more future-proof vs future additions.

catloversg:
I applied the fix to all relevant usages of the dollar sign.

PR #1196, opened 2024-03-28
DEVMENU: Remove SaveFileDev

catloversg:
This is the follow-up PR for #1162.

~~`SaveFileDev` has not been fully implemented. This PR fixes that.~~

Edit: After discussing with our maintainers, I decided to remove this "feature" instead of finishing its incomplete implementation.


src/DevMenu/ui/SaveFileDev.tsx
```diff
@@ -6,48 +6,87 @@ import AccordionDetails from "@mui/material/AccordionDetails";
...
+      throw new Error("Invalid import input");
+    }
+    importInput.current.click();
   }
 
-  function doRestore(): void {
-    const save = JSON.parse(saveFile);
-    // TODO unplanned: "Continue here". Unknown what this todo is for.
-    console.error(save);
+  async function doRestore(): Promise<void> {
```
d0sboots: Does this do anything different from regular import? Is there a reason to keep it in the dev menu?

catloversg: I only implement the needed changes to make this "feature" work.

To be honest, I have never used this "SaveFileDev" and I think its usage is incredibly niche:
- Import without reloading.
- Can exclude other things while importing (scripts, augmentations, SFs).

I'm fine if you want to remove this "SaveFileDev" feature. I use my own tools to save/load while testing.

Edit: For context, this feature was added in 8159dad5fa63ed11b4320d583015060a57fd819b, but the implementation has been incomplete until now. I think that nobody uses this feature, so nobody wants to finish the implementation. It may be fine if we just remove it.

catloversg: For context, this feature was added in 8159dad5fa63ed11b4320d583015060a57fd819b, but the implementation have been incomplete until now. I think that nobody uses this feature, so nobody want to finish the implementation. It may be fine if we just remove it.

d0sboots: I've also never used it when testing. @Snarling?

Edit: My thoughts are specifically that this seems like a maintenance burden for something no one is using.

Snarling: Yeah I've never used it. I agree we should probably just remove it

PR #1199, opened 2024-03-28
UI: Move typescript enum to const enum

### LINKED ISSUE #1198: Enums do not compile correctly to javascript.
The current way that enums are created they do fail when conpiling typescript to javascript giving no reference to @ns.js 

```ts
import { NS, CityName, GymType, LocationName } from "@ns";

export async function main(ns: NS): Promise<void> {
  ns.tprint(CityName.Sector12);
  ns.tprint(LocationName.Sector12PowerhouseGym);
  ns.tprint(GymType.agility);
}
```

this would fail with 

`Error while calculating ram usage for this script. No such file on server: "@ns.js"`

changing the enums to const enum does fix that.

Nepomuceno:
UI: Move typescript `enum` to `const enum`

closes #1198 

# Bug fix

- the code present in the issue now runs correctly.

### DIFF
--- a/src/ScriptEditor/NetscriptDefinitions.d.ts
+++ b/src/ScriptEditor/NetscriptDefinitions.d.ts
@@ -7562,7 +7562,7 @@ type CorpSmartSupplyOption = "leftovers" | "imports" | "none";
 
 /** Names of all cities
  * @public */
-declare enum CityName {
+declare const enum CityName {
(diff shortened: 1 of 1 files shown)


d0sboots:
This fails to build, and will always fail to build, because marking enums "const" removes the runtime aspect from them, making them unusable/useless for JS code.

PR #1204, opened 2024-04-01
UI: Increase margin bottom for BitNode Multipliers in Stats page (BN1)

catloversg:
In BN1, BitNode Multipliers's table is empty, so the bottom border is too near the text.

Before:
before

After:
after


DJ-Laser:
I would suggest also adding some text saying "none" instead of just having an empty multipliers section, or removing it altogether, in addition to this fix.

PR #1205, opened 2024-04-01
MISC: add neofetch command 

DJ-Laser:
This is very WIP but I wanted to submit a pr here so people can see what I'm working on 

Basically I want to add the neofetch command that displays system info like OS and system specs. There could also be Easter eggs from running this on specific servers, such as custom os names or ASCII art

If anyone has ideas for things to display for certain servers or wants to help make ASCII art for operating systems that would be greatly appreciated!

An example of the neofetch command can be found here: [Example](https://www.google.com/search?client=ms-android-google&sca_esv=924f3957077d70b1&sca_upv=1&sxsrf=ACQVn089FZzyVtK9OwlmJdY-_pa_Z3M-EQ:1712001186406&q=neofetch+command&uds=AMwkrPuBGWlH8-V0X-qFY2O9RvTEuCvh0ylbPZ0Wx5dPut72I0xo4D-3FoMWa4NkUWwXlXeIoR0_u1cwHm05X3rCWgctWGglpj4tAGrFeicM4iNB5tH1-MBs-iA86K3ZxaOmldO2TTaH5mwiBviMsj7dKZOnRvyXqOat_TDInVTh13hSSxOSj4WBdnqoUJBZku0Vb76EKw794QMVlPZuGR8SgqPP0DlxczvbXkiOKkopQ5pPOcHdTAjT2e1-0WWPefjnuWCGbtkOUQY9tVZrEpdUnOtXftMViTUZnIt7LXgWMw93Rw5S3_b6On2cjomj4ydpoeJEHYgC&udm=2&prmd=ivsnbmtz&sa=X&ved=2ahUKEwi8_8e05aGFAxU-3skDHUHCDHQQtKgLegQIDRAB&biw=412&bih=784&dpr=2.63&safe=active&ssui=on#vhid=z7M9_EjjBpu9NM&vssid=mosaic)

## Todos:
- [x] Add a terminal command named "neofetch"
- [x] Display a list of information
- [ ] Display some colors from the current theme
- [ ] Customize info based on the server, like installed packages, cpu, and os
- [ ] Display ascii art for each OS


src/Settings/Settings.ts
```diff
@@ -76,6 +76,8 @@ export const Settings = {
   TailRenderInterval: 1000,
   /** Theme colors. */
   theme: { ...defaultTheme },
+  /** Theme Name (needed for neofetch) */
+  themeName: "Default",
```
d0sboots: I don't think this is worth adding just for `neofetch`. Also, since you can have custom themes it's not even guaranteed that your theme will have a name.

d0sboots (commented):
Overall, seems like a fun idea. I think getting a good ascii-art representation of the BB logo (in the proper size) will be key to making this actually work.

Edit: Nitesec already has good ascii art for `avmnite-02h`. Same probably goes for some of the other factions.

src/Settings/Settings.ts
DJ-Laser: Good point. I can remove this as it will display some colors from the theme in the future anyway. I have been thinking about submitting a pr for the theme menu allowing you to save themes with names, so if I get to it I can implement this there.

Hoekstraa:
I'd recommend naming this something other than neofetch, as the names of these *fetch tools change all the time.
Neofetch, screenfetch, pfetch, and of course alternatives ufetch, fastfetch, machina, bfetch... 
Neofetch at this moment is already archived and being forked and rewritten like crazy: https://github.com/dylanaraps/neofetch.
Think of something fun, like burnerfetch, futurefetch, or whatever you can come up with.

d0sboots:
Yeah, I mentioned it on discord but on reflection this is probably better suited as a #project-showcases thing, instead of built-in to the game.

PR #1207, opened 2024-04-04
IPVGO: Bugfix Ensure full name of method is recorded in possibleLogs

ficocelliguy:
Previously the `getFunctionNames()` method in NetscriptFunctions.ts would drop the given prefix if it handled recursive object calls. This would cause certain methods to be listed as e.g. `cheat.destroyNode` instead of `go.cheat.destroyNode`. 

Because of this difference, ns.disableLog() did not work for these functions, and they would log anyway.

This PR fixes the issue to correctly capture the full nested name of functions. (This only affects go.cheat.XX and go.analysis.XX currently)

### DIFF
--- a/src/NetscriptFunctions.ts
+++ b/src/NetscriptFunctions.ts
@@ -1789,10 +1789,10 @@ function getFunctionNames(obj: object, prefix: string): string[] {
   for (const [key, value] of Object.entries(obj)) {
     if (key === "args") {
       continue;
-    } else if (typeof value == "function") {
+    } else if (typeof value === "function") {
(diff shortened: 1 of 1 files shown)


d0sboots:
Wouldn't it also affect stuff like `formulas.hacking.X`? Is it just that no one was trying to disable logs for those?

ficocelliguy:
You are right, technically this affects formulas.hacking.XXX ,  formulas.hacknetNodes.XXX, etc. but none of those methods appear to have any logs to disable. There may be other places where this bug is actually visible, but I don't see any lookign through netscript currently.

PR #1208, opened 2024-04-04
DOCS: Link to stable instead of dev for player-facing docs

ficocelliguy:
Avoid confusion about features that are not yet released or finalized by ensuring player-facing links go to stable branch instead of dev

### DIFF
--- a/src/Documentation/doc/advanced/bitnodes.md
+++ b/src/Documentation/doc/advanced/bitnodes.md
@@ -24,7 +24,7 @@ In this BitNode, most forms of income such as working at a [Company](../basic/co
 [Servers](../basic/servers.md) have less money on them and lowered growth rates, but it is easier to lower their security level using the `weaken` function.
 
 Furthermore, some BitNodes introduce new content and mechanics.
-For example, there is one BitNode that grants access to the [Singularity API](https://github.com/bitburner-official/bitburner-src/blob/dev/markdown/bitburner.singularity.md).
+For example, there is one BitNode that grants access to the [Singularity API](https://github.com/bitburner-official/bitburner-src/blob/stable/markdown/bitburner.singularity.md).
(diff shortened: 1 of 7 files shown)


src/Documentation/doc/advanced/bitnodes.md
```diff
@@ -24,7 +24,7 @@ In this BitNode, most forms of income such as working at a [Company](../basic/co
 [Servers](../basic/servers.md) have less money on them and lowered growth rates, but it is easier to lower their security level using the `weaken` function.
 
 Furthermore, some BitNodes introduce new content and mechanics.
-For example, there is one BitNode that grants access to the [Singularity API](https://github.com/bitburner-official/bitburner-src/blob/dev/markdown/bitburner.singularity.md).
+For example, there is one BitNode that grants access to the [Singularity API](https://github.com/bitburner-official/bitburner-src/blob/stable/markdown/bitburner.singularity.md).
```
d0sboots, line 27: Here and everywhere: do these work as relative links instead? That will ensure they always point correctly regardless of version, as well as working offline.

Snarling: The API markdown is not compiled and included in distribution like the docs in the documentation folder are. We briefly tried implementing this but it kind of sucked, mostly because of the formatting of the API markdown doesn't work well with the ingame doc viewer. I think to fix this we'd basically need to fork the API documenter to make our own custom formatting for the generated docs.

Right now I don't think relative links will work well with the ingame doc viewer unless those relative links stay within the documentation folder. But maybe we can make some changes to how that works, so that relative links that break out of documentation will form correct web links based on the version being played.

d0sboots: That's kinda sad, but I Guess this is the best we can do for now.

PR #1210, opened 2024-04-07
CORPORATION: Add a new API to sell a division

catloversg:
This PR adds 2 APIs:
- restartCorporation: `selfFund` does NOT have a default value.
- sellDivision.

Other changes:
- Change the import of `Player` from `import { Player as player }` to `import { Player }` in many places.
- Change the relative path of `Player`'s import to `@player` in many places.
- Remove `SellCorporationModal.tsx`. Its implementation is merged with `CreateCorporationModal.tsx`.


src/Corporation/ui/modals/CreateCorporationModal.tsx
```diff
@@ -30,55 +33,50 @@ export function CreateCorporationModal(props: IProps): React.ReactElement {
...
-            This <Money money={150e9} /> can either be self-funded, or you can obtain the seed money from the government
-            in exchange for {formatShares(500e6)} shares (a <b>33.3%</b> stake in the company).
+            Would you like to sell your position as CEO and start a new corporation? Everything from your current
+            corporation will be gone and you start fresh.
           </>
         )}
         <br />
         <br />
-        If you would like to start one, please enter a name for your corporation below:
+        If you would like to start {props.restart ? "new" : ""} one, please enter a name for your corporation below:
```
d0sboots:
```suggestion
        If you would like to start {props.restart ? "a new" : ""} one, please enter a name for your corporation below:
```

src/NetscriptFunctions/Corporation.ts
```diff
@@ -803,6 +786,20 @@ export function NetscriptCorporation(): InternalAPI<NSCorporation> {
         CorporationPromise.promise = new Promise<CorpStateName>((res) => (CorporationPromise.resolve = res));
       return CorporationPromise.promise;
     },
+    restartCorporation:
+      (ctx) =>
+      (_corporationName, _selfFund): boolean => {
+        checkAccess(ctx);
+        const corporationName = helpers.string(ctx, "corporationName", _corporationName);
+        const selfFund = !!_selfFund;
+        return createCorporation(corporationName, selfFund, true);
```
d0sboots: I think it's a bit of an API smell to both return true/false *and* throw exceptions for similar types of issues. (For instance, throw if the BN doesn't support corps, but return false if you didn't provide a name: the logical distinction there is arbitrary.)

I know some people don't like dealing with exceptions, but IMO when you call this method you *always* expect it to work, and there's enough different ways for it to fail that the descriptive message of an exception is helpful.

d0sboots (commented):
The refactoring here looks good. 

However, I'm slightly hesitant about sellCorporation. In the past, there were cheese stats for BN3 that revolved around repeatedly selling your corp; those might still exist, or might come back in the future. I'm also not sure what the method is good for *except* cheese strats.

sellDivision is similar, but probably OK.

catloversg:
After discussing with maintainers and other contributors, I decide to remove `restartCorporation` API.
The behavior of `createCorporation` in `Actions.ts` is not changed to avoid a breaking change in `createCorporation` API despite the different behaviors between throwing errors and returning true/false.

PR #1213, opened 2024-04-10
NETSCRIPT: A minorly breaking change around script launch, and refactoring.

d0sboots:
This changes it so that "this" will be unbound when `main()` is called, ~~and also passes `ns.args` as additional arguments to the `main` function.~~

Both of these are technically breaking changes, but in very minor ways: Someone could be calling exported functions/variables with `this.func()`, but given the simpler `func()` is available and works universally for exported *and* unexported functions, it seems unlikely that anyone would do this. It's also a very easy fix.

~~Similarly, adding new parameters to `main()` could break someone if they were calling `main()` recursively and depending on them to be empty, as they are now. Note that they will *still* be empty if the program is called with no arguments, so this is also a tough needle to hit, in exchange for better usability.~~ Adding new parameters to `main()` was reverted.

There is also some refactoring, to use `ScriptArg` consistently everywhere.

I checked that scripts still launch correctly, and args can now be accessed from the extra args. Also, `this` is now `undefined`.

### DIFF
--- a/markdown/bitburner.goopponent.md
+++ b/markdown/bitburner.goopponent.md
@@ -8,5 +8,12 @@
 **Signature:**
 
 ```typescript
-type GoOpponent = "Netburners" | "Slum Snakes" | "The Black Hand" | "Tetrads" | "Daedalus" | "Illuminati" | "????????????";
+type GoOpponent =
+  | "Netburners"
+  | "Slum Snakes"
+  | "The Black Hand"
+  | "Tetrads"
+  | "Daedalus"
+  | "Illuminati"
+  | "????????????";
 ```
--- a/markdown/bitburner.ns.args.md
+++ b/markdown/bitburner.ns.args.md
@@ -9,7 +9,7 @@ Arguments passed into the script.
 **Signature:**
 
 ```typescript
-readonly args: (string | number | boolean)[];
+readonly args: ScriptArg[];
 ```
 
 ## Remarks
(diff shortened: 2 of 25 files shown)


Snarling:
I'm not really sure what the benefit is of having args also get passed directly to main.

I can think of some scripts I would have written in the past that would be broken by this: one common pattern in golfed scripts is to define extra variables as default param values, to avoid using a separate let statement. This would no longer work when the script is run with args, overwriting the default params.

I'm not 100% against the change, but I'd need to see that it's a popular idea. Personally I don't see it really bringing any utility, but maybe I'm in the minority there.

Alpheus:
This will certainly mess with golfed scripts as they rely on parameter defaults to shorten variable. Sadly, it seems those scripts are most popular with players who don't write their own scripts. This might be a hard break.

d0sboots:
OK, you've convinced me. I did the args thing because someone was asking for it and it seemed like a nice usability thing without a real downside, but there are concrete usecases it would break.

I reverted that part and left the other change, plus refactoring.

Alpheus:
Before you give up—there _is_ a middle ground:
- creating a new function name to replace main (on demand) to denote this usage
- creating a second function in addition to main (similar to autocomplete)
- ~~or a bit wilder: check the Function.length parameter of each main script to check how to invoke it~~

If you take a Function object, its **length** property will show you how many defined arguments it has. Your use case is benign when it is exactly one.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/length


———Edit:
Actually, thinking about it—that doesn't work. It will make this change safe to merge but the feature will be turned off for the person who wants it indistinguishable from golfing.

Snarling (approved):
All looks good to me

Snarling:
@Alpheus 
> creating a new function name to replace main (on demand) to denote this usage
creating a second function in addition to main (similar to autocomplete)

I don't see either of these as viable solutions, but  also don't really see what the problem is. The game already provides script arguments to the script instance in a very usable way, and the player is free to place the args array into their own separate variable if they want.

PR #1217, opened 2024-04-17
MISC: Remove file-saver

catloversg:
This PR removes `file-saver` library. We don't need to support old (ancient) browsers or work around old bugs in them, so it's unnecessary.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -27,7 +27,6 @@
         "clsx": "^1.2.1",
         "date-fns": "^2.30.0",
         "escodegen": "^2.1.0",
-        "file-saver": "^2.0.5",
         "js-sha256": "^0.9.0",
         "jszip": "^3.10.1",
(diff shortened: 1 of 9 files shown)


d0sboots:
Before I commit, double checking: Save export, the download command and the recovery screen download still work, right?

(The PR template *does* have a section for testing XD)

catloversg:
I tested those 3 cases (export, download command, recovery screen download), and they work properly on FireFox and Chrome (Windows). I don't have a MacOS machine, so I have not tested on Safari. However, I asked a friend to test the following code on Safari:
```js
const file = new Blob(["123"]);
const a = document.createElement("a");
const url = URL.createObjectURL(file);
a.href = url;
a.download = "text.txt";
a.click();
setTimeout(function () {
  URL.revokeObjectURL(url);
}, 0);
```
They confirmed that it worked, so I assume that it's okay on Safari too.

PR #1218, opened 2024-04-18
MISC: Rename aug to violet Congruity Implant

hydroflame:
The original author of that aug has asked me to rename the aug.

I tried to make it so that if a player already had the augmentation installed or queued it would rename it on the spot. However it seems that the order of loading doesn't really allow for that.

In particular, these lines:
https://github.com/bitburner-official/bitburner-src/blob/dev/src/PersonObjects/Player/PlayerObject.ts#L165-L168

run before these lines:
https://github.com/bitburner-official/bitburner-src/blob/dev/src/SaveObject.ts#L712-L723

and thus it's not possible to rename an aug via save file migration.

I think this is ok for this particular aug because it's not likely installed all that often.

If you can think of any reasons why that wouldn't deploy nicely please let me know.

### DIFF
--- a/.git_blame_ignore_revs
+++ b/.git_blame_ignore_revs
@@ -78,19 +78,19 @@ a7045a234353de795494c0b3f378808b4a6d32aa
 # Date:   Sun Apr 17 17:23:14 2022 -0400
 
 567fcf8fb63bc469ae61013cca511a76eb6f9086
-# Author: nickofolas
+# Author: violet
 # Date:   Wed Apr 13 14:42:07 2022 -0500
 
 f6af85a38d8286aec702556f1296565de5a5067f
-# Author: nickofolas
+# Author: violet
 # Date:   Sun Apr 10 15:57:34 2022 -0500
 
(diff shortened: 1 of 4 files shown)


d0sboots:
The typical way we'd do a savefile migration would be matching on the old string anyway, which I suspect would be undesired in this instance. Agreed that a one-time loss of aug for something people rarely have is ok.

hydroflame:
It was agreed upon that the regexp `/.{10} Congruity Implant/` was a acceptable to use for the purpose of the migration. But I couldn't get the code to work for the reasons described above. I'm just unfamiliar if there's a process in place for renaming augs.

Rename or not is fine though I think.

d0sboots:
You could stuff that in the loading for the augments, right before they get filtered. Or move the filtering to the conversion code, so it has time to be acted on (but that would have to be checked for other side effects)

hydroflame:
I wouldn't want to leave code that would be relevant next update. So I think I'm ok with leaving it as-is, if you all are.

This aug is both rare and also the effect mostly apply as soon as you instal it. So losing it is less a big deal than any other aug.

Issue #795, opened 2023-09-09
"Time skip" in the development menu does not actually skip time


aschmider:
To reproduce this, just enable the development menu, select "Time skip" , and any of the offered choices. This saves and reloads the game, but the message at the start of the game says that just a few seconds have passed. This also means that generating hashes for testing purposes is, as far as I know, not really feasible.

Happens in v2.4.1 (https://github.com/bitburner-official/bitburner-src/commit/de5ebc993bb20634c3f601631017970c8924dc50), for example, but also dev (https://github.com/bitburner-official/bitburner-src/commit/f95ca64e016553d3352525fcab365e7f5523dc27).

![bug-skip-time](https://github.com/bitburner-official/bitburner-src/assets/26329811/84d655f8-62d7-4674-b0f9-ba6652af653a)

I've looked a bit into this and found that SaveObject.saveGame() gets called twice. The first time, Player.lastUpdate hat the decreased value which implements the time skip. When it gets called the second time, however, the value is the current time again. This call apparently comes from the auto save feature.

Since I'm not familiar with the code and the architecture, I don't know what the proper fix would be. I tried just removing the saving and reloading part in the time skip which worked fine in my case, i.e. the game catches up and calculates the hashes gained in the selected time as expected. However, I had a basically fresh game without any scripts running, and I expect performance to be quite poor this way.

For your reference, I have added a stack trace of the two calls to SaveObject.saveGame(). Also see this diff that removes the save+reload after selecting a time skip:
```
diff --git a/src/DevMenu/ui/TimeSkip.tsx b/src/DevMenu/ui/TimeSkip.tsx
index 37a7c8a36..fdafdbb94 100644
--- a/src/DevMenu/ui/TimeSkip.tsx
+++ b/src/DevMenu/ui/TimeSkip.tsx
@@ -18,8 +18,6 @@ export function TimeSkip(): React.ReactElement {
     return () => {
       Player.lastUpdate -= time;
       Engine._lastUpdate -= time;
-      saveObject.saveGame();
-      setTimeout(() => location.reload(), 1000);
     };
   }
```

[localhost-1694268863124.log](https://github.com/bitburner-official/bitburner-src/files/12566009/localhost-1694268863124.log)

Caldwell-74:
could this be rooted in a hacknet server bug?
i used timeskip a few times to accelerate grafting to test out the grafting bug and it worked
and "Time played: Since last Augmentation installation" also correctly adds the skipped time

[...]

PR #1219, opened 2024-04-18
BUGFIX: Issue 795

cjhier:
Fixes Issue: https://github.com/bitburner-official/bitburner-src/issues/795


d0sboots:
This doesn't seem right to me. Every time a new TimeSkip element is created, a new event listener will be added, and they're never cleaned up. In addition, I don't see how adding *extra* autosaves fixes the problem in the bug.

d0sboots:
This issue ended up being subtle, but it's being handled. See #1221 and #1223.

PR #1221, opened 2024-04-19
DEVMENU: Fix bug in Time skip menu

### LINKED ISSUE #795: "Time skip" in the development menu does not actually skip time

catloversg:
This PR fixes a bug in the Time skip menu. The root cause is the autosave feature. The flow:
- `Player.lastUpdate` and `Engine._lastUpdate` are decreased in `TimeSkip.tsx`. [Step 1]
- Engine loop in `Engine.start` runs. It calculates `diff`, **set `Engine._lastUpdate` and `Engine._lastUpdate` to current time** (minus a small offset), then call `Engine.updateGame(diff)`. [Step 2]
- `Engine.updateGame` calls `Engine.decrementAllCounters`.
- `Engine.decrementAllCounters` decreases `Engine.Counters.autoSaveCounter`.
- `Engine.updateGame` calls `Engine.checkCounters`.
- `Engine.checkCounters` checks `Engine.Counters.autoSaveCounter`. `Engine.Counters.autoSaveCounter` is very likely negative at this point, especially if we skip 1 hour/day. Autosave happens, and it saves `Player.lastUpdate` of step 2 (current time), not step 1.

The fix is simple. We only need to counter the decrement of `Engine.Counters.autoSaveCounter`.

Fixes #795.

### DIFF
--- a/src/DevMenu/ui/TimeSkip.tsx
+++ b/src/DevMenu/ui/TimeSkip.tsx
@@ -18,6 +18,9 @@ export function TimeSkip(): React.ReactElement {
     return () => {
       Player.lastUpdate -= time;
       Engine._lastUpdate -= time;
+      // (time / CONSTANTS.MilliPerCycle) is enough to counter the decrement of Engine.Counters.autoSaveCounter, but we
+      // use (time) to ensure that autosave does not happen in the delay time (1 second) before we reload.
+      Engine.Counters.autoSaveCounter += time;
       saveObject.saveGame();
       setTimeout(() => location.reload(), 1000);
     };


d0sboots:
Is autoSaveCounter part of the saved state? This strikes me as rather a hack, although this is a dev menu feature so that might be OK. But if it has the possibility of permanently delaying autosave, I'm not a fan...

The best fix IMO is to fix the interactions in step 2. I feel like it has the possibility of causing *actual* bugs, such as missing offline updates, not just for this feature.

d0sboots:
OK, thinking about it some more I don't think the issue is actually in step 2 at all.

I think the problem is that we are letting `Engine.updateGame` run at all. Basically, there are two possible ways time skip could work, and currently we are doing neither of them:
* We could skip time in an offline fashion. This would involve adjusting `lastUpdate` and *immediately* saving the game. This should then show offline progress when the page reloads.
* We could skip time in an online fashion. This would involve adjusting `lastUpdate` and *never reloading the page*. The normal engine updates will handle applying the time jump.

Right now, we're doing the 2nd but pretending it's like the first.

d0sboots:
OK, I've figured out the *actual* issue with #795. It's a real bug in the code, which causes a small amount of hash loss every time you load, but it's only really noticeable in this usecase.

d0sboots:
See #1223. I think all that remains here is to sort out the reload/no-reload issue.

catloversg:
Let's deal with it in #1223.

PR #1222, opened 2024-04-19
IPVGO: Update opponent in UI when game changes from API

ficocelliguy:
When a new game is started via the API, ensure that the correct opponent label is shown on the gameboard UI

### DIFF
--- a/src/Go/ui/GoGameboardWrapper.tsx
+++ b/src/Go/ui/GoGameboardWrapper.tsx
@@ -34,12 +34,13 @@ interface GoGameboardWrapperProps {
  */
 
 export function GoGameboardWrapper({ showInstructions }: GoGameboardWrapperProps): React.ReactElement {
+  const boardState = Go.currentGame;
   const rerender = useRerender();
   useEffect(() => {
+    setOpponent(boardState.ai);
     return GoEvents.subscribe(rerender);
-  }, [rerender]);
+  }, [boardState.ai, rerender]);
 
-  const boardState = Go.currentGame;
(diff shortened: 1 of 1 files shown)


d0sboots:
Why is opponent even part of the react state?

ficocelliguy:
@d0sboots I've experimented a bit with grabbing the game state object directly (Go.currentGame.ai) instead, but React caches things that aren't reactive/part of state (since it knows they will never need to be changed), meaning certain things never get updated.

For example, the modal to find a new subnet / new game is supposed to default to your current opponent. However, if I remove the opponent from the React state and use the game state object, it snapshots whoever the opponent was when the subnet page was loaded an always uses that, even if it was changed via script later.

(This is an effect of the change to not hold go game state in react state, and instead manually tell react when to rerender with GoEvents.)

Snarling:
I don't see any issue with removing opponent as a state variable.

> For example, the modal to find a new subnet / new game is supposed to default to your current opponent.

Any change in opponent will already trigger a rerender of the GoGameboardWrapper. The only place this can change through the UI is via the resetState function, which already manually calls rerender. Any change of opponent from the API will trigger an event-based rerender. So just getting an up-to-date reference for the opponent during each render is enough to keep it up to date.

I see opponent is sent to the GoScoreModal as a prop. If during a new render cycle of GoGameboardWrapper, the opponent variable changes, it will send an updated prop to GoScoreModal regardless of whether opponent is stored in react state or stored in a plain variable (or just accessed as boardState.ai).

PR #1223, opened 2024-04-19
BUGFIX: (Partial) fix for #795

d0sboots:
Hacknet production wasn't getting saved on the last cycle, because of bad placement in Engine.updateGame. This was especially noticable when using timeskip, but it was a general bug.

This is only a partial resolution of #795 because there's still an issue with how it misleadingly shows no offline production.

Testing: Checked that hacknet production now applies when using dev-menu timeskip.

### DIFF
--- a/src/engine.tsx
+++ b/src/engine.tsx
@@ -114,15 +114,17 @@ const Engine: {
     // Sleeves
     Player.sleeves.forEach((sleeve) => sleeve.process(numCycles));
 
-    // Counters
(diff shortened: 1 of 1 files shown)


catloversg:
With this fix, we can replace the reloading code in `TimeSkip.tsx` with a message box that tells players the "time skip" effect has been applied. Without reloading, we don't have to worry about the confusion of "no offline production".

PR #1230, opened 2024-04-24
EDITOR: Add "arguments" to the keyword list

catloversg:
This change was suggested by @Caldwell-74 on Discord.
`arguments` is not a keyword in the default javascript support of `monaco-editor` (https://github.com/microsoft/monaco-editor/blob/main/src/basic-languages/javascript/javascript.ts). This PR adds it to our custom keyword list.

### DIFF
--- a/src/ScriptEditor/ScriptEditor.ts
+++ b/src/ScriptEditor/ScriptEditor.ts
@@ -45,7 +45,7 @@ export class ScriptEditor {
       l.language.tokenizer.root.unshift([new RegExp("\\bns\\b"), { token: "ns" }]);
       for (const symbol of apiKeys)
         l.language.tokenizer.root.unshift([new RegExp(`\\b${symbol}\\b`), { token: "netscriptfunction" }]);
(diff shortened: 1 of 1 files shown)


Caldwell-74:
just for the reasoning why

> 
> [Making eval and arguments simpler](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode#making_eval_and_arguments_simpler)
> Strict mode makes [arguments](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/arguments) and [eval](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/eval) less bizarrely magical. Both involve a considerable amount of magical behavior in sloppy mode: eval to add or remove bindings and to change binding values, and arguments syncing named arguments with its indexed properties. **Strict mode makes great strides toward treating eval and arguments as keywords**.

and the scripts as js modules run in strict mode
coloring arguments as keyword might prevent that player try to use it as variable name and run into errors that way

PR #1231, opened 2024-04-24
BUGFIX: Fix #795

### LINKED ISSUE #795: "Time skip" in the development menu does not actually skip time

catloversg:
This is the follow-up PR for #1223.
@d0sboots's PR fixes the root cause of #795, but there was a concern about the misleading popup after we reload. "Save+reload" is unnecessary now because the effect will be applied in the next engine loop. This PR replaces the "Save+reload" code with a message dialog.

Fixes #795.

### DIFF
--- a/src/DevMenu/ui/TimeSkip.tsx
+++ b/src/DevMenu/ui/TimeSkip.tsx
@@ -8,8 +8,8 @@ import ExpandMoreIcon from "@mui/icons-material/ExpandMore";
 import Typography from "@mui/material/Typography";
 import Button from "@mui/material/Button";
 import { Player } from "@player";
-import { saveObject } from "../../SaveObject";
 import { Engine } from "../../engine";
+import { dialogBoxCreate } from "../../ui/React/DialogBox";
 
 // Update as additional BitNodes get implemented
 
@@ -18,8 +18,7 @@ export function TimeSkip(): React.ReactElement {
     return () => {
       Player.lastUpdate -= time;
       Engine._lastUpdate -= time;
-      saveObject.saveGame();
-      setTimeout(() => location.reload(), 1000);
+      dialogBoxCreate("Time skip effect has been applied");
     };
   }
 


Snarling:
I think the only difference here is how time skipping affects scripts.

By just changing the last update time with no reload, most game mechanics should accumulate bonus time and function identically to if we were actually offline for X amount of time, but scripts will be unaffected since they don't rely on a last update time at all. Whereas with the current reload approach, scripts should function as if they were offline for X amount of time as well.

Still, there's not really a right or wrong way for this to work, and I don't really like the reload thing, so I'm down with merging this change. Alternately we could make both options available, maybe keeping a timeskip+reload option is helpful to test offline impact of scripts.

catloversg:
If we want to implement the (reload+offline for X), we need to change time skip code a little. In the next engine loop, the "backward" last update time will be overwritten with the current timestamp. If the auto save (which may occur due to the "backward" last update time) and the reload action happen after that, the offline time (after reloading) will be nearly 0. It's a bit troublesome to avoid this problem. For example, we can `await saveObject.saveGame()`, stop the engine loop, and reload immediately (this is just an example).

I can implement a timeskip+reload option if there is demand for it; otherwise, I will keep this PR simple with current approach.

d0sboots:
We might want something to make testing of offline time easier, but right now this doesn't do it (despite seeming like it does). There's a usecase for both, but IMO it's best to have this working the simple way (dealing with reloads is also annoying as a user) and then we can assess if we need an offline timeskip version.

d0sboots:
> For example, we can `await saveObject.saveGame()`, stop the engine loop, and reload immediately (this is just an example).

I don't think you even would need to stop the engine loop, since the reload should be synchronous?

catloversg:
The engine loop **may** still run after the call of `location.reload()`. There is a very small window between `location.reload()` and the moment that the JS engine actually stops due to the page reload. In most cases, it does not matter, because the "delay" between each loop is fairly big (`CONSTANTS.MilliPerCycle - offset`) and it's very unlikely that our engine code can process everything in that small window.

This code show that small window:
```html
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="utf-8" />
    <title>test-tools</title>
  </head>

  <body>
    <script>
      console.log("loaded");
      const random = Math.random();
      const engineLoop = () => {
        console.log(random, Date.now());
        window.setTimeout(engineLoop, 200);
      };
      console.log("start engine loop");
      engineLoop();
      setTimeout(() => {
        console.log("reload");
        location.reload();
      }, 2000);
    </script>
  </body>
</html>
```

Sample log:
```
0.7383995384024831 1714041332123
0.7383995384024831 1714041332324
0.7383995384024831 1714041332525
0.7383995384024831 1714041332726
0.7383995384024831 1714041332927
0.7383995384024831 1714041333128
0.7383995384024831 1714041333330
0.7383995384024831 1714041333531
0.7383995384024831 1714041333734
reload
0.7383995384024831 1714041333938
loaded
start engine loop
0.7883704209928808 1714041333950
Navigated to file:///<Redacted>/test-engine-loop.html
0.7883704209928808 1714041334154
```

With the "delay" of 200, this rarely happens.
If we reduce the "delay" to 1-10, it happens more frequently:
```
0.538676335374972 1714041463883
0.538676335374972 1714041463888
0.538676335374972 1714041463893
0.538676335374972 1714041463898
0.538676335374972 1714041463902
0.538676335374972 1714041463907
0.538676335374972 1714041463912
0.538676335374972 1714041463917
reload
0.538676335374972 1714041463923
0.538676335374972 1714041463928
loaded
start engine loop
0.6298148591624857 1714041463934
0.6298148591624857 1714041463935
0.6298148591624857 1714041463937
0.6298148591624857 1714041463939
0.6298148591624857 1714041463941
0.6298148591624857 1714041463946
0.6298148591624857 1714041463951
Navigated to file:///<Redacted>/test-engine-loop.html
0.6298148591624857 1714041463956
```

catloversg:
The engine loop **may** still run after the call of `location.reload()`. There is a very small window between `location.reload()` and the moment that the JS engine actually stops due to the page reload. In most cases, it does not matter, because the "delay" between each loop is fairly big (`CONSTANTS.MilliPerCycle - offset`) and it's very unlikely that our engine code can process everything in that small window.

This code shows that small window:
```html
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="utf-8" />
    <title>test-tools</title>
  </head>

  <body>
    <script>
      console.log("loaded");
      const random = Math.random();
      const engineLoop = () => {
        console.log(random, Date.now());
        window.setTimeout(engineLoop, 200);
      };
      console.log("start engine loop");
      engineLoop();
      setTimeout(() => {
        console.log("reload");
        location.reload();
      }, 2000);
    </script>
  </body>
</html>
```

Sample log:
```
0.7383995384024831 1714041332123
0.7383995384024831 1714041332324
0.7383995384024831 1714041332525
0.7383995384024831 1714041332726
0.7383995384024831 1714041332927
0.7383995384024831 1714041333128
0.7383995384024831 1714041333330
0.7383995384024831 1714041333531
0.7383995384024831 1714041333734
reload
0.7383995384024831 1714041333938
loaded
start engine loop
0.7883704209928808 1714041333950
Navigated to file:///<Redacted>/test-engine-loop.html
0.7883704209928808 1714041334154
```

With the "delay" of 200, this rarely happens.
If we reduce the "delay" to 1-10, it happens more frequently:
```
0.538676335374972 1714041463883
0.538676335374972 1714041463888
0.538676335374972 1714041463893
0.538676335374972 1714041463898
0.538676335374972 1714041463902
0.538676335374972 1714041463907
0.538676335374972 1714041463912
0.538676335374972 1714041463917
reload
0.538676335374972 1714041463923
0.538676335374972 1714041463928
loaded
start engine loop
0.6298148591624857 1714041463934
0.6298148591624857 1714041463935
0.6298148591624857 1714041463937
0.6298148591624857 1714041463939
0.6298148591624857 1714041463941
0.6298148591624857 1714041463946
0.6298148591624857 1714041463951
Navigated to file:///<Redacted>/test-engine-loop.html
0.6298148591624857 1714041463956
```

PR #1233, opened 2024-04-26
BUGFIX: Fix wrong corruptable text when switching between different companies

### LINKED ISSUE #998: Location names in the new Job Info page mess up with each other when their servers are backdoored
### Step to reproduce
1. Apply job for two locations with different length of name (say Ecorp and Fulcrum Technologies)
2. Backdoor their server.
3. Check the Job page and switch to Fulcrum Technologies
4. Use Previous Job or Next Job to switch to Ecorp

### Result
The shuffled string for Fulcrum Technologies still lingers after switch to Ecorp
image

### Version
Bitburner v2.5.2 (c6141f2ad)

catloversg:
This PR fixes #998.

Before:

https://github.com/bitburner-official/bitburner-src/assets/152669316/392870bd-132c-4b40-8d0f-da75a65192a1

After:

https://github.com/bitburner-official/bitburner-src/assets/152669316/9a6c320d-e450-4b47-b6dd-73fbeb2b4fed

There are comments about the wrong job requirement. It's caused by a different bug. I'll fix it in another PR.

### DIFF
--- a/src/ui/React/CorruptableText.tsx
+++ b/src/ui/React/CorruptableText.tsx
@@ -34,6 +34,8 @@ export function CorruptableText(props: CorruptableTextProps): JSX.Element {
       return;
     }
 
+    setContent(props.content);
+
     let counter = 5;
     const timers: number[] = [];
     const intervalId = setInterval(() => {


d0sboots:
It took me a bit to even understand what was going on here. The structure of CorruptableText is weird.

PR #1236, opened 2024-04-27
MISC: Rework reputation bonus after installing a backdoor

### LINKED ISSUE #1067: OmniTek Reputation
I'm running Bitburner v2.5.2, and was invited to the faction OmniTek Incorporated, which has a requirement of having 400k reputation with OmniTek. When I got the invite, I had just passed 300k reputation with them. Mousing over OmniTek Incorporated in the Factions menu shows that the 400k reputation requirement is checked off, even with my actual reputation too low. The job listing "Head of Software" at OmniTek itself, which also requires 400k reputation, did not show the 400k requirement as checked off at the time, but did some minutes later, despite reputation still being too low.

catloversg:
If players install a backdoor on a company's server, they always have a reputation bonus (100k) when we check the required reputation for joining that company's faction or applying for that company's jobs. There are at least 3 problems:
- Players can apply for a job that has a required reputation of less than 100k when they have 0 reputation.
- The UI of the Job tab does not take account of this reputation bonus.
- `ns.singularity.getFactionInviteRequirements` and `ns.singularity.getCompanyPositionInfo` do not take account of this reputation bonus.

Snarling suggests that we can change this flat reduction to a percentage.

This PR fixes those problems.

Fixes #1067.

### DIFF
--- a/src/Server/ServerHelpers.ts
+++ b/src/Server/ServerHelpers.ts
@@ -10,6 +10,7 @@ import { Person as IPerson } from "@nsdefs";
 import { Server as IServer } from "@nsdefs";
 import { workerScripts } from "../Netscript/WorkerScripts";
 import { killWorkerScriptByPid } from "../Netscript/killWorkerScript";
+import { serverMetadata } from "./data/servers";
 
 /**
  * Constructs a new server, while also ensuring that the new server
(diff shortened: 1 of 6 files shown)


src/Server/ServerHelpers.ts
```diff
@@ -236,6 +237,15 @@ export function isBackdoorInstalled(server: BaseServer): boolean {
   return false;
 }
 
+export function isBackdoorInstalledInCompanyServer(companyName: string): boolean {
```
d0sboots, line 240: Unless you have an immediate compelling use elsewhere, I'd keep this alongside the only thing that's using it (in `Company/utils.ts`), to avoid adding to our cross-helpers dependency web. In fact, I'd probably just merge the two functions together.

d0sboots: Nevermind, since this uses `isBackdoorInstalled` there's no cutting this particular knot. I must have been extra sleepy last night.

PR #1237, opened 2024-04-28
CORPORATION: Update tooltip of storage space

catloversg:
The tooltip of the storage space only shows sizes of materials/products. This is confusing for newbies. They use "Unit" (number of material/product units) when buying materials, but that tooltip only shows sizes without any description.

Before:
before

After:
after

Edit: After discussing it with other contributors, this is how it looks in the end:
after

### DIFF
--- a/src/Corporation/ui/DivisionWarehouse.tsx
+++ b/src/Corporation/ui/DivisionWarehouse.tsx
@@ -11,7 +11,7 @@ import { ProductElem } from "./ProductElem";
 import { MaterialElem } from "./MaterialElem";
 import { MaterialInfo } from "../MaterialInfo";
 import { createProgressBarText } from "../../utils/helpers/createProgressBarText";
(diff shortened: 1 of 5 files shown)


d0sboots:
Hmm. I still think there is a lot (maybe too much) room for confusion here. For instance, "size" could be viewed as the per-unit size - that was my initial assumption when looking at these screenshots.

I think I would change "Size" to either "Total Size" or "Total Space" (There should be enough room since these are aligned on the right), and "Unit" to "Units".

Other than wording, LGTM

Caldwell-74:
agree there with d0s not sure how a third row "size per unit" would look if it bloats the ui or
not but maybe that would be a nice addition that is otherwise only accessable via api

d0sboots:
If we went with three columns (I'm also not sure if that would be too big or not), IMO it should be in the order "Size", "Units", "Total Space".

PR #1240, opened 2024-05-03
DARKWEB: Update DarkWeb.tsx to buy all possible programs using buy -a

### LINKED ISSUE #1239: buy -a command should buy all possible, rather than only buy everything or nothing.
When a user runs "buy -a", they presumably want to buy as much as they can.
At the moment, if you cannot afford Formulas.exe, (and haven't already got it) you get nothing.

I believe that iterating through and making all the possible purchases before reporting if some were too expensive or no more applications are available for purchase would provide an improved user experience.

The available programs are conveniently sorted in general order of value per dollar, even if the details are arguable.
If the buy all purchased each executable in order, users would first get all of the port-opener scripts they can afford (which is probably what they want in the first place to progress), and then some of the utility apps as funds allow.

TheAimMan:
Current implementation of buy -a checks to see if the player can afford every remaining program and then buy all of them.  If they can not, then the function fails.

Changed the command to buy every remaining program in the same order until they can no longer afford them, instead of failing outright.

image

closes https://github.com/bitburner-official/bitburner-src/issues/1239

### DIFF
--- a/src/DarkWeb/DarkWeb.tsx
+++ b/src/DarkWeb/DarkWeb.tsx
@@ -1,6 +1,6 @@
 import React from "react";
 import { DarkWebItems } from "./DarkWebItems";
-
+import { formatMoney } from "../ui/formatNumber";
 import { Player } from "@player";
 import { Terminal } from "../Terminal";
 import { SpecialServers } from "../Server/data/SpecialServers";
@@ -91,7 +91,7 @@ export function buyAllDarkwebItems(): void {
     if (!Player.hasProgram(item.program)) {
       itemsToBuy.push(item);
       if (item.price > Player.money) {
-        Terminal.error("Not enough money to purchase remaining programs");
+        Terminal.error("Need " + formatMoney(item.price - Player.money) + " more to purchase " + item.program);
         return;
       } else {
         buyDarkwebItem(item.program);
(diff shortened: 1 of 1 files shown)


src/DarkWeb/DarkWeb.tsx
```diff
@@ -86,13 +85,17 @@ export function buyDarkwebItem(itemName: string): void {
...
   const itemsToBuy: DarkWebItem[] = [];
-  let cost = 0;
 
   for (const key of Object.keys(DarkWebItems) as (keyof typeof DarkWebItems)[]) {
     const item = DarkWebItems[key];
     if (!Player.hasProgram(item.program)) {
       itemsToBuy.push(item);
-      cost += item.price;
+      if (item.price > Player.money) {
+        Terminal.error("Not enough money to purchase remaining programs");
```
d0sboots:
```suggestion
        Terminal.error("Not enough money to purchase remaining programs, " + formatMoney(cost) + "required for " + item.program);
```
Lint might want to re-wrap that.

TheAimMan: how about telling them how much they actually need and not just the total cost of the program?

image

d0sboots: That works too, I just wanted there to be feedback on what you couldn't buy.

TheAimMan: cool, code submitted

PR #1241, opened 2024-05-03
MISC: Handle error when getting save data

### LINKED ISSUE #799: Invalid string length at JSON.stringify
we had it a couple times that players where unable to save and or the engine crashed (so no progress ingame)
when the savegame became to big through files or running scripts with long names
just want to have an issue open if anyone finds a good way to improve the situation 
even if saving isnt possible the game should notify the player about the problem (not just in the dev console)
and the engine should finish updating

catloversg:
In #799, we discussed situations in which the game engine crashes when it cannot save the game data. When the save data is too large, `JSON.stringify` throws a `RangeError`. Before #1162, `getSaveData` (`src\SaveObject.ts`) is a sync method, so if it throws an exception, the game engine will crash (due to the autosave code when `Engine.updateGame` runs). After #1162, `getSaveData` is an async method, so the exception will be caught by code in `setupUncaughtPromiseHandler`. This mitigates the problem, but it's not ideal:
- The default error message is confusing.
- When players try to import another save file, the "Compare Save" UI is a black screen, because `fetchData` (`src\ui\React\ImportSave\ImportSave.tsx`) does not handle the exception.

This PR changes these things:
- Behave consistently when an exception is thrown in `getSaveData`.
- Show a dialog box with a proper error message.
- Prevent the black screen of "Compare Save" UI.

How to generate save data that throws a `RangeError` when calling `getSaveData`:
Run this code:
```js
    ns.write(`test${i}.txt`, "0".repeat(10e7));
  }
}
```

Test cases:
- "Save game" button.
- "Export Game" button.
- Autosave.
- Import Game -> Compare Save.

Closes #799.

### DIFF
--- a/src/Netscript/ErrorMessages.ts
+++ b/src/Netscript/ErrorMessages.ts
@@ -105,3 +105,14 @@ export function handleUnknownError(e: unknown, ws: WorkerScript | ScriptDeath | 
   }
   dialogBoxCreate(initialText + e);
 }
+
+/** Use this handler to handle the error when we call getSaveData function */
+export function handleGetSaveDataError(error: unknown) {
+  console.error(error);
+  const baseErrorMessage = "Cannot get save data";
+  if (error instanceof RangeError) {
+    dialogBoxCreate(`${baseErrorMessage}. Error: ${error}. This may be because the save data is too large.`);
+  } else {
+    dialogBoxCreate(`${baseErrorMessage}. Error: ${error}.`);
+  }
+}
--- a/src/Netscript/ErrorMessages.ts
+++ b/src/Netscript/ErrorMessages.ts
@@ -109,10 +109,12 @@ export function handleUnknownError(e: unknown, ws: WorkerScript | ScriptDeath | 
 /** Use this handler to handle the error when we call getSaveData function */
 export function handleGetSaveDataError(error: unknown) {
   console.error(error);
-  const baseErrorMessage = "Cannot get save data";
+  let errorMessage = `Cannot get save data. Error: ${error}.`;
   if (error instanceof RangeError) {
-    dialogBoxCreate(`${baseErrorMessage}. Error: ${error}. This may be because the save data is too large.`);
-  } else {
-    dialogBoxCreate(`${baseErrorMessage}. Error: ${error}.`);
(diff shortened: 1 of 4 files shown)


src/Netscript/ErrorMessages.ts
```diff
@@ -105,3 +105,14 @@ export function handleUnknownError(e: unknown, ws: WorkerScript | ScriptDeath |
...
+
+/** Use this handler to handle the error when we call getSaveData function */
+export function handleGetSaveDataError(error: unknown) {
+  console.error(error);
+  const baseErrorMessage = "Cannot get save data";
+  if (error instanceof RangeError) {
+    dialogBoxCreate(`${baseErrorMessage}. Error: ${error}. This may be because the save data is too large.`);
+  } else {
+    dialogBoxCreate(`${baseErrorMessage}. Error: ${error}.`);
+  }
```
d0sboots:
```suggestion
  let errorMessage = `Cannot get save data. Error: ${error}.`;
  if (error instanceof RangeError) {
    errorMessage += " This may be because the save data is too large.");
  }
  if (error.stack) {
    errorMessage += `\nstack:\n${e.stack}`;
  }
  dialogBoxCreate(errorMessage);
```
I didn't test that at all (especially since different browsers treat stack differently), but hopefully you can see the intent XD

src/ui/React/ImportSave/ImportSave.tsx
```diff
@@ -124,14 +125,29 @@ export const ImportSave = (props: { saveData: SaveData; automatic: boolean }): J
   useEffect(() => {
     async function fetchData(): Promise<void> {
       const dataBeingImported = await saveObject.getImportDataFromSaveData(props.saveData);
-      const dataCurrentlyInGame = await saveObject.getImportDataFromSaveData(await saveObject.getSaveData(true));
+      let saveData;
+      try {
+        saveData = await saveObject.getSaveData(true);
+      } catch (error) {
+        return Promise.reject(error);
```
d0sboots: I think this is equivalent to throwing `error`? In which case, I don't see the point in adding try/catch here at all.

d0sboots:
Tricky stuff here, but it looks good, aside from some optional-ish nitpicks.

Issue #826, opened 2023-09-23
Hamming code leading zeroes are confusing


BanovMiroslav:
There are two coding contracts relating to Hamming codes. One of them gives example of how integers are encoded to hamming codes and asks you to encode one, and the other one gives a hamming code with possible flipped bit and asks for the integer.

The issue is that the first one does not expect leading zeroes, and the second one does contain leading zeroes in the hamming code. This is confusing and actually caused me to wonder if the problem was with the endianness instead, which is also mentioned in both of the coding contracts. This was provided for the contract "HammingCodes: Encoded Binary to Integer" - 0100100010111000010001101000001111101011001001111010110100000001

Here is how decoding it and re-encoding without leading zeroes gives different results:

```
{
  hammingInput: '0100100010111000010001101000001111101011001001111010110100000001',
  binaryFromHamming: '000001110001000110100000111101011001001111010110100000001',
  integerFromBinary: 3979415699041537,
  binaryFromInteger: '1110001000110100000111101011001001111010110100000001',
  hammingFromBinary: '00010110000100010101000001111010011001001111010110100000001'
}
```

So I am not sure, maybe zero padding is correct. But I think there shouldn't be two implementations, and only one should be used. Or maybe make a mention in the contract that the value may have zero padding to make it a square matrix or something like that.

Version of game: v2.4.1 (8/26/23)

d0sboots:
*Technically*, the 2nd form (the one asked for in Integer to Encoded Binary) isn't a true Hamming Code, because they operate with a block size of 2^n - 1. (2^n for extended Hamming Codes, which is what the problem is dealing with.) Rather, it's a truncated Hamming Code.

Probably it would make more sense to round up the block size to a power of two, and then the two would match up. However, this would break all the existing contract solvers out there...

gmcew:
Enhancement label for script-breaking consistency change - terminology in UI should be fixed by above PRs

PR #1244, opened 2024-05-05
MISC: Hamming Code Contract description clarification

gmcew:
Rewording of the Hamming code contract wording based on suggestion at [discord](https://discord.com/channels/415207508303544321/415213435974975508/1233491440013672448)
There's further scope on other contracts to clean up inconsistencies in example formatting.

Linked to #826 in that the endianness description is gone, but leading 0 structural issue within the contract is unchanged.


d0sboots:
Sorry for being slow here, I'm working out of vacation backlog.

Although I agree with the goal of making the description stand alone (and I think you've done a good job with it), at the same time Hamming Codes are still complicated and difficult to explain without diagrams. So I think linking to both https://en.wikipedia.org/wiki/Hamming_code and the video previously linked as *supplementary* material (as opposed to "you must watch this to understand") is a good idea.

d0sboots:
Also, one other thing your explanation isn't covering is the confusing bit-ordering that is going on in this problem.
I.e. it is conventional to number bits from left-to-right starting at bit 0 (or 1), and counting up. And that is how it is done in this problem. However, this naturally corresponds to wanting bit 0 to be 2^0, bit 1 = 2^1, bit 2 = 2^2 etc., i.e. a little-endian representation. But this is not how humans conventionally write numbers, at least in English.

As a result, this problem adopts a weird compromise where the bits for the hamming encoding are numbered left-to-right, but the bits for the *binary* encoding of the data are numbered right-to-left.

gmcew:
No apology necessary- thank you for the time and advice. I've incorporated your suggestions - reference to wikipedia and the original youtube video is added/restored, as well as a note on endianness whilst trying to keep it succint. 

EDIT: Didn't sync this branch, sorry

src/data/codingcontracttypes.ts
```diff
@@ -1297,19 +1297,21 @@ export const codingContractTypesMetadata: ICodingContractTypeMetadata[] = [
...
+        "You are given the following decimal value: \n",
+        `${n} \n\n`,
+        "Convert it to a binary representation and encode it as an 'extended Hamming code'.\n ",
+        "The number should be converted to a string of '0' and '1' with no leading zeroes.\n",
+        "Parity bits are inserted at positions 0 and 2^N.\n",
+        "Parity bits are used to make the total number of '1' bits in a given set of data even.\n",
+        "The parity bit at position 0 considers all bits including parity bits.\n",
+        "Each parity bit at position 2^N alternately considers N bits then ignores N bits, starting at position 2^N.\n",
+        "The endianness of the parity bits is reversed compared to the endianness of the data bits:\n",
+        "Data bits are encoded left-to-right (most significant byte first) and the parity bits encoded right-to-left.\n",
```
d0sboots:
```suggestion
        "Data bits are encoded most significant bit first and the parity bits encoded least significant bit first.\n",
```
I feel like it's slightly wrong to say the parity bits are encoded right-to-left, or if not *wrong* exactly then it doesn't get to the heart of the matter. (Same comment for the decoding side.)

src/data/codingcontracttypes.ts
```diff
@@ -1297,19 +1297,21 @@ export const codingContractTypesMetadata: ICodingContractTypeMetadata[] = [
...
+        "The parity bit at position 0 is set last.\n\n",
+        "Examples:\n",
+        "8 in binary is 1000, and encodes to 11110000 (pppdpddd - where p is a parity bit and d is a data bit)\n",
+        "21 in binary is 10101, and encodes to 1001101011 (pppdpdddpd)\n\n",
+        "For more information on the 'rule' of encoding, refer to the Wikipedia entry and the 3Blue1Brown videos on Hamming Codes.",
```
d0sboots:
```suggestion
        "For more information on the 'rule' of encoding, refer to Wikipedia (https://wikipedia.org/wiki/Hamming_code)",
        "or the 3Blue1Brown videos on Hamming Codes. (https://youtube.com/watch?v=X8jsijhllIA)",
```
Adding the links explicitly to make it easy for people (same for the decoding side)

PR #1246, opened 2024-05-06
MISC: Embed JetBrainsMono font

### LINKED ISSUE #17: Embed a font
[theit8514]
unknown

MacOS uses by default a Mac-specific font. These fonts are missing the Unicode characters for minimize and restore. Potential solution is to change the unicode symbols used, or embed a font with those symbols.

catloversg:
Edit: I choose JetBrainsMono font. It looks good. It's also small, so we don't have to worry about the file size.
Capture

The description below is outdated.
---
Lucida Console is our default font, but it's only included by default on Windows. In #17, Snarling suggests embedding a default font, so I embed Lucida Console in this PR.

`LucidaConsole.woff2` was generated by:
- `lucon.ttf`: This is the Lucida Console ttf file on my Windows machine.
- https://github.com/google/woff2: I use this tool to convert ttf to woff2.

Fixes #17.

Tested on Windows and Electron app.

Screenshot:
Capture


Snarling:
Unforuntately, it looks like this would be against the EULA for Lucida Fonts: https://lucidafonts.com/pages/eula

If we're going to embed a font, it needs to be one with an open license.

mctylr-gh:
We could include [Cascadia Code](https://github.com/microsoft/cascadia-code), an open-source (SIL Open Font License) typeface from Microsoft, that is used with Terminal and Visual Studio. I think it may be used in VS Code, but I'm not certain.

Or we could include the pre-patched Nerd Font variant called [CaskaydiaCove Nerd Font](https://github.com/ryanoasis/nerd-fonts/tree/master/patched-fonts/CascadiaCode/Regular). The patching includes many popular icon fonts sets, like awesome font, devicons, weather, seti UI, octicons, font logos, codicons and material design icons.

Snarling:
Last time this was brought up, including a nerd font was also suggested. The filesize is a good bit larger though due to all the embedded extras.

I'm open to either option (nerd font or not) as well as changing up the typeface to something that looks different. Although maybe something safe like Cascadia would be best since it's just a default, and the player can change it themselves to something more stylized if they want.

If we're going to make it a nerdfont (or just any font with a large file size), it would be great if there was a way to only have that font downloaded by the browser if it's actually selected as the game font (i.e. if the player is using a different custom font, don't bother downloading the embedded font). Not sure how feasible that is though.

mctylr-gh:
One thing to reduce the embedded file size is you only need to include the Regular & Bold weights, and if they have a separate Italics file, then that it and Bold-Italics. That would be around 10MB for _CaskaydiaCove Nerd Font Mono_ in those four weights in OpenType format. 

CascadiaCode woff2 file sizes:

```
   986856  2024-04-25 21:44   woff2/CascadiaCodeNF.woff2
   929412  2024-04-25 21:43   woff2/CascadiaCodeNFItalic.woff2

   159692  2024-04-25 21:43   woff2/CascadiaCodeItalic.woff2
   219264  2024-04-25 21:43   woff2/CascadiaCode.woff2
```


[JetBrains Mono](https://github.com/JetBrains/JetBrainsMono) is another professionally designed open-source (SIL OFL) option.

d0sboots:
I like Jetbrains Mono. I'm a bit hesitant about using a Nerd Font, because those stuff all their icons into the private-use area. That's fine if you're a developer - the way you use the icons, you will be able to easily adapt if the mapping changes. But if we expose them through a font in BB, people *will* start using them by codepoint (because otherwise, why did we even include them in the font), which will essentially lock us in to keeping compatibility with that *particular* mapping of icons to PUA codepoints.

Basically, it would be like the same issue Unicode has with standardizing new codepoints: there are no take-backs.

Snarling:
Totally down with this, it is nice and small file size and players can install and use their own nerdfonts if that's something they want.

2 comments:

* We might need to include their [license](https://github.com/JetBrains/JetBrainsMono/blob/master/OFL.txt) if we're going to distribute the font
* If we're embedding a font, there's no need for any of the default settings to have a bunch of fallback options included. The first option (the embedded font) will always be used since it'll always exist.

d0sboots:
It *might* be good to still keep the fallbacks in though, in case either (a) something messes up, or (b) someone is doing something funky. Just playing devil's advocate.

Edit: Also, IANAL but it definitely looks like we'll need to include the license. (Somewhere where it makes it to the final distribution output, which I'm pretty sure the root `dist` directory does?)

catloversg:
New changes:
- Add the license file to `dist` folder. This makes "Validate Title and Check for Dist Changes / Check Files and Title" fail.
- Fallback options for font: I only use Courier New and monospace. "Courier New" is included in Windows and MacOS. "monospace" can handle all other cases.

d0sboots:
> Add the license file to dist folder. This makes "Validate Title and Check for Dist Changes / Check Files and Title" fail.

Hmm, that's probably there for a good reason, maybe Snarling has a better idea of where the license should be put.

Or maybe it's something we just override in this case.

Snarling:
In most cases adding something directly to dist is not appropriate, instead we should add it to src and have a bundler take care of providing the distributed version (similar to how the .woff2 file for the font itself is handled).

But I'm not sure how difficult that would be to set up for a license .txt file that isn't actually used in the code at all. It's not a big issue so I think in this case we can just ignore the test failure.

PR #1248, opened 2024-05-06
API: Bladeburner API break for getCurrentAction when Idle

Snarling:
2.6.1 includes a large refactor of lots of internal Bladeburner code. Among the changes, when the player is not performing an action, their current action is stored as `null` instead of an action identifier for an "idle" task.

This also affected the API in some ways. Previously, the API was providing a fake value `{type: "Idle", name: ""}` for `ns.bladeburner.getCurrentAction()` when the player is not performing an action. Then some, but not all, of the other API functions that take in actions had hardcoded returns for when the idle action was passed in. The fake object return was preserved in the refactor, but the hardcoded behavior of the other functions was not preserved. Because the existing behavior was not helpful and was inconsistent across different functions that take in an action, **instead of reimplementing the previous behavior for all affected functions, the fake action object return is now being removed**.

| `ns.bladeburner` Function | Previous Behavior | New Behavior |
| --: | :-- | :-- |
| `getCurrentAction` | returned `{type: "idle", name: ""}` when idle | returns `null` when idle |
| `getActionCountRemaining` | returned -1 for idle action | can no longer be called with idle |
| `getActionCurrentLevel` | errored for idle action | can no longer be called with idle |
| `getActionEstimatedSuccessChance` | returned `[-1, -1]` for idle action | can no longer be called with idle |
| `getActionMaxLevel` | errored for idle action | can no longer be called with idle |
| `getActionRepGain` | errored for idle action | can no longer be called with idle |
| `getActionSuccesses` | errored for idle action | can no longer be called with idle |
| `getActionTime` | returned -1 for idle action | can no longer be called with idle |

The other action functions all expect strings for type and name, so the new representation of `null` does not provide a way to call these functions for an idle action (since there is no type or name property).

This change now requires players to explicitly handle the null possibility for the return of getCurrentAction before attempting to access its `type` and `name` properties (including through a destructured assignment).


Snarling:
Added some brief instructions for how to fix scripts after the API break at #1249. Before merging this I plan to add an API break notice ingame if usage of getCurrentAction is detected in one of the player's scripts during a numerical version upgrade from <39 (2.6.1) to >=39. Will also provide some basic API break notification utils.

Snarling:
Added a utility for alerting the player to potential impact from API breaks.

When loading a game:
If there are no files on any server where a possible API break is detected, nothing is shown.
Otherwise, details about the API break are saved to a file:
image
A terminal message is shown to alert the player:
image
And a series of dialog boxes are created to display the changes to the player:
image
image
image

Snarling:
@d0sboots (and anyone else who sees this) I'm going to go ahead and merge this, but I'd like some review on the API break utilities included here / formatting of API break data in the 2.6.1 file. Once we can get the formatting finalized, we can start working on formalizing planned changes for 3.0 via the same format.

catloversg:
The new format and utilities look good to me.

Is there any plan to support broken properties (from `setDeprecatedProperties`) with these new utilities? When we remove deprecated properties (after notifying players for at least 1 version), we may use `showAPIBreaks` to show breaking changes to those properties.

d0sboots:
Agreed with Cat, removed properties affect players pretty much identically so it it would be nice to unify them with this system.

Otherwise, formatting looks good!

PR #1250, opened 2024-05-07
UI: Fix clipped numbers when values are too big in Augmentations page

### LINKED ISSUE #614: Augmentations page loses some far right values
When have big numbers some values are not visible/clipped from the right side.

2023-06-15 14_22_16-Bitburner v2 3 1 (100030008) Bug
2023-06-15 14_27_01-Bitburner v2 3 1 (100030008)_Bug2

catloversg:
Before:
before

After:
after

Fixes #614.

### DIFF
--- a/src/Augmentation/ui/PlayerMultipliers.tsx
+++ b/src/Augmentation/ui/PlayerMultipliers.tsx
@@ -24,10 +24,7 @@ interface IBitNodeModifiedStatsProps {
 }
 
 function customFormatPercent(value: number): string {
-  if (value >= 100) {
-    return formatPercent(value, 2, 100);
-  }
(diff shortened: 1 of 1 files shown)


src/Augmentation/ui/PlayerMultipliers.tsx
```diff
@@ -23,14 +23,22 @@ interface IBitNodeModifiedStatsProps {
   color: string;
 }
 
+function customFormatPercent(value: number): string {
+  if (value >= 100) {
```
d0sboots: I don't get the purpose of this `if`. It's duplicating a check that happens inside `formatPercent`; shouldn't `formatPercent(value, 2, 100)` work the same unconditionally?

PR #1254, opened 2024-05-07
IPVGO: Handle the previous AI move even if it was made by a different process

ficocelliguy:
In some cases, there can be multiple different processes interacting with the IPvGO gameboard between player script(s) and the subnet UI. In those cases, if there is a failure to apply an AI move to the board, check if the prior move was made by white, and return its details, instead of logging an error

### DIFF
--- a/src/Go/effects/netscriptGoImplementation.ts
+++ b/src/Go/effects/netscriptGoImplementation.ts
@@ -178,28 +178,12 @@ export async function getAIMove(boardState: BoardState): Promise<Play> {
 
     await sleep(400);
     const aiUpdatedBoard = makeMove(boardState, result.x, result.y, GoColor.white);
-
-    if (!aiUpdatedBoard) {
-      // If the previous move was made by the AI, log it, even if it didn't come from this thread.
-      // This can happen if multiple scripts are run, or if the user opens or interacts with the gameboard UI while a script is running
-      const priorMove = getPreviousMove();
-      if (Go.currentGame.previousPlayer === GoColor.white && priorMove) {
-        return resolve({
-          type: GoPlayType.move,
-          x: priorMove[0],
-          y: priorMove[1],
-        });
-      }
-
-      // If the previous move wasn't made by the AI, handle the AI trying an invalid move. This shouldn't ever happen.
-      boardState.previousPlayer = GoColor.white;
-      console.error(`Invalid AI move attempted: ${result.x}, ${result.y}. This should not happen.`);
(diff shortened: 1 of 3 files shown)


src/Go/effects/netscriptGoImplementation.ts
```diff
@@ -179,8 +179,19 @@ export async function getAIMove(boardState: BoardState): Promise<Play> {
     await sleep(400);
     const aiUpdatedBoard = makeMove(boardState, result.x, result.y, GoColor.white);
 
-    // Handle the AI breaking. This shouldn't ever happen.
     if (!aiUpdatedBoard) {
+      // If the previous move was made by the AI, log it, even if it didn't come from this thread.
```
d0sboots: I don't understand what you mean by "this thread." JS is inherently single-threaded.

src/Go/effects/netscriptGoImplementation.ts
```diff
@@ -179,8 +179,19 @@ export async function getAIMove(boardState: BoardState): Promise<Play> {
...
+      // This can happen if multiple scripts are run, or if the user opens or interacts with the gameboard UI while a script is running
```
d0sboots: Maybe I'm being slow because it's the morning, but I don't fully understand what "this" is in this context, or why two *player* moves would cause the *ai* move to break.

d0sboots (commented):
I suspect the fix is fine, but I don't understand the issue yet.

src/Go/effects/netscriptGoImplementation.ts
ficocelliguy: There is code in place to make sure the AI takes its move when it is its turn, even if the game was saved and loaded where the AI was in mid-thought. This can be triggered in some cases where the player script is actively playing, but the gameboard AI beats it to the punch and makes a move first. (Since the gameboard AI is not part of the player script or netscript , it's like a separate process).

Solution: If the netscript AI tries to make a move but something beat it to it, return white's last move instead of an error.

ficocelliguy:
This has been updated to prevent the race condition from the board UI events side, and simplify the error handling.

d0sboots:
I think we're going to go with Snarling's PR at this point, which will probably conflict hugely with this one. But if this has additional stuff you want, you can rebase it after. Probably best to hold off on more stuff here for the moment, though.

Snarling:
@ficocelliguy I think everything here was addressed a different way in #1260, but we can reopen this if there's something important I'm missing.

PR #1255, opened 2024-05-07
IPVGO: Provide API for getting game stats per opponent

ficocelliguy:
Give users access to wins, losses, stat bonuses, and favor gained


src/Go/effects/netscriptGoImplementation.ts
```diff
@@ -362,6 +364,29 @@ export function resetBoardState(
   return simpleBoardFromBoard(Go.currentGame.board);
 }
 
+/**
+ * Retrieve and clean up stats for each opponent played against
+ */
+export function getStats() {
+  return getRecordKeys(Go.stats).reduce((statDetails, opponent) => {
```
d0sboots:
```suggestion
  const statDetails: Partial<Record<GoOpponent>> = {};
  for (const opponenent of getRecordKeys(Go.stats)) {
```

I might have shanked the syntax (writing from my phone), but IMO using for-of here is simpler/easier to read, and it doesn't need an as-cast for the type.

PR #1256, opened 2024-05-07
UI: Hacknet terminology consistency

### LINKED ISSUE #403: UI: Change name/description of Hacknet-related stats (also Netburner Augmentations)
I like to suggest to change the names/description of Hacknet-related player stats and the Netburner Augmentations.

It's always "Hacknet Nodes" (e.g. "Hacknet Node Production" in Player stats tab). 

The description for BN9 itself states that "Hacknet Mutliplier" are increased.
Maybe "Node" can be removed from the stats and augments.

gmcew:
This is a sweeping change of hacknet terminology to be more consistent per #403.
All UI references should now always refer to "Hacknet Production", "Hacknet Node Purchase Cost", "Hacknet Level Upgrade Cost","Hacknet RAM Upgrade Cost" and "Hacknet Core Upgrade Cost" consistently.
The exception is the multiplier keys are shortened to "hacknet_production", "hacknet_purchase_cost", "hacknet_level_cost", "hacknet_ram_cost", "hacknet_core_cost", making this an API change too.

Things to note, and I'd appreciate feedback on:
- Is the proposed naming approach suitable? The logic is that game refers to the purchased device as a hacknet node, with the upgraded version being called the hacknet server. Hence, all hacknet servers are still nodes when referring to the "physical item" in terms of number owned/purchasing another one. This limits the need to rename any of the ns.hacknet functions and reduces the amount of script-breaking as a result.
- As far as I can tell cache upgrade cost isn't affected by the other cost reductions, is this intentional, and is there any appetite to change this? e.g. Stanek doesn't appear to change cache cost multiplier, nor other augments that reduce "hacknet costs".
- This will break scripts by changing the hacknet-associated keys. Notably ns.getPlayer will break, and ns.getBitNodeMultipliers.
- This required modification of the FullSave snapshot to pass the PlayerSave continuity check for the same reason, which feels like cheating. From playtesting on a local build my pre-commit saves loaded fine.

Closes #403


gmcew:
I think I hunted down all the references that seem worthy of changing so I'm taking this out of draft - let me know thoughts on how to proceed.

d0sboots:
A few high-level comments:
* The generally used terms that I see on Discord are "Hacknet Node" and "Hacknet Server" to refer to two separate things. Sometimes "Hashnet" is used as a shorthand for "Hacknet Server." Also, sometimes just *"Hacknet"* is used as a shorthand for "Hacknet Server," because the Nodes are rather useless and so no one really talks about them or uses them. Never/almost never have I seen "Hacknet Node" used as an umbrella term to refer to all of hacknet.
* Script breaks are to be avoided. Save file format changes are the same way. Both can be done for a good enough reason (and usually synced with a major version change), but I'm not sure this qualifies.
* Putting the two above together: I think improving the docs, and possibly some of the internal variable names could be useful, but probably the player-API-visible names should be left alone. In particular, names like `hacknet_ram_cost` read a little weird to me because ram cost *doesn't* apply to all of hacknet, it only applies to servers. (Obviously the same issue applies to our current names too, but I wouldn't want to introduce a breaking change that isn't improving that.)

gmcew:
I think I agree with your conclusions, and I'll summarise what I understand the direction should be:
- UI labels/Augment effects reworded to remove the "Node" parts
- No changes to API-facing labels/savedata/functions

I'll raise it as an issue separately to keep things clean - interestingly level, RAM and core upgrades all apply to both Nodes and Servers. The only Server specific upgrade is cache level, which doesn't get any cost reduction applied from stanek or augment effects despite what the descriptions imply.

d0sboots:
Current state of this looks good to me, but I'm gonna want @Snarling for a 2nd opinion as well.

Snarling:
Looks good to me, @gmcew did you still want to add more changes? This is still marked as draft status

gmcew:
Sorry for the radio silence @d0sboots and @Snarling , this is all ready if you're happy

PR #1260, opened 2024-05-10
GO: Alternate fix for race conditions

Snarling:
@d0sboots @ficocelliguy I believe this stops any possibility for there to be multiple AI turns pending at the same time, but let me know what you think and if there's some other potential race condition that doesn't seem to be addressed.

Because this PR includes a location change + function name change + changes to the function, it makes more sense to look at the individual race condition 1 and 2 commits for reviewing those specific things, instead of the overall changes for the PR which are hard to see the changes in.

The move and rename commits had no changes other than location and name of the function.

6d01457: Moved getAIMove to a different location since it's not netscript specific. No substantive changes, just a location change.

48bfb2a: Close a weird 300ms gap where it's actually the player's turn because the AI move was already applied to the board, but the Go.nextTurn promise for the previous AI turn has not yet resolved. There was an extra sleep that was causing this weird mismatched in-between time. I slightly increased one of the existing sleeps to compensate removing the extra sleep.

afeb6e7: More directly deals with the possibility of multiple concurrent AI turns due to being triggered from different sources. Now AI turns are disallowed from being triggered until the previous one has resolved. If multiple are triggered simultaneously then they will both return the same turn, instead of being two separate instance of the AI attempting to take a turn.

708ba27: Renamed getAITMove to makeAIMove since it actually triggers the AI to make the move. No substantive changes, just a name change.

3eb4dc7: Removed the functionality from the UI where loading Go will take the AI's turn if a game was loaded where it was already the AI's turn. Instead this is handled by the game automatically on game load. Also un-stated the waitingOnAI variable on the gameboardwrapper, because it's a variable handled by Go internal game state and not React state.

### DIFF
--- a/src/Go/SaveLoad.ts
+++ b/src/Go/SaveLoad.ts
@@ -9,6 +9,7 @@ import { assertLoadingType } from "../utils/TypeAssertion";
 import { getEnumHelper } from "../utils/EnumHelper";
 import { boardSizes } from "./Constants";
 import { isInteger, isNumber } from "../types";
+import { makeAIMove } from "./boardAnalysis/goAI";
 
 type PreviousGameSaveData = { ai: GoOpponent; board: SimpleBoard; previousPlayer: GoColor | null } | null;
 type CurrentGameSaveData = PreviousGameSaveData & {
@@ -77,6 +78,10 @@ export function loadGo(data: unknown): boolean {
   Go.currentGame = currentGame;
   Go.previousGame = previousGame;
   Go.stats = stats;
+
+  // If it's the AI's turn, initiate their turn
+  if (currentGame.previousPlayer === GoColor.black && currentGame.ai !== GoOpponent.none) makeAIMove(currentGame);
+
   return true;
 }
 
--- a/src/Go/SaveLoad.ts
+++ b/src/Go/SaveLoad.ts
@@ -2,9 +2,9 @@ import type { BoardState, OpponentStats, SimpleBoard } from "./Types";
 import type { PartialRecord } from "../Types/Record";
 
 import { Truthy } from "lodash";
-import { GoColor, GoOpponent } from "@enums";
+import { GoColor, GoOpponent, GoPlayType } from "@enums";
 import { Go } from "./Go";
-import { boardStateFromSimpleBoard, simpleBoardFromBoard } from "./boardAnalysis/boardAnalysis";
+import { boardStateFromSimpleBoard, getPreviousMove, simpleBoardFromBoard } from "./boardAnalysis/boardAnalysis";
 import { assertLoadingType } from "../utils/TypeAssertion";
 import { getEnumHelper } from "../utils/EnumHelper";
 import { boardSizes } from "./Constants";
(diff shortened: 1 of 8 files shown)


src/Go/boardAnalysis/goAI.ts
```diff
@@ -21,13 +21,23 @@ import { findAnyMatchedPatterns } from "./patternMatching";
...
+  // If an AI move is already in progress, return the already in progress move
+  if (isAIBusy && Go.nextTurn) return Go.nextTurn;
+  // Disallow new AI moves to be triggered during this one
+  isAIBusy = true;
+  let resolve: (play: Play) => void;
   Go.nextTurn = new Promise<Play>((res) => {
-    resolve = res;
+    resolve = (play) => {
+      // Once the promise resolves, stop blocking new AI moves
+      isAIBusy = false;
```
d0sboots: Subtle point: I think you should set this in a `finally` chained off of `getMove`. I don't *think* it makes a difference now, but it would if any of these promises rejected.

d0sboots:
Great, this is precisely the kind of race-condition avoidance I was thinking of! This will also make it easier to make a change I have vaguely kicking around in my head, which is to move Go AI processing to a Web Worker.

src/Go/boardAnalysis/goAI.ts
Snarling: Good idea, implemented in a471108 / 24cd914

Snarling:
Notable further changes:

* Go.nextTurn is no longer possibly null. It was being initialized as soon as it was accessed anyway and the only null case was on initial game load, so now it is just initialized on load (15ff42a)
* There was a minor bug where if the AI was trying to pass, and the player started a new game before the AI finished determining the next move, the pass would be applied to the current game. That was addressed in b0bbfe7

There is a bug currently where passTurn is being called twice but I'm not sure the best way to fix it. Right now it's being called during getMove (which is not supposed to be actually changing the state of the game), and then called again during makeAIMove. I'd like to remove it from getMove. That'll probably be the last thing included in this PR once I figure out a good way to do that.

Snarling:
Alright, I think I'm done messing around with this. @d0sboots @ficocelliguy FYI in case you want to look over / give feedback on the changes before I merge them

src/Go/SaveLoad.ts
```diff
@@ -77,6 +78,18 @@ export function loadGo(data: unknown): boolean {
   Go.currentGame = currentGame;
   Go.previousGame = previousGame;
   Go.stats = stats;
+
+  // If it's the AI's turn, initiate their turn, which will populate nextTurn
+  if (currentGame.previousPlayer === GoColor.black && currentGame.ai !== GoOpponent.none) makeAIMove(currentGame);
```
d0sboots, line 83: IMO all-on-one-line style for if statements is a bit shady (it's lead to various high-profile bugs in other applications), but especially if you have an else block you should always use curly braces for the if block.

d0sboots (approved):
Nice, "Remove async wrapper for main makeAIMove" and "Unwrap Go.nextTurn" were things I was looking at too, but didn't comment on because I didn't want to expand this PR unduly. :D

PR #1265, opened 2024-05-10
UI: Bladeburner console to use Settings.TimestampsFormat for logging

### LINKED ISSUE #719: UI timestamp format inconsistency - Bladeburner console
Within the game options is the Interface tab allowing custom Timestamp formats (src/Settings/Settings.ts). This does not apply to the Bladeburner console, which only uses a fixed format (src/utils/helpers/getTimestamp.ts).

Present in v2.4.0 (fc86895c3)

BBscreenshot1
BBscreenshot2

gmcew:
This PR changes Bladeburner console to use the `Settings.TimestampsFormat` value for logging, where available. If `Settings.TimestampsFormat` is an empty string it defaults to ISO8601 format. An invalid format string in the settings box results in "format error", same as script timestamps.

This also makes [src/utils/helpers/getTimestamp.ts](src/utils/helpers/getTimestamp.ts) redundant, but I'm reluctant to delete it just yet.

closes #719


d0sboots:
If you know it's redundant, please kill it now :D

PR #1268, opened 2024-05-11
CORPORATION: Refactor bribery

catloversg:
This PR makes these changes:
- Refactor these code:
  - Check valuation.
  - Calculate reputation gain by bribing.
  - Action of bribing.
- Remove the useless restriction `Player.hasGangWith(factionName)`. When the corporation is strong enough to bribe, the gang is useless. This problem was discussed on Discord.
- UI:
  - Use a different color (red in the default theme) when the player set an invalid amount of funds for bribing.
  - Fix a bug: `BribeFactionModal` does not show the last value that the player set in the number input when the modal is reopened.

### DIFF
--- a/src/Corporation/Actions.ts
+++ b/src/Corporation/Actions.ts
@@ -22,8 +22,6 @@ import {
   buybackSharesFailureReason,
   issueNewSharesFailureReason,
   costOfCreatingCorporation,
-  reputationGainByBribing,
-  isValuationHighEnoughToBribe,
 } from "./helpers";
 import { PositiveInteger } from "../types";
 import { currentNodeMults } from "../BitNode/BitNodeMultipliers";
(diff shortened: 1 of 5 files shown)


src/Corporation/Actions.ts
```diff
@@ -628,3 +631,37 @@ export function SetProductMarketTA1(product: Product, on: boolean): void {
 export function SetProductMarketTA2(product: Product, on: boolean): void {
   product.marketTa2 = on;
 }
+
+export function Bribe(
```
d0sboots: Functions should be lowerCamelCase. I get that this will conflict with `bribe` as a function used elsewhere, but IMO that is a feature, because having two functions distinguished only by case is also a bit of a smell.

I don't have enough context yet to suggest how these should be named differently, but clearly they *should*, since one calls the other. Edit: I might call the other function `uiBribe` or something along those lines, since its responsibilities lie with pulling and verifying inputs from the UI components. Maybe you can think of better.

src/Corporation/helpers.ts
```diff
@@ -88,3 +88,11 @@ export function issueNewSharesFailureReason(corp: Corporation, numShares: number
 
   return "";
 }
+
+export function isValuationHighEnoughToBribe(valuation: number) {
```
d0sboots: These two functions are so simple that I'm not sure they need to/should be made functions.

I.e. if I saw `valuation >= corpConstants.bribeThreshold` in the code that would be more clear to me than `isValuationHighEnoughToBribe(valuation)`, and it's roughly the same length.

src/Corporation/Actions.ts
```diff
@@ -628,3 +631,37 @@ export function SetProductMarketTA1(product: Product, on: boolean): void {
...
+  corporation: Corporation,
+  fundsForBribing: number,
+  factionName: FactionName,
+): {
+  success: boolean;
```
d0sboots: Does this need to return an object? Is it not sufficient to assume that 0 rep-gain = failure?

d0sboots (commented):
I like the overall direction, but I have some nitpicks.

PR #1269, opened 2024-05-12
MISC: Modify the source code TS file mode to be a regular file

mctylr-gh:
Correct the file mode of Typescript source file from being 755 to 644

`src/Prestige.ts` was mode 755 (`-rwxr-xr-x`) rather than the regular / normal file mode 644 (`-rw-r--r--`)


d0sboots:
Once the tests pass and this merges, you can have the contributor flair if you want it. Edit: No change too small! :D

mctylr-gh:
Flair? Six pieces, sir.

PR #1270, opened 2024-05-12
IPVGO: Remove unneeded functions from boardState.ts

d0sboots:
There were some functions that duplicate builtins (`floor`, `ceil`) and a pair of conditions that can be collapsed into one, for better efficiency.

Tested that IPvGO still seems to work

@ficocelliguy if this looks good to you, I'll merge on my own authority.

### DIFF
--- a/src/Go/boardAnalysis/boardAnalysis.ts
+++ b/src/Go/boardAnalysis/boardAnalysis.ts
@@ -9,8 +9,7 @@ import {
   getBoardCopy,
   getEmptySpaces,
   getNewBoardState,
-  isDefined,
-  isNotNull,
+  isNotNullish,
   updateCaptures,
   updateChains,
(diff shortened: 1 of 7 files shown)


ficocelliguy:
I had implemented floor/ceil locally so it can't be edited by players to do pseudo-rng-manipulation, but if you don't think that's important I'm fine with this.

d0sboots:
Ah yes, prototype attacks are in the class of techniques that are "allowed exploits," in the sense that we don't try to protect against them. Such techniques can be used to break the rng in the casino, or set your stats to arbitrary values, and that's on you - you could also edit your save to achieve the same thing. 

One of the fun things about BB is that it forces you to define the line of what *you* consider cheating, much more so than other games which set the line themselves. XD

PR #1271, opened 2024-05-12
API: Add FileMetadata like timeOfModification to reconcile file differences when using Remote File API

Hoekstraa:
It's been a while!

Time to make an addition to Remote File API (RFA). This one is to help RFA servers (so the things running on people's computers) decide when to push or pull files from/to the game. It's supposed to solve the problem of the following scenario:

- I've previously used RFA, but don't have it enabled right now.
- I make some changes in Bitburner.
- I connect RFA.
- As RFA server, which file should I see as leading? The local one or the one in-game?


Hoekstraa:
After some testing, this is the metadata behaviour on a Unix system:

**Create**
A - now
M - now
C - now
B - now

**Read contents ("off disk")**
A - now
M - x
C - x
B - x

**Edit**
A - x
M - now
C - now
B - x

**Copy into new file (don't forget that this is a read for the original file)**
A - now
M - now
C - now
B - now

**Move into new file**
A - x
M - x
C - now
B - x

**Copy into existing file**
A - take original file A
M - now
C - now
B - take original file B

**Move into existing file**
A - take copied file A
M - take copied file M
C - now
B - take copied file B

d0sboots:
> After some testing, this is the metadata behaviour on a Unix system:

This is correct, but won't be universally true. You can see the details here (https://man7.org/linux/man-pages/man7/inode.7.html), for instance. The major thing is that `btime` was not historically supported on most linux filesystems, (and it's not specified at all in POSIX), but that's changed relatively recently due to influence from other OSes like Windows.

...
> **Edit** A - x M - now C - now B - x

This shouldn't change `ctime` by what I understand, but from my testing it does. Strange.

src/Paths/FileMetadata.ts
```diff
@@ -0,0 +1,16 @@
+export class FileMetadata {
+  constructor(readonly timeOfAccess: number, readonly timeOfModification: number, readonly timeOfBirth: number) {}
```
d0sboots: I don't think `readonly` is particularly useful on function arguments, the risk of assigning to a parameter is low and it makes the signature *much* more verbose.

src/Paths/FileMetadata.ts
```diff
@@ -0,0 +1,16 @@
+export class FileMetadata {
```
d0sboots, line 3: Your class appears to lack members XD

src/Paths/FileMetadata.ts
```diff
@@ -0,0 +1,16 @@
...
+  constructor(readonly timeOfAccess: number, readonly timeOfModification: number, readonly timeOfBirth: number) {}
+
+  static new(): FileMetadata {
```
d0sboots: I would just roll this into the constructor, which can take optional arguments. Part of why I think that is that this has the very generic name of `new`, indicating it has the same job as the constructor. Another option would be to change the name.

src/Paths/FileMetadata.ts
```diff
@@ -0,0 +1,16 @@
...
+    const now = Date.now();
+    return new FileMetadata(now, now, now);
+  }
+
+  read(): FileMetadata {
+    return new FileMetadata(Date.now(), this.timeOfModification, this.timeOfBirth);
```
d0sboots: Optional: I'd lean towards POSIX-like names of `mtime`, `atime` `ctime` and `btime` personally. Those are admittedly more abbreviated, the main benefit is that they're also rather standard.

src/Script/Script.ts
```diff
@@ -14,6 +15,19 @@ export class Script implements ContentFile {
   filename: ScriptFilePath;
   server: string;
 
+  // Used for the Remote File API,
+  // to resolve conflicts when synchronizing files outside the game
+  #fileMetadata: FileMetadata;
```
d0sboots: We don't really use private members anywhere else, and I'm not sure they're warranted here? Generally we treat most class member vars as private (or at least private-ish) and don't poke them directly.

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -4025,12 +4025,12 @@ export interface Go {
...
    * For example, a 5x5 board might look like this:
    *
-   [<br/>  
-      "XX.O.",<br/>  
-      "X..OO",<br/>  
-      ".XO..",<br/>  
-      "XXO.#",<br/>  
-      ".XO.#",<br/>  
+   [<br/>
+	  "XX.O.",<br/>
```
d0sboots: Was this change intentional?

src/TextFile.ts
```diff
@@ -11,17 +12,33 @@ export class TextFile implements ContentFile {
   /** The content of the file. */
   text: string;
 
+  // Used for the Remote File API,
+  // to resolve conflicts when synchronizing files outside the game
+  #fileMetadata: FileMetadata;
+
+  get metadata(): object {
```
d0sboots: Ugh, having to copy all this stuff is gross and strongly suggests the need for a superclass.

Hoekstraa:
Thanks for the additional info.
As you can see in the commit history, I've moved away from representing it as accurately to POSIX as possible.
Mostly due to the fact that implementing it exactly like how Unix/Linux/POSIX does it, would mean we'd need to make significant changes to how we're managing files right now. Especially due to the characteristics of moving and copying.

So my recommendation would be to keep it simple and only record the information we can easily retrieve. Without, say, rewriting every instance of file modification.

src/Netscript/RamCostGenerator.ts
```diff
@@ -548,6 +548,7 @@ export const RamCosts: RamCostTree<NSFull> = {
   write: 0,
   tryWritePort: 0,
   read: 0,
+  fileMetadata: 0,
```
d0sboots: If we want to be unix-like, we would call this `stat`, and provide a `utime` call to set times. Otherwise, this should probably be `getFileMetadata` and have an accompanying `setFileMetadata`.

src/Paths/FileMetadata.ts
d0sboots: I *suspect* we're going to want to update the metadata in-place, although I suppose either way can work depending on programming style.

d0sboots (commented):
I know this is still a draft, but here's some early feedback on stuff.

Overall I think this is going in a good direction.

d0sboots:
Going through my pending/long-ignored PRs:
How are you doing here? Still drafting, or are you looking for feedback?

Hoekstraa:
Not drafting, no feedback needed.
Currently the RFA-side needs to be tested still, and potentially some CLI commands could be added. The base functionality in my eyes is fine as it is at the moment.

Oh, and some bad test is going red. So gotta deal with that as well.

PR #1276, opened 2024-05-15
EDITOR: Improved infinite loop checking

G4mingJon4s:
The infinite loop checking the game provides is meant to help newer players find infinite loops that might crash the game.
While this is mostly working, there are some cases, where infinite loops might be intentional. The player might want a never-ending script or an infinite generator.

To improve upon the existing detection system, following functionality has been added:
- Testing for another case, where the while loop will never stop, while (1)
- Excluding any while loops, that have been marked by the player with the comment `// @ignore-infinite`
- Showing multiple infinite while loops in one script.

These changes will help newer players spot infinite loops, while giving the intermediate player the option to disable the checking all together.

### DIFF
--- a/src/Documentation/doc/basic/scripts.md
+++ b/src/Documentation/doc/basic/scripts.md
@@ -62,6 +62,21 @@ For example, if a script run with 1 thread is able to hack $10,000, then running
 
 When "multithreading" a script, the total [RAM](ram.md) cost can be calculated by simply multiplying the [RAM](ram.md) cost of a single instance of your script by the number of threads you will use. [See [`ns.getScriptRam()`](https://github.com/bitburner-official/bitburner-src/blob/bec737a25307be29c7efef147fc31effca65eedc/markdown/bitburner.ns.getscriptram.md) or the `mem` terminal command detailed below]
 
+## Never-ending scripts
+
+Sometimes it might be necessary for a script to never end and keep doing a particular task.
+In that case you would want to write your script in a never-ending loop, like `while (true)`.
+
+However, if you are not careful, this can crash your game.
+If the code inside the loop doesn't `await` for some time, it will never give other scripts and the game itself time to process.
+
+<br />
+
+To help you find this potential bug, any `while (true)` loop without any `await` statement inside it will be marked.
+A red decoration will appear on the left side of the script editor, telling you about the issue.
+
+If you are really sure that this is not an oversight, you can suppress the warning using the comment `// @ignore-infinite` directly above the loop.
+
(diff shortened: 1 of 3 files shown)


d0sboots:
I like the idea of checking for multiple infinite loops in one script. But I also sort of like the idea of there being a simple construct (that isn't a comment) to disable the checking. `while (1)` used to have that function. If there's another simple non-awkward construct, I'd be for promoting that, otherwise I feel like we should leave `while (1)` in as the escape-hatch and then leave out the more-obscure comment syntax.

G4mingJon4s:
> But I also sort of like the idea of there being a simple construct (that isn't a comment) to disable the checking. `while (1)` used to have that function.

While I agree that `while (1)` was pretty convenient to disable the check, to me it seems kind of unintuitive.
I added the comment to make it more explicit.
Someone reading the code will understand that the loop is meant to go on forever,
instead of guessing whether or not it is a mistake.

I can remove the `while (1)` check if it bothers you, though I only see benefits to the ignore comment.

d0sboots:
I agree that `while (1)` to ignore the check is unintuitive. However, I think having a magic comment is *also* unintuitive. What do I mean?

Well, the comment *does* document the intent clearly. However, it is very easy to get the comment wrong, or forget how to spell it. Is it `//ignore-infinite`? `// ignore-infinite`? `// @disable-infinite`? `// @ignore-infinite-check`? (The correct answer, of course, is none of those.) All of those document the *intent* equally well, they just won't *function* in terms of disabling the checking. This is my primary objection to magic comments - unlike real syntax, there is an infinite space of potential comments to choose from, but only one of them will work.

In contrast, `while (1)` is unintuitive to hit on, but you will *never* forget how to spell it.

G4mingJon4s:
Okay, if it is about documenting the comment and its use-case, then this is something that can be improved.
Currently, the comment is mentioned in the decoration text on the left, though I can see that being overlooked.
Making a page about infinite loops in the documentation might be an option.

d0sboots:
My comment was more describing why I have an issue with magic comments in general, but if you can improve the doc situation that would solve it. I think a good improvement that adds it right where it needs to be would be including the comment syntax in the tooltip that you see when you hover over the infinite-loop red warning.

Suggestion:
```
Possible infinite loop, `await` something.
If this is a false-positive, use `// @ignore-infinite` to suppress.
```

G4mingJon4s:
I made the decoration text more explicit and added a section explaining the infinite loop detection in the scripts documentation page.
This way, the player is informed of the detection, knows how it works and is explicitly told how to suppress it.

PR #1277, opened 2024-05-15
BUGFIX: Engine does not send all faction invitations

catloversg:
When the player calls `ns.singularity.checkFactionInvitations`, they expect that the returned value is the array of factions for which they satisfy conditions. However, the game engine sends only 1 invitation of a random faction in that array. This behavior does not make sense to me.

### DIFF
--- a/src/engine.tsx
+++ b/src/engine.tsx
@@ -162,9 +162,8 @@ const Engine: {
   checkCounters: function () {
     if (Engine.Counters.checkFactionInvitations <= 0) {
       const invitedFactions = Player.checkForFactionInvitations();
-      if (invitedFactions.length > 0) {
-        const randFaction = invitedFactions[Math.floor(Math.random() * invitedFactions.length)];
-        inviteToFaction(randFaction);
+      for (const invitedFaction of invitedFactions) {
+        inviteToFaction(invitedFaction);
       }
       Engine.Counters.checkFactionInvitations = 100;
     }


d0sboots:
This changes the actual game behavior, not just for `singularity`. Did you check that the pop-up boxes here make sense if you fulfil multiple factions across one tick?

Otherwise, I agree, waiting for random factions to pop never made sense to me either.

catloversg:
Current behavior: each new pop-up dialog (for the new invitation from a random faction) replaces the old one if it still exists.
New behavior: only one pop-up dialog (for the invitation from the last faction in the array) is shown.

In the current version, if the player leaves the UI run for a long time, there will only 1 pop-up because the new one replaces the old one. With this change, all pop-ups are "merged" immediately. I can refactor `InvitationModal` to support multiple dialogs if you want.

d0sboots:
The case I'm worried about is not when the player leaves the UI for a long time (in that case, you *expect* to miss stuff), but rather when you're playing actively. Especially in the transition period, right now everyone is trained that invites take a while and you just have to wait a bit longer to get the 2nd invite. This would instead change it so the 2nd invite *never* appears, and that's what has me slightly concerned.

catloversg:
The best way to deal with this problem is to refactor `InvitationModal` to support multiple dialogs, so that each invitation is a separate dialog. I'll submit another PR for that. We can merge this PR after that.

catloversg:
PR #1287 was merged. Each invitation is a separate dialog now.

PR #1278, opened 2024-05-15
MISC: Remove js-sha256

catloversg:
This PR replaces `js-sha256` with `crypto.subtle.digest`. `crypto.subtle.digest` is supported in all modern browsers, so there is no need to use another library. `js-sha256` also had some bugs that generated wrong hash values in its old versions.

Tested on web app and Electron app.


d0sboots:
Did you test if this works offline (i.e., in a browser using a `file://` URL)?
`cypto.subtle.digest` typically only works in secure contexts, so I suspect we won't be able to rely on it.

catloversg:
I tested it in multiple environments and conditions. It works properly in all of them. I tested these things:
- Chrome, Firefox.
- http (127.0.0.1 and localhost), https (custom domain name), file URL (file://path/test.html)
- Electron app. Our Electron app loads the index file via `file://`.

Quote from MDN: "Locally-delivered resources such as those with `http://127.0.0.1` URLs, `http://localhost` and `http://*.localhost` URLs (e.g. `http://dev.whatever.localhost/`), and file:// URLs are also considered to have been delivered securely."

Reference: https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts

Test code:
```html
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="utf-8" />
    <title>test</title>
  </head>
  <body>
    <script type="module">
      console.log("window.isSecureContext", window.isSecureContext);
      const inputUInt8Array = new TextEncoder().encode("abc");
      const hashBuffer = await crypto.subtle.digest("SHA-256", inputUInt8Array);
      const hashArray = Array.from(new Uint8Array(hashBuffer));
      console.log(hashArray.map((b) => b.toString(16).padStart(2, "0")).join(""));
    </script>
  </body>
</html>
```

URL:
- `http://127.0.0.1/test-tools/test.html`
- `http://localhost/test-tools/test.html`
- `https://bitburner.local/test-tools/test.html`
- `file://<path>/test-tools/test.html`

d0sboots:
So, this definitely breaks *my* dev workflow:

image

I develop on a linux server that is different than the (windows) machine I play from. A little unconventional perhaps, but it works just fine... as long as http continues to work.

AFAIK, this would be the first thing to require a secure context. I would bite the bullet and come up with some proxy workaround if this were introducing some important functionality... but I don't think it's worth breaking http support just to clean up a single dependency.

catloversg:
I have a different opinion. In most cases, I always prefer getting rid of unnecessary dependencies over supporting convenient setups for developers, especially when it's a niche setup (http but without using localhost/loopback IP).

If you think that getting rid of `js-sha256` does not worth it, I suggest updating it. That library had bugs that generated wrong hash values in its old versions. That's one of the reasons that made me notice that library.

Edit: An alternative choice is to use cyrb53 in `StringHelperFunctions.ts`. It's not a cryptographic hash function, but it's still usable for our use case here.

d0sboots:
> I have a different opinion. In most cases, I always prefer getting rid of unnecessary dependencies over supporting convenient setups for developers, especially when it's a niche setup (http but without using localhost/loopback IP).

If it were *only* to support developer usecases, I would generally agree with you. For instance, I wouldn't want to add more dependencies to the (shipped) game just to make development easier. (Dev dependencies are a different matter.)

In this case, I can't shake the feeling that there may be other usecases we don't know of that are non-secure contexts. I already know of the one, but that doesn't mean there aren't more. (There also might not be.) Particularly the fact that we support BB running in a variety of ways (Electron, web page, completely offline *that I know of*) makes me hesitant here.

@Snarling might feel otherwise.

catloversg:
@d0sboots I suggest a new option: use `cyrb53` (`src\utils\StringHelperFunctions.ts`) instead of SHA-256:
- We have already used it for generating the player's identifier.
- We don't need a cryptographic hash function for `AlertManager`. It's extremely unlikely that collisions occur in our usecases. Even when it happens, the player "only" misses a dialog message. It's not good, but it's also not a catastrophe.

d0sboots:
Sounds good to me. Reduce the surface of random hashes we are using (another was just killed for doing a useless UUID)

src/ui/React/AlertManager.tsx
```diff
@@ -46,9 +46,11 @@ export function AlertManager({ hidden }: { hidden: boolean }): React.ReactElemen
 
   const alertMessage = alerts[0]?.text || "No alert to show";
 
-  function getMessageHash(text: string | JSX.Element): string {
-    if (typeof text === "string") return sha256(text);
-    return sha256(JSON.stringify(text.props));
+  async function getMessageHash(text: string | JSX.Element): Promise<string> {
```
d0sboots: This seems needlessly async, and it has potential race condition implications.

catloversg: I forgot removing them.

PR #1282, opened 2024-05-16
BUGFIX: Fix wrong implementation of getRandomInt

catloversg:
This PR fixes problems with the current implementation of getRandomInt.

Test code:
```js
// Old implementation
export function getRandomInt(min, max) {
  const lower = Math.min(min, max);
  const upper = Math.max(min, max);
  return Math.floor(Math.random() * (upper - lower + 1)) + lower;
}

// New implementation
export function getRandomIntInclusive(min, max) {
  const minCeiled = Math.ceil(Math.min(min, max));
  const maxFloored = Math.floor(Math.max(min, max));
  return Math.floor(Math.random() * (maxFloored - minCeiled + 1) + minCeiled);
}
```

Test cases:
Case 1: `min = 0`, `max = 2.9`.
- Old: [0, 3].
- New: [0, 2].

Case 2: `min = 0.1`, `max = 2.9`.
- Old: [0.1, 3.1]. The random number is a float.
- New: [1, 2].

We have not seen any bug until now because we always pass "integers" to the old implementation. It works properly in that case.

Remarkable changes:
- Rename `getRandomInt` to `getRandomIntInclusive`.
- Throw an error if min is greater than max. Current implementation silently swaps min and max if min is greater than max. This may be convenient, but it can lead to subtle bugs. I think it's wrong to silently "mitigate" a problem in the caller instead of throwing an error.

Reference: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/random


src/utils/helpers/getRandomIntInclusive.ts
```diff
@@ -0,0 +1,17 @@
...
+ * Reference: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/random
+ *
+ * @param min The minimum value in the range.
+ * @param max The maximum value in the range.
+ */
+export function getRandomIntInclusive(min: number, max: number): number {
+  if (min > max) {
+    throw new Error(`Min is greater than max. Min: ${min}. Max: ${max}.`);
+  }
+  const minCeiled = Math.ceil(min);
```
d0sboots: If we're adding hard error-checking, I rather feel like passing non-integers should also throw an error, checked by `Number.isInteger`. Alternatively, you could use the integer-type tag that exists and force all callsites to at least assert that they're passing in an integer.

d0sboots: Once you know they're integers, you don't need to floor and ceil.

PR #1285, opened 2024-05-17
BUGFIX: Correct BN10 Sleeve starting Shock

TheAimMan:
Looks like an old bug that was supposed to make Shock be maxed at 25 used the wrong math expression, probably copied from the sync expression.

Not sure if sync should really start so low for BN10 in order to get the player a better start with the sleeve.

### DIFF
--- a/src/PersonObjects/Player/PlayerObjectGeneralMethods.ts
+++ b/src/PersonObjects/Player/PlayerObjectGeneralMethods.ts
@@ -139,7 +139,7 @@ export function prestigeSourceFile(this: PlayerObject): void {
 
   if (this.bitNodeN === 10) {
     for (let i = 0; i < this.sleeves.length; i++) {
-      this.sleeves[i].shock = Math.max(25, this.sleeves[i].shock);
+      this.sleeves[i].shock = Math.min(25, this.sleeves[i].shock);
       this.sleeves[i].sync = Math.max(25, this.sleeves[i].sync);
     }
   }


Zelow79:
Personally I think a starting sync of 50 sounds pretty good for sleeves in bn10 for my opinion. Nice bug catch.

d0sboots:
> Looks like an old bug that was supposed to make Shock be maxed at 25 used the wrong math expression, probably copied from the sync expression.

Did you misword? Shock *was* maxed at 25.

Edit: I think I see what you mean, difference between "maxed" as a verb vs. adjective XD

d0sboots:
So, because I am both lazy and unfamiliar with BN10, can you give me a quick/slightly more detailed justification for why you think it was always meant to be this way? Basically, both`min` and `max` make sense, but in different contexts.

TheAimMan:
> So, because I am both lazy and unfamiliar with BN10, can you give me a quick/slightly more detailed justification for why you think it was always meant to be this way? Basically, both`min` and `max` make sense, but in different contexts.

Because Sleeves always start with 100 shock.  Why would you need a Max of 25 or 100?  It will always be 100.  Sync is the good stat that, without memory always starts at 0 or 1.  That's why you would want the Max of that.

Zelow79:
Also, it's the node for sleeves, so they should get a slight boost in accessibility ease. Just like corps, gangs and almost all mechanics you get from a node.I agree this had to be an error.

PR #1286, opened 2024-05-18
CCT: Improve performance of Valid Math Expressions checker

### LINKED ISSUE #1095: The solution checker for "Find All Valid Math Expressions" is too slow
I noticed that it takes far longer for the game to check the solution to this contract than it does to generate the answer. Just choosing one example, my solver took 55ms to generate a solution and then on submission it took 322ms to check it. It looks like the code is doing an iterative search of the answers against the player's answers using Array.includes. 

Just a suggestion, but perhaps the checker could convert the answer array into a Set instead? That seems like it is a better data structure for this type of problem?

yichizhng:
As the title says. Test build at https://yichizhng.github.io/bitburner-src/

Script for testing:
```js
function helper(res, path, num, target, pos, evaluated, multed) {
    if (pos === num.length) {
        if (target === evaluated) {
            res.push(path);
        }
        return;
    }

    for (let i = pos; i < num.length; ++i) {
        if (i != pos && num[pos] == "0") {
            break;
        }
        const cur = parseInt(num.substring(pos, i + 1));

        if (pos === 0) {
            helper(res, path + cur, num, target, i + 1, cur, cur);
        } else {
            helper(res, path + "+" + cur, num, target, i + 1, evaluated + cur, cur);
            helper(res, path + "-" + cur, num, target, i + 1, evaluated - cur, -cur);
            helper(res, path + "*" + cur, num, target, i + 1, evaluated - multed + multed * cur, multed * cur);
        }
    }
}

/** @param {NS} ns */
    ns.clearLog();
    ns.disableLog('ALL');
    for (let i = 0; i < 1000; ++i) {
        ns.codingcontract.createDummyContract(
            "Find All Valid Math Expressions"
        );
    }
    let files = ns.ls('home', '.cct');
    let start = performance.now();
    let answers = new Map();
    for (let file of files) {
        const result = [];
        let [n, t] = ns.codingcontract.getData(file);
        helper(result, "", n, t, 0, 0, 0);
        answers.set(file, result);
    }
    ns.print(`generated answers in ${performance.now() - start}ms`);
    start = performance.now();
    for (let file of files) {
        ns.codingcontract.attempt(answers.get(file), file);
    }
    ns.print(`checked answers in ${performance.now() - start}ms`);
}
```

closes #1095

### DIFF
--- a/src/data/codingcontracttypes.ts
+++ b/src/data/codingcontracttypes.ts
@@ -1282,8 +1282,9 @@ export const codingContractTypesMetadata: ICodingContractTypeMetadata[] = [
       // Prevent player from providing extra wrong answers and still receiving credit
       if (result.length !== sanitizedPlayerAnsArr.length) return false;
 
-      for (const expr of result) {
-        if (!sanitizedPlayerAnsArr.includes(expr)) {
+      const resultsSet = new Set(result);
+      for (const expr of sanitizedPlayerAnsArr) {
+        if (!resultsSet.has(expr)) {
           return false;
         }
       }


d0sboots:
Does `Find All Valid Math Expressions` have a test? What you wrote there seems like it would be straightforward to adapt (we might not want 1000 iterations, depending on speed), and if we don't have a test for this we really ought to.

If you really don't want to (or it turns into a big pain) I'll take the improvement as-is.

yichizhng:
That script just uses the same code that the game uses to check the solution, so I don't think that it has any value as a test for correctness. A test for performance has too much variance due to the randomness in input size.

The point of the test script is that the difference in the two times printed is the amount of time spent comparing the results (plus whatever other overhead). Some sample results on Chrome (Firefox has better performance on Array and worse on Set, but it's still an improvement)

Before:
```
generated answers in 12703.600000023842ms
checked answers in 29619.900000035763ms
```

After:
```
generated answers in 11085.5ms
checked answers in 11453.300000011921ms
```

PR #1292, opened 2024-05-20
BUGFIX: Fix hidden debt of exp at level 1

### LINKED ISSUE #189: Odd exp penalty upon Aug Install (may be BN10 related?)
As you can see below, I have non-zero amounts of experience (per stats page) but the sidebar shows me as having 0 progress towards a max needed of 32, with 41 remaining.
image

Hitting level 2 requires closer to 45 exp than 32 (once the negative is paid, the bar progresses normally; eg if the true amount needed is 45, then as soon as you have 14 exp the progress will show as 1 out of 32 with 31 remaining). Happens every time I've started from base stats on this bitnode, don't know if it is bitnode 10 specific or not.

- [x] Save file [bitburnerSave_1667529047_BN10x1.zip](https://github.com/bitburner-official/bitburner-src/files/9934581/bitburnerSave_1667529047_BN10x1.zip)
- [x] Reproduction steps:
  1. Open attached save
  2. Install augmentations
  3. Get a job at Joe's FNS and start work
  4. Look at stats page and compare with sidebar
- [x] Version of the game, e.g. Bitburner v2.1.0(8f4636cb)

catloversg:
The player's stat levels start at 1, but the respective exp points start at 0. This creates a weird hidden "debt" of exp at level 1 when mult is smaller than ~0.9914716399982745. It was reported in #189.

Reproducible test case:
- Load a clean save file (Options -> Delete Save).
- Use dev menu to add SF 1.1 (simulate a normal save file that beats BN 1.1).
- Bitflume to BN 10.1.
- We have:
  - `calculateExp(1, Player.mults.hacking * currentNodeMults.HackingLevelMultiplier)` = 24.816093628051703
  - `calculateExp(2, Player.mults.hacking * currentNodeMults.HackingLevelMultiplier)` = 69.52860629547718

As we can see with the UI, the player must pay off the hidden "debt" of level 1. This is confusing for the player. This PR fixes this problem.

Please note that this PR does not fix/change the current formula(s) or how exp/skill works (e.g., level 1->2 is harder than level 2->3). It only "fixes" the problem by returning 0 when skill is 1. It's exactly how it works internally (it needs 0 exp to get to level 1). After applying it, level 1->2 still needs 69.52860629547718.

Fixes #189.

### DIFF
--- a/src/PersonObjects/formulas/skill.ts
+++ b/src/PersonObjects/formulas/skill.ts
@@ -10,6 +10,9 @@ export function calculateSkill(exp: number, mult = 1): number {
 }
 
 export function calculateExp(skill: number, mult = 1): number {
+  if (skill === 1) {
+    return 0;
+  }
   const value = Math.exp((skill / mult + 200) / 32) - 534.6;
   return clampNumber(value, 0);
 }
--- a/src/PersonObjects/formulas/skill.ts
+++ b/src/PersonObjects/formulas/skill.ts
@@ -10,9 +10,6 @@ export function calculateSkill(exp: number, mult = 1): number {
 }
 
 export function calculateExp(skill: number, mult = 1): number {
-  if (skill === 1) {
-    return 0;
-  }
   const value = Math.exp((skill / mult + 200) / 32) - 534.6;
   return clampNumber(value, 0);
 }
--- a/src/Prestige.ts
+++ b/src/Prestige.ts
@@ -25,10 +25,23 @@ import { CONSTANTS } from "./Constants";
 import { LogBoxClearEvents } from "./ui/React/LogBoxManager";
 import { initCircadianModulator } from "./Augmentation/Augmentations";
 import { Go } from "./Go/Go";
+import { calculateExp } from "./PersonObjects/formulas/skill";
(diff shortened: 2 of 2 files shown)


d0sboots:
This fixes *a* bug, but I think it would be better to tackle the root of the bug. I.e. when prestiging etc., set exp such that skill is 1 by the current formula. This fixes the debt and the other oddities.

catloversg:
In the latest commit, I initialize exp points of Player in appropriate locations.

I tested these cases:
- Soft reset.
- Install augmentations.
- Backdoor WD.
- Run b1t_flum3.exe.
- `ns.singularity.softReset()`.
- `ns.singularity.installAugmentations()`.
- `ns.singularity.destroyW0r1dD43m0n()`.

d0sboots:
Nice! Sorry for being slow on these.

I think this might miss the case of "brand new game," since I don't think that goes through the prestige functions. But that's minor if so, I'll merge this now to handle the bulk of it.

catloversg:
@d0sboots I tested that case as well, and it works well. With a new game, `mult` is 1, so the expected exp is 0 for skill level 1. The initial exp is set to 0, so there is nothing to do in that case.

PR #1295, opened 2024-05-21
DOCUMENTATION: Fix wrong examples in NetscriptDefinitions.d.ts

### LINKED ISSUE #1294: Bad examples in markdown
In both bitburner.ns.getpurchasedservercost.md and bitburner.singularity.getdarkwebprogramcost.md, the example at the bottom use ${ns.formatMoney(cost)}

I could find no other reference to this function and when I tried to test it I got back a "function not defined" error.

I went digging and found formatMoney in various places in the code but they are all pulling from src/ui/formatNumber.ts, so not a directly player accessible function.

Can I suggest removing or changing the examples or making some of the other formatX function available?

catloversg:
`formatMoney` is not exposed via NS API, but it's used in the examples of `getPurchasedServerCost` and `getDarkwebProgramCost`. This PR changes it to `formatNumber`.

Fixes #1294.

### DIFF
--- a/markdown/bitburner.codingcontract.attempt.md
+++ b/markdown/bitburner.codingcontract.attempt.md
@@ -36,7 +36,7 @@ Attempts to solve the Coding Contract with the provided solution.
 
 
 ```js
-const reward = codingcontract.attempt(yourSolution, filename, hostname);
+const reward = ns.codingcontract.attempt(yourSolution, filename, hostname);
 if (reward) {
   ns.tprint(`Contract solved successfully! Reward: ${reward}`)
 } else ns.tprint("Failed to solve contract.")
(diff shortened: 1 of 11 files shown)


Shadewing666:
If you are going to change `costs ${ns.formatMoney(cost)}` to `costs ${ns.formatNumber(cost)}`
may I suggest `costs $${ns.formatNumber(cost)}` or `costs ${"$"+ns.formatNumber(cost)}`

This will retain the $ that is lost from not using`formatMoney`

catloversg:
@Shadewing666 It's a part of the latest commit.

mctylr-gh:
The dollar sign needs to be [escaped in a template literal](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#description). 

` $${ns.formatNumber(cost)}` should be `\$${ns.formatNumber(cost)}` when used in a template literal.

catloversg:
There is no need to escape it in this case. We only need to escape it if we want to prevent interpolation.
```js
`$${1}` === "$1" // true
```

d0sboots:
Did you go and try to compile all the examples? Nice work.

PR #1296, opened 2024-05-21
IPVGO: Support playing manually as white against your scripts using the No AI type board

ficocelliguy:
* Prevent the No AI board from making AI moves when accessed via the API
* Ensure player manual moves as white are handled correctly and returned in the Go.nextMove promise
* Added more UI support for playing as white, or playing both colors, manually


src/Go/boardAnalysis/boardAnalysis.ts
```diff
@@ -1,6 +1,6 @@
-import type { Board, BoardState, Neighbor, PointState, SimpleBoard } from "../Types";
+import { Board, BoardState, Neighbor, Play, PointState, SimpleBoard } from "../Types";
```
d0sboots: Although technically the way we have things configured right now there's no issues with importing stuff that we use only for types, it's a good habit. At some point (maybe) we might enable some options to get the codebase clean in that regard.

To that effect: I'd leave the existing ones as `import type`, and and new line for the regular import.

src/Go/boardAnalysis/goAI.ts
```diff
@@ -15,20 +15,29 @@ import {
...
   getAllValidMoves,
+  getPreviousMoveDetails,
 } from "./boardAnalysis";
 import { findDisputedTerritory } from "./controlledTerritory";
 import { findAnyMatchedPatterns } from "./patternMatching";
 import { WHRNG } from "../../Casino/RNG";
 import { Go, GoEvents } from "../Go";
 
 let currentAITurn: Promise<Play> | null = null;
+let currentTurnResolver: ((value: Play | PromiseLike<Play>) => void) | null = null;
```
d0sboots: Why the squishiness on `value`'s type? Most things either should be a promise, or not.

src/Go/boardAnalysis/goAI.ts
```diff
@@ -15,20 +15,29 @@ import {
...
 
 /**
  * Retrieves a move from the current faction in response to the player's move
  */
 export function makeAIMove(boardState: BoardState): Promise<Play> {
   // If AI is already taking their turn, return the existing turn.
-  if (currentAITurn) return currentAITurn;
+  if (currentAITurn) {
+    return Go.nextTurn;
```
d0sboots, line 34: This is a subtle change. Remind me why you're doing it?

src/Go/boardAnalysis/goAI.ts
```diff
@@ -15,20 +15,29 @@ import {
...
 let currentAITurn: Promise<Play> | null = null;
+let currentTurnResolver: ((value: Play | PromiseLike<Play>) => void) | null = null;
 
 /**
  * Retrieves a move from the current faction in response to the player's move
  */
 export function makeAIMove(boardState: BoardState): Promise<Play> {
   // If AI is already taking their turn, return the existing turn.
-  if (currentAITurn) return currentAITurn;
+  if (currentAITurn) {
```
d0sboots: I think this is the only thing that reads `currentAITurn` anymore. Probably it should be turned back into a boolean named `isAIThinking` or smth.

src/Go/boardAnalysis/goAI.ts
```diff
@@ -15,20 +15,29 @@ import {
...
+    return Go.nextTurn;
+  }
+  if (boardState.ai === GoOpponent.none) {
+    GoEvents.emit();
+    return (currentAITurn = Go.nextTurn = new Promise((resolve) => (currentTurnResolver = resolve)));
```
d0sboots: This bypasses the finally, so it never resets currentAITurn. You're also emitting GoEvents synchronously, instead of when the currentTurnResolver resolves?

I think what you want is to conditionally construct either this resolved promise, *or* the `getMove.then` promise below. And then either way, return `that.finally(...)`.

src/Go/boardAnalysis/goAI.ts
```diff
@@ -63,6 +72,18 @@ export function makeAIMove(boardState: BoardState): Promise<Play> {
   return Go.nextTurn;
 }
 
+/**
+ * Resolves the current turn.
+ * This is used for players manually playing against their script on the no-ai board.
+ */
+export function resolveCurrentTurn() {
+  if (currentTurnResolver) {
+    currentTurnResolver(Promise.resolve(getPreviousMoveDetails()));
```
d0sboots: I don't think you need, or particularly want to call `Promise.resolve` here.

d0sboots: Optional: Rather than having things spread out like this, to me it would make more sense if `currentTurnResolver` were a function that takes no arguments. You'd just construct a fresh closure at the point where you're currently making this function, and all the logic would be apparent there.

src/Go/boardAnalysis/goAI.ts
ficocelliguy: The linter should do it for us, but it missed this one it seems. It should be a ternary (or short-circuit &&) for one line conditionals, for all of the standard maintainability reasons

https://www.reddit.com/r/C_Programming/comments/iyu0ei/an_argument_on_why_you_should_always_put_curly/

src/Go/boardAnalysis/goAI.ts
ficocelliguy: The promise needs to not resolve here, though, nor set currentAITurn, yet. It waits until resolveCurrentTurn() is called externally, rather than following the code below. In this case the AI we are waiting on is the player, manually playing white against their script (e.g. a human brain made by humans, which is technically an intelligence and artificial XD).

src/Go/boardAnalysis/goAI.ts
ficocelliguy: I don't think I see how that would work. It needs to be a new promise for each move, then resolve with the player's last move details as white, once they finish moving

src/Go/boardAnalysis/boardAnalysis.ts
ficocelliguy: This one was my IDE changing it, it definitely should all be import type

src/Go/boardAnalysis/goAI.ts
d0sboots: I wasn't clear enough. I'm fully onboard with adding curly braces; the subtle change is the switch from `currentAITurn` to `Go.nextTurn`.

src/Go/boardAnalysis/goAI.ts
d0sboots: I misspoke (again), but I would have thought the Idea would be clear from the code. The promise you are constructing here is *not* resolved (yet), since it is triggered by `currentTurnResolver`.

Later on, you do some custom stuff to perform the action of "when `currentTurnResolver` is called, reset `currentAITurn`." But this is duplicate code; this is already handled by the `finally` promise that chains off the main one. The main promise is not resolved yet either; both are the same in that they are waiting for the turn to be done. This is why it makes sense for them to share the finally.

src/Go/boardAnalysis/goAI.ts
d0sboots: Hmm. I haven't been around much recently, but I should try and catch you on discord. The things I thought would be clear, clearly aren't. XD

src/Go/boardAnalysis/goAI.ts
ficocelliguy: Ah! My first plan was to remove currentAITurn entirely, then it became a boolean flag later, and the above code is the result of that being only partway changed. It makes more sense with the most recent version of the code using `isAiThinking` as the boolean conditional.

src/Go/boardAnalysis/goAI.ts
ficocelliguy: I've changed this to apply the finally() to Go.nextTurn regardless of which branch of the Ai/No AI logic is followed. Let me know if this is closer to what you were thinking.

src/Go/boardAnalysis/goAI.ts
ficocelliguy: I think I understand what you mean now. I have changed currentTurnResolver() to not take any arguments, and instead call the resolver on Go.nextTurn and pass it the previous move's details via a closure built when the No Ai promise is constructed. I've added some comments to try and make that more clear as well

src/Go/boardAnalysis/goAI.ts
d0sboots: Ok, that's all well and good, but it doesn't answer my original question :D

Changing from `currentAITurn` to `Go.nextTurn` here is a subtle change, with subtle implications. They were *almost* the same, but not quite in all circumstances. I can be pretty sure I know what's going on with `currentAITurn`, since it's scoped to this file, but `Go.nextTurn` is effectively global and could be modified anywhere. 

So, why change it?

src/Go/boardAnalysis/goAI.ts
```diff
@@ -63,11 +67,15 @@ export function makeAIMove(boardState: BoardState): Promise<Play> {
...
-    })
-    .finally(() => {
-      currentAITurn = null;
-      GoEvents.emit();
     });
+  }
+
+  // Once the AI moves (or the player playing as white with No AI moves),
+  // clear the isAiThinking semaphore and update the board UI.
+  Go.nextTurn.finally(() => {
```
d0sboots: The new chained promise should be assigned to Go.nextTurn.

src/Go/boardAnalysis/goAI.ts
```diff
@@ -63,11 +67,15 @@ export function makeAIMove(boardState: BoardState): Promise<Play> {
...
+    isAiThinking = false;
+    GoEvents.emit();
```
d0sboots, line 77: I still have a lingering question about the proper timing of this. Clearly, for AI moves the correct time is after the move. I assumed this would be the case for manual moves as well, which is why I encouraged reuse of this block. 

However, you still have the synchronous GoEvents.emit() in the manual case as well. Is it covering something else?

src/Go/boardAnalysis/goAI.ts
d0sboots: Yup, you figured out what I was trying to get at! 👍

src/Go/boardAnalysis/goAI.ts
ficocelliguy: Go.nextTurn is the source of truth for the current turn. In the best case the stuff in this local file stays up-to-date, but (as I discovered during implementing this) this file is not the only place where the white player could start or stop thinking, and having multiple locations editing multiple values that are supposed to be in sync gets messy. If currentAITurn  is truthy it should always be the same as Go.nextTurn, and if it wasn't then it is out-of-date and Go.nextTurn represents the current move's promise.

currentAITurn was added (by snarling, IIRC) because JS doesn't have a way to detect on a promise object if it is resolved or not. I considered making my own promise object type, or always await Go.nextTurn() and then checked on the game state in this file, but both seemed clunky to me

src/Go/boardAnalysis/goAI.ts
ficocelliguy: the "manual case" is the start of the white player's move. The board wrapper UI needs to be updated to reflect the new current player and new score. The pass button in particular changes highlighting and text if they have no legal moves, if passing will end the game, etc. Thus a call to GoEvents.emit()

The finally is the END of the white player's move. The active player has again switched, meaning the above needs to be updated once again to reflect the black player as active and update the pass button, score, etc. 

The millisecond-exact timing doesn't matter super much since its just a UI update, not passing info to another script.

PR #1298, opened 2024-05-21
SLEEVE: Add funds check for sleeve travel

### LINKED ISSUE #991: Can't buy anything after buying first CSEC augmentation
I just bought the first CSEC augmentation and now, after game restart, i can't buy any hacknet servers and on the augmentations screen, available amounts are not checked.
It's like my money wasn't taken into account.

- Version of the game : Bitburner v2.5.1

image
image

gmcew:
This makes sleeve travel consistent with player travel behaviour. This also makes the function description more truthful in that it is now possible to return false.

This does affect game balance by making the debt achievement no longer possible by using sleeve travel; other methods exist.

Closes #991


d0sboots:
In the linked bug, Snarling said this would be more consistent, so I'll take that as a stamp of approval.

PR #1299, opened 2024-05-22
IPVGO: Record full history to avoid infinite ko capture loops on larger boards

ficocelliguy:
Per this discord discussion: https://discord.com/channels/415207508303544321/1221920484111814828/1242562150510362746

The largest board, with its unique shape, lends itself to multiple pairs of capturable stones. If there are enough of these pairs, it creates capture loops longer than the 7 move history currently kept on the board. This change saves the full play history as a simplified string array per boardstate, instead.

Note this is only the boardstate in memory, the history is not preserved on the save file.

### DIFF
--- a/src/Go/boardAnalysis/boardAnalysis.ts
+++ b/src/Go/boardAnalysis/boardAnalysis.ts
@@ -548,7 +548,8 @@ export function findAdjacentLibertiesAndAlliesForPoint(
 }
 
 /**
- * Retrieves a simplified version of the board state. "X" represents black pieces, "O" white, and "." empty points.
+ * Retrieves a simplified version of the board state.
+ * "X" represents black pieces, "O" white, "." empty points, and "#" offline nodes.
  *
  * For example, a 5x5 board might look like this:
  * ```
--- a/src/Go/boardAnalysis/boardAnalysis.ts
+++ b/src/Go/boardAnalysis/boardAnalysis.ts
@@ -44,7 +44,7 @@ export function evaluateIfMoveIsValid(boardState: BoardState, x: number, y: numb
   }
 
   // Detect if the move might be an immediate repeat (only one board of history is saved to check)
-  const possibleRepeat = boardState.previousBoards.find((board) => getColorOnSimpleBoard(board, x, y) === player);
+  const possibleRepeat = boardState.previousBoards.find((board) => getColorOnBoardString(board, x, y) === player);
 
   if (shortcut) {
     // If the current point has some adjacent open spaces, it is not suicide. If the move is not repeated, it is legal
@@ -87,7 +87,7 @@ export function evaluateIfMoveIsValid(boardState: BoardState, x: number, y: numb
   }
   if (possibleRepeat && boardState.previousBoards.length) {
     const simpleEvalBoard = boardStringFromBoard(evaluationBoard);
-    if (boardState.previousBoards.find((board) => simpleEvalBoard === board)) {
+    if (boardState.previousBoards.includes(simpleEvalBoard)) {
       return GoValidity.boardRepeated;
     }
   }
@@ -553,7 +553,13 @@ export function findAdjacentLibertiesAndAlliesForPoint(
  *
  * For example, a 5x5 board might look like this:
  * ```
- *   "XX.O.X..OO.XO..XXO...XOO."
+ * [
+ *   "XX.O.",
+ *   "X..OO",
+ *   ".XO..",
+ *   "XXO..",
+ *   ".XOO.",
+ * ]
  * ```
  *
  * Each string represents a vertical column on the board, and each character in the string represents a point.
@@ -586,6 +592,11 @@ export function simpleBoardFromBoard(board: Board): SimpleBoard {
 /**
  * Returns a string representation of the given board.
  * The string representation is the same as simpleBoardFromBoard() but concatenated into a single string
+ *
+ * For example, a 5x5 board might look like this:
+ * ```
+ *   "XX.O.X..OO.XO..XXO...XOO."
+ * ```
  */
 export function boardStringFromBoard(board: Board): string {
   return simpleBoardFromBoard(board).join("");
@@ -594,6 +605,11 @@ export function boardStringFromBoard(board: Board): string {
 /**
  * Returns a full board object from a string representation of the board.
  * The string representation is the same as simpleBoardFromBoard() but concatenated into a single string
+ *
+ * For example, a 5x5 board might look like this:
+ * ```
+ *   "XX.O.X..OO.XO..XXO...XOO."
+ * ```
  */
 export function boardFromBoardString(boardString: string): Board {
   // Turn the SimpleBoard string into a string array, allowing access of each point via indexes e.g. [0][1]
@@ -643,7 +659,7 @@ export function areSimpleBoardsIdentical(simpleBoard1: SimpleBoard, simpleBoard2
   return simpleBoard1.every((column, x) => column === simpleBoard2[x]);
 }
 
-export function getColorOnSimpleBoard(boardString: string, x: number, y: number): GoColor | null {
+export function getColorOnBoardString(boardString: string, x: number, y: number): GoColor | null {
   const boardSize = Math.round(Math.sqrt(boardString.length));
   const char = boardString[x * boardSize + y];
   if (char === "X") return GoColor.black;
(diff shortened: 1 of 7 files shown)


Snarling:
@ficocelliguy @Sphyxis Any idea how this impacts performance? I imagine it gets taxing to check through the entire history of the game for larger boards late into the game. If the need for this outweighs the performance impact, we can go ahead and merge this, I just want to make sure we've considered the performance side

Sphyxis:
@Snarling @ficocelliguy 

Testing done on the largest size board - ??????

7 board states (current):
-Memory used was about 360mb max for the game 

A max of 65 board states did not produce any infinite loops with 1000's of games played.
-Memory used was about 380mb max for the game

Testing 6500 max board states
-Memory used was about 415mb max for the game

True speed (speed of AI without sleeps) remained the same on all tests, between 0.02ms and 0.4ms

I however did not test this actual change, just played with the old numbers to simulate it.

Kelenius:
We've had a conversation on discord (https://discord.com/channels/415207508303544321/459097632896188436/1244274997292437545) where an idea of packing the board state to reduce memory usage and comparison speed was developed. The idea is to minimize storage and comparison speed, while sacrificing readability, since the data is only used for the repetition rule. Here's my finalized suggestion:

1) Go over the cells in a specific order.
2) Assign a value of 0 to destroyed cell, 1 to empty, 2 to white, and 3 to black. That gives 4 options per cell.
3) Group cells in groups of three. Each group has 64 combinations.
4) Turn the group into a number using `cell1 + cell2 * 4 + cell3 * 16`. This generates a number from 0 to 63.
5) Turn the number into a charcode using `String.fromCharCode(40 + n)`. (I picked 40 somewhat arbitrarily, because 39 is `'`, you can go up to 126 without problems).
6) Combine all chars into one string.

As a result, the entire board state can be turned into a unique string. This should both take up less space than the current realization of `SimpleBoard` (array of strings) and be faster to compare, and packing multiple tiles into one character makes the string shorter.

I originally wanted to have 3 different values for cells and group them into groups of 4, but then I remembered that there are cheats that can change which cells are destroyed.

d0sboots:
I like the idea of storing board states as strings, but I think we should go big or go home: 

* Go big: The state is 2 bits, pack 4 states per byte to fully utilize the string. Strings actually are UTF-16 units, but they always have an optimization for ascii so packing 8 bits per byte or 16 bits per two bytes comes out the same, and sticking to ascii has some readability advantages. 
* Go home: Use descriptive characters like "x", " ", "B", "W" instead. It's 4x the space, but the biggest space savings by far was to switch to a single string instead of an array - this is tiny in comparison. Besides being easier to debug, it would compress well enough that we could probably include it in the save.

ficocelliguy:
I have updated the board history to use a single string per prior board, maintaining the descriptive characters used previously ("X" represents black pieces, "O" white, "." empty points, and "#" offline nodes.)

This is only for the move history: the API accessing the current board still uses a string array, as does the single previous game end state saved to display on the history page, as those do not have multiple items like history does.

src/Go/boardAnalysis/boardAnalysis.ts
```diff
@@ -86,8 +86,8 @@ export function evaluateIfMoveIsValid(boardState: BoardState, x: number, y: numb
     return GoValidity.noSuicide;
   }
   if (possibleRepeat && boardState.previousBoards.length) {
-    const simpleEvalBoard = simpleBoardFromBoard(evaluationBoard);
-    if (boardState.previousBoards.find((board) => areSimpleBoardsIdentical(simpleEvalBoard, board))) {
+    const simpleEvalBoard = boardStringFromBoard(evaluationBoard);
+    if (boardState.previousBoards.find((board) => simpleEvalBoard === board)) {
```
d0sboots: You can use `includes()` here, no need for a closure.

src/Go/boardAnalysis/boardAnalysis.ts
```diff
@@ -548,29 +548,25 @@ export function findAdjacentLibertiesAndAlliesForPoint(
...
  * For example, a 5x5 board might look like this:
  * ```
- * [
- *   "XX.O.",
- *   "X..OO",
- *   ".XO..",
- *   "XXO..",
- *   ".XOO.",
- * ]
+ *   "XX.O.X..OO.XO..XXO...XOO."
```
d0sboots: I think you changed the comments on the wrong place. This is still SimpleBoard, so still the stay form.

src/Go/boardAnalysis/boardAnalysis.ts
```diff
@@ -624,8 +643,9 @@ export function areSimpleBoardsIdentical(simpleBoard1: SimpleBoard, simpleBoard2
   return simpleBoard1.every((column, x) => column === simpleBoard2[x]);
 }
 
-export function getColorOnSimpleBoard(simpleBoard: SimpleBoard, x: number, y: number): GoColor | null {
-  const char = simpleBoard[x]?.[y];
+export function getColorOnSimpleBoard(boardString: string, x: number, y: number): GoColor | null {
```
d0sboots: Function name is wrong, this operates on boardstring

d0sboots (commented):
This doesn't change the savefile format, right?

ficocelliguy:
Correct, the board history is still not saved in the savefile, and the previous board format is unchanged in the save and in memory

Sphyxis:
@ficocelliguy @d0sboots 

I'm done testing my script, so I'm testing this out now.  Currently there is 1 error that appears in the console and the memory usage has jumped.  Console error at the bottom.

Memory: On cleanup it's between 410-460mb
However, the general usage is huge.
Gets to between 600-1250mb usage during the game.

I'll let you know after a few hundred fights against ??? if any loops occur.

Warning: Invalid aria prop `aria-description` on <button> tag. For details, see https://reactjs.org/link/invalid-aria-props
    at button
    at eval (webpack-internal:///./node_modules/@emotion/react/dist/emotion-element-c39617d8.browser.esm.js:55:66)
    at ButtonBase (webpack-internal:///./node_modules/@mui/material/ButtonBase/ButtonBase.js:108:82)
    at eval (webpack-internal:///./node_modules/@emotion/react/dist/emotion-element-c39617d8.browser.esm.js:55:66)
    at IconButton (webpack-internal:///./node_modules/@mui/material/IconButton/IconButton.js:123:82)
    at Tooltip (webpack-internal:///./node_modules/@mui/material/Tooltip/Tooltip.js:245:82)
    at BitNodePortal (webpack-internal:///./src/BitNode/ui/BitverseRoot.tsx:56:86)
    at p
    at eval (webpack-internal:///./node_modules/@emotion/react/dist/emotion-element-c39617d8.browser.esm.js:55:66)
    at Typography (webpack-internal:///./node_modules/@mui/material/Typography/Typography.js:101:87)
    at BitverseRoot (webpack-internal:///./src/BitNode/ui/BitverseRoot.tsx:99:104)
    at div
    at eval (webpack-internal:///./node_modules/@emotion/react/dist/emotion-element-c39617d8.browser.esm.js:55:66)
    at Box (webpack-internal:///./node_modules/@mui/system/esm/createBox.js:35:72)
    at SnackbarProvider (webpack-internal:///./node_modules/notistack/dist/notistack.esm.js:736:24)
    at SnackbarProvider (webpack-internal:///./src/ui/React/Snackbar.tsx:30:19)
    at HistoryProvider (webpack-internal:///./src/ui/React/Documentation.tsx:48:80)
    at BypassWrapper (webpack-internal:///./src/ui/React/BypassWrapper.tsx:7:14)
    at ErrorBoundary (webpack-internal:///./src/ui/ErrorBoundary.tsx:14:5)
    at MathJaxContext (webpack-internal:///./node_modules/better-react-mathjax/esm/MathJaxContext/MathJaxContext.js:6:592)
    at GameRoot (webpack-internal:///./src/ui/GameRoot.tsx:181:19)
    at LoadingScreen (webpack-internal:///./src/ui/LoadingScreen.tsx:27:74)
    at ThemeProvider (webpack-internal:///./node_modules/@mui/private-theming/ThemeProvider/ThemeProvider.js:43:5)
    at ThemeProvider (webpack-internal:///./node_modules/@mui/system/esm/ThemeProvider/ThemeProvider.js:55:5)
    at ThemeProvider (webpack-internal:///./node_modules/@mui/material/styles/ThemeProvider.js:24:14)
    at StyledEngineProvider (webpack-internal:///./node_modules/@mui/styled-engine/StyledEngineProvider/StyledEngineProvider.js:29:5)
    at TTheme (webpack-internal:///./src/Themes/ui/Theme.tsx:365:5)

Sphyxis:
I've gone through over 500 fights on the ??? board.  No stall outs from previous board states.
With that being said, I don't think this is a good fix.  I'd suggest going back to the old format and just raising up the previous boards to 65.  When I did this I never had an issue.  It accounts for 6 flapping points and 1 extra square.

d0sboots:
WDYM by previous format? Do you mean the array-based one? That one should be strictly worse in memory, so I don't think it's recommended.

I'm going to merge this for now, and we can continue to iterate as needed.

PR #1304, opened 2024-05-22
AUGMENTATIONS: Use startingMoney and programs from augmentation definition when prestiging

jjclark1982:
[Reported on Reddit](https://www.reddit.com/r/Bitburner/comments/1cxrqmx/is_this_a_bug_i_should_report/): The `BigDsBigBrain` augmentation includes definitions for `startingMoney` and `programs`, but the prestige function does not check these definitions. The augmentations encountered in normal gameplay all have hard-coded handlers in the prestige function.

This PR uses the values in the augmentation definition to give the player money and programs during prestige.

A small change in behavior is that CashRoot Starter Kit now results in $1,000,000 (from aug) + $1,000 (base) + $151 (donations), instead of a flat $1,000,000. And the money from CashRoot is now tracked by the money source tracker. Does it make sense for this money source to be tracked?

### DIFF
--- a/src/Augmentation/Augmentation.ts
+++ b/src/Augmentation/Augmentation.ts
@@ -195,6 +195,12 @@ export class Augmentation {
   // The Player/Person classes
   mults: Multipliers = defaultMultipliers();
 
+  // Amount of money given to the Player when prestiging with this augmentation.
+  startingMoney: number;
+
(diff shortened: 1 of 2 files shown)


d0sboots:
Augmentations being tracked seems OK to me.

PR #1305, opened 2024-05-23
BUGFIX: fix relative imports

Caldwell-74:
Snarling added relative pathing a while back 
but currently for imports it only works partially

the only part that works is in the editor when hovered over the import 
image

neither the ram calculation nor the compiler uses the relative path

this pr/draw attempts to fix this

caveats:
i havent done much testing with this only used the testscripts ill provide at the end
and variable names are temporary

i just want to get feedback first because this touches sensible areas of the codebase

```js
// test/test2.js
import { test3 } from "test1.js"
/** @param {NS} ns */
  ns.tprint(test3())
}
// test/test1.js
export function test3(){
  return "test"
}
```

### DIFF
--- a/src/Script/RamCalculations.ts
+++ b/src/Script/RamCalculations.ts
@@ -15,7 +15,6 @@ import { Script } from "./Script";
 import { Node } from "../NetscriptJSEvaluator";
 import { ScriptFilePath, resolveScriptFilePath } from "../Paths/ScriptFilePath";
 import { root } from "../Paths/Directory";
-import { FilePath, resolveFilePath } from "../Paths/FilePath";
 
 export interface RamUsageEntry {
   type: "ns" | "dom" | "fn" | "misc";
@@ -57,11 +56,14 @@ function getNumericCost(cost: number | (() => number)): number {
  * Parses code into an AST and walks through it recursively to calculate
  * RAM usage. Also accounts for imported modules.
  * @param otherScripts - All other scripts on the server. Used to account for imported scripts
- * @param code - The code being parsed */
+ * @param code - The code being parsed
+ * @param scriptname - The name of the script that ram needs to be added to
+ * */
+
 function parseOnlyRamCalculate(
   otherScripts: Map<ScriptFilePath, Script>,
   code: string,
-  mainFileName: FilePath,
+  scriptname: ScriptFilePath,
   ns1?: boolean,
 ): RamCalculation {
   /**
@@ -83,7 +85,7 @@ function parseOnlyRamCalculate(
   const parseQueue: string[] = [];
   // Parses a chunk of code with a given module name, and updates parseQueue and dependencyMap.
   function parseCode(code: string, moduleName: string): void {
-    const result = parseOnlyCalculateDeps(code, mainFileName, moduleName);
+    const result = parseOnlyCalculateDeps(code, scriptname, moduleName);
     completedParses.add(moduleName);
 
     // Add any additional modules to the parse queue;
@@ -265,7 +267,7 @@ interface ParseDepsResult {
  * for RAM usage calculations. It also returns an array of additional modules
  * that need to be parsed (i.e. are 'import'ed scripts).
  */
-function parseOnlyCalculateDeps(code: string, filename: FilePath, currentModule: string): ParseDepsResult {
+function parseOnlyCalculateDeps(code: string, filename: ScriptFilePath, currentModule: string): ParseDepsResult {
   const ast = parse(code, { sourceType: "module", ecmaVersion: "latest" });
   // Everything from the global scope goes in ".". Everything else goes in ".function", where only
   // the outermost layer of functions counts.
@@ -344,7 +346,11 @@ function parseOnlyCalculateDeps(code: string, filename: FilePath, currentModule:
     Object.assign(
       {
         ImportDeclaration: (node: Node, st: State) => {
-          const importModuleName = resolveFilePath(node.source.value, filename) as FilePath;
+          const importModuleName = resolveScriptFilePath(node.source.value, filename, ".js");
+          if (!importModuleName)
+            throw new Error(
+              `ScriptFilePath couldnt be resolved in ImportDeclaration. Value: ${node.source.value}  ScriptFilePath: ${filename}`,
+            );
           additionalModules.push(importModuleName);
 
           // This module's global scope refers to that module's global scope, no matter how we
@@ -403,17 +409,18 @@ function parseOnlyCalculateDeps(code: string, filename: FilePath, currentModule:
 /**
  * Calculate's a scripts RAM Usage
  * @param {string} code - The script's code
+ * @param {ScriptFilePath} scriptname - The script's name. Used to resolve relative paths
  * @param {Script[]} otherScripts - All other scripts on the server.
  *                                  Used to account for imported scripts
  */
 export function calculateRamUsage(
   code: string,
-  scriptName: FilePath,
+  scriptname: ScriptFilePath,
   otherScripts: Map<ScriptFilePath, Script>,
   ns1?: boolean,
 ): RamCalculation {
   try {
-    return parseOnlyRamCalculate(otherScripts, code, scriptName, ns1);
+    return parseOnlyRamCalculate(otherScripts, code, scriptname, ns1);
   } catch (e) {
     return {
       errorCode: RamCalculationErrorCode.SyntaxError,
--- a/src/Script/RamCalculations.ts
+++ b/src/Script/RamCalculations.ts
@@ -70,7 +70,7 @@ function parseOnlyRamCalculate(
   /**
    * Maps dependent identifiers to their dependencies.
    *
-   * The initial identifier is __SPECIAL_INITIAL_MODULE__.__GLOBAL__.
+   * The initial identifier is <name of the main script>.__GLOBAL__.
    * It depends on all the functions declared in the module, all the global scopes
    * of its imports, and any identifiers referenced in this global scope. Each
    * function depends on all the identifiers referenced internally.
--- a/src/Script/RamCalculations.ts
+++ b/src/Script/RamCalculations.ts
@@ -114,7 +114,7 @@ function parseOnlyRamCalculate(
     if (!script) {
       return {
         errorCode: RamCalculationErrorCode.ImportError,
-        errorMessage: `File:"${nextModule}" not found on server: ${server}`,
+        errorMessage: `File: "${nextModule}" not found on server: ${server}`,
       };
     }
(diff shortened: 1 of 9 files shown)


src/Script/RamCalculations.ts
```diff
@@ -338,7 +344,7 @@ function parseOnlyCalculateDeps(code: string, currentModule: string): ParseDepsR
     Object.assign(
       {
         ImportDeclaration: (node: Node, st: State) => {
-          const importModuleName = node.source.value;
+          const importModuleName = resolveFilePath(node.source.value, filename) as FilePath;
```
Snarling: Instead of asserting that this is a FilePath instead of null, why not throw if it ends up being null so we can control the error message?

Just based on eyeballing the code, reasons this might end up resolving as null would be player used an invalid character in an import filename, or player used too many ../ and tried going up one level from the root directory, etc. Asserting a real FilePath here if it's possibly null is going to lead to a weird error somewhere else due to us lying to TS

Caldwell-74: oh yeah that was just to get it to work
ill it cleanup today

d0sboots:
My high-level feedback: You're doing good work. If you can get relative paths to work in a sensible, consistent fashion, that would be wonderful. 

However, there is a higher bar than usual for touching this fiddly part of the code: add at least one new test (or test case to an existing test) that fails without your change and passes with it. In other words, a test to validate proper functioning of relative imports, now and into the future.

Caldwell-74:
Thanks.

I will definetly try myself in writing tests. Its just an area i havent ventured in yet so i put it on the end.

Current hold up is the magic string "__SPECIAL_INITIAL_MODULE__"
i need confirmation that this isnt used by another package/the engine 
and that it can safely be replaced by the main scripts file name
because if so it will simplify the code and i wont have to do ugly stuff like 
`parseCode(script.code, nextModule, nextModule);`
CatLover tested it briefly and it does seem to work so i can change that back again and then start doing the test

d0sboots:
I would `grep -r` to be sure, but I'm pretty sure it's only used here (because that's the point).

src/Script/RamCalculations.ts
```diff
@@ -92,7 +101,7 @@ function parseOnlyRamCalculate(otherScripts: Map<ScriptFilePath, Script>, code:
   }
 
   // Parse the initial module, which is the "main" script that is being run
-  const initialModule = "__SPECIAL_INITIAL_MODULE__";
```
d0sboots, line 95: There's some comment references to this that aren't fixed up.

src/Script/RamCalculations.ts
```diff
@@ -101,21 +110,19 @@ function parseOnlyRamCalculate(otherScripts: Map<ScriptFilePath, Script>, code:
...
+        errorMessage: `File: "${nextModule}" not found on server: ${server}`,
+      };
     }
 
     parseCode(script.code, nextModule);
   }
 
   // Finally, walk the reference map and generate a ram cost. The initial set of keys to scan
-  // are those that start with __SPECIAL_INITIAL_MODULE__.
+  // are those that start with the name of the main script.
```
d0sboots, line 125: What happens when the initial script is `/foo.js` and there is `/foo.js.bar.js`?

src/ScriptEditor/ui/ScriptEditorContext.tsx
```diff
@@ -28,14 +31,20 @@ export function ScriptEditorContextProvider({ children, vim }: { children: React
...
-    if (newCode === null || server === null) {
+  const updateRAM: ScriptEditorContextShape["updateRAM"] = (newCode, filename, server) => {
+    if (newCode === null || filename === null || server === null) {
       setRAM("N/A");
       setRamEntries([["N/A", ""]]);
       return;
     }
-
-    const ramUsage = calculateRamUsage(newCode, server.scripts);
+    //temp duplication to turn FilePath into ScriptFilePath
```
d0sboots: Explain this comment, please?

src/ScriptEditor/ui/ScriptEditorRoot.tsx
```diff
@@ -143,6 +143,7 @@ function Root(props: IProps): React.ReactElement {
     infLoop(newCode);
     updateRAM(
       !currentScript || currentScript.isTxt ? null : newCode,
+      currentScript?.path ?? null,
```
d0sboots: `?? null` should be unneeded here?

test/jest/Netscript/StaticRamParsingCalculation.test.ts
```diff
@@ -315,8 +344,119 @@ describe("Parsing NetScript code to work out static RAM costs", function () {
...
+      const libScriptOne = new Script(libNameOne, libCodeOne);
+
+      const libCodeTwo = `
+          export function testRelativeAgain(ns) {
+              return ns.hack("n00dles")
+          }
+        `;
+      const libScriptTwo = new Script(libNameTwo, libCodeTwo);
+
+      const fail_libCodeTwo = `
```
d0sboots: I wouldn't call this ( and variables like it) "fail", because nothing about this file actually fails, does it?

d0sboots (commented):
Looking pretty good!

src/ScriptEditor/ui/ScriptEditorContext.tsx
Caldwell-74: the editor gets scripts and non script files (like txt)
for those something isnt included like no newCode
we cant resolve scriptFilePath before we have checked that its a script 
but after the resolving it its possible that the script is null (shouldnt happen but ts needs a check)
so i need to duplicate the same error handling like right above it because i havent found a better way yet
i planned to just make a dirty fix there and come back to it when iam (mostly/fully) done with the relative path stuff 
and then make a proper

src/Script/RamCalculations.ts
Caldwell-74:
```js
import { test } from "./foo.js.bar"
/** @param {NS} ns */
  ns.tprint(test(ns))
}
//foo.js.bar.js
export function test(ns) {
  return ns.getServer("home")
}
```

tried with this and it seems to just work?
i dont fully understand that lower section of the code where initialModule is used so if there is a special case you see that needs testing other than that let me know

src/Script/RamCalculations.ts
Caldwell-74: changed it to <name of the main script> gonna push it when the other stuff is resolved

src/ScriptEditor/ui/ScriptEditorRoot.tsx
Caldwell-74: `if (newCode === null || filename === null || server === null) {`
it explicity looks for null in updateRAM thats why i assumed it has to be null not undefined
and i tried to keep the changes (for now/this pr) in that section to as little as possible

src/ScriptEditor/ui/ScriptEditorContext.tsx
d0sboots: I think what you're looking for is to add `hasScriptExtension` to the checks above. As a type assertion, you will then be able to use the ContentFilePath as a ScrptFilePath.

test/jest/Netscript/StaticRamParsingCalculation.test.ts
Caldwell-74: the bug i try to test here for was something i completly overlooked at first 
(resolving imports on the second/n layer based of the path of the main script and not the importing scripts path)
called it "fail" because the test would fail when it loads that file by mistake kinda a trap file

src/ScriptEditor/ui/ScriptEditorRoot.tsx
d0sboots: But currentScript can't be undefined? 

If you're worried about it, switch to `== null` to catch both nullish values.

src/Script/RamCalculations.ts
d0sboots: Yeah, I was legitimately asking because I don't know either. I'm vaguely concerned about this case, but I may have to look into it more.

test/jest/Netscript/StaticRamParsingCalculation.test.ts
d0sboots: I would call it 'incorrect" or "notLoaded" instead of "fail", then. Tests should always be written from the perspective of assuming they pass :D (But with appropriate logging to help when they fail)

src/ScriptEditor/ui/ScriptEditorRoot.tsx
Caldwell-74: its intialized with null thats why the other two arguments check for truthyness of currentscript there
its actually a hazard because we have like a dozen of null/truthyness checks for it because of that

i would like to change that when as a whole but assumed thats out of scope of this pr

edit.: 
oh the undefined would come from the optional chaining operator not from openscript itself

test/jest/Netscript/StaticRamParsingCalculation.test.ts
Caldwell-74: but iam not that happy with the naming in the tests anyway i can take a second look on that tomorrow or if you have any names/naming scheme you like to see iam happy to implement that

src/ScriptEditor/ui/ScriptEditorRoot.tsx
d0sboots: Ah, I thought optional chaining passed the value through (i.e. `(null)?.foo === null`), but I see that was wrong. But regardless, you can handle it by changing the check in the function to `== null` (either for just this term or all of them),  which catches null and undefined both.

Caldwell-74: i would go with 
`currentScript && currentScript.path,`
when thats okay cuz then it fits style wise to the other check
i can also change the === null's to == null

d0sboots: `currentScript && currentScript.path` is the same as `currentScript?.path` but less clear/more bug prone IMO. For instance it would be 0 if currentScript were 0; that shouldn't happen, but that's why nullish coalescing exists.

Edit: The other terms have other considerations that make nullish coalescing inappropriate for them (like not wanting to pass undefined to GetServer). I don't think it's important for them to all look the same, but if you want that then OK.

d0sboots:
I think the only thing left here is that I want to do some more detailed testing of my own when I can get back to my computer.

d0sboots:
OK, finally managed to do the testing I wanted. Everything seems to working as expected. Bare imports (`import { foo } from "barename.js"`) resolve to the root directory, which is correct/necessary to prevent breakage.

PR #1307, opened 2024-05-24
MISC: Cap BN 12 Prestige Benefit

jeek:
Caps free NFG gained from beating BN12 to 50 when not in BN12, or 35842 otherwise.

Per TheAimMan, 35843 levels of NFG makes it impossible to hack.


jeek:
Bug. Trying to kill PR now.

PR #1309, opened 2024-05-24
DOCUMENTATION: Fix wrong examples in NetscriptDefinitions.d.ts

catloversg:
This is a follow-up PR for #1295. In this PR, I tried running all examples instead of "eyeballing" them, and I caught some more errors.


d0sboots:
sweeeeeet

PR #1312, opened 2024-05-25
UI: Add a stop button to bladeburner

Kelenius:
It's already possible to stop an action using `ns.bladeburner.stopBladeburnerAction()` command. This PR adds a button to do it from the UI. It appears in place of "Start" on the action that's being performed currently.


d0sboots:
I see a way that I kinda like better that merges the start and stop button, but there's arguments for and against doing it that way, and it would involve changing the whole PR XD

PR #1314, opened 2024-05-25
BUGFIX: BB Sleeves cannot be assigned to contract programmatically

Sphyxis:
Noticed that ns.sleeve.setToBladeburnerAction(1, "Take on contracts", "Tracking") (all contracts) was always returning false.  Tracked things down to a missing contracts parameter.

Added the parameter.
Tested with an alternating sleeve script.  Worked perfectly for all contracts, infiltration and training

### DIFF
--- a/src/NetscriptFunctions/Sleeve.ts
+++ b/src/NetscriptFunctions/Sleeve.ts
@@ -238,24 +238,24 @@ export function NetscriptSleeve(): InternalAPI<NetscriptSleeve> {
       const action = helpers.string(ctx, "action", _action);
       checkSleeveAPIAccess(ctx);
       checkSleeveNumber(ctx, sleeveNumber);
-      let ctract = undefined;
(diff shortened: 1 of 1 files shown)


src/NetscriptFunctions/Sleeve.ts
```diff
@@ -238,8 +238,10 @@ export function NetscriptSleeve(): InternalAPI<NetscriptSleeve> {
       const action = helpers.string(ctx, "action", _action);
       checkSleeveAPIAccess(ctx);
       checkSleeveNumber(ctx, sleeveNumber);
+      let ctract = undefined;
```
d0sboots: Please change this to `contractName` and the other one to `contractEnum`. Other names would work too, as long as they are distinguishing between them and don't have ctractions. :D

Sphyxis:
@d0sboots 
Variable names changed.  Retested OK.

PR #1315, opened 2024-05-26
MISC:  Updated Grafting finish message to remove typo

Sphyxis:
When finishing Grafting you receive a completion message that changes based on if you have Congruity installed.
This has caused a bug report.

Removed the purposeful {" "} and everything looks normal now.

Tested grafting with and without Congruity.


d0sboots:
I'm assuming this was a #bug-reports bug, thus nothing to close?

Sphyxis:
@d0sboots 
Correct.  Nothing to close.

PR #1316, opened 2024-05-26
MISC: Change type of location parameter/property

catloversg:
We use `string` for the type of location/locationName in `NetscriptDefinitions.d.ts`. This PR changes it to `LocationName`.

### DIFF
--- a/src/ScriptEditor/NetscriptDefinitions.d.ts
+++ b/src/ScriptEditor/NetscriptDefinitions.d.ts
@@ -47,7 +47,7 @@ interface Player extends Person {
   jobs: Partial<Record<CompanyName, JobName>>;
   factions: string[];
   totalPlaytime: number;
-  location: string;
+  location: LocationName;
   karma: number;
 }
 
@@ -1611,7 +1611,7 @@ export interface StudyTask {
   type: "CLASS";
   cyclesWorked: number;
   classType: string;
-  location: string;
+  location: LocationName | `${LocationName}`;
 }
 /**
  * Company Work
@@ -2460,7 +2460,7 @@ export interface Singularity {
    * @param locationName - Name of the location.
    * @returns True if the player was moved there, false otherwise.
    */
-  goToLocation(locationName: string): boolean;
+  goToLocation(locationName: LocationName | `${LocationName}`): boolean;
 
   /**
    * Get the current server.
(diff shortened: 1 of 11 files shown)


src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -1611,7 +1611,7 @@ export interface StudyTask {
   type: "CLASS";
   cyclesWorked: number;
   classType: string;
-  location: string;
+  location: LocationName | `${LocationName}`;
```
d0sboots, line 1614: Did your search-and-replace go bad, or is this intentional?

catloversg: In `SleeveClassTask`, I see this:
```typescript
type SleeveClassTask = {
  type: "CLASS";
  classType: UniversityClassType | GymType | `${UniversityClassType}` | `${GymType}`;
  location: LocationName | `${LocationName}`;
};
```
Many `XXXTask` (`SleeveClassTask`, `SleeveCrimeTask`, `SleeveFactionTask`) use `property: Type | ${Type}` format, so I imitated them for `StudyTask`. I'm not sure why they are like that, though. It's pretty weird because `CompanyWorkTask`, `CrimeTask`, `FactionWorkTask` use `property: Type`.

I agree that `property: Type` looks like the proper way here. What's your suggestion?

d0sboots: OK, I finally figured out what's going on here, but I'm not 100% certain of the answer either. 

@Snarling added this syntax ~two years ago, during a big refactor to add more enum use. The actual syntax is "template literal types", a TS feature that (IMO) is rather obscure, and allows you to make complicated string unions. In this case, it is making a union of all the enum values. 

Where this gets weird is that the enum type itself is *already* a union of all the enum values. However, it is a union of them *as enum value types*, rather than as strings.

Further complicating this is that *the type definitions for these functions don't matter*. Internally, all the arguments are changed to `unknown`, and we (have to) do runtime type checking anyway. So (for the functions), the stuff in NetscriptDefinitions is only for documentation. 

Given all that, I think the right course is to rip out the confusing and semi-redundant template literal types. But that's a different refactor; I went ahead and merged this.

catloversg: I'll take note of this problem and make a PR for it later.

PR #1317, opened 2024-05-26
INFILTRATION: Change description of slash game

catloversg:
A player gives valid feedback for the description of the slash game. This PR makes these changes:
- `the guard` -> `the sentinel`. `when the guard drops his guard` feels clunky.
- Reduce the font size of the description and add a line break.

Before:
before

After:
after


d0sboots:
On one hand, I feel like this is a small change almost intentionally to break infil scripts. OTOH, I'm OK with that. XD

PR #1320, opened 2024-05-27
SLEEVES: Add missing faction check when installing through API

yichizhng:
Per title.

To verify the behavior, try the following:

```js
ns.sleeve.purchaseSleeveAug(0, `Stanek's Gift - Serenity`);
```


yichizhng:

PR #1321, opened 2024-05-27
MISC: Refactor favor code

catloversg:
When `favor` is greater than 34866, `favorToRep` returns `Infinity`. This problem creates many bugs and exploits. There are (at least) 3 solutions:
- Use `BigInt`: this is overkill.
- Change the formula: #976 uses this approach. This approach requires careful testing and balancing.
- Clamp the return value in `favorToRep` and `repToFavor`. We must be careful when performing arithmetic operations with this approach; otherwise, the bug will come back.

This PR uses the third approach.

Notable changes:
- Remove `Math.round(raw * 10000) / 10000` in `src\Faction\formulas\favor.ts`. IMO, this rounding code does not do anything useful. hydroflame also does not remember why he added it. `MaxFavor` increases from 34866 to 35331.
- I know that it's weird to create both `setFavor` and `gainFavor` instead of a normal setter. I want to make it clear that modifying `favor` directly is dangerous (e.g., `faction.favor = XXX`, `faction.favor++`, etc.).

### DIFF
--- a/src/Company/Company.ts
+++ b/src/Company/Company.ts
@@ -1,7 +1,8 @@
 import type { CompanyPosition } from "./CompanyPosition";
 
 import { CompanyName, JobName, FactionName } from "@enums";
-import { favorToRep, repToFavor } from "../Faction/formulas/favor";
+import { MaxFavor, calculateFavorAfterResetting } from "../Faction/formulas/favor";
+import { clampNumber } from "../utils/helpers/clampNumber";
 
 export interface CompanyCtorParams {
   name: CompanyName;
@@ -37,7 +38,8 @@ export class Company {
 
   // Dynamic info, loaded from save and updated during game.
   playerReputation = 0;
-  favor = 0;
+
+  #favor = 0;
 
   constructor(p: CompanyCtorParams) {
     this.name = p.name;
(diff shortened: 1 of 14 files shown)


src/Company/Company.ts
```diff
@@ -49,26 +51,38 @@ export class Company {
     if (p.relatedFaction) this.relatedFaction = p.relatedFaction;
   }
 
+  get favor() {
+    return this.#favor;
+  }
+
+  /**
```
d0sboots, line 58: Even though it's duplication, please copy the comments here.

src/Company/Company.ts
```diff
@@ -49,26 +51,38 @@ export class Company {
...
+   */
+  setFavor(value: number) {
+    if (Number.isNaN(value)) {
+      this.#favor = 0;
+      return;
+    }
+    this.#favor = clampNumber(value, 0, MaxFavor);
+  }
+
+  gainFavor(value: number) {
```
d0sboots: I was expecting `gainFavor` to do something more involved, but it's literally just calling `setFavor`. So why do we need it? Having a regular setter and doing `this.favor += value` would do the same thing.

src/Faction/Faction.ts
```diff
@@ -34,6 +35,37 @@ export class Faction {
...
+   * the overflow error of `playerReputation`, so anything that wants to change `favor` must explicitly do that through
+   * `setFavor` and `gainFavor`.
+   *
+   * Only use this function in these cases:
+   * - Initialize favor = 0.
+   * - Set value returned from calculateFavorAfterResetting function.
+   * - Load value from save data.
+   * - Dev menu.
+   *
+   * If you want to increase the favor, you must use the `gainFavor` function.
```
d0sboots: I see the justification here, but I don't agree with it. *Adding* favor is perfectly fine; the clamp in setFavor deals with that. The potential issue comes in if you are doing more complicated math that might divide infinity by infinity, leading to NaN. But those cases can't be handled with gainFavor anyway; they were always going to have to do calculations and then set the value. 

Basically, although I understand your motivation, I think you are only protecting the case that was already safe.

d0sboots (commented):
Although I have some quibbles, overall I really like this approach. This is a good use of private variables.

src/Faction/Faction.ts
d0sboots: Addendum: I can understand intentionally wanting to add some friction so that people don't assume this is a "normal" variable. For that, possibly not having a normal setter (keep it explicit), but drop gainFavor since it doesn't add anything?

PR #1322, opened 2024-05-28
UI: Reverse order of Black Operations list

catloversg:
This PR reverses the order of the Black Operations list. It removes the unnecessary scrolling.

Before:
before

After:
after

### DIFF
--- a/src/Bladeburner/ui/BlackOpPage.tsx
+++ b/src/Bladeburner/ui/BlackOpPage.tsx
@@ -14,7 +14,7 @@ interface BlackOpPageProps {
 }
 
 export function BlackOpPage({ bladeburner }: BlackOpPageProps): React.ReactElement {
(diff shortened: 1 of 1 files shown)


src/Bladeburner/ui/BlackOpPage.tsx
```diff
@@ -14,7 +14,7 @@ interface BlackOpPageProps {
 }
 
 export function BlackOpPage({ bladeburner }: BlackOpPageProps): React.ReactElement {
-  const blackOps = blackOpsArray.slice(0, bladeburner.numBlackOpsComplete + 1);
+  const blackOps = blackOpsArray.slice(0, bladeburner.numBlackOpsComplete + 1).sort((a, b) => b.n - a.n);
```
d0sboots: Can't you just use `reverse()`? If not, this probably deserves a comment.

PR #1324, opened 2024-05-29
BUGFIX: Fix manual hack exploit

catloversg:
When a player hacks a server manually via the terminal, their gain of intelligence exp is based on the gain of hacking exp. This is an oversight in the balance of intelligence gain. It's very easy to gain hacking exp, but intelligence exp is supposed to be incredibly hard to gain. This PR removes the gain of intelligence exp in this case. The reasons are:
- Manually hacking gains intelligence exp, but manually growing/weakening does not.
- The gain rate is absurd. With money from Corporation, I can farm intelligence from 0 to ~750 in ~1-2 hours and 0 to ~800 in < 10 hours.

We can nerf the gain by using a logarithm function, but it's very hard to create a function that generates meaningful exp at a low hacking exp multiplier (it's worth hacking manually to gain intelligence exp) while avoiding exploits at a high hacking exp multiplier.

### DIFF
--- a/src/Constants.ts
+++ b/src/Constants.ts
@@ -24,7 +24,6 @@ export const CONSTANTS: {
   IntelligenceCrimeBaseExpGain: number;
   IntelligenceProgramBaseExpGain: number;
   IntelligenceGraftBaseExpGain: number;
-  IntelligenceTerminalHackBaseExpGain: number;
   IntelligenceSingFnBaseExpGain: number;
   MillisecondsPer20Hours: number;
   GameCyclesPer20Hours: number;
@@ -101,7 +100,6 @@ export const CONSTANTS: {
   IntelligenceCrimeBaseExpGain: 0.05,
   IntelligenceProgramBaseExpGain: 0.1, // Program required hack level divided by this to determine int exp gain
   IntelligenceGraftBaseExpGain: 0.05,
-  IntelligenceTerminalHackBaseExpGain: 200, // Hacking exp divided by this to determine int exp gain
   IntelligenceSingFnBaseExpGain: 1.5,
 
   // Time-related constants
--- a/src/Terminal/Terminal.ts
+++ b/src/Terminal/Terminal.ts
@@ -225,6 +225,9 @@ export class Terminal {
       server.moneyAvailable -= moneyGained;
       Player.gainMoney(moneyGained, "hacking");
       Player.gainHackingExp(expGainedOnSuccess);
+      if (expGainedOnSuccess > 1) {
+        Player.gainIntelligenceExp(Math.log10(expGainedOnSuccess));
+      }
 
       const oldSec = server.hackDifficulty;
       server.fortify(ServerConstants.ServerFortifyAmount);
--- a/src/Terminal/Terminal.ts
+++ b/src/Terminal/Terminal.ts
@@ -226,7 +226,7 @@ export class Terminal {
       Player.gainMoney(moneyGained, "hacking");
       Player.gainHackingExp(expGainedOnSuccess);
       if (expGainedOnSuccess > 1) {
-        Player.gainIntelligenceExp(Math.log10(expGainedOnSuccess));
+        Player.gainIntelligenceExp(4 * Math.log10(expGainedOnSuccess));
       }
 
       const oldSec = server.hackDifficulty;
--- a/src/Terminal/Terminal.ts
+++ b/src/Terminal/Terminal.ts
@@ -211,7 +211,7 @@ export class Terminal {
         Router.toPage(Page.BitVerse, { flume: false, quick: false });
         return;
       }
-      // Manunally check for faction invites
+      // Manually check for faction invitations
       Engine.Counters.checkFactionInvitations = 0;
       Engine.checkCounters();
 
@@ -225,7 +225,6 @@ export class Terminal {
       server.moneyAvailable -= moneyGained;
       Player.gainMoney(moneyGained, "hacking");
       Player.gainHackingExp(expGainedOnSuccess);
-      Player.gainIntelligenceExp(expGainedOnSuccess / CONSTANTS.IntelligenceTerminalHackBaseExpGain);
 
       const oldSec = server.hackDifficulty;
       server.fortify(ServerConstants.ServerFortifyAmount);


d0sboots:
I'm not 100% onboard with just removing this yet; I'd like to explore balancing it more. Part of this comes from the fact that IIUC this is *strictly* from manual hacking: in other words, you can't automate it to get large int gains without getting into exploity territory already. 

I feel like a reasonably tuned log function of hack xp would still give noticeable gains for early int players, while leaving this only about as effective as a travel loop when boosted. It's OK if it's a bit more efficient, as long as it's not the most efficient thing possible, because it requires being willing to exploit.

Caldwell-74:
~~there is a singularity.manualHack (artifact from pre backdoor times) that should give that int aswell?~~

nope it does a tiny flat amount
https://github.com/bitburner-official/bitburner-src/blob/54d099e552102d51d0b90baf0ed164a8b8e06a0e/src/Netscript/NetscriptHelpers.tsx#L467

d0sboots:
Yeah you could make a batcher-like with manualHack, but I think travel loop is already better?

gmcew:
> you can't automate it to get large int gains without getting into exploity territory already.

I feel that's a big part of the balancing issues around this. High BN12 (where all bets are off anyway) or use of the company favor overflow (fixed now, thank you!) give a silly boost to hacking exp and hence huge int gains with this method, but since they break so much of the game wide open I don't think int farming is the biggest problem with those scenarios.

For what it's worth, given the recent tweaks to bladeburner skills capping out and slowing the "traditional" int farming approach at higher levels, I'd be a fan of keeping this (with the log scaling option or similar cap to int rates) to try and spice up int farming diversity, particularly at lower levels.

d0sboots:
For the record, can you give some example numbers (hopefully ~representative situations) for early-game/mid-game/late BN12, to make it easier to evaluate this change? (And/or for future people to understand our reasoning.)

catloversg:
### Overview
There are multiple ways to gain Intelligence exp:
- Hacking manually via the terminal.
- Calling singularity functions.
  - `ns.singularity.manualHack`: 0.005 exp.
  - `ns.singularity.purchaseAugmentation`: 15 exp.
  - `ns.singularity.installAugmentations`: 15 exp.
  - `ns.singularity.goToLocation`: 0.00003 exp.
  - `ns.singularity.travelToCity`: 0.00003 exp.
  - `ns.singularity.purchaseTor`: 0.003 exp.
  - `ns.singularity.purchaseProgram`: 0.0003 exp.
  - `ns.singularity.upgradeHomeCores`: 3 exp.
  - `ns.singularity.upgradeHomeRam`: 15 exp.
  - `ns.singularity.joinFaction`: 7.5 exp.
- Doing Bladeburner actions (Contract, Operation, Black Operation).
- Creating program: 0.02 per `cyclesWorked`.
- Committing crime.
  - robStore: 0.375 exp.
  - larceny: 0.75 exp.
  - bondForgery: 3 exp.
  - grandTheftAuto: 0.8 exp.
  - kidnap: 1.3 exp.
  - assassination: 3.25 exp.
  - heist: 6.5 exp.
- Grafting: 0.001 per `cyclesWorked`.
- Studying: at ZB: [0.008 * `hashManager.getStudyMult()`] per cycle.
- Prestiging SF: 300 exp.

Not all of them are good, so I will give a review of the most common and viable ways. All numbers are tested on my machine. Your mileage may vary.

### Review
Singularity functions: All these ways are simple. 
`ns.singularity.travelToCity`: ~40 exp/second.
`ns.singularity.purchaseProgram` + `ns.rm`: ~195 exp/second.
`ns.singularity.joinFaction` + `ns.singularity.softReset`: 82.5 exp (11 factions)/soft reset. Each soft reset takes ~500-600ms.

Studying: Without Bladeburner, this is the best way to farm Intelligence.
How to do it:
- Buy and upgrade all hash servers. Make sure that your hash production rate is high enough to fill up the hash capacity in only **one** game cycle.
- Buy `Improve Studying` as much as possible every game cycle.
- De-shock your sleeves to at least 90-95.
- Go to Volhaven and take a course at `ZB Institute of Technology` (all courses give the same amount of Intelligence exp). Set the same task for all sleeves.

Test results:
- 1 hour run: 274 Int. `hashManager.getStudyMult()`: ~14900.
- 2 hour run: 309 Int. `hashManager.getStudyMult()`: ~21000.
- 3 hour run: 329 Int. `hashManager.getStudyMult()`: ~25100.

Doing Bladeburner actions: This is the best way to farm Intelligence. The exp gain rate scales with the multiplier from `Hyperdrive` and the level of the action if it's a `LevelableActionClass` (Contract or Operation) (Thanks @Caldwell-74 for the information about the level of the action). This is also the only viable way to farm Intelligence when it's higher than ~500.

### Balance
My opinion about the balancing:
- Hacking manually only gains Intelligence exp since e3991b8795fc9bccd38240e17c99e6d20d53fd4f. That's an ancient commit (15 September 2017), so it's clear that the (massive) amount of exp is **not** intentional.
- Using DOM tricks to automate the manual hack is acceptable. My testing script can do it consistently every ~200ms.
- In order to properly balance the exp gain rate of hacking manually, we need to make it:
  - **Not** ridiculously low.
  - High enough to make it worth the effort. It should be at least as good as the "travel loop".
  - Cannot be exploited even with a massive amount of NFG (deep BN12).

My testing "environment": BN1, All SF, 1830 NFG (from 1e110/s of Corporation), `ns.share()`, all augmentations (except "Easter Egg" augmentations), nonoptimal Stanek's Gift layout. Hacking exp per manual hack is ~1e11.

With the current formula:
- Extremely low hacking exp (e.g., 1e3): 12 exp/hack. 1 heist (10 minutes) only gives 6.5 exp.
- ~1e11 hacking exp: 44 exp/hack. It's roughly the same as 1 second of the "travel loop".
- ~1e11 hacking exp + DOM tricks: 220 exp/second. It's roughly the same as the "purchaseProgram loop".
- 1.7976931348623157E+308 hacking exp (Number.MAX_VALUE): 6165 exp/second. It's only slightly better than studying at ZB with sleeves.

Caldwell-74:
> Doing Bladeburner actions: This is the best way to farm Intelligence. The exp gain rate scales with the multiplier from Hyperdrive. This is also the only viable way to farm Intelligence when it's higher than ~500.

Just for completionist sake:
Hyperdrive is not the only factor for int exp gain from Bladeburner, the Level of the Assassination(or any other levelable and int giving task) has a big effect on it

d0sboots:
Thanks for that detailed writeup! It gives me confidence that this is a) Roughly in line with the other "easy" methods of XP, and b) won't be changing the meta, which I think are the two major goals here.

PR #1327, opened 2024-05-30
IPVGO: Fix non-async promise-returning methods to be more consistent

ficocelliguy:
Based on this conversation: https://discord.com/channels/415207508303544321/1242866251894882396/1245730041565024334

* Set all ns.go methods returning promises to be async in the netscript definitions and in their implementations
* Set determineCheatSuccess to always return a promise, and end the game w/ logging instead of just starting a new game
* Set makeAIMove to be async and be awaited in places where it is called


Sphyxis:
@d0sboots @ficocelliguy 
I've forked this and run it through ~200 fights for each faction.  The issues that I saw previously did not reappear.

src/Go/boardAnalysis/goAI.ts
```diff
@@ -26,7 +26,7 @@ let currentAITurn: Promise<Play> | null = null;
 /**
  * Retrieves a move from the current faction in response to the player's move
  */
-export function makeAIMove(boardState: BoardState): Promise<Play> {
+export async function makeAIMove(boardState: BoardState): Promise<Play> {
```
d0sboots: I don't think this needs to be, or should be, async. This will be a common theme - just because something returns (or handles) promises, doesn't mean it should be async.

src/Go/boardAnalysis/goAI.ts
```diff
@@ -38,7 +38,7 @@ export function makeAIMove(boardState: BoardState): Promise<Play> {
         passTurn(boardState, GoColor.white);
         // if passTurn called endGoGame, or the player has no valid moves left, the move should be shown as a game over
         if (boardState.previousPlayer === null || !getAllValidMoves(boardState, GoColor.black).length) {
-          return { type: GoPlayType.gameOver, x: null, y: null };
+          return Promise.resolve({ type: GoPlayType.gameOver, x: null, y: null });
```
d0sboots: Promise.resolve is unneeded here - the inner closure here is `asyc`, so all return values are wrapped in a promise if they are not already promises.

src/Go/effects/netscriptGoImplementation.ts
```diff
@@ -337,28 +337,25 @@ export async function determineCheatSuccess(
...
+
   // If cheat is successful, run callback
   if ((successRngOverride ?? rng.random()) <= cheatSuccessChance(state.cheatCount)) {
     callback();
-    state.cheatCount++;
     GoEvents.emit();
     return makeAIMove(state);
   }
   // If there have been prior cheat attempts, and the cheat fails, there is a 10% chance of instantly losing
   else if (state.cheatCount && (ejectRngOverride ?? rng.random()) < 0.1) {
```
d0sboots, line 348: The `state.cheatCount` check here will now always be true. Feels like a bug?

src/Go/effects/netscriptGoImplementation.ts
```diff
@@ -389,7 +386,7 @@ export function cheatSuccessChance(cheatCount: number) {
 /**
  * Attempts to remove an existing router from the board. Can fail. If failed, can immediately end the game
  */
-export function cheatRemoveRouter(
+export async function cheatRemoveRouter(
```
d0sboots: None of these cheat functions should be async. Also, they're all missing a return type.

src/NetscriptFunctions/Go.ts
```diff
@@ -41,17 +41,17 @@ export function NetscriptGo(): InternalAPI<NSGo> {
   return {
     makeMove:
       (ctx: NetscriptContext) =>
-      (_x, _y): Promise<Play> => {
+      async (_x, _y): Promise<Play> => {
```
d0sboots: None of these should be async either.

d0sboots (commented):
I think the actual relevant bugfixes here are only 2 or 3 lines XD

src/NetscriptFunctions/Go.ts
ficocelliguy: weaken is async in this way https://github.com/bitburner-official/bitburner-src/blob/dev/src/NetscriptFunctions.ts#L353 m but not hack or grow for some reason

src/Go/effects/netscriptGoImplementation.ts
ficocelliguy: good catch, yeah

src/NetscriptFunctions/Go.ts
d0sboots: weaken probably shouldn't be either. *Or* it could be rewritten to use `await`, which would flow a bit better.

PR #1328, opened 2024-05-30
UI: change time elapsed to time remaining for programs and grafting

TheAimMan:
Current Program/Grafting behavior is to show how long you've been doing the action which isn't useful

image

I calculated the current rate of which the completion of the workTask was being done and added it to the object, and then used that to predict the time remaining for the task

image

Its accurate at the point of consumption, but since other factors can effect how long it takes (lack of focus, more hacking skill, more int) I'm wondering if I need to add "estimated" to the dialog or not.

### DIFF
--- a/src/ui/WorkInProgressRoot.tsx
+++ b/src/ui/WorkInProgressRoot.tsx
@@ -286,7 +286,7 @@ export function WorkInProgressRoot(): React.ReactElement {
   if (isCreateProgramWork(Player.currentWork)) {
     const create = Player.currentWork;
     const completion = (create.unitCompleted / create.unitNeeded()) * 100;
-    const rate = ((create.unitNeeded() - create.unitCompleted) / create.unitRate) * CONSTANTS.MilliPerCycle;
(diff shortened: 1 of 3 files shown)


src/ui/WorkInProgressRoot.tsx
```diff
@@ -286,7 +286,7 @@ export function WorkInProgressRoot(): React.ReactElement {
   if (isCreateProgramWork(Player.currentWork)) {
     const create = Player.currentWork;
     const completion = (create.unitCompleted / create.unitNeeded()) * 100;
-
+    const rate = ((create.unitNeeded() - create.unitCompleted) / create.unitRate) * CONSTANTS.MilliPerCycle;
```
d0sboots: These should be called "remaining", since they're times not rates.

d0sboots (commented):
Sorry for being slow. I really wanted to just merge this, but I am too picky XD

src/ui/WorkInProgressRoot.tsx
TheAimMan: changed to `remainingTime`

PR #1332, opened 2024-06-01
EDITOR: useVimEditor uses Material UI

G4mingJon4s:
The `monaco-vim` package the game uses to have a vim binding of the editor creates a status bar with plain HTML tags.
However, the game renders everything using Material UI, so the status bar looks weird and doesn't use any theme.

While the package doesn't support a direct change of the tags used, overriding the StatusBarClass used allows for handling the status bar outside of the package.

Since the package doesn't have any types, it took me some time to figure out how the package uses the class. I tried to document as much as possible for future contributions.

The current implementation requires the package to not use the second and fourth parameter of `initVimMode` in any other way, since these are no longer what the package intends them to be:
- Instead of `node` being an `HTMLElement` it is used to pass a `React.MutableRefObject` down to get a hold of the spot, where the status bar should be displayed.
- I couldn't figure out what `sanitizer` is intended to do, but since it is only used in the class, I re-purposed it to be the `rerender` hook to update the ScriptEditor, when the status bar changes.

### DIFF
--- a/src/ScriptEditor/ui/StatusBar.tsx
+++ b/src/ScriptEditor/ui/StatusBar.tsx
@@ -1,0 +1,226 @@
+import { Input, Typography } from "@mui/material";
+import { styled } from "@mui/styles";
+import type { editor } from "monaco-editor";
+import React from "react";
+type IStandaloneCodeEditor = editor.IStandaloneCodeEditor;
+
+const StatusBarContainer = styled("div")({
+  display: "flex",
+  flexDirection: "row",
+  alignItems: "center",
+  justifyContent: "space-between",
+  height: 36,
+  marginLeft: 4,
+  marginRight: 4,
+});
+
+const StatusBarLeft = styled("div")({
+  display: "flex",
+  flexDirection: "row",
+  alignItems: "center",
+});
+
+// This Class is injected into the monaco-vim initVimMode function to override the status bar.
+export class StatusBar {
+  // modeInfoNode is used to display the mode in the status bar.
+  // notifNode is used to display notifications in the status bar.
+  // secInfoNode is used to display the input box in the status bar.
+  // keyInfoNode is used to display the operator and count in the status bar.
+  // editor is kept to focus when closing the input box.
+  // sanitizer is weird.
+
+  node: React.MutableRefObject<React.ReactElement | null>;
+  editor: IStandaloneCodeEditor;
+
+  mode = "--NORMAL--";
+  showInput = false;
+  inputValue = "";
+  buffer = "";
+
+  rerender: () => void;
+
+  onCloseHandler: ((query: string) => void) | null;
+  onKeyDownHandler: ((e: KeyboardEvent, query: string, close: () => void) => void) | null;
+  onKeyUpHandler: ((e: KeyboardEvent, query: string, close: () => void) => void) | null;
+
+  // node is used to setup the status bar. However, we use it to forward the resulting status bar to the outside.
+  // sanitizer is weird, so we use it to forward a rerender hook.
+  constructor(
+    node: React.MutableRefObject<React.ReactElement | null>,
+    editor: IStandaloneCodeEditor,
+    rerender: () => void,
+  ) {
+    this.node = node;
+    this.editor = editor;
+
+    this.rerender = rerender;
+
+    this.onCloseHandler = null;
+    this.onKeyDownHandler = null;
+    this.onKeyUpHandler = null;
+
+    this.update();
+  }
+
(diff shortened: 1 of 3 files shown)


src/ScriptEditor/ui/StatusBar.tsx
```diff
@@ -0,0 +1,226 @@
...
+    return;
+  }
+
+  update = () => {
+    this.node.current = this.StatusBar();
+    this.rerender();
+  };
+
+  keyUp = (e: React.KeyboardEvent) => {
+    const event = new KeyboardEvent("keyup", {
```
d0sboots: Can you just do `{...e}`? Do you need to filter out these specific fields?

src/ScriptEditor/ui/StatusBar.tsx
```diff
@@ -0,0 +1,226 @@
...
+      repeat: e.repeat,
+      bubbles: e.bubbles,
+      cancelable: e.cancelable,
+    });
+    if (this.onKeyDownHandler !== null) {
+      this.onKeyDownHandler(event, this.inputValue, this.closeInput);
+    }
+
+    // this handles pressing escape in the input box.
+    if (e.keyCode === 27) {
```
d0sboots: These should be checking key instead of keyCode.

d0sboots (commented):
I'll trust you to have thorough tested that the vim editor still works.

src/ScriptEditor/ui/StatusBar.tsx
G4mingJon4s: Sadly, I haven't found another way to port the `React.KeyboardEvent` to `KeyboardEvent`. There are over 200 entries that aren't on `KeyboardEvent`, so I filtered out *all* of the necessary properties the package uses.

G4mingJon4s:
I haven't removed or changed any functionality the methods of the class had before, so it **should** work.
I tested for some edge-cases that could be problematic, like leaving the editor while typing a command, switching tabs, etc., but I can't guarantee that I have tested for every possibility.

There is only one issue I have found, but that one is not related to the UI implemented:
Adding a count before entering insert mode will duplicate anything typed that many times as expected. However, for a large count, this will crash the game for obvious reasons. Even vim itself crashes for these inputs, so it isn't really game-specific either.

Caldwell-74:
oh seems like reviews need perms before they get "published"
wanted to ask if you can move right away from @mui/styles to @mui/material/styles or @mui/system
https://mui.com/system/styled/
this way we dont add more onto the migration stack

src/ScriptEditor/ui/StatusBar.tsx
Snarling: Why are you using KeyboardEvent at all instead of just keeping it as a React.KeyboardEvent?

Snarling: No errors with these changes, but not sure if there's a reason it has to stay as a KeyboardEvent instead of React.KeyboardEvent

image

G4mingJon4s: Well, the package uses the simple `KeyboardEvent`. I haven't tested simply passing a `React.KeyboardEvent`. Maybe `React.KeyboardEvent` is based on `KeyboardEvent`? If it works, it works.
Looking at the source code of the package, the only "weird" properties it uses are `e.target.selectionStart` and `e.target.selectionEnd`. Others, like `e.target.value` or `e.key`, are probably on `React.KeyboardEvent`.

d0sboots: Similar to Snarling, I don't see/understand why it would be an issue to have extra properties.

G4mingJon4s: I'll try out just passing the event and test for any weird behaviors.

G4mingJon4s: Good thing we discussed this: The current implementation as well as the one with just passing the event throws error for one specific thing, the `q:` operator, or whatever it is called. When pressing this combination, in the current deployment, a text field opens and no errors are thrown when trying to navigate previous commands. However, nothing happens. No idea why or how it is intended to look, but it just doesn't work.
I'll push the changes necessary to avoid the errors for now, but they won't fix this command not working.

G4mingJon4s:
I changed the handlers to just pass down the `React.KeyboardEvent`.
Since the game never supported vim macros and they were causing errors, not just in the PR but also in the current deployment, I re-mapped the macro operators `q` and `@` in normal mode to translate to an empty string. Any input in the key buffer will essentially remove these keys when executing (at least that's what I have observed).

PR #1336, opened 2024-06-01
EDITOR: Fixed vim mode not being set when using the respective command

G4mingJon4s:
### Issue
When opening a script with `nano` or `vim`, the editor setting, whether to use vim, is not overwritten. This means that the initial value, `false` is kept and going to another page and coming back will disable vim. As a side-effect, if the value is `true`, opening a script using `nano` will still enable vim.

### Intended behavior
- Opening a script using `nano` should disable vim mode.
- Opening a script using `vim` should enable vim mode.
- The previously used mode should be kept and only changed when
- - Switching the mode in the editor settings
- - Opening a script with a different command
- Opening a script using `ls` should have no opinion on the mode used

### Changes made
- Removed the `vim` editor option passed between the ScriptEditorRoot and the pages.
- Changed the editor commands to set the mode being used on command to the settings of the editor
- Changed the `ls` command to open the editor similar to the `nano` and `vim` command, ignoring any mode preferences and using the current setting

### DIFF
--- a/src/ScriptEditor/ui/ScriptEditorContext.tsx
+++ b/src/ScriptEditor/ui/ScriptEditorContext.tsx
@@ -80,8 +80,6 @@ export function ScriptEditorContextProvider({ children }: { children: React.Reac
     cursorBlinking: Settings.MonacoCursorBlinking,
   });
 
-  console.log("Context Vim", options.vim, Settings.MonacoVim);
-
   function saveOptions(options: Options) {
     console.log("Saving Vim", options.vim, Settings.MonacoVim);
     setOptions(options);
--- a/src/ScriptEditor/ui/ScriptEditorContext.tsx
+++ b/src/ScriptEditor/ui/ScriptEditorContext.tsx
@@ -24,7 +24,7 @@ export interface ScriptEditorContextShape {
 
 const ScriptEditorContext = React.createContext({} as ScriptEditorContextShape);
 
-export function ScriptEditorContextProvider({ children, vim }: { children: React.ReactNode; vim: boolean }) {
+export function ScriptEditorContextProvider({ children }: { children: React.ReactNode }) {
   const [ram, setRAM] = useState("RAM: ???");
   const [ramEntries, setRamEntries] = useState<string[][]>([["???", ""]]);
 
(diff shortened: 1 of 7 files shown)


src/ScriptEditor/ui/ScriptEditorContext.tsx
```diff
@@ -75,12 +75,15 @@ export function ScriptEditorContextProvider({ children, vim }: { children: React
     fontSize: Settings.MonacoFontSize,
     fontLigatures: Settings.MonacoFontLigatures,
     wordWrap: Settings.MonacoWordWrap,
-    vim: vim || Settings.MonacoVim,
+    vim: Settings.MonacoVim,
     cursorStyle: Settings.MonacoCursorStyle,
     cursorBlinking: Settings.MonacoCursorBlinking,
   });
 
+  console.log("Context Vim", options.vim, Settings.MonacoVim);
```
d0sboots: This looks like debug logging that you forgot to remove.

d0sboots (commented):
I'm not sure that I agree with this change. Generally opening something with a specific editor should not change your default editor; things that have acted this way in the past (think browsers taking over default browser preference) are usually considered annoying.

Shouldn't it just be the case that you can change the default in the settings, and it is respected if you open something with ls?

G4mingJon4s:
The problem is that the game has no setting for a default editor. The `Settings` object does have the `Settings.MonacoVim` option, but that one is not the default, but the current editor being used. Furthermore, the value is not even stored in the save-file, so the game starts out with vim disabled.

Currently, using `vim` will open the editor with vim enabled, but switching pages will reset this.
To me, `nano` and `vim` should open the editor with their respective mode and preserve this mode until it is changed.
I understand that there should be one default the player can set that is respected. Though `nano` and `vim` do different things in my opinion.

One thing I think we can agree on is that switching pages should not change the current mode of the editor, regardless of it being the default or not. The question is if the two different commands for opening a file should affect which mode is being used.

If the commands shouldn't affect the mode, I would say that the editor mode should be saved in the save-file, so the player doesn't have to re-set their preference every time.

Snarling:
Here's how I think it should work:

Player types command `vim test.js && nano test2.js`

test.js should open in vim mode regardless of the player's config, and test2.js should open in nano mode regardless of player's config

Navigating between different pages should not change what mode the script is open in - test.js should stay open in vim mode, and test2.js should remain open in nano mode.

Implementating the above would totally remove the need for Settings.monacoVim. Right now that setting seems to do two things: it makes the nano command function as vim (????? just type vim if that's what you want) and then it also affects the mode that all scripts are open in when the player switches pages.

If each OpenScript can get a new flag for whether it's in vim mode, this would replace the global config.

Edit: maybe keeping a global option is worthwhile just for when scripts are open via clickable `ls` links

G4mingJon4s:
Unless there are any other suggestions on how to deal with this, I'll start to implement each tab having its own mode.
I'm not quite sure what to do with `ls`. Maybe an option in the settings for `ls` specifically would make sense?

d0sboots:
I agree with Snarling (I assumed that's how it already was, having to tried to switch between modes). And then the setting would be strictly for clicking links with ls, and/or other applications where "open with default editor" occurs (in the future?)
IMO it would be best for that to be an explicit setting, but I'd be OK with it changing based on most-recently-used.

G4mingJon4s:
I made a new PR that implements the changes we discussed here, see #1372.

PR #1338, opened 2024-06-01
IMPROVEMENT: partial migration @mui/styles to tss-react

Caldwell-74:
its well known that @mui/styles is deprecated and stops us from updating other (core) packages like react

tss-react is a type safe package that implements a replacement for makeStyle that uses nearly the exact same syntax
its well maintained (>24k commits, closed >17k issues from ~19k, merged >22k PRs)
with a goal to stay seamlessly integrateable with mui
https://github.com/garronej/tss-react

i changed ~~all~~ most occurences of makeStyle, createStyle, withStyle

one problem that occured was an import error thats why i moved the logBoxBaseZIndex into its own constant file to resolve it

~~another (not yet fixed) problem is that the IPvGo board is broken, here i hope to get help from Ficocelli the creator of IPvGo to resolve this aswell~~

the nested style structure of IPvGO makes it harder to migrate away from it thats why i reverted those changes 
the package allows for nested structures but my understanding of the subject is not deep enough to migrate them correctly

so far i havent removed @mui/styles nor upgraded any packages

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -43,7 +43,8 @@
         "react-resizable": "^3.0.5",
         "react-syntax-highlighter": "^15.5.0",
         "remark-gfm": "^3.0.1",
-        "sprintf-js": "^1.1.3"
+        "sprintf-js": "^1.1.3",
+        "tss-react": "^4.9.10"
       },
       "devDependencies": {
         "@babel/core": "^7.23.0",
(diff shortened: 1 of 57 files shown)


Caldwell-74:
sorry for the messy commits 
rebasing/merge conflicts did some wierd stuff and already reversed stuff reappeared

PR #1340, opened 2024-06-02
BLADEBURNER: Remove percentage-based BB operation population changes

Kelenius:
Several of the BB operations currently, on success, reduce the synthoid population in the city by a percentage. This is a problem for two reasons, one mechanical and one fluff:

* Mechanically, these operations can very quickly drop city population, making them practically unusable. This is especially notable in contrast with Assassination/Bounty Hunter/Retirement, which all kill >one< synthoid.
* Fluff-wise, the synthoid population is typically somewhere in the billions, so having an operation that kills 1% means the character has killed tens of millions of them, which is rather _bizarre_.

This PR removes all instances of operations doing that and replaces them with flat numbers that I think make sense for the operation as described:

* Sting: 0.1% -> 5-10
* Raid success: 1% -> 50-100
* Raid failure: 1%-0.5% -> 10-40
* Stealth retirement: 0.5% -> 10-20

This is a buff to BB, because it makes these operations less damaging.

### DIFF
--- a/src/Bladeburner/Bladeburner.ts
+++ b/src/Bladeburner/Bladeburner.ts
@@ -788,35 +788,26 @@ export class Bladeburner {
         break;
       case BladeOperationName.sting:
         if (success) {
-          city.changePopulationByPercentage(-0.1, {
-            changeEstEqually: true,
-            nonZero: true,
-          });
+          const change = getRandomIntInclusive(-10, -5);
+          city.changePopulationByCount(change, { estChange: change, estOffset: 0 });
         }
         city.changeChaosByCount(0.1);
         break;
       case BladeOperationName.raid:
         if (success) {
-          city.changePopulationByPercentage(-1, {
-            changeEstEqually: true,
-            nonZero: true,
-          });
+          const change = getRandomIntInclusive(-100, -50);
+          city.changePopulationByCount(change, { estChange: change, estOffset: 0 });
           --city.comms;
         } else {
-          const change = getRandomIntInclusive(-10, -5) / 10;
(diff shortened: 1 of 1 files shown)


Sphyxis:
If this is going to happen, I'd suggest removing the synthoid community destruction chance as well.  With 0 communities some actions are unavailable.

Kelenius:
Communities are a bit different because they only exist to be destroyed (you need one to do a raid, and a successful raid always destroys one community). It would maybe make sense to remove that, because raids are already limited by the number of contracts, but then we should remove communities altogether. It'd need to be considered separately I think.

d0sboots:
I know nothing about BladeBurner, so I'm not qualified to evaluate this for balance.

Alpheus:
I'm in the process of test-covering portions of bb (notably sleeves, cities, terminal) and will gladly take it up to create some and simulate some data on this to see what the consequences could be.

Alpheus:
Bump, I'm test covering this this week.

I'm in the middle of int farming right now and Raid seems like such a NO-NO both for normal gameplay and endgame.

I'm curious what you two think regarding making Raids more impactful (balance change):
- you can run them without communities present (in which case it only impacts pop)
- it destroys 1-2 communities when possible, boosting the amount of XP and reward it provides to levels greater than assassination
- (optional) Incite Violence should have a chance of destroying comms too
- (optional) Diplomacy should have a chance of creating a community (bigger chance with more synth pop)

This change would encourage treating communities more like a resource that you'd prefer not to deplete, rather than avoiding alltogether.

PR #1343, opened 2024-06-03
TERMINAL: Added grep

muesli4brekkies:
A simple implementation of grep.

There's a few ideas I've thought to add, to make it more grep-like, such as;

* ~regexp support (should be pretty simple),~ :white_check_mark: 
* ~`-n` to disable line-numbers,~ :white_check_mark: 
* ~`-B/A/C n` to provide context,~ :white_check_mark: 

But I thought I'd make the PR and get the ball rolling.

:)


d0sboots:
Nice!
My main feedback would be, I'd suggest to support regex now, and try to make the output match "real" `grep` as much as possible. IMO it's more important for the existing features to work correctly, than to support lots of optional flags (possibly slightly weirdly.)

muesli4brekkies:
Don't you know, I'm making my own POSIX standard, in BitBurner! BOSIX!

Nah you're absolutely right :grin: . I'll give those changes a bash ASAP.


What are your thoughts on the globbing behaviour with no file args passed in? Implementing `/*` syntax should be simple enough,, but is it overkill for the game? 

I think the default "search everything" is a useful feature, if not particularly compliant.

PS - are ANSI codes OK for the colours, or should I be using styles? All I need are the terminal primaries anyway.

d0sboots:
> What are your thoughts on the globbing behaviour with no file args passed in? Implementing `/*` syntax should be simple enough,, but is it overkill for the game?

Hmmm that's a good question. Although it would be technically possible to still use stdin, that's not really a thing with our shell, and we don't have pipelines or redirections. So defaulting to ./* probably makes the most sense.

> PS - are ANSI codes OK for the colours, or should I be using styles? All I need are the terminal primaries anyway.

ANSI codes are fine, as long as the layer you're outputting to renders them :D

src/Terminal/HelpText.ts
```diff
@@ -230,6 +231,24 @@ export const HelpTexts: Record<string, string[]> = {
...
     " ",
   ],
+  grep: [
+    "Usage: grep [OPTION]... PATTERN [FILE]...",
+    " ",
+    "Search for PATTERN in each FILE and print results to terminal.",
+    "Example: grep -n -H 'hello world' file1.js file2.txt",
+    " ",
+    "Pattern selection and interpretation:",
+    "  -G, --basic-regexp   PATTERN is basic regular expression",
```
d0sboots: Somewhere in this help it should explain what PATTERN is if you *don't* pass `-G`.

src/Terminal/HelpText.ts
```diff
@@ -230,6 +231,24 @@ export const HelpTexts: Record<string, string[]> = {
...
+    " ",
+    "Output control:",
+    "  -H --with-filename   print file name with output lines",
+    "  -h --no-filename     do not print file name with output lines. Overrides -H",
```
d0sboots: Should mention that this is the default (I assume it is)

src/Terminal/commands/grep.ts
```diff
@@ -0,0 +1,179 @@
...
+type FileParser = (file: ContentFile | null) => string[] | string;
+type FileTuple = [ContentFilePath, ContentFile];
+
+interface OkArgs {
+  lineNum: string[];
+  regExpr: string[];
+  yesName: string[];
+  notName: string[];
+  toFile: string[];
+  overWrt: string[];
```
d0sboots: Don't abbrvt :D
```suggestion
  overwrite: string[];
```

src/Terminal/commands/grep.ts
```diff
@@ -0,0 +1,179 @@
...
+
+  multiFile: boolean;
+}
+
+const OK_ARGS: OkArgs = {
+  lineNum: ["-n", "--line-number"],
+  regExpr: ["-G", "--basic-regexp"],
+  yesName: ["-H", "--with-filename"],
+  notName: ["-h", "--no-filename"],
+  toFile: ["-O", "--output-number"],
```
d0sboots: You've got extra options in here that aren't in the help yet.

src/Terminal/commands/grep.ts
```diff
@@ -0,0 +1,179 @@
...
+  overWrt: ["-f", "--allow-overwite"],
+  quiet: ["-q", "--quiet", "--silent"],
+};
+
+function getParseFunc(pattern: string | RegExp, options: Options, filename: string, line: string, i: number): string {
+  const RED: string = "\x1b[31m";
+  const DEF: string = "\x1b[0m"; // default
```
d0sboots: I would call this `NORMAL`, or `PLAIN` if you want slightly shorter.

src/Terminal/commands/grep.ts
```diff
@@ -0,0 +1,179 @@
...
+  const lineNo: string = options.lineNum ? `${GREEN}${i + 1}${CYAN}:${DEF}` : "";
+  const prefix: string = name + lineNo;
+
+  return prefix + editedLine;
+}
+
+function parseFile(parseFunc: PtLineParser): FileParser {
+  return function (file: ContentFile | null): string | string[] {
+    if (!file) return "";
+    const content: string = "content" in file ? file["content"] : file["code"];
```
d0sboots: Unneeded, all `ContentFile`s support `.content`.

src/Terminal/commands/grep.ts
```diff
@@ -0,0 +1,179 @@
...
+
+    const parseLine: LineParser = parseFunc.bind(null, file.filename);
+
+    const editedContent: string | string[] = content.split("\n").map(parseLine);
```
d0sboots:
```suggestion
    const editedContent: string[] = [];
    for (const line of file.content.split("\n")) {
      const edited = parseFunc(file.filename, line);
      if (edited !== "") {
        editedContent.push(edited);
      }
    }
```
Optional: This might be a place where being iterative is worthwhile for speed.

src/Terminal/commands/grep.ts
```diff
@@ -0,0 +1,179 @@
...
+  return [
+    [...allContentFiles(server)].map((tuple: FileTuple): ContentFile => tuple[1]),
+    [], // empty array for badFiles
+  ];
+}
+
+function getArgFiles(args: string[]): [(ContentFile | null)[], string[]] {
+  const notFiles: string[] = [];
+  const files: (ContentFile | null)[] = args.map((arg: string): ContentFile | null => {
+    const script: ContentFile | null = hasTextExtension(arg) ? Terminal.getTextFile(arg) : Terminal.getScript(arg);
```
d0sboots: Here and in general: You typically don't need to specify the types on temporaries, TS is pretty good with type inferences. Doing it for function signatures and classes is usually enough.

src/Terminal/commands/grep.ts
```diff
@@ -0,0 +1,179 @@
...
+function getServerFiles(server: BaseServer): [ContentFile[], string[]] {
+  return [
+    [...allContentFiles(server)].map((tuple: FileTuple): ContentFile => tuple[1]),
+    [], // empty array for badFiles
+  ];
+}
+
+function getArgFiles(args: string[]): [(ContentFile | null)[], string[]] {
+  const notFiles: string[] = [];
+  const files: (ContentFile | null)[] = args.map((arg: string): ContentFile | null => {
```
d0sboots: Optional: This map is complicated enough (and it has side effects) such that it feels like it might be better expressed as a for loop.

src/Terminal/commands/grep.ts
```diff
@@ -0,0 +1,179 @@
...
+      notFiles.push(arg);
+    }
+    return script;
+  });
+  return [files, notFiles];
+}
+
+function filterOpts([options, otherArgs]: [Options, string[]], arg: string): [Options, string[]] {
+  const isOption: boolean = Object.keys(OK_ARGS).some((key: string): boolean =>
+    OK_ARGS[key as keyof OkArgs].includes(arg) ? (options[key as keyof Options] = true) : false,
```
d0sboots: Please don't use ternaries for side-effects. Just use an if statement.

src/Terminal/commands/grep.ts
```diff
@@ -0,0 +1,179 @@
...
+  }
+
+  const [options, otherArgs, outFilePath]: [Options, string[], FilePath | null] = filterArgs(args.map(String));
+  if (!outFilePath || !hasTextExtension(outFilePath)) {
+    return Terminal.error(`grep failed: Invalid output file "${outFilePath}". Output file must be a text file.`);
+  }
+
+  console.log(options);
+  if (options.toFile && !options.overWrt) {
+    const overWrtFileExists = [...allContentFiles(server)]
```
d0sboots: Optional: This one makes me so itchy I had to leave a comment XD
This makes two extra copies of the data, when instead you could do this in a single pass with no copies:
```suggestion
    for (const tuple of allContentFiles(server)) {
      if (tuple[1].filename === outFilePath) {
        return Terminal.error(
          `grep failed: Invalid output file "${outFilePath}". Output file must not already exist. Pass -f to force.`,
        );
      }
    }
```

src/Terminal/commands/grep.ts
```diff
@@ -0,0 +1,179 @@
...
+  if (notFiles.length) {
+    return Terminal.error(`Invalid filename(s): ${notFiles.join(", ")}`);
+  }
+
+  try {
+    const pattern: string | RegExp = options.regExpr ? new RegExp(otherArgs[0], "g") : otherArgs[0];
+    const parseFunc: PtLineParser = getParseFunc.bind(null, pattern, options);
+    const result: string = files
+      .flatMap(parseFile(parseFunc))
+      .filter((line: string) => line.length)
```
d0sboots: If you're going to filter unmatched lines here, why even bother including them in the result of `parseFile`?
For large files you're creating a large array only to discard most of it.

d0sboots (commented):
You seem to be having a good time, so I'll review what you've got but feel free to keep going :D

src/Terminal/commands/grep.ts
muesli4brekkies: I have a thing for same-length variable names. :+1:

src/Terminal/commands/grep.ts
muesli4brekkies: Ah yes, this is an atavism. Zoe let me know about the ContentFile type on the Discord. :+1:

src/Terminal/commands/grep.ts
muesli4brekkies: Ah TIL. I'm quite new to TS so tend to be quite paranoid or I get one of those nice wall-of-text errors :smile:

src/Terminal/commands/grep.ts
muesli4brekkies: Ooh neat. I wasn't aware that for of could iterate through that without arrayifying, like the dot methods need. :+1:

muesli4brekkies:
I'm gonna run out of option flags to implement soon :smile:

G4mingJon4s:
About that formatting of the arguments, like when you would do `grep --help` in linux. Do we stick to this `[OPTION]` formatting or do we try to match it to existing commands in the game? I'm asking because I want to "copy" your approach for `rm` to make it more uniform.

Also, I see you parse the arguments yourself. We have a library `import libarg from "arg"` for that, so I'm wondering if you have a reason to not use it?

muesli4brekkies:
> About that formatting of the arguments, like when you would do `grep --help` in linux. Do we stick to this `[OPTION]` formatting or do we try to match it to existing commands in the game?

I figured replicating the command as close as I reasonably could would be ideal and potentially useful, but also most fun and satisfying! 

And I think the other commands should be adjusted to be as close as possible as well, for fun and profit. :grin: 

> Also, I see you parse the arguments yourself. We have a library `import libarg from "arg"` for that, so I'm wondering if you have a reason to not use it?

I was totally unaware of `... from "arg";`! Well hidden, lol - I did cursorily look around the repo for an arg lib, but when not immediately finding one I really didn't fancy pulling a new dep, so I started hacking.

Do you know if it's <https://www.npmjs.com/package/jargs> this? There's no documentation as to what it is (that I can find) but from the few (five) uses it has in the repo, it looks similar.

ps: I'm a little disappointed you pointed that out xD. I'm quite pleased with the shape I got my little parser into haha.

G4mingJon4s:
> Do you know if it's https://www.npmjs.com/package/jargs this? There's no documentation as to what it is (that I can find) but from the few (five) uses it has in the repo, it looks similar.

Sadly, beyond the uses within the game, I have no idea what this package is or what it does.

> I'm a little disappointed you pointed that out xD. I'm quite pleased with the shape I got my little parser into haha.

It's fine if you leave it as is. I'm changing some stuff from the `rm` command and am also not using the library. First, I wanted to, but it doesn't support single flags with multiple characters, e.g. `-rf`, as it's not the standard (I think?). If you don't want to use it, it's probably fine.

muesli4brekkies:
I have multiflags working now anyhow. :smiley: .

While multiflag options might not be POSIX (assuming that they are not), they are definitely Linux. It's something I would miss.

PR #1347, opened 2024-06-04
SLEEVE: Editorial followup on #1314

Snarling:
Just some followup for #1314 to get rid of a double variable that was bothering me. Instead of contractName + contractEnum, it's just contract.


d0sboots:
Ah cool, they looked like different types/usages to me, and I didn't dig deeper.

PR #1348, opened 2024-06-04
IPVGO: Add history, and details to status, to go API

ficocelliguy:
* Add an accessor for the current game's move history as an API method
* Added the current opponent's komi to the getGameState method
* Added the remaining offline cycles to the getGameState method (dependent on #1345 )

### DIFF
--- a/src/BitNode/BitNodeMultipliers.ts
+++ b/src/BitNode/BitNodeMultipliers.ts
@@ -1,5 +1,5 @@
 import { PartialRecord, getRecordEntries } from "../Types/Record";
-
+import { clampNumber } from "../utils/helpers/clampNumber";
 /**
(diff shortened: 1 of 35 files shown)


d0sboots:
How are bonus cycles going to work with Go? I thought it was primarily limited by computation time, not that it was just waiting.

ficocelliguy:
I have some ~300ms sleeps that bookend the start and end of the AI turn, these essentially are skipped and consume an offline cycle each, via #1345 . It still takes time for the AI to run, but this roughly doubles it speed (even more on smaller boards or with few pieces on board)

d0sboots:
Looks like all the offline time stuff got merged already, so this is just history now.

PR #1351, opened 2024-06-04
IPVGO: Use tss-react makeStyles on the IPvGO board and UI

ficocelliguy:
Fix broken IPvGO styles that used a deprecated version of MUI makeStyles

### DIFF
--- a/src/Go/boardState/goStyles.ts
+++ b/src/Go/boardState/goStyles.ts
@@ -1,9 +1,21 @@
 import { Theme } from "@mui/material/styles";
-import makeStyles from "@mui/styles/makeStyles";
-import createStyles from "@mui/styles/createStyles";
+import { makeStyles } from "tss-react/mui";
 
-export const pointStyle = makeStyles((theme: Theme) =>
-  createStyles({
+type sizes = "fiveByFive" | "sevenBySeven" | "nineByNine" | "thirteenByThirteen" | "nineteenByNineteen";
+type points =
+  | "blackPoint"
+  | "whitePoint"
+  | "innerPoint"
+  | "filledPoint"
+  | "emptyPoint"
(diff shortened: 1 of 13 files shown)


src/Go/ui/GoPoint.tsx
```diff
@@ -91,7 +91,9 @@ export function GoPoint({ state, x, y, traditional, hover, valid, emptyPointOwne
 
 export function getSizeClass(
   size: number,
-  classes: ClassNameMap<"fiveByFive" | "sevenBySeven" | "nineByNine" | "thirteenByThirteen" | "nineteenByNineteen">,
```
Caldwell-74, line 94: classes: ReturnType<typeof boardStyles | typeof pointStyle>["classes"]`

could be a better alternative so we dont need the ClassNameMap import

Caldwell-74 (changes requested):

src/Go/boardState/goStyles.ts
```diff
@@ -1,9 +1,31 @@
 import { Theme } from "@mui/material/styles";
-import makeStyles from "@mui/styles/makeStyles";
-import createStyles from "@mui/styles/createStyles";
+import { makeStyles } from "tss-react/mui";
+import { keyframes } from "tss-react";
 
-export const pointStyle = makeStyles((theme: Theme) =>
-  createStyles({
+export type sizes = "fiveByFive" | "sevenBySeven" | "nineByNine" | "thirteenByThirteen" | "nineteenByNineteen";
```
Snarling: Types should be singular and capitalized (e.g. `"fiveByFive"` makes sense to be of type `Size`, not type `sizes`)

It looks like these style category types don't need to be exports since they are only used in this file

PR #1356, opened 2024-06-05
IPVGO: Balance and improvements for offline bonus time cycles

ficocelliguy:
* Do not use hyperspeed bonus time for manual play. The AI near-instantly responding feels a) disrespectful of your moves, and b) not like a real opponent you are playing against
* Rebalance added bonus cycles from offline time, since in-game cycles aren't quite the same as go move times

### DIFF
--- a/src/engine.tsx
+++ b/src/engine.tsx
@@ -331,7 +331,7 @@ const Engine: {
       // Bladeburner offline progress
       if (Player.bladeburner) Player.bladeburner.storeCycles(numCyclesOffline);
 
-      Go.storeCycles(numCyclesOffline);
+      Go.storeCycles(numCyclesOffline / 3);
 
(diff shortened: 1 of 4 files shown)


src/engine.tsx
```diff
@@ -331,7 +331,7 @@ const Engine: {
       // Bladeburner offline progress
       if (Player.bladeburner) Player.bladeburner.storeCycles(numCyclesOffline);
 
-      Go.storeCycles(numCyclesOffline);
+      Go.storeCycles(numCyclesOffline / 3);
```
d0sboots: This seems maybe too harsh? In any case, this should be applied inside `storeCycles`, instead of in engine, for cleanliness.

Edit: Actually, probably best would be to just spend at 3x rate in waitCycle, and then cycles are always cycles.

PR #1358, opened 2024-06-05
UI: Automatically show Bitverse UI if BN is finished

### LINKED ISSUE #1325: When switching bitnodes if the auto save has not occurred and you close the game it could leave you in a bad state.
Bitburner v.2.6.1 (d2c7b2504)

Steps to reproduce:
Destroy bitnode and choose a different bit node, e.g. 1.3 to 2.1.  Quit game before auto save occurs.

Observation:
When starting Bitburner it loads the last save file which occurred before you destroyed the bitnode.  If you have already backdoored world demon you wind up in a stuck state.  No b1tflume.exe either to exit.  You are essentially in a broken state and will need to reload from last save.

Expected:
Game loads fine and maintains previous state. 

Suggestion:
A quick fix for this could be to just autosave when switching bitnodes.  That will ensure the local save file matches and won't leave you in a broken state.

catloversg:
Fixes #1325.

### DIFF
--- a/src/Achievements/Achievements.ts
+++ b/src/Achievements/Achievements.ts
@@ -19,7 +19,6 @@ import { HacknetNode } from "../Hacknet/HacknetNode";
 import { HacknetServer } from "../Hacknet/HacknetServer";
 import { Player } from "@player";
 import { GetAllServers, GetServer } from "../Server/AllServers";
-import { SpecialServers } from "../Server/data/SpecialServers";
 import { Server } from "../Server/Server";
 import { Router } from "../ui/GameRoot";
 import { Page } from "../ui/Router";
@@ -30,7 +29,6 @@ import { workerScripts } from "../Netscript/WorkerScripts";
 
 import { getRecordValues } from "../Types/Record";
 import { ServerConstants } from "../Server/data/Constants";
-import { blackOpsArray } from "../Bladeburner/data/BlackOperations";
 
 // Unable to correctly cast the JSON data into AchievementDataJson type otherwise...
 const achievementData = (<AchievementDataJson>(<unknown>data)).achievements;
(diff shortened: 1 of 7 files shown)


Kelenius:
Am I reading this correctly that if you have finished the last black OP, the game will immediately throw you out of the bitnode?

catloversg:
It will show the Bitverse UI the next time the game is loaded, not immediately after finishing the last BlackOps. I can make a special case for it (only automatically show Bitverse UI if WD is backdoored).

src/PersonObjects/Player/PlayerObject.ts
```diff
@@ -132,6 +132,9 @@ export class PlayerObject extends Person implements IPlayer {
   sourceFileLvl = generalMethods.sourceFileLvl;
   applyEntropy = augmentationMethods.applyEntropy;
   focusPenalty = generalMethods.focusPenalty;
+  bitNodeFinishedState = generalMethods.bitNodeFinishedState;
+  canAccessBitNodeFeature = generalMethods.canAccessBitNodeFeature;
```
d0sboots: What is the purpose of moving these methods onto `Player`? It's a god-object, and IMO we should avoid expanding it further. I can see the logic for sharing `bitNodeFinishedState` (although there are other ways it could be shared, probably), but the other two are only used for the achievements I think?

d0sboots (commented):
I think what would be most consistent would be for "WD backdoored" to be the definitive determination of "you have beaten the node." So clicking the button at the end of Bladeburner would apply backdoor; any other alternate victory conditions would do the same.

This simplifies the logic in `bitNodeFinishedState` and makes it easy to know when we should start showing the selection screen. (It's exactly the same logic.)

src/Achievements/Achievements.ts
```diff
@@ -61,18 +60,11 @@ export interface AchievementData {
...
+function canAccessBitNodeFeature(bitNode: number): boolean {
+  return Player.bitNodeN === bitNode || Player.sourceFileLvl(bitNode) > 0;
 }
 
-function hasAccessToSF(bn: number): boolean {
-  return Player.bitNodeN === bn || Player.sourceFileLvl(bn) > 0;
-}
-
-function knowsAboutBitverse(): boolean {
+function knowAboutBitverse(): boolean {
```
d0sboots: Typo when you reverted this

PR #1360, opened 2024-06-06
IPVGO: Fix crash caused by malformed previous move formatting

ficocelliguy:
Details here:
https://discord.com/channels/415207508303544321/1221920484111814828/1248051409950081144

Due to a malformed array created during loading the move history, the prior move board isn't actually an array of strings and causes recovery ode


d0sboots:
This is a case where being overly functional bites you in the butt. But, OK. XD

PR #1362, opened 2024-06-06
CORPORATION: Refactor markup multiplier

catloversg:
The markup multiplier can be:
- Penalty modifier: $\left(\frac{MarkupLimit}{SellingPrice - MarketPrice}\right)^{2}$ [1]
- Bonus modifier: $\frac{MarketPrice}{SellingPrice}$ [2]
- $10^{12}$ [3]

For a very long time, only [1] was applicable to both materials and products. It creates 2 problems:
- The player cannot discard products. This is fixed by #1330.
- Products do not get a bonus modifier when the selling price is less than the market price.

This PR "unifies" the logic of the markup multiplier for both materials and products and fixes the above problem.


d0sboots:
Yay, more corp refactoring pls

PR #1363, opened 2024-06-06
BLADEBURNER: Add Stop button and refactor code

catloversg:
This PR makes these changes:
- Add Stop button to 3 tabs: Contracts, Operations, and Black Operations.
- Start/Stop button is on the same row as the action name.
- Refactor related code.

### DIFF
--- a/src/Bladeburner/ui/ActionHeader.tsx
+++ b/src/Bladeburner/ui/ActionHeader.tsx
@@ -1,0 +1,55 @@
+import type { Bladeburner } from "../Bladeburner";
+import type { Action } from "../Types";
+
+import React from "react";
+import { Box, Typography } from "@mui/material";
+import { CopyableText } from "../../ui/React/CopyableText";
+import { createProgressBarText } from "../../utils/helpers/createProgressBarText";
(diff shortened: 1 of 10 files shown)


src/Bladeburner/ui/ActionHeader.tsx
```diff
@@ -0,0 +1,55 @@
...
+          <CopyableText value={action.name} />
+          <StopButton bladeburner={bladeburner} rerender={rerender} />
+        </Box>
+        <Typography>
+          (IN PROGRESS - {formatNumberNoSuffix(computedActionTimeCurrent, 0)} /{" "}
+          {formatNumberNoSuffix(bladeburner.actionTimeToComplete, 0)})
+        </Typography>
+        <Typography>
+          {createProgressBarText({
+            progress: computedActionTimeCurrent / bladeburner.actionTimeToComplete,
```
d0sboots, line 41: There was a recent change to grafting and programs so they'd show ETA, does that make sense for BB as well?

catloversg: It's a good suggestion. I'll make a PR for it.

PR #1365, opened 2024-06-06
MISC: Refactor code of traveling and going to location

catloversg:
This PR makes these changes:
- Fix a bug: Sleeves can travel even when there is not enough money. The bug is in `src\PersonObjects\Sleeve\ui\TravelModal.tsx`.
- `ns.sleeve.travel` now prints a warning when the sleeve cannot travel to a city due to insufficient funds.
- Refactor related code.
- Throw an error if `Player.travel` or `Player.gotoLocation` receives an invalid parameter (city or location).
- Use `Player.gotoLocation()` instead of `Player.location = X`.

### DIFF
--- a/src/Locations/ui/CompanyLocation.tsx
+++ b/src/Locations/ui/CompanyLocation.tsx
@@ -57,7 +57,7 @@ export function CompanyLocation(props: IProps): React.ReactElement {
    */
   const currentPosition = jobTitle ? CompanyPositions[jobTitle] : null;
 
-  Player.location = companyNameAsLocationName(props.companyName);
+  Player.gotoLocation(companyNameAsLocationName(props.companyName));
 
   function startInfiltration(e: React.MouseEvent<HTMLElement>): void {
     if (!e.isTrusted) {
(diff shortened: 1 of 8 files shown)


d0sboots:
Didn't the sleeve travel bug already get fixed in a different PR?

d0sboots:
Also, just double-checking that the change to travel here doesn't inadvertently extend checking to a wider class of situations.

catloversg:
> Didn't the sleeve travel bug already get fixed in a different PR?

That PR only fixes the bug in `ns.sleeve.travel`. The player can still exploit the bug in UI. In `src\PersonObjects\Sleeve\ui\TravelModal.tsx`, we check the condition, but do not `return` when it fails.
https://github.com/bitburner-official/bitburner-src/blob/abcd6c545a29928aa98d028c9fba64c95a3e8725/src/PersonObjects/Sleeve/ui/TravelModal.tsx#L22-L24
When I saw that bug, I checked the implementation of traveling. The logic of traveling is duplicated and spread everywhere, so I decided to refactor it.

> Also, just double-checking that the change to travel here doesn't inadvertently extend checking to a wider class of situations.

I tested these cases:
- UI:
  - Travel by Player (Travel tab).
  - Travel by Sleeves (Sleeves tab).
- API:
  - `ns.singularity.travelToCity`.
  - `ns.sleeve.travel`.

`Player.travel` is only called by `ns.singularity.travelToCity` and UI code in `src\Locations\ui\TravelAgencyRoot.tsx`.
`Sleeve.travel` is only called by `ns.sleeve.travel` and UI code in `src\PersonObjects\Sleeve\ui\TravelModal.tsx`.

PR #1372, opened 2024-06-07
EDITOR: Changed editor tabs to have their own mode

G4mingJon4s:
Changes based on feedback from #1336.


d0sboots:
You forgot to change `vim.tsx`. (Probably didn't notice because you had the setting set... you should test all three commands with both sides of the setting to make sure they're all working right.)

G4mingJon4s:
> You forgot to change `vim.tsx`

I didn't have to change the `vim` command at all, because it already sets the `vim` prop. `nano` however assumed not setting this property will result in the default editor being used. Since the monaco editor is no longer the default, but rather the editor mode the player chooses, I have to explicitly provide a mode to be used.

> You should test all three commands with both sides of the setting to make sure they're all working right

Testing out all commands gave predictable behavior:
- `vim` and `nano` ignore the default entirely and set their editor mode for the scripts.
- `ls` uses the default to set the mode for the clicked script.
- Opening a script that is already open will use the mode that is already set.

The last point is debatable: It makes sense for the mode to be preserved and not changed, but it would also make sense for it to change according to the command used

d0sboots:
That's sneaky, I missed that this was just re-using an existing prop.

PR #1375, opened 2024-06-08
BUGFIX: bladeburner.getActionRepGain: mark level param as optional in signature

mrsimo:
This parameter is documented as optional, but the signature is causing my TypeScript compilation to complain. Making this change in my own `NetscriptDefinitions.d.ts` stops the compilation error and the game seems to work fine 😄

Please let me know if I missed something from the contribution process. Thank you for an excellent game!


catloversg:
When you change things in `NetscriptDefinitions.d.ts`, you need to run `npm run doc` to regenerate the documentation.

mrsimo:
Thanks a lot @catloversg! Apologies for missing that. Should be updated now.

PR #1376, opened 2024-06-08
MISC: Fix wrong help text of cd command

catloversg:
> Note that this works even for directories that don't exist.

This is wrong. If the directory does not exist, `cd` will print an error.

> If you change to a directory that does not exist, it will not be created.

This is a really weird description for `cd`. Is there anybody expecting `cd` to create a non-existent directory? I'm not sure if we should keep it, so I did not change or remove it.

### DIFF
--- a/src/Terminal/HelpText.ts
+++ b/src/Terminal/HelpText.ts
@@ -148,7 +148,7 @@ export const HelpTexts: Record<string, string[]> = {
   cd: [
     "Usage: cd [dir]",
     " ",
-    "Change to the specified directory. If you change to a directory that does not exist, it will not be created. Examples:",
+    "Change to the specified directory. You cannot change to a directory that does not exist. Examples:",
     " ",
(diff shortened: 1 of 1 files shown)


src/Terminal/HelpText.ts
```diff
@@ -148,8 +148,7 @@ export const HelpTexts: Record<string, string[]> = {
   cd: [
     "Usage: cd [dir]",
     " ",
-    "Change to the specified directory. Note that this works even for directories that don't exist. If you ",
-    "change to a directory that does not exist, it will not be 'created'. Examples:",
+    "Change to the specified directory. If you change to a directory that does not exist, it will not be created. Examples:",
```
d0sboots: I'd say that trying to change to a directory that doesn't exist is an error. The current wording is weird, but made more sense back when you could cd to non-existing paths.

PR #1377, opened 2024-06-08
BUGFIX: Fix inconsistencies of error handling in promise-based API

catloversg:
When we get a normal server from `GetServer`, we usually check the returned value with `(!(server instanceof Server))` and throw/show an error if it fails the check. This is especially important with promise-based API, e.g., hack, grow, weaken, etc., because the player may put them in a `while (true)` loop. However, we only throw an error with `ns.hack`, but we don't throw an error with other promise-based API. This PR fixes this problem.

```js
/** @param {NS} ns */
  if (!ns.serverExists("hacknet-server-0")) {
    ns.hacknet.purchaseNode();
  }
  while (true) {
    // Ok
    await ns.hack("hacknet-server-0");
    // Freeze
    // await ns.grow("hacknet-server-0");
    // Freeze
    // await ns.weaken("hacknet-server-0");
    // Freeze
    // ns.singularity.connect("hacknet-server-0");
    // await ns.singularity.installBackdoor("hacknet-server-0");
  }
}
```

There are many other API that do not throw when they receive a non-hackable server as a parameter. They are not promise-based API, and most of them are "get-data" type API, so I don't touch them in this PR. They are:
- `ns.hackAnalyzeThreads`.
- `ns.hackAnalyze`.
- `ns.hackAnalyzeSecurity`.
- `ns.hackAnalyzeChance`.
- `ns.growthAnalyze`. Note: there is this comment in the source code: `// Todo 2.3: Make this throw instead of returning 0?`.
- `ns.growthAnalyzeSecurity`.
- `ns.nuke`.
- `ns.brutessh`.
- `ns.ftpcrack`.
- `ns.relaysmtp`.
- `ns.httpworm`.
- `ns.sqlinject`.
- `ns.getServerMoneyAvailable`.
- `ns.getServerSecurityLevel`.
- `ns.getServerBaseSecurityLevel`.
- `ns.getServerMinSecurityLevel`.
- `ns.getServerRequiredHackingLevel`.
- `ns.getServerMaxMoney`.
- `ns.getServerGrowth`.
- `ns.getServerNumPortsRequired`.
- `ns.deleteServer`.
- `ns.getHackTime`.
- `ns.getGrowTime`.
- `ns.getWeakenTime`.


d0sboots:
I'm very wary of this type of change, because it's API-breaking and not for a bugfix reason. It's a little better because hack always threw, so this never fully worked.

catloversg:
I don't have a strong opinion about this one, so it's up to you and Snarling. If it's too risky, you can close it. We may deal with it later in v3.0.

d0sboots:
I've come around on this one. The key bit is that you're not changing the check itself - all these functions already checked !instanceof Server and failed in that case - but you're just changing how that failure manifests. All these cases should be programmer error - I really really cannot think of any usecase where a script would be running successfully and relying on growing/weakening invalid servers. So this shouldn't actually break running code, just help with debugging.

PR #1378, opened 2024-06-08
TERMINAL: Added deleting entire directories using rm

### LINKED ISSUE #48: rm -r should confirm that you want to delete a directory and then do it.
[hydroflame]
null

[Master-Guy]
> Do you want this to be a yes/no prompt in the UI, or a textual yes/no question in the terminal?
> The first is easy to implement, the latter might be very non-trivial

[Undeemiss]
> bump. I can do this if given clear direction (I was planning to write a script to do this for my personal save, but can just write it as a feature instead if you want)

G4mingJon4s:
Currently, `rm` can only delete one single file at a time and can't delete directories. Also, deleting a file requires no confirmation at all. The player could delete something they actually don't want to, if they are not careful.

#### Changes

- `rm` can now accept multiple targets at once.
- `rm` only deletes the targets if the player actively confirms the action, or adds the `-f` flag.
- `rm` can now delete entire directories, if the `-r` flag has been provided.
- `rm` actively prevents deleting the entire file system, unless `--no-preserve-root` is added.
- In general, `rm` is now more like the actual `rm` command.

Closes #48


d0sboots:
Overall I like the added functionality and implementation.

However, I don't think `-f` should be required by default, since that is both an (IMO bothersome) change from current and also not how standard rm works. I think a good compromise would be to only require it if you are deleting more than one file (optional whether this includes empty directories). That doesn't change the current behavior, but still keeps people from shooting themselves in the foot.

An optional enhancement would be to show what *will* be deleted when asking for confirmation.

PR #1381, opened 2024-06-08
TERMINAL: Add grep command

muesli4brekkies:
Me again!

Very sorry! :melting_face: I made a branch and nuked my last PR.

This is a fresh one. Sorry for the bump. My git-fu needs some serious work.


src/Terminal/commands/grep.ts
```diff
@@ -0,0 +1,445 @@
...
+    this.matchCounter = 0;
+  }
+
+  addContext(context: number): Results {
+    const nContext = isNaN(Number(context)) ? 0 : Number(context);
+    for (const [editLineIndex, line] of this.lines.entries()) {
+      if (!line.isMatched) continue;
+      for (let contextLineIndex = 0; contextLineIndex <= nContext; contextLineIndex++) {
+        let contextLine;
+        if (this.options.isPreContext) contextLine = this.lines[editLineIndex - contextLineIndex];
```
d0sboots: Anytime you have `else` or `else if` going on, please use curly braces. One-line if-syntax is very bug-prone, and IMO should only be used for the simplest cases (like your `if (!matched) continue` above.)

src/Terminal/commands/grep.ts
```diff
@@ -0,0 +1,445 @@
...
+    return this;
+  }
+
+  splitAndFilter(): [string[], string[]] {
+    return this.lines.reduce(
+      ([rawResult, prettyResult]: [string[], string[]], lineInfo: Line): [string[], string[]] => {
+        return lineInfo.isPrint === this.options.isInvertMatch
+          ? [rawResult, prettyResult]
+          : [
+              [...rawResult, lineInfo.lines.rawLine],
```
d0sboots: I can't bite my tongue about this. XD Each iteration, you are destructuring the previous array into a new array, and then adding a new item at the end. In other words, you are turning what should be a simple O(N) operation into O(N^2).

In addition, you have this complicated ternary just to create a bypass for the "don't do anything case." This whole thing becomes *much* simpler if you just do it iteratively:
```suggestion
  if (lineInfo.isPrint !== this.options.isInvertMatch) return [[], []];
  const rawResult = [], prettyResult = [];
  for (const lineInfo of this.lines) {
    rawResult.push(lineInfo.lines.rawLine);
    prettyResult.push(lineInfo.lines.prettyLine);
  }
  return [rawResult, prettyResult];
```

src/Terminal/commands/grep.ts
```diff
@@ -0,0 +1,445 @@
...
+      },
+      [[], []],
+    );
+  }
+
+  capMatches(limit: number): Results {
+    if (!this.options.isMaxMatches) return this;
+    for (const line of this.lines) {
+      if (line.isMatched) this.matchCounter += 1;
+      if (this.matchCounter > Number(limit)) line.isMatched = false;
```
d0sboots: limit is a number, so explicitly running it through `Number` should be unneeded?

d0sboots:
It looks like you've got everything you want in, so hopefully this will be the last review round :D

muesli4brekkies:
Aye. :). I think it's pretty "feature complete", for what I can achieve. The big differences to real grep that I can think of now are;

* missing the -o flag to only print matches. This would require quite a rework of the matching algorithm.
* -m n counts LINES instead of matches currently. This is also a consequence of the matching algorithm.

But honestly I think that's about it. Not to mention those options are most useful when piping stdout somewhere else in a bash script or similar.

I think it gets the job done!

d0sboots (commented):
Oh no, I forgot to submit my review so these didn't show up :(

src/Terminal/commands/grep.ts
muesli4brekkies: Agreed.

src/Terminal/commands/grep.ts
muesli4brekkies: No tongue biting necessary. :). I am still quite blind to these things and I'm very glad you point them out!

:+1:

src/Terminal/commands/grep.ts
muesli4brekkies: Correct. :) Good spot. I was passing it as a string before, but I typecheck it earlier in the program now.

d0sboots:
Ship it!

muesli4brekkies:
Amazing! :D Thanks a lot for your help with this!

PR #1387, opened 2024-06-12
API: change Difficulty to Security in getServer and related internal Code

Caldwell-74:
The inconsistency between ns.getServer\<subFunction\> and ns.getServer().property naming is something i would like to remove before 3.0

ns.getServer().baseDifficulty !== ns.getServerBaseSecurity
ns.getServer().minDifficulty  !== ns.getServerMinSecurity
ns.getServer().hackDifficulty !== ns.getServerSecurityLevel

my changes are:

ns.getServer().baseSecurity === ns.getServerBaseSecurity
ns.getServer().minSecurity  === ns.getServerMinSecurity
ns.getServer().security !== ns.getServerSecurityLevel 

for the last i dont see a reason for the "level" addition i think it should be removed from the function name aswell 
if anything it should be currentSecurity and CurrentSecurity but just security is clear enough imo

i also changed the internal side to use Security instead of Difficulty except inside calculations
the reason here beeing that security is the name of the game mechanic component
while difficulty explains more precisely what its role is inside the calculation 

thats ofcourse just my subjective opinion iam open to any change here


d0sboots:
Could you maybe split this in two? The reason being, the change to the internal names I am down for and it can be done immediately (it doesn't impact anything player-facing.) It's also the larger of the two. The change to API names is the one we have to be more careful with.

Caldwell-74:
yes i can just need to see how i do it the formulas section where player facing api and internal collide

d0sboots:
oh yeah that might be a bit spicy

d0sboots:
One other thing, if you're feeling bold: We could try to kill `baseDifficulty` completely XD (There's only a couple uses holding us back rn)

Caldwell-74:
we could derive it from minSecurity but that would mean that the hacknet minSecurity reduction decreases exp gain 
other option would be using current security which might open up for a new more tricky but optimizable exp gain strat?

PR #1388, opened 2024-06-12
DOCUMENTATION: Clarify maximum number of ports

### LINKED ISSUE #1190: FEATURE and DOCUMENTATION REQUEST: Port limit
#1189 : I hit the port limit and the script crashed, and it's not immediately obvious what went wrong nor what I needed to do to fix it.

I'd like to request:
- something in the documentation saying that ports have a limit, and one should clean after their ports with `readPort` or `clearPort`.
- make the error message more clear what's happening - make it say "port limit reached" or something. Otherwise it's not immediately obvious which `Map` is exceeding its limit, especially considering I wasn't using a map myself.
- probably bad idea, but probably have `clearAllPorts` and `clearPortRange` functions so that cleaning up is easier.

catloversg:
Closes #1190.

### DIFF
--- a/markdown/bitburner.ns.writeport.md
+++ b/markdown/bitburner.ns.writeport.md
@@ -31,3 +31,5 @@ RAM cost: 0 GB
 
 Write data to the given Netscript port.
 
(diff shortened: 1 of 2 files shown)


d0sboots:
Do we still have the incorrect port limit in the docs somewhere? I feel like we do, but maybe it got cleaned up and I forgot...

catloversg:
I did not find any obsolete information in `src/Documentation` and `NetscriptDefinitions.d.ts`.

PR #1389, opened 2024-06-12
MISC: Add ns.enums to AutocompleteData

### LINKED ISSUE #581: Script autocomplete should have more data types provided
Some data from the `ns` object would be very useful for setting up autocomplete. For example, I have wanted to use:
- `ns.stock.getSymbols()`
- `ns.enums.CrimeType`
- `ns.corporation.getCorporation().divisions`
- `Object.keys(ns.getPlayer().skills)`
- `ns.read('listOfThings.txt')`
- `ns.readPort(CONFIG_PORT)`

catloversg:
Closes #581.


src/Terminal/getTabCompletionPossibilities.ts
```diff
@@ -324,6 +325,7 @@ export async function getTabCompletionPossibilities(terminalText: string, baseDi
         .map((server) => server.hostname),
       scripts: [...currServ.scripts.keys()],
       txts: [...currServ.textFiles.keys()],
+      enums: structuredClone(enums),
```
d0sboots: Is it possible for enums to be a frozen object such that we don't need to clone it? This isn't in the critical path for module creation, it only gets called for autocomplete, but it would be nice to keep the number of objects created down, if possible.

If there's a technical reason why that isn't easy, this is fine.

catloversg: `enums` is frozen in `src\NetscriptFunctions.ts`, so it's fine to use `enums: enums`. I used it like that at first, but before committing, I got paranoid about `Object.freeze`, so I added `structuredClone`. The usage of `Object.freeze` in `NetscriptFunctions.ts` looks fine, though.

PR #1390, opened 2024-06-12
API: Add getSaveData API

### LINKED ISSUE #851: Concept: Automatic export to GitHub Gists or other online document platform
**Idea**
1. In the option menu, add a login button to GitHub (or Google Docs, Dropbox, etc.). (EDIT: Steam?)
2. When logged in, game [creates a Gist](https://docs.github.com/en/rest/gists/gists?apiVersion=2022-11-28#create-a-gist) or [read](https://docs.github.com/en/free-pro-team@latest/rest/gists/gists?apiVersion=2022-11-28#get-a-gist) current save game and then [update it periodically](https://docs.github.com/en/rest/gists/gists?apiVersion=2022-11-28#update-a-gist).

Optionally add a profile or save game name, in case players would like to switch between saves.

catloversg:
This PR implements my suggestion at https://github.com/bitburner-official/bitburner-src/issues/851#issuecomment-2156607081. ~~There is a small change: the returned data is `Uint8Array | string` instead of `Uint8Array`.~~
Edit: The returned data is still a `Uint8Array`.

I put the new API in the singularity namespace because all automation APIs are in that namespace, including the related `ns.Singularity.exportGame` API.

Closes #851.


src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -1690,6 +1690,15 @@ export type Task = StudyTask | CompanyWorkTask | CreateProgramWorkTask | CrimeTa
  * @public
  */
 export interface Singularity {
+  /**
+   * This function returns the save data. If your browser supports Compression Streams API, the save data is a
```
d0sboots: Can we get rid of this caveat? Converting a string to a Uint8Array is not very hard. Does this have to do with the underlying format being compressed vs. not?

d0sboots:
Codewise, LGTM. I will leave this to @Snarling for the new API/RAM cost, although I think it looks appropriate.

src/ScriptEditor/NetscriptDefinitions.d.ts
catloversg: > Does this have to do with the underlying format being compressed vs. not?

Yes. The data is different:
- `Uint8Array`: compressed json save string. This is our current format.
- `string`: base64-encoded json save string. This is the legacy format. We still export the save data to this format if the player's browser does not support Compression Streams API.

d0sboots: IIRC, we support (and will continue to support) the old format in perpetuity. The issue here is the different data types, which are a needless complication for the user (also I would assume for us, but maybe we are using it to differentiate between format types?)

IMO for an API we should be guaranteeing a single data type. Since the string is already base64-encoded (and thus ASCII), this should be as simple as `new TextEncoder().encode(str)`

catloversg: > maybe we are using it to differentiate between format types?

That's right. I'm not sure if we can "simplify" `export type SaveData = string | Uint8Array;` to `export type SaveData = Uint8Array;` and change related code. I'll take a look at it later. If there is no problem, I'll make a PR for it.

> IMO for an API we should be guaranteeing a single data type. Since the string is already base64-encoded (and thus ASCII), this should be as simple as new TextEncoder().encode(str)

I agree.

catloversg:
@d0sboots This PR has been stuck for a long time. I think we can proceed without waiting.

PR #1393, opened 2024-06-12
BUGFIX: Fix wrong position name in ns.formulas.work.companyGains

### LINKED ISSUE #1220: formulas.companyGains uses incorrect job position for Senior Software Consultant, Senior Business Consultant, Part-time Employee and Part-time Waiter
Can be seen by comparing the values gotten from

```
ns.singularity.getCompanyPositionInfo(company, position).salary
 * ns.getPlayer().mults.work_money
 * bitnode_multiplier_for_work_money
```

vs.

```
ns.formulas.work.companyGains(ns.getPlayer(), company, position, 0).money
```

The values match for all company / job position combinations except for Senior Software Consultant, Senior Business Consultant, Part-time Employee and Part-time Waiter. The result gotten from ns.formulas matches the salary of Software Consultant, Business Consultant, Employee and Waiter, respectively. As far as I can tell, that is because formulas uses findEnumMember from https://github.com/bitburner-official/bitburner-src/blob/v2.6.0/src/utils/helpers/enum.ts and that returns the first enum member that is a substring of the parameter.

singularity.getCompanyPositionInfo doesn't use findEnumMember and gives correct values.

Bitburner version: v2.6.0 (8154a4613) on https://bitburner-official.github.io/

catloversg:
`findEnumMember` was added in #212. It allows fuzzy matching of player input with enum members. In #596, Snarling added `getEnumHelper` that has many features, including fuzzy matching. Nowadays, we mostly use `getEnumHelper.nsGetMember` instead of `findEnumMember`. In my opinion, this is a good move, because allowing the player to use the "fuzzy" value when calling NS APIs will create potential bugs in the future.

`findEnumMember` is still used in these places:
- `fromJSON` in `src\Work\ClassWork.tsx`. This is fine.
- Singularity API: `ns.singularity.universityCourse` and `ns.singularity.gymWorkout`.
- Formulas API: `ns.formulas.work.gymGains`, `ns.formulas.work.universityGains`, `ns.formulas.work.factionGains`, and `ns.formulas.work.companyGains`.

The usage of `findEnumMember` in `ns.formulas.work.companyGains` creates the bug in #1220. This PR fixes that specific bug.

I also want to discuss changing other APIs that use `findEnumMember` (check the list above). It does not make sense to enforce the check of enum values in most places but to allow "fuzzy" values in other places. For example, in `ns.formulas.work.gymGains`, we check the location name with `getEnumHelper("LocationName").nsGetMember`, but allow the "fuzzy" value for `_classType`.
Technically, this is a breaking change, but I think it's a good one.
@Snarling @d0sboots What are your opinions?

I found many other places that use `.toLowerCase` to accept "fuzzy" values, but I will leave them there and make the change in v3.0.

Fixes #1220.

### DIFF
--- a/src/NetscriptFunctions/Formulas.ts
+++ b/src/NetscriptFunctions/Formulas.ts
@@ -26,15 +26,7 @@ import {
   calculateGrowTime,
   calculateWeakenTime,
 } from "../Hacking";
-import {
-  CityName,
-  CompletedProgramName,
-  FactionWorkType,
-  GymType,
-  JobName,
-  LocationName,
-  UniversityClassType,
-} from "@enums";
+import { CityName, CompletedProgramName, FactionWorkType, GymType, LocationName, UniversityClassType } from "@enums";
 import { Formulas as IFormulas, Player as IPlayer, Person as IPerson } from "@nsdefs";
 import {
   calculateRespectGain,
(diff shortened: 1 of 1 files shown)


d0sboots:
I think the general cleanup is a good idea, but it should happen at a major version bump because there *will* be people relying on the current fuzzy matching.

PR #1394, opened 2024-06-12
BLADEBURNER: Fix NaN in getSuccessRange

catloversg:
A player reported that their `Raid` operation showed `Estimated success chance: 0.0% ~ NaN%`. This happens when `popEst=0` and `comms=0`. `popEst=0` makes `r=Infinity`, `comms=0` makes `high=0`, and `0*Infinity=NaN`. This PR fixes this problem.

### DIFF
--- a/src/Bladeburner/Actions/Action.ts
+++ b/src/Bladeburner/Actions/Action.ts
@@ -146,9 +146,19 @@ export abstract class ActionClass {
     let high = real + diff;
     const city = bladeburner.getCurrentCity();
     let r = city.pop / city.popEst;
-    if (Number.isNaN(r)) r = 0;
-    if (r < 1) low *= r;
-    else high *= r;
+    if (Number.isNaN(r)) {
+      r = 0;
+    }
+    if (r < 1) {
+      low *= r;
+    } else {
+      // This happens when the action is Raid, popEst=0, and comms=0.
+      if (high === 0 && r === Infinity) {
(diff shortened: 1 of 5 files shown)


src/Bladeburner/Actions/Action.ts
```diff
@@ -146,9 +146,19 @@ export abstract class ActionClass {
...
-    if (Number.isNaN(r)) r = 0;
-    if (r < 1) low *= r;
-    else high *= r;
+    if (Number.isNaN(r)) {
+      r = 0;
+    }
+    if (r < 1) {
+      low *= r;
+    } else {
+      // This happens when the action is Raid, popEst=0, and comms=0.
```
d0sboots: I think you can simplify this to just `high *= clamp(r)`, which is easier to reason about.

catloversg: Do you mean to simplify the entire `if (high === 0 && r === Infinity) { high = 0; } else { high *= r; }` to `high *= clamp(r)`?
I don't think we can do that. `r` can be greater than 1. For example, if `city.pop` is 2x`city.popEst`, `r` is 2.
We may use the `clampNumber` function: `high *= clampNumber(r)`.

d0sboots: Ah yes, that is what I meant. The new function Caldwell added for dealing with this type of issue.

PR #1399, opened 2024-06-15
MISC: Remove mention of re-sleeving

catloversg:
This PR removes the remaining mentions of the re-sleeving feature.


d0sboots:
Oh huh, *that's* what resleeving was.

PR #1401, opened 2024-06-15
BUGFIX: Re-added rm deleting all files

G4mingJon4s:
This fixes an issue created by #1378, where the player can't delete certain file-types.

### DIFF
--- a/src/Terminal/commands/rm.ts
+++ b/src/Terminal/commands/rm.ts
@@ -1,9 +1,7 @@
 import { Terminal } from "../../Terminal";
 import { BaseServer } from "../../Server/BaseServer";
 import { PromptEvent } from "../../ui/React/PromptManager";
-import { hasScriptExtension } from "../../Paths/ScriptFilePath";
-import { hasTextExtension } from "../../Paths/TextFilePath";
-import type { Directory } from "../../Paths/Directory";
+import { isDirectoryPath, type Directory } from "../../Paths/Directory";
 import type { IReturnStatus } from "../../types";
 import type { FilePath } from "../../Paths/FilePath";
 
@@ -39,7 +37,7 @@ export function rm(args: (string | number | boolean)[], server: BaseServer): voi
   const files: FilePath[] = [];
 
   for (const target of targets) {
-    if (!hasTextExtension(target) && !hasScriptExtension(target)) {
+    if (isDirectoryPath(target) || !/\..+$/.test(target)) {
       const dirPath = Terminal.getDirectory(target);
       if (dirPath === null) return Terminal.error(errors["invalidDir"](target));
       if (!recursive) return Terminal.error(errors["dirsProvided"]());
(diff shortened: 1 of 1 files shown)


src/Terminal/commands/rm.ts
```diff
@@ -57,6 +55,19 @@ export function rm(args: (string | number | boolean)[], server: BaseServer): voi
     for (const file of server.scripts.keys()) {
       if (file.startsWith(dir)) files.push(file);
     }
+    for (const file of server.textFiles.keys()) {
+      if (file.startsWith(dir)) files.push(file);
+    }
+    for (const file of server.messages) {
+      if (file.endsWith(".msg")) continue;
```
d0sboots, line 62: This seems like the intention is to avoid deletion of messages, but in that case, why are you even iterating this array? Especially since `server.removeFiles` doesn't work on messages.

src/Terminal/commands/rm.ts
```diff
@@ -39,7 +37,7 @@ export function rm(args: (string | number | boolean)[], server: BaseServer): voi
   const files: FilePath[] = [];
 
   for (const target of targets) {
-    if (!hasTextExtension(target) && !hasScriptExtension(target)) {
+    if (isDirectoryPath(target) || !/\..+$/.test(target)) {
```
d0sboots, line 40: This won't work.

There is *no* way to tell whether or not something is a directory, just by looking at its filename. For instance, the path `/foo.txt/blar.txt` is valid, so `/foo.txt` could be a file or a directory. So you can only tell that a path is a directory from it being a validly-constructed path (`Terminal.getDirectory` returns non-null) *and* it's not itself a file *and* it is a prefix of some existing files.

d0sboots (commented):
Most of this is good, but there is a significant issue.

d0sboots:
I made my own continuation of this in #1404 because it was rather involved, please take a look and let me know (otherwise, I'll likely submit it in a day or so so that `rm` doesn't stay broken).

src/Terminal/commands/rm.ts
G4mingJon4s: The `server.messages` array contains the `.msg` files and the `.lit` files, so I iterate over the array and filter out all of the `.msg` files, since these can not be deleted.

src/Terminal/commands/rm.ts
G4mingJon4s: I'm not quite sure how to properly handle the directory names, since the file-system is one big mess.
During my testing I found out that there can be directories and files with the same name.
I tried to avoid the problem here and did the following:
- If the target has no extension, it is a directory. That's the check for a match against the regex.
- Else, the function `isDirectoryPath` handles how the input should be properly formatted. If there is `a.js` as a file and a directory, the input `a.js/` will reference the directory and not the file.

d0sboots:
I've merged the follow-on #1404, so this version should be unneeded now.

PR #1407, opened 2024-06-18
MISC: Blade Burner API access and Blades Simulacrum update

Sphyxis:
As per a conversation in Dev:
https://discord.com/channels/415207508303544321/459097632896188436/1251988703132254229

This update unlocked the Blade Burner API upon completion of, and while you are within BN's 6 and 7.
The BN descriptions have been updated to show this change.

As well, the Blades Simulacrum augment is now a reward for completing BN 7.3.  It is only given upon first initialization of Blade Burner.

Looking for conversation around what the BN description should be changed to.

Text changes in the BN's posted as well
BN7Screenshot 2024-06-17 210216
BN6Screenshot 2024-06-17 210151

### DIFF
--- a/src/BitNode/BitNode.tsx
+++ b/src/BitNode/BitNode.tsx
@@ -198,13 +198,14 @@ export function initBitNodes() {
         <br />
         <br />
         In this BitNode you will be able to access the {FactionName.Bladeburners} Division at the NSA, which provides a
-        new mechanic for progression.
+        new mechanic for progression and through the {FactionName.Bladeburners} API, which allows you to access{" "}
+        {FactionName.Bladeburners} functionality through Netscript.
         <br />
         <br />
         Destroying this BitNode will give you Source-File 6, or if you already have this Source-File it will upgrade its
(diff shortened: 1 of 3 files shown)


d0sboots:
I'm for this in general, the whole issue with BN6 feeling redundant (especially since it comes first) is real. I vaguely remember the discussion (it happened across multiple channels), I don't remember if @Snarling signed off on it or not.

Alpheus:
Bump, this is a great change!

d0sboots:
I seem to recall this had a significant community consensus. If this is still the case, I'm good with doing this without waiting for Snarling.

Sphyxis:
I've seen comments as recent as the beginning of this month going over what order to do things in.  It's still being talked about and still an issue afaik.

d0sboots:
Oh, I know for a fact that people doing BN6 by first by "accident" is still an issue (and will continue to be one until we do something about it.) But just because we agree there is a problem, doesn't mean we agree on the solution. XD

I was more referring to, if there is community consensuses that *this* is a good solution to the issue.

PR #1411, opened 2024-06-19
MISC: BitNode options

catloversg:
This PR adds a new feature: BitNode options. When the player enters a BN, they can set custom BitNode options:
- Active level of SF: change the effective level of SFs. For example, if the player has SF12.1000, they can set a custom level and enter a BN without NFG-1000. It also allows the player to create custom "scenarios", e.g., "I have all SFs now. How long do I need to beat BN1 with only SF1, SF4, and SF5?".
- Disable some features. These options are based on the conditions of BN challenges.

Intentional changes:
- `knowsAboutBitverse` is renamed to `knowAboutBitverse` and moved to `BitNodeUtils.ts`. It replaces all checks of `Player.sourceFiles.size > 0`.
- `canAccessBitNodeFeature` is moved to `BitNodeUtils.ts`. It replaces all checks of `Player.bitNodeN === n || Player.sourceFileLvl(n) > 0`.
- All usages of `Player.sourceFileLvl` and `Player.activeSourceFileLvl` are intentional.
- `gainIntelligenceExp`: Remove `this.skills.intelligence > 0`. This check comes from the ancient commit e3991b8795fc9bccd38240e17c99e6d20d53fd4f. It's irrelevant now. The player cannot earn intelligence exp if they are not in BN5 and do not have SF5.
- The refactor of `Sleeve.ts` and `Person.ts`: I moved functions in `[X]Methods.ts` to `[X].ts` when I started to implement the `disableSleeveExpAndAugmentation` option. That change is unnecessary now, so I will revert it if the maintainers want.

### Test case
Normal gameplay: Check SF level and active SF level after entering a new BN:
- Finish the BN:
  - Hack WD via Terminal.
  - `ns.singularity.destroyW0r1dD43m0n`.
  - Bladeburner.
- Bitflume:
  - b1t_flum3.exe.
  - `ns.singularity.b1tflum3`.

Change active SF level:
- Check SF level and active SF level after entering a new BN.
- Check effects of SF. For example, set SF3 to level 0, and check if we can create a corporation outside of BN3.

Change boolean options:
- Restrict Home PC upgrade:
  - Check UI of vendor.
  - `ns.singularity.upgradeHomeRam` and `ns.singularity.upgradeHomeCores`.
- Disable XYZ: Check if XYZ is available. For example:
  - Enter BN3, and check if we can create a corporation.
  - Set SF3 to level 3, enter BN1, and check if we can create a corporation.
- Disable 4S Market Data:
  - Check UI of Stock Market.
  - `ns.stock.purchase4SMarketData` and `ns.stock.purchase4SMarketDataTixApi`.
- Disable Sleeves' exp and augmentation:
  - Check UI of Sleeves.
  - Functions in `ns.formulas.work`.


d0sboots:
High-level comments before I dive into the code:
* Sorry I put this off for a while.
* I like the overall concept!
* The description text confused me at first. Here's my attempt at explaining: "Note: Changing the active level of a SF is temporary, you still permanently own that SF level. For example, if you enter BN 1.3 while having SF 1.2 but with the active level set to 0, you will not get the bonuses from SF1.1 or 1.2 but you will still earn SF1.3 when destroying the BN."
* It would be good to see what the options look like if you *don't* have all SFs yet.
* I think the "Active level" text should only show if it's less than the current max, otherwise it's just noise.

src/Augmentation/ui/SourceFiles.tsx
```diff
@@ -39,57 +57,52 @@ const safeGetSf = (sfNum: number): SourceFile | SfMinus1 | null => {
...
 
 export function SourceFilesElement(): React.ReactElement {
-  const sourceFilesCopy = new Map(Player.sourceFiles);
+  const sourceFileList: SourceFileData[] = [];
+
   const exploits = Player.exploits;
   // Create a fake SF for -1, if "owned"
   if (exploits.length > 0) {
-    sourceFilesCopy.set(-1, exploits.length);
+    sourceFileList.push(getSourceFileData(-1)!);
```
d0sboots: Whaaat syntax is this? Trailing `!` is an unknown operator to me.

Edit: OK, I see this is a TS thing. I've also seen advice not to use it, and I think I agree.

In this case, there is no place that calls `getSourceFileData(-1)` except here. And the handling for `getSourceFileData(-1)` is rather distinct from the handling for a normal SF. So I think it makes more sense to make a dedicated helper for SF -1 that does *not* return null, or just inline that logic here, directly.

src/BitNode/BitNodeUtils.ts
```diff
@@ -9,3 +14,56 @@ export function isBitNodeFinished(): boolean {
...
+  return { valid: true };
+}
+
+export function setBitNodeOptions(bitNodeOptions: BitNodeOptions): void {
+  const validationResultForActiveSourceFiles = validateActiveSourceFiles(bitNodeOptions.activeSourceFiles);
+  if (!validationResultForActiveSourceFiles.valid) {
+    throw new Error(`activeSourceFiles is invalid. Reason: ${validationResultForActiveSourceFiles.message}`);
+  }
+
+  Player.bitNodeOptions.activeSourceFiles = new JSONMap(bitNodeOptions.activeSourceFiles);
```
d0sboots: Why can't you do either `Player.bitNodeOptions = bitNodeOptions` or `Object.assign(Player.bitNodeOptions, bitNodeOptions)`?

src/BitNode/ui/BitNodeAdvancedOptions.tsx
```diff
@@ -0,0 +1,213 @@
...
+  sourceFileLevel,
+  callbacks,
+}: SourceFileButtonRowProps): React.ReactElement {
+  const [sfLevel, setSfLevel] = React.useState<number>(sourceFileLevel);
+  const [inputValue, setInputValue] = React.useState<string>(sourceFileLevel.toString());
+
+  const title = `SF-${sourceFileNumber}`;
+  const sourceFileLevelTool =
+    sourceFileNumber !== 12 ? (
+      Array.from([0, 1, 2, 3]).map((level) => (
```
d0sboots: Shouldn't this be using SourceFileInfo or something?

src/BitNode/ui/PortalModal.tsx
```diff
@@ -21,8 +24,55 @@ export function PortalModal(props: IProps): React.ReactElement {
...
+  if (!props.flume) {
+    const currentSourceFileLevel = Player.sourceFileLvl(props.destroyedBitNode);
+    if (currentSourceFileLevel < 3 || props.destroyedBitNode === 12) {
+      currentSourceFiles.set(props.destroyedBitNode, currentSourceFileLevel + 1);
+    }
+  }
+  currentSourceFiles = new Map([...currentSourceFiles].sort((a, b) => a[0] - b[0]));
+
+  let activeSourceFiles = new Map(currentSourceFiles);
+  let bitNodeBooleanOptions = {
```
d0sboots: Optional: Feels like you could just define these and then call `ResetAll()`.

src/Netscript/NetscriptHelpers.tsx
```diff
@@ -716,3 +724,32 @@ let customElementKey = 0;
 export function wrapUserNode(value: unknown) {
   return <CustomBoundary key={`PlayerContent${customElementKey++}`}>{value}</CustomBoundary>;
 }
+
+function validateBitNodeOptions(ctx: NetscriptContext, bitNodeOptions: unknown): BitNodeOptions {
+  const result = getDefaultBitNodeOptions();
+  if (bitNodeOptions === undefined || bitNodeOptions === null) {
```
d0sboots: You can just do `== null`, which is semi-ubiquitous in our codebase.

src/Netscript/NetscriptHelpers.tsx
```diff
@@ -716,3 +724,32 @@ let customElementKey = 0;
...
+  const options = bitNodeOptions as Unknownify<BitNodeOptions>;
+  if (!(options.activeSourceFiles instanceof Map)) {
+    throw errorMessage(ctx, `activeSourceFiles must be a Map.`);
+  }
+  const validationResultForActiveSourceFiles = validateActiveSourceFiles(options.activeSourceFiles);
+  if (!validationResultForActiveSourceFiles.valid) {
+    throw errorMessage(ctx, `activeSourceFiles is invalid. Reason: ${validationResultForActiveSourceFiles.message}`);
+  }
+
+  result.activeSourceFiles = new Map(options.activeSourceFiles);
```
d0sboots: Feels like this should be a JSONMap? That's strictly more useful, and no more expensive.

src/NetscriptFunctions.ts
```diff
@@ -1757,6 +1764,16 @@ export const ns: InternalAPI<NSFull> = {
     currentNode: Player.bitNodeN,
     ownedAugs: new Map(Player.augmentations.map((aug) => [aug.name, aug.level])),
     ownedSF: new Map(Player.sourceFiles),
+    bitNodeOptions: {
```
d0sboots, line 1810: Unless there's some properties that you're skipping (and if so, those should be commented), you can simplify this to
`{ ...Player.bitNodeOptions, activeSourceFiles: new Map(Player.bitNodeOptions.activeSourceFiles) }`

src/PersonObjects/Person.ts
```diff
@@ -40,17 +42,184 @@ export abstract class Person implements IPerson {
...
-  gainDefenseExp = personMethods.gainDefenseExp;
-  gainDexterityExp = personMethods.gainDexterityExp;
-  gainAgilityExp = personMethods.gainAgilityExp;
-  gainCharismaExp = personMethods.gainCharismaExp;
-  gainIntelligenceExp = personMethods.gainIntelligenceExp;
-  gainStats = personMethods.gainStats;
-  regenerateHp = personMethods.regenerateHp;
-  updateSkillLevels = personMethods.updateSkillLevels;
-  hasAugmentation = personMethods.hasAugmentation;
+  gainHackingExp(exp: number): void {
```
d0sboots: Although this is a potentially worthy refactor, let's save it for a different PR.

src/PersonObjects/Player/PlayerObject.ts
```diff
@@ -181,6 +194,11 @@ export class PlayerObject extends Person implements IPlayer {
         delete player.jobs[loadedCompanyName as CompanyName];
       }
     }
+    // Initialize bitNodeOptions
+    if (!player.bitNodeOptions.initialized) {
```
d0sboots: Ergh. Having an `initialized` flag is something I would really, *really* like to avoid. The moment you do this, *everything* has to check whether it's been initialized already, and it's just awful.

In this case, it seems wholly unneeded as well, because the entirety of the Player object is in the same state: A bunch of valid-but-not-correct default values until fromJSON loads the save, after which it has the correct values. Since an empty map is valid (and also correct, for the default value of `sourceFiles`), I see no reason for this flag to exist.

src/PersonObjects/Sleeve/Sleeve.ts
```diff
@@ -76,8 +80,96 @@ export class Sleeve extends Person implements SleevePerson {
     this.shockRecovery();
   }
 
-  applyAugmentation = sleeveMethods.applyAugmentation;
-  findPurchasableAugs = sleeveMethods.findPurchasableAugs;
+  applyAugmentation(aug: Augmentation): void {
```
d0sboots: Same comment about holding off on this refactor.

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -1682,6 +1684,25 @@ export interface GraftingTask {
...
+}
+
+export interface BitNodeBooleanOptions {
+  restrictHomePCUpgrade: boolean;
+  disableGang: boolean;
+  disableCorporation: boolean;
+  disableBladeburner: boolean;
+  disable4SData: boolean;
+  disableHacknetServer: boolean;
+  disableSleeveExpAndAugmentation: boolean;
```
d0sboots, line 1733: At least this one (probably restrictHomePCUpgrade too) deserves documentation.

src/BitNode/ui/BitNodeAdvancedOptions.tsx
catloversg: I don't understand your question. Can you clarify it?

src/PersonObjects/Player/PlayerObject.ts
catloversg: This check is only useful in `fromJSON`. In `fromJSON`, when we see an empty JSONMap, there are 2 possible cases:
- The player loads an old save, and there is no `bitNodeOptions` in their save data. The empty JSONMap is just a default value, and we need to initialize it with the current `player.sourceFiles`.
- The empty JSONMap is the "real" value. The player intentionally disabled all SFs, so `bitNodeOptions.activeSourceFiles` is empty.

I use `bitNodeOptions.initialized` to determine if we need to initialize `bitNodeOptions.activeSourceFiles`.

Snarling: If it's only useful for old saves, it shouldn't add a new permanent property. One option would be to use savegame conversion code to do the initialization (in SaveObject.ts)

But actually I think it may make sense to rethink the actual structure of the property we're adding to player. Instead of activeSourceFiles, which is an entire clone of the sourceFiles map, it probably makes more sense to just have a sourceFileOverrides (or something like that) which is normally empty except when a player opts to decrease the effective level of their SF. This slims down the player object in the default case (no overrides) and would also mean no initialization required.

src/BitNode/BitNodeUtils.ts
```diff
@@ -9,3 +14,52 @@ export function isBitNodeFinished(): boolean {
   }
   return wd.backdoorInstalled;
 }
+
+export function canAccessBitNodeFeature(bitNode: number): boolean {
+  return Player.bitNodeN === bitNode || Player.activeSourceFileLvl(bitNode) > 0;
+}
+
+export function knowAboutBitverse(): boolean {
+  return Player.bitNodeOptions.activeSourceFiles.size > 0;
```
Snarling: I think this should be using the actual Player.sourceFiles

src/BitNode/ui/BitNodeAdvancedOptions.tsx
d0sboots: I meant instead of hardcoding that everything has 3 SF levels, shouldn't we be getting that from the data map we've created?

d0sboots:
I don't know if you've been following allong in #development, here's some other comments that came from discussion there:

* The toggles should be moved before the detailed SF settings, since they're more likely to be used.
* It would be good to have an "All SFs" setting, even if it just serves to set all the SFs to a specific level without having to click them all manually. But as someone mocked up, it would be nice if you could express the changes as a set of overrides, instead of an individual UI setting for each SF. This plays into what Snarling suggested for the storage representation as well.
* We talked about UI stuff as well, such as using an expand-o box instead of a toggle. I think that would be more natural, but one consideration is how these settings will be preserved (or not) when you beat the BN. Using a toggle means you could save the settings, but have them be default disabled for the next BN.

src/BitNode/BitNodeUtils.ts
d0sboots: I'm not sure I agree? You should be able to disable "knowing about the bitverse" (and the various features that enables) through some means, and turning off all SFs seems like the correct way.

src/BitNode/ui/BitNodeAdvancedOptions.tsx
catloversg: I do it like that to make all SF have the same number of buttons (except SF12). All SF have 3 buttons, but we only enable the relevant ones (with data in the map).

catloversg:
New UI:
Capture

Advanced options are cleared when the player switches between targeted BitNodes. They are also not persisted in Settings. This is intentional. IMO, a minor inconvenience is better than "Oops! I forgot resetting those options!".

d0sboots:
UI is starting to look pretty good, or at the minimum, good compared to many of our UIs XD

I agree with your reasoning on not saving the settings. Especially with the new-style of adding individual SFs.

Some other UI thoughts: We should add tooltips for the toggles. What, exactly, "Restrict home PC upgrade" entails isn't clear without extra words. Extra words can also make it clear that "Disable hacknet servers" means "Reenable hacknet nodes".

For the buttons, IMO we should do more graphically and less with text. I'm assuming "level" means "current owned level of this SF" - that should be expressed by number of buttons. (There's no point having a button you can't click on.) If you don't have the SF, you shouldn't be able to add it, and if you only have 1 level you only get one button. 

Similarly, the currently active level should use a bright highlight or something - standard "button selected" stuff. That eliminates the need for the extra text and IMO will be easier to visually understand.

d0sboots:
One other suggestion that I needed to read the code first to make: IIUC, the "Set all SF" button isn't a setting of its own, but rather it immediately creates/sets the SF setting for all possible values. It's not what I was initially expecting, but I think it's fine. However, in that case it should get an additional "X" or trashcan button to remove all SF settings so that the state is back to default.

jjclark1982:
Should there be a toggle to enter a BitNode with no Intelligence (without disabling SF5 features)? And how should the results merge with previous Intelligence?

A reasonable scenario for comparing speed-runs would be "How long does it take to complete BitNode N with all features unlocked and no starting Intelligence?" so disabling SF5 is not a great option as it removes Formulas.exe at start and Intelligence gain within the node, which are changes to strategy.

catloversg:
New UI:
Capture

UI changes:
- `Restrict Home PC upgrade` -> `Restrict max RAM and core of Home PC`. Change the tooltip text.
- `Disable Hacknet Server`: Change the tooltip text.
- Add `Remove all` + tooltip.
- Add tooltip for `Set all SF`.
- Only show relevant buttons.
- Highlight the active level.

catloversg:
> Should there be a toggle to enter a BitNode with no Intelligence (without disabling SF5 features)? And how should the results merge with previous Intelligence?
> 
> A reasonable scenario for comparing speed-runs would be "How long does it take to complete BitNode N with all features unlocked and no starting Intelligence?" so disabling SF5 is not a great option as it removes Formulas.exe at start and Intelligence gain within the node, which are changes to strategy.

I can add "Disable Intelligence bonus". That option will only disable the bonus of Intelligence (`calculateIntelligenceBonus`). I'll add it if d0sboots and Snarling agree.

d0sboots:
> I can add "Disable Intelligence bonus". That option will only disable the bonus of Intelligence (calculateIntelligenceBonus). I'll add it if d0sboots and Snarling agree.

That seems like a good way to handle it: you still gain int (if applicable), it just has no effect. It might need a UI bit though, otherwise it'll look like you have normal int and won't be obvious that it's disabled. Maybe strikethrough where int is displayed? 

I think @Snarling has yet to sign off on the overall design, but I would encourage him to (I'm not going to merge this one without some kind of thumbs up).

d0sboots (approved):
In terms of code and UI, I'm satisfied with how this looks. Nice job!

catloversg:
Snarling mentioned that they would release v2.6.2 soon. This PR is a bit large, so it's best to merge it after that. I want it to be tested in the dev branch for a bit longer.

src/BitNode/ui/BitNodeAdvancedOptions.tsx
```diff
@@ -0,0 +1,355 @@
...
+            <tbody>
+              <tr>
+                <td>
+                  <Tooltip title="Set active level for all chosen SF">
+                    <Typography minWidth="7rem">Set all SF</Typography>
+                  </Tooltip>
+                </td>
+                <td>
+                  <ButtonGroup>
+                    {Array.from([0, 1, 2, 3]).map((level) => (
```
Snarling: `[0, 1, 2, 3]` is already an array
```suggestion
                    {[0, 1, 2, 3].map((level) => (
```

src/BitNode/BitNodeUtils.ts
```diff
@@ -9,3 +14,63 @@ export function isBitNodeFinished(): boolean {
   }
   return wd.backdoorInstalled;
 }
+
+export function canAccessBitNodeFeature(bitNode: number): boolean {
+  return Player.bitNodeN === bitNode || Player.activeSourceFileLvl(bitNode) > 0;
+}
+
+export function knowAboutBitverse(): boolean {
+  for (const [, sfActiveLevel] of Player.activeSourceFiles) {
```
Snarling: Just a style thing, but when only iterating one side of the map I would recommend using .keys()/.values() instead of using a partial array
```suggestion
  for (const sfActiveLevel of Player.activeSourceFiles.values()) {
```

Snarling (approved):
Looks good, nice work! I just had two extremely minor code style comments while looking through code.

jjclark1982:
A "Disable Intelligence bonus" option would be really helpful for measuring strategy effectiveness and fairly comparing speed-runs.

@Snarling what's your opinion about that option?

Snarling:
> A "Disable Intelligence bonus" option would be really helpful for measuring strategy effectiveness and fairly comparing speed-runs.
> 
> @Snarling what's your opinion about that option?

This would be a good option to add and it would fit in here.

Another option, though slightly harder to implement, would be allowing simulation of different int levels. Checkbox for whether to override intelligence -> enables a numeric input to set the override value.

Internally it could just be a `number | undefined` override, with undefined if that checkbox wasn't checked. If checked, the number would need to be validated as a non-negative integer that is <= the player's intelligence.

Overriding in calculateIntelligenceBonus is a little tricky though, because that is used for both player and sleeves. e.g. if the player is overriding their 1000 int to 50, but their sleeve only has 5 int, then the sleeve should still use 5. It's unclear if the sleeve should be limited by the same override (e.g. another sleeve with 51 int, should it use 50 or 51)

src/BitNode/ui/PortalModal.tsx
```diff
@@ -94,6 +99,7 @@ export function PortalModal(props: IProps): React.ReactElement {
         onClick={() => {
           const bitNodeOptions = {
             sourceFileOverrides: sourceFileOverrides,
+            intelligenceOverride: intelligenceOverride,
```
d0sboots: You can use the shorter object literal syntax for these, where you just list the keys, when the key=value name.

src/ui/React/CharacterOverview.tsx
```diff
@@ -118,6 +118,16 @@ export function Val({ name, color }: ValProps): React.ReactElement {
     return clearSubscription;
   }, [name]);
 
+  if (name === "Int" && Player.bitNodeOptions.intelligenceOverride !== undefined) {
+    return (
+      <Tooltip title={`Intelligence Override: ${formatSkill(Player.bitNodeOptions.intelligenceOverride)}`}>
```
d0sboots: Optional: I think the display of overridden int should be a bit more prominent. Maybe something like "X (Y)", where X is the overridden value and Y is the real value, but only if Y > X. I could see that having issues with fitting in the stats overview, though. 

I do think the overridden value is the more important one to show, since it is what will be used for calculations. But again, only if Y > X.

If this is too complicated/annoying, it can be deferred to a follow-up. The int feature is complicated, and I don't want it to block the overall PR more than it already has.

src/BitNode/ui/BitNodeAdvancedOptions.tsx
```diff
@@ -266,12 +268,40 @@ export function BitNodeAdvancedOptions({
...
+        <Typography display="flex" alignItems="center">
+          Override Intelligence:
+        </Typography>
+        <TextField
+          sx={{ maxWidth: "4rem" }}
+          disabled={intelligenceOverride === ""}
+          value={intelligenceOverride}
+          onChange={(event) => {
+            if (event.target.value === "") {
+              setIntelligenceOverride("0");
```
d0sboots: This feels like a classic case of having state in two places. You should just push the useState up into PortalModal, and pass it down via props.

src/BitNode/ui/BitNodeAdvancedOptions.tsx
```diff
@@ -342,6 +372,44 @@ export function BitNodeAdvancedOptions({
             text="Disable Sleeves' experience and augmentation"
             tooltip="Sleeves cannot gain experience or install augmentations"
           />
+          <OptionSwitch
+            disabled={Player.skills.intelligence <= 0}
+            checked={false}
+            onChange={(value) => {
+              if (!value) {
+                setIntelligenceOverride("");
```
d0sboots: UI-wise, this is a bit awkward: It feels strange for the value in the input to be changed when the option is toggled, rather than just being disabled. 

I think you want the input to start with skills.intelligence as the default value (and be disabled), but it can't be coupled so tightly to the value of the BitnodeOption.

d0sboots (commented):
Some things. 

Overall, I feel like adding the int overrides might have made more sense as a separate PR. Oh well, too late now. Anyway, I'd like to address the most important points now, and we can iterate on the rest later, as people leave feedback.

src/ui/React/CharacterOverview.tsx
catloversg: I made a small change based on your suggestion. I'll make further changes in a follow-up PR if needed.

d0sboots:
Looks good. We can continue to iterate on other stuff in followup(s), if it comes up.

PR #1412, opened 2024-06-20
BUGFIX: Hacknet level base cost multiplier fix

### LINKED ISSUE #746: Hacknet level base cost is 500 but first level purchase is 520, 1 multiplier ahead.
As opposed to first ram and core purchase being their base costs of 30k and 500k respectively.

JamesWilcox-git:
Fixes #746


d0sboots:
Whoops, this fell off my radar somehow.

PR #1414, opened 2024-06-20
INFILTRATION: Handle automated infiltration

catloversg:
Infiltration has never been balanced to be played by an automation script. In our code base, we prohibit it with the checks in `src\Infiltration\ui\Game.tsx` and `src\Infiltration\ui\KeyHandler.tsx`, but our checks are wrong:
- `press`: It does not check if `event` is a "real" event or a synthetic one.
- `press`: It does not call `onFailure({ automated: true })`, so the check `options?.automated` in `onFailure` is useless.

With these bugs, our standpoint on automated infiltration is weird. We prohibit it, but we don't do anything to prevent it. There are 3 ways to deal with it:
- Fix those bugs. Make it clear that infiltration is a non-automatic feature. [1]
- Balance it properly and allow it [2]. For example, we can add special restrictions if we detect automation:
  - Lower reward.
  - Random chance of failure + higher damage.
- Revamp infiltration from scratch. [3]

[3] is a big overhaul, so that's not the job of this PR. This PR uses [1]. This is a controversial change because the automated infiltration script is widespread, and many people like it. If the maintainers want, I will change it to [2].

### DIFF
--- a/src/Infiltration/ui/Game.tsx
+++ b/src/Infiltration/ui/Game.tsx
@@ -97,7 +97,11 @@ export function Game(props: GameProps): React.ReactElement {
       if (options?.automated) {
         damage = Player.hp.current;
         setTimeout(() => {
-          SnackbarEvents.emit("You were hospitalized. Do not try to automate infiltration!", ToastVariant.WARNING, 5000);
+          SnackbarEvents.emit(
+            "You were hospitalized. Do not try to automate infiltration!",
+            ToastVariant.WARNING,
+            5000,
+          );
         }, 500);
       }
       if (Player.takeDamage(damage)) {
--- a/src/Infiltration/ui/Game.tsx
+++ b/src/Infiltration/ui/Game.tsx
@@ -1,6 +1,6 @@
 import { Button, Container, Paper, Typography } from "@mui/material";
 import React, { useCallback, useState } from "react";
-import { FactionName } from "@enums";
+import { FactionName, ToastVariant } from "@enums";
 import { Router } from "../../ui/GameRoot";
 import { Page } from "../../ui/Router";
 import { Player } from "@player";
@@ -15,6 +15,7 @@ import { SlashGame } from "./SlashGame";
 import { Victory } from "./Victory";
 import { WireCuttingGame } from "./WireCuttingGame";
 import { calculateDamageAfterFailingInfiltration } from "../utils";
+import { SnackbarEvents } from "../../ui/React/Snackbar";
 
 type GameProps = {
   StartingDifficulty: number;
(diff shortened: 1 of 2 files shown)


d0sboots:
My personal stance has always been to lean towards (2). For instance, infil rewards could be scaled based on how many successful infiltrations you've done in the past 15 minutes. Done properly, it wouldn't affect manual infil rewards, while reducing the profit of automated infil to balancable levels. This would open the door to infil eventually becoming an official singularity API.

I think @Snarling needs to weigh in here.

otac0n:
As a player, I would really prefer to have some built-in ways to develop this myself rather than downloading someone else's script. I 100% expected to be able to do it as part of singularity.

Snarling:
It's already clear that automating infiltration is not officially supported:

1. There's no API for it
2. Bypassing the anti-automation protections is not impossible but does require specific workarounds.

Why not beef up the anti-automation protections?

1. It is not realistic to patch every possible way the player can access game internals, so our general stance on exploits is not to try and patch them out.
2. Bypassing game protections is fun for players, we don't need to make everything as bulletproof as possible.\\

Why not officially support automation?

1. An API doesn't really make sense for it. If the game state is clearly communicated to the player through an API, then it becomes extremely straightforward to program in the correct response to win a given game. None of the games present any difficulty when given a programmable API.
2. The main challenge of an automated infiltration script (especially if not taking "inspiration" from an existing one) is figuring out how to scrape the game state from the document, and how to spoof the events to play the games.

Other thoughts:

1. Infiltration is currently too strong to allow automation via an official API.
2. The strength of automated infiltration makes players feel more accomplished when they are able to successfully automate it.
3. If automating it is not officially supported, it's fine for it to have the high rewards it currently does, because it's effectively an exploit.

Snarling:
The changes in this PR can also be worked around, so I'd be fine with merging this just to spice things up and give an extra hurdle to players who want to automate infiltration.

src/Infiltration/ui/Game.tsx
```diff
@@ -91,11 +92,14 @@ export function Game(props: GameProps): React.ReactElement {
...
-      // it's clear they're not meant to
-      const damage = options?.automated
-        ? Player.hp.current
-        : calculateDamageAfterFailingInfiltration(props.StartingDifficulty);
+      let damage = calculateDamageAfterFailingInfiltration(props.StartingDifficulty);
+      // Kill the player immediately if they use automation, so it's clear they're not meant to
+      if (options?.automated) {
+        damage = Player.hp.current;
+        setTimeout(() => {
+          SnackbarEvents.emit("You was hospitalized. Do not try to automate infiltration!", ToastVariant.WARNING, 5000);
```
Snarling:
```suggestion
          SnackbarEvents.emit("You were hospitalized. Do not try to automate infiltration!", ToastVariant.WARNING, 5000);
```
Just a typo / grammar fix

PR #1417, opened 2024-06-20
GANG: Show equipments when there is not enough money

### LINKED ISSUE #1409: Gang Equipment | Annoying Wiggling Interface
### TLDR
At Gang > Equipment, the interface wiggles based on money available.
To fix that, equipment should be shown as disabled (not hidden) when the money available is not enough.

### Problem
Sometimes I'm trying to manage my Gang's equipment.
But as my money oscillates, so does the interface. Making it very hard to use the UI.

With a good intention, the Gang > Equipment interface will only show the equipment you can afford.
But that causes the whole interface to shift up and down as my money triggers the next equipment available.
Screen shift (without user's direct input) is very annoying.
And it's a bad UI pattern: This even makes websites rank lower at Google.

This is specially critical at mid-end-game, when my gang is at "turbo" mode, generating billions per second. While my stocks algorithm is consuming those billions. So it's basically impossible to buy an augmentation.

### Solution

I'd like to have every equipment available at my screen, no matter how much money I have.
(of course, only those I haven't bought yet and inside the category selected)

In short:
All equipment I cannot buy should be shown as a disabled button, instead of hidden.
Never hide a button based on external dynamic volatile factors (such as money available).

catloversg:
Screenshot:

Fixes #1409.

### DIFF
--- a/src/Gang/ui/EquipmentsSubpage.tsx
+++ b/src/Gang/ui/EquipmentsSubpage.tsx
@@ -21,31 +21,6 @@ import { Settings } from "../../Settings/Settings";
 import { StatsRow } from "../../ui/React/StatsRow";
 import { useRerender } from "../../ui/React/hooks";
 
-interface INextRevealProps {
-  upgrades: string[];
-  type: UpgradeType;
-}
-
-function NextReveal(props: INextRevealProps): React.ReactElement {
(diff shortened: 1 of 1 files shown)


d0sboots:
Always love it when a UI improvement makes the code simpler

PR #1422, opened 2024-06-21
INFILTRATION: Improve accuracy of slash game UI

### LINKED ISSUE #1402: Shadows of Anarchy Ares augmentation doesn't always work.
image

I'm not completely sure how to reproduce it. I had trouble with this a few days ago, but it was working yesterday. Then it started again today. I'm just doing manual infiltrations, and the timer provided by the Ares augmentation is wrong.

Playing on v2.6.2dev (18ae6ce21).

catloversg:
New change: If the player has `MightOfAres`, the logic of phase 1 only starts after the "special" `GameTimer` expires.

Fixes #1402.

### DIFF
--- a/src/Infiltration/ui/SlashGame.tsx
+++ b/src/Infiltration/ui/SlashGame.tsx
@@ -28,23 +28,24 @@ const difficulties: {
 export function SlashGame({ difficulty, onSuccess, onFailure }: IMinigameProps): React.ReactElement {
   const [phase, setPhase] = useState(0);
   const timeOutId = useRef<number | ReturnType<typeof setTimeout>>(-1);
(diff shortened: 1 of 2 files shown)


src/Infiltration/ui/SlashGame.tsx
```diff
@@ -27,34 +27,52 @@ const difficulties: {
...
-    setPhase(0);
-    setGuardingTime(guardingTime);
-    setHasAugment(Player.hasAugmentation(AugmentationName.MightOfAres, true));
+    return {
+      hasAugment: Player.hasAugmentation(AugmentationName.MightOfAres, true),
+      guardingTime,
+      distractedTime,
+      alertedTime,
+    };
+  }, [difficulty]);
```
d0sboots: You're memoizing the augment state here... isn't it possible for it to change without this component getting re-rendered as a result?

PR #1425, opened 2024-06-21
SINGULARITY: Add ns.singularity.getFactionWorkTypes

### LINKED ISSUE #1420: The function `ns.singularity.getFactionWorkTypes()` is missing?
As of the most recent version of the game, there is no API for determining faction work types, aside from attempting to join them.  This can cause problems if it interrupts some long running work activity (like a slow crime or something else).

For example, you can do something like this:

    let factionWorkTypes = {};
    export function nsGetFactionWorkTypes(ns: NS, faction: string) {
      if (factionWorkTypes.hasOwnProperty(faction)) {
        return factionWorkTypes[faction];
      }
      if (!ns.getPlayer().factions.includes(faction)) return null;
      return factionWorkTypes[faction] = workTypes.map(w => ns.singularity.workForFaction(faction, w, false));
    }

This will cache the result, but it will interrupt existing jobs (with potentially large downsides).

catloversg:
Fixes #1420.


otac0n:
Thank you!

Issue #1009, opened 2024-01-01
Feature request: Obtain estimated success chance for sleeves performing Bladeburner Contracts in NS scripts


cigarmemr:
As far as I've discovered, there is no way to obtain success chance estimation for sleeves performing Bladeburner Contracts. This renders some critical judgement like "Order the sleeve to stop while success chance drops too low" impossible.

Maybe an optional argument in `getActionEstimatedSuccessChance()` is good. Leaving it unset to obtain success chance for player while passing an index for the corresponding sleeve.

gmcew:
Closed by #1428

PR #1428, opened 2024-06-22
API: Get Sleeve Success at BB tasks through existing commands

Sphyxis:
Tweaked the existing BB command of getActionEstimatedSuccessChance(type, name) to also take an optional sleeve number.  With the number added it will return the sleeves success high and low chance of contracts only, (Tracking, Investigation, Bounty Hunter) returning [0, 0] otherwise.

This does not break any existing scripts as it's an optional parameter.

Tested with my sleeve 0 and verified that I received different numbers for myself and my sleeve.

### DIFF
--- a/src/NetscriptFunctions/Bladeburner.ts
+++ b/src/NetscriptFunctions/Bladeburner.ts
@@ -27,6 +27,21 @@ export function NetscriptBladeburner(): InternalAPI<INetscriptBladeburner> {
       throw helpers.errorMessage(ctx, "You must be a member of the Bladeburner division to use this API.");
     return bladeburner;
   };
+  const checkSleeveAPIAccess = function (ctx: NetscriptContext) {
+    if (Player.bitNodeN !== 10 && !Player.sourceFileLvl(10)) {
+      throw helpers.errorMessage(
+        ctx,
+        "You do not currently have access to the Sleeve API. This is either because you are not in BitNode-10 or because you do not have Source-File 10",
+      );
+    }
+  };
+  const checkSleeveNumber = function (ctx: NetscriptContext, sleeveNumber: number) {
+    if (sleeveNumber >= Player.sleeves.length || sleeveNumber < 0) {
+      const msg = `Invalid sleeve number: ${sleeveNumber}`;
+      helpers.log(ctx, () => msg);
+      throw helpers.errorMessage(ctx, msg);
+    }
+  };
 
   function getAction(ctx: NetscriptContext, type: unknown, name: unknown): Action {
     const bladeburner = Player.bladeburner;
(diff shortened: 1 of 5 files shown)


gmcew:
Also closes #1009

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -3114,9 +3114,10 @@ export interface Bladeburner {
    *
    * @param type - Type of action.
    * @param name - Name of action. Must be an exact match.
+   * @param sleeve - Optional.  Sleeve number to check for success.
    * @returns Estimated success chance for the specified action.
    */
-  getActionEstimatedSuccessChance(type: string, name: string): [number, number];
+  getActionEstimatedSuccessChance(type: string, name: string, sleeve: number = undefined): [number, number];
```
```suggestion
  getActionEstimatedSuccessChance(type: string, name: string, sleeve?: number): [number, number];
```

src/NetscriptFunctions/Bladeburner.ts
```diff
@@ -117,10 +132,20 @@ export function NetscriptBladeburner(): InternalAPI<INetscriptBladeburner> {
         1000
       );
     },
-    getActionEstimatedSuccessChance: (ctx) => (type, name) => {
+    getActionEstimatedSuccessChance: (ctx) => (type, name, _sleeve) => {
       const bladeburner = getBladeburner(ctx);
       const action = getAction(ctx, type, name);
-      return action.getSuccessRange(bladeburner, Player);
+      if (_sleeve !== undefined) {
```
d0sboots: I think I'd check against `!= null` here to work with both `undefined` and `null`, in case someone passes `null` directly.

src/NetscriptFunctions/Bladeburner.ts
```diff
@@ -27,6 +27,21 @@ export function NetscriptBladeburner(): InternalAPI<INetscriptBladeburner> {
       throw helpers.errorMessage(ctx, "You must be a member of the Bladeburner division to use this API.");
     return bladeburner;
   };
+  const checkSleeveAPIAccess = function (ctx: NetscriptContext) {
```
d0sboots: I'm pretty sure this function already exists elsewhere... you should import that version.

d0sboots (commented):
Nice! Always good when there's a compatible way to extend these things.

src/NetscriptFunctions/Bladeburner.ts
Sphyxis: @d0sboots 
Everything has been updated.  I was going to export the last 2 but they were defined within a function already.  Since you've given permission though, I've removed those 2 from their main function, defined them above it and exported them.

PR #1430, opened 2024-06-22
CORPORATION: Don't check access for `getConstants`

ilkecan:
I think it is reasonable to want to call `getConstants` before creating a corporation. I also couldn't find a technical reason to disallow it.


d0sboots:
Yeah, seems reasonable.

PR #1431, opened 2024-06-22
TERMINAL: Tweaks and bugfixes to grep

muesli4brekkies:
Halloo.

This is mainly a rework of the arg-parsing. It's now a lot stricter, as previously it was quite slack and would just pass mixed/mispelled flags if they contained a key character.

The typing of the main options is also now far more programmatic, and no longer reliant on careful matching of magic strings.

Also fixed the terminal printing limit, which was just busted!

:blush:


src/Terminal/commands/grep.ts
```diff
@@ -13,20 +13,40 @@ const DEFAULT: string = "\x1b[0m";
 const GREEN: string = "\x1b[32m";
 const MAGENTA: string = "\x1b[35m";
 const CYAN: string = "\x1b[36m";
+const YELLOW: string = "\x1b[33m";
 const WHITE: string = "\x1b[37m";
 
-// Options and ValidArgs key names must correlate
-class ArgStrings {
+type Errors = {
```
d0sboots: Do you need to manually define this? The types exist in the `ERR` object, and you can pull the types if you need to via `typeof`/`keyof` etc.

src/Terminal/commands/grep.ts
```diff
@@ -13,20 +13,40 @@ const DEFAULT: string = "\x1b[0m";
...
+  noArgs: string;
+  noSearchArg: string;
+  badArgs: (str: string[]) => string;
+  badParameter: (opt: string, arg: string) => string;
+  badOutFile: (str: string) => string;
+  outFileExists: (str: string) => string;
+  truncated: () => string;
+};
+
+const ERR: Errors = {
```
d0sboots: Optional: Is there a benefit to defining these within an object, instead of having them just be top-level constants?

src/Terminal/commands/grep.ts
```diff
@@ -101,44 +101,31 @@ const VALID_ARGS: ValidArgs = {
...
-  outFileExists: (path: string) =>
-    `grep file output failed: Invalid output file "${path}". Output file must not already exist. Pass -f/--allow-overwrite to overwrite.`,
-  badOutFile: (path: string) =>
-    `grep file output failed: Invalid output file "${path}". Output file must be a text file.`,
-  truncated: () => `\n${RED}Terminal output truncated to ${Settings.MaxTerminalCapacity} lines (Max terminal capacity)`,
+  isMultiFile: { short: [], long: [] },
+  hasContextFlag: { short: [], long: [] },
 } as const;
 
+type Params = {
```
d0sboots: Instead of naming this type, and creating INIT_PARAMS later, you can just directly assign params their values. Saves a lot of typing, makes the values obvious, avoids declaring types needlessly. Same with INIT_OPTIONS. You might still establish a type alias for `type Options = typeof Args.options` (I think that works).

src/Terminal/commands/grep.ts
```diff
@@ -166,86 +153,100 @@ class Args {
...
+    ])) {
+      this.params[key as keyof Params] = this.spliceParam(args);
+    }
+    return this;
+  }
+
+  private reduceToOptionsAndFiles(): string[] {
+    const stripDash = (arg: string) => arg.slice(1);
+    const argKeys = Object.keys(VALID_ARGS).map((k) => k as keyof Options);
+    const [allValidArgs, validFlagChars] = argKeys.reduce(
```
d0sboots: Optional: Once again, I think this would be more readable as a for-of, rather than a reduce with a state of a 2-length list XD

src/Terminal/commands/grep.ts
```diff
@@ -166,86 +153,100 @@ class Args {
...
+      (ret: [readonly string[], string], key: keyof Options): [readonly string[], string] => {
+        const argStrings = VALID_ARGS[key];
+        return [
+          [...ret[0], ...argStrings.long, ...argStrings.short],
```
d0sboots: Oh yeah, especially since you're doing the O(n^2) thing with list re-concatenation here again.

src/Terminal/commands/grep.ts
muesli4brekkies: But muh functionalisation!

No argument from me here :) Standby.

src/Terminal/commands/grep.ts
muesli4brekkies: I did recognise this repeated smelly-pattern from before, but figured it was acceptable in this instance as these arrays are fixed readonly and short. 

It will be tweaked RE the above review though :)

src/Terminal/commands/grep.ts
muesli4brekkies: No reason besides the slightly nicer syntax of the `Terminal.error(ERR.errorName);` calls.

src/Terminal/commands/grep.ts
muesli4brekkies: For sure I will look into tightening this up as you suggest. The Params object was just slapped on during refactor, repeating the Options pattern.

src/Terminal/commands/grep.ts
```diff
@@ -185,16 +175,15 @@ class Args {
...
-        ];
-      },
-      [[], ""],
-    );
+    const allValidArgs: string[] = [];
+
+    let validFlagChars = "";
+    for (const key of argKeys) {
+      const argString = VALID_ARGS[key];
+      allValidArgs.push(...[...argString.long, ...argString.short]);
```
d0sboots: `push` can take multiple arguments, in fact that is what `...` is doing. So you can just do away the outer array here.

Edit: Sorry for the long back-and-forth here, I wanted to just merge but I couldn't not comment on this. XD

muesli4brekkies: Please don't apologise. :) My code can withstand many reviews!

I *think* it was broken as well haha. It "should" be good now...

src/Terminal/commands/grep.ts
```diff
@@ -163,24 +160,26 @@ class Args {
...
+    for (const [key, validArgs] of Object.entries(VALID_PARAMS)) {
+      this.params[key as keyof Parameters] = this.spliceParam(validArgs);
     }
     return this;
   }
 
   private reduceToOptionsAndFiles(): string[] {
     const stripDash = (arg: string) => arg.slice(1);
     const argKeys = Object.keys(VALID_ARGS).map((k) => k as keyof Options);
+    const paramKeys = Object.keys(VALID_PARAMS).map((k) => k as keyof Parameters);
```
d0sboots, line 172: What do these maps do? They look like identity maps (i.e. no-ops) to me.

You're just using these to iterate over VALID_PARAMS/VALID_ARGS, and you're looking up the entry there as well. It would be better to `for (const value of Object.values(VALID_ARGS))` and/or `for (const [key, value] of Object.entries(VALID_ARGS))`down below, instead.

d0sboots:
Gonna go ahead and commit this which seems to be in an OK state, I suspect you will be refactoring it more XD

muesli4brekkies:
No doubt. :) Thanks again for lending your experience and knowledge.

PR #1433, opened 2024-06-24
CONTRACT: Update Generate IP Address contract description to clarify '0'

### LINKED ISSUE #1432: Description: Generate IP Address contract description nitpick.
>Given the following string containing only digits, return an array with all possible valid IP address combinations that can be created from the string:
>
> 797100113
>
> Note that an octet cannot begin with a '0' unless the number itself is actually 0. For example, '192.168.010.1' is not a valid IP.
>
> Examples:
>
> 25525511135 -> ["255.255.11.135", "255.255.111.35"]
> 1938718066 -> ["193.87.180.66"]

solutions/genip.js: 797100113
solutions/genip.js: ["7.97.100.113","79.7.100.113","79.71.00.113"]

In this case, `79.71.00.113` is not valid although it meets the definitions of the rules.

The text `Note that an octet cannot begin with a '0' unless the number itself is actually 0` should be changed.  Either to have quotes around the second zero, or to specifically rule out strings of length greater than one.

otac0n:
closes #1432

### DIFF
--- a/src/data/codingcontracttypes.ts
+++ b/src/data/codingcontracttypes.ts
@@ -492,7 +492,7 @@ export const codingContractTypesMetadata: ICodingContractTypeMetadata[] = [
         "that can be created from the string:\n\n",
         `${data}\n\n`,
         "Note that an octet cannot begin with a '0' unless the number",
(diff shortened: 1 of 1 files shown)


d0sboots:
Exactly the reword I would have chosen.

PR #1435, opened 2024-06-24
GRAFTING: Add new api for checking ongoing grafting

### LINKED ISSUE #1012: getAugmentationGraftTime returns incorrect value

catloversg:
This PR implements d0sboots's suggestion in #1027.

Closes #1012.


markdown/bitburner.grafting.getaugmentationgrafttime.md
```diff
@@ -4,7 +4,7 @@
 
 ## Grafting.getAugmentationGraftTime() method
 
-Retrieves the time required to graft an aug.
+Retrieves the time required to graft an aug. Do not use this value to determine when the ongoing grafting finishes. The ongoing grafting is affected by current intelligence level and focus bonus. You should use  for that purpose.
```
Kelenius: Looks like the link didn't work properly.

d0sboots:
I'm going to bring up one potential quibble, before this gets locked in stone: Should it reject if the current work is not grafting work?

I can see arguments both ways - on one hand, many players won't be coding to handle promise rejections, but on the other hand, maybe that's a good thing, since if a different type of work is happening, that *should* be an exceptional case. (I would leave "no work" as an instant resolve.)

catloversg:
> Should it reject if the current work is not grafting work?

It makes sense. I considered that behavior at first, but scraped that idea because many people don't like exception/promise rejection.

PR #1437, opened 2024-06-25
UI: Split money earned in stats into income and expenses

Kelenius:
image

See: https://github.com/danielyxie/bitburner/pull/4083. Uses the same approach but different code. I'm going to leave this as a draft because I'm looking for feedback regarding the netscript functions:

Currently, I want to deprecate `ns.getMoneySources()` (so it'll still exist and function like it used to, but give a warning), and replace it with `ns.getMoneyIncome()`, `ns.getMoneyExpenses()`, and `ns.getMoneyTotal()`, which accept two parameters: name of the money source you're looking for, and boolean true/false for money source since aug install/start of the bitnode, and return a single number. Would that be a good approach, or would it be better to make a function similar to `ns.getMoneySources()` that gives you all information in one object?


d0sboots:
I think for compatibility it would be best to have a function that returns a single large object. I think that would also be easiest for most people to work with.

Also, if you go with the "single object" approach, I think it can be done in such a way that `ns.getMoneySources()` can still *be* that object?

PR #1443, opened 2024-06-27
BLADEBURNER: Change formula of skill cost

catloversg:
## Problem
For a long time, `calculateCost` only calculated the cost of the next level. Starting from 908d5e95707bde21faa170b8c8945d8e27a58223, it supported an arbitrary number of levels. However, it still has 2 problems:
- The weird recursive mode for a small `count`. It can be rewritten with only 1 simple loop.
- When `count` is greater than 100, it uses a questionable formula. That formula only calculates an approximation instead of an exact value, and that approximation is not even a good one.

Formulas:

[1]:
$$CostOfNextLevel = \lfloor ((BaseCost + CurrentLevel \ast CostInc) \ast Mult) \rfloor$$

[2]:
$$Cost = \sum_{i = 0}^{Count-1}\lfloor ((BaseCost + (CurrentLevel + i) \ast CostInc) \ast Mult) \rfloor$$

≡

$$Cost = \sum_{i = CurrentLevel}^{CurrentLevel+Count-1}\lfloor ((BaseCost + i \ast CostInc) \ast Mult) \rfloor$$

[1] contains a floor function, so it's really hard to find the closed form of [2].

### Solution 1

Instead of using a recursive function and fall back to the "approximation" formula when `count` is too big, we use a simple loop. This is the simplest solution. It has 2 drawbacks:
- It cannot support a big `count`. This is not a problem, unless the player is in deep BN12 or a long intelligent-grinding session.
- It's really hard to answer the question: "How many skill levels can I buy with X skill points?".

### Solution 2

When we calculate the sum, we can use the floor function at the last step instead of all iteration steps.

[3]:
$$Cost = \lfloor\sum_{i = CurrentLevel}^{CurrentLevel+Count-1} ((BaseCost + i \ast CostInc) \ast Mult) \rfloor$$

**[3] is not equivalent to [2].**

It's easy to find [3]'s closed form, so this solution has 2 upsides:
- It can support a big `count`.
- It's easy to answer the question: "How many skill levels can I buy with X skill points?".

The drawback of this solution is that making many small purchases may be cheaper than making a big purchase. For example, (0->3) + (3->6) is cheaper than (0->6).

[3]'s closed form:

$$Cost = \lfloor\sum_{i = CurrentLevel}^{CurrentLevel+Count-1} ((BaseCost + i \ast CostInc) \ast Mult) \rfloor$$

≡

$$Cost = \lfloor\sum_{i = CurrentLevel}^{CurrentLevel+Count-1} (BaseCost \ast Mult + i \ast CostInc \ast Mult) \rfloor$$

≡

$$Cost = \lfloor (Count \ast BaseCost \ast Mult) + (CostInc \ast Mult \ast \sum_{i = CurrentLevel}^{CurrentLevel+Count-1} (i)) \rfloor$$

≡

$$Cost = \lfloor (Count \ast BaseCost \ast Mult) + (CostInc \ast Mult \ast \frac{Count \ast (2 \ast CurrentLevel + Count - 1)}{2}) \rfloor$$

≡

$$Cost = \lfloor (Count \ast BaseCost \ast Mult) + (CostInc \ast Mult \ast (Count \ast CurrentLevel + \frac{Count \ast (Count - 1)}{2})) \rfloor$$

This PR uses solution 2.

**Edit**: we decide to use `Math.round` instead of `Math.floor`. The final formula:

$$Cost = \lceil (Count \ast BaseCost \ast Mult) + (CostInc \ast Mult \ast (Count \ast CurrentLevel + \frac{Count \ast (Count - 1)}{2})) \rceil$$

≡

$$Cost = \lceil Count \ast Mult \ast (BaseCost + (CostInc \ast (CurrentLevel + \frac{Count - 1}{2}))) \rceil$$

### DIFF
--- a/src/Bladeburner/Skill.ts
+++ b/src/Bladeburner/Skill.ts
@@ -39,6 +39,9 @@ export class Skill {
      * The cost of the next level: (baseCost + currentLevel * costInc) * mult. The cost needs to be an integer, so we
      * need to use Math.floor or Math.round.
      *
+     * Note: there is no notation for Math.round, so I use \lceil and \rceil as alternatives for non-existent \lround
+     * and \rround. When you see \lceil and \rceil, it means Math.round, not Math.ceil.
+     *
      * In order to calculate the cost of "count" levels, we need to run a loop. "count" can be a big number, so it's
      * infeasible to calculate the cost in that way. We need to find the closed forms of:
      *
--- a/src/Bladeburner/Skill.ts
+++ b/src/Bladeburner/Skill.ts
@@ -3,7 +3,7 @@ import type { BladeMultName, BladeSkillName } from "@enums";
 import { currentNodeMults } from "../BitNode/BitNodeMultipliers";
 import { Bladeburner } from "./Bladeburner";
 import { Availability } from "./Types";
-import { PositiveInteger, PositiveSafeInteger, isPositiveInteger } from "../types";
+import { PositiveInteger, isPositiveInteger } from "../types";
 import { PartialRecord, getRecordEntries } from "../Types/Record";
 
 interface SkillParams {
@@ -35,30 +35,10 @@ export class Skill {
   }
 
   calculateCost(currentLevel: number, count = 1 as PositiveInteger): number {
-    if (currentLevel + count > this.maxLvl) return Infinity;
-
-    const recursiveMode = (currentLevel: number, count: PositiveSafeInteger): number => {
-      if (count <= 1) {
-        return Math.floor((this.baseCost + currentLevel * this.costInc) * currentNodeMults.BladeburnerSkillCost);
-      } else {
-        const thisUpgrade = Math.floor(
-          (this.baseCost + currentLevel * this.costInc) * currentNodeMults.BladeburnerSkillCost,
-        );
-        return this.calculateCost(currentLevel + 1, (count - 1) as PositiveSafeInteger) + thisUpgrade;
-      }
-    };
-
-    // Use recursive mode if count is small
-    if (count <= 100) return recursiveMode(currentLevel, count as PositiveSafeInteger);
-    // Use optimized mode if count is large
-    else {
-      //unFloored is roughly equivalent to
-      //(this.baseCost + currentLevel * this.costInc) * BitNodeMultipliers.BladeburnerSkillCost
-      //being repeated for increasing currentLevel
-      const preMult = (count * (2 * this.baseCost + this.costInc * (2 * currentLevel + count + 1))) / 2;
-      const unFloored = preMult * currentNodeMults.BladeburnerSkillCost - count / 2;
-      return Math.floor(unFloored);
-    }
+    return Math.floor(
+      count * this.baseCost * currentNodeMults.BladeburnerSkillCost +
+        this.costInc * currentNodeMults.BladeburnerSkillCost * (count * currentLevel + (count * (count - 1)) / 2),
+    );
   }
 
   canUpgrade(bladeburner: Bladeburner, count = 1): Availability<{ cost: number }> {
--- a/src/Bladeburner/Skill.ts
+++ b/src/Bladeburner/Skill.ts
@@ -35,9 +35,45 @@ export class Skill {
   }
 
   calculateCost(currentLevel: number, count = 1 as PositiveInteger): number {
-    return Math.floor(
-      count * this.baseCost * currentNodeMults.BladeburnerSkillCost +
-        this.costInc * currentNodeMults.BladeburnerSkillCost * (count * currentLevel + (count * (count - 1)) / 2),
+    /**
+     * The cost of the next level: (baseCost + currentLevel * costInc) * mult. The cost needs to be an integer, so we
+     * need to use Math.floor or Math.round.
+     *
+     * In order to calculate the cost of "count" levels, we need to run a loop. "count" can be a big number, so it's
+     * infeasible to calculate the cost in that way. We need to find the closed forms of:
+     *
+     * [1]:
+     * $$Cost = \sum_{i = CurrentLevel}^{CurrentLevel+Count-1}\lfloor ((BaseCost + i \ast CostInc) \ast Mult) \rfloor$$
+     *
+     * Or:
+     *
+     * [2]:
+     * $$Cost = \sum_{i = CurrentLevel}^{CurrentLevel+Count-1}\lceil ((BaseCost + i \ast CostInc) \ast Mult) \rceil$$
+     *
+     * It's really hard to find the closed forms of those two equations, so we switch to these equations:
+     *
+     * [3]:
+     * $$Cost = \lfloor\sum_{i = CurrentLevel}^{CurrentLevel+Count-1} ((BaseCost + i \ast CostInc) \ast Mult) \rfloor$$
+     *
+     * Or
+     *
+     * [4]:
+     * $$Cost = \lceil\sum_{i = CurrentLevel}^{CurrentLevel+Count-1} ((BaseCost + i \ast CostInc) \ast Mult) \rceil$$
+     *
+     * This means that we do the flooring/rounding at the end instead of each iterative step.
+     *
+     * [3] and [4] are not equivalent to [1] and [2] respectively, but it's much easier to find the close forms of [3]
+     * and [4] than [1] and [2]. After testing, we conclude that the cost calculated by [4] is a good approximation of
+     * [2], so we choose [4] to calculate the cost. In order to calculate the cost with a big "count", we accept the
+     * slight inaccuracy.
+     *
+     * The closed form of [4]:
+     *
+     * $$Cost = \lceil Count \ast Mult \ast (BaseCost + (CostInc \ast (CurrentLevel + \frac{Count - 1}{2}))) \rceil$$
+     *
+     */
+    return Math.round(
+      count * currentNodeMults.BladeburnerSkillCost * (this.baseCost + this.costInc * (currentLevel + (count - 1) / 2)),
     );
   }
 


d0sboots:
A couple of comments:
* `Count` and `Mult` can be factored out of the entire expression, which makes the whole thing simpler.
* I think we probably do want the `count/2` floor adjustment, so that doing many at once doesn't drift too far from doing one-at-a-time. Alternatively, we could switch to Math.round instead of Math.floor, which isn't biased when accumulating many terms. (At the cost of making some costs 1 bigger for the one-at-a-time case.)
* The whole thing needs a comment, both discussing what's going on (how the formula is doing an overall sum), *why* it's this way (too slow to do it iteratively), and that we're OK with there being slight inaccuracies for doing lots of terms at once.

catloversg:
>- `Count` and `Mult` can be factored out of the entire expression, which makes the whole thing simpler.
>- The whole thing needs a comment, both discussing what's going on (how the formula is doing an overall sum), _why_ it's this way (too slow to do it iteratively), and that we're OK with there being slight inaccuracies for doing lots of terms at once.

I'll do that after we finalize the solution.

>- I think we probably do want the `count/2` floor adjustment, so that doing many at once doesn't drift too far from doing one-at-a-time. Alternatively, we could switch to Math.round instead of Math.floor, which isn't biased when accumulating many terms. (At the cost of making some costs 1 bigger for the one-at-a-time case.)

I wrote some test code. What do you think about the results?

Code:
```js
(() => {
  let baseCost = 1;
  let costInc = 2.5;
  let mult = 1;

  function currentNonRecursiveCalculation(currentLevel, count) {
    const preMult = (count * (2 * baseCost + costInc * (2 * currentLevel + count + 1))) / 2;
    const unFloored = preMult * mult - count / 2;
    return Math.floor(unFloored);
  }

  function iterationFloor(currentLevel, count) {
    let sum = 0;
    for (let i = currentLevel; i < currentLevel + count; ++i) {
      sum += Math.floor((baseCost + i * costInc) * mult);
    }
    return sum;
  }

  function iterationRound(currentLevel, count) {
    let sum = 0;
    for (let i = currentLevel; i < currentLevel + count; ++i) {
      sum += Math.round((baseCost + i * costInc) * mult);
    }
    return sum;
  }

  function proposedFormula(currentLevel, count) {
    let sum = count * baseCost * mult + costInc * mult * (count * currentLevel + (count * (count - 1)) / 2);
    return Math.floor(sum);
  }

  function proposedFormulaWithAdjustment(currentLevel, count) {
    let sum = count * baseCost * mult + costInc * mult * (count * currentLevel + (count * (count - 1)) / 2);
    return Math.floor(sum - count / 2);
  }

  const currentLevel = 0;
  const count = 123456;

  const baseCostList = [1, 2, 3];
  const costIncList = [1, 1.1, 1.4, 2.1, 2.5, 3];
  const multList = [1, 1.2, 2];

  for (let i = 0; i < baseCostList.length; ++i) {
    for (let j = 0; j < costIncList.length; ++j) {
      for (let k = 0; k < multList.length; ++k) {
        baseCost = baseCostList[i];
        costInc = costIncList[j];
        mult = multList[k];
        console.log(`baseCost: ${baseCost}, costInc: ${costInc}, mult: ${mult}`);
        console.log("iterationFloor".padEnd(30, " "), iterationFloor(currentLevel, count));
        console.log("iterationRound".padEnd(30, " "), iterationRound(currentLevel, count));
        console.log("proposedFormula".padEnd(30, " "), proposedFormula(currentLevel, count));
        console.log(
          "proposedFormulaWithAdjustment".padEnd(30, " "),
          proposedFormulaWithAdjustment(currentLevel, count),
        );
        console.log(
          "currentNonRecursiveCalculation".padEnd(30, " "),
          currentNonRecursiveCalculation(currentLevel, count),
        );
        console.log();
      }
    }
  }
})();
```

Results:
```
baseCost: 1, costInc: 1, mult: 1
iterationFloor                 7620753696
iterationRound                 7620753696
proposedFormula                7620753696
proposedFormulaWithAdjustment  7620691968
currentNonRecursiveCalculation 7620815424

baseCost: 1, costInc: 1, mult: 1.2
iterationFloor                 9144855053
iterationRound                 9144904435
proposedFormula                9144904435
proposedFormulaWithAdjustment  9144842707
currentNonRecursiveCalculation 9144990854

baseCost: 1, costInc: 1, mult: 2
iterationFloor                 15241507392
iterationRound                 15241507392
proposedFormula                15241507392
proposedFormulaWithAdjustment  15241445664
currentNonRecursiveCalculation 15241692576

baseCost: 1, costInc: 1.1, mult: 1
iterationFloor                 8382761166
iterationRound                 8382822892
proposedFormula                8382816720
proposedFormulaWithAdjustment  8382754992
currentNonRecursiveCalculation 8382890793

baseCost: 1, costInc: 1.1, mult: 1.2
iterationFloor                 10059320805
iterationRound                 10059380064
proposedFormula                10059380064
proposedFormulaWithAdjustment  10059318336
currentNonRecursiveCalculation 10059481297

baseCost: 1, costInc: 1.1, mult: 2
iterationFloor                 16765584058
iterationRound                 16765633440
proposedFormula                16765633440
proposedFormulaWithAdjustment  16765571712
currentNonRecursiveCalculation 16765843315

baseCost: 1, costInc: 1.4, mult: 1
iterationFloor                 10668951734
iterationRound                 10669005792
proposedFormula                10669005792
proposedFormulaWithAdjustment  10668944064
currentNonRecursiveCalculation 10669116902

baseCost: 1, costInc: 1.4, mult: 1.2
iterationFloor                 12802746757
iterationRound                 12802806951
proposedFormula                12802806950
proposedFormulaWithAdjustment  12802745222
currentNonRecursiveCalculation 12802952628

baseCost: 1, costInc: 1.4, mult: 2
iterationFloor                 21337957526
iterationRound                 21338011584
proposedFormula                21338011584
proposedFormulaWithAdjustment  21337949856
currentNonRecursiveCalculation 21338295532

baseCost: 1, costInc: 2.1, mult: 1
iterationFloor                 16003391406
iterationRound                 16003453132
proposedFormula                16003446960
proposedFormulaWithAdjustment  16003385232
currentNonRecursiveCalculation 16003644489

baseCost: 1, costInc: 2.1, mult: 1.2
iterationFloor                 19204077093
iterationRound                 19204136352
proposedFormula                19204136352
proposedFormulaWithAdjustment  19204074624
currentNonRecursiveCalculation 19204385733

baseCost: 1, costInc: 2.1, mult: 2
iterationFloor                 32006844538
iterationRound                 32006893920
proposedFormula                32006893920
proposedFormulaWithAdjustment  32006832192
currentNonRecursiveCalculation 32007350707

baseCost: 1, costInc: 2.5, mult: 1
iterationFloor                 19051668192
iterationRound                 19051729920
proposedFormula                19051699056
proposedFormulaWithAdjustment  19051637328
currentNonRecursiveCalculation 19051945968

baseCost: 1, costInc: 2.5, mult: 1.2
iterationFloor                 22862014176
iterationRound                 22862014176
proposedFormula                22862038867
proposedFormulaWithAdjustment  22861977139
currentNonRecursiveCalculation 22862347507

baseCost: 1, costInc: 2.5, mult: 2
iterationFloor                 38103398112
iterationRound                 38103398112
proposedFormula                38103398112
proposedFormulaWithAdjustment  38103336384
currentNonRecursiveCalculation 38103953664

baseCost: 1, costInc: 3, mult: 1
iterationFloor                 22862014176
iterationRound                 22862014176
proposedFormula                22862014176
proposedFormulaWithAdjustment  22861952448
currentNonRecursiveCalculation 22862322816

baseCost: 1, costInc: 3, mult: 1.2
iterationFloor                 27434367629
iterationRound                 27434417011
proposedFormula                27434417011
proposedFormulaWithAdjustment  27434355283
currentNonRecursiveCalculation 27434799724

baseCost: 1, costInc: 3, mult: 2
iterationFloor                 45724028352
iterationRound                 45724028352
proposedFormula                45724028352
proposedFormulaWithAdjustment  45723966624
currentNonRecursiveCalculation 45724707360

baseCost: 2, costInc: 1, mult: 1
iterationFloor                 7620877152
iterationRound                 7620877152
proposedFormula                7620877152
proposedFormulaWithAdjustment  7620815424
currentNonRecursiveCalculation 7620938880

baseCost: 2, costInc: 1, mult: 1.2
iterationFloor                 9145003200
iterationRound                 9145052582
proposedFormula                9145052582
proposedFormulaWithAdjustment  9144990854
currentNonRecursiveCalculation 9145139001

baseCost: 2, costInc: 1, mult: 2
iterationFloor                 15241754304
iterationRound                 15241754304
proposedFormula                15241754304
proposedFormulaWithAdjustment  15241692576
currentNonRecursiveCalculation 15241939488

baseCost: 2, costInc: 1.1, mult: 1
iterationFloor                 8382884622
iterationRound                 8382946348
proposedFormula                8382940176
proposedFormulaWithAdjustment  8382878448
currentNonRecursiveCalculation 8383014249

baseCost: 2, costInc: 1.1, mult: 1.2
iterationFloor                 10059468953
iterationRound                 10059528211
proposedFormula                10059528211
proposedFormulaWithAdjustment  10059466483
currentNonRecursiveCalculation 10059629445

baseCost: 2, costInc: 1.1, mult: 2
iterationFloor                 16765830970
iterationRound                 16765880352
proposedFormula                16765880352
proposedFormulaWithAdjustment  16765818624
currentNonRecursiveCalculation 16766090227

baseCost: 2, costInc: 1.4, mult: 1
iterationFloor                 10669075194
iterationRound                 10669129248
proposedFormula                10669129248
proposedFormulaWithAdjustment  10669067520
currentNonRecursiveCalculation 10669240358

baseCost: 2, costInc: 1.4, mult: 1.2
iterationFloor                 12802894904
iterationRound                 12802955097
proposedFormula                12802955097
proposedFormulaWithAdjustment  12802893369
currentNonRecursiveCalculation 12803100775

baseCost: 2, costInc: 1.4, mult: 2
iterationFloor                 21338204442
iterationRound                 21338258496
proposedFormula                21338258496
proposedFormulaWithAdjustment  21338196768
currentNonRecursiveCalculation 21338542444

baseCost: 2, costInc: 2.1, mult: 1
iterationFloor                 16003514862
iterationRound                 16003576588
proposedFormula                16003570416
proposedFormulaWithAdjustment  16003508688
currentNonRecursiveCalculation 16003767945

baseCost: 2, costInc: 2.1, mult: 1.2
iterationFloor                 19204225240
iterationRound                 19204284498
proposedFormula                19204284499
proposedFormulaWithAdjustment  19204222771
currentNonRecursiveCalculation 19204533880

baseCost: 2, costInc: 2.1, mult: 2
iterationFloor                 32007091450
iterationRound                 32007140832
proposedFormula                32007140832
proposedFormulaWithAdjustment  32007079104
currentNonRecursiveCalculation 32007597619

baseCost: 2, costInc: 2.5, mult: 1
iterationFloor                 19051791648
iterationRound                 19051853376
proposedFormula                19051822512
proposedFormulaWithAdjustment  19051760784
currentNonRecursiveCalculation 19052069424

baseCost: 2, costInc: 2.5, mult: 1.2
iterationFloor                 22862137632
iterationRound                 22862137632
proposedFormula                22862187014
proposedFormulaWithAdjustment  22862125286
currentNonRecursiveCalculation 22862495654

baseCost: 2, costInc: 2.5, mult: 2
iterationFloor                 38103645024
iterationRound                 38103645024
proposedFormula                38103645024
proposedFormulaWithAdjustment  38103583296
currentNonRecursiveCalculation 38104200576

baseCost: 2, costInc: 3, mult: 1
iterationFloor                 22862137632
iterationRound                 22862137632
proposedFormula                22862137632
proposedFormulaWithAdjustment  22862075904
currentNonRecursiveCalculation 22862446272

baseCost: 2, costInc: 3, mult: 1.2
iterationFloor                 27434515776
iterationRound                 27434565158
proposedFormula                27434565158
proposedFormulaWithAdjustment  27434503430
currentNonRecursiveCalculation 27434947872

baseCost: 2, costInc: 3, mult: 2
iterationFloor                 45724275264
iterationRound                 45724275264
proposedFormula                45724275264
proposedFormulaWithAdjustment  45724213536
currentNonRecursiveCalculation 45724954272

baseCost: 3, costInc: 1, mult: 1
iterationFloor                 7621000608
iterationRound                 7621000608
proposedFormula                7621000608
proposedFormulaWithAdjustment  7620938880
currentNonRecursiveCalculation 7621062336

baseCost: 3, costInc: 1, mult: 1.2
iterationFloor                 9145151347
iterationRound                 9145200730
proposedFormula                9145200729
proposedFormulaWithAdjustment  9145139001
currentNonRecursiveCalculation 9145287148

baseCost: 3, costInc: 1, mult: 2
iterationFloor                 15242001216
iterationRound                 15242001216
proposedFormula                15242001216
proposedFormulaWithAdjustment  15241939488
currentNonRecursiveCalculation 15242186400

baseCost: 3, costInc: 1.1, mult: 1
iterationFloor                 8383008078
iterationRound                 8383069804
proposedFormula                8383063632
proposedFormulaWithAdjustment  8383001904
currentNonRecursiveCalculation 8383137705

baseCost: 3, costInc: 1.1, mult: 1.2
iterationFloor                 10059617099
iterationRound                 10059676359
proposedFormula                10059676358
proposedFormulaWithAdjustment  10059614630
currentNonRecursiveCalculation 10059777592

baseCost: 3, costInc: 1.1, mult: 2
iterationFloor                 16766077882
iterationRound                 16766127264
proposedFormula                16766127264
proposedFormulaWithAdjustment  16766065536
currentNonRecursiveCalculation 16766337139

baseCost: 3, costInc: 1.4, mult: 1
iterationFloor                 10669198654
iterationRound                 10669252704
proposedFormula                10669252704
proposedFormulaWithAdjustment  10669190976
currentNonRecursiveCalculation 10669363814

baseCost: 3, costInc: 1.4, mult: 1.2
iterationFloor                 12803043054
iterationRound                 12803103245
proposedFormula                12803103244
proposedFormulaWithAdjustment  12803041516
currentNonRecursiveCalculation 12803248922

baseCost: 3, costInc: 1.4, mult: 2
iterationFloor                 21338451358
iterationRound                 21338505408
proposedFormula                21338505408
proposedFormulaWithAdjustment  21338443680
currentNonRecursiveCalculation 21338789356

baseCost: 3, costInc: 2.1, mult: 1
iterationFloor                 16003638318
iterationRound                 16003700044
proposedFormula                16003693872
proposedFormulaWithAdjustment  16003632144
currentNonRecursiveCalculation 16003891401

baseCost: 3, costInc: 2.1, mult: 1.2
iterationFloor                 19204373388
iterationRound                 19204432647
proposedFormula                19204432646
proposedFormulaWithAdjustment  19204370918
currentNonRecursiveCalculation 19204682027

baseCost: 3, costInc: 2.1, mult: 2
iterationFloor                 32007338362
iterationRound                 32007387744
proposedFormula                32007387744
proposedFormulaWithAdjustment  32007326016
currentNonRecursiveCalculation 32007844531

baseCost: 3, costInc: 2.5, mult: 1
iterationFloor                 19051915104
iterationRound                 19051976832
proposedFormula                19051945968
proposedFormulaWithAdjustment  19051884240
currentNonRecursiveCalculation 19052192880

baseCost: 3, costInc: 2.5, mult: 1.2
iterationFloor                 22862261088
iterationRound                 22862384544
proposedFormula                22862335161
proposedFormulaWithAdjustment  22862273433
currentNonRecursiveCalculation 22862643801

baseCost: 3, costInc: 2.5, mult: 2
iterationFloor                 38103891936
iterationRound                 38103891936
proposedFormula                38103891936
proposedFormulaWithAdjustment  38103830208
currentNonRecursiveCalculation 38104447488

baseCost: 3, costInc: 3, mult: 1
iterationFloor                 22862261088
iterationRound                 22862261088
proposedFormula                22862261088
proposedFormulaWithAdjustment  22862199360
currentNonRecursiveCalculation 22862569728

baseCost: 3, costInc: 3, mult: 1.2
iterationFloor                 27434663923
iterationRound                 27434713306
proposedFormula                27434713305
proposedFormulaWithAdjustment  27434651577
currentNonRecursiveCalculation 27435096019

baseCost: 3, costInc: 3, mult: 2
iterationFloor                 45724522176
iterationRound                 45724522176
proposedFormula                45724522176
proposedFormulaWithAdjustment  45724460448
currentNonRecursiveCalculation 45725201184
```

d0sboots:
You didn't test `proposedFormulaRound`, but it would be almost identical to `proposedFormula`, differing by 1 at most.

Across the results, I see `iterationRound` being identical to, or else very close to `proposedFormula`, which is what I expected. The adjusted formula does not have the same level of accuracy as compared to the floor version. 

So I suggest we keep your proposed formula, but swap out floor for round. This will slightly change the true costs, but not as much as this fix will affect them overall anyway. And I'm the future, if you upgrade one-at-a-time (or in many small bursts), the cost will be well aligned with doing it all at once.

src/Bladeburner/Skill.ts
```diff
@@ -35,9 +35,45 @@ export class Skill {
...
+     * In order to calculate the cost of "count" levels, we need to run a loop. "count" can be a big number, so it's
+     * infeasible to calculate the cost in that way. We need to find the closed forms of:
+     *
+     * [1]:
+     * $$Cost = \sum_{i = CurrentLevel}^{CurrentLevel+Count-1}\lfloor ((BaseCost + i \ast CostInc) \ast Mult) \rfloor$$
+     *
+     * Or:
+     *
+     * [2]:
+     * $$Cost = \sum_{i = CurrentLevel}^{CurrentLevel+Count-1}\lceil ((BaseCost + i \ast CostInc) \ast Mult) \rceil$$
```
d0sboots, line 54: Nice comprehensive comment! Only one nitpick: You're using `\lceil` here, which is for ceil(). AFAIK, there isn't a commonly accepted math symbol for round(), we just write the function out.

catloversg: I added a note for it. I can change `\lceil` and `\rceil` to the literal `Math.round` in the Tex expressions if you want.

d0sboots: Good enough.

PR #1445, opened 2024-06-28
MISC: Simplify graftingIntBonus calculation.

tomprince:
The weight of the intelligence bonus is a multiplier to the percentage increase. So, rather than calculate it with a weight of 3 and then divide by 3, we can just calculate it with a weight of 1.


d0sboots:
Huh, that was pretty silly.

PR #1446, opened 2024-06-29
EDITOR: Use ramOverride() to set compiled script RAM

d0sboots:
To use this, add a line like "ns.ramOverride(2);" as the first statement in main(). Not only will it take effect at runtime, but it will now *also* be parsed at compile time, changing the script's static RAM limit. Since ns.ramOverride is a 0-cost function, the call to it on startup then becomes a no-op.

This is an often-requested feature, and allows for scripts to set their usage without it needing to be explicitly mentioned via args or options when being launched. This also reduces pressure on the static RAM analysis to be perfect all the time. (But certain limits, such as "functions names must be unique across namespaces," remain.)

This also adds a tooltip to the RAM calculation, to make this slightly discoverable.

image

Tested by playing with the tag values a lot, trying nonsense values, etc. Also added some automated test cases. You can set impossibly large values, which I view as OK. (The same is true of the `ramOverride` param.)

### DIFF
--- a/src/Script/RamCalculations.ts
+++ b/src/Script/RamCalculations.ts
@@ -15,6 +15,7 @@ import { Script } from "./Script";
 import { Node } from "../NetscriptJSEvaluator";
 import { ScriptFilePath, resolveScriptFilePath } from "../Paths/ScriptFilePath";
 import { ServerName } from "../Types/strings";
+import { roundToTwo } from "../utils/helpers/roundToTwo";
 
 export interface RamUsageEntry {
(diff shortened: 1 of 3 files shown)


d0sboots:
I have completely overhauled the implementation of this PR. Instead of using magic comments, it is now based on calling `ns.ramOverride()` as the first statement of your script. This is both an easy rule to remember, and guarantees that the runtime effect is equal to the effect that the static analysis is doing. (If it picked it up on later statements, you might have non-trivial computation in the intervening statements that depends on the difference between the normal static RAM limit and the new one being set.)

PR #1448, opened 2024-07-01
MISC: Remove debug code

catloversg:
These codes were only useful for debugging when the PRs were being developed. They should not be in our codebase after that. I do not remove `logBoard` in `src\Go\boardAnalysis\scoring.ts` because it may still be useful.


d0sboots:
@ficocelliguy for heads-up

PR #1450, opened 2024-07-01
BUGFIX: Wrong success chance in ns.bladeburner.getActionEstimatedSuccessChance

catloversg:
If the action type is `BladeActionType.general`, the success chance must be [1, 1].

This PR also fixes the duplicate error message in `checkSleeveNumber`.


src/NetscriptFunctions/Bladeburner.ts
```diff
@@ -124,13 +124,18 @@ export function NetscriptBladeburner(): InternalAPI<INetscriptBladeburner> {
...
+            return [1, 1];
+          case BladeActionType.contract: {
+            const sleevePerson = Player.sleeves[sleeveNumber];
+            return action.getSuccessRange(bladeburner, sleevePerson);
+          }
+          default:
+            return [0, 0];
+        }
       }
+      return action.getSuccessRange(bladeburner, Player);
```
d0sboots: I didn't think to mention this before, but let's invert the outer condition so the short return comes first and we can de-nest.

PR #1451, opened 2024-07-02
CORPORATION: Fix wrong product price calculation

catloversg:
This PR makes 2 changes:
- Remove the check `&& mat.desiredSellPrice` in the material price calculation. I don't get why it's needed. The product price calculation does not have this check.
- Allow the price of products to go lower than MP when the player sets it with a string. Without this change, `MP/2` is the same as `MP`.


d0sboots:
Maybe someday we'll actually unify most of the product/material calculation XD

PR #1452, opened 2024-07-03
UI: Supprt another language

LPerNATTO:
UI: Supprt another language

I want non-English speakers to enjoy this game.
So I think add the function to support another language.
I design the function like this pull request now.
If adding the function is no problem, I continue the develop.
Please consider it.


LPerNATTO:
Sorry to forget paste screenshots.

This is main page in Japanese.
image

I plan to add language setting in Interface Setting.
image

d0sboots:
Unfortunately (as your screenshot shows), the start you've made is only barely scratching the surface. 

There are three major issues here: 
1. AFAIK, none of the major/regular contributors know any languages beyond English. This means that any new features that are added won't end up translated. 
2. The game was fundamentally never written with translation in mind, which means there is a *lot* of stuff that would be hard-to-impossible to resolve. We have lots of places where we dynamically create strings by concatenation, using ternaries etc.; that would all have to be carefully broken down to be able to become translation keys. Conjugating for quantity ("1 module"/"4 modules") is a much bigger deal in some languages than it is in English; you're fortunate that that's not an issue in Japanese XD
3. The game involves programming, and the NS API and names of servers is and always will be in English. (`hack()`, `n00dles`, `tprint()`, etc.). So some English knowledge will always be needed, at a minimum.

I've done exactly this work professionally, so I know how much is involved. It's not impossible, but it would be a *huge* project, involving significant changes across pretty much every single file in the codebase.

If you want to tackle a project that is still huge, but IMO doable, I suggest translating the game's documentation. That doesn't have any of the same technical challenges, and would be helpful for people who know enough English to use the UI but struggle with longer/more complicated explanations. We link to the docs in-game, and if they were fully translated I think we could figure out a way to make that link locale-aware.

LPerNATTO:
Thanks for your advice. And I'm sorry for my poor English.
I'm glad to know that you welcome the idea of localization in another pull request.
I, too, feel that this project will be so big.
So, I should first translate documents or change programs of documents as you said.

Followings are my thoughts towards three major issues.

> 1. AFAIK, none of the major/regular contributors know any languages beyond English. This means that any new features that are added won't end up translated.

I think latest contents often aren't translated or be postponed.
But it may be good to add features on the premise translation.

> 2. The game was fundamentally never written with translation in mind, which means there is a lot of stuff that would be hard-to-impossible to resolve. We have lots of places where we dynamically create strings by concatenation, using ternaries etc.; that would all have to be carefully broken down to be able to become translation keys. Conjugating for quantity ("1 module"/"4 modules") is a much bigger deal in some languages than it is in English; you're fortunate that that's not an issue in Japanese XD

Conjugating for quantity is always the difficult problem.
But postponing difficult parts will be good because I think that players care about interfaces, documents and references.
(It is lucky that there isn't the issue in written Japanese!)

> 3. The game involves programming, and the NS API and names of servers is and always will be in English. (hack(), n00dles, tprint(), etc.). So some English knowledge will always be needed, at a minimum.

It may be no problem that NS APIs, names of servers, names of functions are English.
Because those are English in another language generally.
But I want to translate references. (But it will be postponed because of amount...)

If it's no problem, I'll be glad that there is a opened pull request or issue about localization.

d0sboots:
> If it's no problem, I'll be glad that there is a opened pull request or issue about localization.

Sure. There have been issues opened in the past, I'm not sure if they're still open or not. If you want to open a feature request for translating the documentation, that would be something we can leave open long-term. 

That issue would be a good place to discuss design for how to (re)structure the file tree to encompass multiple languages for our .md content.

Release:
```
export const Locale = {
  Menu: function (key: string) {
    return key in translations[Settings.InterfaceLocale].Menu ? translations[Settings.InterfaceLocale].Menu[key] : key;
  },
  Overview: function (key: string) {
    return key in translations[Settings.InterfaceLocale].Overview
      ? translations[Settings.InterfaceLocale].Overview[key]
      : key;
  },
};
```
I would rewrite this so that I don't have to add a function every time I add a section. Now there are two Menu and Overview, then, for example, Tutorial, Messages and Industry will be added. And the functions are no different except for accessing a specific section.

Release:
> That issue would be a good place to discuss design for how to (re)structure the file tree to encompass multiple languages for our .md content.

To be honest, I don’t understand why it is necessary to (re)structure the file tree. Data files for various languages ​​are already located in a separate `Locales` directory. As well as a script that exports the localization function.

PR #1457, opened 2024-07-04
API: Add a way to get the list of all factions

catloversg:
"Getting the list of all factions" is a frequently requested feature. We have discussed it many times but have not come to a conclusion on how to implement it. There are 3 ways to add this feature:
- `ns.enums.FactionName`: This is the easiest way to implement it. It's also the most discoverable place for new players. The only concern is that some faction names may be spoilers. IMO, faction names are not spoilers. With a new player, there is no difference between the names of `Daedalus` and `NiteSec`. Those are both faction names. There is nothing special about the name of `Daedalus`. The special thing is only revealed when they join it, buy TRP, read the message, look around for special servers, etc. It's the same for other special factions. For example, knowing the name of `Church of the Machine God` does not reveal anything. It's just the name of a faction.
- `ns.singularity.enums.FactionName`: Having access to the list of faction names is most useful for using the singularity API. We also don't have to worry about spoilers. However, due to how the NS proxy works, it's really hard to implement it this way.
- `ns.singularity.getAllFactionNames`: Make it an API function instead of an enum. This is the last resort.

I decided to choose `ns.enums.FactionName`.


d0sboots:
I agree with your reasoning that "just the names of factions are not spoilers." There is somewhat of a counterargument, in that knowing of the existence of factions that you don't have/haven't seen is a slight spoiler. But we already have this in a much more major way with the major mechanics and their namespace; these are at least hidden in an enum that you can avoid seeing. 

This looks good to me, but I'll wait for @Snarling since this has already been contentious.

Alpheus:
Bump @Snarling this is a great change!

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -8256,6 +8256,45 @@ declare enum CompanyName {
   NoodleBar = "Noodle Bar",
 }
 
+/** Names of all factions
```
d0sboots: I am strongly considering just merging this at this time. One thought: Maybe we could square the circle by adding something to this to the effect of "don't click if you don't want to be spoiled" or some such? I'm struggling to come up with something short that still conveys the right idea; basically, that if you don't *want* to know what all the factions are, you shouldn't click through (in the web docs, or in-game, or wherever you are.)

catloversg: I added a warning about spoilers.

PR #1458, opened 2024-07-04
EDITOR: Provide type definitions for `React` and `ReactDOM` in in-game editor.

tomprince:
This also:
-  fixes the type declaration of `!!raw-loader!` modules. This only affects the types, not the actual code that is generated.
- switches from using `raw-loader` to using webpack asset modules.


src/Documentation/pages.ts
```diff
@@ -61,9 +61,8 @@ import file58 from "!!raw-loader!./doc/programming/hackingalgorithms.md";
 import file59 from "!!raw-loader!./doc/programming/learn.md";
 import file60 from "!!raw-loader!./doc/programming/remote_api.md";
 
-interface Document {
-  default: string;
-}
+import type { Document } from "./root.js";
```
d0sboots: Why is this root.js and not root.ts?

If this is correct, it definitely needs a comment.

src/ScriptEditor/ScriptEditor.ts
```diff
@@ -3,7 +3,11 @@ import type { ContentFilePath } from "../Paths/ContentFile";
 import { EventEmitter } from "../utils/EventEmitter";
 import * as monaco from "monaco-editor";
 import { loadThemes, makeTheme, sanitizeTheme } from "./ui/themes";
-import libSource from "!!raw-loader!./NetscriptDefinitions.d.ts";
+import netscriptDefinitions from "!!raw-loader!./NetscriptDefinitions.d.ts";
+// We use a relative path here to bypass the exports in @types/react's package.json
```
d0sboots: Good instinct on the comment, but I'm afraid it has only left me with more questions. XD

src/ScriptEditor/ScriptEditor.ts
```diff
@@ -3,7 +3,11 @@ import type { ContentFilePath } from "../Paths/ContentFile";
...
+// The trailing `?` prevents typescript from complaining about importing a delcaration file.
```
d0sboots: Why is that only needed for this declaration file, and not the others?

src/Documentation/pages.ts
tomprince: No particular reason, other than unfamiliarity with javascript/typescript.

tomprince: It works because typescript will look for `.ts` when asked for `.ts`. I think this partly because typescript tries to not change the dynamic behaviour of programs, so when not bundling, the import would eventually refer to a `.js` file.

src/ScriptEditor/ScriptEditor.ts
d0sboots: Not resolved; I still have all the questions about this XD

src/ScriptEditor/ScriptEditor.ts
tomprince: I'm not entirely sure, I figured out the right incantation mostly by trial-and-error.

After reading your question, I did some more experimentation to see if I could figure out the reason. I wasn't successful, but I did try switching from `raw-loader` (which is deprecated) to [asset modules](https://webpack.js.org/guides/asset-modules/). Making that switch did at least have the benefit of making the behaviour of `react` and `react-dom` be the same (i.e. both requiring a relative path).

I've included the switch to asset modules in this PR as well now.

d0sboots:
Blar. The changes look positive (good to get off deprecated stuff), but changing a core part of asset loading like this means I'm gonna need to manually test some stuff before I'm satisfied.

tomprince:
> Blar. The changes look positive (good to get off deprecated stuff), but changing a core part of asset loading like this means I'm gonna need to manually test some stuff before I'm satisfied.

Note that the only assets impacted by this change are the type definitions (used by the editor and remote file API), and the documentation. The other assets (e.g. images, fonts) are already using assets (although a slightly different variant that returns a URL as the asset module contents).

tomprince:
One thing that one could do is look at the diff between the output of `npm run build:dev` (or `npm run build`). I did this for each of the first two commits, and the only difference is basically some variable names, and whether the asset modules export a dict with a `default` key or not.

d0sboots:
Everything seems to be working fine.

PR #1460, opened 2024-07-05
DOCUMENTATION: Add link to NS API documentation

### LINKED ISSUE #1047: Cannot use input fields after closing alert box created by the game.
After using alert() function in my script to notify me about somethin I can't use input fields after closing alert box. 

Steps to reproduce:
1. Create a new script.
2. Use alert() function
3. Use ns.exit() function (I don't know if this is relevant but I have this in the line after alert)

Version: 2.5.2 (c6141f2ad) Steam
Platform: Windows 11

catloversg:
The link to the NS API documentation is only available in the text editor. This PR adds it to the main page of the Documentation tab.

Closes #1047.

### DIFF
--- a/src/utils/StringUtils.ts
+++ b/src/utils/StringUtils.ts
@@ -1,11 +1,17 @@
+/**
+ * TODO v3.0.0
+ * Move all string-related utility functions in /src/utils to this file.
+ */
+
 import { Settings } from "../Settings/Settings";
+import { CONSTANTS } from "../Constants";
 
 /*
 Converts a date representing time in milliseconds to a string with the format H hours M minutes and S seconds
 e.g.    10000 -> "10 seconds"
         120000 -> "2 minutes and 0 seconds"
 */
-function convertTimeMsToTimeElapsedString(time: number, showMilli = false): string {
+export function convertTimeMsToTimeElapsedString(time: number, showMilli = false): string {
   const negFlag = time < 0;
   time = Math.abs(Math.floor(time));
   const millisecondsPerSecond = 1000;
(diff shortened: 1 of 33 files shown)


src/utils/StringUtils.ts
```diff
@@ -1,11 +1,17 @@
+/**
+ * TODO v3.0.0
+ * Move all string-related utility functions in /src/utils to this file.
```
d0sboots: I'm not sure this is actually a goal. "*Utils" files are something of an anti-pattern, they tend to collect lots of unrelated functionality without rhyme or reason. In our case, this can be problematic for dependency-related reasons. I'm pretty sure Snarling was trying to shard functions into smaller files, not larger ones.

d0sboots (commented):
The link is welcome, and overdue. But it's overshadowed by the (largely) unrelated refactor. XD

src/utils/StringUtils.ts
catloversg: I partially disagree. X-Utils files are okay when they group "related" functions. The problem is how we decide which things are "related". For example, it's fine to group `createRandomString` (`src\utils\helpers\createRandomString.ts`), `trimQuotes` (`src\utils\helpers\string.ts`), `generateRandomString`, `containsAllStrings`, `longestCommonStart`, etc. IMO, "one function per file" (many cases in `src\utils\helpers`) is only good for a "utility" library (like lodash). The concern of "dependency hell" is valid, though.

I'll still revert the change in the filename and the new note. It's best to discuss and implement it in another PR.

d0sboots: I should clarify my position a bit more, which is: I'm not sure I have a position. XD

Specifically, *Utils files are dangerous because they *tend to* accumulate unrelated functions. It doesn't mean that they're *wrong* though, and sometimes even a dangerous pattern is still the right choice. 

The main thing I was objecting to was the TODO, because I'm not sure that we want to go that direction overall. Maybe we do, but it seemed contrary to what I've seen Snarling do. The renaming you did here seemed fine.

PR #1462, opened 2024-07-06
BUGFIX: Disallow scripts from hacking/growing/weakening player-owned servers

yichizhng:
Also covers other hacking related functions, but not `getServer` related ones (since `getServerMoneyAvailable('home')` is an accepted way to get player money).

Tested manually

Test server at https://yichizhng.github.io/bitburner-src/

### DIFF
--- a/src/Netscript/NetscriptHelpers.tsx
+++ b/src/Netscript/NetscriptHelpers.tsx
@@ -420,6 +420,7 @@ function hack(ctx: NetscriptContext, hostname: string, manual: boolean, opts: un
   if (!(server instanceof Server)) {
     throw errorMessage(ctx, "Cannot be executed on this server.");
   }
+  if (server.purchasedByPlayer) throw errorMessage(ctx, "Cannot hack your own machines!");
 
   // Calculate the hacking time
   // This is in seconds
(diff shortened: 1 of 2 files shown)


d0sboots:
I've thought of making this PR myself in the past, but ultimately did not. I'm really of two minds on it. 

Getting the obvious out of the way: this is an API break. Why? Because currently, hack/grow/weaken on your own servers does not have an effect, but it does not *error*, either. 

My maintainer spider-sense is telling me that the moment this gets merged, someone is going to show up with a use case that got broken by it. XD Possibly some batcher-timing-testing case or something.

yichizhng:
The API break could be addressed by having hack return `Promise.resolve(0)` like the other functions (grow and weaken already don't error), though the documentation already indicates that you shouldn't be able to do this.

Per https://github.com/bitburner-official/bitburner-src/blob/dev/markdown/bitburner.ns.grow.md:
```
Hackable servers are any servers not owned by the player.
```

...though looking deeper, weaken is inconsistent with that, and hack has incorrect information (security is increased even on a failed hack)

And looking at comments the source code, the `Server` class is meant to represent hackable servers; presumably `BaseServer` or another derived class thereof was meant to represent player-owned servers so that the `instanceof` check would check for hackability, but instead those are all `Server`s...

d0sboots:
The Server/BaseServer split is pretty much exclusively about hashnet; hacknet servers have some significantly different properties/methods. Of course, all hashnet servers are also player-owned, but the inverse is not true. 

As you've identified, the whole area with hack/grow/weaken is very warty. Some of these warts are what makes matching possible :D

yichizhng:
Eh, I'll close this for now.

PR #1463, opened 2024-07-06
MISC: Add threshold for warning about system clock

catloversg:
The system clock can be altered by many things, including NTP services. The check in #1354 may be a bit too sensitive in some cases; for example, the NTP service may adject the system clock and make it go backward a few seconds.

This PR adds a threshold. We will only show the warning about the system clock if the time difference exceeds that threshold.


d0sboots:
Can we add the amount of the jump to the dialog? That way, when we get future bug reports we can have a better idea of what's going on.

catloversg:
I changed it like this:

Capture

If you want a different message, just let me know.

d0sboots:
Looks good. Anything that gets us the info in screenshots is sufficient.

PR #1466, opened 2024-07-06
UI: Tweak Hacknet summary

catloversg:
There is a complaint on Discord:
> The hacknet summary box is now with dynamic size when the hashes row width overflow, which makes the right aligned numbers on all other rows move around and hard to read.

before

This PR makes the data column left-aligned:
after
When a value changes, it does not affect the positions of other values.


d0sboots:
Two quick questions:
* Why does this issue only affect the summary? 
* Can I see the change in the context of the whole UI? (Being lazy) Just want to double-check that the differing alignment doesn't clash awfully.

catloversg:
The whole UI:
Capture

The Hacknet summary UI is a `StatsTable`. When the data column moves around (left<->right), it only affects the table itself. Everything else is below that table.

d0sboots:
OK, seeing the screenshot answered both my questions at once: The purchased nodes are *already* being left-justified (probably for the same reason).

PR #1467, opened 2024-07-07
EDITOR: Add support for "Go to Definition" for definitions from other open files.

tomprince:
This uses `registerEditorOpener` API introduced in introduced in moncao-editor [v0.38](https://github.com/microsoft/monaco-editor/blob/main/CHANGELOG.md#0380) to allow going to definitions in other opened files.

This also stops disposing of editor models when navigating. This means the language server doesn't forget about all the code in open but not focused tabs.

---

This subsumes part of #1138. I've looked at supporting non-open files, but doing that changes the lifecycle of models, so probably requires a larger refactor.

### DIFF
--- a/src/ScriptEditor/ui/Editor.tsx
+++ b/src/ScriptEditor/ui/Editor.tsx
@@ -4,6 +4,11 @@ import * as monaco from "monaco-editor";
 
 import { useScriptEditorContext } from "./ScriptEditorContext";
 import { scriptEditor } from "../ScriptEditor";
+import { ContentFilePath } from "../../Paths/ContentFile";
+import { hasScriptExtension } from "../../Paths/ScriptFilePath";
+import { hasTextExtension } from "../../Paths/TextFilePath";
+import { asFilePath } from "../../Paths/FilePath";
+import { OpenScript } from "./OpenScript";
 
 interface EditorProps {
   /** Function to be ran after mounting editor */
@@ -12,12 +17,12 @@ interface EditorProps {
   onChange: (newCode?: string) => void;
   /** This function is called before unmounting the editor */
   onUnmount: () => void;
+  openFile: (hostname: string, filename: ContentFilePath) => OpenScript | undefined;
 }
 
-export function Editor({ onMount, onChange, onUnmount }: EditorProps) {
+export function Editor({ onMount, onChange, onUnmount, openFile }: EditorProps) {
   const containerDiv = useRef<HTMLDivElement | null>(null);
   const editorRef = useRef<monaco.editor.IStandaloneCodeEditor | null>(null);
-  const subscription = useRef<monaco.IDisposable | null>(null);
 
   const { options } = useScriptEditorContext();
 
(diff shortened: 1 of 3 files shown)


d0sboots:
For both safety and (possibly) future sanity, I'm gonna ask you to split this into two PRs: one that *only* upgrades Monaco, and the other with your changes. 

There first should be easy to test in isolation. Also, if we notice something going wrong after release, this will make it much easier to bisect to the problem.

tomprince:
> For both safety and (possibly) future sanity, I'm gonna ask you to split this into two PRs: one that _only_ upgrades Monaco, and the other with your changes.
> 
> There first should be easy to test in isolation. Also, if we notice something going wrong after release, this will make it much easier to bisect to the problem.

Done: #1468. I've rebased this PR on top of that one.

d0sboots:
I was holding off on this until the TS change landed, on the assumption that there would be conflicts, but surprisingly enough there are not. Taking another look now...

src/ScriptEditor/ui/Editor.tsx
```diff
@@ -40,6 +47,30 @@ export function Editor({ onMount, onChange, onUnmount }: EditorProps) {
     subscription.current = editorRef.current.onDidChangeModelContent(() => {
       onChange(editorRef.current?.getValue());
     });
+    editorOpener.current = monaco.editor.registerEditorOpener({
+      async openCodeEditor(source, resource, selectionOrPosition) {
+        console.log(resource);
```
d0sboots: Debug code.

src/ScriptEditor/ui/Editor.tsx
```diff
@@ -40,6 +47,30 @@ export function Editor({ onMount, onChange, onUnmount }: EditorProps) {
...
+        if (resource.scheme != "file" || resource.authority === null) {
```
d0sboots: !==, except for the rare times you want loose equality

src/ScriptEditor/ui/ScriptEditorRoot.tsx
```diff
@@ -154,6 +154,51 @@ function Root(props: IProps): React.ReactElement {
     debouncedCodeParsing(newCode);
   };
 
+  function openFile(hostname: string, filename: ContentFilePath, code?: string) {
+    if (!editorRef.current) {
+      return;
+    }
+    if (!code) {
+      const server = GetServer(hostname);
+      if (!server) throw new Error("Server should not be null but it is.");
```
d0sboots: Please add hostname to the message so this is actionable, if it ever occurs.

src/ScriptEditor/ui/Editor.tsx
```diff
@@ -40,6 +47,30 @@ export function Editor({ onMount, onChange, onUnmount }: EditorProps) {
     subscription.current = editorRef.current.onDidChangeModelContent(() => {
       onChange(editorRef.current?.getValue());
     });
+    editorOpener.current = monaco.editor.registerEditorOpener({
```
d0sboots: Why does this need to be defined now? What was doing its job before?

From the title, it's clear that this is somehow related to the PR goal of "jump to definition," but it is entirely unclear how from this changed code.

tomprince: I'm not sure what this comment is referring to. If you are referring to the react ref, then we need to keep the return value of `registerEditorOpener` to unregister  when the component is unmounted. To make this slightly clearer, I switched both this and the subscription above to share a ref as an array of things to dispose on unmount.

d0sboots: I was asking about the entire function/code block.

tomprince: We are providing a hook for when the editor wants to open another file (as described [here](https://microsoft.github.io/monaco-editor/typedoc/functions/editor.registerEditorOpener.html)).

d0sboots: Thanks, that answers my question. I would add a comment here saying that this is a handler for links clicked when editing, as opposed to handling opening the main model. (This is not obvious from the name or code, unless you're familiar with Momaco's internals.)

tomprince: I wouldn't call it part of the internals of Monaco, but rather part of the public interface. I'd be inclined not to add a comment here, given that the behaviour is documented by the docstring of the method, which is available in someone's editor, if it supports that. (This is in contrast to my other PR, which uses obscure bits of Monaco and the typescript language server).

d0sboots: Well, neither my editor, nor browsing the code on GitHub support that.

Although function docstrings are highly useful, you can't assume their presence when someone is reading your code. Because they could be looking from GitHub, or a copied snippet, or even a printout (unlikely, but I've seen it.)

tomprince:
@d0sboots Anything blocking this from landing?

d0sboots:
> @d0sboots Anything blocking this from landing?

Nothing, aside from me getting a chance to look at it. 

One thing that would make my life easier would be to be a little less rebase-happy. I understand the impulse, since I make heavy use of rebase to make meaningful commits myself. XD However, when merging we use squash-merge, which collapses the whole PR to a single commit, so the only purpose to separate commits during the PR is to make it easier to diff progress/changes over the course of the PR. This gets destroyed when you rebase.

tomprince:
> One thing that would make my life easier would be to be a little less rebase-happy.

I can try and avoid rebasing forward, but probably not more than that, since keeping a clean history makes it easier for me to keep track of what I've done.

d0sboots:
> > One thing that would make my life easier would be to be a little less rebase-happy.
> 
> I can try and avoid rebasing forward, but probably not more than that, since keeping a clean history makes it easier for me to keep track of what I've done.

I think a useful way of thinking about it is that when you `git push`, it's like a publish action. You are publishing the state of that branch for others to see. (Well, in a PR. If you're doing stuff in your own private repo, it's your own business.)

If you make subsequent changes, then they logically happen on top of the state that you published. This is both what git does by default (all commits are based on the commits before them), and works well with reviewer workflows. You're welcome to reorder and switch around your commit history to your hearts content in that unpublished state, but (ideally) when you `git push` again it should be with a set of commits that are based off the previously pushed commit (and git will complain if this is not the case).

This might not line up with an idea of splitting up subfeatures into different commits or similar, but (for one thing) a single PR shouldn't be so big that that is necessary. And even if it were, reordering your history for your convenience pales in comparison to forcing your reviewer to re-review *the entire PR* every time there is a change, which is the practical impact of rebasing.

src/ScriptEditor/ui/ScriptEditorRoot.tsx
```diff
@@ -187,6 +187,51 @@ function Root(props: IProps): React.ReactElement {
     debouncedCodeParsing(newCode);
   };
 
+  function openFile(hostname: string, filename: ContentFilePath, code?: string) {
+    if (!editorRef.current) {
+      return;
+    }
+    if (!code) {
```
d0sboots, line 194: You're doing a lot of `!code` checks here, but code could be the empty string.

PR #1468, opened 2024-07-07
EDITOR: Update monaco-editor to 0.50.0.

tomprince:
This is in preparation for supporting cross-file go-to-definition. (See #1467)


d0sboots:
I'm going to assume that you checked and the editor doesn't have any noticable issues.

PR #1470, opened 2024-07-07
EDITOR: Work around bug in monaco-editor

catloversg:
The new version of monaco-editor has a bug: https://github.com/microsoft/monaco-editor/issues/4455.

Dev:
dev
dev-css

Stable:
stable
stable-css

The root cause is the unnecessary outline. That outline might be added mistakenly: https://github.com/microsoft/vscode/pull/209368.


d0sboots:
Thanks for that quick fix! Serves me right for not checking myself, although it does validate my reasoning about splitting the upgrade from functional changes.

PR #1474, opened 2024-07-08
BUGFIX: Fix "Router called before initialization" race

d0sboots:
If the game takes long enough to load, certain counters can become eligible to run as soon as Engine.start() runs. When this happens, eventually Router.page() is called, which throws an Error since Router isn't initialized yet. (Dropping a breakpoint before Engine.start() and waiting at least 30 seconds is enough to reliably repro, but I have seen this both live and in tests.)

This fixes it so that Router.page() is valid immediately, returning a value of Page.LoadingScreen. It also removes the isInitialized field, since this is now redundant. Trying to switch pages is still an error, but that doesn't happen without user input, whereas checking the current page is quite common.

This also consolidates a check for "should we show toasts" behind a function in Router, making the logic central and equal for a few usecases. This means (for instance) that the "autosave is disabled" logic won't run during infiltration. (The toast should have already been suppressed.)

### DIFF
--- a/src/Bladeburner/Bladeburner.ts
+++ b/src/Bladeburner/Bladeburner.ts
@@ -22,6 +22,7 @@ import { Skills } from "./data/Skills";
 import { City } from "./City";
 import { Player } from "@player";
 import { Router } from "../ui/GameRoot";
+import { Page } from "../ui/Router";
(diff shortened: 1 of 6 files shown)


d0sboots:
Submitting right away because it's bugfix.

PR #1475, opened 2024-07-08
BLADEBURNER: Add API to calculate max upgrade count of skill

### LINKED ISSUE #1071: Suggestion: Expose BB skill properties to BB API
I want to max out the number of BB skill upgrades that can be purchased from a given (current) amount of skill points.

This can be done by checking increasing numbers with `ns.bladeburner.getSkillUpgradeCost(skillName,count)`. However, this is basically "trial and error" and not very efficient for large numbers (e.g. Hyperdrive levels in the millions or more).

I checked the code for the `getSkillUpgradeCost` function: You "only" need to solve a pq-formula to get the number of upgrades.
Unfortunately, it requires four variables, of which two are not available for scripting.
- BladeburnerSkillCost --> exposed by Bitnode multipliers
- Current Skill Level --> exposed by bladeburner API
- the skill's BaseCost --> not exposed
- the skill's Cost increase --> not exposed

Would it be possible to expose the two properties to the BB API?

catloversg:
> How many skill levels can I buy with X skill points?

This PR adds a new API to calculate this value. The player won't need to do a "binary search" to get that value (the "mughurTime" script on Discord) or extract constants and formulas from source code.

Closes #1071.

Test code:
```js
(() => {
  let baseCost = 3;
  let costInc = 2.5;
  let mult = 1;

    if (!Number.isInteger(min)) {
      throw new Error(`Min is not an integer. Min: ${min}.`);
    }
    if (!Number.isInteger(max)) {
      throw new Error(`Max is not an integer. Max: ${max}.`);
    }
    if (min > max) {
    }
  }

  function calculateCost(currentLevel, count) {
    return Math.round(count * mult * (baseCost + costInc * (currentLevel + (count - 1) / 2)));
  }

  let adjustmentCount = 0;

  function calculateCount(currentLevel, cost) {
    const e = -2 * baseCost - 2 * costInc * currentLevel + costInc;
    const delta = Math.sqrt(mult * (mult * Math.pow(e, 2) + 8 * costInc * cost));

    // const result = Math.round((mult * e + delta) / (2 * mult * costInc));
    // const costOfResult = calculateCost(currentLevel, result);
    // if (cost < costOfResult) {
    //   ++adjustmentCount;
    //   return result - 1;
    // }
    // return result;

    const result = Math.floor((mult * e + delta) / (2 * mult * costInc));
    const costOfResultPlus1 = calculateCost(currentLevel, result + 1);
      ++adjustmentCount;
      return result + 1;
    }
    return result;
  }

  const baseCostList = [1, 2, 3];
  const costIncList = [1, 1.1, 1.4, 2.1, 2.5, 3];
  const multList = [1, 1.2, 2];

  let testCaseCount = 0;
  let errorCount = 0;
  for (let i = 0; i < baseCostList.length; ++i) {
    for (let j = 0; j < costIncList.length; ++j) {
      for (let k = 0; k < multList.length; ++k) {
        baseCost = baseCostList[i];
        costInc = costIncList[j];
        mult = multList[k];

        for (let m = 0; m < 1e5; ++m) {
          const currentLevel = getRandomIntInclusive(0, 1e3);
          let count = 0;
          let cost = 0;

          // Test 1
          ++testCaseCount;
          const expectedCount = getRandomIntInclusive(0, 1e6);
          cost = calculateCost(currentLevel, expectedCount);
          count = calculateCount(currentLevel, cost);
          if (expectedCount !== count) {
            ++errorCount;
            console.error(`baseCost: ${baseCost}, costInc: ${costInc}, mult: ${mult}, currentLevel: ${currentLevel}`);
            console.error(`cost: ${cost}, count: ${count}, expectedCount: ${expectedCount}`);
            console.log();
          }

          // Test 2
          ++testCaseCount;
          const budget = getRandomIntInclusive(1e9, 1e15);
          count = calculateCount(currentLevel, budget);
          cost = calculateCost(currentLevel, count);
          const costOfCountPlus1 = calculateCost(currentLevel, count + 1);
          if (budget < cost || budget > costOfCountPlus1) {
            ++errorCount;
            console.error(`baseCost: ${baseCost}, costInc: ${costInc}, mult: ${mult}, currentLevel: ${currentLevel}`);
            console.error("count".padEnd(20, " "), count);
            console.error("budget".padEnd(20, " "), budget);
            console.error("cost".padEnd(20, " "), cost);
            console.error("costOfCountPlus1".padEnd(20, " "), costOfCountPlus1);
            console.log();
          }
        }
      }
    }
  }
  console.log(
    `testCaseCount: ${testCaseCount.toExponential()}, errorCount: ${errorCount}, adjustmentCount: ${adjustmentCount.toExponential()}`,
  );
})();
```


d0sboots:
High-level comment: This strongly feels like it belongs in formulas. There might not be bladeburner formulas yet, in which case this would be the first? 

High-level comment 2: Some/all of that test code should be actual tests.

src/Bladeburner/Skill.ts
```diff
@@ -80,6 +80,62 @@ export class Skill {
     );
   }
 
+  calculateMaxUpgradeCount(currentLevel: number, cost: PositiveInteger): number {
+    /**
+     * Note: there is no notation for Math.round, so I use \lceil and \rceil as alternatives for non-existent \lround
```
d0sboots: For this, I would just use `\mathrm{round}(...)`, although the real answer is that I would drop the latex entirely, since it does not render in comments and thus hinders rather than aids understanding.

src/Bladeburner/Skill.ts
```diff
@@ -80,6 +80,62 @@ export class Skill {
...
+     *
+     * $$ x_2 = \frac{a \ast e - Delta}{2 \ast a \ast c} $$
+     *
+     * $a$, $c$ and $y$ are always greater than 0, so $x_2$ is always less than 0. Therefore, $x_1$ is the only
+     * solution.
+     */
+    const e = -2 * this.baseCost - 2 * this.costInc * currentLevel + this.costInc;
+    const delta = Math.sqrt(
+      currentNodeMults.BladeburnerSkillCost *
+        (currentNodeMults.BladeburnerSkillCost * Math.pow(e, 2) + 8 * this.costInc * cost),
```
d0sboots: Just use e * e.

src/Bladeburner/Skill.ts
```diff
@@ -80,6 +80,62 @@ export class Skill {
...
+    );
+    // We ignore Math.round in the previous part, so x2 is just an estimation. In order to not "overestimate" the final
+    // result, we use Math.floor here, then "adjust" it later if necessary.
```
d0sboots: You don't justify that you only need to check 1 level above/below. Indeed, for arbitrary constants I don't think this holds - it only works because the constants are in the ranges they are.

catloversg:
### Floating-point imprecision

At first, let's check this example to see the problem of floating-point imprecision.

```js
(() => {
  const baseCost = 3;
  const costInc = 9.19728658847669;
  const mult = 8.79462273023623;
  const currentLevel = 216;
  const count = 982743;
  function calculateCost(currentLevel, count) {
    return count * mult * (baseCost + costInc * (currentLevel + (count - 1) / 2));
  }
  console.log(`cost        : ${calculateCost(currentLevel, count).toPrecision(21)}`);
})();

(() => {
  // https://github.com/MikeMcl/decimal.js
  Decimal.set({ precision: 100 });
  const dBaseCost = new Decimal("3");
  const dCostInc = new Decimal("9.19728658847669");
  const dMult = new Decimal("8.79462273023623");
  const dCurrentLevel = new Decimal("216");
  const dCount = new Decimal("982743");
  function calculateDecimalCost(dCurrentLevel, dCount) {
    return dCount.times(dMult).times(dBaseCost.plus(dCostInc.times(dCurrentLevel.plus(dCount.minus(1).dividedBy(2)))));
  }
  console.log(`decimal cost: ${calculateDecimalCost(dCurrentLevel, dCount)}`);
})();
```

Output:
```
cost        : 39076672039194.4921875
decimal cost: 39076672039194.5033979264180544313133875967
```

Alternative: paste this expression to a calculator: `982743 * 8.79462273023623 * (3 + 9.19728658847669 * (216 + (982743 - 1) / 2))`.

As we can see here, if the player has 39076672039194 SP, they can upgrade the skill 982743 times due to the floating-point imprecision. In the next part, I'll show a solution that is supposed to work properly, but it does not.

### Solution

Define:
- x = count
- a = currentNodeMults.BladeburnerSkillCost
- b = this.baseCost
- c = this.costInc
- d = currentLevel
- y = cost

Cost function:

$$ f(x) = \mathrm{Round}(x \ast a \ast (b + c \ast (d + \frac{x - 1}{2}))) $$

It's hard to deal with `Math.round(x)`, so we will use `Math.floor(x + 0.5)`. `Math.round(x)` is not exactly the same as `Math.floor(x + 0.5)` in some cases (e.g. `0.49999999999999994`), but it's fine for our cases.

Ref:
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/round
- https://tc39.es/ecma262/multipage/numbers-and-dates.html#sec-math.round

We change it to:

$$ g(x) = \lfloor x \ast a \ast (b + c \ast (d + \frac{x - 1}{2})) + 0.5 \rfloor $$

We need to find the largest positive integer $x_0$ such that $g(x_0) \le y$. Now, let's find the range of $x$ such that $g(x) = y$.

$$ \lfloor x \ast a \ast (b + c \ast (d + \frac{x - 1}{2})) + 0.5 \rfloor = y $$

≡

$$\begin{cases} x \ast a \ast (b + c \ast (d + \frac{x - 1}{2})) + 0.5 \ge y \newline x \ast a \ast (b + c \ast (d + \frac{x - 1}{2})) + 0.5 < y + 1 \end{cases}$$

Define:

$$ e = -2 \ast b - 2 \ast c \ast d + c $$

$$ Delta_1 = \sqrt{a \ast (a \ast {e ^ 2} + 4 \ast c \ast (2 * y - 1))} $$

$$ Delta_2 = \sqrt{a \ast (a \ast {e ^ 2} + 4 \ast c \ast (2 * y + 1))} $$

We have these conditions:
- $a$ and $c$ are always greater than 0.
- $x$ and $y$ must be positive integers.

We have:

$$ x \ast a \ast (b + c \ast (d + \frac{x - 1}{2})) + 0.5 \ge y $$

≡

$$ x \ge \frac{a \ast e + Delta_1}{2 \ast a \ast c} $$

And:

$$ x \ast a \ast (b + c \ast (d + \frac{x - 1}{2})) + 0.5 < y + 1 $$

≡

$$ x < \frac{a \ast e + Delta_2}{2 \ast a \ast c} $$

We have this conclusion: $x_0$ is the largest integer such that:

$$ x_0 < \frac{a \ast e + Delta_2}{2 \ast a \ast c}$$

### Test

Let's test it:

```js
let baseCost = 3;
let costInc = 2.5;
let mult = 1;

  if (!Number.isInteger(min)) {
    throw new Error(`Min is not an integer. Min: ${min}.`);
  }
  if (!Number.isInteger(max)) {
    throw new Error(`Max is not an integer. Max: ${max}.`);
  }
  if (min > max) {
  }
}

function getRandomArbitrary(min, max) {
  if (min > max) {
  }
  return Math.random() * (max - min) + min;
}

function calculateCost(currentLevel, count) {
  return Math.round(count * mult * (baseCost + costInc * (currentLevel + (count - 1) / 2)));
}
```

```js
function calculateCount(currentLevel, cost) {
  const e = -2 * baseCost - 2 * costInc * currentLevel + costInc;
  const delta = Math.sqrt(mult * (mult * Math.pow(e, 2) + 4 * costInc * (2 * cost + 1)));
  return Math.floor((mult * e + delta) / (2 * mult * costInc));
}

const baseCostList = [1, 2, 3];
const costIncList = [1, 1.1, 1.4, 2.1, 2.5, 3];
const multList = [1, 1.2, 2];

let testCaseCount = 0;
let errorCount = 0;
for (let i = 0; i < baseCostList.length; ++i) {
  for (let j = 0; j < costIncList.length; ++j) {
    for (let k = 0; k < multList.length; ++k) {
      baseCost = baseCostList[i];
      costInc = costIncList[j];
      mult = multList[k];

      for (let m = 0; m < 1e5; ++m) {
        const currentLevel = getRandomIntInclusive(0, 1e3);
        let count = 0;
        let cost = 0;

        // Test 1
        ++testCaseCount;
        const expectedCount = getRandomIntInclusive(0, 1e6);
        cost = calculateCost(currentLevel, expectedCount);
        count = calculateCount(currentLevel, cost);
        if (expectedCount !== count) {
          ++errorCount;
          console.error(
            `baseCost: ${baseCost}, costInc: ${costInc.toPrecision(21)}, mult: ${mult.toPrecision(
              21,
            )}, currentLevel: ${currentLevel}`,
          );
          console.error(`cost: ${cost}, count: ${count}, expectedCount: ${expectedCount}`);
          console.log();
        }

        // Test 2
        ++testCaseCount;
        const budget = getRandomIntInclusive(1e9, 1e15);
        count = calculateCount(currentLevel, budget);
        cost = calculateCost(currentLevel, count);
        const costOfCountPlus1 = calculateCost(currentLevel, count + 1);
        if (budget < cost || budget >= costOfCountPlus1) {
          ++errorCount;
          console.error(`baseCost: ${baseCost}, costInc: ${costInc}, mult: ${mult}, currentLevel: ${currentLevel}`);
          console.error("count".padEnd(20, " "), count);
          console.error("budget".padEnd(20, " "), budget);
          console.error("cost".padEnd(20, " "), cost);
          console.error("costOfCountPlus1".padEnd(20, " "), costOfCountPlus1);
          console.log();
        }
      }
    }
  }
}
console.log(`testCaseCount: ${testCaseCount.toExponential()}, errorCount: ${errorCount}`);
```

Output:
```
testCaseCount: 1.08e+7, errorCount: 0
```

It looks good. However, another test shows the problem:
```js
function calculateCount(currentLevel, cost) {
  const e = -2 * baseCost - 2 * costInc * currentLevel + costInc;
  const delta = Math.sqrt(mult * (mult * Math.pow(e, 2) + 4 * costInc * (2 * cost + 1)));
  return Math.floor((mult * e + delta) / (2 * mult * costInc));
}

let testCaseCount = 0;
let errorCount = 0;
  baseCost = getRandomIntInclusive(1, 100);
  for (let j = 0; j < 10; ++j) {
    costInc = getRandomArbitrary(1, 100);
    for (let k = 0; k < 10; ++k) {
      mult = getRandomArbitrary(1, 100);
      for (let m = 0; m < 1e4; ++m) {
        const currentLevel = getRandomIntInclusive(0, 1e3);
        let count = 0;
        let cost = 0;

        // Test 1
        ++testCaseCount;
        const expectedCount = getRandomIntInclusive(0, 1e9);
        cost = calculateCost(currentLevel, expectedCount);
        count = calculateCount(currentLevel, cost);
        if (expectedCount !== count) {
          ++errorCount;
          console.error(
            `baseCost: ${baseCost}, costInc: ${costInc.toPrecision(21)}, mult: ${mult.toPrecision(
              21,
            )}, currentLevel: ${currentLevel}`,
          );
          console.error(`cost: ${cost}, count: ${count}, expectedCount: ${expectedCount}`);
          console.log();
        }

        // Test 2
        ++testCaseCount;
        const budget = getRandomIntInclusive(1e9, 1e15);
        count = calculateCount(currentLevel, budget);
        cost = calculateCost(currentLevel, count);
        const costOfCountPlus1 = calculateCost(currentLevel, count + 1);
        if (budget < cost || budget >= costOfCountPlus1) {
          ++errorCount;
          console.error(`baseCost: ${baseCost}, costInc: ${costInc}, mult: ${mult}, currentLevel: ${currentLevel}`);
          console.error("count".padEnd(20, " "), count);
          console.error("budget".padEnd(20, " "), budget);
          console.error("cost".padEnd(20, " "), cost);
          console.error("costOfCountPlus1".padEnd(20, " "), costOfCountPlus1);
          console.log();
        }
      }
    }
  }
}
console.log(`testCaseCount: ${testCaseCount.toExponential()}, errorCount: ${errorCount}`);
```
Output:

```
testCaseCount: 2e+6, errorCount: 220197
```

Let's enable logging and check a random case:

```
baseCost: 77, costInc: 89.1826908286195561004, mult: 99.6751495532347604467, currentLevel: 592
cost: 265068399506855520, count: 7721948, expectedCount: 7721949
```

Code:

```js
(() => {
  const baseCost = 77;
  const costInc = 89.1826908286195561004;
  const mult = 99.6751495532347604467;
  const currentLevel = 592;
  const count = 7721949;
  function calculateCost(currentLevel, count) {
    return count * mult * (baseCost + costInc * (currentLevel + (count - 1) / 2));
  }
  console.log(`cost        : ${calculateCost(currentLevel, count).toPrecision(21)}`);
})();

(() => {
  // https://github.com/MikeMcl/decimal.js
  Decimal.set({ precision: 100 });
  const dBaseCost = new Decimal("77");
  const dCostInc = new Decimal("89.1826908286195561004");
  const dMult = new Decimal("99.6751495532347604467");
  const dCurrentLevel = new Decimal("592");
  const dCount = new Decimal("7721949");
  function calculateDecimalCost(dCurrentLevel, dCount) {
    return dCount.times(dMult).times(dBaseCost.plus(dCostInc.times(dCurrentLevel.plus(dCount.minus(1).dividedBy(2)))));
  }
  console.log(`decimal cost: ${calculateDecimalCost(dCurrentLevel, dCount)}`);
})();
```

Output:

```
cost        : 265068399506855520.000
decimal cost: 265068399506855539.27885540074000492372076340898914268312
```

This is the same as the example that we see in the first part. Our formula is correct, but the floating-point imprecision forces us to "adjust" the final value.

```js
function calculateCount(currentLevel, cost) {
  const e = -2 * baseCost - 2 * costInc * currentLevel + costInc;
  const delta = Math.sqrt(mult * (mult * Math.pow(e, 2) + 4 * costInc * (2 * cost + 1)));
  const result = Math.floor((mult * e + delta) / (2 * mult * costInc));
  const costOfResultPlus1 = calculateCost(currentLevel, result + 1);
    return result + 1;
  }
  return result;
}
```
Plug this code into the previous test code, and we can see that there is no error.

I don't have a rigorous proof that "plus 1" is enough to "compensate" the floating-point imprecision. The testing shows that "plus 1" is good enough, though.

@d0sboots What's your opinion?

d0sboots:
This is the same approach (essentially) that is used for `formulas.hacking.growThreads`, and it is the only approach that will function over a complete range. It has 2 parts:
* Calculate an approximation that you can guarantee to be within 1 of the correct answer.
* Use the actual function to check the adjacent values to determine the true answer.

Usually a combination of mathematical justification and careful testing goes into justifying the first assumption. I can help provide the mathematical justification; you should solidify your ad-hoc tests into unit tests to satisfy the 2nd part.

I still think this should be in formulas.

d0sboots:
Oh, as a side note: `e` is a bad name for a constant/variable.

d0sboots:
Also, your math has an error in it right now. For instance, with costInc = baseCost = 1 and mult = 1.2, `calculateCount(1, 181.9)` gives 16, despite the fact that `calculateCost(1, 16)` gives 182 (i.e. 181.9 is insufficient).

catloversg:
181.9 is not a valid value for "cost" (skill points). It must be a positive integer because we use `Math.round` in the cost function. SP must be either 181 or 182.

d0sboots:
There's no such limitation for `calculateCount` though, nor should there be. Just because the game never creates a situation with fractional cost, does not mean that asking the question "how many upgrades can I get for this cost?" is only valid/defined for integers. 

I suspect with some work I could also find an example that makes use of floating-point rounding to achieve the same thing for an integer value, it would just take a lot more effort. The underlying issue remains the same, though.

catloversg:
If that's the case, we are back to the current implementation of `calculateMaxUpgradeCount`. In the proof for that function, I "ignore" the `Math.round` in the cost function.

Test code:
```js
(() => {
  const baseCost = 1;
  const costInc = 1;
  const mult = 1.2;
  const currentLevel = 1;
  const count = 16;
  function calculateCount(currentLevel, cost) {
    const e = -2 * baseCost - 2 * costInc * currentLevel + costInc;
    const delta = Math.sqrt(mult * (mult * Math.pow(e, 2) + 8 * costInc * cost));
    const result = Math.floor((mult * e + delta) / (2 * mult * costInc));
    const costOfResultPlus1 = calculateCost(currentLevel, result + 1);
      return result + 1;
    }
    return result;
  }
  function calculateCost(currentLevel, count) {
    return Math.round(count * mult * (baseCost + costInc * (currentLevel + (count - 1) / 2)));
  }
  console.log(`cost : ${calculateCost(currentLevel, count).toPrecision(21)}`);
  console.log(`count: ${calculateCount(currentLevel, 182).toPrecision(21)}`);
  console.log(`count: ${calculateCount(currentLevel, 181.9).toPrecision(21)}`);
  console.log(`count: ${calculateCount(currentLevel, 181).toPrecision(21)}`);
})();
```
```js
function calculateCount(currentLevel, cost) {
  const e = -2 * baseCost - 2 * costInc * currentLevel + costInc;
  const delta = Math.sqrt(mult * (mult * Math.pow(e, 2) + 8 * costInc * cost));
  const result = Math.floor((mult * e + delta) / (2 * mult * costInc));
  const costOfResultPlus1 = calculateCost(currentLevel, result + 1);
    return result + 1;
  }
  return result;
}

let testCaseCount = 0;
let errorCount = 0;
  baseCost = getRandomIntInclusive(1, 100);
  for (let j = 0; j < 10; ++j) {
    costInc = getRandomArbitrary(1, 100);
    for (let k = 0; k < 10; ++k) {
      mult = getRandomArbitrary(1, 100);
      for (let m = 0; m < 1e4; ++m) {
        const currentLevel = getRandomIntInclusive(0, 1e3);
        let count = 0;
        let cost = 0;

        // Test 1
        ++testCaseCount;
        const expectedCount = getRandomIntInclusive(0, 1e9);
        cost = calculateCost(currentLevel, expectedCount);
        count = calculateCount(currentLevel, cost);
        if (expectedCount !== count) {
          ++errorCount;
          console.error(
            `baseCost: ${baseCost}, costInc: ${costInc.toPrecision(21)}, mult: ${mult.toPrecision(
              21,
            )}, currentLevel: ${currentLevel}`,
          );
          console.error(`cost: ${cost}, count: ${count}, expectedCount: ${expectedCount}`);
          console.log();
        }

        // Test 2
        ++testCaseCount;
        const budget = getRandomIntInclusive(1e9, 1e15);
        count = calculateCount(currentLevel, budget);
        cost = calculateCost(currentLevel, count);
        const costOfCountPlus1 = calculateCost(currentLevel, count + 1);
        if (budget < cost || budget >= costOfCountPlus1) {
          ++errorCount;
          console.error(`baseCost: ${baseCost}, costInc: ${costInc}, mult: ${mult}, currentLevel: ${currentLevel}`);
          console.error("count".padEnd(20, " "), count);
          console.error("budget".padEnd(20, " "), budget);
          console.error("cost".padEnd(20, " "), cost);
          console.error("costOfCountPlus1".padEnd(20, " "), costOfCountPlus1);
          console.log();
        }
      }
    }
  }
}
console.log(`testCaseCount: ${testCaseCount.toExponential()}, errorCount: ${errorCount}`);
```

catloversg:
By the way, I'm implementing your other suggestions (move to the formula namespace, add Jest test, etc.). After you write the "mathematical justification" that you mentioned, I'll add it to the comment of `calculateMaxUpgradeCount`.

d0sboots:
A couple other notes while I've been analyzing this: 
* Everything is greatly simplified if you divide out the factor of mult.
* If you divide by the 2 in the denominator, you don't need to artificially multiply the terms in `e` by 2, and it eliminates the multiplication by 4 inside the square root.

d0sboots:
OK, after wearing my numerical analysis hat, here is my conclusion: 
I think you should change the current end of the procedure, which is "floor and then return y+1 or y" to "round, and then return y+1,y or y-1."

The reason is that the current approach potentially leaves 0 margin on the low side. Given a result that is an exact integer due to floating-point error, it could return a value that is 1 too large. Using round and (potentially) checking both sides gives a guaranteed 0.5 margin for error, which is more than sufficient given the accuracy of the calculation.

catloversg:
New changes:
- Simplify the formula based on https://github.com/bitburner-official/bitburner-src/pull/1475#issuecomment-2234350699
- "round, and then return y+1,y or y-1."
- Move to the "formulas" namespace.
- Add Jest test.

I have not written comments for the "round, and then return y+1,y or y-1." in `calculateMaxUpgradeCount`. To be honest, due to the lack of a rigorous proof, I'm not confident in either the old "floor and then return y+1 or y" or the new one. Your reasoning is good, and I'm not against it. Can you write a comment for it? I'll copy it to the code.

src/Bladeburner/Skill.ts
```diff
@@ -108,35 +102,31 @@ export class Skill {
...
      * Solutions:
      *
-     * $$ x_1 = \frac{a \ast e + Delta}{2 \ast a \ast c} $$
+     * $$ x_1 = \frac{m + Delta}{c} $$
      *
-     * $$ x_2 = \frac{a \ast e - Delta}{2 \ast a \ast c} $$
+     * $$ x_2 = \frac{m - Delta}{c} $$
      *
      * $a$, $c$ and $y$ are always greater than 0, so $x_2$ is always less than 0. Therefore, $x_1$ is the only
      * solution.
```
d0sboots, line 116: I've been letting this sit for too long. Here's my attempt at a comment (it's far short of a true proof):

Due to floating-point rounding and edge-cases, we cannot ensure that rounding x_1 will give us the correct integer. In other words, we cannot be sure that x_1 is within 0.5 of the integer value we want. But, we *can* be sure that it is within *1* of the value we want, which means that checking the numbers above and below the rounded value are sufficient to find our correct integer.

src/utils/helpers/getRandomArbitrary.ts
d0sboots, line 1: Oops, I missed that you snuck this in there. The standard naming for this is `randomRange`.

PR #1477, opened 2024-07-09
EDITOR:  Use a custom resolver to match bitburner import semantics.

tomprince:
In working on various bits of the editor (e.g. #1467, #1138) I discovered that the typescript language server doesn't follow the same import resolution that bitburner does. However, it does provide a hook that allows you to override the module resolution algorithm. With a bit of work, it is possible to convince monaco to provide that hook, although I ran into some sharp edges in webpack while doing so.

The code here is somewhat ugly, and probably generates bigger bundles than we'd want, but demonstrates the idea.


d0sboots:
Can you explain more about what problem this is solving? I.e. what is a repro where monaco gets it wrong currently.

Aside from that, there's some complicated stuff going on that will need more explanatory comments, but I suspect this will become more important once The TSX change is merged.

tomprince:
> Can you explain more about what problem this is solving? I.e. what is a repro where monaco gets it wrong currently.

Some cases that I am aware of:
- In `dir/test.js`, `import "/lib/some-lib.js` does not get resolved. You need to use `lib/some-lib.js`, but both work when running.
- If you have both `lib/some-lib.js` and `dir/lib/some-lib.js`, then in `dir/test.js` monaco resolves `import "lib/some-lib.js"` to `dir/lib/some-lib.js` but at runtime it resolves to `lib/some-lib.js`.
- If you have a file `lib.js` on `n00dles`, then monaco will resolve `import "n00dles/lib.js";` to that file anywhere.

In terms of the behavior:
- When resolving an absolute path (i.e. starting with a `/`) relative to a file with a URL filename, it just drops the URL. (I think this is a bug in typescript.
- When resolving a "bare specifier" (i.e. not starting with `/` or `./` or `../`) it looks for the file in the current directory, and then in each parent directory. This is because the typescript language server is defaulting to the [`classic`](https://www.typescriptlang.org/docs/handbook/modules/reference.html#classic) module resolution style. However, we can't solve this by switching to a different module resolution style, since we want to resolve bare specifiers relative to the host of the current file, which would require different options for each host.
- The issue accessing files on other hosts is addressed by changing the URL to be `file://{server}/{path}` instead of `file:///{server}/{path}`.

d0sboots:
I feel like this can possibly be solved be getting off of classic module resolution style, and introducing some other options. However, I haven't looked enough into the monaco system to know if that's viable. Surely, we can solve this from the other direction by just changing up how we map our files to Monaco paths?

Caldwell-74:
iam not 100% certain but i think the only thing we could do "from the other side" is to "fail" the import when its a format which monaco cant correctly process

tomprince:
> I feel like this can possibly be solved be getting off of classic module resolution style, and introducing some other options.

I don't *think* so. All the other module resolution styles look for bare module specifiers either
- in a specific single directory when `baseURL` is specified. (see [here](https://www.typescriptlang.org/docs/handbook/modules/reference.html#baseurl)
- in various `node_modules` directories (see [here](https://www.typescriptlang.org/docs/handbook/modules/reference.html#node_modules-package-lookups)
From my research, the problem boils down partly to the fact that typescript wants to look for bare specifiers in a single directory, but we want to vary where to look for them based on the host of containing script.

> However, I haven't looked enough into the monaco system to know if that's viable. Surely, we can solve this from the other direction by just changing up how we map our files to Monaco paths?

If we got the bug in typescript where typescript throws away the URL when looking up an path starting with `/` (and got that version of typescript incorporated into monaco) *and* we decided to forbid bare module specifiers, then I think the first commit would make things work.

src/Paths/Directory.ts
```diff
@@ -28,7 +26,7 @@ export const root = "" as Directory;
 const invalidCharacters = ["/", "*", "?", "[", "]", "!", "\\", "~", "|", "#", '"', "'", " "];
 
 /** A valid character is any character that is not one of the invalid characters */
-export const oneValidCharacter = `[^${escapeRegExp(invalidCharacters.join(""))}]`;
+export const oneValidCharacter = `[^${invalidCharacters.map((c) => "\\" + c).join("")}]`;
```
tomprince: To easily import the path manipulation function into the worker, there can't be any imports of modules that would go into the vendor bundle. Thus, I just manually escape the characters here.

src/Paths/ListFiles.ts
```diff
@@ -1,6 +1,27 @@
-import { BaseServer } from "../Server/BaseServer";
-import { root } from "./Directory";
-import { ContentFileMap, allContentFiles } from "./ContentFile";
+import type { BaseServer } from "../Server/BaseServer";
+import type { ContentFile, ContentFileMap, ContentFilePath } from "./ContentFile";
+import { type Directory, root } from "./Directory";
+import type { FilePath } from "./FilePath";
+
+/** Generator function to allow iterating through all content files on a server */
+export function* allContentFiles(server: BaseServer): Generator<[ContentFilePath, ContentFile], void, undefined> {
```
tomprince, line 7: I moved these here, to avoid imports from outside `Paths/` (to prevent recursive imports of stuff that goes in the vendor bundle). But, it turns out we only need type imports of those things, so this bit could be dropped.

src/utils/ScriptTransformer.ts
```diff
@@ -110,7 +116,7 @@ export function getModuleScript(
   scripts: Map<ScriptFilePath, Script>,
 ): Script {
   let script;
-  for (const extension of validScriptExtensions) {
+  for (const extension of isLegacyScript(baseModule) ? [legacyScriptExtension] as const: scriptExtensions as ) {
```
tomprince: This is fixing a bug where it would try to import a legacy script file from an non-legacy script. The check for legacy script types is to support the ram calculation logic.

d0sboots:
> > I feel like this can possibly be solved be getting off of classic module resolution style, and introducing some other options.
> 
> I don't _think_ so. All the other module resolution styles look for bare module specifiers either
> 
> * in a specific single directory when `baseURL` is specified. (see [here](https://www.typescriptlang.org/docs/handbook/modules/reference.html#baseurl)
> * in various `node_modules` directories (see [here](https://www.typescriptlang.org/docs/handbook/modules/reference.html#node_modules-package-lookups)
>   From my research, the problem boils down partly to the fact that typescript wants to look for bare specifiers in a single directory, but we want to vary where to look for them based on the host of containing script.

OK, I think I see the shape of the issue. Typescript's module resolution rules aren't the problem. (Well, they are, but IMO they're not the core of the problem.) The problem comes from the fact that you can edit files from multiple servers at the same time.

This has caused a continuous host of bugs and issues. It's clear that BitBurner isn't well-designed to do this: We don't show the server anywhere in the UI, so you're just expected to "know" which server you're editing on. In the past, if you switched servers while editing code, it would even switch which server the file would be saved to!

It's clear that we expect people to "mostly" stick to editing files on home, but at the same time want to support editing files anywhere. However, that doesn't necessarily mean we need to support having files from multiple servers open at the same time. Food for thought.

tomprince:
> We don't show the server anywhere in the UI, so you're just expected to "know" which server you're editing on.

We actually do show the server of each edited file in the editor tab bar.

tsconfig.json
```diff
@@ -9,7 +9,8 @@
     "esModuleInterop": true,
     "isolatedModules": true,
     "jsx": "react",
-    "module": "commonjs",
+    "module": "ES2022",
```
tomprince, line 12: In particular, this tells typescript we can use `import.meta.url` in our code.

tsconfig.json
```diff
@@ -9,7 +9,8 @@
...
+    "moduleResolution": "Bundler",
```
tomprince, line 13: This tells typescript we are using a bundler (i.e. webpack), and so to use the module resolution conventions typical of bundlers.

tomprince (commented):
I think I am more-or-less happy with the state of this. I can imagine that the code could use more comments (which I am happy to add). It probably also needs some more extensive testing than I've done.

src/Paths/Directory.ts
d0sboots: Not all of these form valid escape sequences. For instance, `\ ` is not valid, which becomes an identity escape sequence in Unicode-unaware mode, but an error in Unicode-aware mode.

`invalidCharacters` is *only* used to form this regex, so if you want to avoid lodash I would just construct this regex literally and use .source to get the value (to avoid escaping issues).

```suggestion
export const oneValidCharacterRe = /[^/*?[\]!\\~|#"' ]/;
```

d0sboots: This needs to be commented somewhere, because the restriction is highly non-obvious. Is this for space reasons, or functional reasons? If it's not something that will cause the build to break, it *will* need to be enforced with a test, otherwise this constraint will be violated very quickly.

src/ScriptEditor/worker-url.ts
```diff
@@ -0,0 +1,13 @@
+// Webpack has syntax for passing the URL of a script to be loaded in a web worker to
+// various functions (such as `new Worker(...)`). However, since we just want the URL of
+// the script, we tell webpack that in this module, `String(new URL(...))` should be
+// treated as creating a web worker. That gives us access to the URL of the script,
```
d0sboots: You wrote an extensive comment here, but unfortunately I don't understand what it's trying to say. :(

I *suspect* what this is about is webpack's handling of relative URLs, and your use of the webpackChuckName override, but that doesn't explain why this is in its own module.

src/ScriptEditor/worker.ts
```diff
@@ -0,0 +1,102 @@
...
+    /**
+     * Resolve a list of module identifier relative to the containing module.
+     * @param moduleLiterals List of module identifiers to resolve.
+     * @param containingFile The monaco URI of the containing module, as a string.
+     * @returns An array of module resolutions, in the same order as arguments.
+     */
+    resolveModuleNameLiterals?(
+      moduleLiterals: readonly ts.StringLiteralLike[],
+      containingFile: string,
+      __redirectedReference: ts.ResolvedProjectReference | undefined,
```
d0sboots, line 66: What's with the parameter naming here?

src/ScriptEditor/worker.ts
```diff
@@ -0,0 +1,102 @@
...
+            break;
+          } else {
+            failedLookupLocations.push(candidateURI);
+          }
+        }
+        return {
+          resolvedModule: resolvedFileName
+            ? {
+                resolvedFileName,
+                extension: allScriptExtensions.find((ext) => resolvedFileName!.endsWith(ext))!,
```
d0sboots, line 93: The non-null assertions really shouldn't be needed here. If they still are, I'd rewrite this as an `if`, since it's getting hefty for a ternary anyway, and that should fix it.

d0sboots (commented):
Finally getting around to these. Unfortunately, there's still a lot of testing to be done for such a core change.

src/ScriptEditor/worker-url.ts
tomprince: No, it isn't really about relative URLs.

Webpack has support for creating a bundle of code for web workers (see [here](https://webpack.js.org/guides/web-workers/)). However, it expects a very particular syntax:
```javascript
new Worker(new URL('./worker.js', import.meta.url));
```
In particular, you need to create a new `URL` object with base `import.meta.url` and pass that to `new Worker` (or one of a couple of other things. However, we don't want to create a worker, just pass the URL of a script to load in one.

You can override this (which is what the ` parser: { worker: ["String()"] }` bit in the webpack config is), but I couldn't figure out a better thing to use than `String()`. Since that is so generic, I wanted to limit the amount of code where webpack is looking for that magic syntax, hence the file.

src/ScriptEditor/worker.ts
tomprince: IIRC, eslint.

src/Paths/Directory.ts
tomprince: I guess the comment in `worker.js` got lost in one of the iterations.

The issue is that `worker.ts` needs to *synchronously* define `customTSWorkerFactory`. If it (recursively) imports anything stuff that gets put into `vendor.bundle.js` (i.e. from `node_modules), then it gets defined asynchously.

I'm not sure how to get it to fail the build, but it errors out as soon as you open the editor.

Another option would be to write much messier code to do the imports in `worker.js` (so the imports are not at the top-level), rather than impose this restriction.

tomprince: It looks like this is a bug in webpack (https://github.com/webpack/webpack/issues/18615), though we'll need to work around it for the moment.

src/ScriptEditor/worker-url.ts
d0sboots: OK, I think you're going to need to include most of that explanation in the comment. The file will end up being more comment than code, but that is OK/to be expected, because what you are doing here is very strange and obscure, and thus requires a lot of context.

src/Paths/Directory.ts
d0sboots: OK, so there's two major things that need doing here: 
1. The situation needs to be documented. An explanation of what's going on with synchronous loading in monaco with workers and a link to that bug, at minimum.
2. Either a test needs to be added, or you need to do the complicated thing. "We have to manually check to see if the editor still works on each change" is not an acceptable state of affairs.

webpack.config.js
```diff
@@ -93,120 +93,163 @@ module.exports = (env, argv) => {
         },
   };
 
-  return {
```
d0sboots, line 96: You're changing around a *lot* of stuff in the webpack.config now. How much of this is needed for configuring the worker, vs how much is general reformating/refactoring?

d0sboots:
Also, thank you for preserving the commit history here, it made it *much* easier to review the changes.

webpack.config.js
tomprince: Most of it is whitespace change, most of the rest of it is refactoring to avoid  duplicating code. 

This is using https://webpack.js.org/configuration/configuration-types/#exporting-multiple-configurations to have a second configuration for the custom worker file, which needs to specify some of the same options as the main configuration.

PR #1487, opened 2024-07-15
MISC: Always include stack trace in Recovery Mode

d0sboots:
We are getting some more error reports coming in that don't have enough info in them. It turns out that populating the stack trace was gated behind the dev flag; in reality, production builds are where we need it most. Even if it ends up being obfuscated (source maps should prevent this), we can figure out the actual source lines with enough effort if need be.

This also changes to using the actual stack trace, rather than the "component" trace (the tree of JSX objects), since knowing where the code failed is far more valuable. Also, it ensures we get the full error details when things go wrong in savefile loading.

image

Tested by trying recovery mode in various scenarios. The "file a bug" link also works as expected (still).


d0sboots:
I'm gonna submit this right away, so it doesn't miss any future releases.

PR #1490, opened 2024-07-16
BUGFIX: Add missing files in MathJax

### LINKED ISSUE #1489: Corporation - Changing component formula Math Renderer option crashes game
Changing the Math Renderer from CHTML to SVG causes the game to crash. After changing this option, the game will crash on startup every time due to missing file attempting to be loaded on startup.

To Reproduce:
Right-clicking on the component formula in the Corporation tab and follow the context menu -> Math Settings -> Math Renderer selecting SVG.
bitburner_mathrender_bug1

BitBurner v2.6.2 (633da3830) (July 03 2024 last updated).

Error message indicates trying to load a file named "svg.js" but the full path to file is not included in error prompt.
bitburner_mathrender_bug2

Digging through SteamDB history, looks like this file was included in older MathJax libraries (last found in 3.2.0) but is not included in the MathJax-3.2.2 library any more.
bitburner_mathrender_bug3

Can be resolved by creating an empty file in the MathJax-3.2.2 library for svg.js. Component formulas still render, and allows game to launch and change renderer back to CHTML.
bitburner_mathrender_bug4

catloversg:
It will take a while before I implement things that I and d0sboots talk about in #1489, so I will fix it with this PR and implement those things later.

These files are from https://www.npmjs.com/package/mathjax.

Fixes #1489.


tomprince:
I think it would make more sense to just copy the files from node modules (see #1491), rather than continue to vendor mathjax.

d0sboots:
This goes for both approaches: The things in the base directory (named mml-svg.js, etc.), are bundle files. We only need/want the one we are using (tex-chtml.js), since they are large and only usable by explicitly mentioning them in a script tag or similar. So the only things that should be added are the two files in output/ (and it's questionable if we need some of the other files, or if they're redundant - but that's not for this PR.)

tomprince:
> The things in the base directory (named mml-svg.js, etc.), are bundle files. We only need/want the one we are using (tex-chtml.js), since they are large and only usable by explicitly mentioning them in a script tag or similar.

Assuming this is correct, I've adjusted the copy plugin's configuration to only copy `tex-chtml.js` from the base directory, and copy everything in subdirectories. It is easy enough to tweak this further, but I'm inclined to not try to optimize more than this for the first step.

PR #1491, opened 2024-07-16
BUGFIX: Use mathjax from npm, rather than vendored copy.

### LINKED ISSUE #313: Cleanup dist/ext folder

### LINKED ISSUE #1489: Corporation - Changing component formula Math Renderer option crashes game

tomprince:
Fixes #1489. Closes #1490. Fixes #313.

We have a dependency on mathjax anyway, so rather than vendor all the files from mathjax, we can just grab them as part of the build process from `node_modules`.


d0sboots:
OMG, it didn't even occur to me that these might be checked in to source control, since they're in `dist/` and also that's insane.

tomprince:
Tested this with electron, and it doesn't work there, due to paths being slightly different between the devserver and electron. I suspect this is an easy fix though.

d0sboots:
Not at my computer ATM, but I was planning on looking this over carefully (the build artifacts) before merging.

tomprince:
Fixed so it works in both electron and the dev server. I did this by adjusting the path the dev server uses for copied files to match what gets built.

tomprince:
From https://github.com/bitburner-official/bitburner-src/pull/1490#issuecomment-2231373538
> The things in the base directory (named mml-svg.js, etc.), are bundle files. We only need/want the one we are using (tex-chtml.js), since they are large and only usable by explicitly mentioning them in a script tag or similar.

Assuming this is correct, I've adjusted the copy plugin's configuration to only copy `tex-chtml.js` from the base directory, and copy everything in subdirectories. It is easy enough to tweak this further, but I'm inclined to not try to optimize more than this for the first step.

d0sboots:
I had some unrelated yaks to shave, but eventually I was able to verify that this is working for both dev builds and the full electron build, and switching to SVG mode works OK.

PR #1492, opened 2024-07-17
EDITOR: Tweak typescript language server configuration to match use.

tomprince:
- We allow importing files with `.ts`/`.tsx` extensions.
- We use an file-at-a-time transpiler, so we don't support features that require understanding the full type system.
- We use the classic `React.createElement` transform.


d0sboots:
Unless you think otherwise, I'm gonna ignore this in favor of 1493 that seems to supersede it.

tomprince:
> Unless you think otherwise, I'm gonna ignore this in favor of 1493 that seems to supersede it.

I have no strong feeling either way, but if that one stalls to long, I'd be inclined to land this part first, if there are no issues with this part..

d0sboots:
The main thing that is being discussed in that one is the error code(s), which is also present here.

PR #1493, opened 2024-07-17
EDITOR: Enable strict typechecking of typescript.

### LINKED ISSUE #1488: MISC: Issues with Typescript support
Since the introduction of `.jsx`, `.ts` and `.tsx` file in #1216, players can use typescript. This is honestly one of the greatest changes ever, thank you very much  @catloversg for the great work!

There are some small things I found when playing around. I hope it is clear what is working and not by my comments.
I have two files, both `.ts`: `test.ts` and `lib.ts` on the home server, no complicated folder structure nothing.
```ts
// lib.ts
// This function and its signature is exported correctly
export const sum = (a: number, b: number): number => a + b;

// Type extensions from test.ts are also included here even without importing
const received = globalThis.myValue
```

```ts
// test.ts
import { sum } from "lib"

enum MyTest {
  a = 1,
  b = 2,
  c = 3
}

declare global {
  namespace globalThis {
    var myValue: number;
  }
}

  // Basic types from the game work
  const myServer: Server = {
    hostname: "a",
    ip: "0.0.0.0",
    sshPortOpen: false,
    httpPortOpen: false,
    smtpPortOpen: false,
    sqlPortOpen: false,
    ftpPortOpen: false,
    hasAdminRights: false,
    cpuCores: 1,
    isConnectedTo: false,
    ramUsed: 0,
    maxRam: 1024,
    organizationName: "a",
    purchasedByPlayer: false
  };
  const a: CorpEmployeePosition = "Operations";

  try {
    // RAM allocation as before
    await ns["weaken"]("a");
    await ns.grow("a");
  } catch { }

  // Type extension not only works here, but gets carried over to lib.ts
  globalThis.myValue = 5;
  const value = globalThis.myValue;

  // Enums work
  ns.tprint(MyTest.a);

  // Type signatures are now properly ported over
  const c = sum(1, (() => 1 + 2)());

  // But x is not of type 'Task | null'
  const x = ns.singularity.getCurrentWork();
}

// Implicit any is allowed? This should be an error, not a warning?
function greet(name) {
  console.log("Hello " + name.toUpperCase());
}

// This shouldn't be allowed when using TS, I think
const a = 5 + "10";

// Why does this not throw errors?
let name: string;
name = null;
name = undefined;

interface User {
  name: string;
  age: number;
}
const alice: User = { name: "Alice", age: 99, lastname: "this does give an error! yay!" };
```

To recap, the problems I encountered are:
- API functions that might return null are not typed that way.
- Not typing parameters is allowed, it should only be typed as any explicitly though
- Adding a number and a string should result in a compile error, but it doesn't
- You can assign any value null or undefined and it doesn't give an error

I'm not 100% sure, but I think most if not all of these come from the fact that strict mode is disabled.

tomprince:
Fixes #1488.

This enables strict type checking for typescript (and optionally javascript via `// @ts-check`). It also fixes types that involve null or undefined in both javascript and typescript.

This is stacked on top of #1492 since they trivially conflict.

### DIFF
--- a/src/ScriptEditor/ScriptEditor.ts
+++ b/src/ScriptEditor/ScriptEditor.ts
@@ -70,19 +70,28 @@ export class ScriptEditor {
       languageDefaults.addExtraLib(source, "netscript.d.ts");
       languageDefaults.addExtraLib(reactTypes, "react.d.ts");
       languageDefaults.addExtraLib(reactDomTypes, "react-dom.d.ts");
+      languageDefaults.setCompilerOptions({
+        ...languageDefaults.getCompilerOptions(),
+        // We allow direct importing of `.ts`/`.tsx` files, so tell the typescript language server that.
+        allowImportingTsExtensions: true,
+        // We use file-at-a-time transpiler. See https://www.typescriptlang.org/tsconfig/#isolatedModules
+        isolatedModules: true,
+        // We use the classic (i.e. `React.createElement`:) react runtime.
+        jsx: monaco.languages.typescript.JsxEmit.React,
+        // We define `React` and `ReactDOM` as globals. Don't mark using them as errors.
+        allowUmdGlobalAccess: true,
+      });
+      languageDefaults.setDiagnosticsOptions({
+        ...languageDefaults.getDiagnosticsOptions(),
+        // Ignore these errors in the editor:
+        diagnosticCodesToIgnore: [
+          // We define `React` and `ReactDOM` as globals. Don't mark using them as errors.
+          // - 'React' refers to a UMD global, but the current file is a module. Consider adding an import instead.(2686)
+          2686,
+        ],
+      });
     }
-    monaco.languages.typescript.typescriptDefaults.setCompilerOptions({
-      ...monaco.languages.typescript.typescriptDefaults.getCompilerOptions(),
-      jsx: monaco.languages.typescript.JsxEmit.ReactJSX,
-      allowUmdGlobalAccess: true,
-    });
-    /**
-     * Ignore these errors in the editor:
-     * - Cannot find module ''. Did you mean to set the 'moduleResolution' option to 'nodenext', or to add aliases to the 'paths' option?(2792)
-     */
-    monaco.languages.typescript.typescriptDefaults.setDiagnosticsOptions({
-      diagnosticCodesToIgnore: [2792],
-    });
+
     monaco.languages.json.jsonDefaults.setModeConfiguration({
       ...monaco.languages.json.jsonDefaults.modeConfiguration,
       //completion should be disabled because the
--- a/src/ScriptEditor/ScriptEditor.ts
+++ b/src/ScriptEditor/ScriptEditor.ts
@@ -63,10 +63,10 @@ export class ScriptEditor {
 
     // Add ts definitions for API
     const source = netscriptDefinitions.replace(/export /g, "");
-    for (const languageDefaults of [
-      monaco.languages.typescript.javascriptDefaults,
-      monaco.languages.typescript.typescriptDefaults,
-    ]) {
+    for (const [language, languageDefaults] of [
+      ["javascript", monaco.languages.typescript.javascriptDefaults],
+      ["typescript", monaco.languages.typescript.typescriptDefaults],
+    ] as const) {
       languageDefaults.addExtraLib(source, "netscript.d.ts");
       languageDefaults.addExtraLib(reactTypes, "react.d.ts");
       languageDefaults.addExtraLib(reactDomTypes, "react-dom.d.ts");
@@ -80,12 +80,22 @@ export class ScriptEditor {
         jsx: monaco.languages.typescript.JsxEmit.React,
         // We define `React` and `ReactDOM` as globals. Don't mark using them as errors.
         allowUmdGlobalAccess: true,
+        // Enable strict typechecking.
+        // Note that checking in javascript is disabled by default but can be enabled via `// @ts-check`.
+        // This enables strictNullChecks, which impacts reported types, even in javascript.
+        strict: true,
+        noImplicitAny: language === "typescript",
+        noImplicitReturns: true,
       });
       languageDefaults.setDiagnosticsOptions({
         ...languageDefaults.getDiagnosticsOptions(),
+        // Show semantic errors, even in javascript.
+        // Note that this will only happen if checking is enabled in javascript (e.g. by `// @ts-check`)
+        noSemanticValidation: false,
         // Ignore these errors in the editor:
         diagnosticCodesToIgnore: [
           // We define `React` and `ReactDOM` as globals. Don't mark using them as errors.
+          // Even though we set allowUmdGlobalAccess, it still shows a warning (instead of an error).
           // - 'React' refers to a UMD global, but the current file is a module. Consider adding an import instead.(2686)
           2686,
         ],


src/ScriptEditor/ScriptEditor.ts
```diff
@@ -63,26 +63,44 @@ export class ScriptEditor {
...
+        // Enable strict typechecking.
+        // Note that checking in javascript is disabled by default but can be enabled via `// @ts-check`.
+        // This enables strictNullChecks, which impacts reported types, even in javascript.
+        strict: true,
+        noImplicitAny: language == "typescript",
+        noImplicitReturns: true,
+      });
+      languageDefaults.setDiagnosticsOptions({
+        ...languageDefaults.getDiagnosticsOptions(),
+        // Show semanic errors, even in javascript.
```
d0sboots:
```suggestion
        // Show semantic errors, even in javascript.
```

src/ScriptEditor/ScriptEditor.ts
```diff
@@ -63,26 +63,44 @@ export class ScriptEditor {
...
+        // Note that this will only happen if checking is enabled in javascript (e.g. by `// @ts-check`)
+        noSemanticValidation: false,
+        // Ignore these errors in the editor:
+        diagnosticCodesToIgnore: [
+          // We define `React` and `ReactDOM` as globals. Don't mark using them as errors.
+          // - 'React' refers to a UMD global, but the current file is a module. Consider adding an import instead.(2686)
+          2686,
```
d0sboots, line 100: Why is this still needed, given that UMD access is on? Please add that to the comment.

src/ScriptEditor/ScriptEditor.ts
```diff
@@ -63,26 +63,44 @@ export class ScriptEditor {
...
     // Add ts definitions for API
     const source = netscriptDefinitions.replace(/export /g, "");
-    for (const languageDefaults of [
-      monaco.languages.typescript.javascriptDefaults,
-      monaco.languages.typescript.typescriptDefaults,
-    ]) {
+    for (const [language, languageDefaults] of [
+      ["javascript", monaco.languages.typescript.javascriptDefaults],
+      ["typescript", monaco.languages.typescript.typescriptDefaults],
+    ] as ["javascript" | "typescript", monaco.languages.typescript.LanguageServiceDefaults][]) {
```
d0sboots: Oof, this is a particular ugly cast. Why is it needed? Can you just declare `language` as a string instead?

src/ScriptEditor/ScriptEditor.ts
```diff
@@ -63,26 +63,44 @@ export class ScriptEditor {
...
+        isolatedModules: true,
+        // We use the classic (i.e. `React.createElement`:) react runtime.
+        jsx: monaco.languages.typescript.JsxEmit.React,
+        // We define `React` and `ReactDOM` as globals. Don't mark using them as errors.
+        allowUmdGlobalAccess: true,
+        // Enable strict typechecking.
+        // Note that checking in javascript is disabled by default but can be enabled via `// @ts-check`.
+        // This enables strictNullChecks, which impacts reported types, even in javascript.
+        strict: true,
+        noImplicitAny: language == "typescript",
```
d0sboots:
```suggestion
        noImplicitAny: language === "typescript",
```

d0sboots (commented):
Nice. This goes beyond what was in the bug, but they all look like good settings.

src/ScriptEditor/ScriptEditor.ts
tomprince: Without any cast, it deduces the type of the array as
```typescript
(string | monaco.languages.typescript.LanguageServiceDefaults)[]
```
instead of an array of tuples.

I definitely could just use `string` for `language`, but this will catch typos in tests against language.

tomprince:
> Nice. This goes beyond what was in the bug, but they all look like good settings.

I think the only thing in this PR (as opposed to #1492) is `noImplicitReturns` and explicit support of javascript.

catloversg:
These kind of changes should be tested with a test suite. This is my test suite:
`nano js.js jsWithTsCheck.js jsx.jsx ts.ts tsx.tsx libJsx.jsx libTsx.tsx sum.js multiply.ts`

js.js
```js
import { sum } from "/sum";
import { multiply } from "/multiply";

/** @param {NS} ns */
  ns.tprint(`sum: ${sum(1, 1)}`);
  ns.tprint(`multiply: ${multiply(2, 3)}`);
}
```
jsWithTsCheck.js
```js
// @ts-check
import { sum } from "/sum";
import { multiply } from "/multiply";

/**
 * Test implicit any
 */
function test(s) {
  return s.toString();
}

/** @param {NS} ns */
  ns.tprint(`sum: ${sum(1, 1)}`);
  ns.tprint(`multiply: ${multiply(2, 3)}`);
  let x = undefined;
  x.toString(); // Must show error
}
```
jsx.jsx
```js
import { LibText } from "/libJsx";
import { sum } from "/sum";
import { multiply } from "/multiply";

/** @param {NS} ns */
  ns.tprintRaw(LibText);
  ns.tprintRaw(<>From Jsx</>);
  ns.tprint(`sum: ${sum(1, 1)}`);
  ns.tprint(`multiply: ${multiply(2, 3)}`);
}
```
ts.ts
```js
import { sum } from "/sum";
import { multiply } from "/multiply";

  ns.tprint(`sum: ${sum(1, 1)}`);
  ns.tprint(`multiply: ${multiply(2, 3)}`);
}
```
tsx.tsx
```js
import { LibText } from "/libTsx";
import { sum } from "/sum";
import { multiply } from "/multiply";

  ns.tprintRaw(LibText);
  ns.tprintRaw(<>From Tsx</>);
  ns.tprintRaw(React.createElement("input"));
  ns.tprint(`sum: ${sum(1, 1)}`);
  ns.tprint(`multiply: ${multiply(2, 3)}`);
}
```
libJsx.jsx
```js
export const LibText = <>LibJsx</>;
```
libTsx.tsx
```js
export const LibText = <>LibTsx</>;
```
sum.js
```js
export const sum = (a, b) => a + b;
```
multiply.ts
```js
```

Based on this test suite, we still need to disable 2792.

By the way, why do you disable the `noImplicitAny` check in JS files? By default, type-checking is disabled in JS files. If the player enables it via `// @ts-check`, it means they **really** want to use the type-checking. In that case, the `noImplicitAny` check is very useful. Wanting to use type-checking in JS files while still wanting to use implicit any is a niche use case.

tomprince:
> Based on this test suite, we still need to disable 2792.

Without actually testing this, I'm guessing this is just a symptom of #10, and would be fixed by #1477. If you change the imports to not having leading `/` then things should start working. I don't think it makes sense to disable the error, since it is reporting a problem that the editor is having with getting definitions from the file.

src/ScriptEditor/ScriptEditor.ts
d0sboots: What if you iterate over
```js
for (const {language, languageDefaults} of [
      {language: "javascript", languageDefaults: monaco.languages.typescript.javascriptDefaults},
      {language: "typescript", languageDefaults: monaco.languages.typescript.typescriptDefaults},
]) {
```
? Based on playground, that will require no extra type annotations, and is very clear.

tomprince: Actually, `as const` works, and gives a narrow bound to `language` (specifically, it gives it the type `"javascript" | "typescript"`).

catloversg:
> If you change the imports to not having leading / then things should start working.

We cannot tell our players that. The leading slash has always been a way to "describe" an absolute path, and that's reasonable. Until we fix the problem of importing in monaco, we must disable 2792.

tomprince:
> > If you change the imports to not having leading / then things should start working.
> 
> We cannot tell our players that. The leading slash has always been a way to "describe" an absolute path, and that's reasonable. Until we fix the problem of importing in monaco, we must disable 2792.

Well, I don't think we should silently hide that the editor can't resolve the import (and so get types from the file). I think disabling 2792 does make that silent, so I don't think that is a good option. I'm all for having it show up as something advisory, but I'm not sure how to do that.

d0sboots:
It might just mean that we need to focus on solving the underlying problem first. It wasn't that noticeable before, but obviously that has changed with TS in the mix now.

tomprince:
> It might just mean that we need to focus on solving the underlying problem first. It wasn't that noticeable before, but obviously that has changed with TS in the mix now.

I've been poking at #1477 improving the comments, and hopefully making it more resilient. After that, I'll probably start poking at something along the lines of #1138.

catloversg:
Disabling that error code is a way to avoid confusing our players while we try to fix the root cause. It does not mean we ignore the problem. Without it, players will see (noticeable) errors that are caused by us and have to use a (bad) workaround, which is to remove the leading slash. We cannot tell them: "We add this new cool feature, but our code is buggy. You need to use a workaround or ignore the visual errors."

Edit: The workaround does not work for all cases:

This is `ts.ts`. `sum` is a JS file.
ts

This is `jsWithTsCheck.js`. `multiply` is a TS file.
js

d0sboots:
Are you saying that monaco is processing the imports here, but still warning on them? You didn't show what the warning message is. I didn't realize bare imports (without the extension) even functioned, particularly in monaco.

catloversg:
Monaco will try to solve the imports. In some cases, it cannot and shows the warning: `Cannot find module '<import-path>'. Did you mean to set the 'moduleResolution' option to 'nodenext', or to add aliases to the 'paths' option?(2792)`.

d0sboots:
That definitely does not sound like an error that we should be hiding, to me.

To elaborate: If we hide that, I will the user get *any* feedback if they mess up and import something incorrectly?

catloversg:
No, they won't get any feedback from the editor. In that case, the only thing that shows the error is the RAM calculator (next to the Save button).

tomprince:
I think part of the issue is that monaco uses a different language server for javascript vs typescript, so only each can only see other files of the same kind. I suspect setting all the (non-ns1) files to typescript would fix some of the errors, but I'm note sure what other consequences of that would be.

I'm poking around a bit at the monaco internals, to see if there is a way to work around this.

src/ScriptEditor/ScriptEditor.ts
```diff
@@ -63,26 +63,45 @@ export class ScriptEditor {
...
+        isolatedModules: true,
+        // We use the classic (i.e. `React.createElement`:) react runtime.
+        jsx: monaco.languages.typescript.JsxEmit.React,
+        // We define `React` and `ReactDOM` as globals. Don't mark using them as errors.
+        allowUmdGlobalAccess: true,
+        // Enable strict typechecking.
+        // Note that checking in javascript is disabled by default but can be enabled via `// @ts-check`.
+        // This enables strictNullChecks, which impacts reported types, even in javascript.
+        strict: true,
+        noImplicitAny: language === "typescript",
```
d0sboots, line 87: I think this is the only unresolved point that catlover brought up. Why disable this specifically for JS? Especially since the for loop would become much simpler without it.

tomprince: I don't have an entirely coherent reason for this, but there are a number of things that lead to this choice (though I'm not sure I had considered all this explicitly when I made the choice):
- I was considering the possibility of having an option to turn `checkJS` on globally (likely in the editor options), and though that this would be a better default for that case.
- Given that javascript code is also (I think anyway), syntactically valid typescript code, and we support calling freely between javascript and typescript files, there is an easy way for people get the stricter typechecking, and having different settings for javascript and typescript allows people to choose which strictness they want.
- Partly I thought it would be nice for how I'd be inclined to migrate my code, once this gets to a release:
  - First turn on checking for types that it can deduce or are explicitly delcared, but letting code that doesn't have  types stay as-is.
  - Second, go through and add types for things that aren't already deduced.

d0sboots: I went through and re-read more carefully what `// @ts-check` does. Some notes:
* Unless I got overruled by Snarling, I would reject a global `checkJS` option. Anything that alters the behavior of *code* should be configured in code (or in a file-based config), so that it can be shared if need be. I've encountered too many issues in other games/environments where options and environment settings cause code that works for one person to not work for others; it's the devil, as far as I'm concerned.
* For migration, I think the best thing to do would be to work file-at-a-time and just rename your .js file to .ts, then fixup any type errors that occur. You can bail out to `any` if you need a quick escape hatch for a big section.

That said, `// @ts-check` is a sort of halfway point that will never do a great job of typechecking, because the JSDoc syntax it relies on just isn't expressive enough for that. So I'm OK with leaving the default as implicit-any.

tomprince:
> Unless I got overruled by Snarling, I would reject a global `checkJS` option. Anything that alters the behavior of _code_

`checkJS` doesn't alter the behaviour of code, it just alter the what gets squiggly red underlines in the editor.

d0sboots:
That's true, you can run code without ever touching the editor. Still, it's something that should be configured either per-project (which is a bit awkward, but we might introduce a tsconfig.json if it proves necessary) or per-file (which already exists), IMO.

PR #1496, opened 2024-07-17
SCRIPT: Add sourcemaps to transformed files.

tomprince:
This could perhaps land as-is, but I've barely tested it. Publishing early to avoid duplicating effort, if anybody else is looking at this.

### DIFF
--- a/src/NetscriptJSEvaluator.ts
+++ b/src/NetscriptJSEvaluator.ts
@@ -4,6 +4,7 @@
  */
 import * as walk from "acorn-walk";
 import { parse } from "acorn";
+import * as convertSourceMap from "convert-source-map";
 
 import { LoadedModule, type ScriptURL, type ScriptModule } from "./Script/LoadedModule";
 import type { Script } from "./Script/Script";
@@ -83,7 +84,7 @@ function generateLoadedModule(script: Script, scripts: Map<ScriptFilePath, Scrip
     return script.mod;
   }
 
-  let scriptCode;
+  let scriptCode, sourceMap;
   const fileType = getFileType(script.filename);
   switch (fileType) {
     case FileType.JS:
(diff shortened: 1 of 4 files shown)


d0sboots:
I've thought about this, but never done anything towards it. I don't think anyone else is looking at it.

src/NetscriptJSEvaluator.ts
```diff
@@ -160,7 +158,12 @@ function generateLoadedModule(script: Script, scripts: Map<ScriptFilePath, Scrip
     // servers; it will be listed under the first server it was compiled for.
```
d0sboots, line 158: Oh lol, actually this comment and code is what I was thinking of. I guess I *did* do something with this at some point.

src/NetscriptJSEvaluator.ts
```diff
@@ -160,7 +158,12 @@ function generateLoadedModule(script: Script, scripts: Map<ScriptFilePath, Scrip
...
     // We don't include this in the cache key, so that other instances of the
     // script dedupe properly.
-    const adjustedCode = newCode + `\n//# sourceURL=${script.server}/${script.filename}`;
+    const sourceURL = `bitburner:///${script.server}/${script.filename}`;
```
d0sboots, line 161: What's up with this fake `bitburner:` URI scheme?

src/utils/ScriptTransformer.ts
```diff
@@ -149,28 +147,43 @@ export function transformScript(filename: string, code: string, fileType: FileTy
...
-    return babel.transform(code, { filename: filename, presets: presets }).code;
-  }
-  let parserConfig: ParserConfig;
-  if (fileTypeFeature.isTypeScript) {
-    parserConfig = {
-      syntax: "typescript",
-    };
-    if (fileTypeFeature.isReact) {
-      parserConfig.tsx = true;
+    const result = babel.transform(script.code, {
```
d0sboots, line 150: We should probably go ahead and remove the forceBabelTransform branch at this point, instead of trying to maintain it after its useful life.

d0sboots (commented):
Overall the code and concept looks sound. Bringing in @catloversg since they were just deep in this space.

src/NetscriptJSEvaluator.ts
tomprince: I modelled it after the `webpack://`/`webpack-internal://` scheme. I also chose not to put the hostname of the script in host part of the URL I give here. Doing what I did here leads to `bitburner://` showing up as a new entry in the devtools, with directories for each server.

src/utils/ScriptTransformer.ts
tomprince: :shrug:

I had actually originally implemented this only for babel (getting rid of the swc side of the branch), but it was trivial enough to add swc support as well.

src/NetscriptJSEvaluator.ts
tomprince: I *think* the URL is only really relevant for in the developer tools, and perhaps in tracebacks, so we can change this pretty freely. I experimented with a number of things, and landed on this as a first pass.

PR #1502, opened 2024-07-19
DOCS: WIP: Translate in-game documentation to Chinese

loliconRoot:
Originally posted by @d0sboots in https://github.com/bitburner-official/bitburner-src/pull/1452#issuecomment-2206783646
> If you want to tackle a project that is still huge, but IMO doable, I suggest translating the game's documentation. That doesn't have any of the same technical challenges, and would be helpful for people who know enough English to use the UI but struggle with longer/more complicated explanations. We link to the docs in-game, and if they were fully translated I think we could figure out a way to make that link locale-aware.

I'm trying to translate the in-game documentation into Chinese.
This is a WIP asking if this is possible.
If this work is meaningful, then I will continue to translate other in-game documents.

PR #1505, opened 2024-07-20
DOC: Move all docs into en/ subdirectory

d0sboots:
PR #1502 is working on adding a Chinese translation to the docs. In general, I encouraged this (in #1452) as a path towards getting useful translated content in the game without requiring a massive refactor/rearchitecting of everything.

To support this, this takes the first step of moving our docs into an en/ subdirectory, so that other languages can live alongside. No effort is made at this time to support or select between alternate languages; this is a pure-rename refactor.

Testing: In-game docs still work (for dev server). Ctrl-click is temporarily broken, but will work once this PR is submitted (points at the new location).


d0sboots:
@Snarling for signoff on design/overall goal

catloversg:
This change looks good to me. After merging it, other contributors can start translating documentation.

d0sboots:
Reminder to myself to fix/redo this PR and submit it

d0sboots:
I have updated this PR, and am submitting it now, because it touches all documentation.

I changed it slightly - it does actually look at the user's locale now (is no longer hardcoded), but there is a fallback. I tested with "ja" as the locale and it seems to work fine. We can patch it after if there is an issue; I don't want to block the whole thing since the docs refactor is big.

PR #1509, opened 2024-07-21
BLADEBURNER: Allow upgrading skill level over max safe integer

catloversg:
The maximum level of a skill is limited to `Number.MAX_SAFE_INTEGER`. This is an unintentional change caused by #1154. This PR removes that limit.

### DIFF
--- a/src/Bladeburner/Bladeburner.ts
+++ b/src/Bladeburner/Bladeburner.ts
@@ -143,7 +143,7 @@ export class Bladeburner {
 
   /** Directly sets a skill level, with no validation */
   setSkillLevel(skillName: BladeSkillName, value: number) {
-    this.skills[skillName] = clampInteger(value, 0);
+    this.skills[skillName] = clampNumber(value, 0);
     this.updateSkillMultipliers();
   }
 
--- a/src/Bladeburner/Bladeburner.ts
+++ b/src/Bladeburner/Bladeburner.ts
@@ -143,17 +143,29 @@ export class Bladeburner {
 
   /** Directly sets a skill level, with no validation */
   setSkillLevel(skillName: BladeSkillName, value: number) {
-    this.skills[skillName] = clampNumber(value, 0);
+    this.skills[skillName] = clampInteger(value, 0, Number.MAX_VALUE);
     this.updateSkillMultipliers();
   }
 
   /** Attempts to perform a skill upgrade, gives a message on both success and failure */
   upgradeSkill(skillName: BladeSkillName, count = 1): Attempt<{ message: string }> {
-    const availability = Skills[skillName].canUpgrade(this, count);
-    if (!availability.available) return { message: `Cannot upgrade ${skillName}: ${availability.error}` };
+    const currentSkillLevel = this.skills[skillName] ?? 0;
+    const actualCount = currentSkillLevel + count - currentSkillLevel;
+    if (actualCount === 0) {
+      return {
+        message: `Cannot upgrade ${skillName}: Due to floating-point inaccuracy and the small value of specified "count", your skill cannot be upgraded.`,
+      };
+    }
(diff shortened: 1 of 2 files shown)


src/Bladeburner/Bladeburner.ts
```diff
@@ -143,7 +143,7 @@ export class Bladeburner {
 
   /** Directly sets a skill level, with no validation */
   setSkillLevel(skillName: BladeSkillName, value: number) {
-    this.skills[skillName] = clampInteger(value, 0);
+    this.skills[skillName] = clampNumber(value, 0);
```
d0sboots:
```suggestion
    this.skills[skillName] = clampInteger(value, 0, Number.MAX_VALUE);
```
It's still important that skills be an integer value.

d0sboots (commented):
If you're going to (re)extend skills past the safe integer range, then upgradeSkill should be fixed. Otherwise, you can take advantage of rounding to increase skills by more than what is requested. In the limit, you only need to pay for half the levels. 

To fix this, you need to compute `const actualCount = (this.skills[skillName] + count) - this.skills[skillName];` and use that consistently. You also should be prepared for the possibility that actualCount is 0.

tomprince:
> You also should be prepared for the possibility that actualCount is 0.

It was suggested in discord that things should fail in this case. I think this should at least not cost the player anything, though that may lead to some player code going into an infinite loop.

catloversg:
> If you're going to (re)extend skills past the safe integer range, then upgradeSkill should be fixed. Otherwise, you can take advantage of rounding to increase skills by more than what is requested. In the limit, you only need to pay for half the levels.
> 
> To fix this, you need to compute `const actualCount = (this.skills[skillName] + count) - this.skills[skillName];` and use that consistently. You also should be prepared for the possibility that actualCount is 0.

What should we do when `actualCount` is 0 or different from `count`?
- Throw an error.
- Return false and print a log.

d0sboots:
Based on discord consensus, it seems like:
* `actualCount === 0` should return false (and have a useful error message, even though that will be ignored by netscript. But useful for the BB console, I guess.
* `actualCount !== count` is fine for other cases.
* `skill.calculateCost` should use the same logic for determining `actualCount` as `upgradeSkill`.

The last one has some interesting implications, although it seems important to me. For instance, the UI will show 0 cost at some point (the point at which manually upgrading becomes impossible).

An alternative would be to special-case the 0 case to round up, so that as long as you have specified a non-zero count you will always attempt to upgrade at least a little.

catloversg:
Test case:
- Skill: Hyperdrive.
- Current level: 1e17.
- Cost of "+9" (before the change): 2250000000000000000.00
- Cost of "+16" (before the change): 4000000000000000000.00
- At 1e17:
  - Count 0->8: error.
  - Count 9->23: as if count = 16.

Test code:
```js
/** @param {NS} ns */
  ns.clearLog();
  Player.bladeburner.skillPoints = 1e20;
  Player.bladeburner.skills.Hyperdrive = 1e17;
  const currentSP = ns.bladeburner.getSkillPoints();
  ns.bladeburner.upgradeSkill("Hyperdrive", 9);
  ns.print("Current level: ", ns.bladeburner.getSkillLevel("Hyperdrive").toPrecision(21));
  ns.print("Used: ", (currentSP - ns.bladeburner.getSkillPoints()).toPrecision(21));
}
```

Output:
```
bladeburner.upgradeSkill: Upgraded skill Hyperdrive by 16 levels
Current level: 100000000000000016.000
Used: 4000000000000000000.00
```

> The last one has some interesting implications, although it seems important to me. For instance, the UI will show 0 cost at some point (the point at which manually upgrading becomes impossible).

When the current level and "count" are at the point that floating-point inaccuracy becomes a problem, the UI (the "+" button, the "Skill Points required" info, etc.) is useless anyway, so I don't think we should care about that.

PR #1510, opened 2024-07-21
BUGFIX: Correctly show remaining grafting/programming time left when cycles are skipped.

tomprince:
I tried to test this with grafting, but not creating programs.

It seemed to work, but getting the bug to trigger consistently was tricky in my dev environment.

The issue this is fixing is the time display being wildly low if some cycles are skipped.


d0sboots:
This wouldn't work for mechanics with bonus time, but this is not a mechanic that has that.

PR #1511, opened 2024-07-22
MISC: Add some more types for developement.

tomprince:
- Show values of constants in type hint by making the types literal types.
- Add some more types to `webpack.config.js`.
  -  Also fix some errors detected when turning on checking for it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -113,7 +113,7 @@
     "format": "prettier -c --write .",
     "format:report": "prettier -c .",
     "start": "electron .app/index.html",
-    "start:dev": "webpack serve --progress --env devServer --mode development",
+    "start:dev": "webpack serve --progress --mode development",
     "build": "bash ./tools/build.sh production",
     "build:dev": "bash ./tools/build.sh development",
     "lint": "eslint --fix --ext js,jsx,ts,tsx --max-warnings 0 src",
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -7,15 +7,16 @@ const ForkTsCheckerWebpackPlugin = require("fork-ts-checker-webpack-plugin");
 const ReactRefreshWebpackPlugin = require("@pmmmwh/react-refresh-webpack-plugin");
 const CopyPlugin = require("copy-webpack-plugin");
 
-/** @type import("webpack-cli").CallableOption} */
+/** @type {import("webpack-cli").CallableOption} */
 module.exports = (env, argv) => {
-  const isDevServer = (env || {}).devServer === true;
+  const isDevServer = (env || {}).WEBPACK_SERVE === true;
   const runInContainer = (env || {}).runInContainer === true;
   const isDevelopment = argv.mode === "development";
   const enableReactRefresh = (env || {}).enableReactRefresh === true;
   const outputDirectory = "dist";
   const entry = "./src/index.tsx";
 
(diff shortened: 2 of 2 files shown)


webpack.config.js
```diff
@@ -192,6 +191,10 @@ module.exports = (env, argv) => {
       },
     },
     devServer: devServerSettings,
+    watchOptions: {
+      // When running in a container, we can't necesarily watch filesystem events.
+      poll: runInContainer ? true : undefined,
+    },
```
tomprince, line 197: I haven't tested this change, but tried to find the right option by looking at documentation.

package.json
```diff
@@ -113,7 +113,7 @@
     "format": "prettier -c --write .",
     "format:report": "prettier -c .",
     "start": "electron .app/index.html",
-    "start:dev": "webpack serve --progress --env devServer --mode development",
+    "start:dev": "webpack serve --progress --mode development",
```
d0sboots, line 116: What's up with this?

webpack.config.js
d0sboots: I am (understandably) leery of changes that haven't been tested. Especially config changes XD

d0sboots (commented):
This feels like two separate PRs mashed together: One that adds const to types, and one that changes stuff in webpack.config.js.

package.json
tomprince: Webpack provides a native way to detect whether we are running the dev server, so we don't also need to specify it manually.

```diff
-  const isDevServer = (env || {}).devServer === true;
+  const isDevServer = (env || {}).WEBPACK_SERVE === true;
```
See [here](https://webpack.js.org/api/cli/#cli-environment-variables) for details.

webpack.config.js
tomprince: Well, what is untested in an undocumented development option that appears to have bit-rotted. That it the kind of thing where support is on a best effort basis, and if it breaks for someone, the answer is for them to send a patch to fix it.

Note that only the `runInContainer == true` case in untested, since I'm not exactly sure how to run that, and I don't particularly want to set it up if I did.

tomprince:
> This feels like two separate PRs mashed together: One that adds const to types, and one that changes stuff in webpack.config.js.

They are both adding better types for things that are only seen by developers. But, they are separate commits, so it'd be easy enough to split into two PRs. The changes in the webpack config were addressing errors from the better types there.

d0sboots:
> They are both adding better types for things that are only seen by developers. But, they are separate commits, so it'd be easy enough to split into two PRs. The changes in the webpack config were addressing errors from the better types there.

How does adding `const` to types expose errors in the webpack config?!

Although I'd slightly prefer it, you don't have to split if you don't want to. The const part I can merge trivially, the webpack changes I will have to play around with to verify.

tomprince:
> How does adding `const` to types expose errors in the webpack config?!

It doesn't. I also added some more `@type`s to `webpack.config.js`, which revealed the errors. I originally added (some of) them when figuring out #1501, and thought they'd be useful for others.

> Although I'd slightly prefer it, you don't have to split if you don't want to. The const part I can merge trivially, the webpack changes I will have to play around with to verify.

I split the const part into #1515, and left the webpack changes here.

d0sboots:
still works with my nonstandard setup, so merge = yes

PR #1512, opened 2024-07-22
UI: Sync UI updates to game updates.

tomprince:
There are a bunch of React components that update at the same rate that the game engine processes cycles. Rather than have each place that does so start its own timer to update that often, add a new react hook that triggers an update shortly after the engine completes a cycle.

---

When trying to test #1510, I noticed that there were a bunch of things that appeared to be trying to update themselves every game cycle. 

I've tested a few of the screen this affects, but not all of them.

### DIFF
--- a/src/utils/EventEmitter.ts
+++ b/src/utils/EventEmitter.ts
@@ -17,4 +17,8 @@ export class EventEmitter<T extends any[]> {
       sub(...args);
     }
   }
+
+  hasSubscibers(): boolean {
+    return this.subscribers.size > 0;
+  }
 }
--- a/src/utils/EventEmitter.ts
+++ b/src/utils/EventEmitter.ts
@@ -1,30 +1,19 @@
-function uuidv4(): string {
-  return "xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx".replace(/[xy]/g, function (c) {
-    const r = (Math.random() * 16) | 0,
-      v = c == "x" ? r : (r & 0x3) | 0x8;
-    return v.toString(16);
-  });
-}
-
 /** Generic Event Emitter class following a subscribe/publish paradigm. */
 export class EventEmitter<T extends any[]> {
-  subscribers: Record<string, (...args: [...T]) => void | undefined> = {};
+  private subscribers: Set<(...args: [...T]) => void | undefined> = new Set();
+
+  constructor() {}
 
   subscribe(s: (...args: [...T]) => void): () => void {
-    let uuid = uuidv4();
-    while (this.subscribers[uuid] !== undefined) uuid = uuidv4();
-    this.subscribers[uuid] = s;
+    this.subscribers.add(s);
 
     return () => {
-      delete this.subscribers[uuid];
+      this.subscribers.delete(s);
     };
   }
 
   emit(...args: [...T]): void {
-    for (const s in this.subscribers) {
-      const sub = this.subscribers[s];
-      if (sub === undefined) continue;
-
+    for (const sub of this.subscribers) {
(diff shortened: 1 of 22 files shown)


src/utils/EventEmitter.ts
```diff
@@ -8,24 +8,30 @@ function uuidv4(): string {
...
-      const sub = this.subscribers[s];
-      if (sub === undefined) continue;
-
+    this.subscribers.forEach((sub) => {
       sub(...args);
+    });
+  }
+
+  /** Emits the event soon (i.e. sometime shortly after control returns to the browser).  */
+  emitSoon(...args: [...T]): void {
```
d0sboots: Is this necessary? React already defers updates in order to batch them instead of performing them synchronously.

src/utils/EventEmitter.ts
```diff
@@ -8,24 +8,30 @@ function uuidv4(): string {
...
+      this.subscribers.delete(uuid);
     };
   }
 
   emit(...args: [...T]): void {
-    for (const s in this.subscribers) {
-      const sub = this.subscribers[s];
-      if (sub === undefined) continue;
-
+    this.subscribers.forEach((sub) => {
```
d0sboots: This is a case where I'll insist on the imperative for-of style for efficiency, because it's in a base library.

src/utils/EventEmitter.ts
```diff
@@ -8,24 +8,30 @@ function uuidv4(): string {
...
-  subscribers: Record<string, (...args: [...T]) => void | undefined> = {};
+  private subscribers: Map<string, (...args: [...T]) => void | undefined> = new Map();
+
+  constructor() {}
 
   subscribe(s: (...args: [...T]) => void): () => void {
     let uuid = uuidv4();
-    while (this.subscribers[uuid] !== undefined) uuid = uuidv4();
-    this.subscribers[uuid] = s;
+    while (this.subscribers.get(uuid) !== undefined) uuid = uuidv4();
```
d0sboots: If you're switching to a Map, you can just switch to a Set and use the fact that all objects have unique identity.

src/utils/EventEmitter.ts
```diff
@@ -8,24 +8,30 @@ function uuidv4(): string {
 
 /** Generic Event Emitter class following a subscribe/publish paradigm. */
 export class EventEmitter<T extends any[]> {
-  subscribers: Record<string, (...args: [...T]) => void | undefined> = {};
+  private subscribers: Map<string, (...args: [...T]) => void | undefined> = new Map();
```
d0sboots: Optional: I don't think we use typescript's access-control declarations (private, protected) elsewhere in the codebase. At minimum, they're very rare. I can certainly see the value if they were used consistently, but if they're only used in a couple of places they could get confused with JS's private variables. 
Anyway, up to you.

src/utils/EventEmitter.ts
tomprince: It looks like this is *not* true in React 17. We can call `ReactDOM.unstable_batchedUpdates` around updates to opt into that behaviour though.

d0sboots: OK, I went and read up and learned a bunch more about how React batches updates. Indeed, the batching behavior in 17 (and 16 before it) was more complicated; the official docs glossed over it by simply saying "it batches updates in various circumstances" without going into detail on the circumstances.

But yes, the point of all of this (I assume) is batching the updates; otherwise, there's no reason not to let them continue to run on their own timers. In which case, there's no reason to throw on an extra delay.

d0sboots:
sigh, force-pushing again. You're breaking my heart XD

PR #1513, opened 2024-07-23
MISC: Remove unsed attribues of internal `ScriptDeath`.

tomprince:
As far as I can tell, these attributes are entirely unused, and `ScriptDeath` is not part of the public interface.


d0sboots:
This is the sort of change where I'm very glad we have TypeScript.

PR #1518, opened 2024-07-25
EDITOR: Support cross-language imports in editor.

tomprince:
@catloversg This should fix the issue you raised in #1493.

### DIFF
--- a/src/ScriptEditor/ScriptEditor.ts
+++ b/src/ScriptEditor/ScriptEditor.ts
@@ -13,6 +13,7 @@ import { Settings } from "../Settings/Settings";
 import { NetscriptExtra } from "../NetscriptFunctions/Extra";
 import * as enums from "../Enums";
 import { ns } from "../NetscriptFunctions";
+import { isLegacyScript } from "../Paths/ScriptFilePath";
 
 /** Event emitter used for tracking when changes have been made to a content file. */
 export const fileEditEvents = new EventEmitter<[hostname: string, filename: ContentFilePath]>();
@@ -63,9 +64,9 @@ export class ScriptEditor {
 
     // Add ts definitions for API
     const source = netscriptDefinitions.replace(/export /g, "");
-    for (const [language, languageDefaults] of [
-      ["javascript", monaco.languages.typescript.javascriptDefaults],
-      ["typescript", monaco.languages.typescript.typescriptDefaults],
+    for (const [language, languageDefaults, getLanguageWorker] of [
+      ["javascript", monaco.languages.typescript.javascriptDefaults, monaco.languages.typescript.getJavaScriptWorker],
+      ["typescript", monaco.languages.typescript.typescriptDefaults, monaco.languages.typescript.getTypeScriptWorker],
     ] as const) {
       languageDefaults.addExtraLib(source, "netscript.d.ts");
       languageDefaults.addExtraLib(reactTypes, "react.d.ts");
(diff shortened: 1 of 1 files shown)


src/ScriptEditor/ScriptEditor.ts
```diff
@@ -100,6 +103,25 @@ export class ScriptEditor {
           2686,
         ],
       });
+      // Sync all javascript and typescript text models to both language servers.
+      {
```
d0sboots: What's up with this extra block just hanging out here? Do we need this scope for some reason?

src/ScriptEditor/ScriptEditor.ts
```diff
@@ -100,6 +103,25 @@ export class ScriptEditor {
...
+        // `monaco.languages.typescript.get{Java,Type}ScriptWorker` returns a promise that
+        // fires with a function that takes a list of `monaco.Uri`s and sync's them with the
+        // worker. (It also returns the worker, but we don't care about that.)
+        const languageWorker = new Promise<(...uris: monaco.Uri[]) => unknown>((resolve) =>
+          monaco.languages.onLanguage(language, () => getLanguageWorker().then(resolve)),
+        );
+        // Whenever a model is created, arange for it synced to the language server.
+        monaco.editor.onDidCreateModel((model) => {
+          if (language === "typescript" && isLegacyScript(model.uri.path)) {
```
d0sboots: I think we don't want *either* type synced? IIRC, .script files can only import other .script files.

src/ScriptEditor/ScriptEditor.ts
tomprince: It is to scope the comment.

src/ScriptEditor/ScriptEditor.ts
tomprince: Ideally we'd have a custom language for script files, but we tell monaco that they are `javascript`, so it will get synced to the javascript worker at some point eventually anyway.

We could probably get away with just syncing to the opposite language server (so when `language != model.getLanguageId()`), but it doesn't hurt to tell them to sync an extra time, and I haven't dug into the timing of when they get synced, and I think it is reasonable to eargerly sync them if they aren't already.

src/ScriptEditor/ScriptEditor.ts
d0sboots: I think this's the first time I've ever seen that. 

It's pretty clear what the scope of the comment is without being that explicit; more importantly, there is no harm done if someone misinterprets the scope slightly.
You can use an extra newline to indicate "this is not directly part of the next comment," and then you don't need this while extra level of indentation.

src/ScriptEditor/ScriptEditor.ts
```diff
@@ -100,6 +103,25 @@ export class ScriptEditor {
           2686,
         ],
       });
+      // Sync all javascript and typescript text models to both language servers.
+      {
+        // `monaco.languages.typescript.get{Java,Type}ScriptWorker` returns a promise that
+        // fires with a function that takes a list of `monaco.Uri`s and sync's them with the
+        // worker. (It also returns the worker, but we don't care about that.)
+        const languageWorker = new Promise<(...uris: monaco.Uri[]) => unknown>((resolve) =>
+          monaco.languages.onLanguage(language, () => getLanguageWorker().then(resolve)),
```
d0sboots: Why do you need to jump through these extra callback hoops with onLanguage? Ultimately, `languageWorker` is just `getLanguageWorker()` with extra steps.

tomprince: Because `getLanguageWorker()` errors if the language has not yet been loaded, unfortunately.  I think I had meant to include that in the comment, but it appears that I missed adding that bit.

d0sboots: man wtf. It's a flipping promise, it can wait to resolve XD

PR #1519, opened 2024-07-26
API: Hide spoiler properties in getPlayer

yichizhng:
Per description. The properties are still available to avoid API breaks.

The documentation on getPlayer also needs work - several of the properties are not listed there, and the existence of entropy is spoiled.

### DIFF
--- a/src/NetscriptFunctions.ts
+++ b/src/NetscriptFunctions.ts
@@ -111,7 +111,7 @@ import { ServerConstants } from "./Server/data/Constants";
 import { assertFunction } from "./Netscript/TypeAssertion";
 import { Router } from "./ui/GameRoot";
 import { Page } from "./ui/Router";
-import { canAccessBitNodeFeature, validBitNodes } from "./BitNode/BitNodeUtils";
+import { canAccessBitNodeFeature, validBitNodes, knowAboutBitverse } from "./BitNode/BitNodeUtils";
 
 export const enums: NSEnums = {
   CityName,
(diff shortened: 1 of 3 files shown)


d0sboots:
I don't think I agree with the premise of this change. I fundamentally don't think it is the job of the API to "hide" parts of itself until you have access to them. Rather, it is the job of the API to *restrict* parts of itself until you have access to them. 
Things like the various SourceFile mechanics are very out there in the open if you choose to read the public docs, but calling them returns an opaque "you can't use this yet, and it will be obvious when you are able to."

I think this strikes the right balance, because although we can do stuff like scramble text in the game, stuff like the API documentation exists outside then game and thus outside of any context of your progress. Therefore, it *must* document everything, although it can still be done in a spoiler-sensitive way. (For instance, knowing that the Player has a field called "entropy" tells you nothing without further context, and we shouldn't explain that context in the docs for Player. Either they know what entropy is, or they don't.)

yichizhng:
Why are they hidden in the UI then? It really defeats the purpose of making these things spoilers if the simplest thing you can think of to do on getPlayer() will reveal them.

d0sboots:
> Why are they hidden in the UI then? It really defeats the purpose of making these things spoilers if the simplest thing you can think of to do on getPlayer() will reveal them.

Two reasons:
1. Because they can be. There's no reason to go out of our way to spoil things needlessly.
2. The places in the UI where things are revealed have all the context about them. That makes them significantly greater spoilers than having a field named "bandwidth" (I'm intentionally using a non-real field as an example here) where you know what the word means generally, but not what it means in the game. The most it reveals is that *some* other mechanic exists that you don't know about yet.

In general, "spoiler" is not a binary thing. It's a continuum. On one end, there is having *zero* knowledge about anything you haven't directly experienced yet. I don't think this is a practical level to try and support. On the other end, is telling people about BNs and advanced strategies like int grinding and BN12 looping while they're still in early game. We intentionally segregate channels in discord to prevent that level of spoilers.

What I've seen from most people is that they want to avoid "spoilers" in the form of other people's code, which is easily accomplished. Also, to avoid the "big reveals" such as the existence of BNs, which is also easily (and currently) accomplished. What seems less practical is to try and hide that future content/mechanics *exists at all*. Just knowing that it exists, but not what it is, or when it will be unlocked, seems like a reasonable line to draw, and is *much* easier to accomplish.

Edit: I will note that there are other games (in the narrative/exploration genres, for instance) where just *knowing that future content exists* can still be a major spoiler, even if you don't know what it is. When playing Fez, or Outer Wilds, or Tunic, or games of that nature I *would* want a true zero-knowledge experience. But Bitburner is not that type of game.

PR #1521, opened 2024-07-26
BUGFIX: Stop current work when starting a program with Singulatiry

TheAimMan:
Current implementation of the CreateProgram singularity API doesn't stop the current work before calling the CreateProgramWork function.  The issue with this is that that function checks to see if you're resuming an incomplete program.  Since the work done on an incomplete program isn't written to the users home directory until the work is stopped, this check fails, ending up with 2 incomplete programs

image

My fix is, once singularity verifies you can actually create a program, it checks to see if you're busy and if so, stops it, then creates the program.  

I could reproduce this before by just calling the same singulatiry API twice, but can't after my fix.

### DIFF
--- a/src/NetscriptFunctions/Singularity.ts
+++ b/src/NetscriptFunctions/Singularity.ts
@@ -1026,6 +1026,9 @@ export function NetscriptSingularity(): InternalAPI<ISingularity> {
           helpers.log(ctx, () => `Hacking level is too low to create '${p.name}' (level ${create.level} req)`);
           return false;
         }
+        if (Player.currentWork) {
+          Player.finishWork(true);
+        }
 
         Player.startWork(
           new CreateProgramWork({


d0sboots:
What was happening of you were doing a job or some other kind of work? Was it secretly still going?

TheAimMan:
I don't believe so.  It just appears to be a potential race condition between when the original createProgram work stops and the next one starts, the program just hasn't been written to home.

I tested again just to make sure I wasn't insane because I feel like it *should* be stopping, as `Player.startWork` appears to be checking to see if the player is busy or not, and if so to stop it.

Doing these steps
1. Manually create a program and don't stop.  Let it get to 1%
2. Run the createProgram singularity function to create the program

With fix, the focus go back to creating the program at the same percent.  1 file is on home when you stop work
Without fix, focus starts at 0.  Stopping work and `ls` home shows to programs in progress with different values.

Not sure I want to dig much more into it as the fix seemed pretty simple and appears to work.

d0sboots:
Ah, I see the issue now and it is specific to the singularity code.
The  `CreateProgramWork` class picks up incomplete progress and deletes the file in *the constructor*. But the incomplete progress is only written when the work is stopped. Thus, it is important that the worked be stopped *before* the new work is constructed, even though `Player.startWork` makes sure to stop any pending work.

TheAimMan:
I'm glad you get it but I don't :)
But my assumption was correct so yay!

PR #1523, opened 2024-07-27
BUGFIX: Cap Gang recruit member calculation

TheAimMan:
The current calculation for recruit-able gang members is
`Math.floor(Math.log(this.respect) / Math.log(recruitCostBase)) + numFreeMembers - this.members.length; `
The issue with this is that `Math.log(this.respect) / Math.log(recruitCostBase)` has no ceiling.  It's possible for this number to actually be greater than the current possible 12 gang members after the complete calculation.  This causes an issue where the UI claims there are more gang members available than there are possible to recruit

image

I just added a check to see which one was smaller, the already existing calculation, or the actual limit of gang members.

### DIFF
--- a/src/Gang/Gang.ts
+++ b/src/Gang/Gang.ts
@@ -320,19 +320,10 @@ export class Gang {
     if (this.members.length >= GangConstants.MaximumGangMembers) {
       return 0;
     }
-    const numFreeMembers = 3;
-    const recruitCostBase = 5;
-    if (this.members.length < numFreeMembers && this.respect < Math.pow(recruitCostBase, numFreeMembers)) {
-      return numFreeMembers - this.members.length; // if the max possible is less than freeMembers
-    }
-    return (
-      Math.min(
-        Math.floor(Math.log(this.respect) / Math.log(recruitCostBase)),
-        GangConstants.MaximumGangMembers - numFreeMembers,
-      ) +
-      numFreeMembers -
-      this.members.length
-    ); //else
+    const numFreeMembers = GangConstants.numFreeMembers;
+    const recruitCostBase = GangConstants.recruitThresholdBase;
+    const membersRecruitabile = Math.floor(Math.log(this.respect) / Math.log(recruitCostBase)) + numFreeMembers;
+    return Math.min(membersRecruitabile, GangConstants.MaximumGangMembers) - this.members.length;
   }
 
   recruitMember(name: string): boolean {
--- a/src/Gang/Gang.ts
+++ b/src/Gang/Gang.ts
@@ -325,7 +325,14 @@ export class Gang {
     if (this.members.length < numFreeMembers && this.respect < Math.pow(recruitCostBase, numFreeMembers)) {
       return numFreeMembers - this.members.length; // if the max possible is less than freeMembers
     }
-    return Math.floor(Math.log(this.respect) / Math.log(recruitCostBase)) + numFreeMembers - this.members.length; //else
+    return (
+      Math.min(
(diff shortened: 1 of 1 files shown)


src/Gang/Gang.ts
```diff
@@ -325,7 +325,14 @@ export class Gang {
...
     }
-    return Math.floor(Math.log(this.respect) / Math.log(recruitCostBase)) + numFreeMembers - this.members.length; //else
+    return (
+      Math.min(
+        Math.floor(Math.log(this.respect) / Math.log(recruitCostBase)),
+        GangConstants.MaximumGangMembers - numFreeMembers,
+      ) +
+      numFreeMembers -
+      this.members.length
+    ); //else
```
d0sboots: There's no `if` for this to be an "else" to. (I see this was pre-existing, but it should be killed because it's weird.)

src/Gang/Gang.ts
```diff
@@ -325,7 +325,14 @@ export class Gang {
     if (this.members.length < numFreeMembers && this.respect < Math.pow(recruitCostBase, numFreeMembers)) {
```
d0sboots: There's actually an issue here too: If respect is 10 (for instance), that is < 5**3, so this line will trigger. However, the correct result is `1 + numFreeMembers - this.members.length`, i.e. it should go to the next part.

The real condition here should be `< recruitCostBase`.

src/Gang/Gang.ts
```diff
@@ -325,7 +325,14 @@ export class Gang {
...
       return numFreeMembers - this.members.length; // if the max possible is less than freeMembers
     }
-    return Math.floor(Math.log(this.respect) / Math.log(recruitCostBase)) + numFreeMembers - this.members.length; //else
+    return (
+      Math.min(
+        Math.floor(Math.log(this.respect) / Math.log(recruitCostBase)),
```
d0sboots: A slight re-arrangement that both simplifies the expression and makes more intuitive sense (to me) is to bring `numFreeMembers` inside the `min` by adding it to both sides. This leaves an expression for total gang members on one branch, and `GangConstants.MaximumGangMembers` on the other.

Since that brings most of the stuff inside the min, I think you won't need the surrounding parens after.

src/Gang/Gang.ts
TheAimMan: Removed

src/Gang/Gang.ts
TheAimMan: Based on my changes to the next check, I just completely removed this statement.  I'm already account for the free members in the next check by default.  It also fixes the bug where the available count would be off if you didn't recruit the first three members before gaining enough respect for the next member.

src/Gang/Gang.ts
TheAimMan: Since we're going at it with a butchers knife, I changed some of the existing declared variables to equal their constants via the constant's file instead of the odd hardcoded values of 3 and 5.

I accounted for the free members in the 'membersRecruitable' calc so we could remove the previous check.

src/Gang/Gang.ts
d0sboots: There is one case that was covering that is still important, or you need to upgrade the `max` to a `clamp`: When respect is 0, the log becomes -Infinity, leading to -Infinity recruitable members currently.

TheAimMan: It looks like you start with 1 respect.  I'm not sure if it's even possible to lose respect as I thought when a gang member dies/ascends you just lose the respect they've gained.  Regardless, I believe I followed what you were suggesting and added the check to the appropriate place.

PR #1526, opened 2024-07-27
INFILTRATION: Minesweepergame minor bugfix, made rounding behavior for height, width and mine count consistent

### LINKED ISSUE #1520: [BUG?] Inconsistent rounding behavior in MinesweeperGame
I was looking through the code of the MinesweeperGame and I noticed that the intention in its setup seemed to be that the grid is rectangular, see the difficulty config below, in each case both height and width are consistent

https://github.com/bitburner-official/bitburner-src/blob/dd5fcc077ac6fddfc32ada8de4ce108ca23d444f/src/Infiltration/ui/MinesweeperGame.tsx#L29-L32

However when playing the game the grid was not square:

image

This made me wonder if there was an off by 1 error somewhere in the code, so I started looking for that. Seems the culprit is an inconsistent rounding issue instead in the following code:

https://github.com/bitburner-official/bitburner-src/blob/dd5fcc077ac6fddfc32ada8de4ce108ca23d444f/src/Infiltration/ui/MinesweeperGame.tsx#L167-L187

Turns out the difficulty's height and width values are interpolated in the same manner, however the floating point value resulting from this (e.g. `4.418230611480418` for both width and height) is handled differently between the setup of the columns and rows. In the line below you can see that for the number columns the interpolated difficulty's width is `Math.round`ed :

https://github.com/bitburner-official/bitburner-src/blob/dd5fcc077ac6fddfc32ada8de4ce108ca23d444f/src/Infiltration/ui/MinesweeperGame.tsx#L170

And in this line below you can see that for the number rows the interpolated difficulty's height is not `Math.round`ed, resulting in some cases in a height that is 1 higher than the column count (because of the way the for loops are set up, doing no rounding will result in behavior similar to a `Math.ceil`):

https://github.com/bitburner-official/bitburner-src/blob/dd5fcc077ac6fddfc32ada8de4ce108ca23d444f/src/Infiltration/ui/MinesweeperGame.tsx#L169

I don't know if this is intentional? If not the behavior should probably be made consistent (either round them both or not round them both). If it is the case that it should always favor the higher difficulty (as it is it doing when not rounding) it might be nice to `Math.ceil` the values explicitly (this will make the behavior more clear in the code)?

Also note that this behavior is inconsistent with the generation for the   grid-template-columns and  grid-template-rows css properties which are both rounded (although in this specific case and with the way css grid's work the grid-template-rows does not actually do anything and could be removed entirely):

https://github.com/bitburner-official/bitburner-src/blob/dd5fcc077ac6fddfc32ada8de4ce108ca23d444f/src/Infiltration/ui/MinesweeperGame.tsx#L113-L114

On a side note, the amount of mine's is also not rounded or ceiled, it might be also good to do that for consistency sake:

https://github.com/bitburner-official/bitburner-src/blob/dd5fcc077ac6fddfc32ada8de4ce108ca23d444f/src/Infiltration/ui/MinesweeperGame.tsx#L177

If a maintainer agrees with my findings I would be happy to raise a pull request which make's this behavior consistent?

mmjr-x:
A detailed description of the bug that this PR fixes can be found in #1520 .

# Linked issues

fixes #1520 

# Bug fix

Tested it with and without this change:

Without:
image

With:
image

Screenshot's confirm the fix, all data to reproduce should also be in the screenshots.

### DIFF
--- a/src/Infiltration/ui/MinesweeperGame.tsx
+++ b/src/Infiltration/ui/MinesweeperGame.tsx
@@ -166,7 +166,7 @@ function fieldEquals(a: boolean[][], b: boolean[][]): boolean {
 
 function generateEmptyField(difficulty: Difficulty): boolean[][] {
   const field = [];
-  for (let i = 0; i < difficulty.height; i++) {
+  for (let i = 0; i < Math.round(difficulty.height); i++) {
     field.push(new Array(Math.round(difficulty.width)).fill(false));
   }
   return field;
@@ -174,7 +174,7 @@ function generateEmptyField(difficulty: Difficulty): boolean[][] {
 
 function generateMinefield(difficulty: Difficulty): boolean[][] {
   const field = generateEmptyField(difficulty);
-  for (let i = 0; i < difficulty.mines; i++) {
+  for (let i = 0; i < Math.round(difficulty.mines); i++) {
     const x = Math.floor(Math.random() * field.length);
     const y = Math.floor(Math.random() * field[0].length);
     if (field[x][y]) {


d0sboots:
Let me know when you've tested it. (Dev has dev menu which should make it easier to get the exact stats needed to trigger the bug.)

mmjr-x:
@d0sboots I have tested and confirmed the fix, see the screenshots for more details :) let me know if any other changes are required?

PR #1530, opened 2024-07-31
UI: Fix misleading favor numbers

catloversg:
Favor is not an integer, but we "pretend" that it is by using `Math.floor` in many places in UI. The problem is that the usage of `Math.floor` is inconsistent, and UI shows "misleading" numbers in some cases.

How to reproduce it:
- Join a faction.
- Set favor to 149 and current reputation to 5000.
- Hover the mouse over `Reputation:`. It shows "You will have 150 faction favor ...".
- Soft reset. Check `Faction Favor:`. It shows "149".

Fix: remove the usage of `Math.floor` and change `formatFavor`.
Alternative fix in case we want to show Favor as an integer (I don't like this approach): use `Math.floor` consistently.

### DIFF
--- a/src/Faction/ui/AugmentationsPage.tsx
+++ b/src/Faction/ui/AugmentationsPage.tsx
@@ -207,7 +207,7 @@ export function AugmentationsPage({ faction }: { faction: Faction }): React.Reac
             <Typography>
               <b>Reputation:</b> <Reputation reputation={faction.playerReputation} />
               <br />
-              <b>Favor:</b> <Favor favor={Math.floor(faction.favor)} />
+              <b>Favor:</b> <Favor favor={faction.favor} />
             </Typography>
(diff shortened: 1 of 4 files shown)


d0sboots:
I agree with the direction of this change, but could you please add screenshots of the new favor formatting? I could go build the branch, but 
1. I'm lazy, and
2. It's a good habit

catloversg:
Screenshots:

Before:
before-1
before-2
before-3
before-4
before-5

After:
after-1
after-2
after-3
after-4
after-5

bupjae:
Just my 2 cent

If favor is no longer integer, we may want to remove "floor" symbol from MathJax.

catloversg:
The formula in the tooltip is outdated. I fixed it in #1554.

PR #1532, opened 2024-08-01
MISC: Refactor Person.ts and Sleeve.ts

catloversg:
This was a part of #1411. I held off according to the reviews at https://github.com/bitburner-official/bitburner-src/pull/1411#discussion_r1651970680 and https://github.com/bitburner-official/bitburner-src/pull/1411#discussion_r1651974645.


src/PersonObjects/Person.ts
```diff
@@ -40,18 +43,195 @@ export abstract class Person implements IPerson {
...
+    }
+    return false;
+  }
+
+  travel(cityName: CityName): boolean {
+    if (!Player.canAfford(CONSTANTS.TravelCost)) {
+      return false;
+    }
+
+    Player.loseMoney(CONSTANTS.TravelCost, this instanceof PlayerObject ? "other" : "sleeves");
```
d0sboots, line 229: Not for this PR (let's keep this one simple), but Person.ts has a dependency on PlayerObject that should be wholly unnecessary (and is a rather nasty dependency loop) because of this line. This is extra gross because what's being done here can be trivially done by adding a new method to Person itself to get the `moneySourceLabel()`.

PR #1535, opened 2024-08-02
MISC: Clarify deprecation warning of ns.getTimeSinceLastAug()

### LINKED ISSUE #1533: ns.getResetInfo().lastAugReset returns incorrect value
### Bug Description: ###

ns.getResetInfo().lastAugReset incorrectly returns absolute value instead of being direct replacement of the old deprecated function ns.getTimeSinceLastAug() as it should be (if the plan is to remove this function)

In the example script below i print the old deprecated value from ns.getTimeSinceLastAug() which returns correct value that is supposed to be the value for ns.getResetInfo().lastAugReset

### Script: ###

image

### Output: ###
image

### Time played screenshot ### 
This is only an approximate match for the script obviously as i can't really do 2 things at once - run script and do screenshot.

image

### Bitburner version: ### 
v2.6.2 633da3830

catloversg:
Closes #1533.

### DIFF
--- a/src/NetscriptFunctions.ts
+++ b/src/NetscriptFunctions.ts
@@ -1633,7 +1633,11 @@ export const ns: InternalAPI<NSFull> = {
     return convertTimeMsToTimeElapsedString(milliseconds, milliPrecision);
   },
   getTimeSinceLastAug: () => () => {
-    deprecationWarning("ns.getTimeSinceLastAug()", "Use ns.getResetInfo().lastAugReset instead.");
+    deprecationWarning(
+      "ns.getTimeSinceLastAug()",
+      "Use ns.getResetInfo().lastAugReset instead. Please note that ns.getResetInfo().lastAugReset does NOT return the " +
+        "same value as ns.getTimeSinceLastAug(). Check the NS API documentation for details.",
+    );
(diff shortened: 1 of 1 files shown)


src/NetscriptFunctions.ts
```diff
@@ -1633,7 +1633,11 @@ export const ns: InternalAPI<NSFull> = {
     return convertTimeMsToTimeElapsedString(milliseconds, milliPrecision);
   },
   getTimeSinceLastAug: () => () => {
-    deprecationWarning("ns.getTimeSinceLastAug()", "Use ns.getResetInfo().lastAugReset instead.");
+    deprecationWarning(
+      "ns.getTimeSinceLastAug()",
+      "Use ns.getResetInfo().lastAugReset instead. Please note that ns.getResetInfo().lastAugReset does NOT return the " +
+        "same value as ns.getTimeSinceLastAug(). Check the NS API documentation for details.",
```
```suggestion
      "Use  `Date.now() - ns.getResetInfo().lastAugReset` instead. Please note that ns.getResetInfo().lastAugReset does NOT return the " +
        "same value as ns.getTimeSinceLastAug(). Check the NS API documentation for details.",
```

PR #1536, opened 2024-08-02
BUGFIX: Duplicated augmentation when buying after grafting

catloversg:
This bug can be exploited to duplicate all graftable augmentations. How to reproduce it:
- Load a clean save with SF10.
- Graft QLink.
- Join Illuminati. Buy QLink.
- Wait until the grafting QLink finishes, or use the Dev tool to "time skip".
- Check the augmentation list.

Capture-1
Capture-2

### DIFF
--- a/src/PersonObjects/Grafting/ui/GraftingRoot.tsx
+++ b/src/PersonObjects/Grafting/ui/GraftingRoot.tsx
@@ -156,19 +156,17 @@ export const GraftingRoot = (): React.ReactElement => {
                   Router.toPage(Page.Work);
                 }}
                 confirmationText={
-                  <>
+                  <Typography component="div" paddingBottom="1rem">
                     Cancelling grafting will <b>not</b> save grafting progress, and the money you spend will <b>not</b>{" "}
                     be returned.
-                    <br />
-                    <br />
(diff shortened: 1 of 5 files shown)


d0sboots:
I could be convinced otherwise, but I think this fix is closing a potentially legitimate usecase.

Let's say someone is grafting Aug A. They might *also* queue aug A, being unsure at the present time if it will be better to stick with grafting or soft-reset to install augs. Queuing A does *not* make grafting A pointless, since the effect of A is not present yet, and it will be as soon as grafting is finished.

IMO the proper bugfix would be for (finished) grafting of an already-queued aug to remove it from the queue, in addition to gaining it as an installed aug. In other words, only install it once instead of twice.

src/PersonObjects/Grafting/ui/GraftingRoot.tsx
d0sboots, line 1: Revert this file, it's just got extra whitespace now unless I'm missing something.

catloversg: I used those `<br />` to create an empty space before the "Confirm" button.

Before:
before

After:
after

In the latest commit, I used "paddingBottom" instead of those  `<br />`.

PR #1537, opened 2024-08-03
BUGFIX: Tab completion uses wrong command list

catloversg:
When I checked `getTabCompletionPossibilities.ts`, I saw this comment: `TODO: this shouldn't be hardcoded in two places with no typechecks to verify equivalence`. `gameCommands` is not the same as `commands` in `executeCommand` (`Terminal.ts`). The discrepancy:
- Missing commands: buy, grep, history, unalias, wget
- Nonexistent commands: ifconfig, theme

This PR refactors the command list.


d0sboots:
Just the other day I was noticing `wget` tab completion doesn't work...

PR #1542, opened 2024-08-03
TERMINAL: Stop terminal scp from revealing and copying to unreachable servers

yichizhng:
Currently tab autocomplete on terminal `scp` spoils the existence of WD before installing TRP. This PR also makes terminal `scp` unable to copy to unreachable servers for consistency with `ns.scp`.


d0sboots:
We should probably use that new function in a bunch more places, but you don't need to commit to that refactor here.

PR #1558, opened 2024-08-08
IPVGO: add scoring rules explanations to how to play page and score modal

### LINKED ISSUE #1538: IPVGO: Concerns regarding in-game "How to Play" instruction
Bitburner v2.6.2 (633da3830)

1. This game uses Chinese scoring system (counts Empty Nodes + my Routers still on board), while [linked guide](https://way-to-go.gitlab.io/#/en/capture-stones) uses Japanese scoring system (counts Empty Nodes - my Routers removed from board)

2. This game does not remove "dead" Routers when game ends; the owner of the territory must place their Routers to capture manually before pressing Pass Turn or End Game. This fact should be mentioned in the insturction.

ficocelliguy:
closes #1538 

* Add detailed scoring rules to the "how to play" page to avoid confusion with the more complex modern Japanese Go scoring system
* Added a modal to show score rule details from the post-game score report, so it is easily accessible if a game's score does not turn out as expected

### DIFF
--- a/src/Go/boardState/goStyles.ts
+++ b/src/Go/boardState/goStyles.ts
@@ -585,6 +585,9 @@ export const boardStyles = makeStyles<void, Size | "background">({ uniqId: "boar
     scoreModal: {
       width: "400px",
     },
+    scoreExplanationModal: {
+      width: "80vw",
+    },
     centeredText: {
       textAlign: "center",
     },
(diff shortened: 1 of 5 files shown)


gmcew:
A lore based thought; in the 2070s (where the IPvGO intro starts us) "modern Japanese territory scoring" might have a different context, and "modern shortcuts to make physical games quicker" feels a strange concept when you've just installed a bunch of super-human body augments to boost your physical speed.

"Common 21st century" may fit the flavour better than "modern", but I accept may be confusing to a player who is just after scoring clarity instead of cyber-punky lore!

ficocelliguy:
@gmcew I tweaked some of the phrasing with that in mind, let me know if it fits better in-world now.

PR #1563, opened 2024-08-10
ELECTRON: Use steamworks.js to integrate with Steamworks SDK

catloversg:
As we discussed on Discord, this PR replaces greenworks with steamworks.js (https://github.com/ceifa/steamworks.js).

Currently, steamworks.js does not have 2 APIs that we need. I opened 2 PRs to add those APIs (https://github.com/ceifa/steamworks.js/pull/157 and https://github.com/ceifa/steamworks.js/pull/158). I don't know when they will be merged, so we will use my custom build in the meantime.

How to verify the binaries:
- Check my fork at https://github.com/catloversg/steamworks.js.
- Check the build logs and artifacts at https://github.com/catloversg/steamworks.js/actions/runs/10323373419.

Tested: Windows 10 and Debian 12.

Tested Steam's features:
- Cloud Sync.
- Achievements.


electron/achievements.js
```diff
@@ -1,52 +1,51 @@
...
+  }
 
   // This is backward but the game fills in an array called `document.achievements` and we retrieve it from
   // here. Hey if it works it works.
-  const steamAchievements = greenworks.getAchievementNames();
+  const steamAchievements = steamworksClient.apps.achievements();
   log.silly(`All Steam achievements ${JSON.stringify(steamAchievements)}`);
-  const playerAchieved = (await Promise.all(steamAchievements.map(checkSteamAchievement))).filter((name) => !!name);
+  const playerAchieved = steamAchievements.filter((achievement) => {
+    return steamworksClient.achievement.isActivated(achievement);
```
d0sboots: You don't need the braces and return here.

electron/achievements.js
```diff
@@ -1,52 +1,51 @@
...
+    return;
+  }
 
   // This is backward but the game fills in an array called `document.achievements` and we retrieve it from
   // here. Hey if it works it works.
-  const steamAchievements = greenworks.getAchievementNames();
+  const steamAchievements = steamworksClient.apps.achievements();
   log.silly(`All Steam achievements ${JSON.stringify(steamAchievements)}`);
-  const playerAchieved = (await Promise.all(steamAchievements.map(checkSteamAchievement))).filter((name) => !!name);
+  const playerAchieved = steamAchievements.filter((achievement) => {
```
d0sboots: This sound very similar to "playerAchievements", I'd call it steamAchievements instead. To make that work, I'd rename the current thing to allSteamAchievements.

electron/main.js
```diff
@@ -37,23 +37,9 @@ log.transports.console.level = store.get("console-log-level", "debug");
 
 log.info(`Started app: ${JSON.stringify(process.argv)}`);
 
-// We want to fail gracefully if we cannot connect to Steam
```
d0sboots, line 39: We should still be logging the status to the logfile.

electron/steamworks.js
```diff
@@ -0,0 +1,103 @@
+/**
```
d0sboots: This should be an npm module or similar, not something we copy directly into our source tree.

d0sboots:
There's some core architectural decisions made in steamworks.js that are baffling to me; I don't see how so much of the API can be promise-free when it *must* be making network requests. But, I'll take that devil in exchange for something that's maintained. 

I'm a bit concerned that the removal of "async" everywhere is premature.

LJNeon:
Both steamworks.js PRs have been merged.

d0sboots:
Yes, this would be a good thing to merge for 3.0 once we get the patch release out.

catloversg:
The concern about sync APIs is totally valid. My guess is that those APIs only get/set local files and let the Steam app sync those files later. For example, `steamworksClient.cloud.writeFile` *may* only write to a local file (via Steam SDK API) instead of calling Steam's HTTP-based APIs. That's only a guess, though. I'll try to take a look at it later.

catloversg:
In most cases, `steamworks-rs` (the underlying Rust library) calls Steam's sync APIs:
- steamworksClient.apps.achievements(): https://partner.steamgames.com/doc/api/ISteamUserStats#GetAchievementName
- steamworksClient.achievement.isActivated(): https://partner.steamgames.com/doc/api/ISteamUserStats#GetAchievement
- steamworksClient.achievement.activate(): https://partner.steamgames.com/doc/api/ISteamUserStats#SetAchievement
- steamworksClient.cloud.listFiles(): https://partner.steamgames.com/doc/api/ISteamRemoteStorage#GetFileCount and https://partner.steamgames.com/doc/api/ISteamRemoteStorage#GetFileNameAndSize
- steamworksClient.cloud.isEnabledForAccount(): https://partner.steamgames.com/doc/api/ISteamRemoteStorage#IsCloudEnabledForAccount
- steamworksClient.cloud.isEnabledForApp(): https://partner.steamgames.com/doc/api/ISteamRemoteStorage#IsCloudEnabledForApp
- steamworksClient.cloud.setEnabledForApp(): https://partner.steamgames.com/doc/api/ISteamRemoteStorage#SetCloudEnabledForApp
- steamworksClient.cloud.writeFile(): https://partner.steamgames.com/doc/api/ISteamRemoteStorage#FileWriteStreamOpen
- steamworksClient.cloud.deleteFile(): https://partner.steamgames.com/doc/api/ISteamRemoteStorage#FileDelete

The only async usage is this one: steamworksClient.cloud.readFile()
https://github.com/Noxime/steamworks-rs/blob/a0b0f97acc7d1b2a5420599947e062d52f7eacf9/src/remote_storage.rs#L220-L262
It calls an async Steam API, then polls for the result in a while loop. I'm not sure how problematic it is if we use this API in our JS side as a sync API (steamworksClient.cloud.readFile is sync in JS). I don't know much about Rust<->JS, so I'll let @d0sboots make the decision here.

d0sboots:
As LJ said, the PRs have been merged so we can use official upstream. Also, there are a bunch of pending comments. But, the timing of this looks good to do now. I think the sync issue will be fine.

catloversg:
Using upstream is blocked until https://github.com/ceifa/steamworks.js/pull/182 is merged.

I updated binaries at https://github.com/bitburner-official/bitburner-src/pull/1563/commits/5264d81848198718e8169976247ce4e0943441fb.
You can verify those files by checking https://github.com/catloversg/steamworks.js/actions/runs/14170389666.

d0sboots:
I'm going to keep sitting on this one for now, I'd really like for us to be using an npm package (probably pinned to a specific version) as opposed to including opaque binary blobs in our repo.
If it takes super long, I might reconsider.

catloversg:
The maintainer of `steamworks.js` is too busy now (https://github.com/ceifa/steamworks.js/issues/176), so we have to wait for a long time. I want to push this change a bit faster. In the latest commit, I switch to my published version (https://www.npmjs.com/package/@catloversg/steamworks.js). I pin the version at `0.0.1` and enable build provenance.

d0sboots (commented):
Well crap. I just realized these comments have been pending the whole time.

electron/steamworks.js
d0sboots: This was done.

electron/main.js
catloversg: The error is logged in: https://github.com/bitburner-official/bitburner-src/blob/68999521f52cf4404e18aff21f7c0c9ffc08d13c/electron/steamworksUtils.js#L6-L16

```
[2025-04-xx xx:38:09.397] [warn]  Cannot create IPC pipe to Steam client process.  Steam is probably not running.
[2025-04-xx xx:38:09.704] [info]  Started app: ["path\\.build\\bitburner-win32-x64\\bitburner.exe"]
[2025-04-xx xx:38:09.858] [info]  Application is ready!
[2025-04-xx xx:38:15.847] [error] Could not retrieve steam file
```

PR #1565, opened 2024-08-11
MISC: Increase minimum Node version

catloversg:
There is no need to support older Node versions.

Tested: Windows 10 - Node v20.11.1 and Node v18.20.4.


catloversg:
Our CI workflow is using Node.js 16.13.1. We should update them to v20: https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/.

@d0sboots If you agree, I'll make another PR to do that.

d0sboots:
Hehe, I think you'll have to upgrade the CI *first*, before we can merge this XD

catloversg:
@d0sboots #1578 was merged. We can merge this one now.

d0sboots:
Even though it doesn't show anything to merge, I think you need to `git pull`/merge the updated git workflows into your branch so that the tests will pass - they're still using the old versions.

Given the nature of this, I don't want to submit without everything being green.

PR #1566, opened 2024-08-11
DOCUMENTATION: Fix errors and warnings shown by api-extractor

catloversg:
Wrong setup of api-extractor.json:
```
Error: src/Augmentation/Enums.ts:1:1 - (ae-wrong-input-file-type) Incorrect file type; API Extractor expects to analyze compiler outputs with the .d.ts file extension. Troubleshooting tips: https://api-extractor.com/link/dts-error
Warning: src/Documentation/pages.ts:64:31 - (TS5097) An import path can only end with a '.ts' extension when 'allowImportingTsExtensions' is enabled.
```

Wrong usage of TSDoc:
```
Warning: src/ScriptEditor/NetscriptDefinitions.d.ts:4140:6 - (tsdoc-malformed-inline-tag) Expecting a TSDoc tag starting with "{@"
Warning: src/ScriptEditor/NetscriptDefinitions.d.ts:4141:24 - (tsdoc-malformed-inline-tag) Expecting a TSDoc tag starting with "{@"
Warning: src/ScriptEditor/NetscriptDefinitions.d.ts:4149:8 - (tsdoc-escape-right-brace) The "}" character should be escaped using a backslash to avoid confusion with a TSDoc inline tag
Warning: src/ScriptEditor/NetscriptDefinitions.d.ts:4150:6 - (tsdoc-escape-right-brace) The "}" character should be escaped using a backslash to avoid confusion with a TSDoc inline tag
```

Outdated api-extractor:
```
Warning: src/ui/MD/MD.tsx:20:41 - (TS2322) Type '"instant"' is not assignable to type 'ScrollBehavior | undefined'.
Warning: src/utils/SaveDataUtils.ts:19:84 - (TS2304) Cannot find name 'CompressionStream'.
Warning: src/utils/SaveDataUtils.ts:24:86 - (TS2304) Cannot find name 'DecompressionStream'.
```

### api-extractor.json
api-extractor will read tsconfig.json and process relevant files. With our tsconfig.json, api-extractor will try to process **everything** instead of just `NetscriptDefinitions.d.ts`. Please note that `mainEntryPointFilePath` is just the starting point for analysis.

### api-documenter
I have to "pin" it to v7.23.x. Starting from v7.24.0, api-documenter has a new behavior change: "Emit HTML tags for tables instead of Markdown code" (https://github.com/microsoft/rushstack/blob/main/apps/api-documenter/CHANGELOG.md#7240). This new change causes some bugs. One of them affects us. I created an issue in their repository: https://github.com/microsoft/rushstack/issues/4878.


api-extractor.json
```diff
@@ -88,9 +88,22 @@
      *
      * DEFAULT VALUE: no overrideTsconfig section
      */
-    // "overrideTsconfig": {
-    //   . . .
-    // }
+    "overrideTsconfig": {
```
d0sboots, line 95: Can you add a comment here describing what is actually being overridden? (I assume it's the "include", but it's not obvious, and will become less so if tsconfig changes.)

Alternately, if this overridden tsconfig can be slimmed down, that could help.

package.json
```diff
@@ -54,8 +54,8 @@
     "@babel/preset-env": "^7.22.20",
     "@babel/preset-react": "^7.22.15",
     "@babel/preset-typescript": "^7.23.0",
-    "@microsoft/api-documenter": "^7.23.9",
-    "@microsoft/api-extractor": "^7.38.0",
+    "@microsoft/api-documenter": "~7.23.38",
```
d0sboots, line 57: Can we add a comment about why this one is pinned?

catloversg: package.json does not support comments. There are some hacky (and ugly) workarounds that only work in some specific cases. AFAIK, there is no good way to add comments to "dependencies" and "devDependencies". Technically, this trick works:
```json
  "comments": {
    "devDependencies": {
      "@microsoft/api-documenter": "comment here"
    }
  },
```
I really don't like it. I'll only use it as a last resort.

d0sboots: Bleh, you're right, that's JSON for you.

I guess I'll be satisfied with blame pointing to this discussion.

PR #1568, opened 2024-08-11
UI: Add apostrophe to Stanek's Gift strings

Faenre:
# UI: Add apostrophe to Stanek's Gift strings

Stanek is the name of a person, and their gift is a possessive term. This PR changes a couple of instances of "Staneks Gift" to "Stanek's Gift" in the UI and in a couple of comments.

# Documentation

No functional or behavioral changes.

# Bug fix

## Testing:
Built the app, loaded a save file, visited the Church, accepted Stanek's gift, compared UI before and after.

## Before:
image

## After:
image

### DIFF
--- a/src/CotMG/StaneksGift.ts
+++ b/src/CotMG/StaneksGift.ts
@@ -243,12 +243,12 @@ export class StaneksGift extends BaseGift {
 
   static includedProperties = getKeyList(StaneksGift, { removedKeys: ["justCharged"] });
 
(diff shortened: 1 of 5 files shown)


d0sboots:
Changing the enum in SimplePage has the most potential for disruption. Unfortunately, it is also the most important one to change. 

I've looked, and I don't *think* this value is exposed via the NS api anywhere. So it should be safe to change. This comment serves as the place to say "you missed a spot" if I was wrong.

PR #1569, opened 2024-08-12
DEVMENU: Use Autocomplete instead of Select in augmentation tool

catloversg:
It's more convenient now.

Before:

https://github.com/user-attachments/assets/878f5161-1afc-4d9a-86ab-af5267dbbb2d

After:

https://github.com/user-attachments/assets/67f1514c-5cdd-4b0b-af04-d5f324d9d11b


d0sboots:
Nice! There are probably other places this could be used...

PR #1570, opened 2024-08-12
UI: Allow filtering graftable augmentations

catloversg:
The filter implementation is different from the one in `AugmentationsPage.tsx`. In `AugmentationsPage.tsx`, we filter by name, info, and stats of augmentations. I don't agree with that implementation. In this one, I only filter by name.

Screenshots:
Capture 1
Capture 2


d0sboots:
I can see the reason for filtering on all of those: It's very useful to be able to see all the "dexterity" augs, for instance. So I think I slightly prefer that style.

But much more than that, I think aug filtering should work the same in different places, even if that means it's maybe not what we most prefer.

PR #1571, opened 2024-08-12
UI: Notify players about documentation tab after getting SF1.1

catloversg:
Most pages in the "Advanced Mechanics" section are inaccessible until the player has at least SF 1.1. It's not obvious that those pages are unlocked when they get SF 1.1. I have seen many new players do not know that we have extensive documentation for BN mechanics.

This PR shows new players a message to notify them about new unlocked documentation pages when they get SF 1.1.


d0sboots:
Mmm nice, that's an elegant solution.

PR #1573, opened 2024-08-13
CORPORATION: Fix NaN Total Assets caused by bug in bulkPurchase API

catloversg:
This code is in `bulkPurchase`:
```js
material.averagePrice =
    (material.averagePrice * material.stored + material.marketPrice * amt) / (material.stored + amt);
```
When both `material.stored` and `amt` are 0, `material.averagePrice` will be NaN. This NaN value will be used in `updateTotalAssets`.

How to reproduce it:
- Create a corporation. Create a division.
- Use `bulkPurchase` (API or UI) with amount = 0.
- Check "Total Assets".

I think we should enforce the "positive number" constraint on the "amt" parameter of `bulkPurchase` in the next major version.

### DIFF
--- a/src/Corporation/Actions.ts
+++ b/src/Corporation/Actions.ts
@@ -333,7 +333,9 @@ export function buyMaterial(division: Division, material: Material, amt: number)
     throw new Error(`${material.name} is not a relevant material for industry ${division.type}`);
   }
   if (!Number.isFinite(amt) || amt < 0) {
-    throw new Error("Invalid value for amount field! Must be numeric and greater than or equal to 0");
+    throw new Error(
+      `Invalid amount '${amt}' to buy material '${material.name}'. Must be numeric and greater than or equal to 0`,
+    );
   }
   material.buyAmount = amt;
 }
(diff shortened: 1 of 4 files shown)


src/Corporation/Actions.ts
```diff
@@ -332,8 +332,8 @@ export function buyMaterial(division: Division, material: Material, amt: number)
   if (!isRelevantMaterial(material.name, division)) {
     throw new Error(`${material.name} is not a relevant material for industry ${division.type}`);
   }
-  if (isNaN(amt) || amt < 0) {
-    throw new Error(`Invalid amount '${amt}' to buy material '${material.name}'`);
+  if (!Number.isFinite(amt) || amt < 0) {
+    throw new Error("Invalid value for amount field! Must be numeric and greater than or equal to 0");
```
d0sboots: Let's keep the actual amount and material in the error, although specifying the proper constraints is good too.

src/Corporation/Actions.ts
```diff
@@ -350,22 +350,25 @@ export function bulkPurchase(
   }
   const matSize = MaterialInfo[material.name].size;
   const maxAmount = (warehouse.size - warehouse.sizeUsed) / matSize;
-  if (isNaN(amt) || amt < 0) {
-    throw new Error(`Invalid input amount`);
+  if (!Number.isFinite(amt) || amt < 0) {
+    throw new Error("Invalid value for amount field! Must be numeric and greater than or equal to 0");
```
d0sboots: Ditto.

catloversg:
The CI test failed in `test/jest/StockMarket.test.ts` again. We can ignore it.

PR #1574, opened 2024-08-13
DOCUMENTATION: Remove wrong information in ns.weaken

### LINKED ISSUE #1572: DOCUMENTATION: API reference ns.weaken() contains "constant" 0.05 which may not be true
Bitburner v2.6.2 (633da3830)

https://github.com/bitburner-official/bitburner-src/blob/633da383016ad0521f9f1c17cdd99478d2701e41/markdown/bitburner.ns.weaken.md?plain=1#L32

The API document implies that the amount of security decreased is constant 0.05.

However, it may not be true when one of the following is true:
* The number of threads is not 1
* The number of cores is not 1
* BitNode has ServerWeakenRate. (i.e. 11 and 12)

catloversg:
Fixes #1572.

### DIFF
--- a/markdown/bitburner.ns.hack.md
+++ b/markdown/bitburner.ns.hack.md
@@ -33,7 +33,7 @@ Function that is used to try and hack servers to steal money and gain hacking ex
 
 A script can hack a server from anywhere. It does not need to be running on the same server to hack that server. For example, you can create a script that hacks the `foodnstuff` server and run that script on any server in the game.
 
-A successful `hack()` on a server will raise that server’s security level by 0.002.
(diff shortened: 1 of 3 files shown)


gmcew:
In case you haven't seen the follow-up comment yet; #1572 also notes `ns.hack()` has a misleading security reference too. `ns.grow()` appears to correctly refer to the `growthAnalyzeSecurity` function for the relevant information.

For consistency and new-player friendliness, I would suggest they all point to their respective `analyze`/`analyzeSecurity` functions rather than a complete removal.

d0sboots:
I also left some comments on the issue (probably best to keep discussion there)

PR #1578, opened 2024-08-14
CI: Update Node version

catloversg:
@d0sboots Should we update `actions/checkout` and `actions/setup-node`?

### DIFF
--- a/.github/workflows/bump-version.yml
+++ b/.github/workflows/bump-version.yml
@@ -42,10 +42,10 @@ jobs:
         uses: actions/checkout@v2
         with:
           fetch-depth: 0
-      - name: Use Node.js 16.13.1
+      - name: Use Node.js 20
         uses: actions/setup-node@v2
         with:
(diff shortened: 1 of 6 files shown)


d0sboots:
> @d0sboots Should we update `actions/checkout` and `actions/setup-node`?

If we're poking the CI, might as well update everything that's badly out of date

catloversg:
Some notable changes that I found when checking the changelog:
- `actions/upload-artifact`:
  - `Due to how Artifacts are created in this new version, it is no longer possible to upload to the same named Artifact multiple times. You must either split the uploads into multiple Artifacts with different names, or only upload once. Otherwise you will encounter an error.`.
  - I don't think that this change affects us.
- `tj-actions/changed-files`:
  - `Directory patterns now require explicit specification of the globstar pattern to match all sub paths.` (v40).
  - I changed the pattern from `dist/*` to `dist/**`.

PR #1580, opened 2024-08-14
BUGFIX: Wrong error message when failing to recruit gang member

catloversg:
There are 4 reasons that the player cannot recruit a gang member:
- Empty name.
- Duplicated name.
- Exceed max number. [1]
- Not enough respect. [2]

`recruitMember` shows a wrong error message in [1] and [2].

### DIFF
--- a/src/Gang/Gang.ts
+++ b/src/Gang/Gang.ts
@@ -27,7 +27,8 @@ import { PowerMultiplier } from "./data/power";
 import { FactionName } from "@enums";
 import { CONSTANTS } from "../Constants";
 
-export enum RecruitmentErrorType {
+export enum RecruitmentResult {
+  Success = "Success",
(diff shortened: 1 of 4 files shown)


src/Gang/Gang.ts
```diff
@@ -301,9 +308,22 @@ export class Gang {
     }
   }
 
-  canRecruitMember(): boolean {
-    if (this.members.length >= GangConstants.MaximumGangMembers) return false;
-    return this.respect >= this.respectForNextRecruit();
+  canRecruitMember(): { isSuccess: boolean; errorType?: RecruitmentErrorType } {
```
d0sboots: Rather than have a compound type here, I think it would make more sense to return `RecruitmentResult`, which would be your current error enum plus "Success".

PR #1584, opened 2024-08-15
DOCUMENTATION: Clarify security increase of ns.hack and ns.grow

catloversg:
This PR makes 2 changes:
- `ns.grow`: add new information: the security increase is 0.004 per thread.
- `ns.grow` and `ns.hack`: "per thread" is not exactly true. I change it to "per used thread" and explain what "used thread" means. It answers these questions:
  - "If it's as simple as 'X per thread', why do I need to use other functions (`hackAnalyzeSecurity` and `growthAnalyzeSecurity`)?"
  - "I call `ns.grow` to grow server X, but the security increase is not 0.004 per thread".
  - "I call `ns.hack` to hack server X, but the security increase is not 0.002 per thread".

### DIFF
--- a/markdown/bitburner.ns.grow.md
+++ b/markdown/bitburner.ns.grow.md
@@ -47,7 +47,9 @@ To determine how many threads are needed to return a server to max money, obtain
 
 Like [hack](./bitburner.ns.hack.md)<!-- -->, `grow` can be called on any hackable server, regardless of where the script is running. Hackable servers are any servers not owned by the player.
 
-The grow() command requires root access to the target server, but there is no required hacking level to run the command. It also raises the security level of the target server based on the number of threads. The security increase can be determined using [growthAnalyzeSecurity](./bitburner.ns.growthanalyzesecurity.md)<!-- -->.
+The grow() command requires root access to the target server, but there is no required hacking level to run the command. It also raises the security level of the target server based on the number of threads.
+
+The security level is increased by 0.004 per "used thread". "used thread" may not be the same as opts.threads. Let's say you call this function with opts.threads = 1000, but it only takes 500 threads to grow the targeted server to the max money. In this case, "used thread" is 500. You should use [growthAnalyzeSecurity](./bitburner.ns.growthanalyzesecurity.md) to calculate the security increase.
 
 ## Example
 
--- a/markdown/bitburner.ns.hack.md
+++ b/markdown/bitburner.ns.hack.md
@@ -33,7 +33,7 @@ Function that is used to try and hack servers to steal money and gain hacking ex
 
 A script can hack a server from anywhere. It does not need to be running on the same server to hack that server. For example, you can create a script that hacks the `foodnstuff` server and run that script on any server in the game.
 
-A successful `hack()` on a server will raise that server’s security level by 0.002 per thread. You can use [hackAnalyzeSecurity](./bitburner.ns.hackanalyzesecurity.md) to calculate the security increase for a number of threads.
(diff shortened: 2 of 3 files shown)


markdown/bitburner.ns.grow.md
```diff
@@ -47,7 +47,9 @@ To determine how many threads are needed to return a server to max money, obtain
 
 Like [hack](./bitburner.ns.hack.md)<!-- -->, `grow` can be called on any hackable server, regardless of where the script is running. Hackable servers are any servers not owned by the player.
 
-The grow() command requires root access to the target server, but there is no required hacking level to run the command. It also raises the security level of the target server based on the number of threads. The security increase can be determined using [growthAnalyzeSecurity](./bitburner.ns.growthanalyzesecurity.md)<!-- -->.
+The grow() command requires root access to the target server, but there is no required hacking level to run the command. It also raises the security level of the target server based on the number of threads.
+
+The security level is increased by 0.004 per "used thread". "used thread" may not be the same as opts.threads. Let's say you call this function with opts.threads = 1000, but it only takes 500 threads to grow the targeted server to the max money. In this case, "used thread" is 500. You should use [growthAnalyzeSecurity](./bitburner.ns.growthanalyzesecurity.md) to calculate the security increase.
```
d0sboots, line 52:
```suggestion
The security level is increased by 0.004 per "used thread". "Used threads" may not be the same as opts.threads. Let's say you call this function with opts.threads = 1000, but it only takes 500 threads to grow the targeted server to the max money. In this case, "used threads" is 500. You can use [growthAnalyzeSecurity](./bitburner.ns.growthanalyzesecurity.md) to calculate the security increase.
```

If you're getting into the gritty details (which I agree is a good thing), do you want to talk about fractional threads and their relation to used threads?

markdown/bitburner.ns.hack.md
```diff
@@ -33,7 +33,7 @@ Function that is used to try and hack servers to steal money and gain hacking ex
 
 A script can hack a server from anywhere. It does not need to be running on the same server to hack that server. For example, you can create a script that hacks the `foodnstuff` server and run that script on any server in the game.
 
-A successful `hack()` on a server will raise that server’s security level by 0.002 per thread. You can use [hackAnalyzeSecurity](./bitburner.ns.hackanalyzesecurity.md) to calculate the security increase for a number of threads.
+A successful `hack()` on a server will raise that server’s security level by 0.002 per "used thread". "used thread" may not be the same as opts.threads. Let's say you call this function with opts.threads = 1000, but it only takes 500 threads to hack all money in the targeted server. In this case, "used thread" is 500. You should use [hackAnalyzeSecurity](./bitburner.ns.hackanalyzesecurity.md) to calculate the security increase.
```
d0sboots, line 36:
```suggestion
A successful `hack()` on a server will raise that server’s security level by 0.002 per "used thread". "Used threads" may not be the same as opts.threads. Let's say you call this function with opts.threads = 1000, but it only takes 500 threads to hack all money in the targeted server. In this case, "used threads" is 500. You should use [hackAnalyzeSecurity](./bitburner.ns.hackanalyzesecurity.md) to calculate the security increase.
```

There is a subtle point with `hack()` that this glosses over: While `grow` is capped based on the number of threads needed to grow to max money, `hack` is capped based on _the number of threads needed to hack from full to 0_. In other words, even if the server has 1 money, you can still "use" up to 500 threads (in your example). There's also an early-out special case for 0 money where used threads will be 0. I.e. the situation is quite complicated for hack.

d0sboots (commented):
I'm all for more docs, but these functions are weird and knobbly (partly by accident, but now intentionally frozen that way), so documenting them is a bit fraught.

markdown/bitburner.ns.hack.md
Alpheus: If it's a rule engine (rather than formula) documenting as bullet points makes more sense.

PR #1585, opened 2024-08-15
BUGFIX: Block hacking-related actions on player-owned servers

catloversg:
Implementations of hack, grow, weaken, and backdoor are inconsistent between CLI and API. In CLI, we check `server.purchasedByPlayer`, but we do not check it in API. This PR fixes it. As we discussed on Discord, this change is pretty harmless.

On Discord, I talked about this case: `await ns.singularity.installBackdoor("hacknet-node-0")` shows `singularity.installBackdoor: Successfully installed backdoor on 'home'.` Just ignore it. I forgot that `installBackdoor` is a no-parameter API, and we need to connect to the targeted server before using it. We cannot connect to a Hacknet Node (not Hacknet Server), so it won't be a problem.

Although the PR's title says "hacking-related actions", I don't touch nuke and port-opener APIs in this PR. It seems unnecessary to change them. If you think we should change them too, I'll do it.

Test code:
```js
/** @param {NS} ns */
  ns.tail();
  ns.clearLog();
  await ns.hack("home");
  await ns.grow("home");
  await ns.weaken("home");
  ns.singularity.connect("home"); await ns.singularity.installBackdoor();

  await ns.hack("hacknet-node-0");

  await ns.hack("hacknet-server-0");
  await ns.grow("hacknet-server-0");
  await ns.weaken("hacknet-server-0");
  ns.singularity.connect("hacknet-server-0"); await ns.singularity.installBackdoor();
}
```

### DIFF
--- a/src/Hacking/netscriptCanHack.ts
+++ b/src/Hacking/netscriptCanHack.ts
@@ -9,28 +9,28 @@ import { IReturnStatus } from "../types";
 import { Player } from "@player";
 import { Server } from "../Server/Server";
 
-function baseCheck(server: Server, fnName: string): IReturnStatus {
+function baseCheck(server: Server, actionName: string): IReturnStatus {
   const hostname = server.hostname;
 
-  if (!("requiredHackingSkill" in server)) {
+  if (server.purchasedByPlayer) {
     return {
       res: false,
-      msg: `Cannot ${fnName} ${hostname} server because it is a Hacknet Node`,
+      msg: `Cannot ${actionName} ${hostname} server because it is your server`,
     };
   }
 
   if (!server.hasAdminRights) {
     return {
       res: false,
-      msg: `Cannot ${fnName} ${hostname} server because you do not have root access`,
+      msg: `Cannot ${actionName} ${hostname} server because you do not have root access`,
     };
   }
 
   return { res: true };
 }
 
-export function netscriptCanHack(server: Server): IReturnStatus {
-  const initialCheck = baseCheck(server, "hack");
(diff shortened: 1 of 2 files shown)


src/Hacking/netscriptCanHack.ts
```diff
@@ -9,28 +9,28 @@ import { IReturnStatus } from "../types";
 import { Player } from "@player";
 import { Server } from "../Server/Server";
 
-function baseCheck(server: Server, fnName: string): IReturnStatus {
+function baseCheck(server: Server, actionName: string): IReturnStatus {
   const hostname = server.hostname;
 
-  if (!("requiredHackingSkill" in server)) {
+  if (server.purchasedByPlayer) {
```
d0sboots, line 15: Just to double-check: There are no cases where we are loosening the check, right? The message claims the requiredHackingSkill check was only for hacknet, but that doesn't mean it's right.

catloversg: The old check is useless. Before we call the triplet `netscriptCan-X`, we always check `!(server instanceof Server)` (the type of the first parameter of `baseCheck` is `Server`). Hacknet Node and Hacknet Server are not `Server`.

d0sboots: Ah, this was my bad. I was thinking this was checking against `undefined`, but it's actually checking `"requiredHackingSkill" in server` which is the subtly different thing that can still be true even if the value is undefined. Since `Server` always sets the property in the constructor, it really is equivalent to checking `!(server istanceof Server)`, just more awkwardly.

PR #1586, opened 2024-08-15
BLADE: Adjust IV action to current-city affect rather than global

Faenre:
# Adjust Bladeburner IV action to affect a single city

This PR makes the Bladeburner general action "Incite Violence" affect only the current city, rather than all cities on the map. This change opens up potential room for strategic behavior from the player, while removing a trap that can very easily lead to ruining an entire bitnode mid-play with very little effort or guidance.

image

# Context

Choice of player city is a core aspect of Bladeburner, where anything that affects any city, affects the player's Bladeburner HQ city. Typically, for most actions, you want to choose a city with higher populations, because they make your tasks easier; and due to Chaos penalties, you typically need to perform Diplomacy actions any time Chaos approaches 50, otherwise it becomes increasingly difficult to progress.

The Incite Violence general action has the stated goal of providing new contracts for the player, at the expense of increased Chaos. However, _unlike all other Bladeburner actions,_ right now this affects **all** cities. This is counter-intuitive and non-obvious to the user, as well as being non-visible to the player unless they have scripted a dashboard for themself using the Bladeburner API (which is not immediately available).

Current implementation:
- All Bladeburner mechanics except for Incite Violence are targeted to specific regions
- Incite Violence causes all cities grow chaos at both a static and inverse-logarithmic rate
- There is no player skill scaling in either contract gain or violence increase
- The static increase `chaos += 10` is more drastic already than any other player action

This leads to the following results:
- The player's choice of city is irrelevant for this mechanic, which goes against all other Bladeburner mechanics
- The player can lock themselves out of hours of gameplay with only a few cycles
- A character with end-node-levels of high charisma can essentially ignore this penalty, but early-mid-node, the chaos penalty locks you out of **all** Bladeburner activities quite quickly, because the Chaos penalty to Contracts and Operations can quickly bring you into < 10% success rate on even Tracking contracts

For comparison with other related mechanics:
- Choice-of-city is directly relevant to the Bladeburner mechanic design. Every other action that affects the world, affects the _current_ city, with difficulties and impacts relating to actions there
- The highest chaos changes from other mechanics are:
 - `+ 1.00 + getRandomIntInclusive(5, 20) / 100` from synthoid riots random event, at 20% chance
 - `+ 0.01` flat gain from Sting operations
 - `+ 1-5%` percentage gain from Raid operations
 - `+/- up to 5%` from Assassination operations
 - `+ 0.02` from Bounty Hunter contracts
 - `+ 0.04` from Retirement contracts 
 - All of the above are _single-city_ adjustments

Incite Violence has **both** a flat gain that is **250x** as powerful as the next most disruptive player event, and a **minimum +10% scaling** that is more than double the Raid operations, with no ceiling. Meanwhile, the benefit is a **constant, non-scaling** amount of change. 

# Solution

The changes made at time of writing remove the global factors to the city population gains.

# Linked issues

Per Discord conversation:
- [Venusaur's comment on August 14 asking if this is a bug](https://discord.com/channels/415207508303544321/923445914461413376/1273465421433143408)
- More people thinking this is a bug ([list created by FicocelliGuy](https://discord.com/channels/415207508303544321/923445914461413376/1273681258865492048)): [1](https://discord.com/channels/415207508303544321/415207923506216971/1157269063198855229) [2](https://discord.com/channels/415207508303544321/923445914461413376/1235822552740270131) [3](https://discord.com/channels/415207508303544321/415207923506216971/1231866450084364379) [4](https://discord.com/channels/415207508303544321/923445914461413376/1226607705561763873) [5](https://discord.com/channels/415207508303544321/923445914461413376/1199803564642549780) [6](https://discord.com/channels/415207508303544321/923445914461413376/1183834409330286602) [7](https://discord.com/channels/415207508303544321/415207923506216971/1147959675929772063)
- [CatLover's response that "it's intentional, but it does not mean that it's a good idea](https://discord.com/channels/415207508303544321/923445914461413376/1273581155907538987)

Conversation ensues from that point.

# Documentation

Clarity is added to the general action in-game description.

# Testing

After changes, before inciting violence:
Screenshot 2024-08-15 at 11 30 36 AM

After changes, after 1 incite violence:
image

After changes, after getting up and going to the restroom and coming straight back:
image

Before the changes, this affected all cities in the game equally.

### DIFF
--- a/src/Bladeburner/Bladeburner.ts
+++ b/src/Bladeburner/Bladeburner.ts
@@ -1203,7 +1203,7 @@ export class Bladeburner {
             }
             const city = this.getCurrentCity();
             city.chaos += 10;
-            city.chaos += city.chaos / Math.log(city.chaos / 10);
+            city.chaos += city.chaos / (Math.log(city.chaos) / Math.log(10));
             break;
           }
           default: {
--- a/src/Bladeburner/Bladeburner.ts
+++ b/src/Bladeburner/Bladeburner.ts
@@ -646,7 +646,7 @@ export class Bladeburner {
     } else if (chance <= 0.7) {
       // Synthoid Riots (+chaos), 20%
       sourceCity.changeChaosByCount(1);
-      sourceCity.changeChaosByPercentage(getRandomIntInclusive(5, 20) / 100);
+      sourceCity.changeChaosByPercentage(getRandomIntInclusive(5, 20));
       if (this.logging.events) {
         this.log("Tensions between Synthoids and humans lead to riots in " + sourceCityName + "! Chaos increased");
       }
--- a/src/Bladeburner/Bladeburner.ts
+++ b/src/Bladeburner/Bladeburner.ts
@@ -1201,11 +1201,9 @@ export class Bladeburner {
             if (this.logging.general) {
               this.log(`${person.whoAmI()}: Incited violence in the synthoid communities.`);
             }
-            for (const cityName of Object.values(CityName)) {
-              const city = this.cities[cityName];
-              city.chaos += 10;
-              city.chaos += city.chaos / (Math.log(city.chaos) / Math.log(10));
-            }
+            const city = this.getCurrentCity();
+            city.chaos += 10;
+            city.chaos += city.chaos / Math.log(city.chaos / 10);
             break;
           }
           default: {
--- a/src/Bladeburner/Bladeburner.ts
+++ b/src/Bladeburner/Bladeburner.ts
@@ -1203,7 +1203,7 @@ export class Bladeburner {
             }
             const city = this.getCurrentCity();
             city.changeChaosByCount(10);
-            city.changeChaosByPercentage(getRandomIntInclusive(2, 10));
+            city.changeChaosByCount(city.chaos / Math.log(city.chaos) / Math.log(10));
             break;
           }
           default: {
--- a/src/Bladeburner/Bladeburner.ts
+++ b/src/Bladeburner/Bladeburner.ts
@@ -645,8 +645,8 @@ export class Bladeburner {
       }
     } else if (chance <= 0.7) {
       // Synthoid Riots (+chaos), 20%
-      sourceCity.chaos += 1;
-      sourceCity.chaos *= 1 + getRandomIntInclusive(5, 20) / 100;
+      sourceCity.changeChaosByCount(1);
+      sourceCity.changeChaosByPercentage(getRandomIntInclusive(5, 20) / 100);
       if (this.logging.events) {
         this.log("Tensions between Synthoids and humans lead to riots in " + sourceCityName + "! Chaos increased");
       }
@@ -1202,8 +1202,8 @@ export class Bladeburner {
               this.log(`${person.whoAmI()}: Incited violence in the synthoid communities.`);
             }
             const city = this.getCurrentCity();
-            city.chaos += 10;
-            city.chaos += city.chaos / (Math.log(city.chaos) / Math.log(10));
+            city.changeChaosByCount(10);
+            city.changeChaosByPercentage(getRandomIntInclusive(2, 10));
             break;
           }
           default: {
--- a/src/Bladeburner/data/GeneralActions.ts
+++ b/src/Bladeburner/data/GeneralActions.ts
@@ -53,6 +53,6 @@ export const GeneralActions: Record<BladeGeneralActionName, GeneralAction> = {
     getActionTime: () => 60,
     desc:
       "Purposefully stir trouble in the synthoid community in order to gain a political edge. This will generate " +
-      "additional contracts and operations, at the cost of increased Chaos.",
+      "additional contracts and operations, at the cost of increased Chaos in your current city.",
   }),
 };
--- a/src/Bladeburner/data/GeneralActions.ts
+++ b/src/Bladeburner/data/GeneralActions.ts
@@ -53,6 +53,6 @@ export const GeneralActions: Record<BladeGeneralActionName, GeneralAction> = {
     getActionTime: () => 60,
     desc:
       "Purposefully stir trouble in the synthoid community in order to gain a political edge. This will generate " +
-      "additional contracts and operations, at the cost of increased Chaos in your current city.",
+      "additional contracts and operations. Completing this action will increase the Chaos level in your current city.",
   }),
 };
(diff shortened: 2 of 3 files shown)


src/Bladeburner/Bladeburner.ts
```diff
@@ -1201,11 +1201,9 @@ export class Bladeburner {
...
               this.log(`${person.whoAmI()}: Incited violence in the synthoid communities.`);
             }
-            for (const cityName of Object.values(CityName)) {
-              const city = this.cities[cityName];
-              city.chaos += 10;
-              city.chaos += city.chaos / (Math.log(city.chaos) / Math.log(10));
-            }
+            const city = this.getCurrentCity();
+            city.changeChaosByCount(10);
+            city.changeChaosByCount(city.chaos / Math.log(city.chaos) / Math.log(10));
```
Faenre, line 1206: This part is the logical change. The scaling stays the same in this PR, but the penalty now only affects the player's current city.

Also, the change to `changeChaosByCount` handles integer safety with a clamp that defaults to `Number.MAX_VALUE`.

d0sboots:
I will be straight with you: Bladeburner is one of the areas I am least familiar with, since I've never done it myself. So, I can review this code-wise, but I don't have the context to tell if it makes sense game-wise. That'll require either @Snarling, or a community consensus that this is a good idea.

src/Bladeburner/Bladeburner.ts
d0sboots, line 1206: You dropped the parens here, which changes order-of-ops. Probably the best thing is to use Math.log10 directly.

d0sboots (commented):
I would suggest (but not require) that you split this into two parts. There's some cleanup/refactoring that I can merge immediately (and that distracts from the logical change), and then the actual change, which I suspect will be debated for a bit.

src/Bladeburner/Bladeburner.ts
Faenre: Fair point, I'll put a changeChaosByCount bugfix together separately and make this one just about the city change.

Faenre: https://github.com/bitburner-official/bitburner-src/pull/1588

Snarling (commented):
From a balance perspective, this basically makes it no-cost to incite violence - have a burner city for generating contracts by inciting violence, and then actually perform the contracts in a different city.

If contract / operation counts were per-city, then this change would make sense, but since they are global I lean towards incite violence making sense as having a global impact since it is generating more global contract/operation counts.

That said I am open to other viewpoints.

Faenre:
> From a balance perspective, this basically makes it no-cost to incite violence - have a burner city for generating contracts by inciting violence, and then actually perform the contracts in a different city.
> 
> If contract / operation counts were per-city, then this change would make sense, but since they are global I lean towards incite violence making sense as having a global impact since it is generating more global contract/operation counts.
> 
> That said I am open to other viewpoints.

Is that a problem, though?

If the goal is to put a limiter on players generating contracts for int-farming, perhaps we can do that another way? As it is, the global chaos gain is a horrible player experience.

- Because _Diplomacy_ scales on charisma and applies a percentage change, players who are int-farming don't care anyway. [They just get however much charisma is needed, then wipe 100% of the chaos per diplomacy](https://discord.com/channels/415207508303544321/923445914461413376/1273646377297121280)
- Early- and mid-node, this can quickly burn not just one city, but every city, to the point of Bladeburner being entirely unusable. You essentially can't come back to it later on without spending hours to build charisma and whittle away at it.

Changing it to be single-city is a strategic investment that is potentially punishing, but it means players who don't have sleeves yet have a method to generate contracts and operations that doesn't brick their run.

Basically the global-hit limits/punishes the wrong people. It's not powerful enough to be worth it if you're just trying to clear a node, and it does nothing to slow people down who are int-farming

Snarling:
Beating a node with Bladeburner is time gated and stat gated - you need time to generate enough contracts / operations to reach enough rank for the later blackops, and you need stats to be able to actually complete those tasks.

Tools have been added to ease the time gate - Incite Violence and the sleeve "Infiltrate Synthoids" task, but I don't think it's healthy for the game to erode the time gate too much. Incite violence is supposed to have a real cost.

If it's going to only raise chaos in one city, then there needs to be some tradeoff, a reason to not just ignore chaos management and concentrate all the chaos in one bad city. For instance the task could become less effective if chaos is higher in the current city.

Faenre:
After a day of lots of discussion, [a consensus seems to be](https://discord.com/channels/415207508303544321/923445914461413376/1273807781790748695) not to adjust IV to only affect 1 city, but that it should be a flat scaling (perhaps +10 to all cities), plus have a chance to trigger a random event. Potentially, it's been suggested that Diplomacy receive an early-game buff as well, perhaps a flat amount plus the scaling (maybe -2).

Getting rid of the scaling factor solves the exponential growth which leads to bricked/bitflumed circumstances:
> OK, so I just ran the numbers. If your cha is 1000, it will take 4650 runs of diplomacy to get down from 1e50 to 50 chaos
> That counts as "bricked" to me
—[d0sboots](https://discord.com/channels/415207508303544321/923445914461413376/1273796746065215499)

As well, adding a small flat amount to the Diplomacy action will help the player at lower charisma, without making it busted OP. Otherwise, _even at 1000 charisma_, chaos in the 50-100 range will only recover 1-2 points per minute. 

Before I get to coding it, how does this sound?

d0sboots:
Chiming in with a late addition from the discussion that CatLover has requested some time to look at the mechanics too, so probably we can chill on this for a few days and resume discussion later. (As I suspected, even the "simple" things are not always simple or fast XD)

Alpheus:
I vote to merge the refactor and split the functional change to a new PR.

I would like to test cover this separately to make it easier to reason about.

d0sboots:
Oh yeah, we never really got back to this. Also, I forgot there was other refactor here (although it's mostly incidental.)

src/Bladeburner/Bladeburner.ts
Alpheus: To continue on the discord discussion: I vote for combining the two logics. This is not a perfect, final solution, but merely something that I'm confident will move this PR forward and allow us to make adjustments that have a positive effect on playability.

- give the old, heavy penalty to the current city, but not others
- apply the flat penalty to all cities

This way the cost of generating contracts remains global, but the escalating effect only stays on the one you are currently looking at (ie. you won't accidentally end up with your non-current city having 1e50 chaos in two days).

Alpheus: I request separating this commit out to a new PR and reverting it in this one so we can merge the surrounding refactors.

PR #1587, opened 2024-08-15
DOCUMENTATION: Update documentation of Corporation API

### LINKED ISSUE #1582: DOCUMENTATION: behavior of ns.corporation.createCorporation does not match with document.
Bitburner v2.6.2 (633da3830)

After I read https://github.com/bitburner-official/bitburner-src/blob/633da383016ad0521f9f1c17cdd99478d2701e41/markdown/bitburner.corporation.createcorporation.md ,
I expected that `ns.corporation.createCorporation` would return false if the creation of corporation has been failed in anyway. i.e.

1) You don't have access to corporation. (i.e. you don't have sf3 and you are not in bn3)
2) You already have corporation
3) Corporation name is not given
4) You selected to seed-money but you are not in bn3
5) Current bn doesn't allow to create corporation
6) You selected to self-fund but you don't have enough money

I expected that all of these case returns false.

However, in current implementation https://github.com/bitburner-official/bitburner-src/blob/633da383016ad0521f9f1c17cdd99478d2701e41/src/Corporation/Actions.ts , 1, 2, 3, 6 returns false, while 4, 5 throws error.

### LINKED ISSUE #1583: DOCUMENTATION: API document doesn't mention RAM usage of ns.corporation functions.
Bitburner v2.6.2 (633da3830)

image

Except `nextUpdate`, these corporation function doesn't mention RAM usage on API document.

catloversg:
Fixes #1582.
Fixes #1583.

### DIFF
--- a/markdown/bitburner.corporation.createcorporation.md
+++ b/markdown/bitburner.corporation.createcorporation.md
@@ -29,5 +29,9 @@ true if created and false if not
 
 RAM cost: 20 GB
 
-This function throws an error if: - Try to self-fund outside BitNode 3. - Be in a BitNode that has CorporationSoftcap (a BN modifier) less than 0.15. Use [getBitNodeMultipliers](./bitburner.ns.getbitnodemultipliers.md) to get the value of this modifier.
+This function throws an error if:
+
+- Try to self-fund outside BitNode 3.
+
+- Be in a BitNode that has CorporationSoftcap (a BN modifier) less than 0.15. Use [getBitNodeMultipliers](./bitburner.ns.getbitnodemultipliers.md) to get the value of this modifier.
 
--- a/markdown/bitburner.corporation.createcorporation.md
+++ b/markdown/bitburner.corporation.createcorporation.md
@@ -4,7 +4,7 @@
 
 ## Corporation.createCorporation() method
 
-Create a Corporation
+Create a Corporation.
 
 **Signature:**
 
@@ -17,11 +17,17 @@ createCorporation(corporationName: string, selfFund: boolean): boolean;
 |  Parameter | Type | Description |
 |  --- | --- | --- |
 |  corporationName | string | Name of the corporation |
-|  selfFund | boolean | If you should self fund, defaults to true, false will only work on BitNode 3 |
+|  selfFund | boolean | If you want to self-fund. Defaults to true, false will only work in BitNode 3. |
 
 **Returns:**
 
 boolean
 
 true if created and false if not
+
+## Remarks
+
+RAM cost: 20 GB
+
+This function throws an error if: - Try to self-fund outside BitNode 3. - Be in a BitNode that has CorporationSoftcap (a BN modifier) less than 0.15. Use [getBitNodeMultipliers](./bitburner.ns.getbitnodemultipliers.md) to get the value of this modifier.
 
--- a/src/Script/RamCalculations.ts
+++ b/src/Script/RamCalculations.ts
@@ -161,7 +161,7 @@ function parseOnlyRamCalculate(
 
   // Finally, walk the reference map and generate a ram cost. The initial set of keys to scan
   // are those that start with the name of the main script.
-  let ram = RamCostConstants.Base;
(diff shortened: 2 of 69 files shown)


markdown/bitburner.corporation.createcorporation.md
```diff
@@ -17,11 +17,17 @@ createCorporation(corporationName: string, selfFund: boolean): boolean;
...
 
 boolean
 
 true if created and false if not
 
+## Remarks
+
+RAM cost: 20 GB
+
+This function throws an error if: - Try to self-fund outside BitNode 3. - Be in a BitNode that has CorporationSoftcap (a BN modifier) less than 0.15. Use [getBitNodeMultipliers](./bitburner.ns.getbitnodemultipliers.md) to get the value of this modifier.
```
d0sboots: Bulleted lists get munged

src/Script/RamCalculations.ts
```diff
@@ -161,7 +161,7 @@ function parseOnlyRamCalculate(
 
   // Finally, walk the reference map and generate a ram cost. The initial set of keys to scan
   // are those that start with the name of the main script.
-  let ram = RamCostConstants.Base;
+  let ram: number = RamCostConstants.Base;
```
d0sboots, line 164: What changed such that this was necessary? I know that "as const" allows for tighter inferences, but AFAIK it should be the same as initializing this to 1.6, which would infer the type number.

d0sboots (commented):
Good docs, only one issue.

src/Script/RamCalculations.ts
catloversg: When we use `as const`, the type is narrowed down as tight as possible. For example:
```typescript
let ram = RamCostConstants.Base;
```
In this case, the type of `ram` is not `number`, it's `1.6`. We cannot assign any other values. You can check it at https://github.com/bitburner-official/bitburner-src/actions/runs/10408238673/job/28825245929.

markdown/bitburner.corporation.createcorporation.md
bupjae: Regarding CorporationSoftcap,

The game uses `CorporationSoftcap < 0.15` to mark that corp is disabled in current bn. (Currently, only bn8 has CorporationSoftcap === 0).

https://github.com/bitburner-official/bitburner-src/blob/633da383016ad0521f9f1c17cdd99478d2701e41/src/BitNode/ui/BitnodeMultipliersDescription.tsx#L333-L339

Maybe number 0.15 is here because they want to prevent "tax" formular breakout when "tax exponent" goes to negative.

However, I think `CorporationSoftcap < 0.15` is implementation details, and there should be some method to provide whether Corporation is disabled in this bn without refering CorporationSoftcap. (accessing CorporationSoftcap needs sf5)

d0sboots: You can try to create a corporation and see if it succeeds. :D

d0sboots:
It's a little awkward that the list will be broken into separate paragraphs, but good enough.

src/Script/RamCalculations.ts
d0sboots: Hmm, I guess it's acting like `let ram = 1.6 as const`, i.e. the const becomes transitive. Makes sense I suppose, but not what I was expecting.

markdown/bitburner.corporation.createcorporation.md
catloversg: I agree with bupjae:
- Accessing CorporationSoftcap requires SF5.
- Need to know the "magic" number 0.15 (by reading the TSDoc). Implementation details like this should not matter for players.
- Technically, they can bypass this problem by wrapping the API in try-catch, but it looks hacky and ugly.

@d0sboots Can I add `ns.corporation.isCorporationDisabled`?

d0sboots: I'd call it isCorporationEnabled or (perhaps better) canCreateCorporation, which potentially could test for more cases. 

But yeah, go for it if you want.

PR #1588, opened 2024-08-15
BUGFIX: Fix Bladeburner city chaos reaching Infinity/NaN

Faenre:
# Fix Bladeburner cities from reaching Infinity/NaN Chaos

It's currently possible to accidentally brick the Bladeburner mechanic by hitting Infinity chaos, which is irrecoverable from. This is possible because of places that manually `+=` or `*=` on the Chaos value without clamping the number between 0 and `Number.MAX_VALUE` (1.798e+308). This clamping feature is performed automatically by the provided functions `changeChaosByCount` and `changeChaosByPercentage` which are already built-in to the City class, but some locations are improperly modifying the chaos attribute directly. This PR fixes that.

# Bug fix

Tested using the dev console to add Infinity levels of chaos, then performing Incite Violence and other actions.

This is the result:

image

## Diplomacy efficiency -> percentage

D0sboots suggested renaming the `getDiplomacyEfficiency` to `getDiplomacyPercentage` to clean some of the code up. I have attached a test showing that 

Testing with the following servers:
image

Per the charisma efficiency, I would expect a ~1.82% change (calculated at 500 charisma):
image

In-game testing results demonstrate the correct percentage:
image

### DIFF
--- a/src/Bladeburner/Bladeburner.ts
+++ b/src/Bladeburner/Bladeburner.ts
@@ -700,14 +700,14 @@ export class Bladeburner {
     };
   }
 
-  getDiplomacyEffectiveness(person: Person): number {
-    // Returns a decimal by which the city's chaos level should be multiplied (e.g. 0.98)
+  getDiplomacyPercentage(person: Person): number {
+    // Returns a percentage by which the city's chaos level should be modified (e.g. 2 for 2%)
     const CharismaLinearFactor = 1e3;
     const CharismaExponentialFactor = 0.045;
 
     const charismaEff =
       Math.pow(person.skills.charisma, CharismaExponentialFactor) + person.skills.charisma / CharismaLinearFactor;
-    return (100 - charismaEff) / 100;
+    return charismaEff;
   }
 
   getRecruitmentSuccessChance(person: Person): number {
@@ -1155,12 +1155,12 @@ export class Bladeburner {
             break;
           }
           case BladeGeneralActionName.diplomacy: {
-            const eff = this.getDiplomacyEffectiveness(person);
-            this.getCurrentCity().changeChaosByPercentage((1 - eff) * 100);
+            const diplomacyPct = this.getDiplomacyPercentage(person);
+            this.getCurrentCity().changeChaosByPercentage(-diplomacyPct);
             if (this.logging.general) {
               this.log(
                 `${person.whoAmI()}: Diplomacy completed. Chaos levels in the current city fell by ${formatPercent(
-                  1 - eff,
+                  diplomacyPct / 100,
                 )}.`,
               );
             }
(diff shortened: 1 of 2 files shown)


src/Bladeburner/Bladeburner.ts
```diff
@@ -1156,10 +1156,7 @@ export class Bladeburner {
           }
           case BladeGeneralActionName.diplomacy: {
             const eff = this.getDiplomacyEffectiveness(person);
-            this.getCurrentCity().chaos *= eff;
-            if (this.getCurrentCity().chaos < 0) {
-              this.getCurrentCity().chaos = 0;
-            }
+            this.getCurrentCity().changeChaosByPercentage((1 - eff) * 100);
```
Faenre: `eff` is provided as a decimal multiplier intended to be multiplied against the chaos, e.g. `0.98` to apply a 2% change. However, changeChaosByPercentage wants the percentage, not the decimal. So, (1 - eff) * 100 converts the decimal multiplier to the percentage differential.

src/Bladeburner/Bladeburner.ts
```diff
@@ -645,8 +645,8 @@ export class Bladeburner {
       }
     } else if (chance <= 0.7) {
       // Synthoid Riots (+chaos), 20%
-      sourceCity.chaos += 1;
-      sourceCity.chaos *= 1 + getRandomIntInclusive(5, 20) / 100;
+      sourceCity.changeChaosByCount(1);
+      sourceCity.changeChaosByPercentage(getRandomIntInclusive(5, 20));
```
Faenre, line 649: Similar to the diplomacy change, `changeChaosByPercentage` wants percentage points, so we can just take the `getRandomIntInclusive` directly.

src/Bladeburner/Bladeburner.ts
d0sboots: This one is complicated. For both numerical reasons (numbers near zero have precision in floating-point, which is lost if you add/subtract them from one), and also because it makes the calculation much harder to understand, I think this one should be kept as a straight multiply, and then use `clampNumber(chaos, 0)` after.

d0sboots: Actually, hold on one second....

d0sboots: OK, I have another option that might be better. I think getDiplomacyEffectiveness is only used here, in this one place. Also, it internally is calculating a percentage and then converting it to effectiveness. So you could change the function to getDiplomacyPercentage and just have it return the percentage directly, and then feed it directly to changeChaosByPercentage.

If you don't want to do all that, the original option stands. Doing more means doing more testing/verification work.

Faenre: I like the idea of changing it! I'll put that together.

I really don't want the attribute being manually adjusted outside of the intended functions, personally.

Faenre: Implemented!

PR #1589, opened 2024-08-16
NETSCRIPT: Rework disableLog for efficiency

d0sboots:
TL;DR: I'm seeing almost a 3x reduction in script size and a 10x performance increase in script launch + `disableLog("ALL")`. Numbers will be less impressive when actual batching logic is added. Old one could only get to ~1m scripts, new can easily get to 3m.

----

The current implementation was naive; disableLog("ALL") was storing a key for every function, and iterating over a different object to do it (when iterating over objects is quite slow).

The common cases of Bitburner (and especially batching, where efficiency matters most) are either never disabling anything, or disabling "ALL". This optimizes for these two cases, at the expense of slightly more complicated code to deal with the less-common edge cases.

# Tests

Added a new test to verify some of the potential new edge cases

# Performance

`benchmark_disableLog.js`
```js
/** @param {NS} ns */
    const perf = globalThis.performance;
    const startMem = perf.memory.usedJSHeapSize;
    ns.tprintf("Starting memory: %d", startMem);
    let speedMax = 0;
    const totalScripts = Number(ns.args[0]);
    let start = perf.now();
    for (let i = 0; i < totalScripts;) {
        const k = Math.min(i + 10000, totalScripts);
        for (; i < k; i++) {
            if (ns.run("lib/disableLog.js", {temporary: true}) <= 0) {
              throw new Error("Out of memory: " + i);
            }
        }
        await new Promise(setTimeout);
        const end = perf.now();
        const speed = 1e7 / (end - start);
        start = end;
        speedMax = speed > speedMax ? speed : speedMax;
        ns.tprintf("Iter %6d: %.3f/sec", i, speed);
    }
    ns.tprintf("Best: %.3f/sec", speedMax);
    const endMem = perf.memory.usedJSHeapSize;
    ns.tprintf("Ending memory: %d", endMem);
    ns.tprintf("Used ~%.1f bytes/script", (endMem - startMem) / totalScripts);
}
```
`lib/disableLog.js`
```js
/** @param {NS} ns */
export function main(ns) {
  ns.disableLog("ALL");
  // This will never resolve, and doesn't require defining a closure.
  return new Promise(Boolean);
}
```

I ran these tests in my browser, against dev BB before/after the change. Before:
```
run benchmark_disableLog.js 1000000
...
Iter 970000: 5863.039/sec
Iter 980000: 5844.194/sec
Iter 990000: 2724.350/sec
Iter 1000000: 4868.549/sec
Best: 7647.599/sec
Ending memory: 3383249365
Used ~2892.2 bytes/script
```

After:
```
run benchmark_disableLog.js 2000000
...
Iter 1970000: 41580.042/sec
Iter 1980000: 43029.260/sec
Iter 1990000: 44642.857/sec
Iter 2000000: 66137.566/sec
Best: 82576.383/sec
Ending memory: 2643886332
Used ~1083.6 bytes/script
```

### DIFF
--- a/src/Netscript/WorkerScript.ts
+++ b/src/Netscript/WorkerScript.ts
@@ -131,7 +131,7 @@ export class WorkerScript {
   }
 
   shouldLog(fn: string): boolean {
-    return this.disableLogs[fn] == null;
+    return !(this.disableLogs.ALL || this.disableLogs[fn]);
   }
 
   log(func: string, txt: () => string): void {
--- a/src/NetscriptFunctions.ts
+++ b/src/NetscriptFunctions.ts
@@ -495,37 +495,46 @@ export const ns: InternalAPI<NSFull> = {
   },
   disableLog: (ctx) => (_fn) => {
     const fn = helpers.string(ctx, "fn", _fn);
-    if (fn === "ALL") {
-      for (const fn of Object.keys(possibleLogs)) {
-        ctx.workerScript.disableLogs[fn] = true;
-      }
-      helpers.log(ctx, () => `Disabled logging for all functions`);
-    } else if (possibleLogs[fn] === undefined) {
+    if (possibleLogs[fn] === undefined) {
       throw helpers.errorMessage(ctx, `Invalid argument: ${fn}.`);
+    }
+    if (fn === "ALL") {
+      ctx.workerScript.disableLogs = allDisabled;
+      // No need to log here, it's been disabled.
     } else {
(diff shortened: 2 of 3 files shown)


d0sboots:
@yichizhng for review, since you were the one to discover the issue.

yichizhng:
LGTM

I was considering an optimization where you store a default + keys which vary from the default, so that disabling all logs except one is more efficient, but I don't think that use case is important enough to optimize for.

PR #1590, opened 2024-08-16
NETSCRIPT: Moved formatting functions to their own interface

G4mingJon4s:
# Changes
- Removed deprecated function `nFormat`
- Moved all other formatting functions to `ns.format`
- Removed unnecessary dependency `numeral`

`numeral` was only used by `nFormat` which was deprecated since v2.4, I think.

Do note, this is a breaking change. My idea was that this gets merged with v3.0, so there's no need to keep the deprecated functions.


d0sboots:
My opinion on this: Even in a major version bump, I have reservations. There wasn't anything particularly wrong with the old API that we're fixing here, so the benefit is marginal. OTOH, these are common functions used by *lots* of players, so the amount of churn will be large. 

I think a conversion that rewrites player's scripts so that they don't have to do it manually is probably needed to make this worthwhile. There's examples of this for the v1 and v2 transitions; we'll likely need one for the v3 upgrade regardless.

d0sboots:
Looks good to me. You'll want to update package_lock.json by doing npm -i, but that will need to be redone later since this is on hold until 3.0.

G4mingJon4s:
So, I kinda blocked my dev branch with this PR. I will re-open the PR from another branch.😅

PR #1594, opened 2024-08-17
MISC: Improve several things relating to PID lookups

d0sboots:
findRunningScriptByPid needlessly took a "server" argument. This caused there to be a "getRunningScriptByPid" version that did *not*, and it was looping through all servers in order to function, which is needlessly inefficient.

By removing the parameter and the needless inefficient helper method, the following changes:
- Many Netscript functions such as isRunning() and getScript() become faster.
- The terminal "tail" command now works by pid regardless of the current server. Note that "kill" already worked this way.

I also improved the docs around "tail", since the pid argument wasn't in the help.

Testing:
I tried tail, it now works cross-server. `ns.isRunning` still functions properly.


Alpheus (approved):
🫡

PR #1597, opened 2024-08-17
MISC: Follow-up of #1565

catloversg:
I forgot committing this part in #1565.


d0sboots:
Fireball got it in their PR.

PR #1602, opened 2024-08-19
UI: Fix incorrect information given in pserv upgrade error msg.

Nolshine:
### UI: Fix incorrect information given in pserv upgrade error message.

beforeafter

Testing steps (done both before and after change):

1. build and run local dev built
2. give myself infinite money
3. buy an 8GB server from Alpha-Enterprises (can also buy with script)
4. Use script shown below to "upgrade" server to a smaller size

```
/** @param {NS} ns */
  ns.tail();
  ns.upgradePurchasedServer("pserv-0", 2) // smaller than current RAM size of 8
}```

### DIFF
--- a/src/Server/ServerPurchases.ts
+++ b/src/Server/ServerPurchases.ts
@@ -47,7 +47,7 @@ export const getPurchasedServerUpgradeCost = (hostname: string, ram: number): nu
   if (isNaN(ram) || !isPowerOfTwo(ram) || !(Math.sign(ram) === 1))
     throw new Error(`${ram} is not a positive power of 2`);
   if (server.maxRam >= ram)
-    throw new Error(`New RAM (${ram}) should be bigger than server (${hostname}) current RAM (${server.maxRam}).`);
+    throw new Error(`The new ram of '${hostname}' (${ram}) must be bigger than its current ram (${server.maxRam}).`);
   return getPurchaseServerCost(ram) - getPurchaseServerCost(server.maxRam);
 };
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by Faenre** [comment]
Good change,

Could we logically rephrase it? There's a negative comparison and the player's action is at the end rather than the front of the sentence, making it a little difficult to follow.

How about something like:

```js
    throw new Error(`new ram (${ram}) should be greater than '${hostname}' current ram (${server.maxRam})`);
```

src/Server/ServerPurchases.ts
```diff
@@ -47,7 +47,7 @@ export const getPurchasedServerUpgradeCost = (hostname: string, ram: number): nu
   if (isNaN(ram) || !isPowerOfTwo(ram) || !(Math.sign(ram) === 1))
     throw new Error(`${ram} is not a positive power of 2`);
   if (server.maxRam >= ram)
-    throw new Error(`'${hostname}' current ram (${server.maxRam}) is not bigger than new ram (${ram})`);
+    throw new Error(`'${hostname}' current ram (${server.maxRam}) is not smaller than new ram (${ram})`);
```
catloversg: My suggestion:
```js
throw new Error(`The new ram of '${hostname}' (${ram}) must be bigger than its current ram (${server.maxRam}).`);
```
When I skimmed through this change, I thought your change was wrong. I only realize that you are right when I read it the second time. Our old way of phrasing "not + bigger|smaller" trips me up. I think "this is the requirement" is clearer than "this is why it failed".

Nolshine: You're right, I was considering that when I made the change and irrationally decided against it. Made the change now. (ignore 2363eb4 and allow me to pretend it does not exist)

catloversg (approved):

src/Server/ServerPurchases.ts
d0sboots: I was also going to suggest this, so being lazy let others get there first XD

PR #1604, opened 2024-08-19
API: Adjust RAM cost for ns.getPurchasedServers

### LINKED ISSUE #1600: API: Adjust ns.getPurchasedServers() RAM cost
(Version 2.6.2)

## Context

Discord user Nolshine noticed that the cost for `NS.getPurchasedServers()` is disproportional to its value and the other functions in its design space.

| Function | RAM cost |
| --- | --- |
| [NS.getPurchasedServerLimit()](https://github.com/bitburner-official/bitburner-src/blob/dev/markdown/bitburner.ns.getpurchasedserverlimit.md) | 0.05 GB
| [NS.getPurchasedServerMaxRam()](https://github.com/bitburner-official/bitburner-src/blob/dev/markdown/bitburner.ns.getpurchasedservermaxram.md) | 0.05 GB
| [NS.getPurchasedServerUpgradeCost()](https://github.com/bitburner-official/bitburner-src/blob/dev/markdown/bitburner.ns.getpurchasedserverupgradecost.md) | 0.1 GB
| [NS.getPurchasedServerCost()](https://github.com/bitburner-official/bitburner-src/blob/dev/markdown/bitburner.ns.getpurchasedservercost.md) | 0.25 GB
| [NS.getPurchasedServers()](https://github.com/bitburner-official/bitburner-src/blob/dev/markdown/bitburner.ns.getpurchasedservers.md) | **2.25 GB**
| [NS.scan()](https://github.com/bitburner-official/bitburner-src/blob/dev/markdown/bitburner.ns.scan.md) | 0.2 GB

Worth noting, there are many ways to avoid paying this price together, including:
- Scanning `home` and filtering the results (0.2 GB)
- Wrapping `getPurchasedServerUpgradeCost` in a `try { } catch { ; }` block and incrementing a counter (0.1 GB)
- Maintaining a `pservs.txt` list (0 GB)
- Writing data into ports for each server (0 GB)
- Hard-coding your pserv names into the manager (0 GB)

## Problem

Several of the "solutions" require re-architecting code, or having deeper game awareness, to solve what should be a very simple task: getting a list of purchased servers and then moving on. This doesn't seem like it _should_ be a Problem though, here, unless you're getting into deep RAM optimization, which should generally be more important than workers than managers (unless you're getting into wide multi-manager networks, but that's deep into endgame territory).

I feel like it's an antipattern to teach players to avoid using the API, and while optimization and ram-reduction can definitely be rewarding, the steeper price of 2.25 sends the message that "I should do everything myself". Which is fine, rolling your own can often be better. Besides, hard-coding is free, why not do that? But libraries are supposed to be meaningfully convenient. Managers are an opportunity to teach players better coding habits though, like relying on libraries when they are helpful and optimizing after the fact; especially since the problem space for managers is supposed to be about formulaic thinking, not performance-squeezing.

## Solution

I propose the `NS.getPurchasedServers()` function be reduced to 0.5GB, 1 GB, or 1.5 GB. These feel in line with the fact that `NS.scan('home')` costs 0.2 GB and also returns these values, and the only hard part is filtering them (which can be done for about 0.05-0.1 GB). The values I suggested are all more affordable, while still offering room for DIY improvement. Just as importantly, they encourage healthy API awareness without looking at the API as being a player trap.

I'm happy to make the PR if this sounds agreeable and we can agree on a number.

Faenre:
# Adjust RAM cost for ns.getPurchasedServers

The reasoning for this change are laid out in Issue #1600 but to summarize, getPurchasedServers is overpriced compared to the information it brings. It is a 2.25 GB ram function that provides a list of names of servers the player has already purchased. If you want anything else, that costs extra.

It is trivial for a player who knows how to minimize RAM and cares to do so, to get this information for cheap or free:
- parse all servers via `ns.scan` and filter by name (0.2gb)
- maintain a written list of `pservs.txt` (free)
- write a list of names to a port (free)

The function is purely informational and offers nothing new to the player other than an API method of easily accessing information that the user already has available to them. It seems weird to overprice this knowledge to such an extent. However, several players argue that there is benefit in having useful-but-inefficient pitfalls, so this aims to protect that.

---

This PR brings the cost closer to its related informational functions while still being overpriced as a "convenience tax". Further, as an additional signpost to pique players' curiosity, rather than rounding to 1 GB, it ends with 1.05 GB. The idea is that the extra 0.05 will actually be a _more_ noticeable thorn when players are looking at how to minimize their RAM costs, and serving as a signpost that hopefully stands out.

Before:
image
image

After:
image
image

# Linked issues

closes #1600 

# Documentation


d0sboots:
I think you didn't sync beforehand, or something... the merge conflict I'm seeing is surprising so I'm gonna ask you to look into that instead of me trying to deal with it from the web interface.

Faenre:
This branch is being weird, I will close and re-open this with a clean branch.

PR #1608, opened 2024-08-19
IPVGO: Prevent issues caused by resetting the board while the go AI is in flight

ficocelliguy:
Currently, if `ns.go.resetBoardState` is called while the white faction AI is thinking, and player moves are made, the original AI move from before the reset is still seen as alive, preventing other AI from playing. This protection was intended to prevent kicking off multiple AI thinking at once, but stops the game from progressing in this case.

This PR ensures that resetting the board also resets the in-flight AI protections. 
It also makes AI check that the move count on the board is what they expected before attempting to make a move, preventing orphaned AI processes from editing the game state once they resolve.

Original user report that led to this edge case being discovered:
https://discord.com/channels/415207508303544321/1221920484111814828/1275222654965514360


src/Go/boardAnalysis/goAI.ts
```diff
@@ -57,6 +56,11 @@ export function makeAIMove(boardState: BoardState, useOfflineCycles = true): Pro
           return play;
         }
 
+        if (currentMoveCount !== Go.currentGame.previousBoards.length) {
+          console.error("AI move attempted, but the board state has changed.");
```
d0sboots: Let's downgrade this to a warn, it's something to know about but doesn't indicate an error in anyone's (including the game's) code.

src/Go/boardAnalysis/goAI.ts
```diff
@@ -37,12 +37,11 @@ export function makeAIMove(boardState: BoardState, useOfflineCycles = true): Pro
...
-    // Update currentTurnResolver to call Go.nextTurn's resolve function with the last played move's details
-    Go.nextTurn = new Promise((resolve) => (currentTurnResolver = () => resolve(getPreviousMoveDetails())));
+    resetAI();
   }
   // If an AI is in use, find the faction's move in response, and resolve the Go.nextTurn promise once it is found and played.
   else {
+    const currentMoveCount = Go.currentGame.previousBoards.length;
     Go.nextTurn = getMove(boardState, GoColor.white, Go.currentGame.ai, useOfflineCycles).then(
       async (play): Promise<Play> => {
         if (boardState !== Go.currentGame) return play; //Stale game
```
d0sboots: You already have a check here. How does it differ from checking the current move count? Does one check cover more or less than the other?

src/Go/boardAnalysis/goAI.ts
```diff
@@ -57,6 +56,11 @@ export function makeAIMove(boardState: BoardState, useOfflineCycles = true): Pro
           return play;
         }
 
+        if (currentMoveCount !== Go.currentGame.previousBoards.length) {
+          console.error("AI move attempted, but the board state has changed.");
+          return play;
+        }
+
         // Handle AI making a move
         await waitCycle(useOfflineCycles);
```
d0sboots: I think there is a race where you could do the check and then the state could change over the duration of this `waitCycle`, it should be moved to after here.

src/Go/boardAnalysis/goAI.ts
```diff
@@ -82,6 +86,13 @@ export function makeAIMove(boardState: BoardState, useOfflineCycles = true): Pro
   return Go.nextTurn;
 }
 
+export function resetAI(thinking = true) {
+  isAiThinking = thinking;
```
d0sboots, line 99: Because `isAiThinking` is a boolean, it is prone to racy behavior. For instance, it will unconditionally be set to false in the finally block, even for a stale AI promise, leading to possible issues.

I think you're going to have to go back to storing a copy of `nextTurn` here instead of a boolean. That way, you can check if its value is still the same as what you started with, so you only clear it when it is correct to do so.

src/Go/boardAnalysis/goAI.ts
ficocelliguy: The boardState check is just object pointer comparison, it just looks for entirely new games. The move count is to make sure nothing has changed in the current game due to potential race conditions. I'd like to think there aren't any that could trigger this, but it doesn't hurt. I unified the checks made further down after some calls have been awaited, but the move count couldn't have changed yet in this highlighted line.

src/Go/boardAnalysis/goAI.ts
ficocelliguy: I tweaked this to not update the semaphore if it was determined that the current AI process is pointed at a stale game state. which prevents orphaned processes from incorrectly updating that flag.

I tested a number of ways to set isAiThinking to be synched with nextTurn, but I ran into issues trying to keep it up-to-date when balancing the manual-play-vs-a-script flow with the script-vs-ai flow, and decided to identify the stale-game errors instead.

d0sboots:
Sorry I've been really slow on this. I still have a vague unease, but I can't think of any concrete issues, so let's go with this until proven otherwise. XD

PR #1615, opened 2024-08-22
BUGFIX: Wrong skill multipliers in Bladeburner

catloversg:
I found this bug when I checked an int grinding save file of a player. This bug was introduced when we refactored Bladeburner in 2.6.1.

Test code:
```js
/** @param {NS} ns */
  ns.tail();
  ns.clearLog();
  Player.bladeburner.skillPoints = 1e6;
  Player.bladeburner.skills = {};
  ns.bladeburner.upgradeSkill("Reaper", 10);
  ns.bladeburner.upgradeSkill("Evasive System", 10);
}
```

How to reproduce it:
- Load a clean save file. Use the dev menu to add SF7. Expose the Player object.
- Run test code.
- Check Effective Dexterity and Effective Agility in the `Skills` tab.
  - v2.6.0: x1.680
  - v2.6.3: x1.600

v2.6.0:
```js
updateSkillMultipliers(): void {
  this.resetSkillMultipliers();
  for (const skillName of Object.keys(this.skills)) {
    if (Object.hasOwn(this.skills, skillName)) {
      const skill = Skills[skillName];
      if (skill == null) {
        throw new Error("Could not find Skill Object for: " + skillName);
      }
      const level = this.skills[skillName];
      if (level == null || level <= 0) {
        continue;
      } //Not upgraded

      const multiplierNames = Object.keys(this.skillMultipliers);
      for (let i = 0; i < multiplierNames.length; ++i) {
        const multiplierName = multiplierNames[i];
        if (skill.getMultiplier(multiplierName) != null && !isNaN(skill.getMultiplier(multiplierName))) {
          const value = skill.getMultiplier(multiplierName) * level;
          let multiplierValue = 1 + value / 100;
          if (multiplierName === "actionTime") {
            multiplierValue = 1 - value / 100;
          }
          this.skillMultipliers[multiplierName] *= multiplierValue;
        }
      }
    }
  }
}
```

v2.6.3:
```js
updateSkillMultipliers(): void {
  this.skillMultipliers = {};
  for (const skill of Object.values(Skills)) {
    const level = this.getSkillLevel(skill.name);
    if (!level) continue;
    for (const [name, baseMult] of getRecordEntries(skill.mults)) {
      const mult = baseMult * level;
      this.skillMultipliers[name] = clampNumber(this.getSkillMult(name) + mult / 100, 0);
    }
  }
}
```

I also make a small change in the UI:
v2.6.0:

v2.6.3:

That change restores the number format of v2.6.0.

Note: the way that we multiply effects from different skills makes Dexterity and Agility much "better"/"more efficient" than Strength and Defense. This PR is a bugfix one, so I don't touch that part.

### DIFF
--- a/src/Bladeburner/Bladeburner.ts
+++ b/src/Bladeburner/Bladeburner.ts
@@ -751,7 +751,7 @@ export class Bladeburner {
       const level = this.getSkillLevel(skill.name);
       if (!level) continue;
       for (const [name, baseMult] of getRecordEntries(skill.mults)) {
-        const mult = 1 + baseMult * level / 100;
+        const mult = 1 + (baseMult * level) / 100;
         this.skillMultipliers[name] = clampNumber(this.getSkillMult(name) * mult, 0);
       }
     }
--- a/src/Bladeburner/Bladeburner.ts
+++ b/src/Bladeburner/Bladeburner.ts
@@ -254,7 +254,7 @@ export class Bladeburner {
   getSkillMultsDisplay(): string[] {
     const display: string[] = [];
     for (const [multName, mult] of getRecordEntries(this.skillMultipliers)) {
-      display.push(`${multName}: x${formatNumberNoSuffix(mult, 3)}`);
+      display.push(`${multName}: x${formatBigNumber(mult)}`);
     }
     return display;
   }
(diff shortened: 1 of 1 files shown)


d0sboots:
What's the logic on reverting the number format? Those numbers are really big, and I imagine they can get even bigger.

catloversg:
v2.6.0 used `formatBigNumber`, but v2.6.1 onwards used `formatNumberNoSuffix(mult, 3)`. As you can see on the screenshot, currently we show a really long number when it's smaller than 1e33 (check Effective Strength in v2.6.3). IMO, "x2e24" looks more "useful" than "x2,000,000,000,000,000,000,000,000". That long number will only "collapse" to the shorter form when it's greater than 1e33.

d0sboots:
Oh, I misinterpreted your 2.6.0 and 2.6.3 as "before" and "after." OK, all clear now.

PR #1616, opened 2024-08-22
NETSCRIPT: Added new CLI

G4mingJon4s:
# Idea
Using `ns.flags` and `autocomplete` to create a CLI is a bit tedious and not as straight-forward. It is not type-safe and error-prone.
Defining the way the player manually runs the script in a different way allows for an easier setup of a CLI, enables type-safe arguments and options, implicit auto-completion and automatic generation of a help page.

# Changes
Added a new function that can be exported: `cli`. To explain how it works, it's best to look at an example:
```ts
/** @param {CLIBuilder} builder @param {CLIData} data */
export const cli = (builder, data) => builder.addName("PizzaMaker3000")
  .addVersion("4.2.0")
  .addDescription("Make your own pizza from the terminal!")
  .addArgument("type", "string", "What type of pizza do you want?")
  .addOption("c", "boolean", "Would you like to have cheese", true)
  .addAction(async (ns, args, options) => {
    ns.tprint(args[0]); // args is of type '[string]', since the only argument is 'type'
    ns.tprint(options.c); // options.c is of type 'boolean'
  })
  .build();

/** @param {NS} ns */
export function main(ns) {
  ns.tprint("I'm not affected at all!");
}
```
The `cli` function allows the player to create a structured CLI that is independent of their `main` function. Note, this only overrides `main` if the `cli` function is defined and the script was ran by the player from the terminal. The `cli` function is entirely opt-in.

## Current issues and WIP stuff
- Auto-completion is pretty dumb at the moment.
- The `data` input includes information like all servers, which can be used inside the `action`s defined.
- The `--help` and other special flags are highly WIP.
- Sub-commands (`builder.addCommand(...)` are implemented poorly at the moment and should probably allow for deeper-nested ones.


d0sboots:
So, as a high-level design thing: why does this need to be built-in to the game as another entry point? 

Having a builder pattern for arg parsing is extremely common, no issues there. But I don't understand why you couldn't just do `args.parse(ns.args)` at the top of main to invoke arg parsing and `args.autocomplete()` inside the autocomplete function. You could even provide `args.wrap(main, autocomplete)` as a convenience helper to reduce boilerplate to almost zero. 

Basically, I see a lot of upsides to doing this as a user library (fast iteration, uses existing interfaces, ability for users to upgrade on their own schedule) and no downsides or lack of power.

G4mingJon4s:
I hope I understood this correctly: You want something like this?
```ts
const args = /* some defined CLI */

export function main(ns) {
  const parsed = args.parse(ns.args);
}

export function autocomplete() {
  return args.autocomplete();
}
```
While this is another way to do it, I see some small benefits to the approach with new entry points:
- The arguments and options might be different depending on which command was executed. Therefore, the types of these and the action associated may be different. Handling all of this in one place seems weird to me. There's probably a way to get over this with pretty advanced typings and a lot of  conditionals. Simply defining what should happen on the command itself would remove that.
```ts
export function main(ns) {
  const { command, parsedArgs, options } = args.parse(ns.args);
  if (command === args.command[0].name) {
    // Execute first command
  } else if (/* ... */) {
    // and so on
  }

  // Maybe put all functions into one object?
  const actions = {
    [args.command[0].name]: () => {}
  };
  const callback = actions[command] ?? (() => {});
  callback();
}
```
- Basically, my concern is that `main` gets too crowded and can't be followed during the execution easily. Again, there's probably a way to shorten this and make it much more readable. However, for the average player, I think it would be more convenient to split the entry points up a bit.

Of course it would be best if all of this was one user library that is completely configurable and extendable. Chances are that it would even be superior to something implemented by the game. I see this addition more as a QOL feature. You could write all of this yourself, just like formatting functions etc, it's just nice to have if you want a quick and easy solution.

One really great argument by @shyguy1412 was that this is not a difficult problem but something that takes a lot of time to implement.

d0sboots:
> Basically, my concern is that main gets too crowded and can't be followed during the execution easily.

Anything you can do in your cli builder, you can do in main() (and vice versa). You can define the exact same structure as you currently have, and a single line runs it all - I don't see how that is too crowded. 

My *only* point is that you aren't gaining anything special by introducing a new entry point. I'm not making any judgement about the structure of the builder or args parsing itself - that's complicated and filled with trade-offs.

But if you can write this as a user library (and I think you can), that solves a *lot* of logistical challenges. One of the nastiest things about the existing flags "library" is that we can't really make changes to it, since it would break compatibility. This spectre looms over *everything* that is built-in to the game. In contrast, user libraries can be upgraded *when the user chooses to*, which makes breaking changes possible.

G4mingJon4s:
I do agree, it is possible to abstract everything to one line and even as a user library. If everyone could do this, is willing to spend the time implementing it and had their own opinion on how it should behave, then I wouldn't be making this feature.

I would say, giving the player a simple solution they can use if they don't feel like implementing it themselves is a good idea.

Also, one thing this does allow is a script that does different stuff depending on if it was launched by another script or ran from the terminal.
I'm not sure this is possible without separate entry points?
Inputs from the terminal and from another script can't be distinguished without some kind of special argument?

Defining your script to take certain flags and arguments with autocomplete is really handy.
However, correctly passing arguments to that script from other scripts is hard.
In this case, separate entry points solve the problem really nicely by allowing a separation between the two cases.
And if you really want to, you can just fuse all of these back together to one `main` with optional parameters.

shyguy1412:
I dont agree with trying to make one solution that does both the argument parsing and the autocomplete. These features are fairly isolated from each other so I believe there should be two solutions, one for each.

I think we benefit more from two proper solutions rather than one that tries to do both

G4mingJon4s:
> I dont agree with trying to make one solution that does both the argument parsing and the autocomplete.

That's interesting. I thought the way the CLI is defined dictates both at the same time?
I agree that it might be difficult to get everything correct on the first try, but completely isolating these is unintuitive to me.

d0sboots:
> I do agree, it is possible to abstract everything to one line and even as a user library. If everyone could do this, is willing to spend the time implementing it and had their own opinion on how it should behave, then I wouldn't be making this feature.

The great thing about having a community is that we have a place to share user libraries like this, so everyone doesn't have to implement their own.

G4mingJon4s:
Sure, if you want some kind of CLI at the moment, you can use one another player has created.

The thing is, this can be said for everything:
You don't want to solve Go? Use one of the provided scripts from other players.
You don't want to fiddle with the stock market? Use the scripts other players have created.

I'm not saying this is a bad thing. I actually think this sharing is one of the best aspects of the community.
Still, some out of the box solution would remove the need to depend on others.
Additionally, it would define some default others can relate to or even build on top of.

d0sboots:
> Still, some out of the box solution would remove the need to depend on others.

People act like there's some magical difference between the code in the game and the code other players write, but there really isn't. The only difference is that the game's code has been "blessed" in some sense, by being accepted as a PR. But it was still written (and thus you are still depending) on others.

I think that deep down, people understand this: If the game provided functions for automating all of gangs as a black box (for instance), I think a number of people would refuse to use them, in favor of writing their own. Similarly, people who are otherwise strict about writing all their own code will use box.js, or my script that keeps the browser tab awake.

> Additionally, it would define some default others can relate to or even build on top of.

You can't really build on top of the game's code, because it's unmodifiable. User libraries are much better in this regard, because you can fork and tweak them to sit your needs.

Because of those considerations, when I think about "should this be part of the game or not" the first question is usually "does it require access to data or features that aren't available to a regular script?" If the answer is no (like here), I'm already leaning towards "user script." Other questions are "is this a common need for most players?" (Several, but not most, so neutral there.) "Is this something that players will have different opinions as to how it should be done?" (Definitely) "Does this introduce API surface, and if so is there a risk the API will need to change in the future?" (Yes, high risk) 

I'd be happy to bless a particular CLI library as "recommended", but for a host of reasons I don't think it's a good fit to include in the game.

G4mingJon4s:
Right, I totally agree that there are number of reasons why user libraries are better here. Since there isn't really a way to distinguish between manually running a script or executing it via other scripts, I'll implement something like that instead.

PR #1620, opened 2024-08-23
BLADE: Fix population estimate percent growth when estimate is 0

Faenre:
# BLADE: Fix population estimate percent growth when estimate is 0

In Bladeburners, there are 4 methods to improve the player's population estimate, 3 of which are percentage-based. However, if the estimate is at 0.00000 population, then Field Analysis, Investigation, and Undercover Operation are all unable to improve estimate counts, even if the actual estimate recovers.

This PR clamps the estimate change to a minimum of +1 in the circumstance that the actual population is a positive number and the estimated population is less. 

This does NOT change the percentage calculations, it only clamps that a "too low" population estimate increase is guaranteed to discover at least 1.00 new synthoid.

Tested by reducing Sector 12 to 0 and then letting passive population growth kick in during bonus time.

Before: 
image

After:
image

# Bug fix

### DIFF
--- a/src/Bladeburner/City.ts
+++ b/src/Bladeburner/City.ts
@@ -35,9 +35,9 @@ export class City {
   improvePopulationEstimateByCount(n: number): void {
     n = clampInteger(n, 0);
     const diff = Math.abs(this.popEst - this.pop);
-    // Chgnge would overshoot actual population -> make estimate accurate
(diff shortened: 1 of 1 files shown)


d0sboots:
I have an alternate plan here, but I want to grab the comment fixes and the min is a step in the right direction, so I'm gonna take this as-is.

PR #1624, opened 2024-08-24
BLADE: Use "grow" semantics for population estimate

d0sboots:
This changes the way percantage-based population estimates work, so that they have a similar mechanism to grow(). This allows natural recovery from 0, and also falling to 0 in finite steps.

This also changes the formula for reducing estimates to `est / (1 + p)` instead of `est * (1 - p)`. The latter becomes overpowered at large p, we've seen this formula issue elsewhere (such as BN modifiers).

## Testing

I did some basic Bladeburner testing, and populate estimates seem to still work.


d0sboots:
@Faenre This is what I was thinking of. (It seemed a bit complicated for a comment.)

d0sboots:
This one fell off discussion, but there were some thumbs up and no objections in discord, so I'm gonna merge it.

Issue #1621, opened 2024-08-23
API: Make string and ID enums more accessible to users


Faenre:
_[This is to encourage Programming Best Practices, as well as bring us closer to offering localization in the future. It also helps enable us to change strings once they've been in the wild for some time.]_

We have a bunch of constants made available in the API, some of which are available in `ns.enums`, many of which are locked behind functions like `getSkillNames()`, and many of which have RAM costs associated, for what is essentially just getting a list of API-recognized strings for a given function. They don't provide logic, and their scattered use encourages people to just hardcode names and numbers. This also makes it difficult in case we ever want to change strings for any reason at all. Because there are not player-facing methods to avoid problems, any player-facing strings and IDs become permanent fixtures that cannot be later changed. I'd like to, well, _fix that._

After a [brief discussion in Discord](https://discord.com/channels/415207508303544321/415207923506216971/1276595579786498269), I've been given the clear to reduce the string-constant functions to 0 RAM:

- [Infiltration.getPossibleLocations()](https://github.com/bitburner-official/bitburner-src/blob/dev/markdown/bitburner.infiltration.getpossiblelocations.md) down from 5 GB
- [Gang.getEquipmentNames()](https://github.com/bitburner-official/bitburner-src/blob/dev/markdown/bitburner.gang.getequipmentnames.md) down from 1 GB
- [Gang.getTaskNames()](https://github.com/bitburner-official/bitburner-src/blob/dev/markdown/bitburner.gang.gettasknames.md) down from 1 GB
- [CodingContract.getContractTypes()](https://github.com/bitburner-official/bitburner-src/blob/dev/markdown/bitburner.codingcontract.getcontracttypes.md) down from 2 GB
- [Bladeburner.getBlackOpNames()](https://github.com/bitburner-official/bitburner-src/blob/dev/markdown/bitburner.bladeburner.getblackopnames.md) down from 0.4 GB
- [Bladeburner.getContractNames()](https://github.com/bitburner-official/bitburner-src/blob/dev/markdown/bitburner.bladeburner.getcontractnames.md) down from 0.4 GB
- [Bladeburner.getGeneralActionNames()](https://github.com/bitburner-official/bitburner-src/blob/dev/markdown/bitburner.bladeburner.getgeneralactionnames.md) down from 0.4 GB
[...]

PR #1628, opened 2024-08-25
API: Reduce NS costs for static information functions to 0

Faenre:
# API: Reduce NS costs for static information functions to 0

This PR implements the top change outlined in #1621 by reducing NS costs for functions which serve only serve to provide the user with constant string values.

Currently, the workflow for these functions is "run it once, copy the data, save as a hardcoded array/hash, never use it again." This is a type of antipattern behavior.

Turning the data into useful attribute-constants will be a follow-up PR at a later date; today, the goal is simply to reduce a barrier-of-entry by zeroing-out these costs. This fulfills the top portion of the linked issue.

# Documentation

image

### DIFF
--- a/markdown/bitburner.bladeburner.getblackopnames.md
+++ b/markdown/bitburner.bladeburner.getblackopnames.md
@@ -19,7 +19,7 @@ Array of strings containing the names of all Bladeburner Black Ops.
 
 ## Remarks
 
-RAM cost: 0.4 GB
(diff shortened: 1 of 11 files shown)


Faenre:
If I missed any, let me know so we can round them up in one PR

d0sboots:
I agree, and this was remarkably uncontroversial.

PR #1635, opened 2024-08-27
NETSCRIPT: Moved formatting functions to their own interface

G4mingJon4s:
# Changes
- Removed deprecated function `nFormat`
- Moved all other formatting functions to `ns.format`
- Removed unnecessary dependency numeral
- Added a conversion from the old API call to the new one via the interface
- Added a API break message for v3.0

`numeral` was only used by `nFormat` which was deprecated since v2.4.

(Re-opened from #1590)


G4mingJon4s:
I rebased this PR to the latest commit and removed the last occurrences of the old naming. There are some (most likely) unintentional changes to `codingcontractenumtype.md`. I should probably remove those from this PR, right?

d0sboots:
The cct changes are fine, it's because of a submitted PR where a final change wasn't paired with `npm run doc`

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -7991,79 +8065,6 @@ export interface NS {
...
-   * Format a number.
-   * @remarks
-   * RAM cost: 0 GB
-   *
-   * Converts a number into a numeric string with the specified format options.
-   * This is the same function that the game itself uses to display numbers. The format also depends on the Numeric
-   * Display settings (all options on the "Numeric Display" options page)
-   * To format ram or percentages, see {@link NS.formatRam | formatRam} and {@link NS.formatPercent | formatPercent}.
-   *
-   * This function has some quirky undocumented behaviors. This is a non-exhaustive list of those behaviors:
```
d0sboots, line 8004: You dropped this part. It makes me worried that you're missing other stuff that was updated in the docs.

d0sboots (commented):
This is a major one, but auto-migration should be able the handle it, so I'm not too concerned. Make sure the docs are up-to-date, and it should be good to go.

G4mingJon4s:
So, I went over the changes to any formatting function that happened after I first made this PR. I think that behavior notice was the only thing I missed. Looking at the diff, this should be good to go.

PR #1636, opened 2024-08-27
NETSCRIPT: Added various utility info

G4mingJon4s:
In regard to #1616.

# Changes
- Added `ns.filename`, giving direct access to the filename of the script. This info was pretty much hidden before. The only way was to find the running script via the process info. There shouldn't be a RAM cost for this basic info, so it is a field on the API now.
- Added `ns.parent` giving access to the PID of the parent script that launched the script. This value will be 0, if it was launched through the terminal.
- Added `AutcompleteData.filename`, `AutocompleteData.hostname` and `AutocompleteData.processes`. This information allows a wider range of autocompletion suggestions or specific suggestions based on the filename or host.


shyguy1412:
This looks great. I think it would also be cool to have a function to get the parent of another script. That way if a child script spawns more children those can traverse parents to find the root process

G4mingJon4s:
I can look into that. Should the field remain then? The direct parent would cost no RAM, but getting the parent of any other script could cost a small amount?

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -9496,6 +9509,10 @@ interface AutocompleteData {
   txts: string[];
   enums: NSEnums;
   flags(schema: [string, string | number | boolean | string[]][]): { [key: string]: ScriptArg | string[] };
+  hostname: string;
+  filename: string;
+  /** The processes running on the host */
+  processes: ProcessInfo[];
```
d0sboots, line 9524: This is a level of dynamic data I'm not 100% sure about exposing to the RAM-less autocomplete. At minimum, it gives you a method to bypass RAM for process checking by tab-completing a script repeatedly.

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -9496,6 +9509,10 @@ interface AutocompleteData {
   txts: string[];
   enums: NSEnums;
   flags(schema: [string, string | number | boolean | string[]][]): { [key: string]: ScriptArg | string[] };
+  hostname: string;
+  filename: string;
```
d0sboots, line 9522: These need comments. (Really, they all do, but I'll only insist for new fields.)

d0sboots (commented):
I'm OK with the general idea, but I think the architecture should be different. 

I don't like all the scattered fields being added to the top-level ns object; it feels like they belong in an object. It turns out there is *already* an object that has exactly these fields (and more, that people might want to add in the future): RunningScript (the API version).

I propose we add parent to RunningScript (both versions), which solves the suggestion of other scripts being able to look it up. Then we add a 0-cost `ns.self()` that returns the current running script.

Justification: Everything that's in there is stuff you could pass in args or otherwise track for yourself for 0 cost. The biggest buff is the free access to money trackers, but those are more of a convenience than an actual gameplay benefit.

src/ScriptEditor/NetscriptDefinitions.d.ts
G4mingJon4s: Yes, you can dodge the RAM cost this way, but you can already do that to get all servers and scripts. Since the server data is mostly static it is even easier than using this to get the process data for free.

My idea with providing this data is to allow suggestions like pids so you can do `run script.js --tracking scriptPid`. This isn't really possible at the moment.
I can remove it if it really bothers you, but to me this would also mean that server and script data can't be provided either.

G4mingJon4s:
To recap, currently implemented:
- `ns.self()`, returning the `RunningScript` without any cost.
- `RunningScript` has a `parent` field with the pid of the parent script. This allows not only getting the own parent, but also the parent of other scripts.
- The `AutcompleteData` has the new fields `hostname`, `filename` and `processes`. All fields of the `AutocompleteData` now have a description.

src/ScriptEditor/NetscriptDefinitions.d.ts
d0sboots: It being limited to the current host mitigates many of my concerns. I think it's probably OK.

src/NetscriptFunctions.ts
```diff
@@ -1842,7 +1847,11 @@ setRemovedFunctions(ns, {
 });
 
 export function NetscriptFunctions(ws: WorkerScript): NSFull {
-  return NSProxy(ws, ns, [], { args: ws.args.slice(), pid: ws.pid, enums });
+  return NSProxy(ws, ns, [], {
```
d0sboots: Is this what npm format wants? It feels like a partial revert.

src/Script/RunningScript.ts
```diff
@@ -58,6 +58,9 @@ export class RunningScript {
   // Process ID. Must be an integer and equals the PID of corresponding WorkerScript
   pid = -1;
 
+  // Process ID of the parent process. Defaults to 0.
```
d0sboots:
```suggestion
  // Process ID of the parent process. 0 indicates no parent (such as run from terminal).
```
It is self-evident that the default is 0. XD

d0sboots (commented):
Looks like you have to fix the test data slightly.

src/NetscriptFunctions.ts
G4mingJon4s: I don't know why formatting didn't revert back to the one line there. I simply forced it back to one line.

d0sboots:
You've still got a failing test to deal with. (It should be an easy fix)

G4mingJon4s:
Not sure if I fixed the tests correctly. At least it doesn't fail anymore.

d0sboots:
Sorry for the repeated back-and-forth, I was on autopilot and wasn't thinking about this properly.

Specifically, the tests were doing their job right, but I wasn't thinking about what the failure meant. It was complaining because you needed to adjust the test for a new saved field on `RunningScript`, but the field you added shouldn't be saved at all. (Since pids aren't saved across reloads, and so on a reload scripts won't have parents.)

This, the proper fix is to add `parent` to the list of ignored keys at the bottom of `RunningScript.ts`. Once you do that, the test should go back to failing (until you revert the most recent change).

G4mingJon4s:
Ah, that makes more sense. Thanks a ton for the help!

PR #1637, opened 2024-08-27
CONTRACT: Clarify applicibility of wiki entry in Hamming contracts.

### LINKED ISSUE #1610: CODINGCONTRACT: HammingCode: Clarify difference between Wikipedia implementaion and in-game implementaion
The Coding Contract description for HammingCode mentions [wikipedia article](https://wikipedia.org/wiki/Hamming_code) as refernece.
However, in-game uses slighty different coding scheme than wikipedia article.
Namely, we add one extra parity bit at the very beginning of binary string, compared to wikipedia article.

It seems that in-game scheme is also used widely as it has its own advantage (i.e. SECDED).
However, for players, they have hard time when their implementaion which follows wikipedia article doesn't work well.

I think clarifying difference on scheme between in-game and wikipedia article will be helpful.

gmcew:
closes #1610 
Wording changes here are to emphasise the "extended" part of the contract structure. In parallel, there has been positive feedback on the 3b1b video's presence on discord, and I think the wiki entry still has some value in the very entry-level understanding of what a Hamming code is.

It's draft since I'd appreciate any comments/suggestions on collating further improvements to limit follow-on corrections.

### DIFF
--- a/src/Go/ui/GoTutorialChallenge.tsx
+++ b/src/Go/ui/GoTutorialChallenge.tsx
@@ -1,4 +1,4 @@
-import React, { useRef, useState } from "react";
+import React, { useRef, useState, useEffect, useCallback } from "react";
 import { Typography, Button } from "@mui/material";
 
 import { GoColor, GoValidity, ToastVariant } from "@enums";
@@ -65,12 +65,17 @@ export function GoTutorialChallenge({
     }
   };
 
-  const reset = () => {
+  const reset = useCallback(() => {
     stateRef.current = getStateCopy(state);
     stateRef.current.previousBoards = [];
     setDisplayText(description);
     setShowReset(false);
-  };
+  }, [state, description]);
+
+  // Ensure that the challenge is reset on unmount / mount
+  useEffect(() => {
+    reset();
+  }, [reset]);
 
   return (
     <div>


d0sboots:
Seems good to me. Probably marginally better, but these contracts will always have issues. (Fundamentally, the concepts are more complicated than the code here.)

d0sboots:
I'm happy with this as is; comment or do something if you're happy to have it merged, otherwise I'll forget about it.

bupjae:
> An 'extended Hamming code' has an additional parity bit to enhance error detection and correction.

To be more precise, the additional parity bit of "extended" hamming code enhances the error detection when compared to plain hamming code, but not for correction.

* plain hamming code can detect and correct 1-bit error, but cannot detect 2-bit or more error.
* extended hamming code can still correct only 1-bit error, but able to detect up to 2-bit error. (so-called SECDED)

As our contract required to correct up to 1-bit error, this won't be affected for implementing solver. However, as we are trying to clarify the problem description, it would be good to be more precise.

gmcew:
> I'm happy with this as is; comment or do something if you're happy to have it merged, otherwise I'll forget about it.

Thanks, I'm happy to have it merged. I completely agree this is not likely to be the last revision of this description; it is a complex concept to distill down effectively.

d0sboots:
Shoot, I completely missed that this one was ready to go a while ago.

It looks like someone touched the description here slightly in the meantime (it really is an impossible circle to square), but if you want to resolve the changes I'll merge this.

PR #1638, opened 2024-08-27
UI: Simplify Corps LimitMaterial Buttons – Remove Non-Functional and Rename Functional to LimitProduction

cmsautter:
## Documentation

Consider removing the "Limit Material" button from materials that aren't being produced and renaming it to "Limit Production" to match the existing button for products.
Corps can be a bit overwhelming for new players, and having buttons that don't do anything or are mislabeled doesn't help. I understand that the production limit feature isn't working as intended, but a good first step might be to remove the button where it’s not needed.

Tobacco:
grafik

Agriculture:
grafik

Just to make my point a bit clearer. This is the UI element that opens, regardless of whether you click the "Limit Material" or the "Limit Production" Button.
grafik

Also, it looks like there's no trace of a real "Limit Material" feature in the code. As a player, when I see that button, I'd expect it to limit the space in the warehouse that could be filled by that material - whether through buying, producing, or importing. It would be a cool feature to have, but as it stands, it's not in the game.

That's why I’m suggesting we rename or remove the existing buttons. If we decide to introduce a real "Limit Material" feature in the future, we can always create new buttons for it then.

This PR removes nonfunctional buttons and renames the functional ones to "Limit Production":
grafik

### DIFF
--- a/src/Corporation/ui/MaterialElem.tsx
+++ b/src/Corporation/ui/MaterialElem.tsx
@@ -74,10 +74,11 @@ export function MaterialElem(props: IMaterialProps): React.ReactElement {
   }
 
   // Limit Production button
-  let limitMaterialButtonText = "Limit Material";
+  let limitProductionButtonText = "Limit Production";
   if (mat.productionLimit !== null) {
-    limitMaterialButtonText += " (" + formatCorpStat(mat.productionLimit) + ")";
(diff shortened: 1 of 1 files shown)


cmsautter:
CatLover had an even better idea in my opinion:
https://discord.com/channels/415207508303544321/415213435974975508/1278217635984834592

d0sboots:
Yup, I'd rather fix the issue for real.

PR #1639, opened 2024-08-28
UI: Synchronize Corps Material Production Limits

cmsautter:
## Documentation

Currently, the material production limits in Corps are not functioning correctly when a division produces multiple materials, such as ore and minerals in mining. If you set the production limit for the first material, both materials' production will be limited. However, if you set the limit for the second material, no production limits will be applied. This issue arises because the production of these materials is currently synchronized, meaning that you cannot produce one without producing the other. Whether this synchronization is intended game design is not the focus of this PR.

Additionally, a separate issue in division.ts causes only the production limit for the first output material to be checked and applied.

This PR addresses these issues by synchronizing the production limits for all materials that a division produces. This change will provide players with accurate feedback, indicating that setting a limit for one material will affect the others, and ensure that the displayed production limits match the applied limits.

Note: The term "set production limit" refers to the amount being set using the "Limit Material" button. This button controls the production limits in the city, not the material quantities in the warehouse. A different PR addresses this separate ssue: https://github.com/bitburner-official/bitburner-src/pull/1638

## Screenshot visualising the solution:

grafik

**Before:** Both limits could be set separately, but only one limit had an effect on both materials' production.

**After:** The limits are synchronized; changing the limit for one material now updates the limit for the other material as well.

### DIFF
--- a/src/Corporation/Actions.ts
+++ b/src/Corporation/Actions.ts
@@ -613,11 +613,19 @@ export function limitProductProduction(product: Product, cityName: CityName, qua
   }
 }
 
-export function limitMaterialProduction(material: Material, quantity: number): void {
-  if (quantity < 0 || isNaN(quantity)) {
-    material.productionLimit = null;
-  } else {
-    material.productionLimit = quantity;
+export function limitMaterialProduction(
+  material: Material,
+  division: Division,
+  cityName: CityName,
+  quantity: number,
+): void {
+  const limit = quantity < 0 || isNaN(quantity) ? null : quantity;
+  material.productionLimit = limit;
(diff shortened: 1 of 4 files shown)


cmsautter:
CatLover had an even better idea in my opinion:
https://discord.com/channels/415207508303544321/415213435974975508/1278217635984834592

d0sboots:
Do you want to close this, or take a stab at implementing what they talked about?

cmsautter:
We can close this now. I just wanted to make sure you're on the same page with what we were planning and that there are no major objections. I'll create a new PR for this combined idea, taking it a small step further.

PR #1641, opened 2024-08-30
API: Add script access to autolink

### LINKED ISSUE #1640: API Feature Request: After access to AutoLink.exe allow script use of links
Add an API method to allow turning a server hostname into a `Link`.

Essentially something like `ns.serverLink(hostname: string): Link` could be added and usable in scripts after creation/purchase of `AutoLink.exe`. Then upon printing to terminal that hostname would be clickable to connect to as would be expected.

This feature would be super useful for any script that returns server names to the terminal in general. Because of it's usefulness I do think it should be a function that comes with a RAM cost rather than being free. Potentially `0.05` similar to other EXEs that are exposed as script functions like nuke and such.

mallardduck:
# API: Add script access to autolink

Formatted as such:
CATEGORY: Change Description

# Linked issues

closes https://github.com/bitburner-official/bitburner-src/issues/1640

# Documentation

-TODO: run `npm run doc` 

# Bug fix

- Include how it was tested: not yet...still learning that part (PR made via github UI)
- Include screenshot / gif (if possible)

Make sure you run `npm run format` and `npm run lint` before pushing.

### DIFF
--- a/src/Netscript/RamCostGenerator.ts
+++ b/src/Netscript/RamCostGenerator.ts
@@ -511,6 +511,7 @@ export const RamCosts: RamCostTree<NSFull> = {
   getScriptLogs: 0,
   hasTorRouter: 0.05,
   nuke: RamCostConstants.PortProgram,
+  autoLink: RamCostConstants.PortProgram,
   brutessh: RamCostConstants.PortProgram,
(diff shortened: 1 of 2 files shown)


mallardduck:
I will get back to this and properly pull the files locally to build and correct the PR this weekend.

PR #1643, opened 2024-09-01
DOCUMENTATION: Fix mangled NS API TSDoc

catloversg:
This PR makes 2 changes:

Change 1:
In #1566, I replaced `<pre lang="javascript">` with triple ticks to fix warnings of `api-extractor`, but `api-documenter` v7.23.x cannot render the table properly with the usage of triple ticks (check `markdown\bitburner.go.md`). In this PR, I revert that change. The warning of `api-extractor` is "harmless", so it's not a big deal now. We can find another way to "fix" it later.

Change 2:
#1642 copies comments from `src\BitNode\BitNodeMultipliers.ts` to `src\ScriptEditor\NetscriptDefinitions.d.ts`. In this PR, I made some editorial changes in `BitNodeMultipliers`:
- Capitalization, space, etc.
- `Influences the respect gain and money gain of your gang.` explains the modifier better than `Reduces gangs earning.`

#1642 also forgot running `npm run doc`. We should notify contributors when they forget doing that.


d0sboots:
#1642 was by a new contributor, I opted to take it as -is and let someone more familiar deal with running `npm doc`.

PR #1645, opened 2024-09-02
JEST: Test faster by removing barrels

cmsautter:
On request from FicocelliGuy in development: https://discord.com/channels/415207508303544321/459097632896188436/1280207617062469662 

This is just fiddling arround with [babel-plugin-transform-barrels](https://github.com/cmfrydos/babel-plugin-transform-barrels)
following this guide: https://dev.to/fogel/potential-issues-with-barrel-files-in-jest-1nkl
it currently has many problems:
- I struggle to put babelTransform.js anywhere else than the root directory. (Edit: fixed)
- Tests now only work from the root-directory (working directory cannot be /src or /test) (Edit: only for jest command)
- Using the debugger in Visual Studio Code is broken now. (Could later not be verified)

So this needs some work to fix these issues first.
I suspect the first two being related to false pathing, and the second one due to changing line numbering.

It's faster though:
Here are some test-performance-tests running a 5950X with 32GB (18GB free).
I excluded test\jest\Netscript\Bladeburner.test.ts since it's running stupid amounts of testcases (10 000 000), which just take a lot of time on their own. So to meassure jest startup time, which is the real problem here, I excluded it.

Before (all tests) : 50.644s / 15GB usage
grafik

After (all tests): 15.721s / 10GB usage
grafik

Ram-Screenshot
grafik

After (all tests with --maxWorkers=4 param): 16.18 s / 4GB usage
grafik
**This might be the ideal config for people with 16GB or less RAM.**

Before (one test): 20.835s (Corporation.test.ts)
grafik

After (one test): 6.803s (Corporation.test.ts)
grafik

Further ToDo:
- [x] check whether opening a subbash in root directory when calling `npm run test` will fix the plugins pathing issues.
(That's actually package.jsons default behaviour)
- [x] investigate the ast node generation and why the source maps are not generated properly
- [x] ~~try make the plugin work with Webpack to potentially reduce output bundle size and improve performance~~ No need.


catloversg:
You must not exclude any test. If you want to tune a test, you have to make a separate PR.

cmsautter:
Argh, this needs way more work than expected:
I've noticed there are some cache-shenanigans happening, and the current state of the PR fails to run with
`jest --no-cache`. I'll try to find the problem.

EDIT: Ok, disabling the caches helped a lot for debugging. Now the tests run fine and don't show as many unusual behaviours as before.

package.json
```diff
@@ -25,6 +25,7 @@
     "acorn-jsx-walk": "^2.0.0",
     "acorn-walk": "^8.3.2",
     "arg": "^5.0.2",
+    "babel-plugin-transform-barrels": "github:cmfrydos/babel-plugin-transform-barrels",
```
d0sboots: This should be a dev dependency. That should also fix up a lot of unneeded changes in package.lock.

jest.config.js
```diff
@@ -2,7 +2,8 @@ module.exports = {
   roots: ["<rootDir>/src/", "<rootDir>/test/"],
   moduleFileExtensions: ["ts", "tsx", "js", "jsx"],
   transform: {
-    "^.+\\.(js|jsx|ts|tsx)$": "babel-jest",
+    //"^.+\\.(ts|tsx)$": "babel-jest",
```
d0sboots: This is ok when developing, but commented-out lines shouldn't be committed.

babelTransform.js
```diff
@@ -0,0 +1,18 @@
...
+
+module.exports = {
+  process(sourceText, sourcePath, options) {
+    const alias = options?.config?.moduleNameMapper;
+    const extensions = options.config.moduleFileExtensions;
+    const babelTransformer = babelJest.createTransformer({
+      presets: [["@babel/preset-react"], ["@babel/preset-env"], ["@babel/preset-typescript"]],
+      plugins: [
+        ["transform-barrels", { executorName: "jest", alias: alias, extensions: extensions, isCacheEnabled: false }],
+      ], // set to true when all else worked
```
d0sboots: I don't understand this comment; what does it mean?

d0sboots (commented):
Sorry I've been slow, but I'm back on the case now.

package.json
cmsautter: Didn't know the meaning of dev packages until I googled it yesterday. But good catch, would probably have forgotten to fix it.

jest.config.js
cmsautter: True. During Testing it was quite helpful, since this is the best place to change it back and forth between the old transformer and the new.

babelTransform.js
cmsautter: I've found out about the isCacheEnabled parameter, and wanted to leave me a note here to try to enable it when I fixed the other problems. During testing, caching was not helping at all, since it wouldn't delete the cache when you've made changes to the configuration. 

---
I did try this parameter in the meantime though and it didn't yield any performance gains, but would save some temporary files. So we shouldn't enable this. Jest-Caching, enabled by default, on the other hand is giving quite a boost, and would probably also circumvent any plugin caching.

d0sboots:
Of course, the biggest obstacle to merging this right now is that the automated tests don't run XD

cmsautter:
ah, the tests failing is not a good sign. It worked at my computer, even without caching. I'll investigate this though tomorrow.

Also, I couldn't really fix the problem that the jest command now fails at any directory but the root dir. I've tried, and removed many process.cwd calls, or ones to path.transform("."). Calling npm run test works everywhere though.

The plugin could potentially also reduce the webpack size and improve general game performance. I wanted to test this, but somehow this also introduces new errors I didn't worked through.

The last point to check would be to assure the source-maps are created correctly such that debugging testruns would be possible (again).

@d0sboots 
And would you trust this plugin enough, including my changes to it, to let it transform all our tests and codebase? It may also introduce failing tests in the future, just because some obscure js syntax is throwing it off. For me, these drawbacks aren't really worth it for the 2-3x speedup, atleast for others who did not experiment with it as I did the last days and knows potential error sources.

Edit: Maybe having a test:fast command could mitigate those concerns a bit.

d0sboots:
I'm not going to trust some random thing that isn't even published as an npm module (which anyone can do) but instead is just someone's random github blob, to operate as part of our release process. There's also significantly less need; the speed of building is not a bottleneck there, and all that matters is the final produced output. By contrast, running tests is much more about speed, and the size of the output is not a concern since it will be thrown away after the tests run.

It *is* important that all the tests be able to run, but if there's a minor issue we can work around it. That's not the issue that I saw in the test logs though, it seemed like something much more systematic (but hopefully easy to fix).

cmsautter:
To be precise, it is a npm module, but I needed to fork it to make it compatible with this codebase. 
(Maybe you should fork it from me after all issues are resolved, so that I could not install any backdoor on your testing setup 😉. Edit: Actually, in package.json, you can specify the hash of a specific commit you'd like to install, which might be the best solution)

It is a rather "unknown" babel plugin though, looking at the amount of issues of the original repo, which explains the troubles it has with a larger codebase covering some more unlikely edge cases.

cmsautter:
The tests should pass now. I accidentally pushed [babel-plugin-transform-barrels](https://github.com/cmfrydos/babel-plugin-transform-barrels) in a broken state earlier. This likely happened because I first edited the plugin files directly and only updated the fork afterward.

Now, I just need to fix the AST source-map generation for this plugin.

One last point to clarify: should I modify the transformer plugin to only 'activate' on the `npm run test:fast` command and leave the npm test command as it was? Or would it be acceptable to make this faster version the new standard test procedure, with the drawback that tests may unexpectedly fail when, for example, a new module is installed that uses additional unused JS import/export syntax? In that case, you'd encounter an error similar to the one in the last failed test: https://github.com/bitburner-official/bitburner-src/pull/1645/commits/63d696fadb185ddd6947554284c59e0f5013527f

d0sboots:
> One last point to clarify: should I modify the transformer plugin to only 'activate' on the `npm run test:fast` command and leave the npm test command as it was? Or would it be acceptable to make this faster version the new standard test procedure, with the drawback that tests may unexpectedly fail when, for example, a new module is installed that uses additional unused JS import/export syntax?

This should work with `npm test`, and we should not introduce a new `npm test:fast`. There needs to be only one way to run tests, and it needs to be able to run all the tests. But it's fine if it introduces (reasonable) restrictions on the ways we can write new tests.

d0sboots:
As a separate note, the success of this makes me think we should look into refactoring for removal of barrel file usage. I don't think we have much of it in the BB codebase directly, but MUI has a lot of it and it's completely optional - we could import only what we use relatively easily.

cmsautter:
That's a good point. We should analyse the plugins output and compare it with the current implementation. I also prefer a more verbose, but detailed, import section over  just importing a wildcard. Maybe the plugin already transformed the import sections in the files in an appropriate way and there isn't much more manual work to do. I'll come back here, when I tested that and fixed the remaining source-mapping issue (shouldn't be too hard).

cmsautter:
I've counted them! Those are all the files / import statements that use barrel import in bb:
Total .js/.ts/.tsx/.jsx files: 298
Total imports: 623

So this would be another +- 600 line refactor xD I can make a pr with all of them if you wish, but we can also move this to a later stage.

d0sboots:
Oh, definitely later stage XD

I'd like to focus on landing this first, then we can look at refactoring to make it maybe not needed. I'm this case, if it's a strictly mechanical refactor that is only touching import lines, that's pretty easy to handle even if it is a lot of lines/files.

d0sboots:
(also - this is still marked as draft. I'd there anything left to do here still, in your opinion?)

cmsautter:
Ok, then let's skip that for now.

I believe I've fixed the AST source map generation, but I'd like to verify this. Previously, I noticed some strange debugging behavior, such as breakpoints not stopping at the correct line. I'd like to replicate this with the last version to see if my recent changes have indeed fixed the issue. If so, we could go live with the minor drawback, as I previously mentioned, that the `jest` command will only work in the root directory and not in any subdirectories. However, `npm run test` will continue to work from anywhere, as before.

cmsautter:
Also, I'd like to cleanup my last push to the plugin-fork, and then we can set the version to this specific commit, such that it cannot be changed. I've already noticed the maintainer of the plugin about our issues with it, and hopefully my fixes will somehow go into their official code in a more robust way.

cmsautter:
I'll undraft this as soon as I believe this can be merged with no further edits.

cmsautter:
Somehow, I couldn't replicate the debugging issues I had initially. I had struggled a lot with the caching at that time, and the disguised issues might have interfered with the debugging process. I read that including line information in the AST generation is recommended, which wasn't the case with this plugin. I've added this now, but I'm not sure if it was actually necessary.

I've pinned the version to the latest commit I made to the plugin. This is ready for merging from my end.

For anyone testing this, I recommend starting with the `jest --clear-cache` command to ensure that the new, faster transformer is used. Then, based on your system's memory:
- If you have 16GB of RAM or less, run ` npm run test -- -- --maxWorkers=4`.
- If memory is not a constraint, simply use `npm run test`.

The initial test run may be slow as Jest rebuilds its cache, but subsequent runs should be way faster.

d0sboots:
Tests pass, looks good

PR #1646, opened 2024-09-08
REFACTOR: BLADEBURNER Identifier Lookup

Alpheus:
# REFACTOR: BLADEBURNER Identifier Lookup

relates to PR#1606

Our prior conversation inspired me to help with refactoring the large BB logic. This is my first contribution, treating it as a small win to go through review and the process before making larger changes.

# Context
While benign for internal usage in the BladeBurner logic, this is also used in the BB terminal and NetScript for player access. The fuzzy matcher is used to guarantee all supported (documented) usage scenarios, including the main enums.

# Motivation
I am moving as much of the non-essential logic outside of the 1500-line long Bladeburner file to ease gradual refactoring.

No external API changes.

# Coverage

Decided to test-cover it after discussions with CatLover and muesli.
image


src/Bladeburner/utils/loadActionIdentifier.ts
```diff
@@ -23,3 +23,49 @@ export function loadActionIdentifier(identifier: unknown): ActionIdentifier | nu
...
+    "black operations",
+    "black op",
+    "black ops",
+    "blackop",
+    "blackops",
+  ],
+  general: <readonly string[]>["general", "general action", "gen"],
+} as const;
+
+export function fuzzyActionIdentifier(typeShorthand: string, name: string): ActionIdentifier | null {
```
cmsautter: `fuzzyActionIdentifier`'s body seems a bit repetitive. How about iterating over `string[]: BladeBurnerActionType` key-value pairs to clean it up a bit? Also, you can use access the enums via `` `Bladeburner${actionType}Name` `` or another map.

```ts
export function fuzzyActionIdentifier(typeShorthand: string, name: string): ActionIdentifier | null {
  const type = typeShorthand.toLowerCase().trim();
  
  const actionMap = {
    contracts: BladeburnerActionType.Contract,
    operations: BladeburnerActionType.Operation,
    blackops: BladeburnerActionType.BlackOp,
    general: BladeburnerActionType.General
  };
  
  for (const [key, actionType] of Object.entries(actionMap)) {
    if (BladeBurnerTerminalShorthands[key].includes(type) && getEnumHelper(`Bladeburner${actionType}Name`).isMember(name)) {
      return { type: actionType, name };
    }
  }

  return null;
}
```

or even cleaner, why not share keys?

```ts
export const BladeBurnerTerminalShorthands = {
  Contract: <readonly string[]>["contract", "contracts", "contr"],
  Operation: <readonly string[]>["operation", "operations", "op", "ops"],
  BlackOp: <readonly string[]>[
    "blackoperation",
    "black operation",
    "black operations",
    "black op",
    "black ops",
    "blackop",
    "blackops",
  ],
  General: <readonly string[]>["general", "general action", "gen"],
} as const;

export function fuzzyActionIdentifier(typeShorthand: string, name: string): ActionIdentifier | null {
  const type = typeShorthand.toLowerCase().trim();
  
  for (const [actionType, shorthands] of Object.entries(BladeBurnerTerminalShorthands)) {
    if (shorthands.includes(type) && getEnumHelper(`Bladeburner${actionType}Name`).isMember(name)) {
      return { type: BladeburnerActionType[actionType as keyof typeof BladeburnerActionType], name };
    }
  }

  return null;
}
```

cmsautter: Also, it's best to name functions with a verb that describes what they are doing. Naming it 'fuzzyActionIdentifier' seems a bit unclear to me - it's not related to fuzzy logic, is it?
What about `resolveActionShorthand`?

cmsautter: BladeBurnerTerminalShorthands and fuzzyActionIdentifier do not need to be exported.
In your tests you can use `(Bladeburner as any).privateMethodOrField`.

cmsautter (changes requested):
I generally like the idea of removing the big switch, but here are some further improvements that could be made:

src/Bladeburner/utils/loadActionIdentifier.ts
cmsautter: Oh, I see, you moved those methods into a separate file. In that case, using `export` on `fuzzyActionIdentifier` is fine. However, I don't agree with the filename "load...". It would be better to use something like "resolve..." or "match..." instead.

Alpheus: Yeah I moved it to a file that had cohesive logic and kept it there till I learn more. These type lookups occur on a few more places so am looking for a simpler naming for the entire thing.

Alpheus: Easily, fixed. Good feedback.

Alpheus: I tried this and it made it both more complex and harder to read. I'll stick with what's there till we find something better.

src/Bladeburner/Actions/ActionIdentifier.ts
```diff
@@ -0,0 +1,66 @@
...
+      return id;
+    }
+  }
+
+  return null;
+}
+
+/** These shorthands match those documented in the BB Terminal Help */
+export const TerminalShorthands = {
+  [BladeburnerActionType.Contract]: <readonly string[]>["contract", "contracts", "contr"],
```
d0sboots: You shouldn't need `<readonly string[]>` here anymore, it's implied by `as const`.

src/Bladeburner/Actions/ActionIdentifier.ts
```diff
@@ -0,0 +1,66 @@
...
+  let id = resolveActionIdentifier({ name });
+
+  if (id && !TerminalShorthands[id.type].includes(matchedType)) {
+    id = null;
+  }
+
+  return id;
+}
+
+const TypeToActionMatchers = [
```
d0sboots: Rather than this being a bunch of closures with redundant info, I think it would be better to have it be a simple array of pairs; The name of the enum and the ActionType are the only things that are changing here. The rest of the code can be pulled to resolveActionIdentifier.

You can look at the code I recently posted in #suggestions for guidance if you are having type issues.

d0sboots (commented):
Sorry for being slow on this one. I had to pull it down locally to diff it with -M20.

Looks good, aside from a couple nits.

src/Bladeburner/Actions/ActionIdentifier.ts
cmsautter: the lambda list was also something I was worrying about, maybe something like this could be a middle ground:
```ts
function getTypedKeys<T extends object>(obj: T): Array<keyof T> {
  return Object.keys(obj) as Array<keyof T>;
}

const ActionEnums = {
  General: BladeburnerGeneralActionName,
  Contract: BladeburnerContractName,
  Operation: BladeburnerOperationName,
  BlackOp: BladeburnerBlackOpName,
};

const getActionEnum = (key: keyof typeof BladeburnerActionType) => ActionEnums[key];

const inActionEnum = (key: keyof typeof BladeburnerActionType, name: string) =>
  Object.values(getActionEnum(key)).includes(name);

const matchTypeToAction = (key: keyof typeof BladeburnerActionType, name: string): ActionIdentifier | null =>
  inActionEnum(key, name) ? ({ type: BladeburnerActionType[key], name: name } as ActionIdentifier) : null;
  // I know 'as' is circumventing ts here, but you should be able to see that this typechecks.

export const resolveActionIdentifier = (identifier: string): ActionIdentifier | null => {
  const keys = getTypedKeys(BladeburnerActionType);

  for (const key of keys) {
    const matchedAction = matchTypeToAction(key, identifier);
    if (matchedAction) {
      return matchedAction;
    }
  }

  return null; // Return null if no match is found

  // I'd prefer reduce ;D but I know that's not everyone's cup of tea
  return getTypedKeys(BladeburnerActionType).reduce<ActionIdentifier | null>((acc, key) => {
    return acc ?? matchTypeToAction(key, identifier);
  }, null);
};
```

Alpheus: > Rather than this being a bunch of closures with redundant info, I think it would be better to have it be a simple array of pairs; The name of the enum and the ActionType are the only things that are changing here. The rest of the code can be pulled to resolveActionIdentifier.
> 
> You can look at the code I recently posted in #suggestions for guidance if you are having type issues.

Can you provide an example?

> The name of the enum and the ActionType are the only things that are changing here

Three things are changing: the type, the enum and the name. The name is also typed and forms a tuple with the type (type, name) that is being discriminated against in wide areas of the code base. That's what's making it challenging, we don't have a typeguard for the tuple, only for the enum. I'll write something up that may be a good compromise.

Alpheus: Nevermind! I did some root cause analysis and it looks like we're solving the wrong problem. I'll add type discriminator guards to where the types are used, rather than the factory.

d0sboots:
```js
updateSkillLevels(this: Person): void {
  const skillsAndMults: readonly (readonly [keyof Skills & keyof Multipliers, keyof BitNodeMultipliers])[] = [
    ["hacking", "HackingLevelMultiplier"],
    ["strength", "StrengthLevelMultiplier"],
    ["defense", "DefenseLevelMultiplier"],
    ["dexterity", "DexterityLevelMultiplier"],
    ["agility", "AgilityLevelMultiplier"],
    ["charisma", "CharismaLevelMultiplier"],
  ] as const;
  for (const [skill, bnMult] of skillsAndMults) {
    this.skills[skill] = Math.max(
      1,
      Math.floor(this.calculateSkill(this.exp[skill], this.mults[skill] * currentNodeMults[bnMult])),
    );
  }
  
  const ratio: number = Math.min(this.hp.current / this.hp.max, 1);
  this.hp.max = Math.floor(10 + this.skills.defense / 10);
  this.hp.current = Math.round(this.hp.max * ratio);
}
```

Here's the code I was referring to. It's solving a slightly different problem, but it might give you some ideas.

Alpheus: Ah I see. My case is a bit different. We have 3 types: The Enum, the name and the type.

The type and name form a discriminated union to cover the 4 types as the names are all type-specific.

The reason I put it in a closure previously is to maintain the link between all three. Using your example only 2 of the inferences are maintained, and one has to be cast. I'm hesitant to go to a solution that involves more complex typing while losing precision. I'm leaning to simply deduplicate the closures. How will this affect your stance?

Alpheus: The root cause of the issue is that a discriminated union type is being used to extract runtime information about the type and put the type back into the specific context. The union doesn't share a parent class, so it's not subtype polymorphism, nor is the parent type desired. The union is only used together or specifically. I'll address that in the next update.

The issue is that when the union is being iterated over, the subtype properties (type and name) are being split into a discriminated union as well and lose the pairing. This makes it impossible to put them back together without a type guard that checks both or an unsafe cast. I've created the closures to act as a type guard over the whole. But I agree they have too much logic.

cmsautter: I'd still love to know why
`inActionEnum(key, name) ? ({ type: BladeburnerActionType[key], name: name } as ActionIdentifier) : null;`
is bad. ActionIdentifier acts as a glue between two enums: BladeburnerActionType and enums for each of it's values. So, in every other 'normal' programming language that would just a pair of integers (BladeburnerActionType, BladeburnerActionSubType). In ts string enums we now have to check for validity of those pairs at runtime. The user inputs those values via the console - they cannot be typechecked at compile time.
That only valid pairs can be constructued is completly obvious in:
`inActionEnum(key, name) ? ({ type: BladeburnerActionType[key], name: name } as ActionIdentifier) : null;`
It's saying: Is `name` in the BladeburnerActionType sub-enum `key`? If so, form the **valid** pair. 

To prevent unnoticed strange changes to the ActionIdentifier type I'd suggest moving this check closely to where the type is defined.

----

If I was to completly rewrite bb, I'd drop the name enum(s) altogether. It is, in it's own, a rip out of the action data files (like in bladeburner/data/contracts.ts): `[BladeburnerContractName.Tracking]: new Contract({
      name: BladeburnerContractName.Tracking`, which should simply be part of a list [new Contract{name : "Tracking"}], and all other needed datastructures, like a list of all Contrac names, should be derived from that.

Alpheus: That's a great question. AI would be fine if it was a supertype or at least common denomitator. That's what it _looks like_ at first glance.

But it's not. It is a discriminated union (or tagged type), so there is no "all of them" state. It's specifically _excatly one of_ the four, not a denominator that can hold _either one of_
That detail is crucial because it makes the type erasure when going inside a loop split the properties (name and type) also a union, rather than a denominator.

This problem just keeps getting worse the more hard-casts and separation we keep putting in to circumvent it because it requires switch statements on both ends.

Instead, I'm making the problem go away without having to keep carrying around. Ultimately, I think we have the same direction in mind but your root cause analysis is incomplete.

Alpheus: New update is up, made it simpler.

d0sboots:
A side comment: It helps if you don't rebase+force push changes. Doing that means I have to re-review the entire PR every time, instead of being able to look at only what's changed.

src/Bladeburner/Actions/ActionIdentifier.ts
d0sboots, line 1: Optional: This file does not define the type ActionIdentifier, so it seems a bit odd for it to be called Action/ActionIdentifier.ts. I'm also not clear on why you changed the function from loadActionIdentifier to resolveActionIdentifier; they both seem equally good to me. IMO this change could have been much less invasive.

src/Bladeburner/Actions/BlackOperation.ts
```diff
@@ -1,9 +1,10 @@
 import type { Bladeburner } from "../Bladeburner";
-import type { Availability, ActionIdentifier } from "../Types";
+import { Availability, TypeOfActionId } from "../Types";
```
d0sboots: Here and everywhere: this should still be import type.

src/Bladeburner/Actions/BlackOperation.ts
```diff
@@ -12,13 +13,22 @@ interface BlackOpParams {
...
   reqdRank: number;
   teamCount = 0;
-  get id(): ActionIdentifier {
-    return { type: this.type, name: this.name };
+
+  get id() {
+    return BlackOperation.ActionIdentifier(this.name);
+  }
+
+  static ActionIdentifier(name: BladeburnerBlackOpName): TypeOfActionId<BlackOperation> {
```
d0sboots: The name of this function is confusing, and it's also totally unneeded (it's only called once, immediately above here). Just inline this into get().

src/Bladeburner/Bladeburner.ts
```diff
@@ -1,6 +1,6 @@
 import type { PromisePair } from "../Types/Promises";
-import type { BlackOperation, Contract, GeneralAction, Operation } from "./Actions";
-import type { ActionIdentifier, Action, Attempt } from "./Types";
+import { BlackOperation, Contract, GeneralAction, Operation } from "./Actions";
```
d0sboots: Most of these should still be type too.

src/Bladeburner/Bladeburner.ts
```diff
@@ -38,9 +46,9 @@ import { formatTime } from "../utils/helpers/formatTime";
 import { joinFaction } from "../Faction/FactionHelpers";
 import { isSleeveInfiltrateWork } from "../PersonObjects/Sleeve/Work/SleeveInfiltrateWork";
 import { isSleeveSupportWork } from "../PersonObjects/Sleeve/Work/SleeveSupportWork";
-import { WorkStats, newWorkStats } from "../Work/WorkStats";
+import { newWorkStats, WorkStats } from "../Work/WorkStats";
 import { getEnumHelper } from "../utils/EnumHelper";
-import { PartialRecord, createEnumKeyedRecord, getRecordEntries } from "../Types/Record";
+import { createEnumKeyedRecord, getRecordEntries, PartialRecord } from "../Types/Record";
```
d0sboots: Did prettier do this? Why are so many import lines changing?

src/Bladeburner/Bladeburner.ts
```diff
@@ -135,7 +144,7 @@ export class Bladeburner {
     if (!availability.available) {
       return { message: `Could not start action ${action.name}: ${availability.error}` };
     }
-    this.action = actionId;
+    this.action = action.id;
```
d0sboots: Why does your PR keep growing in scope? :(

This seems like a very subtle change, and it is not immediately obvious how it's related to everything else.

src/Bladeburner/Types.ts
```diff
@@ -1,11 +1,4 @@
-import type { BlackOperation, Contract, GeneralAction, Operation } from "./Actions";
-import type {
-  BladeburnerActionType,
-  BladeburnerBlackOpName,
-  BladeburnerContractName,
-  BladeburnerOperationName,
-  BladeburnerGeneralActionName,
-} from "@enums";
+import { BlackOperation, Contract, GeneralAction, Operation } from "./Actions";
```
d0sboots: It is *definitely* the case that Types.ts can only use import type.

d0sboots (commented):
I've got more to say, but I ran out of time right now.

src/Bladeburner/Actions/ActionIdentifier.ts
Alpheus: No biggie, can rename it back. It seemed to be named after the JSON Reviver but then expanded in use. I figured it would be cohesive. Not attached to the name, can undo that.

src/Bladeburner/Bladeburner.ts
Alpheus: The two actions are not of the same type.

src/Bladeburner/Actions/ActionIdentifier.ts
```diff
@@ -0,0 +1,52 @@
+import { ActionIdentifier } from "../Types";
+import { BladeburnerActionType } from "@enums";
+import { assertLoadingType } from "../../utils/TypeAssertion";
+import { Contract } from "./Contract";
+import { BlackOperation } from "./BlackOperation";
+import { GeneralAction } from "./GeneralAction";
+import { Operation } from "./Operation";
+
+/** Loads an action identifier
+ * This is used for loading ActionIdentifier class objects from pre-2.6.1
```
d0sboots: Is this comment out of date? I'm not seeing any loading differences across versions.

src/Bladeburner/Types.ts
```diff
@@ -21,11 +14,12 @@ type AttemptFailure = { success?: undefined; message: string };
 export type Attempt<T extends object = object> = AttemptSuccess<T> | AttemptFailure;
 
 export type Action = Contract | Operation | BlackOperation | GeneralAction;
+export type TypeOfActionId<ActionType extends Action> = Pick<ActionType, "type" | "name">;
```
d0sboots: This name sounds like it's literally `typeof ActionId`, which isn't accurate. I'd lean towards something like `ActionIdFor` or just `ActionId` itself.

Edit: `ActionIdentifier` already exists, so `ActionIdFor` would probably be best.

src/Bladeburner/Types.ts
Alpheus: I'll remove the nice-to-haves from the PR

src/Bladeburner/Bladeburner.ts
d0sboots: I still don't get it. `this.action` is an `ActionIdentifier | null`. There is no type advantage to assigning it a narrower type.

d0sboots (commented):
OK, I switched to computer, which made the rest easier to review. Overall, I like the changes, even though they expanded the scope. Using `TypeOfActionId` (which I hope will be renamed) is a good change to tie the types together better.

d0sboots:
Oh, and lastly the tests fail, so clearly there's still an issue somewhere.

src/Bladeburner/Actions/ActionIdentifier.ts
Alpheus: Kept it there, I believe that was the previous version of it.

src/Bladeburner/Types.ts
Alpheus: ActionIdFor sounds good.

src/Bladeburner/Bladeburner.ts
Alpheus: Global IDE settings on my part. Want me to undo it?

src/Bladeburner/Bladeburner.ts
Alpheus: Leftover from the previous iteration. No longer necessary. ty

Alpheus:
> A side comment: It helps if you don't rebase+force push changes. Doing that means I have to re-review the entire PR every time, instead of being able to look at only what's changed.

I tried to squash them so the last commit is the only thing that changes. I figured it would help :D

Alpheus:
> Oh, and lastly the tests fail, so clearly there's still an issue somewhere.

Looking into it.

Edit: the tests are failing due to a cyclical import through JSONReviver/loadIdentifier. Fixing.

Edit: done.

src/Bladeburner/Actions/ActionIdentifier.ts
Alpheus: It is tied to functionality in JSONReviver—I'm not sure how relevant it is.

src/Bladeburner/Actions/BlackOperation.ts
Alpheus: It's used twice, which is why I pulled it out (here and the factory). I'll add a verb to it to help connect it.

src/Bladeburner/Actions/ActionIdentifier.ts
Alpheus: Reverted.

src/Bladeburner/Bladeburner.ts
d0sboots: Either way is fine, but in the future you should configure your IDE to not change things you aren't explicitly touching.

d0sboots:
Great! This version is a lot more straightforward (less renaming of stuff to make it more confusing to follow.)

PR #1649, opened 2024-09-12
FIX: Resolve Alias Recursion Issue by Preventing Duplicate Rule Application

### LINKED ISSUE #1647: recursive alias causes error
version of the game: Bitburner v2.6.2 (633da3830)

I have aliased `buy="buy -l"`. Now, when I type "buy" into the terminal and press enter, the game hangs briefly, then gives the following error:

```
UNCAUGHT PROMISE ERROR
You forgot to await a promise
maybe hack / grow / weaken ?

RUNTIME ERROR:

too much recursion
stack:
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18541
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
h@https://bitburner-official.github.io/dist/main.bundle.js:2:18658
```

[bitburnerSave_1725813539_BN1x1.json.gz](https://github.com/user-attachments/files/16922972/bitburnerSave_1725813539_BN1x1.json.gz)

cmsautter:
Closes #1647

Previously, the check was only to ensure that the right side of an alias rule does not repeat, which is not effective for endlessly expanding aliases such as buy="buy -l". Now, the implementation ensures a single alias rule is not recursively applied multiple times by remembering and checking the entire rule.

Updates expected values in tests; this change may disrupt existing setups with endless recursive aliases. Please verify if the new expected values align with your expectations. It would be possible to avoid applying the last substitution if it would result in infinite recursion, but this approach would not comply with the POSIX standard.

### DIFF
--- a/src/Alias.ts
+++ b/src/Alias.ts
@@ -83,7 +83,7 @@ export function substituteAliases(origCommand: string): string {
  * @param currentlyProcessingAliases any aliases that have been applied in the recursive evaluation leading to this point
  * @return { string } the provided command with all of its referenced aliases evaluated
  */
-function applyAliases(origCommand: string, depth = 0, currentlyProcessingAliases: string[] = []) {
+function applyAliases(origCommand: string, depth = 0, currentlyProcessingAliases: Set<string> = new Set()) {
   if (!origCommand) {
     return origCommand;
   }
@@ -98,17 +98,17 @@ function applyAliases(origCommand: string, depth = 0, currentlyProcessingAliases
   // (unless there are any reference loops or the reference chain is too deep)
   const localAlias = Aliases.get(commandArray[0]);
   const localrule = commandArray[0] + "->" + localAlias + "(l)";
-  if (localAlias && !currentlyProcessingAliases.includes(localrule)) {
-    const appliedAlias = applyAliases(localAlias, depth + 1, [localrule, ...currentlyProcessingAliases]);
+  if (localAlias && !currentlyProcessingAliases.has(localrule)) {
+    const appliedAlias = applyAliases(localAlias, depth + 1, new Set([localrule, ...currentlyProcessingAliases]));
     commandArray.splice(0, 1, ...appliedAlias.split(" "));
   }
 
   // Once local aliasing is complete (or if none are present) handle any global aliases
   const processedCommands = commandArray.reduce((resolvedCommandArray: string[], command) => {
     const globalAlias = GlobalAliases.get(command);
-    const rule = command + "->" + globalAlias + "(g)";
-    if (globalAlias && !currentlyProcessingAliases.includes(rule)) {
-      const appliedAlias = applyAliases(globalAlias, depth + 1, [rule, ...currentlyProcessingAliases]);
+    const globalrule = command + "->" + globalAlias + "(g)";
+    if (globalAlias && !currentlyProcessingAliases.has(globalrule)) {
+      const appliedAlias = applyAliases(globalAlias, depth + 1, new Set([globalrule, ...currentlyProcessingAliases]));
       resolvedCommandArray.push(appliedAlias);
     } else {
       // If there is no alias, or if the alias has a circular reference, leave the command as-is
--- a/src/Alias.ts
+++ b/src/Alias.ts
@@ -93,7 +93,7 @@ function applyAliases(origCommand: string, depth = 0, currentlyProcessingAliases
   // First get non-global aliases, and recursively apply them
   // (unless there are any reference loops or the reference chain is too deep)
   const localAlias = Aliases.get(commandArray[0]);
-  const localrule = commandArray[0] + "->" + localAlias + "(g)";
+  const localrule = commandArray[0] + "->" + localAlias + "(l)";
   if (localAlias && !currentlyProcessingAliases.includes(localrule)) {
     const appliedAlias = applyAliases(localAlias, depth + 1, [localrule, ...currentlyProcessingAliases]);
     commandArray.splice(0, 1, ...appliedAlias.split(" "));
--- a/src/Alias.ts
+++ b/src/Alias.ts
@@ -62,9 +62,13 @@ export function removeAlias(name: string): boolean {
   return hadAlias;
 }
 
+export function clearAliases(): void {
+  Aliases.clear();
+  GlobalAliases.clear();
+}
+
 /**
  * Returns the original string with any aliases substituted in.
- * Aliases are only applied to "whole words", one level deep
  * @param origCommand the original command string
  */
 export function substituteAliases(origCommand: string): string {
--- a/src/Alias.ts
+++ b/src/Alias.ts
@@ -93,16 +93,18 @@ function applyAliases(origCommand: string, depth = 0, currentlyProcessingAliases
   // First get non-global aliases, and recursively apply them
   // (unless there are any reference loops or the reference chain is too deep)
   const localAlias = Aliases.get(commandArray[0]);
-  if (localAlias && !currentlyProcessingAliases.includes(localAlias)) {
-    const appliedAlias = applyAliases(localAlias, depth + 1, [commandArray[0], ...currentlyProcessingAliases]);
+  const localrule = commandArray[0] + "->" + localAlias + "(g)";
+  if (localAlias && !currentlyProcessingAliases.includes(localrule)) {
+    const appliedAlias = applyAliases(localAlias, depth + 1, [localrule, ...currentlyProcessingAliases]);
     commandArray.splice(0, 1, ...appliedAlias.split(" "));
   }
 
   // Once local aliasing is complete (or if none are present) handle any global aliases
   const processedCommands = commandArray.reduce((resolvedCommandArray: string[], command) => {
     const globalAlias = GlobalAliases.get(command);
-    if (globalAlias && !currentlyProcessingAliases.includes(globalAlias)) {
-      const appliedAlias = applyAliases(globalAlias, depth + 1, [command, ...currentlyProcessingAliases]);
+    const rule = command + "->" + globalAlias + "(g)";
+    if (globalAlias && !currentlyProcessingAliases.includes(rule)) {
+      const appliedAlias = applyAliases(globalAlias, depth + 1, [rule, ...currentlyProcessingAliases]);
       resolvedCommandArray.push(appliedAlias);
     } else {
       // If there is no alias, or if the alias has a circular reference, leave the command as-is
--- a/test/jest/Alias/Alias.test.ts
+++ b/test/jest/Alias/Alias.test.ts
@@ -7,7 +7,19 @@ describe("substituteAliases Tests", () => {
     parseAliasDeclaration("d=recursiveAlias");
 
     const result = substituteAliases("recursiveAlias");
-    expect(result).toEqual("d");
+    expect(result).toEqual("recursiveAlias");
+  });
+
+  it("Should gracefully handle recursive local aliases II", () => {
+    parseAliasDeclaration("recursiveAlias=recursiveAlias");
+    const result = substituteAliases("recursiveAlias");
+    expect(result).toEqual("recursiveAlias");
+  });
+
+  it("Should gracefully handle recursive local aliases II", () => {
+    parseAliasDeclaration('recursiveAlias="recursiveAlias -l"');
+    const result = substituteAliases("recursiveAlias");
+    expect(result).toEqual("recursiveAlias -l");
   });
 
   it("Should only change local aliases if they are the start of the command", () => {
@@ -27,7 +39,7 @@ describe("substituteAliases Tests", () => {
     parseAliasDeclaration("d=a", true);
 
     const result = substituteAliases("a b c d");
-    expect(result).toEqual("d a b c");
+    expect(result).toEqual("a b c d");
   });
 
   it("Should gracefully handle recursive mixed local and global aliases", () => {
@@ -37,7 +49,7 @@ describe("substituteAliases Tests", () => {
     parseAliasDeclaration("d=recursiveAlias", false);
 
     const result = substituteAliases("recursiveAlias");
-    expect(result).toEqual("d");
+    expect(result).toEqual("recursiveAlias");
   });
 
   it("Should replace chained aliases", () => {
--- a/test/jest/Alias/Alias.test.ts
+++ b/test/jest/Alias/Alias.test.ts
@@ -1,6 +1,9 @@
-import { substituteAliases, parseAliasDeclaration } from "../../../src/Alias";
+import { substituteAliases, parseAliasDeclaration, clearAliases } from "../../../src/Alias";
 describe("substituteAliases Tests", () => {
-  it("Should gracefully handle recursive local aliases", () => {
+  beforeEach(() => {
+    clearAliases();
+  });
+  it("Should gracefully handle recursive local aliases I", () => {
     parseAliasDeclaration("recursiveAlias=b");
     parseAliasDeclaration("b=c");
     parseAliasDeclaration("c=d");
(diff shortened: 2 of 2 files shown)


cmsautter:
> Previously, the check was only to ensure that the right side of an alias rule does not repeat. 

Actually, it was more flawed than that, but it's hard to put into words. Whenever it processed a rule like `a=b`, it checked whether it already had seen the left side `a`, but then 'remembers' the ride side `b`. This has the same effect, that when the right side expands, i.e. `a=a b`, it will run into endless recursion, since it always checks whether it had seen `a`, but then remember `a b`, so it will never see just `a` on the right side and thus apply the rule endlessly.

cmsautter:
> Updates expected values in tests

Basically for rules like (`a=b`, `b=a`) previously `applyAlias('a')` returned `b`, now it returns `a`, since it stopps as soon as it tries to apply the first rules a second time.

Previously it did following shenanigans (see post above):
input: `a`, rule: `a=b`, visited: `{}`, `check a in {}`, `remember {b}`, `output b`
input: `b`, rule: `b=a`, visited: `{b}`, `check b in {b}`, `stop`, `output input b`

cmsautter:
The new expected values also correspond to the behavior observed with aliases under Ubuntu:
grafik

cmsautter:
And here a screenshot that the alias mentioned in #1647 now works as expected, regardless of whether its useful:
grafik

test/jest/Alias/Alias.test.ts
```diff
@@ -1,13 +1,54 @@
...
+    const result = substituteAliases("'a'");
+    expect(result).toEqual("'a'");
+  });
+
+  it("Should not substitute quoted commands III", () => {
+    parseAliasDeclaration("a=b");
+    parseAliasDeclaration("b='c'");
+    parseAliasDeclaration("c=d");
+    const result = substituteAliases("a");
+    //expect(result).toEqual("'c'"); // Currently FAILS
```
d0sboots: You can write this test as `skip` or `failing` (they have different semantics; `failing` is *expected* to fail, `skip` is just not run since the result is ignored).

src/Alias.ts
```diff
@@ -93,16 +97,18 @@ function applyAliases(origCommand: string, depth = 0, currentlyProcessingAliases
   // First get non-global aliases, and recursively apply them
   // (unless there are any reference loops or the reference chain is too deep)
   const localAlias = Aliases.get(commandArray[0]);
-  if (localAlias && !currentlyProcessingAliases.includes(localAlias)) {
-    const appliedAlias = applyAliases(localAlias, depth + 1, [commandArray[0], ...currentlyProcessingAliases]);
+  const localrule = commandArray[0] + "->" + localAlias + "(l)";
+  if (localAlias && !currentlyProcessingAliases.includes(localrule)) {
+    const appliedAlias = applyAliases(localAlias, depth + 1, [localrule, ...currentlyProcessingAliases]);
```
d0sboots: Optional: Using an array for this is kinda cruddy. It would be pretty simple/faster to switch to an object (everything here has string keys) with set = `currentlyProccesingAliases[localrule] = true` and test being `&& !currentlyProcessingAliases[localrule]`

d0sboots (commented):
Nice tests!

src/Alias.ts
cmsautter: I'l research that. Lists are normally faster than Hashmaps / Sets for small n though, but you are definitly right for larger n. I'm interested in checking how js objects internally work, but they probably are just Hashmaps. Syntax-wise I prefer `!currentlyProcessingAliases.includes(localrule)` but from the semantics I'd prefer the hashing to stay in constant time. So I'll come back to this, when I know more.

d0sboots: If you like that syntax better you could also use a Set, which tend to be slightly slower than objects (for this purpose) on some browsers but only slightly (these days).

cmsautter: After having done some research, I'd prefer using a Set. That's exactly their purpose, holding a unique set of keys. Sets outperform Maps, since they are practically Maps without values, and Maps outperform Objects in Scenarios where the content changes more dynamically, like it is the case here. JS Objects are also just implemented as Maps in the different browser engines. Using Sets and Maps over Objects also seems to be the more modern approach when you don't need a fully fledged object with a class/prototype.

test/jest/Alias/Alias.test.ts
cmsautter: there is also xtest or xit, which I find funny, but I'd use test.skip here.

ficocelliguy:
Good catch on this. Nice test coverage, too.

cmsautter:
I've incorporated the feedback from d0sboots. Please let me know if there’s anything else to adjust!

src/Alias.ts
d0sboots: Just a few comments:
Sets aren't always faster than Maps, or faster than objects in this purpose. It heavily depends on browser, also version. Ultimately, only benchmarking will tell you.
Also, Objects definitely are *not* just implemented as Maps under the covers; there's a lot more going on there. JS interpreter/compilers have some crazy stuff going on under the hood.

None of that changes this though; I'm fine with using Set here.

src/Alias.ts
```diff
@@ -98,17 +98,17 @@ function applyAliases(origCommand: string, depth = 0, currentlyProcessingAliases
   // (unless there are any reference loops or the reference chain is too deep)
   const localAlias = Aliases.get(commandArray[0]);
   const localrule = commandArray[0] + "->" + localAlias + "(l)";
-  if (localAlias && !currentlyProcessingAliases.includes(localrule)) {
-    const appliedAlias = applyAliases(localAlias, depth + 1, [localrule, ...currentlyProcessingAliases]);
+  if (localAlias && !currentlyProcessingAliases.has(localrule)) {
+    const appliedAlias = applyAliases(localAlias, depth + 1, new Set([localrule, ...currentlyProcessingAliases]));
```
d0sboots, line 102: I really wanted to just merge, but I can't let this go... XD

Don't create a brand-new Set, that will be the same O(N^2) behavior as the array you just replaced. Add the item to the set you have.

Edit: For this to work properly (because there are other checks that happen after), you also have to remove it after you recurse.

cmsautter: I need to think this through, I'm just in love with immutables. But you're right ofc that this is a little expensive. 
For the global rules you have to copy, since a global should be able to be applied multiple times. For the local one it might suffice to just add it.

d0sboots: If you really can't stand to do mutable stuff, I'm fine with you changing it back to an array. There's no point in doing something more verbose if you're going to be N^2 anyway, and most people won't have lots of deep aliases.

cmsautter: Oh don't merge, there is another bug that will make local rules be applied globally I believe. Sorry that I haven't thought this through carefully before. I was just focused on this single bug.

cmsautter: It shouldn't be hard to fix though, but I'd like to write some more tests to catch it.

d0sboots: based attitude

cmsautter: There is a point in doing it more verbose: readability. Sets, in contrasts to Arrays, are unique and often used for lookup. So it helps with getting the codes intention.

I think mutating the set for the local aliases is totally fine, but copying is needed for the global ones. I don't believe you can circumvent that.

I will probably finish this tomorrow. It is easy to mess up.
After global rules are applied, subsequent calls to applyAlias should just consider global rules with the exception of a global rule modifying the first token/command.  I believe the function might need some restructuring of the execution flow to handle all edge cases.

d0sboots: Regardless of how you end up restructuring, the following should work for mutation without copying: Add the key to the set immediately before recursion, and remove it immediately after. 

A quick sketch of why this works: You abort if the key was already present, so there's never an issue of double-removing a key. Thus, the removal always properly reverts to the prior state.

cmsautter: Ah that's a great approach. I'll probably use that, but then also "unify" the application of local and global rules inside one loop / reduce, to avoid certain edge cases.

PR #1653, opened 2024-09-15
MISC: Refactor coding contracts for type safety

d0sboots:
This refactor does three things without any behavior changes:
* Adds type safety by introducing generic type parameters to the coding contract definitions, so they can use concrete types instead of "unknown". This also eliminates a bunch of boilerplate casts.
* Removes the unneeded CodingContractType class. We can use the metadata type directly.
* Introduces a new hidden state to coding contracts. Instead of generating and storing the data (which is what is shown to the user as the problem's input), the state is stored instead. This allows problems to (for instance) generate the answer up-front, and check the solution directly against the answer, instead of needing to embed a solver in the problem (which can then easily be ripped from the source code).

For compatibility, state == data by default. I plan to make use of this feature in a followup, but it is unused currently.

## Testing

Tested that dummy contracts still work and can be solved.

Tested that a savegame from the current version can be loaded, and that a pre-existing contract shows its data correctly.

### DIFF
--- a/src/CodingContracts.ts
+++ b/src/CodingContracts.ts
@@ -9,7 +9,11 @@ import { ContractFilePath, resolveContractFilePath } from "./Paths/ContractFileP
 export const CodingContractTypes: Record<string, CodingContractType<unknown>> = {};
 
 for (const md of codingContractTypesMetadata) {
-  CodingContractTypes[md.name] = md;
+  // Because functions are contravariant with their parameters, we can't
+  // consider arbitrary CodingContractTypes as CodingContractType<unknown>
+  // directly. However, we do want that as the final type, to enforce that the
+  // state and data are unknown. So we cast through CodingContractType<any>.
+  CodingContractTypes[md.name] = md as CodingContractType<any>;
 }
 
 // Numeric enum
--- a/src/CodingContracts.ts
+++ b/src/CodingContracts.ts
@@ -1,59 +1,15 @@
 import type { FactionName } from "@enums";
-import { codingContractTypesMetadata, DescriptionFunc, GeneratorFunc, SolverFunc } from "./data/codingcontracttypes";
+import { codingContractTypesMetadata, type CodingContractType } from "./data/codingcontracttypes";
 
 import { Generic_fromJSON, Generic_toJSON, IReviverValue, constructorsForReviver } from "./utils/JSONReviver";
 import { CodingContractEvent } from "./ui/React/CodingContractModal";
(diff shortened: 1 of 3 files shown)


src/data/codingcontracttypes.ts
```diff
@@ -85,10 +91,9 @@ export const codingContractTypesMetadata: ICodingContractTypeMetadata[] = [
 
       return (n === 1 ? fac - 1 : n) === parseInt(ans, 10);
     },
-  },
+  } as CodingContractType<number>,
```
G4mingJon4s: Not sure if this *works*, but `satisfies CodingContractType<number>` should still check for the object to match the specified form, instead of just saying it can't be coerced?

src/CodingContracts.ts
```diff
@@ -95,8 +51,8 @@ export type ICodingContractReward =
  * The player receives a reward if the problem is solved correctly
  */
 export class CodingContract {
```
G4mingJon4s, line 57: Can't the `CodingContractTypes[type]` be on the `CodingContract`? It would remove checking for it being undefined or assuming it's state type to be `unknown`.

G4mingJon4s (commented):
Just some small stuff. Would also be fine without my suggestions.

src/data/codingcontracttypes.ts
```diff
@@ -4,23 +4,31 @@ import { comprGenChar, comprLZGenerate, comprLZEncode, comprLZDecode } from "../
...
+   */
+  getData?: (state: State) => Data;
+  /** Name of the problem. Used to request contracts of this type. */
   name: string;
-  numTries: number;
-  solver: SolverFunc;
+  /** How many tries you get. Defaults to 10. */
+  numTries?: number;
+  /** Function that checks if the provided solution is correct. */
+  solver: (state: State, answer: string) => boolean;
```
Alpheus, line 31: I like this change.

src/data/codingcontracttypes.ts
d0sboots: Good call.

src/CodingContracts.ts
d0sboots: It took me a second, but I see what you mean now: Passing in the `CodingContractType` object directly, instead of the string `type` and needing to look it up.

Normally, that would be a better design, but I know why it's this way. The `CodingContract` class is serialized as part of the savegame, and this makes it both smaller (only the name is saved) and also more future-proof (if we need to bugfix a cct, that bugfix actually applies to existing saves, instead of the bugged version being persisted in old ccts).

PR #1654, opened 2024-09-16
BLADEBURNER: Move bladeburner team losses to Casualties

Alpheus:
# Motivation
I am moving as much of the non-essential logic outside of the 1500-line long Bladeburner file to ease gradual refactoring.

No external or NS API changes.

# Coverage
image


test/jest/Bladeburner/BladeburnerCasualties.test.ts
```diff
@@ -0,0 +1,90 @@
...
+  KillBladeburnerCasualties,
+  OperationCasualtyOutcome,
+} from "../../../src/Bladeburner/Actions/Casualties";
+
+import { getRandomIntInclusive } from "../../../src/utils/helpers/getRandomIntInclusive";
+
+describe("Bladeburner Team", () => {
+  beforeEach(() => {
+    // Return the high end of all random rolls
+    getRandomIntInclusive.mockImplementationOnce((_: number, high: number) => high);
```
d0sboots: Why not just use `mockImplementation`? You're not doing something complicated such that the "Once" part is important.

src/Bladeburner/Actions/Casualties.ts
```diff
@@ -0,0 +1,51 @@
+import { getRandomIntInclusive } from "../../utils/helpers/getRandomIntInclusive";
+import { OperationTeam } from "../Bladeburner";
```
d0sboots: import type

src/Bladeburner/Bladeburner.ts
```diff
@@ -50,10 +50,19 @@ import { GeneralActions } from "./data/GeneralActions";
 import { PlayerObject } from "../PersonObjects/Player/PlayerObject";
 import { Sleeve } from "../PersonObjects/Sleeve/Sleeve";
 import { autoCompleteTypeShorthand } from "./utils/terminalShorthands";
+import { CasualtyReport, KillBladeburnerCasualties, OperationCasualtyOutcome } from "./Actions/Casualties";
 
 export const BladeburnerPromise: PromisePair<number> = { promise: null, resolve: null };
 
-export class Bladeburner {
+export interface OperationTeam {
```
d0sboots: It's arguable, but this interface feels like it belongs to Casualties.ts.

src/Bladeburner/Actions/Casualties.ts
```diff
@@ -0,0 +1,51 @@
...
+  deaths: number; // Actual roll
+  damagedSleeves: number; // Sleeves included in actual deaths
+};
+
+type CasualtyRolls = {
+  worstCase: number; // High end roll for potential deaths
+  bestCase: number; // Low end roll for potential deaths
+};
+
+export enum OperationCasualtyOutcome {
```
d0sboots: This one is also arguable, but in this particular case I feel like the two-member enum is adding more extra verbosity than it is helping. Just using a boolean "isSevere" feels like it would cut out a lot of boilerplate.

src/Bladeburner/Actions/Casualties.ts
```diff
@@ -0,0 +1,51 @@
+import { getRandomIntInclusive } from "../../utils/helpers/getRandomIntInclusive";
+import { OperationTeam } from "../Bladeburner";
+
+export type Casualties = {
+  deaths: number; // Actual roll
+  damagedSleeves: number; // Sleeves included in actual deaths
+};
+
+type CasualtyRolls = {
```
d0sboots: I don't see any actual uses of this type? If it's just for testing, it probably shouldn't be added - you should think of a different way to get effective tests.

test/jest/Bladeburner/BladeburnerCasualties.test.ts
Alpheus: Since it's not injected, I want the test to fail in case the underlying code ever gets expanded with more than one use.

src/Bladeburner/Actions/Casualties.ts
Alpheus: Blimey I've never been called out on this so often. Do you have import -> import type automated by any chance in your IDE? It seems like this should be covered by the linter or formatter.

Thanks for being so dilligent.

src/Bladeburner/Bladeburner.ts
Alpheus: I wasn't sure. I _think_ this belong to whatever the underlying link between Operations and BlackOperations is. So until we refactor that, I'd be most comfortable sticking it into Action. Action is already aware of "Team Success Bonus", even though it's not aware of team size (That's a discrimanted union on Ops/BOPS). So I suppose out of here is good.

Afterall, purpose of these refactors is to slim down non-essentials from Bladeburner.ts

src/Bladeburner/Actions/Casualties.ts
Alpheus: I had the same argument in mind. Just because it has two values, doesn't make it a good candidate for a boolean. I think boolean types should be used specifically for a YES / NO situation. In a LOW-HIGH situation I'd rather have LOW and HIGH.

If you insight, I can turn it specifically into 0.5 and 1 (1 | 0.5) and pass around the factor, but that's opening doors for architecture we don't need to support right now.

However, since this is mostly just data, I could stick it into Bladeburner constants as the default loss rate and not give a choice at all (just pass in success).

Which one would you prefer?

src/Bladeburner/Actions/Casualties.ts
Alpheus: The kill doesn't need the rolls, I prefer this over ducktyping.

src/Bladeburner/Actions/Casualties.ts
d0sboots: I am a robot XD

We can't turn on lint for this, because there's way too many existing violations. A vague plan is to try to incrementally clean those up so we can enforce this automatically. (It would let us enable some better options in tsconfig)

Alpheus: Ok knowing that I've forced this one ON in my IDE (I turned it _off_ last time as you called me out on re-sorting the imports), but I'm only changing letting it through on imports that are _added_ as a result of me typing code. It's not bullet proof but should lessen the burden on the reviewer.

src/Bladeburner/Actions/Casualties.ts
Alpheus: The intention is to separate the types from "what could happen, ie. potential" from "what did happen"

I can separate the types, but the correct hinting results in more code being added on the Bladeburner side. Though now that I'm looking at it from a fresh perspective, I may just move this to a method and hint the type of the parameter (similar to how Bladeburner is being passed to Action.ts getters, though a step I disagree with design-wise)

src/Bladeburner/Actions/TeamCasualties.ts
```diff
@@ -0,0 +1,49 @@
...
+ * Caveats: TeamSize = Human Team + Supporting Sleeves set to this action
+ */
+export class TeamCasualties {
+  constructor(private severity: CasualtyFactor, private teamSizeUsed: number, private supportingSleeves: number) {}
+
+  rollOutcome(random: typeof getRandomIntInclusive, team: OperationTeam) {
+    const deaths = random(this.bestCase, this.worstCase);
+    const humans = this.teamSizeUsed - this.supportingSleeves;
+    const humanDeaths = Math.min(humans, deaths);
+    const damagedSleeves = deaths - humanDeaths;
```
catloversg: You should double-check the logic here.

`teamSizeUsed` = `action.teamCount` and `supportingSleeves` = `this.sleeveSize`. Let's say we have 4 team members (from recruitment), 2 supporting sleeves and we set team count to 4 when doing the operation: `teamSizeUsed` = 4, `supportingSleeves` = 2, and `this.teamSize` = 6.

If `deaths` = 4, `damagedSleeves` should be 0, but, in your implementation, `damagedSleeves` is 2 (`deaths - Math.min(teamSizeUsed - supportingSleeves, deaths)` = 4 - min(4 - 2, 2) = 2.

In the current implementation, `damagedSleeves` is calculated based on `this.teamSize`, not `action.teamCount`. It's `this.teamSize - losses - this.sleeveSize`.

The current logic: when there is danger, we use team members (from recruitment) as cannon fodder and let them die before our sleeves. If you want to change this logic, you need to state that in the PR, because it's a balance change.

src/Bladeburner/Actions/TeamCasualties.ts
```diff
@@ -0,0 +1,49 @@
...
+
+    /** Supporting Sleeves take damage when they are part of losses,
+     *   e.g. 8 sleeves + 3 team members with 4 losses -> 1 sleeve takes damage */
+    team.killSupportingSleeves(damagedSleeves);
+
+    team.teamSize -= humanDeaths;
+    team.teamLost += deaths;
```
catloversg: The current implementation is wrong. It was reported, but nobody bothers fixing it. It's best to fix it here.

Currently, it's:
```js
this.teamSize -= losses;
if (this.teamSize < this.sleeveSize) {
  const sup = Player.sleeves.filter((x) => isSleeveSupportWork(x.currentWork));
  for (let i = 0; i > this.teamSize - this.sleeveSize; i--) {
    const r = Math.floor(Math.random() * sup.length);
    sup[r].takeDamage(sup[r].hp.max);
    sup.splice(r, 1);
  }
  this.teamSize += this.sleeveSize;
}
this.teamLost += losses;
```
`this.teamSize -= losses;` is fine, but `this.teamSize += this.sleeveSize` is wrong.

Let's check this example: We have 2 team members (from recruitment), 4 supporting sleeves and we set team count to 2 when doing the operation. `this.teamSize` = 6. `this.sleeveSize` = 4. `teamSizeUsed` = 2.
If `deaths` = 2 (all team members die; 0 supporting sleeves take damage), `this.teamSize` should be 4 (0 supporting sleeves take damage; all supporting sleeves still support the main body), but:
- In the current implementation, it's 8: `this.teamSize - losses + this.sleeveSize` = 6 - 2 + 4 = 8.
- In your implementation, it's also 8: `team.teamSize - Math.min(teamSizeUsed - supportingSleeves, deaths)` = 6 - min(2 - 4, 2) = 6 - min(-2, 2) = 8.

The proper fix **may** be:
```js
// Deduct the losses
this.teamSize -= losses;
// If this happens, all team members died and there are "damaged sleeves". In this case, teamSize = sleeveSize.
if (this.teamSize < this.sleeveSize) {
  this.teamSize = this.sleeveSize;
}
team.teamLost += losses;
```
@d0sboots Please double-check this logic.

src/Bladeburner/Bladeburner.ts
```diff
@@ -750,31 +751,26 @@ export class Bladeburner {
...
 
+  private getActionCasualties(action: Operation | BlackOperation, success: boolean) {
+    const severity = success ? CasualtyFactor.LOW_CASUALTIES : CasualtyFactor.HIGH_CASUALTIES;
+    const casualties = new TeamCasualties(severity, action.teamCount, this.sleeveSize);
+    return casualties.rollOutcome(getRandomIntInclusive, this);
+  }
+
+  public killSupportingSleeves(sleeveDeaths: number) {
+    const sup = Player.sleeves.filter((x) => isSleeveSupportWork(x.currentWork));
+    const damagedIndices = [...Array(Math.min(sup.length, sleeveDeaths)).keys()].sort(() => 0.5 - Math.random());
```
catloversg: This is just a nitpick.

If you want to shuffle an array, you should use Fisher–Yates shuffle, not this `0.5 - Math.random()` trick. We have duplicated implementation of Fisher–Yates shuffle in our code, and we even import `shuffle` from `lodash`. If you want to use Fisher–Yates shuffle, you can use one of them. I'll clean them up later. If you think that it's unnecessary, you can ignore this comment. It's not a big deal in our use case. There are other problems with our randomization functions that I turn a blind eye to.

src/Bladeburner/Bladeburner.ts
```diff
@@ -992,8 +988,7 @@ export class Bladeburner {
         }
 
         // Team loss variables
```
catloversg: This comment is outdated.

test/jest/Bladeburner/TeamCasualties.test.ts
```diff
@@ -0,0 +1,83 @@
...
+    });
+
+    it("at worst half the team when low", () => {
+      const losses = CasualtyReport(5, CasualtyFactor.LOW_CASUALTIES, 0);
+      expect(losses.worstCase).toBeLessThan(5);
+      expect(losses.worstCase).toBeGreaterThan(0);
+    });
+  });
+
+  function TeamWithShockingSleeves(teamSize: number = 0) {
```
catloversg: Function names should be camel case.

catloversg (commented):
I have to rewrite this entire feedback after rereading everything due to your force-push, so there may be some copy-paste errors. I put feedback into relevant parts of the changes, but there is a small concern for the overall code style. It's opinion-based, so feel free to ignore it.

I think some parts of the PR tend to become a bit overengineered over time. I can guess why they are like that, but I'm not sure if it's necessary. For example: the change of `worstCase`.

In the first implementation of this PR, it looked like this:
```js
const worstCaseFraction = outcome === OperationCasualtyOutcome.LOW_CASUALTIES ? 0.5 : 1;
const worstCaseOp = Math[outcome === OperationCasualtyOutcome.LOW_CASUALTIES ? "ceil" : "floor"];
const worstCase = worstCaseOp(teamSizeUsedForAction * worstCaseFraction);
```
IMO, it's overengineered. The old implementation is just `const maxLosses = success ? Math.ceil(teamCount / 2) : Math.floor(teamCount);`. The new change is much harder to read than:
```js
const worstCase = outcome === OperationCasualtyOutcome.LOW_CASUALTIES
  ? Math.ceil(teamSizeUsedForAction / 2)
  : Math.floor(teamSizeUsedForAction);
```

In the current implementation, it was put in `TeamCasualties` and this class's constructor even receives a random function as a parameter. Now, we can "customize" many things: CasualtyFactor, random function, etc. This flexibility is good for testing (no special mocking), but if we overuse it, the code becomes much harder to read. Please note that I'm **not** saying that your code is hard to read. My point is that we should maintain the balance between the flexibility (tons of options) and the simplicity.

With all being said, there is no need to change your code, though. It comes with proper tests, and the flexibility may bring more benefits than troubles.

Alpheus:
Cheers, simple fix. Anything else?

src/Bladeburner/Bladeburner.ts
Alpheus: Can you show me a sample usage of our internal fisher-yates?

src/Bladeburner/Actions/TeamCasualties.ts
```diff
@@ -0,0 +1,49 @@
...
+ * Some actions (Operations and Black Operations) use teams for success bonus
+ * and may result in casualties, reducing the player's hp, killing team members
+ * and killing sleeves (to shock them, sleeves are immortal)
+ *
+ * Caveats: TeamSize = Human Team + Supporting Sleeves set to this action
+ */
+export class TeamCasualties {
+  constructor(private severity: CasualtyFactor, private teamSizeUsed: number, private supportingSleeves: number) {}
+
+  rollOutcome(random: typeof getRandomIntInclusive, team: OperationTeam) {
```
cmsautter: I don't like this 'throwaway' class. If you'd like to parametrize, and then keep it this way, you can do so using a function returning a parametrized function. But you're using it as follows:

```ts
const casualties = new TeamCasualties(severity, action.teamCount, this.sleeveSize);
return casualties.rollOutcome(getRandomIntInclusive, this);
```
casualities get's then thrown away the next time the garbage collector runs. The only thing that I can imagine that would somehow make sense to keep parametrized is random, the random number generator, to be able too keep the events deterministic. But that's currently not a thing in the whole codebase. Also, you are using another random function to shuffle the array, so having one as a parameter seems not consistent. What was your intent?

I'd recommend to drop the random parameter, and just use the stack for all 4 remaining arguments:
**Make this a static function.**

What I really like is the interface for the BladeBurner object (OperationTeam)!

src/Bladeburner/Actions/TeamCasualties.ts
Alpheus: This PR is a refactor. I'm not doing any bugfixes and I'm testing any behavior as-is, including test-covering currently known bugs.

src/Bladeburner/Actions/TeamCasualties.ts
Alpheus: I haven't considered killing off human team members that did not partake in the action. I will check with the current dev branch and have that behavior match. I can't confirm whether you're correct outside of that.

src/Bladeburner/Bladeburner.ts
```diff
@@ -750,31 +751,26 @@ export class Bladeburner {
...
+    const severity = success ? CasualtyFactor.LOW_CASUALTIES : CasualtyFactor.HIGH_CASUALTIES;
+    const casualties = new TeamCasualties(severity, action.teamCount, this.sleeveSize);
+    return casualties.rollOutcome(getRandomIntInclusive, this);
+  }
+
+  public killSupportingSleeves(sleeveDeaths: number) {
+    const sup = Player.sleeves.filter((x) => isSleeveSupportWork(x.currentWork));
+    const damagedIndices = [...Array(Math.min(sup.length, sleeveDeaths)).keys()].sort(() => 0.5 - Math.random());
+    damagedIndices.forEach((idx) => sup[idx].kill());
+  }
```
cmsautter: I would love to see this also excluded to TeamCasualities.ts
Maybe provide a wrapper to get the sleeves that are doing support work still from bladeburner, meaning keeping this 'bad' code out: 
```ts
Player.sleeves.filter((x) => isSleeveSupportWork(x.currentWork)); 
``` 

and just provide SleeveTeamMembers or getSleeveTeamMembers to the TeamCasualities.rollOutcome function.

I hope the god singleton object Player will get refactored later to have some less responsibilities xD.

src/Bladeburner/Bladeburner.ts
catloversg: You can search the usages of `shuffleArray` in `src\Infiltration\ui\BribeGame.tsx` and `shuffle` in `src\data\codingcontracttypes.ts`.

Alpheus: Will do, thanks!

test/jest/Bladeburner/TeamCasualties.test.ts
cmsautter: and be a verb.

cmsautter (changes requested):
I like rollOutcome now being overall more readable, but have some suggestions on how to further reorder things slightly. 
This is just meant as an addition to catloversg's points, he made some good remarks on how to reuse shuffle functions and his note, that your implementation does not reproduce a known bug. I'd suggest keeping the bugged behaviour in your new implementation, and move fixing the bug, which should then be a one-liner, to another Pr.

Edit: It seems I missunderstood CatLoversG. I'm happy if your refactor reproduces the bug.
Maybe adding a failing testcase that catches the bug would be nice? (mark with .failing):
```ts
it.failing("should not spawn team members out of thin air") {...}
```
😉

src/Bladeburner/Actions/TeamCasualties.ts
d0sboots: FWIW I agree with the stance of having this being refactor-only, and doing any bugfixes either before or after.

d0sboots:
> I have to rewrite this entire feedback after rereading everything due to your force-push, so there may be some copy-paste errors.

I strongly agree with this part in particular. Please don't force-push (which comes from rebasing) during PRs, it messes up the review flow. There is no point in trying to make the individual commits "pretty"; everything will be squashed together into a single commit at the end anyway. It is better to think of the commit history of a PR branch as recording the history of evolution of thinking on the feature, because that is what it *is*.

(You can freely rebase as much as you want before publishing the PR, of course)

src/Bladeburner/Actions/TeamCasualties.ts
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * Some actions (Operations and Black Operations) use teams for success bonus
+ * and may result in casualties, reducing the player's hp, killing team members
+ * and killing sleeves (to shock them, sleeves are immortal)
+ *
+ * Caveats: TeamSize = Human Team + Supporting Sleeves set to this action
+ */
+export class TeamCasualties {
+  constructor(private severity: CasualtyFactor, private teamSizeUsed: number, private supportingSleeves: number) {}
```
d0sboots: I had to go look up parameter properties to understand what the heck was going on here.

I'm not a fan of this syntax. Almost everything in TypeScript does not affect the runtime semantics; you can trivially strip it out and the JS program remains unchanged. This is different; if you remove the "private"s here, it means something different. IMO we should explicitly use regular JS syntax here, for readability.

src/Bladeburner/Actions/TeamCasualties.ts
Alpheus: Checked dev, team members not present also get killed off. Will restore original behavior. As d0s mentioned the fix can be a one-liner as a follow-up, plus the test.

Alpheus:
> I like rollOutcome now being overall more readable, but have some suggestions on how to further reorder things slightly. This is just meant as an addition to catloversg's points, he made some good remarks on how to reuse shuffle functions and his note, that your implementation does not reproduce a known bug. I'd suggest keeping the bugged behaviour in your new implementation, and move fixing the bug, which should then be a one-liner, to another Pr.
> 
> Edit: It seems I missunderstood CatLoversG. I'm happy if your refactor reproduces the bug. Maybe adding a failing testcase that catches the bug would be nice? (mark with .failing):
> 
> ```ts
> it.failing("should not spawn team members out of thin air") {...}
> ```
> 
> 😉

I will test cover the intended bugged behavior. it.failing involves double-negation and makes it confusing to understanding what the intended behavior is.

Alpheus:
> > I have to rewrite this entire feedback after rereading everything due to your force-push, so there may be some copy-paste errors.
> 
> I strongly agree with this part in particular. Please don't force-push (which comes from rebasing) during PRs, it messes up the review flow. There is no point in trying to make the individual commits "pretty"; everything will be squashed together into a single commit at the end anyway. It is better to think of the commit history of a PR branch as recording the history of evolution of thinking on the feature, because that is what it _is_.
> 
> (You can freely rebase as much as you want before publishing the PR, of course)

Sorry, I'm used to stacked-diffing shortlived commits and it's muscle memory at this point. Will do my best in the future.

catloversg:
I remember that I did that in some big PRs (save data compression, TS support, etc.). I feel guilty now.

src/Bladeburner/Bladeburner.ts
Alpheus: These two are tricky. I'm not sure if the heuristic of picking X random sleeves from the player should be that deep in casualties.

I agree that there should be a heuristic to highlight that the sleeves killed **are random**, ie. killRandomSleeves, not just killXSleevesInOrder, but for the implementation I'd rather keep in on BB or Player till that mess gets moved somewhere else (I have a feeling I'll be the one doing that anyway).

src/Bladeburner/Actions/TeamCasualties.ts
d0sboots: I agree. This is a case of making the implementation awkward in order to try and do testing.

Note I say "try" because an important precept of unit testing is that you should be testing the *public* interfaces of units, not internal implementation details. This is because the internal details can change, via refactoring; trying to test the internal details is too granular, and also doesn't cover what is actually important, which is the public API. Sometimes you need to expose internal knobs in order to make mocks and make testing easier or possible, but that is different from testing the internals directly.

Alpheus: Regarding the static—I feel like I'm playing devil's advocate here but there isn't a clean solution for this. We have 2 components merging into data flow: the rolls (low, high, size, used), and the outcomes (randomGenerator, deaths, sleeves killed (specifically)).

If I make it static, then the function call has to be curried for it to have segmented return values. If I don't curry it, then the final return injects test-only data into the production code. The class was a sort of middle ground as d0s already commented on this, but I'm afraid I can't make both of you happy.

I disagree on the garbage collection aspect, this calculation is no different than using objects for actionId or returning [nextState, prevState] object arrays in corporations. The _state_ object is a byproduct of the actionObject, which gets created once per action, with relevant bits cloned. This is one of those "cloned" bits.

The separation of the interfaces is specific to allow the random generator to be injected without mocking at test time as the randomness of things in Bladeburner are notoriously error-prone.

d0sboots:
> Sorry, I'm used to stacked-diffing shortlived commits and it's muscle memory at this point. Will do my best in the future.

This also work out if you don't reflexively push all the time XD
(I can't stand to push "messy" commits either, and compromise by tidying up my work before pushing the latest commit)

src/Bladeburner/Actions/TeamCasualties.ts
Alpheus: Is this a blocker or a nit?

Imho if we don't want this syntax we should automate the check in the linter. The JS target we are using will convert this to javascript in the form of private, non-enumerable properties. Natural JS syntax would involve using hashed properties to denote them private, which goes against the scoping in typescript (as you can make them public).

src/Bladeburner/Actions/TeamCasualties.ts
d0sboots: Hmm... I feel a little differently. My initial gut reaction is that we *should* just be mocking random(), providing a series of values, expecting them all to be used, and expecting a certain result from Bladeburner. This is what I meant by "you might need to poke some internal details to do testing, but you should test the external interface."

d0sboots:
I realized we should take a step back and probably agree on what we're trying to accomplish here.

IMO the goals and non-goals are: Goals:
* Make bladeburner code easier to read.
* Improve bladeburner test coverage.

Non-goals:
* Split up bladeburner.ts.
  * It is very possible (even likely) that this needs to happen in order to make the code easier to read, but it's not a goal in and of itself.
* Increase testing coverage at the expense of code readability.
  * Our overall test coverage is too poor for this to be a goal. There's so much low-hanging fruit in terms of testability, there's no reason to sacrifice readability here to get a little bit more in this particular area.

src/Bladeburner/Actions/TeamCasualties.ts
catloversg: I agree with d0sboots about the mocking of Math.random. In its current state, the custom random function is only useful for testing while it reduces the readability. I don't think it's worth the trade-off.

Alpheus: I don't follow. I'm replacing the outside interface here, not the internal one. _Random_ is the internal bit. I didn't want to expose _how_ Math.random (float) gets converted to a range. Which is why I allowed the higher-cohesion unit to be injected and stubbed (getRandomIntInclusive) rather than Math.random.

On the topic of testing the *public* interfaces, I disagree. The codebase is not in a state where each publicly accessible component can be isolated and controlled. We'll get there, but it may take a few hundred PRs. For the time being I'm increasing what little coverage we can get in lean pieces of code **without explicitly touching** the bladeburner public API, so that these two worlds won't mix.

I find it difficult to even talk about public/internal/external APIs when for most classes all accessors are public and there's little notion of where the boundaries are (case in point: Player, Sleeves, Bladeburner).

But anyway—what are we doing about this? I can give it another design round to try to make everyone happy.

src/Bladeburner/Actions/TeamCasualties.ts
Alpheus: I checked the codebase and the only one other place constructor promotion is used is
../Casino/CardDeck/Deck.ts
8:  constructor(private numDecks = 1) {

So since this is "rare" in this codebase, I'm happy to drop it in favor of an object parameter. Though we're redesigning this bit anyway by the looks of it.

src/Bladeburner/Actions/TeamCasualties.ts
Alpheus: Okay, I see the conversation diverged a bit. There's 4 issues:
- the surface area of the ROLLING random generator
- the surface area of the SHUFFLING SLEEVES
- the surface area of the Casualties constructor (and calls)
- whether to keep this single-function behavior in a class or function (static or no)

Which one here is most important to get this PR moving forward?

src/Bladeburner/Actions/TeamCasualties.ts
Alpheus: Side-note, If there's an open issue on this change, I'd love to reference it in the future PR so we can keep the discussions separate.

src/Bladeburner/Actions/TeamCasualties.ts
cmsautter: > I don't follow. I'm replacing the outside interface here, not the internal one. _Random_ is the internal bit. I didn't want to expose _how_ Math.random (float) gets converted to a range. Which is why I allowed the higher-cohesion unit to be injected and stubbed (getRandomIntInclusive) rather than Math.random.

Why not use and mock getRandomIntInclusive then?
Same for the shuffle function.

Alpheus:
> I realized we should take a step back and probably agree on what we're trying to accomplish here.
> 
> IMO the goals and non-goals are: Goals:
> 
> * Make bladeburner code easier to read.
> * Improve bladeburner test coverage.
> 
> Non-goals:
> 
> * Split up bladeburner.ts.
>   
>   * It is very possible (even likely) that this needs to happen in order to make the code easier to read, but it's not a goal in and of itself.
> * Increase testing coverage at the expense of code readability.
>   
>   * Our overall test coverage is too poor for this to be a goal. There's so much low-hanging fruit in terms of testability, there's no reason to sacrifice readability here to get a little bit more in this particular area.

This is very helpful.

I have a few PRs lined up that split out things like city changes and BB console (it's very isolated to begin with). Is that a reasonable goal for a different kind of PR or are these non-goals global for the foreseeable future?

src/Bladeburner/Actions/TeamCasualties.ts
cmsautter: > If I make it static, then the function call has to be curried for it to have segmented return values. If I don't curry it, then the final return injects test-only data into the production code. The class was a sort of middle ground as d0s already commented on this, but I'm afraid I can't make both of you happy.

Here I cannot really follow. I would't curry here at all, have a simple function that returns deaths and a list / a count of all damaged sleeves. To test the function, just create an object that implements the Operation team interface , and mock random/ getRandomIntInclusive / shuffle.

cmsautter: What I mean is giving it the following signature:
```ts
static TeamCasualtiesRollOutcome(team: OperationTeam, severity: CasualtyFactor, teamSizeUsed: number, supportingSleeves: List<SleeveObject>) {
... 
    return {deadPeople, deadSleeves } 
}
```
Or, to simplify things further, put teamSizeUsed and supportingSleeves inside the OperationTeam interface.
Maybe even make it a class, put the roll function in there, to stay OO. It would mean more refactoring work though, so this could be a separate pr.

src/Bladeburner/Bladeburner.ts
cmsautter: I meant that casualties should pick random sleeves from the team thats in the operation.
You already pick random people from that team in casualties. The only difference is, that they are just a number / annonym, while the sleeves are objects.

src/Bladeburner/Actions/TeamCasualties.ts
Alpheus: That path already got rejected. And we did the mock version as well in the first two comits, the mock got rejected as well.

Look I appreciate what you guys are doing but it gets tricky when all of a sudden everyone passionately shows up and wants to chip in with sentimental nits or personal preferences.

I presented a working solution with a very small compromise. Non of your suggestions are compromise-free.

Ie. if I follow your suggestion, then the method goes from 3 parameters to 4, and to 5 if I hook in the random or the mock. If I replace that with an object because the method has too many parameters I end up here again.

I'll zoom out and consider an alternative approach.

src/Bladeburner/Bladeburner.ts
Alpheus: What's the problem with that? I may as well move this to SleevePerson or Person.

cmsautter:
> Goals:
> 
>     * Make bladeburner code easier to read.
>     * Improve bladeburner test coverage.

I'd say there should be only the first goal:
Write programs for humans, not machines. 
(With the exception of performance crititcal parts.)

I'd see test coverage separatly. Refactors, or new code, should be test-covered of course. 
Also sometimes you need to refactor to make a part testable, or to fix bugs.
But like you said, it must not make the code harder to read.

Alpheus:
> > Goals:
> > ```
> > * Make bladeburner code easier to read.
> > * Improve bladeburner test coverage.
> > ```
> 
> I'd say there should be only the first goal: Write programs for humans, not machines. (With the exception of performance crititcal parts.)
> 
> I'd see test coverage separatly. Refactors, or new code, should be test-covered of course. Also sometimes you need to refactor to make a part testable, or to fix bugs. But like you said, it must not make the code harder to read.

I disagree. We should strive to make it as readable as possible as long as it is testable. If we keep comparing PRs to the readability of untested code then we keep going back to statics and public-everything interfaces. Those are cute, but partly the maintenance burden stems from there being too large hubs of dependencies centered around the Bitnode processes. Those have to be split up. There is no "testable and more readable" without splitting up the largest files, of which Bladeburner is a heavy contender.

cmsautter:
> > ```ts
> > it.failing("should not spawn team members out of thin air") {...}
> > ```
> 
> I will test cover the intended bugged behavior. it.failing involves double-negation and makes it confusing to understanding what the intended behavior is.

I see that. You can use skip to show that a test currently fails, but should not fail.
The argument against it.skip would be, that it doesn't then prove that your refactor shows the same bugged behaviour.
I think that having it it.failing until the very last commit, where you change it to it.skip, could be the optimal approach.

Alpheus:
> > > ```ts
> > > it.failing("should not spawn team members out of thin air") {...}
> > > ```
> > 
> > 
> > I will test cover the intended bugged behavior. it.failing involves double-negation and makes it confusing to understanding what the intended behavior is.
> 
> I see that. You can use skip to show that a test currently fails, but should not fail. The argument against it.skip would be, that it doesn't then prove that your refactor shows the same bugged behaviour. I think that having it it.failing until the very last commit, where you change it to it.skip, could be the optimal approach.

this is totally out of scope.

cmsautter:
> > I'd say there should be only the first goal: Write programs for humans, not machines. (With the exception of performance crititcal parts.)
> > I'd see test coverage separatly. Refactors, or new code, should be test-covered of course. Also sometimes you need to refactor to make a part testable, or to fix bugs. But like you said, it must not make the code harder to read.
> 
> I disagree. We should strive to make it as readable as possible as long as it is testable. If we keep comparing PRs to the readability of untested code then we keep going back to statics and public-everything interfaces. Those are cute, but partly the maintenance burden stems from there being too large hubs of dependencies centered around the Bitnode processes. Those have to be split up. There is no "testable and more readable" without splitting up the largest files, of which Bladeburner is a heavy contender.

Sorry, but I don't follow. I believe we have the same goals, but you are throwing a few things around here; did you quote the right thing? 
My opinion to splitting up bladeburner is that it naturally follows as soon as you identify logical sub-components of what it is doing. The team of an operation could be one of those things. It consists of multiple members in bladeburner that could be packed into a new class. This then automatically improves readability. Same with static functions: If they operate on a certain datatype, they should be part of a class. Also improves readability. Public-Everything makes it easier to **write**/sketch code, but not to **read** it. It's really helpful when you know which class members are private and thus internally managed, and not part of the "api"/"interface" it provides.

cmsautter:
> this is totally out of scope.

I can see that this was quite a specific thing that doesn't belong to this conversation. I just noticed that you cannot really comment on comments, so next time I'll make this a separate review point, i.e. at this empty line were we could have another test.

src/Bladeburner/Actions/TeamCasualties.ts
cmsautter: > That path already got rejected. And we did the mock version as well in the first two comits, the mock got rejected as well.

Sorry if that is the case, but I cannot find the rejection. Can you provide a link to that? I somehow lost the overview here, too many code reviews. I also struggle to read them, since somehow the code lines are messed up. I could imagine the force pushes overwrote the reviewed commit on the server.

Alpheus:
> > > I'd say there should be only the first goal: Write programs for humans, not machines. (With the exception of performance crititcal parts.)
> > > I'd see test coverage separatly. Refactors, or new code, should be test-covered of course. Also sometimes you need to refactor to make a part testable, or to fix bugs. But like you said, it must not make the code harder to read.
> > 
> > 
> > I disagree. We should strive to make it as readable as possible as long as it is testable. If we keep comparing PRs to the readability of untested code then we keep going back to statics and public-everything interfaces. Those are cute, but partly the maintenance burden stems from there being too large hubs of dependencies centered around the Bitnode processes. Those have to be split up. There is no "testable and more readable" without splitting up the largest files, of which Bladeburner is a heavy contender.
> 
> Sorry, but I don't follow. I believe we have the same goals, but you are throwing a few things around here; did you quote the right thing? My opinion to splitting up bladeburner is that it naturally follows as soon as you identify logical sub-components of what it is doing. The team of an operation could be one of those things. It consists of multiple members in bladeburner that could be packed into a new class. This then automatically improves readability. Same with static functions: If they operate on a certain datatype, they should be part of a class. Also improves readability. Public-Everything makes it easier to **write**/sketch code, but not to **read** it. It's really helpful when you know which class members are private and thus internally managed, and not part of the "api"/"interface" it provides.

We don't need to pack the data in a class. For the sake of reviving these bitnode hubs from a savegame, the best place for data is in the core interfaces. That's 99% of what it does. Where we disagree is the seams. I try to separate out behavior, without impacting data flow too much. I don't care for static functions (on classes) as we have reasonable first-class function support.

I disagree on readability, we don't get better readability just by splitting things up. The frequency of change and cohesion must match the intent of the reader. This is very difficult.

For example, in this PR, most of what "is changing" is limited to 3 files which are very close together. That's intentional. Most of what we're doing here on these refactor PRs are temporary at best. We're refactoring semi-testless to be able to get into a minimal state where testing is sensible (ie. the tests don't require 20 mocks each). Only once we get there with a full control of all moving ports and interfaces, can we make larger architectural decisions.

src/Bladeburner/Actions/TeamCasualties.ts
d0sboots: > Look I appreciate what you guys are doing but it gets tricky when all of a sudden everyone passionately shows up and wants to chip in with sentimental nits or personal preferences.

I've been trying to keep my comments down for this reason. XD

There's some stuff I like, and some stuff I don't, but I don't have a compromise-free solution to suggest either. I'll try to break some more time free later today and see if I can come up with a more positive proposal, as opposed to "I don't like some of this." :D

Alpheus:
I've got some time to work on this again.
Here's what I'm going to do:

- I'm widening the testing surface to interface from Bladeburner.completeAction
- I'll curry the casualties functionality, ditching the class and move them to BlackOperations / Operations
- I'll inject the roll itself as a seed rather than injecting the random. From what I understand Math.random cannot be seeded and I'd rather not mock it. This is the simplest (least bad) alternative.
- I'll "visit" the casualties check on the action itself, to inject only into the relevant subtypes (this is new).

Alpheus:
New update is up. It got somewhat larger unfortunately, but it did address most feedback so far.

src/Bladeburner/Bladeburner.ts
```diff
@@ -1041,26 +1031,10 @@ export class Bladeburner {
...
-          }
-          this.teamLost += losses;
-          if (this.logging.blackops) {
-            this.log(
-              `${person.whoAmI()}:  You lost ${formatNumberNoSuffix(losses, 0)} team members during ${action.name}.`,
-            );
-          }
+        if (this.logging.blackops && deaths > 0) {
+          this.log(
+            `${person.whoAmI()}:  You lost ${formatNumberNoSuffix(deaths, 0)} team members during ${action.name}.`,
```
Alpheus, line 1037: I always wondered... is this double-space intentional?

cmsautter:
Not a full review yet, but one of your comments made me remember a small detail:
>  /** Best case is always no deaths */

That's actually not the case. This is (the?) one constant where the two 'duplicates' differ:

https://github.com/bitburner-official/bitburner-src/blob/ade92ed29750215f1f3cbdf4913717f98f7050ff/src/Bladeburner/Bladeburner.ts#L763

https://github.com/bitburner-official/bitburner-src/blob/ade92ed29750215f1f3cbdf4913717f98f7050ff/src/Bladeburner/Bladeburner.ts#L1046

I hope to find some time tomorrow to do a full review.

src/Bladeburner/Bladeburner.ts
```diff
@@ -50,10 +49,12 @@ import { GeneralActions } from "./data/GeneralActions";
 import { PlayerObject } from "../PersonObjects/Player/PlayerObject";
 import { Sleeve } from "../PersonObjects/Sleeve/Sleeve";
 import { autoCompleteTypeShorthand } from "./utils/terminalShorthands";
+import type { OperationTeam } from "./Actions/TeamCasualties";
+import { shuffleArray } from "../Infiltration/ui/BribeGame";
```
d0sboots, line 53: This probably should go into its own helper, but that can also wait for later if you want.

src/Bladeburner/Bladeburner.ts
d0sboots: I'd assume it is, although if it bugs you I wouldn't mind if you changed it.

Honestly, I'd be OK with completely unifying the logging between the two versions and ironing out the minor differences in wording, to help factor out more stuff. But the 0 vs 1 discrepancy that Marvin brought up throws a big wrench into things. (OK, I guess it wasn't that big.)

src/Bladeburner/Actions/TeamCasualties.ts
```diff
@@ -0,0 +1,47 @@
...
+  /** Supporting Sleeves take damage when they are part of losses,
+   *   e.g. 8 sleeves + 3 team members with 4 losses -> 1 sleeve takes damage */
+  team.killRandomSupportingSleeves(damagedSleeves);
+
+  /** Clamped, bugfix for PR#1659
+   * "BUGFIX: Wrong team size when all team members die in Bladeburner's action" */
+  team.teamSize = Math.max(team.teamSize - humanDeaths, team.sleeveSize);
+  team.teamLost += deaths;
+
+  return { deaths, team, damagedSleeves };
```
d0sboots: I continue to have the same feelings about the return type of this function that I did before: The only thing it should actually be returning is `deaths`. The other bits are test-only, and *by definition* test-only data should not be returned: If it is being used for tests, it means you are testing code paths that are irrelevant to the real code, and therefore you have useless tests. (Possibly worse than useless, because they might be giving you false security.) If you *aren't* using them for tests, then they're just completely irrelevant.

`team`, in particular, is one of the parameters and so has no need to be part of the output. And the proper way to capture `damagedSleeves` if you want to verify its value in tests is to turn `killRandomSupportingSleeves` into a spy - this also verifies that it is being called correctly.

Alpheus:
> Not a full review yet, but one of your comments made me remember a small detail:
> 
> > /** Best case is always no deaths */
> 
> That's actually not the case. This is (the?) one constant where the two 'duplicates' differ:
> 
> https://github.com/bitburner-official/bitburner-src/blob/ade92ed29750215f1f3cbdf4913717f98f7050ff/src/Bladeburner/Bladeburner.ts#L763
> 
> https://github.com/bitburner-official/bitburner-src/blob/ade92ed29750215f1f3cbdf4913717f98f7050ff/src/Bladeburner/Bladeburner.ts#L1046
> 
> I hope to find some time tomorrow to do a full review.

Oh wow. I never spotted this. That's an easy fix, thanks!

src/Bladeburner/Actions/BlackOperation.ts
```diff
@@ -60,4 +62,8 @@ export class BlackOperation extends ActionClass {
   getTeamSuccessBonus = operationTeamSuccessBonus;
 
   getActionTypeSkillSuccessBonus = operationSkillSuccessBonus;
+
+  getTeamCasualtiesRoll = getRandomIntInclusive;
+
+  resolveTeamCasualties = resolveTeamCasualties;
```
d0sboots: This is more interface complexity that exists only for testing.

`getTeamCasualtiesRoll` obfuscates the intent a fair amount (it gets plumbed through a class and an interface to ultimately be a simple call to `getRandomIntInclusive`. However, this is balanced by the fact that it would be a bit trickier to mock around without it.

AFAICT, `resolveTeamCasualties` isn't even *used* as a test injection point, and is thus completely unneeded. If you got rid of both of these variables, you could also completely remove the `TeamActionWithCasualties` interface and `resolveTeamCasualties` could be a simple function that takes `teamCount` as a number instead of semi-awkwardly through `this`. It would also be just as straightforward to mock, if you ever needed/wanted to in the future.

EDIT: Just saw your change to fix the 0/1 issue. I guess the interface has a stay on execution. XD You *could* pass that in as another free parameter to the function, but that's getting more ugly... it tips the balance back towards it being worth it to have an interface. I still don't think you need the instance variable for the function here, though. It can just be a free function that takes a `TeamActionWithCasualties` as first parameter. (Normally you would just put it on the base class, but that doesn't work here when you don't have a base class for just this pair of classes.)

Alpheus:
> Not a full review yet, but one of your comments made me remember a small detail:
> 
> > /** Best case is always no deaths */
> 
> That's actually not the case. This is (the?) one constant where the two 'duplicates' differ:
> 
> https://github.com/bitburner-official/bitburner-src/blob/ade92ed29750215f1f3cbdf4913717f98f7050ff/src/Bladeburner/Bladeburner.ts#L763
> 
> https://github.com/bitburner-official/bitburner-src/blob/ade92ed29750215f1f3cbdf4913717f98f7050ff/src/Bladeburner/Bladeburner.ts#L1046
> 
> I hope to find some time tomorrow to do a full review.

fixed.

src/Bladeburner/Bladeburner.ts
```diff
@@ -750,32 +751,20 @@ export class Bladeburner {
     }
   }
 
+  public killRandomSupportingSleeves(n: number) {
```
d0sboots: Isn't `public` the default? Does it need to be stated?

src/Bladeburner/Bladeburner.ts
```diff
@@ -750,32 +751,20 @@ export class Bladeburner {
...
+    const sup = [...Player.sleevesSupportingBladeburner()]; // Explicit shallow copy
+    shuffleArray(sup);
+    sup.slice(0, Math.min(sup.length, n)).forEach((sleeve) => sleeve.kill());
```
d0sboots, line 758: Optional: I find it easier to read the mutation action in non-functional style with `for..of`, where the `sleeve.kill()` would end up on its own line instead of all mushed together with `forEach`, but it's more of a preference thing.

src/Bladeburner/Bladeburner.ts
d0sboots: Optional: This function doesn't depend on *any* data in Bladeburner. (I.e. no `this` or anything like that.) Does it need to be part of Bladeburner? Does somewhere else make more sense?

d0sboots (commented):
I've got some nits, but overall I'm pretty happy with the flow of this. Sorry it took me so long to give it the full review it deserved.

src/Bladeburner/Bladeburner.ts
Alpheus: By the looks of it we may have a second pass to go over things after this is merged. I'd rather get this one into dev and then do another smaller pass. The PR is getting too stale.

src/Bladeburner/Bladeburner.ts
Alpheus: Agreed. I'd prefer to not touch non-BB refactors at the moment. Next pass?

src/Bladeburner/Actions/TeamCasualties.ts
Alpheus: Clumsiness on my end, these all became obsolete as I'm testing shock directly.

src/Bladeburner/Actions/BlackOperation.ts
Alpheus: > AFAICT, resolveTeamCasualties isn't even used as a test injection point, and is thus completely unneeded. If you got rid of both of these variables, you could also completely remove the TeamActionWithCasualties interface and resolveTeamCasualties could be a simple function that takes teamCount as a number instead of semi-awkwardly through this. It would also be just as straightforward to mock, if you ever needed/wanted to in the future.

It won't work. It needs teamCount, sleeveSize and teamSize (not to be confused with teamCount).  I tried it and this is the minimal complexity I've been able to scale it down to. It's **so easy** to confuse teamCount and teamSize and I wanted it to be handled explicitly. A narrower type (the interface) but generously wide instance (Bladeburner, Action) to make it user-friendly on the developer side who's working on CompleteAction.

I tried mocking it. But ended up with a lot more complexity (and automocking on the module level which made the tests much slower).

However, if you insist I _could_ make it a public module export on TeamCasualties and then mock that, but then the tests get exposed to much more "implementation detail" that isn't captured by the interface. It's a hard balance. Anything touching entropy needs to be controlled.

**Okay, a bit more context**
I need to be able to control the random range to test it properly. If we don't want this interface shenanigans, then I'd rather throw away the tests that are flaky when random. But that means we lose 2/3rds of the tests. The issue isn't Math.random or something similar. The problem is that our numbers cause the rolls to be exhaustive at random. It's exhaustive if casualties are 0% or 100%, and the ranges can randomly range to 50%-100% or 0%-100% depending on what the attempt roll was. This makes it incredibly difficult to test, even with monte-carlo simulations. I either need control over an non-exhaustive range **or** a forced scope to the roll (MIN,MAX). I opted for the latter, but it requires **one** injection point in production scope.

Now, I'm aware you may consider injection scope on an interface more invasive than on a package module, but from the perspective of production code, **there will be** a public function somewhere to roll the range (either exported, or on a class/interface). There is no way avoiding having it be "mentioned" in the final output. If this is a huge blocker, I'd much rather remove the tests, than try to pretend we're not affecting public scope. As I mentioned in the earlier round of reviews, this is one of those problems that cannot be **without compromise**. That's the nature of random elements.

src/Bladeburner/Bladeburner.ts
Alpheus: I'll skip this, possibly reviewing on a second pass (different PR).

src/Bladeburner/Bladeburner.ts
Alpheus: Hm.... Yes it needs to be here. It's hard to spot but the coupling is **efferent** (going out) to the _Player_ global static.

There's a dependency loop here: Player has a Bladeburner instance. The bladeburners have actions. The actions are performed by a person (sleeve or player). The sleeves have a player owner (not referenced). The sleeve kills need to be performed by the player owning _this_ bladeburner instance. This is not referenced so the efferent coupling is implied. It's very difficult to spot without a major refactor (also causes babel to shit itself if you import them in the wrong order).

Should we at some point resolve the usage of the static global we'd have to inject the PlayerObject explicitly (not the person performing the action, as sleeve actions cannot have a teamCount > 0). There isn't a simple way of handling that right now without making a mess, but moving this out of here will make it more difficult to do so.

I have a PR for completeAction test/refactor lined up that does address this to some degree.

src/Bladeburner/Actions/BlackOperation.ts
Alpheus: Edit: As a compromise if you want I can put the roll on Bladeburner instead (on the OperationTeam interface). That way it requires less care on the Action data side.

d0sboots: OK, as you said it's a compromise but I like this compromise better.

However, I feel like you didn't see the other part of my comment? Specifically, regarding the `resolveTeamCasualties` function itself, that it doesn't need to be part of the interface but can be a free function that takes an interface object as a param. And that should be just as easy to mock (you can mock out the free function in the tests, if needed.)

src/Bladeburner/Bladeburner.ts
d0sboots: You marked this as resolved, but nothing changed here, did you forget?

Alpheus: Human error.

src/Bladeburner/Actions/BlackOperation.ts
Alpheus: I misunderstood. I thought you meant the resolve as being part of the interface, not the function declaration being referenced and obsolete after the 0/1 fixed gave legit purpose for the namespace. Let me re-read.

Alpheus: I did comment on it, quoted below your original. You mean to have it be a global function? Doesn't that make the invocations more verbose?

Alpheus: Oh, you mean it can _stay_ on the class, but not be part of the interface? Yeah it seems we lost that structural need somewhere during refactoring. I'll try it, see how it looks.

Edit: Yeah, that's completely benign. Done!

Alpheus: _I still see you referring to an object with an interface but I'm not quite sure what that's highlighting. Could you give me a quick example or code pin?_

Alpheus: I _think_ I got what you meant, let me know!

d0sboots: Yup, that was what I was talking about!

PR #1655, opened 2024-09-16
MISC: Update blood donation

hydroflame:


d0sboots:
I thought the test failing was a bug, but it's not:

https://github.com/bitburner-official/bitburner-src/blob/dev/src/PersonObjects/Player/PlayerObjectGeneralMethods.ts#L96
```js
this.money = 1000 + CONSTANTS.Donations;
```

The line to fix is in `test/jest/__snapshots__/FullSave.test.ts.snap`.

d0sboots:
catlover took care of the test update.

PR #1656, opened 2024-09-17
CCT: Add new Square Root coding contract

d0sboots:
To make this simpler, there is now a general-purpose util for getting random bigints. I don't know if anyone else will use it, but it makes this code simpler.

I also rejiggered the type-checking on contracts a little more, because the previous refactor didn't quite catch all the mistakes that could be made.

## Testing

Tested adding extra keys, dropping needed keys, and leaving `getData` out when it was needed. The typechecking catches all of these, as long as the `satisfies` conditions are used right.

The new contract functions properly (tested hundreds of iterations on dummy contracts). The nature of the CCT inherently tests edge cases more.

### DIFF
--- a/src/CodingContracts.ts
+++ b/src/CodingContracts.ts
@@ -1,20 +1,12 @@
 import type { FactionName } from "@enums";
-import { codingContractTypesMetadata, type CodingContractType } from "./data/codingcontracttypes";
+import { codingContractTypesMetadata } from "./data/codingcontracttypes";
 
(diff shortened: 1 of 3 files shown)


Alpheus:
Is this considered a balance change?

d0sboots:
> Is this considered a balance change?

Why would it be? A CCTs are generated at random, this doesn't affect that rate.

Alpheus:
> > Is this considered a balance change?
> 
> Why would it be? A CCTs are generated at random, this doesn't affect that rate.

Looks good then.

PR #1657, opened 2024-09-18
BUGFIX: Change generation rate of CCTs

### LINKED ISSUE #1629: Infinite loading when date is too far in the past
[bitburnerSave_777811909_BN1x1.json.gz](https://github.com/user-attachments/files/16740969/bitburnerSave_777811909_BN1x1.json.gz)

I changed the date on the pc to 1994, trying to get an achievement, and got an infinite loading
this save file causes the same problem in the web version

Bitburner v2.6.2 (216bf616)

image

catloversg:
When the game loads a save file that has too much bonus time, it tries to generate a massive amount of CCTs. This may take a very long time, and the player will think that the game hangs forever. This PR fixes this problem by changing how we generate CCTs at loading time.

New change: We have a new threshold: 30 days. If the offline time is below this threshold, we generate CCTs as normal; otherwise, we use a new formula to calculate the number of "bonus" CCTs.

Current generation rate: 25% chance of 1 CCT per 10 minutes. Average: 1.5 CCT/hour.

We have 2 formulas:
Current [1]:
$$y = 24 \ast 1.5 \ast x$$
New [2]:
$$y = \frac{6500}{1 + e ^ {-0.02 \ast x + 2.2}}$$

Graph:
Capture

Desmos link: https://www.desmos.com/calculator/pxm0lgovvp

Rationale behind the new change:
- No change in the generation rate when the player is online or when the offline time is below the threshold, so there is no need to debate about the balance.
- After being offline for 1 month (~30 days), the player will receive ~1100 bonus CCTs in [1]. Having 1100 bonus CCTs is a decent offline bonus.
- The cap of [2] is 6500. 6500 CCTs is the amount of CCTs that the player will receive if they are online for half a year **without** resetting. There is no point in rewarding them more than that. Generating 6500 CCTs is also very fast. On my 8-year-old machine, it takes only ~300ms to do that.

Fixes #1629.

### DIFF
--- a/src/CodingContractGenerator.ts
+++ b/src/CodingContractGenerator.ts
@@ -14,6 +14,24 @@ import { BaseServer } from "./Server/BaseServer";
 
 import { getRandomIntInclusive } from "./utils/helpers/getRandomIntInclusive";
 import { ContractFilePath, resolveContractFilePath } from "./Paths/ContractFilePath";
+import { clampNumber } from "./utils/helpers/clampNumber";
+
+export function tryGeneratingRandomContract(numberOfTries: number): void {
+  // We try to generate a contract every 10 minutes. 525600 is the number of tries in 10 years. There is no reason to
+  // support anything above that.
+  numberOfTries = clampNumber(0, 525600);
+  let currentNumberOfContracts = GetAllServers().reduce((sum, server) => {
+    return sum + server.contracts.length;
+  }, 0);
+  for (let i = 0; i < numberOfTries; ++i) {
+    const random = Math.random();
+    if (random > 0.25 / (1 + Math.exp(0.0012 * currentNumberOfContracts - 6))) {
+      continue;
+    }
+    generateRandomContract();
+    ++currentNumberOfContracts;
+  }
+}
 
 export function generateRandomContract(): void {
   // First select a random problem type
--- a/src/CodingContractGenerator.ts
+++ b/src/CodingContractGenerator.ts
@@ -17,8 +17,11 @@ import { ContractFilePath, resolveContractFilePath } from "./Paths/ContractFileP
 import { clampNumber } from "./utils/helpers/clampNumber";
 
 export function tryGeneratingRandomContract(numberOfTries: number): void {
-  // We try to generate a contract every 10 minutes. 525600 is the number of tries in 10 years. There is no reason to
-  // support anything above that.
+  /**
+   * We try to generate a contract every 10 minutes. 525600 is the number of tries in 10 years. There is no reason to
+   * support anything above that. We tested this number (525600) on a very old machine. It took only 300-350ms to
+   * loop 525600 times and generate ~9137 contracts on that machine.
+   */
   numberOfTries = clampNumber(Math.floor(numberOfTries), 0, 525600);
   if (numberOfTries < 1) {
     return;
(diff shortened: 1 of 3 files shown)


d0sboots:
A couple of design comments:
* I think this should be in the online calculation, not the offline one. (And the offline calculation should just fall through to online calculation.) This both prevents duplication of code and offline-only bugs/behavior, but also fixes the issue where people will adjust the date (setting it to 0 is common when doing exploits, or if you messed up) and then the game freezes attempting to do CCTs.
* I'd prefer if the backoff were based on current contracts that exist, rather than elapsed time. That way it would serve to absolutely prevent contract spam in the form of multiple large jumps. But that's a nice-to-have.

catloversg:
New update:
- Unify code of online and offline generation.
- Chance of generation is based on the number of current contracts.

Chance:
Capture1

Desmos link: https://www.desmos.com/calculator/zzdf958b9g

Number of generated contracts if:
- Start generating when there are no contracts on any servers.
- X axis is "month".

Capture2

Raw data:
```
	Old	New
1	1089.2	1072
2	2173	2123
3	3225.8	3140.1
4	4313.3	4047.8
5	5425.6	4773.7
6	6479.9	5297.9
7	7570.4	5696.9
8	8669.5	5959.3
9	9709.1	6187.8
10	10811.2	6391.4
11	11853.2	6523
12	12953.3	6677.5
13	13991.2	6795.8
14	15151.3	6917.8
15	16202.5	7001.3
16	17311.3	7088.1
17	18407.5	7171.6
18	19458.7	7236.7
19	20513.6	7309.8
20	21621.9	7374
21	22690.6	7428.5
22	23675	7484.8
23	24910.1	7528.7
24	25887.2	7584.2
25	27046.5	7629.4
26	28075.1	7668.6
27	29279.4	7699.9
28	30226.4	7753.9
29	31330.3	7789.9
30	32414	7811.7
31	33572.8	7854.7
32	34544.4	7881.2
33	35634.1	7912.9
34	36679.9	7940.8
35	37838.1	7987.8
36	38922.1	8020.2
```

src/CodingContractGenerator.ts
```diff
@@ -14,6 +14,27 @@ import { BaseServer } from "./Server/BaseServer";
...
+  numberOfTries = clampNumber(Math.floor(numberOfTries), 0, 525600);
+  if (numberOfTries < 1) {
+    return;
+  }
+  let currentNumberOfContracts = GetAllServers().reduce((sum, server) => {
+    return sum + server.contracts.length;
+  }, 0);
+  for (let i = 0; i < numberOfTries; ++i) {
+    const random = Math.random();
+    if (random > 0.25 / (1 + Math.exp(0.0012 * currentNumberOfContracts - 6))) {
```
d0sboots: We should have a comment here explaining this formula. Relevant things:
* That the chance is 0.25 for "reasonable" amounts of contracts
* That the total number of contracts goes to an asymptote (and what that value is)
* The time period/general shape of the curve, as an explanation for why this formula was chosen

d0sboots: After analysis, I have a slightly different set of constants to suggest: `100 / (399 + Math.exp(0.0012 * currentNumberOfContracts))`. Primarily to exactly match the probability/slope of 0.25 at 0 contracts.

Analysis also reveals that there *is* no asymptote, the behavior is logarithmic. IMO this is a good thing, it's what I was asking for.

https://www.desmos.com/calculator/zmpohx9eps

src/CodingContractGenerator.ts
```diff
@@ -14,6 +14,27 @@ import { BaseServer } from "./Server/BaseServer";
 
 import { getRandomIntInclusive } from "./utils/helpers/getRandomIntInclusive";
 import { ContractFilePath, resolveContractFilePath } from "./Paths/ContractFilePath";
+import { clampNumber } from "./utils/helpers/clampNumber";
+
+export function tryGeneratingRandomContract(numberOfTries: number): void {
+  // We try to generate a contract every 10 minutes. 525600 is the number of tries in 10 years. There is no reason to
```
d0sboots: I'm assuming this many tries turned out to have reasonable performance?

d0sboots (commented):
I was surprised to see that the number of contracts generated online is fixed at 1. Does that mean that the online issue isn't actually this? If you exploit and change the date while BB is running, does it still hang or does this fix it?

src/CodingContractGenerator.ts
catloversg: Yes. I added the information to the comment in the latest commit.

catloversg:
> I was surprised to see that the number of contracts generated online is fixed at 1. Does that mean that the online issue isn't actually this? If you exploit and change the date while BB is running, does it still hang or does this fix it?

AFAIK, we don't have a problem with the online generation. All reports that I have seen only show the problem of loading a save file. If the player changes the date **while** Bitburner is running, the game will not hang due to the CCT generation.

src/CodingContractGenerator.ts
d0sboots: Awesome detail, ty

PR #1658, opened 2024-09-19
BUGFIX: Wrong team size when all team members die in Bladeburner's action

catloversg:
I mentioned this bugfix in https://github.com/bitburner-official/bitburner-src/pull/1654#pullrequestreview-2310084828.

If all team members die, the team size must be the number of supporting sleeves. However, instead of an assignment, we do an addition.

### DIFF
--- a/src/Bladeburner/Bladeburner.ts
+++ b/src/Bladeburner/Bladeburner.ts
@@ -1052,7 +1052,8 @@ export class Bladeburner {
               sup[r].takeDamage(sup[r].hp.max);
               sup.splice(r, 1);
             }
-            this.teamSize += this.sleeveSize;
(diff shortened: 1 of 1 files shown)


Alpheus:
You forgot to apply this fix to the second duplicate of this function on L762.

I'd ask for a more thorough review next time, at least in line with what's happening in #1654 since you referenced it.

d0sboots:
> You forgot to apply this fix to the second duplicate of this function on L762.
> 
> I'd ask for a more thorough review next time, at least in line with what's happening in #1654 since you referenced it.

On small fixes like this, I'm verifying that it does what it claims and doesn't break stuff, not that it fixes all possible instances of the issue. Checking the former only requires looking at the changed lines (and nearby context), doing the latter requires a thorough audit of *all* the code. The need for the latter is the reason bigger refactors take longer to review.

PR #1659, opened 2024-09-20
BUGFIX: Wrong size of table cell in Import Save Comparison UI

catloversg:
Before:
before

After:
after


d0sboots:
They're also unaligned vertically (even on the new one) and now I can't unsee it XD

catloversg:
New:
Capture

PR #1660, opened 2024-09-21
FEAT: add code navigation to script editor

lucebac:
Adds code navigation by registering a CodeEditorOpener to monaco.

### DIFF
--- a/src/ScriptEditor/ui/ScriptEditorRoot.tsx
+++ b/src/ScriptEditor/ui/ScriptEditorRoot.tsx
@@ -32,7 +32,7 @@ import { useVimEditor } from "./useVimEditor";
 import { useCallback } from "react";
 import { type AST, getFileType, parseAST } from "../../utils/ScriptTransformer";
 import { RamCalculationErrorCode } from "../../Script/RamCalculationErrorCodes";
-import { hasScriptExtension, isLegacyScript } from "../../Paths/ScriptFilePath";
+import { hasScriptExtension, isLegacyScript, resolveScriptFilePath, ScriptFilePath } from "../../Paths/ScriptFilePath";
 
 interface IProps {
   // Map of filename -> code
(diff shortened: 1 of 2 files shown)


src/ScriptEditor/ui/ScriptEditorRoot.tsx
```diff
@@ -42,6 +42,74 @@ interface IProps {
...
+      openScripts.push(newScript);
+      currentScript = newScript;
+      source.setModel(newScript.model);
+    }
+
+    if (selectionOrPosition === undefined) {
+      return true;
+    }
+
+    if ((selectionOrPosition as monaco.IRange) !== undefined) {
```
d0sboots: `as` is a compile-time-only thing, it doesn't do anything at runtime. You'll need to do something like `startLineNumber in selectionOrPosition`

src/ScriptEditor/ui/ScriptEditorRoot.tsx
```diff
@@ -187,19 +255,40 @@ function Root(props: IProps): React.ReactElement {
...
+
+      const script = openScripts.find((s) => s.hostname === hostname && s.path === path);
+      if (script) {
+        parseCode(script.code);
+      }
+
+      rerender();
+      removeOutlineOfEditor();
+    });
+    vim = props.vim;
```
d0sboots, line 282: Why is this changing?

d0sboots (commented):
Sorry I've been so slow here too. The editor is weird and complicated, but this mostly looks OK - aside from one definite issue.

Also, if you could add a screenshot showing this in action, that would be nice.

src/ScriptEditor/ui/ScriptEditorRoot.tsx
lucebac: When using code navigation, monaco might open a file that has not been loaded yet. Originally, `props.vim` is set by the parent ReactNode when calling OnMount; unfortunately, the CodeEditorOpener doesn't know the props and hence I save the most recent value so that new files are opened with the latest `props.vim` setting instead of falling back to a fixed `true` or `false`.

src/ScriptEditor/ui/ScriptEditorRoot.tsx
lucebac: Good to know, thanks for the hint.

PR #1661, opened 2024-09-21
FIX: import resolution in code editor

lucebac:
After switching to the dev build to be able to use Typescript, I noticed that import resolution fails most of the time. Not always, though. This is what it looks like:
image

As you can see in the screenshot, it doesn't matter whether or not a file extension is provided. Additionally, I noticed that if an imported file is opened before the 'importee' file, the import resolution does seem to work fine. Unfortunately, I've not come around with why this happens and what the underlying problem is. I believe it's related to the virtual filesystem which Bitburner implements and that monaco doesn't know how to find source files automatically in there.

As a workaround, I opted to implement a callback which loads all imported files into the typescript worker in the background whenever a model changes. The callback does not reload scripts which are already open, so dirty files which are currently open will not be overwritten.

With the fix in place, import resolution and type annotations for imported types/functions now works correctly:
image
image

### DIFF
--- a/src/ScriptEditor/ui/ScriptEditorRoot.tsx
+++ b/src/ScriptEditor/ui/ScriptEditorRoot.tsx
@@ -162,8 +162,12 @@ function Root(props: IProps): React.ReactElement {
   }
 
   function loadAllServerScripts(): void {
-    let server;
-    if (!currentScript || !(server = GetServer(currentScript.hostname))) {
+    if (!currentScript) {
+      return;
+    }
+
+    const server = GetServer(currentScript.hostname);
+    if (!server) {
       return;
     }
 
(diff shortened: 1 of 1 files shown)


src/ScriptEditor/ui/ScriptEditorRoot.tsx
```diff
@@ -161,6 +161,28 @@ function Root(props: IProps): React.ReactElement {
     }
   }
 
+  function loadAllServerScripts(): void {
+    let server;
+    if (!currentScript || !(server = GetServer(currentScript.hostname))) {
+      return;
+    }
```
d0sboots:
```suggestion
    if (!currentScript) {
      return;
    }
    const server = GetServer(currentScript.hostname);
    if (!server) {
      return;
    }
```
I'd prefer to not set variables inside a condition, it makes it harder to reason about their values.

d0sboots (commented):
Looks good, this was much simpler than I feared. Sorry for being slow.

lucebac:
No worries, all good. I've applied your suggestion as it makes absolute sense to do it that way.

PR #1662, opened 2024-09-21
BUGFIX: Follow-up of #1658 (Wrong team size in Bladeburner)

catloversg:
This PR applies the same fix in #1658 to the duplicate code in `completeOperation`. In a normal situation, I'll refactor those duplicate parts, but I'll leave that task to #1654 in this case.


Alpheus:
Thank you <3

Alpheus (approved):

PR #1664, opened 2024-09-23
BLADEBURNER: test cover console commands

Alpheus:
# Motivation
In line with recent refactoring PRs, this one includes _only_ the test coverage for existing code.
I've noticed certain areas were hiding bugs and I made a few clerical errors along the way. Going forward I'd like to start with test coverage on the old implementation first. 

# New Coverage
- bladeburner console: help for all commands
- bladeburner console: behaviors of action commands
- bladeburner console: command concatenation

### DIFF
--- a/test/jest/Bladeburner/BladeburnerConsole.test.ts
+++ b/test/jest/Bladeburner/BladeburnerConsole.test.ts
@@ -6,52 +6,53 @@ import { Contract } from "../../../src/Bladeburner/Actions";
 
 describe("Bladeburner Console", () => {
   let inst: Bladeburner;
-  const FEEDBACK_CMDS = [["skill"], ["start"], ["help"], ["log"], ["automate"]];
-  const SILENT_CMDS = [["cls"], ["clear"], ["stop"]];
-  const CLEAR_CMDS = [["cls"], ["clear"]];
+
+  function initBladeburner(player: PlayerObject): player is PlayerObject & { bladeburner: Bladeburner } {
+    player.init();
+    player.startBladeburner();
+    return true;
+  }
 
   beforeEach(() => {
     setPlayer(new PlayerObject());
-    Player.init();
-    Player.startBladeburner();
-
-    if (!Player.bladeburner) throw new Error();
-    inst = Player.bladeburner;
-    inst.clearConsole();
+    if (initBladeburner(Player)) {
+      inst = Player.bladeburner;
+      inst.clearConsole();
+    }
   });
 
   it("May concatenate multiple commands with ';'", () => {
     inst.startAction(Contract.createId(BladeburnerContractName.Tracking));
-    bb("stop;start contract Retirement;help cls;");
+    execute("stop;start contract Retirement;help cls;");
     expect(inst.consoleLogs).not.toContainEqual("Invalid console command");
   });
 
-  describe.each(FEEDBACK_CMDS)("%s", (cmd: string) => {
+  describe.each(["skill", "start", "help", "log", "automate"])("%s", (cmd: string) => {
     it("provides valid console feedback", () => {
-      bb(cmd);
+      execute(cmd);
       expect(inst.consoleLogs.length).toBeGreaterThan(0);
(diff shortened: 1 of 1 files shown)


test/jest/Bladeburner/BladeburnerConsole.test.ts
```diff
@@ -0,0 +1,92 @@
+import type { Bladeburner } from "../../../src/Bladeburner/Bladeburner";
+import { Player, setPlayer } from "@player";
+import { PlayerObject } from "../../../src/PersonObjects/Player/PlayerObject";
+import { BladeburnerContractName } from "@enums";
+import { Contract } from "../../../src/Bladeburner/Actions";
+
+describe("Bladeburner Console", () => {
+  let inst: Bladeburner;
+  const FEEDBACK_CMDS = [["skill"], ["start"], ["help"], ["log"], ["automate"]];
```
d0sboots: Why are these all single-element lists? I'm pretty sure if `each` takes a single argument, they don't need to be wrapped like this.

test/jest/Bladeburner/BladeburnerConsole.test.ts
```diff
@@ -0,0 +1,92 @@
...
+  const SILENT_CMDS = [["cls"], ["clear"], ["stop"]];
```
d0sboots: Optional-ish: I think it would be more clear to have these lists directly inside the `each`, so it is more clear what exactly is being tested. This is a general principle in tests; it is often worth it to be more verbose (sometimes to the point of duplication) in order to get extreme clarity.

At the very least, `SILENT_CMDS` is only used once, so really should be inlined.

test/jest/Bladeburner/BladeburnerConsole.test.ts
```diff
@@ -0,0 +1,92 @@
...
+  const CLEAR_CMDS = [["cls"], ["clear"]];
+
+  beforeEach(() => {
+    setPlayer(new PlayerObject());
+    Player.init();
+    Player.startBladeburner();
+
+    if (!Player.bladeburner) throw new Error();
```
d0sboots: It should never happen, but just in case let's put a message in this error.

test/jest/Bladeburner/BladeburnerConsole.test.ts
```diff
@@ -0,0 +1,92 @@
...
+      expect(inst.automateActionHigh).toMatchObject(Contract.createId(BladeburnerContractName.Tracking));
+    });
+
+    it("queues action at low threshold", () => {
+      bb("automate contract Tracking low");
+      expect(inst.automateActionLow).toMatchObject(Contract.createId(BladeburnerContractName.Tracking));
+    });
+  });
+
+  function bb(cmd: string) {
```
d0sboots: Optional: This alias exists only to save typing, it doesn't really reduce the amount of code or help with understanding. I'd lean towards removing it.

d0sboots (commented):
Nice comprehensive tests. Hopefully they aren't adding badly to our test times XD

test/jest/Bladeburner/BladeburnerConsole.test.ts
Alpheus: Hey I wasn't aware of that. Cool!

test/jest/Bladeburner/BladeburnerConsole.test.ts
Alpheus: As you said, this never happens. It's there as a typeguard, I'll flip it over to be an explicit type guard.

test/jest/Bladeburner/BladeburnerConsole.test.ts
Alpheus: I kept the alias for clr in the actual functionality check, and dropped the help check.

test/jest/Bladeburner/BladeburnerConsole.test.ts
Alpheus: It's for brevity to keep the tests quicker to read. But 'bb' is quite obtuse, I'll give it a meaningful name.

PR #1668, opened 2024-09-28
BLADEBURNER: Refactor BB console commands out of Bladeburner.ts

Alpheus:
No NS api changes. Impacted functionality is fully test-covered.

# Motivation
This is the biggest move out of Bladeburner.ts so far.
We're almost sub-1k lines from originally 1500, a solid 25% decrease.

Cohesion has increased considerably with this change as the console commands can all run through a single public API and do not require all the implementation details or help in the main Bladeburner file.

# Review suggestions
The added files in Console/Commands are kept original verbatim except for translating **this.** to **blade.**
Github diff doesn't show that very well.


Alpheus:
I was hoping the tests would expedite this change. Seems I was wrong, will re-do.

PR #1672, opened 2024-09-30
BUGFIX: Wrong calculation in team casualties of Bladeburner action

catloversg:
There are at least 2 regression bugs caused by the refactor in #1654.

Bug 1:

`resolveTeamCasualties` does not short-circuit when `teamCount` is 0.

With BlackOps, `minCasualties` is 1. If `teamCount` is 0, `worstCase` is 0. When we call `team.getTeamCasualtiesRoll(action.getMinimumCasualties(), worstCase)`, it will call `getRandomIntInclusive(1, 0)`. `getRandomIntInclusive` only accepts parameters in the form of `(min, max)`.

Exception

Bug 2:

`humans` is not "clamped" properly. It should be `const humans = Math.max(action.teamCount - team.sleeveSize, 0);`.
If `humans` becomes negative, all following calculations are wrong.
For example: TeamSize = 3; SleeveSize = 3; TeamCount = 0; deaths = 1. In this case:
- `humans` is `-3`.
- `humanDeaths` is `-3`.
- `team.killRandomSupportingSleeves` is called with `1 - (-3) = 4`.
- `team.teamSize` = Math.max(3 - (-3), 3) = 6;

These bugs were introduced because `resolveTeamCasualties` does not reuse the old logic:
- It does not short-circuit when `teamCount` is 0.
- It introduces the new logic of `humans` and `humanDeaths`.

This PR rewrites the entire `resolveTeamCasualties` and reuses the old logic.

For quick reference, this is how it was implemented:

Operation:
```js
const teamCount = action.teamCount;
if (teamCount >= 1) {
  const maxLosses = success ? Math.ceil(teamCount / 2) : Math.floor(teamCount);
  const losses = getRandomIntInclusive(0, maxLosses);
  this.teamSize -= losses;
  if (this.teamSize < this.sleeveSize) {
    const sup = Player.sleeves.filter((x) => isSleeveSupportWork(x.currentWork));
    for (let i = 0; i > this.teamSize - this.sleeveSize; i--) {
      const r = Math.floor(Math.random() * sup.length);
      sup[r].takeDamage(sup[r].hp.max);
      sup.splice(r, 1);
    }
    // If this happens, all team members died and some sleeves took damage. In this case, teamSize = sleeveSize.
    this.teamSize = this.sleeveSize;
  }
  this.teamLost += losses;
  if (this.logging.ops && losses > 0) {
    this.log("Lost " + formatNumberNoSuffix(losses, 0) + " team members during this " + action.name);
  }
}
```

Black Operation:
```js
if (teamCount >= 1) {
  const losses = getRandomIntInclusive(1, teamLossMax);
  this.teamSize -= losses;
  if (this.teamSize < this.sleeveSize) {
    const sup = Player.sleeves.filter((x) => isSleeveSupportWork(x.currentWork));
    for (let i = 0; i > this.teamSize - this.sleeveSize; i--) {
      const r = Math.floor(Math.random() * sup.length);
      sup[r].takeDamage(sup[r].hp.max);
      sup.splice(r, 1);
    }
    // If this happens, all team members died and some sleeves took damage. In this case, teamSize = sleeveSize.
    this.teamSize = this.sleeveSize;
  }
  this.teamLost += losses;
  if (this.logging.blackops) {
    this.log(
      `${person.whoAmI()}:  You lost ${formatNumberNoSuffix(losses, 0)} team members during ${action.name}.`,
    );
  }
}
```

### DIFF
--- a/src/Bladeburner/Actions/TeamCasualties.ts
+++ b/src/Bladeburner/Actions/TeamCasualties.ts
@@ -26,22 +26,33 @@ export interface TeamActionWithCasualties {
  * and may result in casualties, reducing the player's hp, killing team members
  * and killing sleeves (to shock them, sleeves are immortal) *
  */
-export function resolveTeamCasualties(action: TeamActionWithCasualties, team: OperationTeam, success: boolean) {
-  const severity = success ? CasualtyFactor.LOW_CASUALTIES : CasualtyFactor.HIGH_CASUALTIES;
-  const radius = action.teamCount * severity;
-  const worstCase = severity < 1 ? Math.ceil(radius) : Math.floor(radius);
-  /** Best case is always no deaths */
-  const deaths = team.getTeamCasualtiesRoll(action.getMinimumCasualties(), worstCase);
-  const humans = action.teamCount - team.sleeveSize;
-  const humanDeaths = Math.min(humans, deaths);
-  /** Supporting Sleeves take damage when they are part of losses,
-   *   e.g. 8 sleeves + 3 team members with 4 losses -> 1 sleeve takes damage */
-  team.killRandomSupportingSleeves(deaths - humanDeaths);
+export function resolveTeamCasualties(action: TeamActionWithCasualties, team: OperationTeam, success: boolean): number {
+  if (action.teamCount <= 0) {
+    return 0;
+  }
 
-  /** Clamped, bugfix for PR#1659
-   * "BUGFIX: Wrong team size when all team members die in Bladeburner's action" */
-  team.teamSize = Math.max(team.teamSize - humanDeaths, team.sleeveSize);
-  team.teamLost += deaths;
+  // Operation actions and Black Operation actions have different min casualties: Min of Ops = 0. Min of BlackOps = 1.
+  const minCasualties = action.getMinimumCasualties();
+  const maxCasualties = success
+    ? Math.ceil(action.teamCount * CasualtyFactor.LOW_CASUALTIES)
+    : Math.floor(action.teamCount * CasualtyFactor.HIGH_CASUALTIES);
+  /**
+   * In the current state, it's safe to assume that minCasualties <= maxCasualties. However, in the future, if we change
+   * min casualties, LOW_CASUALTIES, or HIGH_CASUALTIES, the call of getTeamCasualtiesRoll may crash.
+   * getTeamCasualtiesRoll is just getRandomIntInclusive, and that function's parameters need to be in the form of
+   * (min, max).
+   */
+  const losses =
+    minCasualties <= maxCasualties
+      ? team.getTeamCasualtiesRoll(minCasualties, maxCasualties)
+      : team.getTeamCasualtiesRoll(maxCasualties, minCasualties);
+  team.teamSize -= losses;
+  if (team.teamSize < team.sleeveSize) {
+    team.killRandomSupportingSleeves(team.sleeveSize - team.teamSize);
+    // If this happens, all team members died and some sleeves took damage. In this case, teamSize = sleeveSize.
+    team.teamSize = team.sleeveSize;
+  }
+  team.teamLost += losses;
 
-  return deaths;
+  return losses;
 }
--- a/src/Bladeburner/Actions/TeamCasualties.ts
+++ b/src/Bladeburner/Actions/TeamCasualties.ts
@@ -43,9 +43,7 @@ export function resolveTeamCasualties(action: TeamActionWithCasualties, team: Op
    * (min, max).
    */
   const losses =
-    minCasualties <= maxCasualties
-      ? team.getTeamCasualtiesRoll(minCasualties, maxCasualties)
-      : team.getTeamCasualtiesRoll(maxCasualties, minCasualties);
+    minCasualties <= maxCasualties ? team.getTeamCasualtiesRoll(minCasualties, maxCasualties) : minCasualties;
   team.teamSize -= losses;
   if (team.teamSize < team.sleeveSize) {
     team.killRandomSupportingSleeves(team.sleeveSize - team.teamSize);
--- a/test/jest/Bladeburner/TeamCasualties.test.ts
+++ b/test/jest/Bladeburner/TeamCasualties.test.ts
@@ -18,7 +18,7 @@ import { PlayerObject } from "../../../src/PersonObjects/Player/PlayerObject";
  */
 describe("Bladeburner Team", () => {
   const MAX_ROLL = (_: number, high: number) => high;
-  const MIN_ROLL = (low: number, _: number) => low;
+  const MIN_ROLL = (low: number, __: number) => low;
   const BLACK_OP = BlackOperation.createId(BladeburnerBlackOpName.OperationAnnihilus);
   const OP = Operation.createId(BladeburnerOperationName.Assassination);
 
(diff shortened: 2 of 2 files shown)


src/Bladeburner/Actions/TeamCasualties.ts
```diff
@@ -26,22 +26,33 @@ export interface TeamActionWithCasualties {
...
+  /**
+   * In the current state, it's safe to assume that minCasualties <= maxCasualties. However, in the future, if we change
+   * min casualties, LOW_CASUALTIES, or HIGH_CASUALTIES, the call of getTeamCasualtiesRoll may crash.
+   * getTeamCasualtiesRoll is just getRandomIntInclusive, and that function's parameters need to be in the form of
+   * (min, max).
+   */
+  const losses =
+    minCasualties <= maxCasualties
+      ? team.getTeamCasualtiesRoll(minCasualties, maxCasualties)
+      : team.getTeamCasualtiesRoll(maxCasualties, minCasualties);
```
d0sboots: I don't think this is the right thing to do in this case. Possible things that would make sense would include "only minCasualties" or "early return 0". (We can decide what makes the most sense, since this doesn't happen currently.)

test/jest/Bladeburner/TeamCasualties.test.ts
```diff
@@ -96,6 +96,15 @@ describe("Bladeburner Team", () => {
   });
 
   describe("Casualties", () => {
+    it.each([[OP], [BLACK_OP]])(
+      "no change in team size when not using team. Action: %s",
```
Alpheus, line 100: Great to see you're continuing the trend. Kuddos!

src/Bladeburner/Actions/TeamCasualties.ts
```diff
@@ -26,22 +26,31 @@ export interface TeamActionWithCasualties {
...
-   *   e.g. 8 sleeves + 3 team members with 4 losses -> 1 sleeve takes damage */
-  team.killRandomSupportingSleeves(deaths - humanDeaths);
-
-  /** Clamped, bugfix for PR#1659
-   * "BUGFIX: Wrong team size when all team members die in Bladeburner's action" */
-  team.teamSize = Math.max(team.teamSize - humanDeaths, team.sleeveSize);
-  team.teamLost += deaths;
-
-  return deaths;
+export function resolveTeamCasualties(action: TeamActionWithCasualties, team: OperationTeam, success: boolean): number {
```
Alpheus, line 29: The refactor revert is over-engineered. Here is the minimal fix needed by following your clamp suggestions in the PR description (insert at pre-change Line number 34):

```ts
  const bestCase = Math.max(0, Math.min(action.teamCount, action.getMinimumCasualties()));
  const deaths = team.getTeamCasualtiesRoll(bestCase, bestCase < worstCase ? worstCase : bestCase);
  const humans = Math.max(0, action.teamCount - team.sleeveSize);
```

But please keep the test, it's the most important piece here.

Alpheus (changes requested):

src/Bladeburner/Actions/TeamCasualties.ts
catloversg: Unless @d0sboots demands, I won't change the implementation of this PR. It uses the old logic and looks very similar to the old code. That's the whole point of the fix.

I also disagree about the "over-engineered" part. For example:
```js
const severity = success ? CasualtyFactor.LOW_CASUALTIES : CasualtyFactor.HIGH_CASUALTIES;
const radius = action.teamCount * severity;
const worstCase = severity < 1 ? Math.ceil(radius) : Math.floor(radius);
```
vs
```js
const maxCasualties = success
    ? Math.ceil(action.teamCount * CasualtyFactor.LOW_CASUALTIES)
    : Math.floor(action.teamCount * CasualtyFactor.HIGH_CASUALTIES);
```
The latter code is how it was implemented in the past, and it looks cleaner.

Alpheus: Your focus on cleanliness I agree with. Let's work with the data: 6 different people missed this bug after 3 separate PRs over 4 weeks. I invite you to use neither this implementation nor the old one if you want to focus on making it more readable.

But also let's focus on making small improvements in the same vein as you suggested your comments of the original PR: either fix the bug or refactor it, but not both.

Personally I prefer the math problem being laid out flat, rather than resulting to control structures and early returns. But that's just my preference (and bias), no need to use my version. However, I invite you not to give into the fear-based option of "the old version worked fine as long as you don't touch it".

If you think the refactor was a fluke that all the reviewers, yourself included, missed then revert that commit instead of gradually reintroducing the old confusion.

catloversg: To make sure that this discussion does not go awry, I'll make this clear:
- I have no hard feelings about the refactor. Please check the last paragraph. Your refactor is a good addition/change.
- I don't want to change the implementation, but it's **not** because I think my code is "better" or anything. I choose to stick with the old implementation because it's bug-free (AFAIK) while still being easy to understand (this is an opinion, not a fact). Please note that I don't have any implication when saying that.

"X people do not find problems for Y <PRs/weeks/months/years>": I have seen many people use this phrase in their arguments in many cases, and it never makes sense to me. One example: OpenSSL, in spite of being an extremely important project and being "reviewed" by innumerable experts, still has nasty bugs. Back in the "era" of Heartbleed, many people asked: How has this bug not been found out when < Insert tons of reasons here > before it's merged? To be honest, I was not surprised at all. "Having many eyes on it" is not the silver bullet.

I know that comparing Bitburner with OpenSSL (or other massive projects) is silly (to make this message succinct, I won't write a useless wall of text about it). The point is that we all make mistakes. We try to minimize the chance of having bugs by writing test cases, keeping the code clean and easy to read, etc. (Please note that I don't imply that your code is buggy or anything, just in case you misunderstand my point). These bugs, which are outlined in this PR, are caused by the new logic, so I reuse the old logic to fix it (more about this in the next paragraph). I pointed out the potential problem of the new logic in https://github.com/bitburner-official/bitburner-src/pull/1654#discussion_r1763441473, but I did not check it again after that, before d0sboots merged it. If there is a person to blame, it's me. There is no need to pull other people in. It's my fault for not checking it again before it's merged, and I'll take the blame for it.

This PR is a bugfix one, and I do that by reusing the old code. The old implementation is clean and easy to understand, so I decided to reuse it, not because "the old version worked fine as long as you don't touch it". This is just my opinion, not a fact, so it's fine if you disagree. We have made our points clear, so I don't think we should argue about which one is "better". I'll let the maintainer decide the implementation. One small point: I did not write those old code, so I have no special "feelings" about them. I'm fine if the maintainer tells me to use X code instead of Y code.

The last point: your refactor is not a "fluke". I have never said or thought about it like that. I have also never said that I would revert it or implement another "better" version. I spotted some bugs, then I made a PR to fix them. No hard feelings. In the past, I have also introduced some nasty bugs in this game (corrupted save data in the "Save compression" PR, "unlimited" SF in the "BN option" PR, etc.) when I submitted my PRs. We spot bugs, we fix them, then we move on to other bugs/suggestions.

Alpheus: I don't mind either way, just sad to see the old control flow reintroduced as it made (in my opinion) the sleeved parts much harder to follow.

d0sboots: Random points:
* When a bug is found in a refactor, a revert (or partial revert) should generally be the go-to response. So in that sense, what CatLover is doing makes perfect sense to me. Especially since our tests are... not the greatest.
* I don't have a particular opinion about either way being easier to read.
* Kudos for keeping this focused on the code.

PR #1677, opened 2024-10-04
CONTRACT: Fix incorrect description of Hamming Codes in Integer to Encoded Binary.

zorbathut:
Note that this brings Integer-To-Binary's description to match Binary-To-Integer's description.

For some reason I don't pretend to understand, the Steam release has the same typo, but *only* in the Encoded Binary to Integer contract. I don't understand how this typo moved before releasing it on Steam (or after releasing it on Steam?)


d0sboots:
It's not exactly a typo, the issue is that referring to positions here with 2^N while using N earlier is confusing. It was halfway in-between. But, this is at least consistent across both now.

PR #1678, opened 2024-10-05
BLADEBURNER: Separate Bladeburner Console Automation from script API (refactor).

Alpheus:
No NS api changes. Impacted functionality is fully test-covered.

### Motivation
Aiming to extract these functions, but would like a review of the scope change first. After review, the extraction will create a cohesive bundle for the Bladeburner Automation Console

### Next Steps for Draft
- [ ] Extract the private static methods to a separate js module.
- [ ] (optional) Provide an interface to Bladeburner to denote all public touch points used by the Console

### Review suggestions
The added files in Console/Commands are kept original verbatim except for translating this. to blade.
Github diff doesn't show that very well.


d0sboots:
I'm not sure that this is actually an improvement in its current form. It seems to just be pulling a lot of functions out as `private static`, which... a) why even do that instead of just having them be free functions? (They'll automatically be "private" be being file-scoped.) b) These seem like they "belong" to the bladeburner class, so extracting them in this way seems anti-OOP.

Alpheus:
> I'm not sure that this is actually an improvement in its current form. It seems to just be pulling a lot of functions out as `private static`, which... a) why even do that instead of just having them be free functions? (They'll automatically be "private" be being file-scoped.) b) These seem like they "belong" to the bladeburner class, so extracting them in this way seems anti-OOP.

Alright

PR #1694, opened 2024-10-16
MISC: Validate bet input of casino mini games

catloversg:
The "bet input" of 3 games (Coin flip, Slot machine, Roulette) has many problems:
- The validation logic is duplicated.
- The validation logic does not check the current player's money.
- When the bet amount is "out-of-range", it's silently changed to the min/max value.
- When the bet amount is invalid, there is no error shown in the UI.
- There is no way to know the max bet amount of a game.

Blackjack does not have these problems, so this PR pulls the "bet input" of Blackjack out and reuses it in other games.

Before:
before

After:
after


src/Casino/Roulette.tsx
```diff
@@ -1,15 +1,15 @@
-import React, { useState, useEffect } from "react";
+import React, { useEffect, useState } from "react";
```
Alpheus, line 1: Hey my IDE does this too!

d0sboots: Expanding on the discussion in #development: I would change your IDE settings to not reorder unchanged lines like this, it adds noise to PRs. No need to revert in this case, though.

src/Casino/CoinFlip.tsx
```diff
@@ -1,61 +1,53 @@
...
+          color={correct ? "primary" : "error"}
+        >
           {letter}
         </Typography>
-      </Box>,
+        ,
+      </div>,
     );
-    setStatus(correct ? " win!" : "lose!");
+    setStatus(correct ? " Win" : "Lose");
```
d0sboots: Any reason the text here is changing?

src/Casino/BetInput.tsx
```diff
@@ -0,0 +1,85 @@
...
+    const bet = Math.round(parseFloat(betInput));
+    let isValid = false;
+    if (isNaN(bet)) {
+      setBet(0);
+      setHelperText("Not a valid number");
+    } else if (bet <= 0) {
+      setBet(0);
+      setHelperText("Must bet a positive amount");
+    } else if (bet > maxBet) {
+      setBet(0);
```
d0sboots: I kind of liked the behavior of ceiling the bet to the maximum, it made it easier to enter by just putting in a bunch of 9s.

d0sboots:
Nice, the greater consistency looks good. This will of course mess up existing Casino scripts (again) XD

d0sboots (commented):
Oh no I forgot to send these pending comments from forever ago

src/Casino/BetInput.tsx
```diff
@@ -0,0 +1,85 @@
...
+      setHelperText("Exceed max bet");
+    } else if (!Player.canAfford(bet)) {
```
d0sboots: I'm not sure that we should be checking canAfford at this point... The player's money can (and does) change between this check and when the bet is actually used.

IMO setting the bet value should be to any valid number, and then either the play button would be disabled if you can't afford, or it would fail on clicking.

src/Casino/CoinFlip.tsx
catloversg: It has been so long, so I don't remember the reason now. It may not be important, so I reverted the change in the latest commit.

catloversg:
I updated based on your feedback. I also made 2 changes to make the behaviors of 4 games more consistent.
Currently, their behaviors are like this:
- All games enable the play button without checking the bet, except Blackjack.
- Roulette and slot machine call `reachedLimit()` *after* playing. Coin flip and blackjack do not.

My changes:
- All games enable the play button without checking the bet. We will check the bet *after* the player presses the play button.
- No game calls `reachedLimit()` *after* playing. We only need to call it *after* the player presses the play button.

PR #1695, opened 2024-10-17
BLADEBURNER: Test cover action completion

Alpheus:
# BLADEBURNER: Test cover action completion

## Things covered
- main effect of general actions
- stat exp, int exp, money and rank rewards for contracts, operations, black ops
- simulacrum effect on work
- scaling and proportion of reward vs risk (ie. black ops give more rank than ops, stages of scaling of charisma with diplomacy and incite violence depending on chaos levels)

## Not covered in this PR
- the process loop. Has a dependency to the React router. It's a minor refactor but I was avoiding all source code changes in this stage
- random city events
- stat rewards on failure. Without **explicitly** measuring exact formulas and random rolls, the detection between low and high reward is very flaky
- precise measurements of stat gain and modifiers, can test that separately through skills
- player and sleeve damage—that's mostly already covered by casualties tests
- sleeve complete action—sadly a bit more involved to control and simulate due to the process loops overlapping. Requires a minor refactor. Next round.
- skill rankups and various scaling

## Outcomes

Bladeburner line coverage is now up to 60%. Branch cover is slightly lower than that due to the branches produced by overlaps with data and react components.

This PR makes it easier to perform changes in the following areas:
- Stat XP distribution for actions
- Chaos changes and stat scaling for things like Diplomacy and Incite Violence
- Larger refactors in the action scheduling area
- Adding new missions without requiring involved balance simulations

# Linked issues

- related to PR-1586 https://github.com/bitburner-official/bitburner-src/pull/1586
- related to PR-1664 https://github.com/bitburner-official/bitburner-src/pull/1664

# Documentation

No NS api changes or source changes in any production-facing code.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -7,6 +7,7 @@ node_modules/
 dist
 input
 assets
+coverage
 doc
 markdown
 electron
--- a/test/jest/Bladeburner/Actions.test.ts
+++ b/test/jest/Bladeburner/Actions.test.ts
@@ -1,20 +1,72 @@
 import type { Bladeburner } from "../../../src/Bladeburner/Bladeburner";
 import { PlayerObject } from "../../../src/PersonObjects/Player/PlayerObject";
 import { Player, setPlayer } from "@player";
-import { Contract } from "../../../src/Bladeburner/Actions";
-import { BladeburnerContractName } from "@enums";
+import { Contract, GeneralAction, Operation } from "../../../src/Bladeburner/Actions";
+import { BladeburnerContractName, BladeburnerGeneralActionName, BladeburnerOperationName, CrimeType } from "@enums";
 import { FormatsNeedToChange } from "../../../src/ui/formatNumber";
+import { CrimeWork } from "../../../src/Work/CrimeWork";
+import type { ActionIdentifier } from "../../../src/Bladeburner/Types";
+import type { Skills } from "../../../src/PersonObjects/Skills";
+
+type StatKey = keyof Skills;
 
 describe("Bladeburner Actions", () => {
-  const REASONABLE_TIME_TO_COMPLETE_ANY_ACTION = 1e50;
+  const Tracking = Contract.createId(BladeburnerContractName.Tracking);
+  const Diplomacy = GeneralAction.createId(BladeburnerGeneralActionName.Diplomacy);
+  const Assassination = Operation.createId(BladeburnerOperationName.Assassination);
+  const Recruitment = GeneralAction.createId(BladeburnerGeneralActionName.Recruitment);
+  const ENOUGH_TIME_TO_FINISH_ACTION = 1e5;
 
   let inst: Bladeburner;
 
-  function initBladeburner(player: PlayerObject): player is PlayerObject & { bladeburner: Bladeburner } {
-    player.init();
-    player.startBladeburner();
-    return true;
-  }
+  /** All the tests depend on this assumption */
+  it("always succeeds with optimal stats, rank, stamina and city chaos levels", () => {
+    guaranteeSuccess(), start(Assassination), finish();
+    expect(inst.getActionObject(Assassination).getSuccessChance(inst, Player, { est: false })).toBe(1);
+  });
+
+  describe("Without Simulacrum", () => {
+    it("Starting an action cancels player's work immediately", () => {
+      Player.startWork(new CrimeWork({ crimeType: CrimeType.assassination, singularity: false }));
+      start(Diplomacy);
+      expect(Player.currentWork).toBeNull();
+    });
+  });
+
+  describe("Upon successful completion", () => {
+    it("Contracts give the player money", () => {
+      const moneyBefore = Player.money;
+      guaranteeSuccess(), start(Tracking), finish();
+      expect(Player.money).toBeGreaterThan(moneyBefore);
+    });
+
+    it.each(<StatKey[]>["strength", "dexterity", "defense", "dexterity"])(
+      "Operations provide combat skill XP: %s",
+      (stat: StatKey) => {
+        const statsBefore = Object.assign({}, Player.skills);
+        guaranteeSuccess(), start(Assassination), finish();
+        expect(Player.skills[stat]).toBeGreaterThan(statsBefore[stat]);
+      },
+    );
+
+    describe("Recruitment", () => {
+      it("provides charisma", () => {
+        const { charisma } = Player.skills;
+        guaranteeSuccess(), start(Recruitment), finish();
+        expect(Player.skills.charisma).toBeGreaterThan(charisma);
+      });
+
+      it("hires team member", () => {
+        guaranteeSuccess(), start(Recruitment), finish();
+        expect(inst.teamSize).toBeGreaterThan(0);
+      });
+    });
+  });
+
+  it("have a minimum duration of 1 second", () => {
+    start(Tracking);
+    expect(inst.actionTimeToComplete).toBeGreaterThanOrEqual(1);
+  });
 
   beforeAll(() => {
     /* Initialise Formatters. Dependency of Bladeburner */
@@ -29,10 +81,32 @@ describe("Bladeburner Actions", () => {
     }
   });
 
-  it("an action is never instant", () => {
-    inst.startAction(Contract.createId(BladeburnerContractName.Tracking));
-    inst.getActionObject(Contract.createId(BladeburnerContractName.Tracking)).count = 5;
-    inst.clearConsole();
-    inst.processAction(0);
-  });
+  function initBladeburner(player: PlayerObject): player is PlayerObject & { bladeburner: Bladeburner } {
+    player.init();
+    player.startBladeburner();
+    return true;
+  }
+
+  function guaranteeSuccess() {
+    Player.gainStrengthExp(1e200);
+    Player.gainAgilityExp(1e200);
+    Player.gainDexterityExp(1e200);
+    Player.gainDefenseExp(1e200);
+    inst.calculateMaxStamina();
+    inst.stamina = inst.maxStamina;
+    inst.rank = 1e10;
+    inst.cities[inst.city].chaos = 0;
+    inst.cities[inst.city].comms = 100;
+    inst.cities[inst.city].pop = 1e9;
+  }
+
+  function start(id: ActionIdentifier) {
+    const action = inst.getActionObject(id);
+    if ("count" in action) action.count = 1;
+    inst.startAction(id);
+  }
+
+  function finish() {
+    inst.processAction(ENOUGH_TIME_TO_FINISH_ACTION);
+  }
 });
--- a/test/jest/Bladeburner/Actions.test.ts
+++ b/test/jest/Bladeburner/Actions.test.ts
@@ -1,0 +1,38 @@
+import type { Bladeburner } from "../../../src/Bladeburner/Bladeburner";
+import { PlayerObject } from "../../../src/PersonObjects/Player/PlayerObject";
+import { Player, setPlayer } from "@player";
+import { Contract } from "../../../src/Bladeburner/Actions";
+import { BladeburnerContractName } from "@enums";
+import { FormatsNeedToChange } from "../../../src/ui/formatNumber";
+
+describe("Bladeburner Actions", () => {
+  const REASONABLE_TIME_TO_COMPLETE_ANY_ACTION = 1e50;
+
+  let inst: Bladeburner;
+
+  function initBladeburner(player: PlayerObject): player is PlayerObject & { bladeburner: Bladeburner } {
+    player.init();
+    player.startBladeburner();
+    return true;
+  }
+
+  beforeAll(() => {
+    /* Initialise Formatters. Dependency of Bladeburner */
+    FormatsNeedToChange.emit();
+  });
+
+  beforeEach(() => {
+    setPlayer(new PlayerObject());
+    if (initBladeburner(Player)) {
+      inst = Player.bladeburner;
+      inst.clearConsole();
+    }
+  });
+
+  it("an action is never instant", () => {
+    inst.startAction(Contract.createId(BladeburnerContractName.Tracking));
+    inst.getActionObject(Contract.createId(BladeburnerContractName.Tracking)).count = 5;
+    inst.clearConsole();
+    inst.processAction(0);
+  });
+});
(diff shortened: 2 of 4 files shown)


.eslintignore
```diff
@@ -7,6 +7,7 @@ node_modules/
 dist
 input
 assets
+coverage
```
Alpheus, line 10: Added `coverage` folder to ignore as I was using the coverage report to verify the PR's claims while also checking for any loose ends.

test/jest/Bladeburner/Actions.test.ts
```diff
@@ -0,0 +1,358 @@
...
+
+  /** Why 1e14? The stat tests need to be able to detect an EXP change of cca. 0.01
+   * 1e14 is the last threshold where 1e14 + 0.01 > 1e14
+   *
+   * Bug? Coincidentally this also means that stat gains stop at that level unless
+   * exp multipliers play a significant role
+   *
+   * Example: 1e14 strength exp gives ~~831 strength skill without augs/modifiers
+   */
+  const HIGH_STAT_EXP = 1e14;
```
Alpheus: Where necessary to guarantee 100% success rate giving the player high combat stats through EXP.

test/jest/Bladeburner/Actions.test.ts
```diff
@@ -0,0 +1,358 @@
...
+  const LOW_STAT_EXP = 1e6;
+
+  let inst: Bladeburner;
+
+  const instanceUsedForTestGeneration = new Bladeburner();
+  const CITIES = <CityName[]>Object.keys(instanceUsedForTestGeneration.cities);
+
+  /** All the tests depend on this assumption */
+  it("always succeeds with optimal stats, rank, stamina and city chaos levels", () => {
```
Alpheus: This is not a test directly, but rather a shared assumption that breaks all reliability should this one not pass.

Expect this to change only if we significantly alter balance in favor of making BB harder on combat stats, which I deemed unlikely.

test/jest/Bladeburner/Actions.test.ts
```diff
@@ -0,0 +1,358 @@
...
+
+  describe("Without Simulacrum", () => {
+    it("Starting an action cancels player's work immediately", () => {
+      Player.startWork(new CrimeWork({ crimeType: CrimeType.assassination, singularity: false }));
+      start(SampleGeneralAction);
+      expect(Player.currentWork).toBeNull();
+    });
+  });
+
+  describe("Upon successful completion", () => {
```
Alpheus, line 44: There's plenty of orthogonal concepts:
- the type of action: general, op, black op, contract
- failure vs success
- custom behavior
- black ops need setup
- contracts award money

To make test organization easier, I am using the Operation name where applicable to keep the tests in source code separate, but in the test runner they will merge according to the common describe block headers.

Ie. 2 separate tests for 
- describe Recruitment behavior
- describe Recruitment stat gain

Will merge as such:
image

test/jest/Bladeburner/Actions.test.ts
```diff
@@ -0,0 +1,358 @@
...
+    describe.each([...actionId(contracts())])("$id.name", ({ id }) => {
+      it("all contracts award money", () => {
+        guaranteeSuccess(), (before = Player.money), complete(id);
+        expect(Player.money).toBeGreaterThan(before);
+      });
+    });
+
+    /** Stat EXP check for all actions */
+    /** Checking all of them to avoid regressions */
+    describe.each([...actionIdWithIndividualStat(nonGeneralActions())])("$id.name", ({ id, stat }) => {
```
Alpheus: This is the most complex bit. It generates pairs of Action + Stat depending on which stat weights are non-zero for each given action. This allows for a quick rundown of all possible combos to make sure we're not missing awards for things like hacking exp and int exp.

These are not hardcoded to the Actions per se, but using the data files through the static Bladeburner instance. It is somewhat coupled to implementation detail but I figured it was the lesser of two evils rather than writing a test suite of all the combinations manually.

test/jest/Bladeburner/Actions.test.ts
```diff
@@ -0,0 +1,358 @@
...
+    Player.gainStrengthExp(LOW_STAT_EXP);
+    Player.gainDefenseExp(LOW_STAT_EXP);
+    Player.gainAgilityExp(LOW_STAT_EXP);
+    Player.gainDexterityExp(LOW_STAT_EXP);
+
+    inst.calculateMaxStamina();
+    inst.stamina = inst.maxStamina;
+
+    resetCity();
+  }
```
Alpheus: Player setup where we want to keep numbers small in order to more precisely measure floating point operations (usually exp rewards).

test/jest/Bladeburner/Actions.test.ts
```diff
@@ -0,0 +1,358 @@
...
+    inst.setSkillLevel(BladeburnerSkillName.BladesIntuition, 1e12);
+    inst.setSkillLevel(BladeburnerSkillName.Cloak, 1e12);
+    inst.setSkillLevel(BladeburnerSkillName.DigitalObserver, 1e12);
+    inst.setSkillLevel(BladeburnerSkillName.Reaper, 1e12);
+
+    inst.calculateMaxStamina();
+    inst.stamina = inst.maxStamina;
+
+    resetCity();
+  }
```
Alpheus: Player setup where we want actions to run as fast as possible (in simulated time) with 100% success chance

test/jest/Bladeburner/Actions.test.ts
```diff
@@ -0,0 +1,358 @@
...
+
+  function guaranteeFailure() {
+    inst.rank = 1;
+    inst.changeRank(Player, 400e3); // Minimum to attempt to fail hardest op
+    Player.gainAgilityExp(1e50);
+    inst.calculateMaxStamina();
+    inst.stamina = 0;
+
+    resetCity();
+  }
```
Alpheus: Guarantee failure on non-general actions by dropping stamina to zero.

test/jest/Bladeburner/Actions.test.ts
```diff
@@ -0,0 +1,358 @@
...
+
+  function resetCity() {
+    inst.cities[inst.city].chaos = 0;
+    inst.cities[inst.city].comms = 100;
+    inst.cities[inst.city].pop = 1e9;
+
+    /** Disable random event */
+    inst.randomEventCounter = Infinity;
```
Alpheus: Random events trigger when the counter is `<= zero`
Don't want random events to interfere with checks for chaos, communities, pops or estimation figures

test/jest/Bladeburner/Actions.test.ts
```diff
@@ -0,0 +1,358 @@
...
+  }
+
+  function* actionIdWithIndividualStat(actions: Iterable<Action>) {
+    for (const action of actions) {
+      yield* Object.entries(action.weights)
+        .filter(([__, value]) => value > 0)
+        .map(([stat]) => ({ id: action.id, stat } as { id: ActionIdentifier; stat: keyof Skills }));
+    }
+  }
+});
```
Alpheus: These are generators. I am using them solely for the purpose of permutating test cases as they simplify nicely to [...contracts()] and can be composed.

Alpheus:
New testing PR up @d0sboots. Did a quick self-review. No code changes or early refactor this time. This relates to a few outstanding BB PRs where changes where deemed risky or need further testing. Hoping to get this merged sooner rather than later.

To my knowledge, none of the tests are flaky.

test/jest/Bladeburner/Actions.test.ts
```diff
@@ -0,0 +1,358 @@
...
+  const SampleContract = Contract.createId(BladeburnerContractName.Tracking);
+  const SampleGeneralAction = GeneralAction.createId(BladeburnerGeneralActionName.Diplomacy);
+  const SampleOperation = Operation.createId(BladeburnerOperationName.Assassination);
+  const SampleBlackOp = BlackOperations["Operation Centurion"].id;
+  const ENOUGH_TIME_TO_FINISH_ACTION = 1e5;
+
+  /** Why 1e14? The stat tests need to be able to detect an EXP change of cca. 0.01
+   * 1e14 is the last threshold where 1e14 + 0.01 > 1e14
+   *
+   * Bug? Coincidentally this also means that stat gains stop at that level unless
```
d0sboots: I would take out this part; it's not a bug but rather just how things work. The rest of the explanation is solid.

test/jest/Bladeburner/Actions.test.ts
d0sboots: Generator functions are a very powerful concept, and they certainly have their place where doing something equivalent without using them requires cumbersome code. However, I don't think that place is here.

In particular, rather than taking and returning Iterables, you can take and return Arrays instead. Then these just become normal functions, still composable, and you don't need [...spread()] syntax when you call them. Also, the `contracts()` and `operations()` functions become trivial to the point of "don't define them" at that point.

Alternatively, you could restructure this functionally. In that case, you'd just remove the `for..of` loops and have these all be functors to be applied with `map()`. Your test definitions would then look like `nonGeneralActions().map(actionIdWithIndividualStat)`

test/jest/Bladeburner/Actions.test.ts
```diff
@@ -0,0 +1,358 @@
...
+    if (id.type === "Black Operations") inst.numBlackOpsComplete = (<BlackOperation>action).n;
+    inst.startAction(id);
+  }
+
+  function finish() {
+    inst.processAction(ENOUGH_TIME_TO_FINISH_ACTION);
+    inst.calculateMaxStamina();
+  }
+
+  function* nonGeneralActions() {
```
d0sboots: As an example of what I'm talking about below, as a non-generator this is as simple as:
`return [bb.contracts, bb.operations, BlackOperations].flatMap(Object.values);`

test/jest/Bladeburner/Actions.test.ts
```diff
@@ -0,0 +1,358 @@
...
+   * exp multipliers play a significant role
+   *
+   * Example: 1e14 strength exp gives ~~831 strength skill without augs/modifiers
+   */
+  const HIGH_STAT_EXP = 1e14;
+  const LOW_STAT_EXP = 1e6;
+
+  let inst: Bladeburner;
+
+  const instanceUsedForTestGeneration = new Bladeburner();
```
d0sboots: Despite how wordy this is, I don't think it communicates as much as `bb` or `blade`.

test/jest/Bladeburner/Actions.test.ts
```diff
@@ -0,0 +1,358 @@
...
+      expect(Player.currentWork).toBeNull();
+    });
+  });
+
+  describe("Upon successful completion", () => {
+    /** Repetitive snapshot declarations in most tests below */
+    let pop, before, after;
+
+    describe(BladeburnerGeneralActionName.Training, () => {
+      const Training = GeneralAction.createId(BladeburnerGeneralActionName.Training);
```
d0sboots: Here and elsewhere: If these aren't classes, they shouldn't be uppercase. Use lowerCamelCase, or YELLING_CASE for true constants (although we do that rarely).

test/jest/Bladeburner/Actions.test.ts
```diff
@@ -0,0 +1,358 @@
...
+        expect(inst.getActionObject(SampleContract).count).toBeGreaterThan(count);
+      });
+
+      it("generates available operations", () => {
+        const { count } = inst.getActionObject(SampleOperation);
+        guaranteeSuccess(), complete(Incite);
+        expect(inst.getActionObject(SampleOperation).count).toBeGreaterThan(count);
+      });
+
+      /** Relates to all issues mentioned in PR-1586 */
```
d0sboots: Write out the issues here, it shouldn't require cross-referencing to understand what's being talked about.

d0sboots (commented):
Good tests. Some coding/style stuff, but largely LGTM.

test/jest/Bladeburner/Actions.test.ts
Alpheus: Addressed with a mock instead, opting instead to test within "reasonable exp territory"

test/jest/Bladeburner/Actions.test.ts
Alpheus: Are you referring to inst or the static one?

Alpheus:
e8afd1959 — covers all review feedback I believe.

Alpheus:
> Good tests. Some coding/style stuff, but largely LGTM.

Cheers for going over it so thoroughly!

test/jest/Bladeburner/Actions.test.ts
d0sboots: Oh, I was referring to the static one. Although, do you really need two?

Edit: I see you did remove the 2nd one, nice.

test/jest/Bladeburner/Actions.test.ts
```diff
@@ -61,85 +46,81 @@ describe("Bladeburner Actions", () => {
...
+        expect(bb.getCurrentCity().chaos).toBe(0);
       });
 
       it("does NOT affect chaos in other cities", () => {
-        const otherCity = <CityName>CITIES.find((c) => c !== inst.getCurrentCity().name);
+        const otherCity = <CityName>cities.find((c) => c !== bb.getCurrentCity().name);
         /** Testing against a guaranteed 0-chaos level of charisma */
-        basicStats(), Player.gainCharismaExp(1e500), allCitiesHighChaos(), complete(Diplomacy);
-        expect(inst.cities[otherCity].chaos).toBeGreaterThan(0);
+        Player.gainCharismaExp(1e500), allCitiesHighChaos(), complete(diplomacy);
```
d0sboots: I was apparently not looking closely enough... can you convert these to separate lines instead of using commas for sequencing?
Yes, it'll make the tests a lot longer (by linecount), but we don't really want to golf our tests.

test/jest/Bladeburner/Actions.test.ts
```diff
@@ -222,15 +209,15 @@ describe("Bladeburner Actions", () => {
...
     });
   });
 
   it("have a minimum duration of 1 second", () => {
     complete(SampleContract);
-    expect(inst.actionTimeToComplete).toBeGreaterThanOrEqual(1);
+    expect(bb.actionTimeToComplete).toBeGreaterThanOrEqual(1);
   });
 
   beforeAll(() => {
```
d0sboots, line 249: Optional: IMO it flows better if your setup code is moved to before the test code, but up to you.

test/jest/Bladeburner/Actions.test.ts
d0sboots: Most of these PR comments should probably be code comments, if they're worth noting.

test/jest/Bladeburner/Actions.test.ts
Alpheus: Yes, boss

PR #1697, opened 2024-10-17
MISC: Improve tutorial, documentation and discoverability of NS API documentation

### LINKED ISSUE #1696: Official code documentation is hard to find
The [link to the official documentation](https://github.com/bitburner-official/bitburner-src/blob/stable/markdown/bitburner.ns.md) for NS is only located in the `Scripts Editor` ingame. 
This is confusing for new users, as it should also be included in the `Documentation` page of the sidebar, especially because of the deprecated Beta RTD. 

Suggestions:

- Add the link to the `Resources` section of the ingame Documentation page
- Add the link to the main `README.md`
-  Add a disclaimer that https://bitburner-beta.readthedocs.io/en/latest/index.html is deprecated, as it is the first result on google

catloversg:
### Tutorial
- Change the example of the hack script:
  - Add `/** @param {NS} ns */`.
  - Change the tab character to double space characters.
- Mention "NS API documentation".

Capture1

### Documentation
Update README.md.

### Discoverability of NS API documentation:
Script Editor tab:
Before:
before
After:
after

Documentation tab:
Before:
before
After:
after

Note: There is a "side-effect" of this change: all external links will be changed in the same way as the NS API documentation link. For example:
Capture

Closes #1696.

### DIFF
--- a/src/ui/InteractiveTutorial/InteractiveTutorialRoot.tsx
+++ b/src/ui/InteractiveTutorial/InteractiveTutorialRoot.tsx
@@ -25,6 +25,7 @@ import {
 } from "../../InteractiveTutorial";
 import { useRerender } from "../React/hooks";
 import { Settings } from "../../Settings/Settings";
+import { NsApiDocumentationLink } from "../React/NsApiDocumentationLink";
 
 interface IContent {
   content: React.ReactElement;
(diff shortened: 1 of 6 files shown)


Alpheus (approved):

src/ui/InteractiveTutorial/InteractiveTutorialRoot.tsx
```diff
@@ -352,6 +353,13 @@ export function InteractiveTutorialRoot(): React.ReactElement {
             continuously hack the n00dles server.
             <br />
             <br />
+            To access{" "}
+            <Typography component="span" color={Settings.theme.warning}>
```
d0sboots: I think either this shouldn't be colored differently, or you should go all the way and make this a `<Link>` too.

d0sboots (commented):
I think the "side-effect" of external links being styled a little differently is a good one. It shouldn't be a style difference for just the NS API, it is better for it to be a systematic reason.

PR #1699, opened 2024-10-18
BLADEBURNER: UI now shows tooltips on action success chance to show stat scaling

Alpheus:
# BLADEBURNER: UI now shows tooltips on action success chance to show stat scaling

Hovering over the "estimated success chance" text now shows what stats it scales from. Plus description for Black Ops to keep them focused on the lore text.

# Motivation

Joining bladeburner requires 100 combat stats but it's never explained how or why it is relevant to gameplay. Hoping this tooltip will help connect the dots.

Looking for review/feedback on:
- wording, spacing
- tooltip alternatives
- tooltip shape
- (optional) whether to generate the tooltips from weights

### DIFF
--- a/src/Bladeburner/Actions/Action.ts
+++ b/src/Bladeburner/Actions/Action.ts
@@ -12,6 +12,7 @@ import { clampNumber } from "../../utils/helpers/clampNumber";
 
 export interface ActionParams {
   desc: string;
+  successScaling?: string;
   baseDifficulty?: number;
   rewardFac?: number;
   rankGain?: number;
@@ -25,6 +26,7 @@ export interface ActionParams {
 
 export abstract class ActionClass {
   desc = "";
+  successScaling = "";
   // For LevelableActions, the base difficulty can be increased based on action level
   baseDifficulty = 100;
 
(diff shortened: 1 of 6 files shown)


d0sboots:
Can you add a screenshot of the tooltip in action? Also, what happens when there's no text? It looks like it'll make an empty tooltip, probably better to not have a tooltip in that case.

Aside from that, looks good.

Alpheus:
> Can you add a screenshot of the tooltip in action? Also, what happens when there's no text? It looks like it'll make an empty tooltip, probably better to not have a tooltip in that case.
> 
> Aside from that, looks good.

I checked the docs for nullish titles and this is what they claim:
> 
-- | -- | -- | --
Tooltip title. **Zero-length titles string, undefined, null and false are never displayed.**
-- | -- | -- | --

from: https://mui.com/material-ui/api/tooltip/

Is the wrapped empty component an issue? I can cycle it out with a ternary but it becomes much harder to read (relative to the existing simplicity)

Screenshot: 
image

d0sboots:
> Is the wrapped empty component an issue? I can cycle it out with a ternary but it becomes much harder to read (relative to the existing simplicity)

Nope, that's fine then.

PR #1703, opened 2024-10-19
UI: Added the ability to change the font size

G4mingJon4s:
Changed the following:
- Made the input fields in the `StyleEditorModal` one generic component
- Added an option to change the font size of the entire UI
- Added an option to change the tail font size

The default font size of the entire UI is different from the font size of the tail windows, because they are applied in a different way.
MUI applies a scaling on the `fontSize` value the input sets. It has an internal font size of 16px and a `fontSize` value of `14` is scaled to the default rem value of `1rem`. The tail window font size is applied through `styles`, so it is a direct pixel value. If this is something that bothers people, I can also add a scaling to the tail window font size to make it uniform with the general font size.

I also added `display: inline-block` to the span of each tail window line. This ensures the line height is correct for very small or very big font sizes. I couldn't find any styling implications from this, so nothing else should have changed. If I missed any issues that this creates, I'd be happy to fix them.


d0sboots:
I was worried about the display: inline-block thing, but I see it only applies to string log output, and is skipped for rawPrint stuff. So that's fine.

Could you add some screenshots to the PR?

src/ui/React/LogBoxManager.tsx
```diff
@@ -398,7 +398,20 @@ function LogWindow({ hidden, script, onClose }: LogWindowProps): React.ReactElem
...
+                      line
+                    ) : (
+                      <ANSIITypography
+                        key={i}
+                        text={line}
+                        color={lineColor(line)}
+                        styles={{
+                          display: "inline-block",
+                          fontSize: Settings.styles.tailFontSize,
+                          lineHeight: `calc(${Settings.styles.tailFontSize}px * ${Settings.styles.lineHeight})`,
```
d0sboots: This didn't need to happen before; why is it needed now?

G4mingJon4s:
Added Options:
image

Default font size of 16px (14 in the options due to conversions):
image

Setting the general font size to 10 (a displayed font size of about 11px):
image

Setting the general font size to 20 (a displayed font size of about 23px):
image

Default tail font size of 16px:
image

Setting the tail font size to 12px:
image

How the tail font size of 12px would look like without `display: inline-block`:
image

Setting the tail font size to 20px:
image

And setting the line height to 2 on top of that:
image

So, after playing around a bit, I realized that it's pretty hard to get the tail and general font size to match, if it is not the default.
I will convert the general font size input to be in pixels shortly.

src/ui/React/LogBoxManager.tsx
G4mingJon4s: The inline block is needed to make the line height match what was set by the user for smaller and bigger font sizes than the default (see the images in my comment).
The line height might not be needed since the set value should cascade down. I'll play around a bit and remove it, if it isn't necessary.

G4mingJon4s:
Also, I set a minimum font size of 5, which is about 6px. During my testing I set font sizes like 2 or 3, which made the UI unusable. To stop the user from messing up their UI to the point where they can't even revert them to the default, I found that about 6px is just on the edge where that's possible. If you think this limitation should be moved or isn't needed, I can remove it.

d0sboots:
> So, after playing around a bit, I realized that it's pretty hard to get the tail and general font size to match, if it is not the default.
I will convert the general font size input to be in pixels shortly.

Agreed, I think it not being in pixels will confuse everyone, so I think that's the only thing I'm holding this back for.

G4mingJon4s:
- All font size inputs are now in pixels
- Removed setting the font size explicitly, since it is inherited from the typography component
- Fixed a small issue where separate print calls had a different line height than what was intended for the font size

src/Themes/ui/StyleEditorModal.tsx
```diff
@@ -106,8 +106,10 @@ export function StyleEditorModal(props: IProps): React.ReactElement {
         <StyleField<"fontSize">
           name="Font Size"
           type="number"
-          value={customStyle.fontSize}
-          onChange={(value, error) => update({ ...customStyle, fontSize: Math.max(5, Number(value) ?? 8) }, error)}
+          value={customStyle.fontSize * (16 / 14)}
+          onChange={(value, error) =>
+            update({ ...customStyle, fontSize: Math.max(5, (Number(value) ?? 8) * (14 / 16)) }, error)
```
d0sboots, line 116: What's up with this 14/16 adjustment? It's really weird. I'm pretty sure what you actually want to do is `+ "px"`

d0sboots (commented):
Sorry! This somehow slipped through the cracks.

src/Themes/ui/StyleEditorModal.tsx
G4mingJon4s: MUI has an internal font size of 14, which then gets translated to 16px inside the typography.
The font size that "overwrites" the tail font size is directly added by the styling. This value is in pixels.
The inputs need to match, as two differently scaling inputs are hard to work with.
To the user, both inputs are in pixels. The value MUI uses to set the font size needs to have that weird scaling of 16 to 14, so it will correctly scale back to 16px.
Maybe MUI accepts the input in "14px" etc., but it doesn't look like that to me. [Here](https://github.com/G4mingJon4s/bitburner-src/blob/cf900638e6ca02daa97d2f9661cc8aab0791befb/src/Themes/ui/Theme.tsx#L117) the type of the property should be `number | undefined`, so I went with the scaling. There are a few other times the font size gets altered in that file, but those have a different type. Looking into the type def of the typography property didn't help either, it's pretty generic.

d0sboots: That makes sense. I added this as a comment to the source, it's weird enough that we shouldn't lose track of it.

PR #1710, opened 2024-10-21
API: Improve built-in print APIs when printing Promise objects

### LINKED ISSUE #1706: API: ns.singularity.getCurrentWork().completion (when grafting) returns an empty object.
When grafting an augmentation, `ns.singularity.getCurrentWork()` returns an object such as this:
```js
{
  "type": "GRAFTING",
  "cyclesWorked": 3692,
  "augmentation": "Bionic Legs",
  "completion": {}
}
```
It looks like "completion" is just an empty object. I expected it to be e.g. a decimal between 0.0 and 1.0

It's also not terribly clear what "cyclesWorked" infers. An expert player might know that a cycle is 200ms, in which case a player can use this in combination with `ns.grafting.getAugmentationGraftTime` to infer the completion / time remaining, but that's quite a bit to infer.

All that said, an argument could be made that the ability to `await ns.grafting.waitForOngoingGrafting()`, as well as the progress that is automatically displayed on the HUD, makes it relatively low-priority to surface this information to the API.

catloversg:
When the player prints Promise objects, those objects are serialized into `{}`. This may be confusing in some cases. This PR adds a special case in `argsToString` to handle this situation.

Closes #1706.

Before:
before

After:
after

### DIFF
--- a/src/Netscript/NetscriptHelpers.tsx
+++ b/src/Netscript/NetscriptHelpers.tsx
@@ -256,7 +256,16 @@ function argsToString(args: unknown[]): string {
       return (out += `< Set: ${[...nativeArg].join("; ")} >`);
     }
     if (typeof nativeArg === "object") {
-      return (out += JSON.stringify(nativeArg));
+      return (out += JSON.stringify(nativeArg, (_, value) => {
+        /**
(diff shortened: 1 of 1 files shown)


d0sboots:
Adding a replacer also gives us a hook to add printing for Sets/Maps if we want to add that later.

PR #1715, opened 2024-10-22
DEVMENU: Reset other gangs when leaving gang

catloversg:
When we use the "Leave Gang" button, other gangs' data like power and territory are not reset. This PR fixes that.

### DIFF
--- a/src/Gang/AllGangs.ts
+++ b/src/Gang/AllGangs.ts
@@ -6,39 +6,8 @@ interface GangTerritory {
   territory: number;
 }
 
-export let AllGangs: Record<string, GangTerritory> = {
-  [FactionName.SlumSnakes]: {
-    power: 1,
-    territory: 1 / 7,
-  },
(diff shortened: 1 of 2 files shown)


Alpheus (approved):

src/Gang/AllGangs.ts
```diff
@@ -70,6 +39,12 @@ export function resetGangs(): void {
   };
 }
 
+export let AllGangs: Record<string, GangTerritory> = getDefaultAllGangs();
```
d0sboots, line 42: Can this be const? (I know it was pre-existing)

d0sboots: No, it can't, I am a dumb

Alpheus: Technically it could be with some extra work. The mutation is not used outside of this package. The only mutation happens below from the reviver.

PR #1716, opened 2024-10-22
IPVGO: Add optional board state argument to the go analysis functions

ficocelliguy:
Adds support to the go analysis functions to provide an optional boardstate to analyze, instead of only supporting analysis of the current board state. This makes it easier to build lookahead or other move search tools.

This feature was requested here:
https://discord.com/channels/415207508303544321/1221920484111814828/1297387013036576818


Alpheus:
Seems reminiscent to how we manage `formulas`. A good candidate feature for **ns.formulas.go**?

src/Go/effects/netscriptGoImplementation.ts
```diff
@@ -322,6 +332,49 @@ export function getStats() {
   return statDetails;
 }
 
+export const boardValidity = {
```
d0sboots: I don't think you're using this export anywhere? In which case, you can also change this to a tighter
```js
const boardValidity = {
...
} const;
```
Declaration (making it a true constant for TypeScript)

src/Go/effects/netscriptGoImplementation.ts
```diff
@@ -322,6 +332,49 @@ export function getStats() {
...
+    error(boardValidity.badShape);
+  }
+  if (![5, 7, 9, 13, 19].includes(boardState.length)) {
+    error(boardValidity.badSize);
+  }
+  if (boardState.find((row) => row.match(/[^XO#.]/))) {
+    error(boardValidity.badCharacters);
+  }
+  try {
+    const board = boardFromSimpleBoard(boardState);
```
d0sboots: All the places where you use this end up wanting the full board. Why not return the full board, and take that as the optional argument (on the internal, not external side only)?

src/Go/boardAnalysis/boardAnalysis.ts
```diff
@@ -635,14 +635,16 @@ export function simpleBoardFromBoardString(boardString: string): SimpleBoard {
...
-  return simpleBoard.map((column, x) =>
+  const board = simpleBoard.map((column, x) =>
     column.split("").map((char, y) => {
       if (char === "#") return null;
       if (char === "X") return blankPointState(GoColor.black, x, y);
       if (char === "O") return blankPointState(GoColor.white, x, y);
       return blankPointState(GoColor.empty, x, y);
     }),
   );
+  updateChains(board);
```
d0sboots, line 654: Why is this added? The comment explicitly says "no analytics."

src/Go/boardState/boardState.ts
```diff
@@ -59,6 +59,13 @@ export function getNewBoardState(
   return newBoardState;
 }
 
+export function getNewBoardStateFromSimpleBoard(
```
d0sboots, line 66: You only use this once, I think it makes more sense to just do this where you use it instead of exporting a new function.

src/Go/effects/netscriptGoImplementation.ts
```diff
@@ -150,8 +158,8 @@ export async function getOpponentNextMove(logOpponentMove = true, logger: (s: st
 /**
  * Returns a grid of booleans indicating if the coordinates at that location are a valid move for the player (black pieces)
  */
-export function getValidMoves() {
-  const boardState = Go.currentGame;
+export function getValidMoves(_boardState?: SimpleBoard) {
+  const boardState = _boardState ? getNewBoardStateFromSimpleBoard(_boardState) : Go.currentGame;
```
d0sboots: The fact that you have to transform this into a BoardState, unlike the others, hints at the underlying issue with giving this an optional parameter of a SimpleBoard: Determining valid moves requires knowing if certain moves are illegal due to ko, and this simply cannot be represented from a SimpleBoard representation. Given the subtlety of this particular foot-gun, I think this one would be better to omit (unless you can come up with a way to provide a full BoardState as input.)

d0sboots:
> Seems reminiscent to how we manage `formulas`. A good candidate feature for **ns.formulas.go**?

On one hand, I kind of agree. On the other hand, I hate having to pay for formulas, and the design of these fits in well extending the existing analysis functions, so I'm fine just doing it this way. :D

src/Go/boardState/boardState.ts
ficocelliguy: I played around with this, but after the other changes it is fairly large, and also the "where you use it" is now the netscript input validation implementation. Keeping this as a contained utility function here alongside similar ones ended up being cleaner.

src/Go/effects/netscriptGoImplementation.ts
ficocelliguy: The valid moves api was the particular feature requested when this came up. As a way to mitigate some of that issue, I accept the prior board state too (optionally) to support simple ko rules. I note in the docs that superko on custom boards will be an exercise for the reader.

PR #1725, opened 2024-10-23
DOCUMENTATION: Clarify cyclesWorked of Task

catloversg:
alainbryden mentioned `cyclesWorked` of Task in #1706. I think that it's good to document what it is in the NS API documentation.


src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -1646,63 +1646,81 @@ export interface TIX {
...
 }
 
+/**
+ * Base interface of all tasks.
+ *
+ * @public
+ */
+export interface BaseTask {
+  /**
+   * The number of game engine cycles has passed since this task started. 1 engine cycle = 200ms.
```
```suggestion
   * The number of game engine cycles that has passed since this task started. 1 engine cycle = 200ms.
```

PR #1726, opened 2024-10-23
BUGFIX: Running scripts may be loaded before main UI

### LINKED ISSUE #1707: API: Scripts using singularity on startup can throw Error: "Router called before initialization - toPage(..."
Just capturing what's been discussed in Discord.

Player scripts invoked on startup (either because their running state is part of the save, or they are set as the startup script) can run before the game is fully initialized, and this can lead to an error if one of them triggers routing.

Normally, this error silently goes unnoticed (because the game's error handling isn't in place yet?), so the end-result is that the script that should have run on startup crashes and appears to have not run.

If the user script has any kind of error-handling though, they would catch an error like this:
image

Here's a sample script that can be used to reproduce the error fairly reliably just by running it, saving, then refreshing the game:
```js
    try {
        ns.singularity.workForFaction("Tian Di Hui", "hacking");
        while (true)
            await ns.sleep(10000);
    }
    catch (err) {
        ns.tprint("ERROR: An unexpected error occurred and now this script must terminate:\n" + (`${err.message}\n${err.stack}` ?? err))
    }
}
```
You might have to update the faction to one you've joined.

---

As discussed, some possible solutions are:
- Don't start any user scripts until after the game is done initializing (speed runners might be mad)
- Just for the handful of singularity scripts which route, wrap the attempt to route in a try/catch. You can then either:
  - ignore the attempt to route and move on (might have weird side-effects e.g. relating to focus penalties)
  - Store the route destination in a variable, and once the UI is initialized, fulfill the route request ASAP. If multiple route-attempts occur, just save the most recent one? (I can only imagine this messing with scripts that try to interact with the UI, like auto-infiltration)

catloversg:
When we load running scripts, those scripts may run before the main UI is loaded. In that case, anything that interacts with the router will fail to run. For example, Singularity APIs that call `Router.toPage()` will throw an error, and that error may not be shown to the player because the main UI is not loaded.

How to reproduce:
- Join Daedalus.
- Run test code.
- Reload and check the console.

If you cannot reproduce it consistently, you can forcefully delay the main UI: In `src\ui\LoadingScreen.tsx`, add another promise (e.g., `new Promise((resolve) => setTimeout(resolve, 1000))`) to `[initSwc(), Engine.load(saveData)]`.

Test code:
```js
/** @param {NS} ns */
  try {
    ns.singularity.workForFaction("Daedalus", "hacking");
  } catch (e) {
    console.error(e);
  }
  while (true) {
    await ns.sleep(1000);
  }
}
```
Closes #1707.


d0sboots:
This fixes *a* bug, so if push comes to shove I'll accept it. However, this is only fixing it for the specific case of script loading, it's not fixing the root of the issue. For instance, this won't fix the game (sometimes) crashing at load, or our tests flaking.

Edit: What I'm hoping for is a way for the router to *actually* be initialized before everything, in some guaranteed deterministic fashion.

catloversg:
In the latest commit, I add a queue that stores router actions when the main UI is not loaded. That queue will be processed after the main UI is loaded. This change should fix those crashes that happen because the router is used before the initialization.

How to test it:
- In `src\ui\LoadingScreen.tsx`, delay loading the main UI.
- In `engine.tsx`, before `Engine.start()`, call `Router.toPage()` and `Router.back()`.

Instead of a crash, the game shows this warning popup:
Capture

d0sboots:
FWIW, I haven't (completely) forgotten about this. I want to do a deep dive on the code in question, and that hasn't worked out so far.

d0sboots:
I have (finally) looked at this deeply enough to really grok why there isn't going to be a nice, simple solution. Resolve conflicts one more time, and I'll merge it.

PR #1727, opened 2024-10-24
CI: Linter strict rules, re-enabled with baseline

Alpheus:
# New Package
- eslint-baseline

# Usage
- `npm run lint` now runs `eslint-baseline` instead of `eslint` (parameters remain)
- `npm run lint:full` bypasses the baseline (equivalent to old behavior)
- `npm run lint:reset-baseline` creates a new baseline file

# Motivation
This gives contributors a chance to get used to and evaluate to strict rules without having to fix everything in the older parts of the codebase.

## Caveats
- **extract**, **move** and **rename** refactors will bypass the baseline match, thus forcing the developer to address old issues in changed files. I think this is healthy, but some individuals will understandably be annoyed by this
- be mindful of your IDE's eslint automation, it will bypass the baseline unless you configure it to runit
- the output of the baseline'd lint commands will change to json. This is a requirement to the plugin itself

# Linked issues

continues on #501
relates to #474


Alpheus:
~~Behavior seems inconsistent when running in github actions workflow. Investigating.~~

My `dev` branch was out of date when generating baseline.

Alpheus:
Should be good now.

Note to reviewer: if this branch gets very old, the baseline needs to be reset with the latest pull from dev prior to merging

Alpheus:
as @catloversg mentioned the new rules _will_ show up on your IDE immediately despite the baseline. If that's annoying I can remove the linter rules but keep the baseline, see where we go from there after merging?

d0sboots:
There is currently an ongoing push to resolve many of the eslint issues. I think this is a good pragmatic solution, but still ultimately less preferred than not needing it. Let's see how far the current effort goes, and then submit this if/when it stalls.

Alpheus:
> There is currently an ongoing push to resolve many of the eslint issues. I think this is a good pragmatic solution, but still ultimately less preferred than not needing it. Let's see how far the current effort goes, and then submit this if/when it stalls.

Would those efforts be easier if we merged this without the enabled configs?

PR #1729, opened 2024-10-24
UI: Disable font ligatures by default

catloversg:
Using font ligatures may confuse the player:
- Ligatures may not be what they want. For example: `| >` will look like a triangle.
- The progress bar looks weird when it's near 100%.

This PR disables font ligatures by default. If the player wants to use that "feature", they can enable it in the Options page.

Example code:
```js
/** @param {NS} ns */
  ns.tprint("|>");
  ns.tprint("|||||-");
}
```

Before:
before

After:
after

### DIFF
--- a/src/GameOptions/ui/InterfacePage.tsx
+++ b/src/GameOptions/ui/InterfacePage.tsx
@@ -18,7 +18,7 @@ export const InterfacePage = (): React.ReactElement => {
         checked={Settings.DisableASCIIArt}
         onChange={(newValue) => (Settings.DisableASCIIArt = newValue)}
         text="Disable ascii art"
-        tooltip={<>If this is set all ASCII art will be disabled.</>}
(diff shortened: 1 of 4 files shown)


d0sboots:
I am really torn on this one. I did a deep dive on why this was even happening in the first place, so I now know it's due to a "feature" in the new font that we're using that is designed for coding, but can cause problems elsewhere in exactly this way. Given that we *already* have an option for enabling/disabling ligatures for coding, is there any purpose to leaving them on in the rest of the game? I.e. is there even a reason to add a new option for them?

catloversg:
In this PR's draft, I only set the necessary CSS to disable the font ligatures without the new setting. Before submitting that, I wondered if the player would want to use ligatures in the main UI (for whatever reasons), so I decided to add a setting for that. Personally, I don't think it's necessary at all. It's kind of "if you don't like this change, you can disable/enable it". It seems that you also think that the new setting is unnecessary, so I'll remove it.

Issue #1730, opened 2024-10-25
CODEBASE: Enable new lint rules and fix all lint errors


catloversg:
We don't enable `plugin:@typescript-eslint/recommended-requiring-type-checking` and `plugin:@typescript-eslint/strict` because there will be too many lint errors (973 errors at this point) if we do that. Snarling handled tons of them in #501. I'll continue fixing all these lint errors until we can enable those rules.

Because there are too many lint errors, I'll create a series of PR instead of a massive one. I'll close this issue after all those PRs are merged.

Reference: #474 and #501.

PR #1732, opened 2024-10-26
CODEBASE: Fix lint errors 1

catloversg:
This is a part of the series of PRs for #1730. It mostly deals with simple lint errors.


src/Corporation/Actions.ts
```diff
@@ -207,8 +207,8 @@ export function sellMaterial(material: Material, amount: string, price: string):
   try {
     if (temp.includes("MP")) throw "Only one reference to MP is allowed in sell price.";
     temp = eval?.(temp);
-  } catch (e) {
-    throw new Error("Invalid value or expression for sell price field: " + e);
+  } catch (error) {
+    throw new Error(`Invalid value or expression for sell price field: ${String(error)}`);
```
d0sboots: Here and everywhere you are making this type of change:
I think the better change would be
```suggestion
    throw new Error("Invalid value or expression for sell price field", {cause: error});
```
Here is what the difference looks like to me:
image

This will be a lot of changes (although mechanical), so if you don't think it's right or you don't want to do it, that's fine with me.

src/Corporation/Actions.ts
```diff
@@ -591,7 +591,7 @@ Your input: ${amount}
 Sanitized input: ${sanitizedAmt}
 Input after replacement: ${replaced}
 Evaluated value: ${evaluated}
-Error encountered: ${error}`);
+Error encountered: ${String(error)}`);
```
d0sboots: This is an example of a place where I wouldn't try to change anything. (I.e. it is correct as you have it.)

src/Corporation/ui/ExpandNewCity.tsx
```diff
@@ -29,8 +29,8 @@ export function ExpandNewCity(props: IProps): React.ReactElement {
   function expand(): void {
     try {
       purchaseOffice(corp, division, city);
-    } catch (err) {
-      dialogBoxCreate(err + "");
+    } catch (error) {
+      dialogBoxCreate(String(error));
```
d0sboots, line 33: This is on the fence. Stringifying the error loses info. Is that intentional here (and the several other places where this is happening?) It feels intentional, but I'm unsure if it's correct.

src/Gang/Gang.ts
```diff
@@ -112,7 +112,7 @@ export class Gang {
       this.processTerritoryAndPowerGains(cycles);
       this.storedCycles -= cycles;
     } catch (e: unknown) {
-      console.error(`Exception caught when processing Gang: ${e}`);
+      console.error(`Exception caught when processing Gang: ${String(e)}`);
```
d0sboots: Definitely in the case of `console.error`, best to pass the unstringified `e` as a 2nd argument.

src/Go/boardAnalysis/example.js
```diff
@@ -1,3 +1,9 @@
+/* eslint-disable */
```
d0sboots: I'm fine with this, but we might want to delete this or move this file elsewhere.

src/Netscript/NetscriptHelpers.tsx
```diff
@@ -223,7 +223,7 @@ function spawnOptions(ctx: NetscriptContext, threadOrOption: unknown): CompleteS
   if (spawnDelay !== undefined) {
     result.spawnDelay = number(ctx, "spawnDelay", spawnDelay);
     if (result.spawnDelay < 0) {
-      throw errorMessage(ctx, `spawnDelay must be non-negative, got ${spawnDelay}`);
+      throw errorMessage(ctx, `spawnDelay must be non-negative, got ${String(spawnDelay)}`);
```
d0sboots: After looking at a bunch of these... I'm not actually certain this is a good lint rule for our codebase.

Specifically, what this is triggering on a lot is `error  Invalid type "unknown" of template literal expression`. But this is precisely the time when we frequently *want* to use template literal expressions, particularly in error messages. I don't think wrapping these in `String()` adds anything, rather I think it obfuscates to little purpose.

If we can add an exception for "unknown" that would be good, but it might be too core of a type to do that. (And while there are flags for "allowAny" there isn't one for "allowUnknown". In that case, we might be better off without this rule entirely.

d0sboots (commented):
I'm glad you're doing these in chunks. There's a large part of this that I'm fine with, but then an equally large part that I'm not so sure about.

src/Netscript/NetscriptHelpers.tsx
Alpheus: Isn't this incorrect typing? It has nothing to do with the stringification. This is the entire context:

```ts
const { spawnDelay } = threadOrOption as Unknownify<CompleteSpawnOptions>;
  if (spawnDelay !== undefined) {
    result.spawnDelay = number(ctx, "spawnDelay", spawnDelay);
    if (result.spawnDelay < 0) {
      throw errorMessage(ctx, `spawnDelay must be non-negative, got ${spawnDelay}`);
    }
  }
```

Unknownify is being used plenty in Netscript Helpers to assert a certain sub-shape from the user-space where it is 'unknown' by default, then it is narrowed to a "type", and usually that type has optional params. So when it is implied by structure that a parameter is likely present, you can re-type it, guard it or do this unknownify trick.

In this case the type guard for undefined narrows it incorrectly to the **type** {} | null (from {} | undefined | null via unknown) and then explicitly validated and cast with another guard `number` which returns a new variable (but doesn't guard the original).

So the correct usage would be

```suggestion
      throw errorMessage(ctx, `spawnDelay must be non-negative, got ${result.spawnDelay}`);
```

src/Corporation/Actions.ts
Alpheus: Isn't this silly? We're throwing a string, not an error above. Following the TSC4.4 unknown in catch clause, this particular scenario should be type-checked.

If it's an error, print error.message, if it's a string, concatenate it to the Message in the catch block.

src/Corporation/Actions.ts
Alpheus: Here and elsewhere—this is another unknown error type that is being rethrown. It can be both Error | string and is being used for concatenation onto a new Error.

This is a variant where the error will be user-facing, relating to something on the UI. That is the most important detail—as such errors should _not_ have stack traces or technical jargon.

I agree with d0s in that the template itself should remain unchanged.
However, the catch block on line 584 should runtime-check the error and pass on the appropriate message as the error variable is being used as a string.

(it also gets rid of the annoying Error: prefix)

src/Corporation/Division.ts
```diff
@@ -645,7 +645,9 @@ export class Division {
                   amt = eval?.(amtStr);
                 } catch (e) {
                   dialogBoxCreate(
-                    `Calculating export for ${mat.name} in ${this.name}'s ${city} division failed with error: ${e}`,
+                    `Calculating export for ${mat.name} in ${this.name}'s ${city} division failed with error: ${String(
```
Alpheus: Similar to other user-facing rethrows: the users deserves a clean message and doesn't need a stack trace as it is UI-related.

The error should be runtime-type checked and the message used in plain form, ie. `e instanceof Error ? e.message : String(e)`

We have plenty of such examples that are worth replicating in src/utils/ErrorHelper.ts

src/Corporation/ui/ExpandNewCity.tsx
Alpheus: Can we type check this? Office only throws correctly typed Errors (no strings or objects anywhere).

src/Go/SaveLoad.ts
```diff
@@ -102,7 +102,7 @@ function loadCurrentGame(currentGame: unknown): BoardState | string {
   if (!currentGame) return "Savedata did not contain a currentGame";
   assertLoadingType<CurrentGameSaveData>(currentGame);
   const ai = getEnumHelper("GoOpponent").getMember(currentGame.ai);
-  if (!ai) return `currentGame had an invalid opponent: ${currentGame.ai}`;
+  if (!ai) return `currentGame had an invalid opponent: ${String(currentGame.ai)}`;
```
Alpheus: Can we type-check currentGame.ai against string?

Here and elsewhere with usage from Unknownify (assertLoadingType and other guards), the resulting shape is a Record<knownKeys, unknown>. You still need to type-check for the type of currentGame.ai. assertLoadingType only checks the type of the root element, all children are still unknown.

If you look at the function below this one, previous game board checks for Array and string values correctly on currentGame.board.

src/Netscript/ErrorMessages.ts
```diff
@@ -96,16 +96,16 @@ export function handleUnknownError(e: unknown, ws: WorkerScript | null = null, i
   if (typeof e !== "string") {
     console.error("Unexpected error:", e);
     const msg = `Unexpected type of error thrown. This error was likely thrown manually within a script.
-      Error has been logged to the console.\n\nType of error: ${typeof e}\nValue of error: ${e}`;
+      Error has been logged to the console.\n\nType of error: ${typeof e}\nValue of error: ${String(e)}`;
```
Alpheus: This is one of those rare cases where the error is already heavily type guarded, so all other options got exhausted.

If you wish to maintain the stringified Error object usage for the rethrow, then the correct `type: msg` usage is:
```ts
const newE = new Error(e);
newE.name = typeof e;
```

However, in _most_ cases these errors are thrown Promises or functions which do not stringify in a helpful manner, but the console.error already takes care of the debugging ergonomics on it.

src/StockMarket/BuyingAndSelling.tsx
```diff
@@ -48,9 +48,10 @@ export function buyStock(
   if (shares <= 0) {
     return false;
   }
+  // TODO: Validation code in this function and other ones (sellStock, shortStock, sellShort) are bad. I'll rewrite them later.
```
Alpheus: The PR is large enough as is. Can you revert this file and we tackle it when we have time to address what you wrote in the todo?

src/PersonObjects/Sleeve/Sleeve.ts
```diff
@@ -543,7 +543,7 @@ export class Sleeve extends Person implements SleevePerson {
 
   takeDamage(amt: number): boolean {
     if (typeof amt !== "number") {
-      console.warn(`Player.takeDamage() called without a numeric argument: ${amt}`);
+      console.warn(`Player.takeDamage() called without a numeric argument: ${String(amt)}`);
       return false;
     }
```
Alpheus: Is this necessary? This seems to be inherited from the PlayerObject prototype but seems unnecessarily defensive. The save/load mechanism already protects against non-numeric HP values.

src/PersonObjects/Player/PlayerObjectGangMethods.ts
```diff
@@ -36,9 +36,9 @@ export function getGangFaction(this: PlayerObject): Faction {
   return fac;
 }
 
-export function getGangName(this: PlayerObject): string {
+export function getGangName(this: PlayerObject): FactionName | null {
   const gang = this.gang;
-  return gang ? gang.facName : "";
+  return gang ? gang.facName : null;
```
Alpheus, line 41: Isn't this a behavior change?

src/NetscriptFunctions/Formulas.ts
```diff
@@ -380,37 +380,37 @@ export function NetscriptFormulas(): InternalAPI<IFormulas> {
         checkFormulasAccess(ctx);
         const person = helpers.person(ctx, _person);
         const crime = findCrime(helpers.string(ctx, "crimeType", _crimeType));
-        if (!crime) throw new Error(`Invalid crime type: ${_crimeType}`);
+        if (!crime) throw new Error(`Invalid crime type: ${String(_crimeType)}`);
```
Alpheus: **Optional: Here and elsewhere in formulas.**

Assign the return value from helpers.string above and use it as a type guard.

The line above uses `helpers.string` that will type guard against string-like primitives (numbers, strings mostly) and throw an error if it is not such a type.
So the only way you can come into the "not enumMember" branch is if  `_crime` _was_ a string, but is not a member, thus allowing you to keep onto the type without having to cast it again. This pattern repeats for all such cases in this file.

src/Netscript/NetscriptHelpers.tsx
```diff
@@ -296,17 +297,17 @@ function validateHGWOptions(ctx: NetscriptContext, opts: unknown): CompleteHGWOp
...
-    throw errorMessage(ctx, `BasicHGWOptions must be an object if specified, was ${opts}`);
+    throw errorMessage(ctx, `BasicHGWOptions must be an object if specified, was ${String(opts)}`);
   }
   // Safe assertion since threadOrOption type has been narrowed to a non-null object
   const options = opts as Unknownify<CompleteHGWOptions>;
   result.stock = !!options.stock;
   result.additionalMsec = number(ctx, "opts.additionalMsec", options.additionalMsec ?? 0);
   if (result.additionalMsec < 0) {
-    throw errorMessage(ctx, `additionalMsec must be non-negative, got ${options.additionalMsec}`);
+    throw errorMessage(ctx, `additionalMsec must be non-negative, got ${String(options.additionalMsec)}`);
```
Alpheus: Did you mean `result.additionalMsec`?

```suggestion
    throw errorMessage(ctx, `additionalMsec must be non-negative, got ${result.additionalMsec}`);
```

src/Netscript/NetscriptHelpers.tsx
```diff
@@ -264,12 +264,13 @@ function argsToString(args: unknown[]): string {
...
     if (typeof nativeArg === "object") {
-      return (out += JSON.stringify(nativeArg, (_, value) => {
+      return (out += JSON.stringify(nativeArg, (_, value: unknown) => {
         /**
          * If the property is a promise, we will return a string that clearly states that it's a promise object, not a
          * normal object. If we don't do that, all promises will be serialized into "{}".
          */
         if (value instanceof Promise) {
+          // eslint-disable-next-line @typescript-eslint/no-base-to-string -- "[object Promise]" is exactly the string that we want.
           return value.toString();
```
Alpheus, line 274: Can't we just `return "[object Promise]"` as a static string?

If [object Promise] is the string we want, then let's be explicit, rather than linter suppression and the comment.

Alpheus (changes requested):

src/Netscript/NetscriptHelpers.tsx
catloversg: @d0sboots What's your opinion?

src/PersonObjects/Player/PlayerObjectGangMethods.ts
catloversg: AFAIK, no. I checked all callers.

Alpheus: Thanks for checking!

catloversg:
After discussing on Discord, I made these changes:
- Disable `@typescript-eslint/restrict-template-expressions` and revert all related changes.
- `@typescript-eslint/no-base-to-string` and `@typescript-eslint/restrict-plus-operands`: Generally, these rules are not really useful for our case, but we will disable them line-by-line instead of disabling the rules.
- When we pass `error: unknown` to a function that expects a string, we can use `func(String(error))`. One notable case is the `dialogBoxCreate(String(error));` call in `src\Corporation\ui`.
- `console.error()`: pass `error` as the second parameter.

Edit: I also implemented the usage of `cause` when rethrowing the error, as d0sboots suggested. If we change our minds on that, I'll revert those changes later.

src/Netscript/NetscriptHelpers.tsx
```diff
@@ -297,17 +297,18 @@ function validateHGWOptions(ctx: NetscriptContext, opts: unknown): CompleteHGWOp
     return result;
   }
   if (typeof opts !== "object") {
-    throw errorMessage(ctx, `BasicHGWOptions must be an object if specified, was ${String(opts)}`);
+    // eslint-disable-next-line @typescript-eslint/no-base-to-string
+    throw errorMessage(ctx, `BasicHGWOptions must be an object if specified, was ${opts}`);
```
Alpheus, line 301: Can we play nice with the linter?

The reason the rule no-base-to-string complains is because it is not obvious from the narrowed type that it has non-machine toString/toLocaleString (ie. one that doesn't return [object Object]).

Since in this case it was *not* an object and you expect one, the appropriate measure would be to JSON.stringify.

src/Netscript/NetscriptHelpers.tsx
```diff
@@ -297,17 +297,18 @@ function validateHGWOptions(ctx: NetscriptContext, opts: unknown): CompleteHGWOp
...
   }
   // Safe assertion since threadOrOption type has been narrowed to a non-null object
   const options = opts as Unknownify<CompleteHGWOptions>;
   result.stock = !!options.stock;
   result.additionalMsec = number(ctx, "opts.additionalMsec", options.additionalMsec ?? 0);
   if (result.additionalMsec < 0) {
-    throw errorMessage(ctx, `additionalMsec must be non-negative, got ${String(options.additionalMsec)}`);
+    throw errorMessage(ctx, `additionalMsec must be non-negative, got ${options.additionalMsec}`);
```
Alpheus: Did you mean result.additionalMsec?

options.additionalMsec is the unknown input, result.additionalMsec is the number. If you suspect it's a non-numeric, nullish, and non-string value then it's best to JSON.stringify it rather than string interpolate

src/Script/RunningScript.ts
```diff
@@ -106,7 +106,8 @@ export class RunningScript {
 
     let logEntry = txt;
     if (Settings.TimestampsFormat && typeof txt === "string") {
-      logEntry = "[" + formatTime(Settings.TimestampsFormat) + "] " + String(logEntry);
+      // eslint-disable-next-line @typescript-eslint/restrict-plus-operands
```
Alpheus: Optional: Since you disabled the lint check for template string, by using template format here too you can skip the operand suppression.

src/StockMarket/BuyingAndSelling.tsx
```diff
@@ -51,7 +51,8 @@ export function buyStock(
   // TODO: Validation code in this function and other ones (sellStock, shortStock, sellShort) are bad. I'll rewrite them later.
   if (stock == null || isNaN(shares)) {
     if (ctx) {
-      helpers.log(ctx, () => `Invalid arguments: stock='${stock != null ? stock.name : stock}' shares='${shares}'`);
+      // eslint-disable-next-line @typescript-eslint/no-base-to-string
```
Alpheus: Careful, this lint was raising an error correctly. You are interpolating a stock Object (of class Stock), not a symbol. The Stock class does not have a toString implementation. This linter rule is meant to catch that.

I vote for keeping the unsuppressed linter behavior and simply adding toString to Stock.

src/Terminal/commands/download.ts
```diff
@@ -25,7 +25,7 @@ export function exportScripts(pattern: string, server: BaseServer, currDir = roo
     .then((content: Blob) => downloadContentAsFile(content, filename))
     .catch((error) => {
       console.error(error);
-      Terminal.error(`Cannot compress scripts with pattern ${pattern} on ${server.hostname}. Error: ${String(error)}`);
+      Terminal.error(`Cannot compress scripts with pattern ${pattern} on ${server.hostname}. Error: ${error}.`);
```
Alpheus: Small nit: This seems like a revert, but you added a period. Not a huge deal, but please check that **either all or none** of the similar cases in these files have periods in the message.

src/NetscriptFunctions/Singularity.ts
```diff
@@ -1142,7 +1142,7 @@ export function NetscriptSingularity(): InternalAPI<ISingularity> {
         ? resolveScriptFilePath(helpers.string(ctx, "cbScript", _cbScript), ctx.workerScript.name)
         : false;
       if (cbScript === null) {
-        throw helpers.errorMessage(ctx, `Could not resolve file path: ${String(_cbScript)}`);
+        throw helpers.errorMessage(ctx, `Could not resolve file path. callbackScript is null.`);
```
Alpheus, line 1145: I like how you handled this 👍

Alpheus:
> After discussing on Discord, I made these changes:
> 
> * Disable `@typescript-eslint/restrict-template-expressions` and revert all related changes.
> * `@typescript-eslint/no-base-to-string` and `@typescript-eslint/restrict-plus-operands`: Generally, these rules are not really useful for our case, but we will disable them line-by-line instead of disabling the rules.
> * When we pass `error: unknown` to a function that expects a string, we can use `func(String(error))`. One notable case is the `dialogBoxCreate(String(error));` call in `src\Corporation\ui`.
> * `console.error()`: pass `error` as the second parameter.
> 
> Edit: I also implemented the usage of `cause` when rethrowing the error, as d0sboots suggested. If we change our minds on that, I'll revert those changes later.

Regarding this one
**When we pass error: unknown to a function that expects a string, we can use func(String(error)). One notable case is the dialogBoxCreate(String(error)); call in src\Corporation\ui.**

I was also confused by why it gets triggered on the caught exceptions so I dug a bit deeper. The string interpolation for `${e}` is completely fine for normal use cases. Interpolation itself is not the problem. The rule gets raised when the type in question does not have a guarantee for having a non-trivial toString/toLocaleString implementation.

Most of our use cases are string | number | Error where we check for unknown types and the object types have already been checked for, **which are all legit for the linter.** We could safely create a helper to type guard against most of these use cases, especially in regards to console.error, log, rethrow or showing errors on the UI.

The only issue is that narrowing primitives or generic objects from 'unknown' still lacks the toString guarantee, it would have to be explicitly checked for known string-compatible types or stringify them with JSON.stringify.

The reason I am mentioning this is that everywhere that
```ts
${String(e)}
```

solves the problem of 'no-base-to-string' for
```ts
${e}
```

the linter should also complain, and it isn't. I'm not sure where that got disabled, though it may be accidental.

Source: https://typescript-eslint.io/rules/no-base-to-string/

src/Netscript/NetscriptHelpers.tsx
catloversg: No. When we detect invalid data from the player, we always print the exact value that they give us, not the converted value.

Alpheus: Shouldn't it be stringified then through something that can survive unknown values? JSON.stringify or equivalent? In this case only since it's a property of a structured object.

src/Netscript/NetscriptHelpers.tsx
catloversg: I don't particularly disagree with the usage of JSON.stringify, but I'll only do that if d0sboots agrees. `${opts}` is only problematic when:
- `opts` is not an object.
- `opts`'s type is also not a built-in type with a nice toString implementation.

Off the top of my head, I cannot think of a normal use case.

Alpheus: to be fair, opts _may be an object_. Object and object are not the same (emphasis: capitalization).

Something that is typeof !== 'object' (lowercase for emphasis) means that it is a non-null primitive. (null is also object).
- Symbol
- undefined
- primitive string (without new String())
- primitive number (without new Number())
- primitive boolean (without Boolean)
- functions

There is a loose check for null, so that removes undefined and null, only having to contend with the primitives that break the linter, which I believe is only Symbol.

To be exhaustive against the intended usage we could also check for Object.keys(opts).length in addition to typeof (strings are the only false positives on the keys check, the rest we have narrowed already).

The obvious culprit are Promises which notoriously simplify down to '{}' and can easily confuse a beginner.

src/Netscript/NetscriptHelpers.tsx
catloversg: Using JSON.stringify is useful when we try to print unknown data, and that data may be an object. In this case, `options.additionalMsec` is guaranteed to be a negative number. Theoretically, when we have that information, it may be better to just use the "type-narrowed" variable (`result.additionalMsec`). The usage of the original variable is kind of a "default action". IMO, printing the original variable is the better action by default. We can use JSON.stringify when it's useful.

Alpheus: My comment above was related to that. Since this particular check knows that it's already a negative number you can use the error with result.additionalMsec to carry over the correct type, rather than using the more "DRY pattern" that then requires a type cast. Pragmatism and readability over purity.

catloversg:
For the context of that decision:
- `dialogBoxCreate` expects a string in this context (`JSX.Element` is irrelevant), so we have to pass a string to it.
- That folder `src\Corporation\ui` always calls `dialogBoxCreate` with 2 "patterns":
  - `${err}`
  - `err + ""`
- They trigger different types of lint errors (back when we have not made any decisions about those rules). To fix those not-helpful lint errors, the easiest way is to use `String(error)`.
- Practically, in this context, there is no difference between `String(error)` and those old "patterns".

src/Netscript/NetscriptHelpers.tsx
catloversg: I'm not sure what you want to do here. As I said, `${opts}` is fine, maybe except some edge cases. If you mean there is a **practical** use case that we must replace `${opts}` with `${JSON.stringify(opts)}`, you have to point that out. The only thing that can fail is the Symbol, and Symbol is definitely not the practical use case.

Alpheus: No change needed outside of suggestion for stringification. The rest is discussing alternatives.

src/Script/RunningScript.ts
catloversg: Using template literals here will trigger "no-base-to-string".

Alpheus: Yes, that is correct. The linter is highlighting an issue.

this.logs is an array of react nodes, not strings. If you want literal text, these need to be converted to JSX. If you want string[], the logs need to be typed differently.

For context: a reactnode is this:
```ts
type ReactNode = React.ReactElement | string | number | React.ReactFragment | React.ReactPortal | boolean | null | undefined;
```

The text here is being constrained down to a string and then used in the wider ReactNode[] type.

The linter _should_ pick up such discrepancies as they can cause issues. Look at the if statement above. txt = lopEntry; typeof txt === 'string'.

If the narrowing is correct, why is the template triggering the rule? Check the narrowing again.

Alpheus:
```suggestion
    let logEntry = txt;
    if (Settings.TimestampsFormat && typeof txt === "string") {
      logEntry = `[${formatTime(Settings.TimestampsFormat)}] ${txt}`;
    }

    this.logs.push(logEntry);
```

catloversg: At first, I'm sorry if I sound rude here. That's not my intention. This is the same case as your previous comment of `${opts}`. I read your message, and I don't get what you want.

This is just my opinion, so feel free to skip it. When I read a code review, I want to know exactly what the reviewer wants. If everything is good, there is no need to add a review; we can just merge. If there is a review, it means that something must/should be changed. In that case, I appreciate this type of review:
- I disagree with your X, because I think Y is better.
- Suggestion: Do X instead of Y because Z.
- I'll veto this change due to X.
- etc.

The key thing here is the "action". I want to know the expected "action" of the reviewer. It's normal to disagree with each other, so your expected "action" may not be what I want to do, but I need to know what it is before I can think about it. d0sboots's review is usually excellent in this matter. They always tell me clearly what they want to change ("You do X, but I want Y"). Having extended discussion ("food for thought") is totally fine, but it should come after that.

Back to this review, what do you want me to change?

Edit: after submitting and reloading, I have seen your "Suggested change" comment.

Alpheus: No worries :)

src/Netscript/NetscriptHelpers.tsx
d0sboots: Agreed with catlover on two points:
* Denis, it was unclear to me what you were asking for (if anything). It seemed like you were asking for a change, but then you said not? It pays to be clear on this during a code review.
* I agree that JSON.stringify doesn't make sense here. It only works for a narrow set of types, and throws when it encounters other types. It is *not* something you want to be feeding `unknown` user data to in an error-handling scenario.

src/Netscript/NetscriptHelpers.tsx
```diff
@@ -297,17 +297,18 @@ function validateHGWOptions(ctx: NetscriptContext, opts: unknown): CompleteHGWOp
     return result;
   }
   if (typeof opts !== "object") {
-    throw errorMessage(ctx, `BasicHGWOptions must be an object if specified, was ${String(opts)}`);
+    // eslint-disable-next-line @typescript-eslint/no-base-to-string
```
d0sboots, line 300: Oh how obnoxious. `opts` *was* unknown, but by narrowing it we encounter this lint, instead. Oh well.

PR #1734, opened 2024-10-27
IPVGO: Makes the cheat API lookahead-friendly.

ballardrog:
- Adds getCheatCount for looking up the current number of cheat attempts.
- Updates getCheatSuccessChance to optionally take in an override for the number of cheat attempts.

Discussed at https://discord.com/channels/415207508303544321/1221920484111814828/1299910143114477648

Tested in-game with the following script:
```js
/** @param {NS} ns */
  ns.tprint("Cheat count table:");
  for (let cheatCount = 0; cheatCount <= 2; cheatCount++) {
    ns.tprint(`${cheatCount}: ${ns.formatNumber(ns.go.cheat.getCheatSuccessChance(cheatCount))}`);
  }
  ns.tprint("");

  ns.go.resetBoardState("Netburners", 5);
  printCheatStats(ns);
  await ns.go.cheat.playTwoMoves(1, 1, 2, 2);
  printCheatStats(ns);
  await ns.go.makeMove(3, 3);
  await ns.go.cheat.removeRouter(3, 3);
  printCheatStats(ns);
}

/** @param {NS} ns **/
function printCheatStats(ns) {
  ns.tprint(`Current cheat count: ${ns.go.cheat.getCheatCount()}`);
  ns.tprint(`Current cheat success chance: ${ns.formatNumber(ns.go.cheat.getCheatSuccessChance())}`);
}
```


src/Netscript/RamCostGenerator.ts
```diff
@@ -272,6 +272,7 @@ const go = {
   },
   cheat: {
     getCheatSuccessChance: 1,
+    getCheatCount: 0,
```
catloversg: Minor suggestion: set the RAM cost equal to `getCheatSuccessChance`. If @ficocelliguy and @d0sboots think that 0 RAM cost is better, I'm fine with it.

ficocelliguy: I like the idea of making it have a small cost. You can keep track of cheat count yourself in memory (or in a file) if you wanted to cut down on that ram cost, but also it's a nice convenience if you want to bump the ram slightly.

ballardrog: Makes sense to me. I bumped the ram cost to 1GB.

d0sboots: 1GB seems like a good cost to me, cheap in the context of the (generally expensive) Go functions, but enough to be worth tracking yourself if you are trying to optimize.

PR #1735, opened 2024-10-27
BUGFIX: Crash in theme editor modal

catloversg:
This PR fixes a crash in the theme editor modal.

How to reproduce it:
- Open that modal.
- Paste "null" into the text box.

The recovery mode will be activated after the crash.


src/Themes/ui/ThemeEditorModal.tsx
```diff
@@ -81,15 +82,22 @@ export function ThemeEditorModal(props: IProps): React.ReactElement {
...
       for (const key of Object.keys(importedTheme)) {
         Settings.theme[key] = importedTheme[key];
       }
       ThemeEvents.emit();
-    } catch (err) {
-      // ignore
+    } catch (error) {
+      console.error(`Theme data is invalid. Data: ${event.target.value}.`);
+      console.error(error);
+      dialogBoxCreate(`Invalid theme. ${String(error)}`);
```
d0sboots: Per disucussion on the other PR, possibly still just:
```suggestion
      dialogBoxCreate(`Invalid theme. ${error}`);
```

src/ScriptEditor/ui/ThemeEditorModal.tsx
```diff
@@ -74,12 +75,19 @@ export function ThemeEditorModal(props: ThemeEditorProps): React.ReactElement {
...
-      const importedTheme = JSON.parse(event.target.value);
-      if (typeof importedTheme !== "object") return;
+      const importedTheme = JSON.parse(event.target.value) as typeof Settings.EditorTheme;
+      if (importedTheme == null) {
+        throw new Error("Theme data must not be null or undefined.");
+      }
+      if (typeof importedTheme !== "object") {
+        throw new Error(`Theme data is invalid.`);
+      }
       Settings.EditorTheme = importedTheme;
```
d0sboots, line 85: Technically you should do the `as` cast here, since it's a bit gross to assert it's an EditorTheme and then be checking if it's null after.

catloversg: Noted. I plan to make a PR that validates the input theme data instead of bypassing the type check by casting the result of `JSON.parse`. Doing that may be a bit complicated (the JSON schema of the editor theme is more complicated than the game theme's JSON), so I have not done it.

d0sboots: Yeah, properly validating JSON schemas is tricky and annoying.

PR #1736, opened 2024-10-27
MISC: Update blood donation

catloversg:
hydroflame made this change in #1655, but that PR needs to update the Jest snapshot. hydroflame is not active anymore, so that PR may be stuck for a long time. This PR just makes the same change.


d0sboots:
Thanks, I was considering doing the same but forgot about it.

PR #1739, opened 2024-10-29
MISC: Remove or change popup shown when system clock moved backward

### LINKED ISSUE #1738: Weird backtracking of Bitburner clock
Weird. I was just at the Rothman University taking the Algorithms course (it was on Focus since I wasn't doing anything at the time), and then I went downstairs in real life to get a drink, and when I came back up, and opened the windows to take a look at Bitburner, I saw a dialogue saying:-

WARNING: The system clock moved backwards.... something or other.

Now, if it was just a minute at most, I'd just consider it a synchronization problem. But it went back a whole 4 days+, nearly 5 days! Which, by the way, in case you're wondering, there's nothing wrong with my PC's system clock. It's been working fine all this while, even now it's still working fine! My younger brother and I are both in Computer Science, so we handle and take care of our computers, smartphones, electronics, etc. ourselves. Sure beats taking it to the useless "repair facilities" here. And heaven forbid buying a Dell and taking it for "repairs" at a Dell service centre. I had friends who worked there for their industrial attachment, and they told me what actually goes on there. This is why my sibling and I assemble, install and maintain our own computers. I also used to fix up the old computers at the old Chinese companies I used to work at with super old computers, like those where I literally had to overlay the HDD before formatting it or it wouldn't let me use its full secondary storage capacity, which was already very little to begin with, and way less if I didn't overlay it first. It'd be like only 8MB instead of 20MB? Lol.

catloversg:
When we detect that the system clock moved backward, we will show a popup warning the player. After this change, we have received some reports from players asking what it means. In the dev branch, the threshold was changed to 5 minutes to make it less sensitive to NTP. When I look at this change again, I wonder if it's a useful feature or just a new source of confusion for players.
- Some (or many) players do not know why the system clock matters and why it may be changed without their notice.
- When we receive reports like that, there is nothing that we can do. The problem is simple; it's exactly what the popup says: the system clock moved backward. However, there is no way for us to "fix" that. It's the problem with the player's machine. In edge cases, if they modify Date's prototype or do other unexpected things, this popup can still be shown.

IMO, this warning popup does not help the player at all. It tells something that many players do not comprehend. Even if they understand what happened, there are not many things that they (or we) can do. Those reports just become unsolvable problems. It's very likely that it's just a problem with their machines, but it's hard to tell them, "That's your problem, not ours.".

Based on what I saw on Discord after I posted this suggestion:
- The popup is not helpful.
- Alternative option: change from a popup to a toast.

I'll change it to a toast if the maintainer wants.

Closes #1738.

### DIFF
--- a/src/engine.tsx
+++ b/src/engine.tsx
@@ -63,15 +63,6 @@ declare global {
   };
 }
 
-// Only show warning if the time diff is greater than this value.
-const thresholdOfTimeDiffForShowingWarningAboutSystemClock = CONSTANTS.MillisecondsPerFiveMinutes;
-
-function showWarningAboutSystemClock(timeDiff: number) {
-  AlertEvents.emit(
-    `Warning: The system clock moved backward: ${convertTimeMsToTimeElapsedString(Math.abs(timeDiff))}.`,
-  );
-}
-
 export const GameCycleEvents = new EventEmitter<[]>();
 
 /** Game engine. Handles the main game loop. */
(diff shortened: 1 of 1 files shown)


d0sboots:
Agreed. This is a case where ignorance is bliss.

PR #1743, opened 2024-10-31
DOCUMENTATION: Update NetscriptDefinitions.d.ts

ghost:
Added line 5934: using hackAnalyzeThreads on a server you cannot hack returns -1.


d0sboots:
That is still not an exhaustive list, but it is closer now. :D

PR #1748, opened 2024-11-03
MISC: Add generic type as returned type for action and checking

catloversg:
Edit: Originally, this PR moves Availability and Attempt to a more generic place. After discussing, we decide to create a new type instead of moving Availability and Attempt.

---
In the past, we sometimes used a returned string to indicate success/failure. For example:
- `purchaseAugmentation` (`src\Faction\FactionHelpers.tsx`) returns a string. The caller will check if it starts with "You purchased" to see whether it succeeds.
- `purchaseUnlock` and `purchaseUpgrade` return strings to indicate a failure.

When refactoring those functions, I use `IReturnStatus` (`src\types.ts`), then I see `Availability` and `Attempt`. I think these two types are pretty nice and can be used outside Bladeburner. This PR moves them to `src\types.ts`.


src/types.ts
```diff
@@ -24,13 +24,23 @@ export type Member<T> = T extends (infer arrayMember)[] ? arrayMember : T[keyof
...
 export interface IReturnStatus {
   res: boolean;
   msg?: string;
 }
 
+type AvailabilitySuccess<T extends object> = { available: true } & T;
+type AvailabilityFailure = { available?: undefined; error: string };
+export type Availability<T extends object = object> = AvailabilitySuccess<T> | AvailabilityFailure;
+
+type AttemptSuccess<T extends object> = { success: true; message?: string } & T;
```
d0sboots: These two are the same type, just with different labels. I think we should try to standardize on one, and "gently" deprecate the other. (I.e., leave it in Bladeburner.)

Edit: I have no preference as to which one to keep.

src/types.ts
```diff
@@ -24,13 +24,23 @@ export type Member<T> = T extends (infer arrayMember)[] ? arrayMember : T[keyof
...
+ * Status object for functions that return a boolean indicating success/failure
+ * and an optional message
+ */
 export interface IReturnStatus {
   res: boolean;
   msg?: string;
 }
 
+type AvailabilitySuccess<T extends object> = { available: true } & T;
+type AvailabilityFailure = { available?: undefined; error: string };
```
d0sboots: It's a bit gross that "available" isn't always defined. Can it be made at least optional-false (instead of optional-undefined) without screwing up existing uses? Ideally it would be always true or false, but that might be too big of a refactor to chew here.

PR #1750, opened 2024-11-03
MISC: Include React component stack in Recovery Mode report

catloversg:
This component stack may not be useful in most cases, but having it in the report is harmless.

I use `String(metadata.error)` to suppress the false-positive `no-base-to-string` lint error. `metadata.error` is an instance of Error in most cases, but we specify the type as `Record<string, unknown>` for reasons that I don't know of.

### DIFF
--- a/src/utils/ErrorHelper.ts
+++ b/src/utils/ErrorHelper.ts
@@ -85,7 +85,7 @@ export function getErrorMetadata(error: unknown, errorInfo?: React.ErrorInfo, pa
 
 export function getErrorForDisplay(error: unknown, errorInfo?: React.ErrorInfo, page?: Page): IErrorData {
   const metadata = getErrorMetadata(error, errorInfo, page);
-  const fileName = (metadata.error as any).fileName;
+  const fileName = String(metadata.error.fileName);
   const features =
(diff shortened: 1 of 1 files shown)


src/utils/ErrorHelper.ts
```diff
@@ -118,6 +118,11 @@ Please fill this information with details if relevant.
 ${metadata.error.stack}
 \`\`\`
 
+### React Component Stack
```
d0sboots, line 121: Can we make this whole section optional based on errorInfo presence? (Is that a useful thing to do? Is it present most of the time, or only sometimes?)

catloversg: AFAIK, it's presence in most cases. When the recovery mode UI is activated by `ErrorBoundary`, `errorInfo` will be set. The only exception that I know is when we activate that mode from `LoadingScreen` due to loading error.

Nonetheless, I can make your suggested change if you want, even when it's presense in most cases. If that's the case, just let me know.

d0sboots: I suppose we don't really have much more than our intuition on how many RecoveryScreen views are from `LoadingScreen` vs the rest of the game. I'm guessing it actually tilts towards `LoadingScreen`, but it is also just a guess.

PR #1744, opened 2024-11-02
BLADEBURNER: Change skill.canUpgrade and Skills UI

catloversg:
This PR implements changes discussed on Discord:
- Move the check of `actualCount` to `canUpgrade`.
- Disable "+1 button" when the skill level reaches `Number.MAX_SAFE_INTEGER`.

Test code:
```js
/** @param {NS} ns */
export async function main(ns) {
  Player.bladeburner.skillPoints = 1e300;
  Player.bladeburner.skills.Hyperdrive = 0;
  ns.bladeburner.upgradeSkill("Hyperdrive", 1); // OK
  Player.bladeburner.skills.Hyperdrive = 1e100;
  ns.bladeburner.upgradeSkill("Hyperdrive", 1); // Error
  Player.bladeburner.skills.Overclock = 0;
  ns.bladeburner.upgradeSkill("Overclock", 90); // OK
  ns.bladeburner.upgradeSkill("Overclock", 1); // Error
}
```

Screenshot:
Capture

PR #1751, opened 2024-11-04
MISC: Minor code change in SkillElem.tsx

catloversg:
This PR makes the suggested change at https://github.com/bitburner-official/bitburner-src/pull/1744#discussion_r1827266831.


d0sboots:
Sorry that I wasn't clear enough: I felt the need to comment (for some reason), but I wasn't actually requesting that you make a change after-the-fact. If I felt strongly enough about it, I wouldn't have merged it as-is.

catloversg:
Don't worry. I appreciate your feedback there. I was also not happy with the combination of ternary and nullish coalescing, but I was too lazy to find a better way to write them. In the end, I submitted it as it was.

PR #1754, opened 2024-11-06
MISC: Always add script's earnings to its parent

catloversg:
Currently, we only add the script's earnings to its parent script if the script quits "gracefully". This PR changes that behavior. Now we always do that.

MRE:
hack.js
```js
/** @param {NS} ns */
  await ns.hack("n00dles");
  throw new Error("error");
}
```

test.js
```js
/** @param {NS} ns */
  ns.tail();
  while (true) {
    ns.exec("hack.js", "home");
    await ns.sleep(1000);
  }
}
```

### DIFF
--- a/src/NetscriptWorker.ts
+++ b/src/NetscriptWorker.ts
@@ -330,21 +330,23 @@ Otherwise, this can also occur if you have attempted to launch a script from a t
 
   // Start the script's execution using the correct function for file type
   (isLegacyScript(workerScript.name) ? startNetscript1Script : startNetscript2Script)(workerScript)
-    // Once the code finishes (either resolved or rejected, doesnt matter), set its
+    // Once the code finishes (either resolved or rejected, doesn't matter), set its
     // running status to false
     .then(function () {
-      // On natural death, the earnings are transferred to the parent if it still exists.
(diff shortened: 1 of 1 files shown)


src/NetscriptWorker.ts
```diff
@@ -330,21 +330,23 @@ Otherwise, this can also occur if you have attempted to launch a script from a t
...
-        parent.scriptRef.onlineMoneyMade += runningScriptObj.onlineMoneyMade;
-      }
       killWorkerScript(workerScript);
       workerScript.log("", () => "Script finished running");
     })
     .catch(function (e) {
       handleUnknownError(e, workerScript);
-      workerScript.log("", () => (e instanceof ScriptDeath ? "Script killed." : "Script crashed due to an error."));
       killWorkerScript(workerScript);
+      workerScript.log("", () => (e instanceof ScriptDeath ? "Script killed." : "Script crashed due to an error."));
```
d0sboots, line 342: This is and interesting and subtle change. I'm assuming it flips the order of two log lines in a specific case?

catloversg: Yes. I move this line to make it consistent with the `then` block. Now both of them kill WS, then print the log message.

PR #1758, opened 2024-11-07
CODEBASE: Fix lint errors 3

catloversg:
This is a part of the series of PRs for #1730.

Providing type check for usages of `libarg` is too messy, so I gave up and disabled lint rules for those cases.

In `NetscriptDefinitions.d.ts`, we have some `declare enum` looked like this:
```js
declare enum UniversityLocationName {
  AevumSummitUniversity = LocationName.AevumSummitUniversity,
  Sector12RothmanUniversity = LocationName.Sector12RothmanUniversity,
  VolhavenZBInstituteOfTechnology = LocationName.VolhavenZBInstituteOfTechnology,
}
```
These enums will trigger the `@typescript-eslint/prefer-literal-enum-member` rule. IMO, this kind of usage is fine, so I disable that rule for `NetscriptDefinitions.d.ts`.

Typing for `IReviverValue` is not very good. I'm not satisfied with the way I did, but I have not found a better way. The type of `data` can be:
- Normal object (`Record<string, any>`): This is good for most classes that will be serialized to JSON in the save data.
- `Array<any>`: JSONSet.
- `Array<[any, any]>`: JSONMap.

~~I have not found a way to define T in `IReviverValue`, so I "lie" to TS and set it as `Record<string, any>`, and then typecasting in JSONSet and JSONMap. I'm open to other suggestions.~~
Edit: I tried to improve the type checking in 1f8ed0231d24eb0482a4247efc9cb021c3aba748.


.eslintrc.js
```diff
@@ -42,4 +42,12 @@ module.exports = {
       version: "detect",
     },
   },
+  overrides: [
+    {
+      files: ["src/ScriptEditor/NetscriptDefinitions.d.ts"],
+      rules: {
```
d0sboots, line 54: Since this is .js and not .json, let's include the rationale in a comment.

src/Gang/GangMember.ts
```diff
@@ -100,10 +100,16 @@ export class GangMember {
   }
 
   getTask(): GangMemberTask {
+    /**
+     * TODO: It's best to remove this code. It tries to provide "backward compatibility", but it's a futile attempt.
```
d0sboots: I was going to say, can we simply remove this? But I see that it will result in returning Unassigned, which is fine and safe.

I'd be OK with you removing this right now, if you want to.

src/SaveObject.ts
```diff
@@ -282,8 +288,69 @@ function convert(code: string, changes: [RegExp, string][]): string {
 // Makes necessary changes to the loaded/imported data to ensure
 // the game stills works with new versions
 function evaluateVersionCompatibility(ver: string | number): void {
-  // We have to do this because ts won't let us otherwise
-  const anyPlayer = Player as any;
+  type LegacyPlayer = {
```
d0sboots: I think I prefer a solution where the type is "any" or "unknown" and we deal with lint errors differently.

The reason is that the savegame data *can* be any type (and we do see this, when players do save-file editing), so it isn't correct to assert that it has this particular structure. Also, in general this is a lot of boilerplate.

src/SaveObject.ts
```diff
@@ -369,7 +436,7 @@ function evaluateVersionCompatibility(ver: string | number): void {
     Player.reapplyAllSourceFiles();
   }
   if (ver < 3) {
-    anyPlayer.money = parseFloat(anyPlayer.money);
+    anyPlayer.money = parseFloat(anyPlayer.money as string);
```
d0sboots: What lint error triggers here? `parseFloat` should be fine to run on any type.

src/SaveObject.ts
```diff
@@ -750,7 +819,21 @@ async function loadGame(saveData: SaveData): Promise<boolean> {
   if (!saveData) return false;
   const jsonSaveString = await decodeSaveData(saveData);
 
-  const saveObj = JSON.parse(jsonSaveString, Reviver);
+  const saveObj = JSON.parse(jsonSaveString, Reviver) as {
```
d0sboots: Same argument as before, but even moreso.

The current code is treating these as string, but we can't actually guarantee that. Any lint errors here are almost certainly valid.

src/SaveObject.ts
```diff
@@ -807,7 +890,7 @@ async function loadGame(saveData: SaveData): Promise<boolean> {
   }
   if (Object.hasOwn(saveObj, "LastExportBonus")) {
     try {
-      ExportBonus.setLastExportBonus(JSON.parse(saveObj.LastExportBonus));
+      ExportBonus.setLastExportBonus(JSON.parse(saveObj.LastExportBonus) as number);
```
d0sboots: We should either leave these as unfixed, or actually fix them properly here in the load code IMO. In this case that would look something like throwing an error if the bonus is not a number, and letting the catch block handle setting a reasonable default.

src/SaveObject.ts
```diff
@@ -822,7 +905,7 @@ async function loadGame(saveData: SaveData): Promise<boolean> {
   }
   if (Object.hasOwn(saveObj, "VersionSave")) {
     try {
-      const ver = JSON.parse(saveObj.VersionSave, Reviver);
+      const ver = JSON.parse(saveObj.VersionSave, Reviver) as string | number;
```
d0sboots: same

src/Server/AllServers.ts
```diff
@@ -188,13 +190,14 @@ export function initForeignServers(homeComputer: Server): void {
...
   for (const member of Object.keys(AllServers)) {
+    // eslint-disable-next-line @typescript-eslint/no-dynamic-delete
     delete AllServers[member];
   }
   AllServers = {};
 }
 
 export function loadAllServers(saveString: string): void {
-  AllServers = JSON.parse(saveString, Reviver);
+  AllServers = JSON.parse(saveString, Reviver) as typeof AllServers;
```
d0sboots: I think I've actually seen a crash on load due to this specific type assumption. You don't have to fix it now, but I'd rather not paper over it with a cast.

src/utils/JSONReviver.ts
```diff
@@ -8,12 +8,22 @@ type JsonableClass = (new () => { toJSON: () => IReviverValue }) & {
   validationData?: ObjectValidator<any>;
 };
 
-export interface IReviverValue<T = any> {
```
d0sboots, line 11: This really gets at the heart of it. The old definition was more correct; after `isReviverValue`, we only know that it has a `ctor` and `data` field, but know *nothing* about the type of the data. The only thing wrong with the old type was that T defaulted to "any" instead of "unknown".

d0sboots (commented):
I haven't finished the whole thing, but I'm gonna stop and send comments now because there's really two parts to this PR:
1. A bunch of minor lint fixes that I have no issues with.
2. A type refactoring for the save/load stuff. My issue here (as mentioned in some comments) is that our loading is *not* typesafe in actuality; anything can be in the savegame (and semi-frequently is, when people do savegame editing ineptly). As a result, representing all of these JSON'd structures as specific types is simply a lie, and the lie is occurring at all the points where you do casts. I'm actually more comfortable with the old state of the code, where we were doing a bad thing but at least we weren't lying about it. The new lint errors are (rightly) warning us about the bad things we're doing here, and I'd rather they continue to do that until we fix our issues. (Or, we disable them line-by-line and admit that we won't.)

catloversg:
Before I start making changes, I want to make sure that we agree on how to deal with these problems:

### IReviverValue
If we leave `<T = any>` as it is right now, I'll disable `@typescript-eslint/no-unsafe-argument` line-by-line in all `toJSON` and `fromJSON`.
If we change it to `<T = unknown>`, I'll have to use `// @ts-expect-error -- reason` line-by-line in all `toJSON` and `fromJSON`.
You have to make a decision here. In the case that you choose `<T = unknown>`, if you don't mind, please write the "reason" for me. I'll copy it.

### evaluateVersionCompatibility in SaveObject.ts
I agree with your comment, but we still have to find a viable way to deal with the lint errors here. If we don't lie to TS, there will be tons of lint errors. Disabling all of them line-by-line is tedious and "weird". Another option is to move this function to a new file, then disable rules in that file. This option is not ideal, but it's still better than dumping tons of `// eslint-disable-next-line` to this function.

d0sboots:
There is a third option, but it might be too much work (it would belong in its own PR for sure): Fixing the lint errors by fixing the code, i.e. properly checking types on load.

catloversg:
Gradually dealing with `fromJSON` and `toJSON` is possible, but what about `evaluateVersionCompatibility` in `SaveObject.ts`? Checking all possibly legacy properties of `anyPlayer` sounds like a hellish job. I'm not even sure how to start with it.

d0sboots:
Well, the idea would be to check at time-of-use. I.e., you don't assert it has a particular shape up-front, but rather let it be `unknown` and check for the presence of fields as you go. This is how the NetscriptHelper functions work, for instance. And the js code already *mostly* works this way, it's just making some assumptions about "if it has this field, it must be this object that also has these fields."

At the same time, I'm not too concerned about the legacy conversion code; it gets run rarely and is very hard to test, so changing it minimally is usually best. I'm fine with wholesale disabling lint on it one way or another. It's the contemporary loading code that I'd potentially like to tighten up.

src/SaveObject.ts
catloversg: `parseFloat` coerces the parameter to a string, so technically it's fine to pass anything and rely on JS's type coercion. However, in TypeScript, `parseFloat` explicitly requires a string.

src/Bladeburner/Actions/Operation.ts
```diff
@@ -82,6 +83,7 @@ export class Operation extends LevelableActionClass implements TeamActionWithCas
   }
 
   static fromJSON(value: IReviverValue): Operation {
+    throwErrorIfNotObject(value.data);
```
d0sboots: It feels like this is something that belongs inside Generic_fromJSON.

Reading further, I can see why you didn't: there are times there is conversion code that wants this assertion. However, some of that is actually OK JS (convincing TS is another matter): for instance `(5)?.field` is `undefined`, not an error, so conversions based on that work on all types without need for assertion (and assertion would actually be a minor behavior change).

`in` only works for objects though, so I'm that case you'll need to duplicate the assertion. Doing that once or twice is much better IMO than adding it as boilerplate everywhere.

src/SaveObject.ts
```diff
@@ -84,7 +66,78 @@ export interface ImportPlayerData {
...
+  SettingsSave?: string;
+  VersionSave?: string;
+  AllGangsSave?: string;
+  LastExportBonus?: string;
+  StaneksGiftSave: string;
+  GoSave: unknown; // "loadGo" function can process unknown data
+};
+
+/**
+ * This function asserts the unknown saveObject. It's called from "loadGame", and contains validation logic originally
```
d0sboots: Please remove this line and the next. These are PR comments (explaining the logic of why you made particular *changes*), as opposed to code comments (explaining the logic/purpose of the code *as-is*.) Especially since most of the time, the reader will have no idea (or care) who "I" is.

src/SaveObject.ts
```diff
@@ -84,7 +66,78 @@ export interface ImportPlayerData {
...
+ * from that function. Please note that this logic is flawed, so I made some changes.
+ *
+ * In "loadGame", we parse a json save string to saveObject, then load data from this object. When we do that, we have
+ * to ensure that this object contains valid data. The validation logic in "loadGame" was really lenient. This is how
```
d0sboots: The language here should be cleaned up too.

src/SaveObject.ts
```diff
@@ -84,7 +66,78 @@ export interface ImportPlayerData {
...
+ * "loadGame" did it (I'll change these behaviors in the implementation of this function if needed):
+ * - Some properties of the object are "mandatory". We assume that they are strings without performing any checks.
```
d0sboots: It would be nice to check that they're strings, but probably not in this PR. It's gotten too much scope creep already. (Totally my fault)

src/Types/Jsonable.ts
```diff
@@ -1,21 +1,33 @@
...
-  toJSON(): IReviverArrayValue {
+  toJSON(): IReviverValue {
     return { ctor: "JSONMap", data: Array.from(this) };
   }
 
-  static fromJSON(value: IReviverArrayValue): JSONMap<any, any> {
-    return new JSONMap(value.data);
+  static fromJSON(value: IReviverValue): JSONMap<any, any> {
+    throwErrorIfNotArray(value.data);
+    for (const item of value.data) {
```
d0sboots, line 22: Is any of this necessary? Can't we just let JSONMap throw (via the underlying Map constructor)?

You could (rightly) ask, in that case what is the point of any of this? If we're just going to throw anyway, why bother explicitly typechecking? In my mind it comes down to two things: 
1. *When* do we throw? Lots of the loading validation deals with cases where we might otherwise uncritically stick values in fields that have the wrong type. This error then won't be discovered until later, leading to a confusing/useless exception. That is not the case here, where the exception will always happen in the same control flow.
2. *What* do we throw? Often a low-level exception lacks a good explanation or context for what went wrong. But I'm this case, we are already so low-level that our context is no better.

src/utils/SaveDataMigrationUtils.ts
```diff
@@ -0,0 +1,506 @@
+/**
+ * This file contains utility functions that migrate save data. Originally, they were in SaveObject.ts. It's too hard to
```
d0sboots: I agree with this decision. However, to properly diff these changes I'll need my computer, and I only have phone access for the next several days

src/utils/helpers/typeAssertion.ts
```diff
@@ -22,6 +22,30 @@ export function getFriendlyType(v: unknown): string {
...
+export function throwErrorIfNotObject(v: unknown): asserts v is Record<string, unknown> {
+  const type = getFriendlyType(v);
+  if (type !== "object") {
+    console.error("The value is not an object. Value:", v);
+    throw new Error(`The value is not an object. Its type is ${type}. Its string value is ${String(v)}.`);
+  }
+}
+
+/**
+ * This function is different from arrayAssert:
```
d0sboots: I don't think this is a good enough reason to duplicate the function. Since this is an assert, what is thrown (almost) doesn't matter. IMO a better plan would be to change arrayAssert to throw a better Error in a later PR. (The same applies to objectAssert, if the type issue is not critical.)

catloversg:
New changes:
- Revert e7de9688868bf8e1cc04a9b8f7bd499f90db4a56 and some other changes in `SaveObject.ts`. I'll make those changes in another PR.
- Remove usages of `throwErrorIfNotObject` in most `fromJSON`. Some classes still call that function because they need to assert `value.data` before using it (besides `Generic_fromJSON`).
- Change `objectAssert`, `arrayAssert`, `stringAssert`: throw an error instead of a string. I did it in a way that it would not affect current callers of those functions.

About `fromJSON` in `Jsonable.ts`: `value.data` is `unknown`, so TS won't let us pass it `JSONMap` (same thing for `JSONSet`). We have 2 choices:
- Typecasting without checking. Error handling will rely on Map/Set constructor. [1]
- Typecasting after our own checking. [2]

I'm not particularly against [1], but I'm hesitant about doing it. This is an example of how things may go wrong (it's about `JSONSet`, not `JSONMap`):
- Create a corporation.
- Edit the save file: Search: "JSONSet" of "researched\". Change data from `[]` to `\"test\"`.
- Load the edited save file.
- Check the console: `ResearchTree.research() did not find the specified Research node for:`

The game loads normally, but the research data is corrupted.

I know that the example is weird. If the player improperly edits their save file, that's their problem. My point is that the Set constructor does not check the same thing like we do (it will take a string and use it as an array of characters). Maybe there are problems with the Map constructor too.

Nonetheless, I have not found a problem with the Map constructor, so I can do [1] if you insist, but I still prefer [2].

d0sboots:
OK, that's a sufficiently compelling example. Technically it is perfectly legal to create a set from a string, but is rarely what is wanted, and always incorrect in our usage.

src/Bladeburner/Bladeburner.ts
```diff
@@ -1403,9 +1404,10 @@ export class Bladeburner implements OperationTeam {
...
   /** Initializes a Bladeburner object from a JSON save state. */
   static fromJSON(value: IReviverValue): Bladeburner {
+    objectAssert(value.data);
     // operations and contracts are not loaded directly from the save, we load them in using a different method
-    const contractsData = value.data?.contracts;
-    const operationsData = value.data?.operations;
+    const contractsData = value.data.contracts;
+    const operationsData = value.data.operations;
     const bladeburner = Generic_fromJSON(Bladeburner, value.data, Bladeburner.keysToLoad);
     // Loading this way allows better typesafety and also allows faithfully reconstructing contracts/operations
```
d0sboots, line 1412: Unrelated to this PR, but: I don't buy the rationale of this comment. Once again, what the hell, bladeburner.

src/utils/JSONReviver.ts
```diff
@@ -87,16 +88,23 @@ export function Generic_toJSON<T extends Record<string, any>>(
...
-    for (const key of keys) {
+    /**
+     * The type of key is "keyof T", but the type of data is Record<string, unknown>. TypeScript won't allow us to use
+     * key as the index of data, so we need to typecast here.
+     */
+    for (const key of keys as string[]) {
       const val = data[key];
-      if (val !== undefined) obj[key] = val;
+      if (val !== undefined) {
+        // @ts-expect-error -- TypeScript won't allow this action: Type 'T' is generic and can only be indexed for reading.
```
d0sboots, line 105: Can you eliminate this by deferring the cast to `const val = data[key as string]`? Or something similar? 
Basically, this used to work, it seems like the issue comes from the type of key being string.

d0sboots (commented):
I have one comment, but f it. Let's goooooo

src/utils/JSONReviver.ts
catloversg: I tried to mix and match many ways of typecasting, but TS always complained about something. In the past, data is `Record<keyof T, any>`, so we don't have any problems with key (`keyof T`). Now we change data to `unknown`. After asserting, data is `Record<string, unknown>`. I tried `const val = data[key as string]`, but TS complained:
```
Type '{} | null' is not assignable to type 'T[keyof T]'.
  '{} | null' is assignable to the constraint of type 'T[keyof T]', but 'T[keyof T]' could be instantiated with a different subtype of constraint 'any'.
    Type 'null' is not assignable to type 'T[keyof T]'.
      'null' is assignable to the constraint of type 'T[keyof T]', but 'T[keyof T]' could be instantiated with a different subtype of constraint 'any'.
```
If I change the check from `val !== undefined` to `val != null`, TS will complain:
```
Type '{}' is not assignable to type 'T[keyof T]'.
  '{}' is assignable to the constraint of type 'T[keyof T]', but 'T[keyof T]' could be instantiated with a different subtype of constraint 'any'.
```

d0sboots: I think the loop below provides the missing piece: a cast of `val` to `T[keyof T]`.

We should probably also note that the cast (or TS-disable) is *not* safe, in the sense that we can load wrongly-typed data here at runtime. I don't see a way around that without some massive changes.

catloversg: You are right. I'll add comments to this function in a new PR.

src/ThirdParty/sprintf-js.d.ts
gmcew, line 1: I'll happily admit I am well out of my depth here; but isn't this still needed by the first import line of `NetscriptFunctions.ts`? The tests don't get upset, but trying to run the game without it throws an error for me.

d0sboots: I was under the assumption that CatLover had tested thoroughly, since he is usually quite thorough about these things. Could it be a cached artifact issue (does it still happen from a clean tree)?

d0sboots: On the latest head, after `npm install` `npm run start:dev` seems to work fine for me.

catloversg: @gmcew It may be a problem with your `node_modules` folder. Can you delete it, then run `npm i` + `npm run build` again?

gmcew: Thank you both for your patience and help; should have known it was a local issue - clearing it out and reinstalling works fine.

PR #1760, opened 2024-11-08
GANG: Show confirmation popup if player enables territory clashes while being too weak

catloversg:
In the past, the player could enable territory clashes in the main page of Gang UI (Management tab). We removed that toggle, so now the toggle is only on the Territory tab.

I accidentally stumbled on the old (heated) discussion about this change. I agree that having that toggle on the main page is a newbie trap. After reading the whole discussion, I want to implement the second point of Snarling's idea in this comment: https://discord.com/channels/415207508303544321/459097632896188436/1149047761044512799.

Quote:
> Warning dialog? I think it would be a huge improvement to implement a conditional warning dialog for enabling territory warfare if the player has less than a 50% chance to win clashes. Attempting to enable while you have a <50% chance of winning is a pretty good indicator that the player does not know what they're doing and are about to ruin their gang. If they have a >=50% chance to win clashes, no reason to show a dialog because the player seems to know what they are doing. Simple dialog like "You are about to enable Territory Warfare, but you only have a 42% chance to win Territory Clashes, are you sure? Y/N". I don't think the cooldown where clash chance slowly falls back down to 0% is the important thing to highlight here, since that's not what makes enabling warfare a bad decision.

Screenshot:
Capture

### DIFF
--- a/src/Gang/ui/TerritorySubpage.tsx
+++ b/src/Gang/ui/TerritorySubpage.tsx
@@ -5,10 +5,11 @@ import { Help } from "@mui/icons-material";
 
 import { formatNumberNoSuffix, formatPercent } from "../../ui/formatNumber";
 
-import { AllGangs } from "../AllGangs";
+import { AllGangs, getClashWinChance } from "../AllGangs";
 
 import { useGang } from "./Context";
 import { TerritoryInfoModal } from "./TerritoryInfoModal";
+import { PromptEvent } from "../../ui/React/PromptManager";
 
 /** React Component for the territory subpage. */
 export function TerritorySubpage(): React.ReactElement {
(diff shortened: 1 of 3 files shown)


src/Gang/ui/TerritorySubpage.tsx
```diff
@@ -37,7 +38,34 @@ export function TerritorySubpage(): React.ReactElement {
           control={
             <Switch
               checked={gang.territoryWarfareEngaged}
```
d0sboots, line 40: Correct me if I'm wrong, but based on what I know of React, this situation will occur:
* You try to enable warfare
* You don't have enough power, so the prompt comes up
* You select "Cancel" (don't continue with clashes)
* This component manages its own state. The *initial* checked state is based on `{gang.territoryWarfareEngaged}`, but after that it does its own thing. So now the `Switch` will be checked, but warfare is disabled.

d0sboots:
Ah yes, that infamous discussion.

Glad you circled around to address this. Everything looks good, aside from that one quirk.

src/Gang/ui/TerritorySubpage.tsx
catloversg: We are using the `Switch` component as a controlled component (https://legacy.reactjs.org/docs/forms.html#controlled-components) here. Its value is controlled by `gang.territoryWarfareEngaged`, so if we don't set that value, the `Switch` value won't be changed. For example, in the `onChange` handler, if we return immediately without changing `gang.territoryWarfareEngaged`, the `Switch` component will be stuck forever at the initial value of `gang.territoryWarfareEngaged`, even when the player tries to press it.

Video:

https://github.com/user-attachments/assets/76420df7-e365-4d6c-ab09-8ef3d2e6ae6b

d0sboots: I *knew* that this was a controlled component, but I still got the semantics wrong. *Grumble grumble* stupid React *grumble grumble*.

PR #1763, opened 2024-11-08
GANG: Show error popup when there are errors

catloversg:
This PR makes 2 changes:
- Remove the useless check `isNaN(numCycles)`:
  - `process` is only called from `src\engine.tsx`. I checked its callers. `numCycles` cannot be invalid.
  - If `numCycles` is invalid, that check only prints to the console without doing anything else. It's useless anyway.
- When there are errors while processing gang data, show an error popup instead of printing to the console.

### DIFF
--- a/src/Gang/Gang.ts
+++ b/src/Gang/Gang.ts
@@ -97,9 +97,6 @@ export class Gang {
 
   /** Main process function called by the engine loop every game cycle */
   process(numCycles = 1): void {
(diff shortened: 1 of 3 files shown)


d0sboots:
Since this is a popup, won't it end up popping up every 200ms, thus making gameplay impossible?

I'm all for surfacing error messages better, but I feel like there needs to be a middle ground here.

d0sboots:
Great!

PR #1768, opened 2024-11-09
UI: Auto focus hashnet upgrade modal

catloversg:
A player asked why they could not use the Esc button to close the hashnet upgrade modal (the modal that shows hashnet upgrades). All modal dialogs can be closed by pressing the Esc button if they are focused, but that modal is not focused by default. This PR fixes that by auto-focusing the modal.

The usage of `focusRipple` in the button is intentional. Without it, after closing the modal by pressing the Esc button, the button has a weird ripple effect (only on Chrome).
Capture
The documentation says that `focusRipple` is false by default, but I have to explicitly set it to false to fix this weird ripple effect.

### DIFF
--- a/src/Hacknet/ui/HacknetRoot.tsx
+++ b/src/Hacknet/ui/HacknetRoot.tsx
@@ -110,7 +110,7 @@ export function HacknetRoot(): React.ReactElement {
 
       {hasHacknetServers() && (
         <>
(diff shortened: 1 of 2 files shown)


d0sboots:
> The usage of focusRipple in the button is intentional. ...

Please include this as a comment in the source, it's knowledge that belongs with the code itself.

PR #1774, opened 2024-11-14
CODEBASE: Improve type checks in SaveObject.ts

catloversg:
This PR was a part of #1758. That PR is too large now, so I split these changes to a separate PR.

Generally, I try to not make behavior changes, but there are still some exceptions:
- String check: All "loader" functions require a string value to pass into `JSON.parse`. This is why I add the string check in all 3 "groups".
- Default value for `AliasesSave` and `GlobalAliasesSave`: Let's use `AliasesSave` as the example. The old behavior: If `saveObj` does not have `AliasesSave`, use an empty string as the default value. If `loadAliases` fails to load `AliasesSave`, call `loadAliases` again with an empty string. This is nonsense. `loadAliases` cannot process an empty string, so using an empty string just makes it crash. This is why I put them in the "mandatory" group instead of the "optional 1" group.

Note that the change in `typeAssertion.ts` was cherry-picked from #1758.

I know that you cannot do a proper diff for `SaveDataMigrationUtils.ts` in the next few days. There is no need to rush. This PR can wait.

### DIFF
--- a/src/SaveObject.ts
+++ b/src/SaveObject.ts
@@ -25,7 +25,7 @@ import { SaveData } from "./types";
 import { SaveDataError, canUseBinaryFormat, decodeSaveData, encodeJsonSaveString } from "./utils/SaveDataUtils";
 import { isBinaryFormat } from "../electron/saveDataBinaryFormat";
 import { downloadContentAsFile } from "./utils/FileUtils";
-import { handleGetSaveDataInfoError } from "./Netscript/ErrorMessages";
+import { handleGetSaveDataInfoError } from "./utils/ErrorHandler";
 import { isObject, objectAssert } from "./utils/helpers/typeAssertion";
 import { evaluateVersionCompatibility } from "./utils/SaveDataMigrationUtils";
 import { Reviver } from "./utils/GenericReviver";
--- a/src/SaveObject.ts
+++ b/src/SaveObject.ts
@@ -6,14 +6,7 @@ import { CONSTANTS } from "./Constants";
 import { getFactionsSave, loadFactions } from "./Faction/Factions";
 import { loadAllGangs, AllGangs } from "./Gang/AllGangs";
 import { Player, setPlayer, loadPlayer } from "./Player";
-import {
-  saveAllServers,
-  loadAllServers,
-  GetAllServers,
-  createUniqueRandomIp,
-  AddToAllServers,
-  GetServer,
-} from "./Server/AllServers";
+import { saveAllServers, loadAllServers } from "./Server/AllServers";
 import { Settings } from "./Settings/Settings";
 import { loadStockMarket, StockMarket } from "./StockMarket/StockMarket";
 import { staneksGift, loadStaneksGift } from "./CotMG/Helper";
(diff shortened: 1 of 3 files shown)


src/SaveObject.ts
```diff
@@ -83,7 +66,76 @@ export interface ImportPlayerData {
...
+ */
+function assertBitburnerSaveObjectType(saveObject: unknown): asserts saveObject is BitburnerSaveObjectType {
+  objectAssert(saveObject);
+
+  const mandatoryKeysOfSaveObj = [
+    "PlayerSave",
+    "AllServersSave",
+    "CompaniesSave",
+    "FactionsSave",
+    "AliasesSave",
```
d0sboots, line 109: Note to both of us: The first four definitely seem mandatory. The two alias sections might have been added later (I think were?) We need to audit those code paths to ensure there's no sneaky way old saves being migrated could end up failing this check.

d0sboots (commented):
Another thing I will do when I get to my computer is build this and test it against the collection of old player saves I have from various bug reports.

catloversg:
I tried to look at our Git history to find out why the loading code of `Aliases`, `GlobalAliases` and other things in "loadGame" is so weird/inconsistent. This is what I found:

a3c104ca4e670e080034f5f3d394283658505be5 - v0.14 - 23 May 2017
`Aliases` was added in this commit. `SaveObject.js` was really simple at this point. We called `JSON.stringify` in `saveGame` and `JSON.parse` in `loadGame`. There was no try-catch. `Aliases` was an object (not a Map as it is now).

9b8303518734298647775172e4f1fa3884102d2e - v0.14 - 23 May 2017
We implemented a "safety net" for loading `Aliases`:
```js
if (saveObj.hasOwnProperty(AliasesSave)) {
  try {
      Aliases = JSON.parse(saveObj.AliasesSave, Reviver);
  } catch(e) {
      Aliases = {};
  }
} else {
  Aliases = {};
}
```
Starting from this commit, we have added "safety nets" for all saving/loading code. It's usually looked like this:
```js
if (saveObj.hasOwnProperty("something")) {
  try {
      Something = JSON.parse(saveObj.something, Reviver);
  } catch(e) {
      Something = DefautObject/ParseEmptyString;
  }
} else {
  Something = DefautObject/ParseEmptyString;
}
```

6bfe6a549d8adfcd27da9d83c90f11e2cc1b10b7 - v0.22.1 - 30 Jun 2017
`GlobalAliases` was added in this commit. Its saving/loading code was implemented exactly like `Aliases`.

7a05d3585aad55881b753e9513ea15eab1e68a2a - v0.28.0 - 31 Aug 2017
We implemented a special handler for the empty string:
```js
function loadAliases(saveString) {
  if (saveString === "") {
      Aliases = {};
  } else {
      Aliases = JSON.parse(saveString);
  }
}

function loadGlobalAliases(saveString) {
  if (saveString === "") {
      GlobalAliases = {};
  } else {
      GlobalAliases = JSON.parse(saveString);
  }
}
```

ebae35b1fb118b0fb0a399d40750a166f06bc854 - v2.3.0 - 5 May 2023
The type of `Aliases` and `GlobalAliases` was changed to Map. The special handler for the empty string was also removed.

Based on the history, this is my opinion:
- Save data of `Aliases` and `GlobalAliases` has always been a non-empty string (json string of an object) since these features were added in 2017.
- We have had "safety net" code for a long time since 2017, which was the era of rapid development and, IMO, buggy code. Now, those "safety nets" are just legacy code. They are "harmless" and may catch errors in ancient save files, so maybe we will leave them there. They are not super useful, though.
- The special handler for the empty string was gone in 2023. After that, `AliasesSave` and `GlobalAliasesSave` are "mandatory". If they cannot be parsed by `JSON.parse`, the game will crash.

d0sboots:
Good investigation! I agree with your conclusions. I am undecided about keeping or removing the safety net, I can see arguments both ways. (It mostly has to do with the impacts of a bug in the loading code.)

catloversg:
I think it's best to not touch them now. After we refactor "loader" functions and ensure a safe way to load legacy data, we can gradually deal with the legacy code. I have an example of a new implementation of `loadStaneksGift` in #1775.

PR #1775, opened 2024-11-15
CODEBASE: Recheck all usages of typecasting with JSON.parse

catloversg:
While we search and fix lint errors, we (kind of) agree that we should not typecast the result of `JSON.parse` (e.g., `JSON.parse(saveString, Reviver) as Something`. This PR searches all references of `JSON.parse` and checks if they are typecasting in a questionable way. My guideline:
- It's fine to typecast to `unknown`.
- It's fine to typecast after or immediately before the validation code.
- Compromises:
  - If invalid data comes from a third-party tool (RFA client).
  - If proper validation requires an overcomplicated and risky refactor (the god object PlayerObject).

### DIFF
--- a/src/CotMG/Helper.tsx
+++ b/src/CotMG/Helper.tsx
@@ -7,20 +7,21 @@ import { StaneksGift } from "./StaneksGift";
 export let staneksGift = new StaneksGift();
 
 export function loadStaneksGift(saveString: string): void {
-  if (saveString) {
-    const staneksGiftData: unknown = JSON.parse(saveString, Reviver);
-    if (!(staneksGiftData instanceof StaneksGift)) {
-      console.error("Invalid StaneksGiftSave:", saveString);
-      setTimeout(() => {
-        dialogBoxCreate("Cannot load data of Stanek's Gift from the save string. StaneksGift is reset.");
-      }, 1000);
-      staneksGift = new StaneksGift();
-      return;
-    }
-    staneksGift = staneksGiftData;
-  } else {
-    staneksGift = new StaneksGift();
+  let staneksGiftData: unknown;
+  try {
+    staneksGiftData = JSON.parse(saveString, Reviver);
+  } catch (error) {
+    console.error(error);
   }
+  if (!(staneksGiftData instanceof StaneksGift)) {
+    console.error("Invalid StaneksGiftSave:", saveString);
+    staneksGift = new StaneksGift();
+    setTimeout(() => {
+      dialogBoxCreate("Cannot load data of Stanek's Gift. Stanek's Gift is reset.");
+    }, 1000);
+    return;
+  }
+  staneksGift = staneksGiftData;
 }
 
 export function zeros(width: number, height: number): number[][] {
--- a/src/CotMG/Helper.tsx
+++ b/src/CotMG/Helper.tsx
@@ -7,7 +7,11 @@ export let staneksGift = new StaneksGift();
 
 export function loadStaneksGift(saveString: string): void {
   if (saveString) {
-    staneksGift = JSON.parse(saveString, Reviver) as StaneksGift;
+    const staneksGiftData: unknown = JSON.parse(saveString, Reviver);
+    if (!(staneksGiftData instanceof StaneksGift)) {
+      throw new Error(`Cannot load data of Stanek's Gift from the save string. saveString: ${saveString}`);
+    }
+    staneksGift = staneksGiftData;
   } else {
     staneksGift = new StaneksGift();
   }
--- a/src/CotMG/Helper.tsx
+++ b/src/CotMG/Helper.tsx
@@ -1,3 +1,4 @@
+import { dialogBoxCreate } from "../ui/React/DialogBox";
 import { Reviver } from "../utils/JSONReviver";
 import { BaseGift } from "./BaseGift";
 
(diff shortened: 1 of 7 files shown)


catloversg:
In the latest commit, I refactored `loadStaneksGift`. There is a behavior change:
- Old: If `saveString` is an empty string, we load default data; otherwise, we parse `saveString` and pretend that everything is okay. If there is an error, the caller (`loadGame`) will handle it with a try-catch.
- New: Always pass `saveString` to JSON.parse in a try-catch. Validate the result. If it's invalid, we load default data, then show a dialog.

Rationale:
- BN13 was added in 793d9b34ce9848485216502ace7cda727267355e - v0.54.0 - 26 Sep 2021. `saveString` can only be invalid (empty string or invalid data) if the save file is older than v0.54.0 or it's a badly edited one.
- Proper validation.
- When the save string is invalid, instead of silently reset data, we show a dialog to warn the player. This is not a problem, unless they try to load an ancient/invalid save file.

From now on, I'll use this code for other "loader" functions if I need to change them (I'm testing another PR that changes `loadAllGangs`):
```js
let somethingData: unknown;
try {
  somethingData = JSON.parse(saveString, Reviver);
} catch (error) {
  console.error(error);
}
if (!validate(somethingData)) {
  console.error("Invalid somethingData:", saveString);
  somethingData = DefaultData;
  setTimeout(() => {
    dialogBoxCreate("Cannot load data of something. Something is reset.");
  }, 1000);
  return;
}
something = somethingData;
```

After #1774 is merged, with proper "loader" functions, we can gradually deal with legacy code in `loadGame` in a safe way.

src/CotMG/Helper.tsx
```diff
@@ -6,11 +7,21 @@ import { StaneksGift } from "./StaneksGift";
...
 
 export function loadStaneksGift(saveString: string): void {
-  if (saveString) {
-    staneksGift = JSON.parse(saveString, Reviver) as StaneksGift;
-  } else {
+  let staneksGiftData: unknown;
+  try {
+    staneksGiftData = JSON.parse(saveString, Reviver);
+  } catch (error) {
+    console.error(error);
```
d0sboots, line 17: Can we merge this condition into the other error handling? If someone edits their file incompetently, it would be good to be notified (more than just on the console)

src/SaveObject.ts
```diff
@@ -220,11 +220,13 @@ class BitburnerSaveObject {
...
 
     if (!decodedSaveData || decodedSaveData === "") {
-      return Promise.reject(new Error("Save game is invalid"));
+      console.error("decodedSaveData:", decodedSaveData);
+      return Promise.reject(new Error("Save game is invalid. The save data cannot be decoded."));
     }
 
     let parsedSaveData;
     try {
+      // Typecasting here is fine. We will validate parsedSaveData immediately after this line.
```
d0sboots: I don't think we actually need this cast, or it should be possible to get rid of with some code tweaks. The "if" block *does* do the same thing; TS should be able to see from the control-flow that the type is narrowed to this structure after this point. 

Doing it without the cast ensures we haven't missed an edge case.

test/jest/FullSave.test.ts
```diff
@@ -11,12 +11,14 @@ import { Companies } from "../../src/Company/Companies";
 describe("Check Save File Continuity", () => {
   establishInitialConditions();
   // Calling getSaveString forces save info to update
-  saveObject.getSaveData();
+  saveObject.getSaveData().catch((error) => {
```
d0sboots: Since this is a test, we want this error to propagate. Also, presumably we want this sequenced. So, `await` send like the answer here, but the test will have to become async.

d0sboots (commented):
Note to self: Have to test this on the old saves too

test/jest/FullSave.test.ts
catloversg: I'll use `beforeAll`. Other solutions that do not work:
- async `describe`: `describe` does not accept an async function.
- If we put test code inside `then`, Jest does not recognize the tests.

catloversg:
Unrelated question:
We have [`src\utils\helpers\typeAssertion.ts`](https://github.com/bitburner-official/bitburner-src/blob/75cf9c88b54479dffb0aa6146184ba337d92ec8e/src/utils/helpers/typeAssertion.ts) and [`src\utils\TypeAssertion.ts`](https://github.com/bitburner-official/bitburner-src/blob/75cf9c88b54479dffb0aa6146184ba337d92ec8e/src/utils/TypeAssertion.ts). Can I merge them later in another PR (`src\utils\helpers\typeAssertion.ts` -> `src\utils\TypeAssertion.ts`)?

d0sboots:
> Unrelated question: We have [`src\utils\helpers\typeAssertion.ts`](https://github.com/bitburner-official/bitburner-src/blob/75cf9c88b54479dffb0aa6146184ba337d92ec8e/src/utils/helpers/typeAssertion.ts) and [`src\utils\TypeAssertion.ts`](https://github.com/bitburner-official/bitburner-src/blob/75cf9c88b54479dffb0aa6146184ba337d92ec8e/src/utils/TypeAssertion.ts). Can I merge them later in another PR (`src\utils\helpers\typeAssertion.ts` -> `src\utils\TypeAssertion.ts`)?

Please do, that's just silly. *Unless* there are circular dependency issues, but those should be solved by breaking individual functions out, not having two of the same file.

d0sboots:
In testing this change with the old saves I have, I ran into this issue. I don't think this is a problem with this change directly, but rather an earlier undetected regression:

> Error: WorldDaemon is not a normal server. This is a bug. Please contact developers. (at "Loading Screen")

[test_save_0.47.2.json](https://github.com/user-attachments/files/17841573/test_save_0.47.2.json)

d0sboots:
I confirmed that it has the same issue on `dev`. So, an uncaught regression. This PR seems fine across all the test saves I have.

PR #1780, opened 2024-11-19
REFACTOR: Better casting in JSONReviver.ts

d0sboots:
No real code change, only casting changes.


d0sboots:
@catloversg this is the casting I was talking about, that seems to work out

catloversg (approved):
LGTM.

PR #1782, opened 2024-11-19
BUGFIX: Hashserver UI shows wrong server list when purchasing upgrades

catloversg:
#1779 accidentally reverts the fix in #1665 and adds unexpected servers to the server list. How to reproduce this regression bug:
- Load a clean save file with SF9.
- Buy a hashserver.
- Buy a private server (p-server-0).
- Hacknet -> Spend Hashes on Upgrades -> Reduce Minimum Security -> Open the drop-down list. It will contain home, p-server-0, and hacknet-server-0.

The problem is in this code:
```js
return s.hostname !== "home" && !purchased && !Player.hasAugmentation(AugmentationName.TheRedPill, true)
  ? s.hostname !== "w0r1d_d43m0n"
  : true;
```
When s is home, private servers, or hashservers, it always returns true.

While fixing this bug, I spotted other wrong logic in this file, so I fixed all of them in this PR. Those code paths are currently not used (`ServerDropdown` is only used by `src\Hacknet\ui\HacknetUpgradeElem.tsx`, and `HacknetUpgradeElem.tsx` only uses `ServerType.Foreign`.), so it's not a problem.

Code in `isValidServer` does not check members of ServerType properly:
- `purchased` is checked by `(s instanceof Server && s.purchasedByPlayer) || s instanceof HacknetServer`, but home also satisfies `s instanceof Server && s.purchasedByPlayer`, so `purchased` is in fact `isOwnedServer` (home, private servers, hashservers), not `Everything from Owned except home computer`.
- `s.hostname !== "home"` is used in wrong places.

@d0sboots To avoid making a breaking change with `ns.hacknet.spendHashes` (https://github.com/bitburner-official/bitburner-src/blob/d824cd4fa6692ca1e04173b892278b79587a88fb/src/Hacknet/HacknetHelpers.tsx), I don't touch it in this PR, but I want to hear your opinions about it:
- When it sees a hashserver, it returns false without printing a warning.
- Home and private servers are valid targets. It's inconsistent with how we determine the target in the UI. It's not exploitable, so it's not urgent.

Screenshot:
Capture

### DIFF
--- a/src/ui/React/ServerDropdown.tsx
+++ b/src/ui/React/ServerDropdown.tsx
@@ -38,8 +38,8 @@ export function ServerDropdown(props: IProps): React.ReactElement {
    */
   function isValidServer(baseServer: BaseServer): boolean {
     /**
-     * isOwnedServer is true if baseServer is home, private servers, or hacknet servers. Note that home satisfies the
-     * condition: (baseServer instanceof Server && baseServer.purchasedByPlayer).
+     * isOwnedServer is true if baseServer is home, private servers, or hacknet servers. Note that, with home computer,
+     * baseServer.purchasedByPlayer is true.
      */
     const isOwnedServer =
       (baseServer instanceof Server && baseServer.purchasedByPlayer) || baseServer instanceof HacknetServer;
@@ -53,10 +53,10 @@ export function ServerDropdown(props: IProps): React.ReactElement {
           return false;
         }
         // If the player has not installed TRP, exclude WD server.
-        if (!Player.hasAugmentation(AugmentationName.TheRedPill, true)) {
-          return baseServer.hostname !== SpecialServers.WorldDaemon;
-        }
-        return true;
+        return (
+          Player.hasAugmentation(AugmentationName.TheRedPill, true) ||
+          baseServer.hostname !== SpecialServers.WorldDaemon
+        );
       case ServerType.Owned:
         return isOwnedServer;
       case ServerType.Purchased:
--- a/src/ui/React/ServerDropdown.tsx
+++ b/src/ui/React/ServerDropdown.tsx
@@ -13,19 +13,20 @@ import Select, { SelectChangeEvent } from "@mui/material/Select";
 import MenuItem from "@mui/material/MenuItem";
 import Button from "@mui/material/Button";
 import { AugmentationName } from "@enums";
+import { SpecialServers } from "../../Server/data/SpecialServers";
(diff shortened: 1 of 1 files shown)


gmcew:
I'm sorry, I didn't check that logic at all well. Thank you for fixing my error, and highlighting the various others with it!
For the inconsistent hashnet/home/private server handling, and use of "home" string identified in #1781 - I'll raise these as separate issues (#1785 and #1786) for cleaner discussion when these are merged. If that's overkill we can close them.

src/ui/React/ServerDropdown.tsx
```diff
@@ -35,24 +36,35 @@ export function ServerDropdown(props: IProps): React.ReactElement {
    * Checks if the server should be shown in the dropdown menu, based on the
    * 'serverType' property
    */
-  function isValidServer(s: BaseServer): boolean {
-    const purchased = (s instanceof Server && s.purchasedByPlayer) || s instanceof HacknetServer;
+  function isValidServer(baseServer: BaseServer): boolean {
+    /**
+     * isOwnedServer is true if baseServer is home, private servers, or hacknet servers. Note that home satisfies the
+     * condition: (baseServer instanceof Server && baseServer.purchasedByPlayer).
```
d0sboots: I would specifically call out purchasedByPlayer, since the other part should be obvious.

src/ui/React/ServerDropdown.tsx
```diff
@@ -35,24 +36,35 @@ export function ServerDropdown(props: IProps): React.ReactElement {
...
       case ServerType.Foreign:
-        return s.hostname !== "home" && !purchased && !Player.hasAugmentation(AugmentationName.TheRedPill, true)
-          ? s.hostname !== "w0r1d_d43m0n"
-          : true;
+        // Exclude home, private servers, hacknet servers.
+        if (isOwnedServer) {
+          return false;
+        }
+        // If the player has not installed TRP, exclude WD server.
+        if (!Player.hasAugmentation(AugmentationName.TheRedPill, true)) {
```
d0sboots: Optional: Can be simplified to (pseudocode) `return hasAugmentation || hostname !== WorlDaemon`

d0sboots:
Oh, can you show a screenshot of where home used to be in the dropdown and now isn't? We've had a bunch of fixes related to this code, so it would be good to document.

catloversg:
I added a screenshot in PR's description.

PR #1783, opened 2024-11-19
CODEBASE: Validate AllGangs and StockMarket after loading with JSON.parse

catloversg:
Ideally, when we validate loaded data from JSON.parse, we should validate the actual data (checking the data with JSON schemas), not just the "shape" of it (using instanceOf, checking if it's an object with typeof, etc.). Doing that for all json-able classes requires a tremendous effort while the benefit is too small (invalid data indicates buggy saving/loading code or badly edited save file). That's the reason why we have not validated the actual data until now. Originally, I only planned to validate theme data, because it can be set directly by the player via UI. In #1775, I see that AllGangs and StockMarket can also be validated this way, so I made 4 exceptions for that PR:
- AllGangs.
- StockMarket.
- Theme.
- Editor theme.

This PR handles AllGangs and StockMarket.

Test:
AllGangs:
`AllGangs`'s structure has not been changed since at least 9137c2427415c4dc344876a3c30ee445d92ebe4f (v0.43.1). I tested with a save file from v1.0.0.

StockMarket:
`StockMarket`'s structure has not been changed since 1b41e3326ee4098183f7888f652236500ad2d510 (this commit was included in v1.3.0). I tested with a save file from v1.3.0.

Webpack Bundle Analyzer:
Capture

### DIFF
--- a/src/JsonSchema/Data/AllGangsSchema.ts
+++ b/src/JsonSchema/Data/AllGangsSchema.ts
@@ -1,4 +1,4 @@
-import { JSONSchemaType } from "ajv";
+import type { JSONSchemaType } from "ajv";
 import type { AllGangs } from "../../Gang/AllGangs";
 import { GangConstants } from "../../Gang/data/Constants";
 
--- a/src/JsonSchema/Data/AllGangsSchema.ts
+++ b/src/JsonSchema/Data/AllGangsSchema.ts
@@ -3,10 +3,11 @@ import type { AllGangs } from "../../Gang/AllGangs";
 import { GangConstants } from "../../Gang/data/Constants";
 
 /**
- * If we add/remove gangs, we must change 3 places:
+ * If we add/remove gangs, we must change 4 things:
  * - src\Gang\AllGangs.ts: getDefaultAllGangs
  * - src\Gang\data\Constants.ts: GangConstants.Names
  * - src\Gang\data\power.ts: PowerMultiplier
+ * - Save file migration code.
  *
  * Gang code assumes that save data contains exactly gangs defined in these places.
  */
--- a/src/JsonSchema/Data/AllGangsSchema.ts
+++ b/src/JsonSchema/Data/AllGangsSchema.ts
@@ -1,7 +1,15 @@
 import { JSONSchemaType } from "ajv";
 import type { AllGangs } from "../../Gang/AllGangs";
-import { FactionName } from "@enums";
+import { GangConstants } from "../../Gang/data/Constants";
 
+/**
+ * If we add/remove gangs, we must change 3 places:
+ * - src\Gang\AllGangs.ts: getDefaultAllGangs
+ * - src\Gang\data\Constants.ts: GangConstants.Names
+ * - src\Gang\data\power.ts: PowerMultiplier
+ *
+ * Gang code assumes that save data contains exactly gangs defined in these places.
+ */
 export const AllGangsSchema: JSONSchemaType<typeof AllGangs> = {
   $schema: "http://json-schema.org/draft-07/schema#",
   type: "object",
@@ -20,23 +28,7 @@ export const AllGangsSchema: JSONSchemaType<typeof AllGangs> = {
     },
   },
   propertyNames: {
-    enum: [
-      FactionName.SlumSnakes,
-      FactionName.Tetrads,
-      FactionName.TheSyndicate,
-      FactionName.TheDarkArmy,
-      FactionName.SpeakersForTheDead,
-      FactionName.NiteSec,
-      FactionName.TheBlackHand,
-    ],
+    enum: GangConstants.Names,
   },
-  required: [
-    FactionName.SlumSnakes,
-    FactionName.Tetrads,
-    FactionName.TheSyndicate,
-    FactionName.TheDarkArmy,
-    FactionName.SpeakersForTheDead,
-    FactionName.NiteSec,
-    FactionName.TheBlackHand,
-  ],
+  required: GangConstants.Names,
 };
--- a/src/JsonSchema/Data/AllGangsSchema.ts
+++ b/src/JsonSchema/Data/AllGangsSchema.ts
@@ -1,0 +1,42 @@
+import { JSONSchemaType } from "ajv";
+import type { AllGangs } from "../../Gang/AllGangs";
+import { FactionName } from "@enums";
+
+export const AllGangsSchema: JSONSchemaType<typeof AllGangs> = {
+  $schema: "http://json-schema.org/draft-07/schema#",
+  type: "object",
+  patternProperties: {
+    ".*": {
+      type: "object",
+      properties: {
+        power: {
+          type: "number",
+        },
+        territory: {
+          type: "number",
+        },
+      },
+      required: ["power", "territory"],
+    },
+  },
+  propertyNames: {
+    enum: [
+      FactionName.SlumSnakes,
+      FactionName.Tetrads,
+      FactionName.TheSyndicate,
+      FactionName.TheDarkArmy,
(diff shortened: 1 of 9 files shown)


src/JsonSchema/Data/AllGangsSchema.ts
```diff
@@ -0,0 +1,42 @@
...
+      FactionName.SlumSnakes,
+      FactionName.Tetrads,
+      FactionName.TheSyndicate,
+      FactionName.TheDarkArmy,
+      FactionName.SpeakersForTheDead,
+      FactionName.NiteSec,
+      FactionName.TheBlackHand,
+    ],
+  },
+  required: [
```
d0sboots: In the hypothetical universe where we added a new gang faction... I think we would fail validation here once we loaded an old save. How bad is it to be missing a faction? If my suspicion is correct that they start with default values, I think we should drop this clause.

src/JsonSchema/Data/StockMarketSchema.ts
```diff
@@ -0,0 +1,138 @@
...
+ * validation code uses "typeof" to check the data. "typeof functionName" is "function", not "object", so ajv sees it as
+ * invalid data. There are some ways to work around this problem, but all of them are complicated. In the end,
+ * JSONSchemaType is only a utility type. If our schema is designed and tested properly, after the validation, we can
+ * typecast the data.
+ */
+export const StockMarketSchema = {
+  $schema: "http://json-schema.org/draft-07/schema#",
+  type: "object",
+  patternProperties: {
+    "^(?!.*(lastUpdate|Orders|storedCycles|ticksUntilCycle))": {
```
d0sboots: A rule I set for all but the most trivial regex: Explain in a comment what it is intending to match, and give an illustrative example.

src/JsonSchema/Data/StockMarketSchema.ts
```diff
@@ -0,0 +1,138 @@
...
+      type: "object",
+      additionalProperties: false,
```
d0sboots: We don't know about future upgrades, so don't be too rigid. Additional properties don't hurt anything. The specific issue here is loading a newer save from an older game.

src/JsonSchema/Data/StockMarketSchema.ts
```diff
@@ -0,0 +1,138 @@
...
+        },
+        symbol: {
+          type: "string",
+          enum: [...stockSymbolValues],
+        },
+        totalShares: {
+          type: "number",
+        },
+      },
+      required: [...stockKeys],
```
d0sboots: Same issue as with gangs: if we add a new faction, this will fail for old saves.

src/JsonSchema/Data/StockMarketSchema.ts
d0sboots: What's with the ".*" here? It seems incorrect to me.

src/JsonSchema/Data/StockMarketSchema.ts
```diff
@@ -0,0 +1,138 @@
...
+  properties: {
+    lastUpdate: { type: "number" },
+    Orders: {
+      type: "object",
+      patternProperties: {
+        ".*": {
+          type: "array",
+          items: {
+            type: "object",
+            additionalProperties: false,
```
d0sboots: Remove

src/JsonSchema/Data/StockMarketSchema.ts
```diff
@@ -0,0 +1,138 @@
...
+              },
+            },
+            required: [...orderKeys],
+          },
+        },
+      },
+      propertyNames: {
+        enum: [...stockSymbolValues],
+      },
+      required: [],
```
d0sboots: Isn't this a no-op?

src/JsonSchema/Data/StockMarketSchema.ts
```diff
@@ -0,0 +1,138 @@
...
+                type: "string",
+                enum: [OrderType.LimitBuy, OrderType.LimitSell, OrderType.StopBuy, OrderType.StopSell],
+              },
+            },
+            required: [...orderKeys],
+          },
+        },
+      },
+      propertyNames: {
+        enum: [...stockSymbolValues],
```
d0sboots: This also feels like it has potential to bite is on upgrade/downgrade cases. I would give so long hard thought to both loading old saves and loading newer save on an older game, vs what will happen if we load keys that are outside the factions enum.

src/StockMarket/StockMarket.tsx
```diff
@@ -130,27 +136,41 @@ export function cancelOrder(params: ICancelOrderParams, ctx?: NetscriptContext):
...
-    lastUpdate: 0,
-    Orders: {},
-    storedCycles: 0,
-    ticksUntilCycle: 0,
-  } as IStockMarket;
+  StockMarket = getDefaultEmptyStockMarket();
 }
 
 export function initStockMarket(): void {
+  console.log(Object.getOwnPropertyNames(StockMarket));
```
d0sboots: Leftover debug logging

d0sboots (commented):
Very exciting! I wasn't expecting you (or anyone) to take this on, which we'll need to be really careful with because it has lots of implications for old saves.

catloversg:
### AllGangs
`enum` and `required` use `GangConstants.Names` instead of a hardcoded array. Currently, gang code cannot handle invalid data:
- Missing gangs: Gang UI crashes. Errors in `processTerritoryAndPowerGains` and `Achievements.ts` (GANG_TERRITORY).
- Have unknown gangs: Errors in `processTerritoryAndPowerGains`.

If we add or move gangs in the future, we have to refactor our code to deal with missing gangs and unknown gangs. Loading a newer save file with a different gang list in our current code results in bugs.

### StockMarket
I removed many requirements as you requested. Tested:
- Load a valid save file from v1.3.0.
- Add a new stock (ECorp1 with symbol ECP1). Load the game with the save file containing the new (unknown) stock.

src/JsonSchema/Data/AllGangsSchema.ts
```diff
@@ -1,7 +1,15 @@
...
 import type { AllGangs } from "../../Gang/AllGangs";
-import { FactionName } from "@enums";
+import { GangConstants } from "../../Gang/data/Constants";
 
+/**
+ * If we add/remove gangs, we must change 3 places:
+ * - src\Gang\AllGangs.ts: getDefaultAllGangs
+ * - src\Gang\data\Constants.ts: GangConstants.Names
+ * - src\Gang\data\power.ts: PowerMultiplier
+ *
```
d0sboots, line 11: I would also add "save migration code" to this list (doesn't need to be an exact filename, since the exact file might change), because due to these constraints a coordination is *required* if we ever decide to add/remove gangs.

src/JsonSchema/Data/StockMarketSchema.ts
```diff
@@ -25,9 +23,18 @@ export const StockMarketSchema = {
...
+    /**
+     * IStockMarket is an intersection type combining:
+     * - Record<string, Stock> [1]
+     * - {lastUpdate: number;Orders: IOrderBook;storedCycles: number;ticksUntilCycle: number;} [2]
+     *
+     * StockMarketSchema contains:
+     * - patternProperties: Defines [1]. The following regex matches all properties that are not in [2]. It defines the
+     * map of "Full stock name -> Stock".
+     * - properties: Define [2].
+     */
```
d0sboots, line 42: I would add (here or somewhere more appropriate) a result of your investigation into the keys, specifically that having extras is tolerated (and thus adding new ones shouldn't require a save migration).

d0sboots (approved):
Great! Aside from the minor business with comments, this is just waiting on me testing it against old saves.

PR #1789, opened 2024-11-20
CODEBASE: Validate theme, editor theme, and styles

catloversg:
In #1783, we validated AllGangs and StockMarket. This PR validates other things.

This PR is WIP. It's not ready for reviewing now. After we agree on how to deal with debatable things, I'll finish the implementation. After #1783 is merged, I'll rebase this PR (or merge with dev), then mark it as ready for reviewing.

@d0sboots These are the list of debatable things. I want to know your opinions/decisions before continuing with the implementation.
### How strict the schemas should be?
In #1783, we reduced the number of conditions to ensure backward/forward compatibility. With themes/styles, the situation is different:
- Themes and styles can be changed easily by the player via UI/NS APIs, so having invalid data is expected. We may need to use stricter schemas.
- In current code, we are trying to validate/sanitize the input data, so it makes sense to use stricter schemas.
  - Validation in NS APIs (https://github.com/bitburner-official/bitburner-src/blob/04611a30b607aaad2d671206217367c8034ddd2c/src/NetscriptFunctions/UserInterface.ts):
We only use the input theme (from the player) if it matches exactly with the structure of current theme data. If the input theme has a key that the current theme does not have, we tell the player that the input theme is invalid. It's the same with styles.
  - Sanitization attempt in editor theme (https://github.com/bitburner-official/bitburner-src/blob/04611a30b607aaad2d671206217367c8034ddd2c/src/ScriptEditor/ui/themes.ts): We try to "sanitize" the input theme (`sanitizeTheme`). The implementation is flawed and questionable, though. More about it later.

Question: Should we enforce more conditions with the "required" keyword? We have already tried to do that in `UserInterface.ts`. If we don't use the "required" keyword, the strict check in `UserInterface.ts` looks weird. It's as if we are trying to reimplement the "required" keyword of ajv.
### sanitizeTheme - Is it a good idea to use FF0000 as the default value for invalid data?
I don't like this "feature". If it's invalid data, it's best to warn the player, then use a default theme.
### sanitizeTheme - Do we need it now?
Currently, it's used in 2 cases:
- Sanitizing the editor theme when the editor is initializing (`initialize` in `src\ScriptEditor\ScriptEditor.ts`).
- Sanitizing the editor theme after the theme is changed. This is unnecessary after we put the validation logic into `onThemeChange` (`src\ScriptEditor\ui\ThemeEditorModal.tsx`).

The implementation is also flawed. I found bugs in them.

My suggestion: Remove this function. In `initialize` (`src\ScriptEditor\ScriptEditor.ts`), we use the validator to check `Settings.EditorTheme`. If it's invalid, we use a default theme.


d0sboots:
My main overall feeling:
We should be type-correct here. That means, either accept that theme is potentially invalid, have its type be `unknown`, and validate it on use, or give it a full type and validate it thoroughly in all the places where it can be ingested. Right now it sounds like we are in an awkward half-state of both.

Given the number of places that we use the theme, doing the latter (properly validating on ingestion) seems both easier and closer to what is currently happening. If we can share the same functions to do the validation, that would be better. (For instance, if we could re-use the JSON schema validator you plan on adding to also thoroughly validate data coming from the NS API.) I agree with tossing existing implementations, especially if they are buggy - a JSON schema version will be more robust, both in readability, error reporting, and catching edge cases.

d0sboots:
Also, I want to point you to a thing that is simultaneously similar yet also different from what we are doing here: https://stackoverflow.com/questions/31801257/why-required-and-optional-is-removed-in-protocol-buffers-3

I don't know if you have ever encountered Protocol Buffers, but it's something I am very familiar with, and the fact that including `required` as a keyword was considered such a misfeature that it was removed in a later version is something that sticks with me. At the same time, what we are doing here with JSON schema is a subtly but critically different context: Instead of defining a data transfer format, we are defining the *validation* for an existing (implicit) transfer format. Part of the importance of this distinction is that rather than a binary "message is valid"/"message is invalid", *we* control the semantics of what happens if validation fails, because the data has already been parsed. We can choose to use all, some, or none of the object.

This is a long-winded way of saying that I think we *should* define a strict validation for the Theme, but the result of the validation may not be to simply discard the whole theme. (Or maybe it will be.) Missing fields ought to be filled with defaults, at the minimum.

catloversg:
To summarize the problem:
- The validation needs to be strict because the invalid data can be given by the player easily via UI or NS APIs. In current code:
  - The game crashes when it loads an invalid theme/editor theme.
  - When the theme/styles data has unknown properties, we reject the entire theme/styles.
- Too strict validation may not be good. For example, if the given theme has unknown properties, rejecting the entire theme is a bit extreme.
- Using the "required" keyword is not ideal.

My solution:
- Use the `removeAdditional` option (https://ajv.js.org/guide/modifying-data.html#removing-additional-properties) to remove unknown properties. The validation is still strict. We only accept properties defined in schemas. Other unknown properties are stripped.
- Validate and sanitize unknown data in 2 cases:
  - Loading data from the save file: Ensure that the game uses valid data.
  - Receiving unknown data from the player via UI or NS APIs: Ensure that the current valid data is not overwritten by invalid user-provided data.

d0sboots:
I think that approach is good. Let me know when this is not a draft.

catloversg:
@d0sboots It's ready for reviewing now.

Note:
- `src\JsonSchema\Data\StockMarketSchema.ts`: "required" is a hardcoded list now. I had made that change before I implemented #1792. I think it's good to keep that change. `StockMarketSchema` is a utility class, so I should not depend on `Stock` and `Order`.
- I found a minor bug in the editor theme UI. That bug is also in the current stable and dev version, so it's not a regression bug. To avoid making too many changes in this PR, I'll fix it later.
- Feel free to postpone merging my non-bugfix PRs until Snarling releases v2.7.0.

d0sboots:
https://i.kym-cdn.com/entries/icons/original/000/045/422/84_years_meme.jpg

PR #1791, opened 2024-11-21
BUGFIX: Wrong usage of delete operator in Settings.load

catloversg:
I found this bug while implementing #1789.

Test code:
```js
/** @param {NS} ns */
  ns.tail();
  const object = {
    prop1: 1,
    prop2: 2,
  };
  delete object.prop1, object.prop2; // Wrong
  // delete object.prop1; delete object.prop2; // Correct
  ns.print(object);
}
```

### DIFF
--- a/src/Settings/Settings.ts
+++ b/src/Settings/Settings.ts
@@ -164,11 +164,12 @@ export const Settings = {
     save.styles && Object.assign(Settings.styles, save.styles);
     save.overview && Object.assign(Settings.overview, save.overview);
     save.EditorTheme && Object.assign(Settings.EditorTheme, save.EditorTheme);
-    delete save.theme;
-    delete save.styles;
-    delete save.overview;
-    delete save.EditorTheme;
-    Object.assign(Settings, save);
+    Object.assign(Settings, save, {
+      theme: Settings.theme,
+      styles: Settings.styles,
+      overview: Settings.overview,
+      EditorTheme: Settings.EditorTheme,
+    });
     /**
      * The hostname and port of RFA have not been validated properly, so the save data may contain invalid data. In that
      * case, we set them to the default value.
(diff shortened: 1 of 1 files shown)


d0sboots:
Slightly surprised lint doesn't complain about this, but I guess it's considered covered by "don't use delete."

src/Settings/Settings.ts
```diff
@@ -164,7 +164,10 @@ export const Settings = {
     save.styles && Object.assign(Settings.styles, save.styles);
     save.overview && Object.assign(Settings.overview, save.overview);
     save.EditorTheme && Object.assign(Settings.EditorTheme, save.EditorTheme);
-    delete save.theme, save.styles, save.overview, save.EditorTheme;
+    delete save.theme;
+    delete save.styles;
+    delete save.overview;
+    delete save.EditorTheme;
     Object.assign(Settings, save);
```
d0sboots: If the `assign` is written this way, the undesirable save properties will be re-overwritten with the ones from Settings, eliminating the need to delete anything (and thus avoiding a future lint issue).
```suggestion
    Object.assign(Settings, save, {Settings.theme, Settings.styles, Settings.overview, Settings.EditorTheme});
```

Edit: On the whole, the use of assign here is a bit awkward, but possibly necessary if things are keeping references to the settings object. And it's hard to know (without auditing the whole codebase) if that's hapenning.

d0sboots: I see you spelled out all the keys, did the shorthand version not work for some reason?

catloversg: The shorthand version is a syntax error.
Test code:
```js
/** @param {NS} ns */
  const Settings = {
    theme: {
      color: "#fff"
    }
  };
  const save = {
    theme: {
      color: "#000"
    }
  };
  save.theme && Object.assign(Settings.theme, save.theme);
  // Object.assign(Settings, save, { Settings.theme }); // Syntax error
  Object.assign(Settings, save, {
    theme: Settings.theme,
  });
}
```

PR #1795, opened 2024-11-23
MISC: Change current version to 2.7.0

catloversg:
After a long dev cycle, the current version contains tons of bugfixes, tweaks, and 2 major new features. It's best to make it a new minor version instead of a patch version. Snarling agreed on Discord.

Changing the version number is usually handled by Snarling at the start of a dev cycle (commits with `Start x.y.z dev cycle` message), but this time, we decided to do that near the end of the current dev cycle. To make sure that we don't forget doing it at release time, I decided to make this change beforehand.

### DIFF
--- a/src/Constants.ts
+++ b/src/Constants.ts
@@ -4,7 +4,7 @@
  * Constants for specific mechanics or features will NOT be here.
  */
 export const CONSTANTS = {
-  VersionString: "2.6.3dev",
+  VersionString: "2.7.0dev",
   isDevBranch: true,
   VersionNumber: 40,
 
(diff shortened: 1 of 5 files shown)


src/Constants.ts
```diff
@@ -109,7 +109,7 @@ export const CONSTANTS = {
 
   // Also update Documentation/doc/changelog.md when appropriate (when doing a release)
   LatestUpdate: `
-## v2.6.3 Dev: Last updated 15 August 2024
+## v2.7.0 Dev: Last updated 15 August 2024
```
Alpheus, line 112: Should probably update the date on this line as well.

catloversg: We usually update it when Snarling or d0sboots update the content of changelog.

catloversg: We usually update it when Snarling or d0sboots updates the content of changelog.

PR #1802, opened 2024-11-26
JEST: Add tests for b1tflum3 and destroyW0r1dD43m0n API

catloversg:
It's just as the title said.

### DIFF
--- a/test/jest/Netscript/RunScript.test.ts
+++ b/test/jest/Netscript/RunScript.test.ts
@@ -13,9 +13,9 @@ declare const importActual: (typeof EvaluatorConfig)["doImport"];
 // Replace Blob/ObjectURL functions, because they don't work natively in Jest
 global.Blob = class extends Blob {
   code: string;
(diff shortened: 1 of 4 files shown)


test/jest/Netscript/RunScript.test.ts
```diff
@@ -77,11 +77,11 @@ test.each([
   },
 ])("Netscript execution: $name", async function ({ expected: expectedLog, scripts }) {
   global.URL.createObjectURL = function (blob) {
```
d0sboots, line 79: Huh, surprised you couldn't just put the type here.

catloversg: TS wants blob to be exactly `Blob | MediaSource` (signature of `global.URL.createObjectURL`). I tried `global.URL.createObjectURL = function (blob: { code: string }) {`, but TS does not allow doing that without `// @ts-expect-error`.

PR #1803, opened 2024-11-26
MISC: Update changelog

catloversg:
This changelog is updated up to 9d8ac65aaf233db62afc2c8ca3ae6d53adc40d7a.

We don't have a "policy" for writing changelogs, so I use this guideline:
- UI: Only contains changes directly affect UI. Non-UI changes go to MISC. Non-spoiler.
- MISC: Non-spoiler.
- SPOILER CHANGES: All BN-related (spoiler) changes go here.
- CODEBASE/REFACTOR: New section. Players usually don't know or care about things like webpack, lint, CI, Jest, refactoring, etc. All changes that don't affect "gameplay" go here.
- Use prefix "Fix: " for bugfix changes in all sections (except DOCUMENTATION).

Note: I removed this entry: "Add threshold for warning about system clock" and skipped the change in #1739. That feature was implemented by me, but I removed it later. I think it's unnecessary to mention that "feature". I'll revert this change if you want.

This PR conflicts with #1795. I'll resolve conflicts after one of them is merged.


d0sboots:
Thank you very much for doing this! I agree with the "policy" you adopted. I spot-checked the entries, but obviously didn't double-check every single one.

PR #1807, opened 2024-11-28
BUGFIX: Sleeves UI shows and sets wrong task

catloversg:
This PR fixes 2 bugs.

Bug 1:
How to reproduce it:
- Set a sleeve's task to non-Idle.
- Switch to another tab, then switch back to Sleeves tab.
- The current task is correct, but when you press "Set Task", the current task is always set to Idle, regardless of the current task.

Bug 2:
Test code:
```js
  if (min > max) {
  }
}

/** @param {NS} ns */
  ns.tail();
  switch (getRandomIntInclusive(0, 7)) {
    case 0:
      ns.print("setToIdle");
      ns.sleeve.setToIdle(0);
      break;
    case 1:
      ns.print("setToSynchronize");
      ns.sleeve.setToSynchronize(0);
      break;
    case 2:
      ns.print("Crime Shoplift");
      ns.sleeve.setToCommitCrime(0, "Shoplift");
      break;
    case 3:
      ns.print("Crime Mug");
      ns.sleeve.setToCommitCrime(0, "Mug");
      break;
    case 4:
      ns.print("Gym str");
      ns.sleeve.setToGymWorkout(0, "Powerhouse Gym", "str");
      break;
    case 5:
      ns.print("Gym agi");
      ns.sleeve.setToGymWorkout(0, "Powerhouse Gym", "agi");
      break;
    case 6:
      ns.print("Bladeburner Chamber");
      ns.sleeve.setToBladeburnerAction(0, "Hyperbolic Regeneration Chamber");
      break;
    case 7:
      ns.print("Bladeburner Tracking");
      break;
  }
}
```
How to reproduce it:
- Open Sleeves tab.
- Run test code.
- "Earnings" and the description are changed, but 3 dropdown inputs stay the same.

@d0sboots We are going to release v2.7.0 soon. I updated the changelog to reflect the latest change (this PR and #1806). Maybe I'll also update the changelog in my bugfix PRs until Snarling releases the new version. If you want me to move the change in changelog to other PRs, I'll revert the change in this PR.


src/PersonObjects/Sleeve/ui/TaskSelector.tsx
```diff
@@ -290,10 +290,16 @@ function getABC(sleeve: Sleeve): [string, string, string] {
...
 export function TaskSelector(props: IProps): React.ReactElement {
-  const abc = getABC(props.sleeve);
-  const [s0, setS0] = useState(abc[0]);
-  const [s1, setS1] = useState(abc[1]);
-  const [s2, setS2] = useState(abc[2]);
+  const [s0, setS0] = useState("Idle");
+  const [s1, setS1] = useState("------");
+  const [s2, setS2] = useState("------");
+
+  useEffect(() => {
```
d0sboots: This has gone from subtle to even more subtle, and I'm pretty sure the reason is because it was designed wrong in the first place. The state here is attempting to be stored *both* locally in the S0/S1/S2 variables, and also in the parent via setABC. I'm pretty sure that duplication is leading to some of these bugs; it should only be in the parent.

src/PersonObjects/Sleeve/ui/TaskSelector.tsx
```diff
@@ -247,89 +240,49 @@ const canDo: {
...
+  const s2 = props.abc[2];
 
-  const validActions = Object.keys(canDo).filter((k) => (canDo[k] as (sleeve: Sleeve) => boolean)(props.sleeve));
+  const validActions = Object.keys(canDo).filter((taskType) => {
+    const canDoTask = canDo[taskType];
+    if (canDoTask === undefined) {
+      return false;
+    }
+    return canDoTask(props.sleeve);
+  });
```
d0sboots, line 254:
```suggestion
  const validActions = Object.keys(canDo).filter((taskType) => canDo[taskType]?.(props.sleeve));
```
The optional chaining shouldn't even be needed, see if that goes away with my suggestion above.

src/PersonObjects/Sleeve/ui/TaskSelector.tsx
```diff
@@ -247,89 +240,49 @@ const canDo: {
   Synchronize: (sleeve: Sleeve) => sleeve.sync < 100,
 };
```
d0sboots, line 241: Adding `as const` here will fix typing issues below, also you can remove the whole explicit type section.

d0sboots (commented):
Non-blocking cleanup suggestions

PR #1810, opened 2024-11-29
CODEBASE: Use "esnext" target when transforming code with swc

catloversg:
d0sboots requested this change at https://github.com/bitburner-official/bitburner-src/pull/1216#discussion_r1669358105. There is a bug in swc-wasm (https://github.com/swc-project/swc/issues/9495) blocking this change. It's only a minor bug in the definition file, so I decided to just ignore it and make this change.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
       "hasInstallScript": true,
       "license": "SEE LICENSE IN license.txt",
       "dependencies": {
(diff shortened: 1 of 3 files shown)


d0sboots:
I'm going to grab this one because it includes a rather sizable version bump for `swc`, to include in the release. If that bites us in the butt, it's on me.

PR #1812, opened 2024-11-29
MISC: Add source map to transformed scripts

catloversg:
This PR is an alternative for #1496.

Screenshots:
Capture1
Capture2


catloversg:
In the latest commit, I decided to print the error to the console. Please check the comment to see the reason.

Context: https://discord.com/channels/415207508303544321/415207508303544323/1329116690797035672

Capture

src/utils/ErrorHandler.ts
```diff
@@ -18,7 +18,19 @@ export function handleUnknownError(e: unknown, ws: WorkerScript | null = null, i
...
-    if (e.name === "Canceled" && e.message === "Canceled") return;
+    if (e.name === "Canceled" && e.message === "Canceled") {
+      return;
+    }
+    if (ws) {
+      console.error(`An error was thrown in your script. Hostname: ${ws.hostname}, script name: ${ws.name}.`);
+    }
+    /**
+     * If e is an instance of Error, we print it to the console. This is especially useful when debugging a TypeScript
+     * script. The stack trace in the error popup contains only the trace of the transpiled code. Even with a source
```
d0sboots, line 29: The situation with the transpiled trace is particularly unsatisfying right now because it's showing transpiled line numbers, but the original source name. This discrepancy is the biggest issue IMO; I think it will lead to lots of confusion for people using TS/TSX. If the source file names were `.js` or `.transformedJs` or something to make it obvious that the line numbers aren't the same, it would be OK.

d0sboots:
I still have reservations about the situation with the tracebacks in the popup, but it's sort of orthogonal and this is strictly an improvement, so I'm merging as-is.

PR #1815, opened 2024-11-30
API: Add ns.renderTail

catloversg:
Currently, the player can only update the tail window by calling tail-related APIs (`setTitle`, `moveTail`, `resizeTail`). This kind of trick is only known by people who looked at source code. This PR adds a new API to rerender the tail window.


d0sboots:
Gonna sit on this one a little longer, pending the outcome of the poll.

d0sboots:
Merging this as-is, but we should consider moving it per the results of the poll before next release.

PR #1818, opened 2024-12-01
NETSCRIPT: Log script kill immediately and identify the guilty script.

yichizhng:
This PR changes the timing of script death messages; the "Script killed" message is logged immediately by `killWorkerScriptByPid`, instead of waiting for the ScriptDeath exception on the next Netscript function call. The current behavior causes confusing log timings, e.g. if a script is killed during `asleep`.

### DIFF
--- a/src/NetscriptFunctions.ts
+++ b/src/NetscriptFunctions.ts
@@ -294,16 +294,15 @@ export const ns: InternalAPI<NSFull> = {
       if (host === null) {
         throw helpers.errorMessage(ctx, `Cannot find host of WorkerScript. Hostname: ${ctx.workerScript.hostname}.`);
       }
-      const moneyBefore = server.moneyAvailable <= 0 ? 1 : server.moneyAvailable;
-      processSingleServerGrowth(server, threads, host.cpuCores);
+      const moneyBefore = server.moneyAvailable;
+      const growth = processSingleServerGrowth(server, threads, host.cpuCores);
       const moneyAfter = server.moneyAvailable;
       ctx.workerScript.scriptRef.recordGrow(server.hostname, threads);
       const expGain = calculateHackingExpGain(server, Player) * threads;
-      const logGrowPercent = moneyAfter / moneyBefore - 1;
       helpers.log(
         ctx,
         () =>
-          `Available money on '${server.hostname}' grown by ${formatPercent(logGrowPercent, 6)}. Gained ${formatExp(
+          `Available money on '${server.hostname}' grown by ${formatPercent(growth - 1, 6)}. Gained ${formatExp(
             expGain,
           )} hacking exp (t=${formatThreads(threads)}).`,
       );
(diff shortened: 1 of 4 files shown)


d0sboots:
FYI I'm holding off on PRs right now while we ensure 2.7 is stable, and also I think this might conflict with #1813 (which I'll merge first most likely)

catloversg:
`scriptKill` also needs to be patched.
https://github.com/bitburner-official/bitburner-src/blob/6e1848dd79551654d89a40b22b070c981bea6572/src/NetscriptFunctions.ts#L1487

Optional: Change the messages of `spawn` ("About to exit...") and `exit` ("Exiting...") to "Script killed.".
- Before: Those messages are followed by "Script killed." due to the catch block in `src\NetscriptWorker.ts`.
- After: The catch block in `src\NetscriptWorker.ts` is changed, so there is no "Script killed." message.

Printing only "Exiting..." may be confusing for the player: "It's exiting, but when will the "exit" finish?".

src/NetscriptFunctions.ts
```diff
@@ -801,7 +801,10 @@ export const ns: InternalAPI<NSFull> = {
       const killByPid = typeof ident === "number";
       if (killByPid) {
         // Kill by pid
-        res = killWorkerScriptByPid(ident);
+        res = killWorkerScriptByPid(
+          ident,
+          `Script killed by script ${ctx.workerScript.name} with PID ${ctx.workerScript.pid}`,
```
d0sboots: This log line is identical and duplicated, I'd construct it in `killWorkerScriptByPid` instead and pass `ctx.workerScript` as "killer".

src/Netscript/WorkerScript.ts
```diff
@@ -146,6 +146,10 @@ export class WorkerScript {
     }
   }
 
+  logKill(reason: string): void {
```
d0sboots: This function is tiny and only used once, just inline the impl into `killWorkerScriptByPid`.

src/NetscriptWorker.ts
```diff
@@ -352,7 +352,9 @@ Otherwise, this can also occur if you have attempted to launch a script from a t
     .catch(function (e) {
       handleUnknownError(e, workerScript);
       killWorkerScript(workerScript);
-      workerScript.log("", () => (e instanceof ScriptDeath ? "Script killed." : "Script crashed due to an error."));
```
d0sboots: I think we should leave this alone. As CatLover said, there are cases that will be missed if you take it out. Also, the timing with `sleep` and `asleep` is actually an important edge case that IMO we *should* be highlighting: `sleep` will "truly die" immediately, because it will get `delayReject`'ed. `asleep` may not, and if it does it could come much later. The only way of seeing this difference is the timing of the "Script killed" log message.

However, it might be appropriate to change the message, now that (afaik) all cases have an *additional* message at the time when killing is initiated. Perhaps "main() terminated."?

d0sboots (commented):
OK, the other PR is now (finally) merged, and I'm looking at this.

PR #1819, opened 2024-12-01
CORPORATION: Fix wrong warning of sellAmt being negative

catloversg:
In SALE state, we need to calculate and use `sellAmt` like this (pseudocode):
```
Parse from desiredSellAmount
Check typeof number and isFinite
Check if it's negative. If it is, set to 0. [1]
Calculate maxSellPerCycle (material) or maxSellAmount (product)
Calculate sellAmt:
- Limit by maxSellPerCycle
- x10
- Limit by stored
Check if it's negative again. If it is, print warning, set actualSellAmount to 0, then continue/break. [2]
```
We implement it differently between material and product:
- Material does [2] but not [1].
- Product does [1] but not [2].

This PR adds missing steps for both material and product.

How to reproduce this issue:
- Go to "a" division. Set the sell amout of Plants (or Food) to "MAX - 10".
- Check console.

Save file: [test-sellAmt.gz](https://github.com/user-attachments/files/17968608/test-sellAmt.gz)

### DIFF
--- a/src/Corporation/Division.ts
+++ b/src/Corporation/Division.ts
@@ -559,6 +559,10 @@ export class Division {
               );
               continue;
             }
+            // sellAmt must be a non-negative number.
+            if (sellAmt < 0) {
+              sellAmt = 0;
+            }
 
             // Determine the cost that the material will be sold at
             const markupLimit = mat.getMarkupLimit();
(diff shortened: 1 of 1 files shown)


d0sboots:
Can we use this opportunity to merge the logic of these two? 
Either way, gonna hold off on this one until post release, unless you think it's critical.

catloversg:
I refactored it as you wanted. I have played for a while with my scripts. It seems that it works normally.

This change is not urgent. We can merge it later.

d0sboots:
Well, it certainly is later. XD

Great job!

PR #1823, opened 2024-12-03
UI: Change description and add tooltip for HackMoney-related multipliers

catloversg:
When a player enters BN8, they usually ask us why their hacking scripts do not gain money despite "Script Hack Money" being 30%. With the current description of HackMoney-related multipliers, it's hard to know the reason. This PR changes the description and adds tooltips for those multipliers.

Before:
before

After:
after

### DIFF
--- a/markdown/bitburner.bitnodemultipliers.manualhackmoney.md
+++ b/markdown/bitburner.bitnodemultipliers.manualhackmoney.md
@@ -4,7 +4,7 @@
 
 ## BitNodeMultipliers.ManualHackMoney property
 
-Influences how much money can be stolen from a server when the player performs a hack against it through the Terminal.
+Influences how much money is stolen from a server when the player performs a hack against it through the terminal.
 
 **Signature:**
 
--- a/markdown/bitburner.bitnodemultipliers.manualhackmoney.md
+++ b/markdown/bitburner.bitnodemultipliers.manualhackmoney.md
@@ -4,7 +4,7 @@
 
 ## BitNodeMultipliers.ManualHackMoney property
 
-Influences how much money is stolen from a server when the player performs a hack against it through the terminal.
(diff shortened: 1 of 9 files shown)


d0sboots:
If you're fixing the docs here, you should fix `ManualHackMoney` too. Observe https://github.com/bitburner-official/bitburner-src/blob/dev/src/Netscript/NetscriptHelpers.tsx#L526

```js
      let moneyGained = moneyDrained * currentNodeMults.ScriptHackMoneyGain;
      if (manual) {
        moneyGained = moneyDrained * currentNodeMults.ManualHackMoney;
      }
```
Despite the naming, ManualHackMoney is the counterpart to *ScriptHackMoneyGain*.

Note also that (by my reading) there is a difference between NS and terminal here: NS will still drain the money, whereas the terminal will not.

catloversg:
The current logic of 3 hacking methods is:
- Manual hack via the terminal:
  - Server loses money: ScriptHackMoney + ManualHackMoney [1]
  - Player gains money: ScriptHackMoney + ManualHackMoney
- `ns.hack`:
  - Server loses money: ScriptHackMoney
  - Player gains money: ScriptHackMoney + ScriptHackMoneyGain
- `ns.singularity.manualHack`:
  - Server loses money: ScriptHackMoney
  - Player gains money: ScriptHackMoney + ManualHackMoney

https://github.com/bitburner-official/bitburner-src/blob/d9dba2a44280d88263ce24c8a4ac97f07477fe3d/src/Terminal/Terminal.ts#L266-L274
https://github.com/bitburner-official/bitburner-src/blob/d9dba2a44280d88263ce24c8a4ac97f07477fe3d/src/Netscript/NetscriptHelpers.tsx#L509-L529

I think [1] is an oversight. As you said, `ManualHackMoney` is the counterpart of `ScriptHackMoneyGain`. It should affect only "Player gains money". What do you think about it?

In the latest commit, I updated the description of all 3 multipliers.

d0sboots:
I agree that [1] is an oversight, although a minor one because it doesn't provide any advantage (as far as I can tell), and in the few circumstances where it's relevant people aren't hacking from the terminal anymore.

PR #1830, opened 2024-12-08
MISC: Add key binding feature

### LINKED ISSUE #420: Implement a custom keybinding system
E.g. the documentation states that `Alt + t` switches to the terminal, however if I press that, I end up on the factions page because on my keyboard layout (Colemak) T is where F would be on Querty.

I think it would make more sense to use the keyboard event's `key` to make sure the shortcuts match what is documented/expected.

catloversg:
This feature has been requested for a long time. It's the only way to properly fix wrong behaviors of shortcuts.

Tested: I played for a while with US, US Intl, and US Dvorak.

Screenshots:
Capture1
Capture2
Capture3
Capture4

Closes #420.


src/ScriptEditor/ui/ScriptEditorRoot.tsx
```diff
@@ -211,18 +216,28 @@ function Root(props: IProps): React.ReactElement {
...
+      if (Settings.DisableHotkeys) {
+        return;
       }
-
-      // CTRL/CMD + S
-      if (event.code == "KeyS" && (event.ctrlKey || event.metaKey)) {
-        event.preventDefault();
-        event.stopPropagation();
-        save();
+      if (event.getModifierState(event.key)) {
```
d0sboots: IIUC this checks to see if the keypress is a modifier key? Why is that needed? Shouldn't `determineKeyBindingTypes` return an empty array in this case?

src/Settings/Settings.ts
```diff
@@ -209,11 +212,13 @@ export const Settings = {
     } catch (error) {
       console.error(error);
     }
```
d0sboots, line 220: Probably because you wrote this before the sanitization stuff, but there should be a sanitize for key bindings too, right?

src/Sidebar/ui/SidebarRoot.tsx
```diff
@@ -180,81 +189,112 @@ export function SidebarRoot(props: { page: Page }): React.ReactElement {
...
+        case SimplePage.StaneksGift:
+          return canStaneksGift;
+        case SimplePage.Factions:
+          return canOpenFactions;
+        case SimplePage.Augmentations:
+          return canOpenAugmentations;
+        case SimplePage.Sleeves:
+          return canOpenSleeves;
+        case SimplePage.Grafting:
+          // TODO: Change this after PR 1809 is merged.
```
d0sboots: has been merged

src/utils/helpers/keyCodes.ts
```diff
@@ -76,6 +76,7 @@ export enum KEYCODE {
   //CTRL: 17, // Check by `&& event.ctrlKey`
```
d0sboots: This entire enum should be deleted, because all use of keycodes should be gone.

src/utils/KeyBindingUtils.ts
```diff
@@ -0,0 +1,388 @@
...
+] as const;
+
+export type KeyBindingType = (typeof KeyBindingTypes)[number];
+
+export type KeyCombination = {
+  control: boolean;
+  alt: boolean;
+  shift: boolean;
+  meta: boolean;
+  code: string;
```
d0sboots: Here and everywhere: All references to code should be gone. We should be using key only.

src/utils/KeyBindingUtils.ts
```diff
@@ -0,0 +1,388 @@
...
+  const result = new Set<KeyBindingType>();
+  for (const [keyBindingType, combinations] of getRecordEntries(keyBindings)) {
+    for (const combination of combinations) {
+      if (
+        !combination ||
+        combination.control !== keyCombination.control ||
+        combination.alt !== keyCombination.alt ||
+        combination.shift !== keyCombination.shift ||
+        combination.meta !== keyCombination.meta ||
+        combination.code !== keyCombination.code
```
d0sboots: Keybinding *definitely* needs to be based on key, not code. That is the entire impetus of the bug report that led to this.

If you do it like this, then if you switch keyboard layouts the hotkey will remain mapped to the same physical key on the keyboard, instead of being mapped to the logical key. For instance, if I bind save to Ctrl-S while in dvorak (Ctrl-; as labeled by physical keys), then switching to US layout I would have to hit Ctrl-; to save instead of Ctrl-S.

src/Sidebar/ui/SidebarRoot.tsx
```diff
@@ -180,81 +189,112 @@ export function SidebarRoot(props: { page: Page }): React.ReactElement {
...
+      }
+      if ((Player.currentWork && Player.focus) || Router.page() === Page.BitVerse) {
+        return;
+      }
+      const keyBindingTypes = determineKeyBindingTypes(
+        Settings.KeyBindings,
+        convertKeyboardEventToKeyCombination(event),
+      );
+      for (const keyBindingType of keyBindingTypes) {
+        if (keyBindingType === ScriptEditorAction.Save || keyBindingType === ScriptEditorAction.GoToTerminal) {
```
d0sboots: Seems superfluous, given that canGoToPage will return false.

src/ScriptEditor/ui/ScriptEditorRoot.tsx
```diff
@@ -211,18 +216,28 @@ function Root(props: IProps): React.ReactElement {
...
+        return;
+      }
+      const keyBindingTypes = determineKeyBindingTypes(
+        Settings.KeyBindings,
+        convertKeyboardEventToKeyCombination(event),
+      );
+      for (const keyBindingType of keyBindingTypes) {
```
d0sboots: Since you have these as a set, you can do `if keyBindingTypes.has()` blocks instead of a loop+switch. Slightly shorter and also more efficient. It also guarantees that overlap between save+gototerminal will be handled in the order we want.

src/GameOptions/ui/KeyBindingPage.tsx
```diff
@@ -0,0 +1,308 @@
...
+    if (
+      isKeyCombinationPressed(newCombination, { control: true, code: KEYCODE.C }) ||
+      isKeyCombinationPressed(newCombination, { control: true, code: KEYCODE.V }) ||
+      isKeyCombinationPressed(newCombination, { control: true, code: KEYCODE.X })
+    ) {
+      conflicts.add("Common hotkeys");
+    }
+  }
+  // Terminal-ClearScreen
+  if (isKeyCombinationPressed(newCombination, { control: true, code: KEYCODE.L })) {
```
d0sboots: We should probably make this one bindable, but that can be for another day.

src/GameOptions/ui/KeyBindingPage.tsx
```diff
@@ -0,0 +1,308 @@
...
+    conflicts.add("Terminal-ClearScreen");
+  }
+  // Bash hotkeys
+  if (
+    Settings.EnableBashHotkeys &&
+    (isKeyCombinationPressed(newCombination, { control: true, code: KEYCODE.M }) ||
```
d0sboots: The bash hotkeys *definitely* should be bindable, but that can also be for another day.

src/GameOptions/ui/KeyBindingPage.tsx
```diff
@@ -0,0 +1,308 @@
...
+    setCombination(currentKeyCombination);
+    setConflicts(
+      currentKeyCombination
+        ? determineConflictKeys(keyBindingType, isPrimary, currentKeyCombination)
+        : new Set<string>(),
+    );
+    // Add/remove handler and emit an event that notifies subscribers if the player is setting up key bindings.
+    if (open) {
+      document.addEventListener("keydown", handler);
+      KeyBindingEvents.emit(KeyBindingEventType.StartSettingUp);
```
d0sboots, line 178: I'm not clear on why we need these events. Keybindings are disabled while in bitverse; can't the same approach be taken for the keybindings screen? Or is there something I'm missing?

src/GameOptions/ui/KeyBindingPage.tsx
```diff
@@ -0,0 +1,308 @@
...
+          combination is currently used for features that you have not unlocked.
+        </Typography>
+        <br />
+        <Typography>
+          On non-Apple keyboards, the "Windows" key (other names: win, start, super, meta, etc.) is shown as ⊞. On Apple
+          keyboards, the command key is shown as ⌘.
+        </Typography>
+        <br />
+        <Typography>
+          Do NOT use the right Alt key and the AltGr key, especially if you don't use the US keyboard layout. On many
```
d0sboots: You can remove this part. This is not an issue when using key-based binding (as opposed to keycode binding). Or at least, they are no more an issue than any other key that is done through more than a single keystroke. (Compose-based characters, function keys on laptops with a Fn key, etc.)

d0sboots (commented):
I'm *finally* getting around to this.

High-level comments:
First off, great job! The screenshots of the UI look great.

Actually, that's it for high-level comments, the rest is specifics XD

src/GameOptions/ui/KeyBindingPage.tsx
catloversg: I updated the comment to explain it.

src/Sidebar/ui/SidebarRoot.tsx
catloversg: keyBindingType can be a Page (SimplePage or ComplexPage) or a ScriptEditorAction. That check narrows its type down to a Page, so we can pass it to clickPage. In the latest commit, I changed the signature of `canGoToPage` to do that and removed this redundant check.

catloversg:
It's ready for the next review now.

I moved `SimplePage` and `ComplexPage` out of `src\ui\Router.ts` to minimize the dependency chain.
I'll add key bindings for Terminal-ClearScreen and Bash hotkeys in another PR.

d0sboots:
This looks good to me. One last thing is, can you use the deploy action in GitHub to host this at https://catloversg.github.io/bitburner-src (or maybe it's already hosted somewhere else) so we can play with it? I'd like to, but more importantly Zoë expressed a strong desire to have a heads-up on keybinding-related changes, and they're the only person I am certain of who has a Mac to test this on.

catloversg:
It's available at https://catloversg.github.io/bitburner-src/.

d0sboots:
There's one last thing I think we should add before we merge; the rest can be added iteratively later. That is Denis' suggestion of only saving explicitly set keybinds, so that future changes to the defaults won't end up locked in.

This is a chunk of additional complexity (you'll probably need to keep a merged and unmerged version for different purposes), but it's something that lots of other games do and one that I think we will regret it we don't have. And it has to be done now, to be effective. 

Note that this pretty much requires that hitting "default" on the secondary bind restores the secondary default, since hitting "default" will be how you clear a "set" keybind. (Add opposed to "clear", which sets it to clear.)

src/GameOptions/ui/KeyBindingPage.tsx
```diff
@@ -259,45 +283,65 @@ export function KeyBindingPage(): React.ReactElement {
     );
   };
 
+  const keyBindingRows = getRecordKeys(CurrentKeyBindings)
+    .filter(
+      (keyBindingType) =>
+        knowAboutBitverse() || !(SpoilerKeyBindingTypes as unknown as string[]).includes(keyBindingType),
+    )
+    .map((keyBindingType) => {
+      const primaryKeyCombination = CurrentKeyBindings[keyBindingType][0] ? (
```
d0sboots, line 292: Optional: Would this inner part be cleaner as a loop from 0 to 2? It *looks* like everything is just duplicated, but I might have missed something that's subtly different between them.

d0sboots (commented):
Tested the new iteration at https://catloversg.github.io/bitburner-src/

Everything is working as expected now. It still had all the old (lowercase) bindings saved from before, so I was able to test the new default and merge behavior.

Optional comment is for a potential followup. I don't want to delay this any longer, it is in a working state and should get more eyes on it for any issues.

PR #1838, opened 2024-12-10
BUGFIX: Tail log does not render multiline logs properly in edge cases

catloversg:
MRE:
```js
/** @param {NS} ns */
  ns.tail();
  ns.clearLog();
  ns.print("a\n\u001b[39mb");
}
```
In pre-v2.7.0, "a" and "b" are printed on 2 lines. In v2.7.0, they are printed on the same line. This is a regression bug caused by the `display: "inline-block"` style added in #1703.

`inline-block` is supposed to ensure that line height is displayed properly. However, when I test many combinations of tail font size and line height, it seems that we don't need to use `inline-block`.

Test code 1:
```js
/** @param {NS} ns */
  ns.tail();
  ns.clearLog();
  ns.print("| a");
  ns.print("| b");
}
```
Test code 2 (provided by @G4mingJon4s): https://gist.github.com/G4mingJon4s/ba804da307e356268ee1f01f20b4af19

Screenshots:
Tail font size 16, line height 1.5

Tail font size 20, line height 1.5

Tail font size 20, line height 2

Tail font size 25, line height 3

### DIFF
--- a/src/ui/React/LogBoxManager.tsx
+++ b/src/ui/React/LogBoxManager.tsx
@@ -415,7 +415,6 @@ function LogWindow({ hidden, script, onClose }: LogWindowProps): React.ReactElem
                         text={line}
                         color={lineColor(line)}
                         styles={{
-                          display: "inline-block",
                           fontSize: Settings.styles.tailFontSize,
                         }}
                       />


G4mingJon4s:
Good catch! Honestly, I was not sure what the `inline-block` did in the first place. It once fixed line heights for bigger font sizes, but that issue seems to be gone.

d0sboots:
Is this something we should think about cherry-picking, or just try for a faster release cycle this time around?

catloversg:
This PR is not urgent. AFAIK, it only affects edge cases of `ns.print`. Most people print many log lines by calling `ns.print` multiple times instead of combining `\n` with ANSI escape sequences or other special combinations. It's fine to release it in the next version instead of a hotfix patch.

PR #1843, opened 2024-12-11
CODEBASE: Change signature of interpolate function in Infiltration code

catloversg:
This function is pretty weird. It receives an object ("out"), does its job, mutates "out" to output the result, then returns the "min setting". All callers use it to mutate their internal "difficulty" value and ignore the returned value. When I read its implementation, I thought that it was wrong (why does it mutate "out" but then return "a"?), until I found out that all callers ignored the returned value.


d0sboots:
That *is* a weird one!

PR #1852, opened 2024-12-15
NETSCRIPT: Added changing tail font size through scripts

G4mingJon4s:
It is now possible to change the font size for each tail window through the API. The font size is stored on the `RunningScript`, also giving access to the current font size value through `ns.self` and other API functions.

This adds another function to the general netscript functions. Since there are a lot of tail specific functions, it would make sense to move them to their own interface. If this is desired, I would be more than happy to move them in this PR.

Test code:
```ts
  ns.disableLog("ALL");
  ns.clearLog();
  ns.tail();

  ns.print("Test string");
  ns.print("Multi\nLine\nStuff");

  while (true) {
    await ns.asleep(1000);
    ns.setTailFontSize(); // omitting resets the font size to the default, set in the style editor

    await ns.asleep(1000);
    ns.setTailFontSize(20);

    await ns.asleep(1000);
    ns.setTailFontSize(16);

    await ns.asleep(1000);
    ns.setTailFontSize(12);
  }
}
```


d0sboots:
> Since there are a lot of tail specific functions, it would make sense to move them to their own interface. If this is desired, I would be more than happy to move them in this PR.

I made a poll about this, but regardless it would be something for a different PR.

markdown/bitburner.ns.settailfontsize.md
```diff
@@ -0,0 +1,37 @@
...
+
+void
+
+## Remarks
+
+RAM cost: 0 GB
+
+This overwrites the tail font size and forces an update of the tail window's contents.
+
+The font size is saved across restarts.
```
d0sboots, line 34: This has a variety of implications that I'm not so sure about.

For one thing, `temporary` scripts won't be saved (although I think they also don't have logs, or at least you don't tend to look at them.)

In the more general sense, saving this as a property in this way, particularly as a *persistent* property, means that changing the font size in settings won't do anything (because the font size is already specified on all the scripts).

Also also, we already have tailProps, which are non-persisted. This feels like something that belongs there?

src/NetscriptFunctions.ts
```diff
@@ -613,6 +614,19 @@ export const ns: InternalAPI<NSFull> = {
       runningScriptObj.title = typeof title === "string" ? title : wrapUserNode(title);
       runningScriptObj.tailProps?.rerender();
     },
+  setTailFontSize:
+    (ctx) =>
+    (_pixel = Settings.styles.tailFontSize, _pid = ctx.workerScript.scriptRef.pid) => {
+      const pixel = helpers.number(ctx, "pixel", _pixel);
+      const pid = helpers.number(ctx, "pid", _pid);
+      const runningSCriptObj = helpers.getRunningScript(ctx, pid);
```
```suggestion
      const runningScriptObj = helpers.getRunningScript(ctx, pid);
```

src/NetscriptFunctions.ts
```diff
@@ -613,6 +614,19 @@ export const ns: InternalAPI<NSFull> = {
       runningScriptObj.title = typeof title === "string" ? title : wrapUserNode(title);
       runningScriptObj.tailProps?.rerender();
     },
+  setTailFontSize:
+    (ctx) =>
+    (_pixel = Settings.styles.tailFontSize, _pid = ctx.workerScript.scriptRef.pid) => {
```
d0sboots: Optional: Many of our functions (kill, getRunningScript, etc.) operate in the form of a *script ID*, which can either be a pid or a scriptname, hostname and args (some or all of which are optional). There's standard helpers for dealing with those, check out the implementation of `tail` or `getScriptLogs`. It might make sense for this to use the same setup, for API conformity.

d0sboots (commented):
An important part of this design that needs to be considered is "what happens when you change the font size in settings, and have log windows open?" IMO the font size ought to change, at least for tails that haven't been touched by this API.

src/NetscriptFunctions.ts
G4mingJon4s: Sure, I can do that! I just copied the above function `setTitle`. Seems like that didn't use the standard script ID stuff.

markdown/bitburner.ns.settailfontsize.md
G4mingJon4s: I can certainly move the property to the tailProps.
At first I thought it would make more sense to have that stored, since this will mostly come in handy for custom UIs that are just that tiny bit too big and don't fit in the width and height set.
Those properties are lost, so that logic doesn't work out.

That also solves the temporary scripts stuff.
However, the style editor changing the font size is a bit more tricky:
Quickly changing the tail font size should probably stay.
My first idea was to just make the tail font size typed as `number | undefined` and only set it if the font size has been altered. Then the log box will use the font size `fontSize: runningScriptObj.tailProps.fontSize ?? Settings.styles.tailFontSize`.
This way any scripts that don't have custom font sizes will use the default set by the style editor and show the changes, but custom UIs don't get affected and possibly messed up. It wouldn't make sense to scale those as well.

G4mingJon4s:
- Moved the tail font size property to tailProps, no longer making it persistent
- Made tail windows without a custom font size be changed through the style editor
- Changed pid identifier to scriptID

d0sboots:
I was waiting a bit for the resolution of the tail-methods poll (and vacation), but we can (and should) handle moving this in a different PR, if we do it.

PR #1854, opened 2024-12-16
MISC: Convert Literature entries and helper functions to TSX

NagaOuroboros:
Following the conclusion of discussion on the Discord, I have brought the Literature-related files in-line with the recommendations made so that, going forward, contribution and maintenance of Literature files should be both easier and more flexible.

### Literatures.ts -> Literatures.tsx
- Reformatted every `Literature` entry's `text` property to hold a JSX Element.
- Miscellaneous edits to entries to use JSX features and fix parsing errors so the final output is unchanged.
- Replaced single quotes within the entries with double quotes, since the `text` property is no longer comprised of string literals.
### LiteratureHelpers.ts -> LiteratureHelpers.tsx
- Updated `showLiterature()` to pass a JSX Element to `dialogueBoxCreate()` instead of a string.
### Literature.ts
- Changed the `LiteratureConstructorParams` interface's `text` field from type `string` to type `JSX.Element`.
- Changed the `Literature` class's `text` field from type `string` to type `JSX.Element`.

### DIFF
--- a/src/Literature/LiteratureHelpers.tsx
+++ b/src/Literature/LiteratureHelpers.tsx
@@ -2,7 +2,7 @@ import { Literatures } from "./Literatures";
 import { dialogBoxCreate } from "../ui/React/DialogBox";
 import { LiteratureName } from "@enums";
 import { Player } from "@player";
-import React from 'react'
+import React from "react";
 
 export function showLiterature(fn: LiteratureName): void {
   const litObj = Literatures[fn];
@@ -12,6 +12,13 @@ export function showLiterature(fn: LiteratureName): void {
   for (const factionName of litObj.factionRumors) {
     Player.receiveRumor(factionName);
   }
-  const txt = <><i>{litObj.title}</i><br/><br/>{litObj.text}</>;
+  const txt = (
+    <>
+      <i>{litObj.title}</i>
+      <br />
+      <br />
+      {litObj.text}
+    </>
+  );
   dialogBoxCreate(txt, true);
 }
--- a/src/Literature/LiteratureHelpers.tsx
+++ b/src/Literature/LiteratureHelpers.tsx
@@ -20,5 +20,5 @@ export function showLiterature(fn: LiteratureName): void {
       {litObj.text}
     </>
   );
-  dialogBoxCreate(txt, true);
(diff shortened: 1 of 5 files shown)


src/Literature/LiteratureHelpers.tsx
```diff
@@ -11,6 +12,13 @@ export function showLiterature(fn: LiteratureName): void {
...
-  const txt = `<i>${litObj.title}</i><br><br>${litObj.text}`;
+  const txt = (
+    <>
+      <i>{litObj.title}</i>
+      <br />
+      <br />
+      {litObj.text}
+    </>
+  );
   dialogBoxCreate(txt, true);
```
catloversg:
```suggestion
  dialogBoxCreate(txt);
```
The second parameter should not be used. We only use it in 2 cases:
- Literature. We concatenate strings to create an HTML string, so we need HTML support in `dialogBoxCreate`.
- `ns.alert`. Legacy support.

Literature content is a React element now, so there is no need to call `dialogBoxCreate` with html=true.

NagaOuroboros: Done!

I only left it in due to my own uncertainty over what it controlled, and didn't try removing it, as it didn't produce any erroneous output.  Thanks for the pointer!

src/Literature/LiteratureHelpers.tsx
```diff
@@ -2,6 +2,7 @@ import { Literatures } from "./Literatures";
 import { dialogBoxCreate } from "../ui/React/DialogBox";
 import { LiteratureName } from "@enums";
 import { Player } from "@player";
+import React from "react";
```
d0sboots, line 5: I don't think this is actually needed? (You're not referencing any react types.)

catloversg: I'm pretty sure they need to add it to fix this error:
```
'React' must be in scope when using JSX eslintreact/react-in-jsx-scope
'React' refers to a UMD global, but the current file is a module. Consider adding an import instead.ts(2686)
```

d0sboots: Ah, I didn't realize it was needed just to use JSX syntax.

PR #1856, opened 2024-12-16
BLADEBURNER: Add visual cues to warn player of dangerous actions and status of population, chaos

catloversg:
There are 2 things that can destroy a Bladeburner run:
- Kill too much population.
- Not manage chaos.

There is no visual cue that warns the player about dangerous actions. Bladeburner has some problems with its balance (unbalanced chaos of IV, most ops kill pop too fast, etc.). Ideally, we should rebalance Bladeburner to fix these issues, but it takes time to analyze and test potential solutions. I'll do that later.

This PR focuses on helping the player avoid making common mistakes by providing them more information and warning them about dangerous things in Bladeburner.

Capture1
Capture2
Capture3
Capture4

### DIFF
--- a/src/Bladeburner/data/Operations.ts
+++ b/src/Bladeburner/data/Operations.ts
@@ -9,8 +9,8 @@ export function createOperations(): Record<BladeburnerOperationName, Operation> 
     [BladeburnerOperationName.Investigation]: new Operation({
       name: BladeburnerOperationName.Investigation,
       desc:
-        "As a field agent, investigate and identify Synthoid populations, movements, and operations.\n\n" +
-        "Successful Investigation ops will increase the accuracy of your synthoid data.\n\n" +
+        "As a field agent, investigate and identify Synthoid populations, movements, and operations.\n" +
+        "Successful Investigation ops will increase the accuracy of your synthoid data.\n" +
         "You will NOT lose HP from failed Investigation ops.",
       successScaling: "Significantly affected by Hacking skill and Charisma. Minor bonus from combat stats.",
       baseDifficulty: 400,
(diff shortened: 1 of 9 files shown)


src/Bladeburner/data/Operations.ts
```diff
@@ -148,7 +153,9 @@ export function createOperations(): Record<BladeburnerOperationName, Operation>
       name: BladeburnerOperationName.StealthRetirement,
       desc:
         "Lead a covert operation to retire Synthoids. The objective is to complete the task without drawing any " +
-        "attention. Stealth and discretion are key.",
+        "attention. Stealth and discretion are key.\n" +
+        "Completing this operation will DECREASE the chaos level of your current city. If you complete it successfully, it will decrease the Synthoid population of your current city.",
```
gmcew, line 157: This is the only all caps "DECREASE" - change to lower case for consistency?

gmcew (commented):
The changes look clean, the wording's good!

I have a minor concern that the individual warning labels detailing which actions are percentage based rather than constants risks spoiling some of the experimentation around the mechanic, particularly if we're considering future rebalancing for these actions.

Would it be reasonable to only show those warnings (or the full detail) when the respective chaos/population condition is in "warning" or "error" state, rather than continuously?

catloversg:
These dangerous actions tend to decrease population and increase chaos too fast, so I worry that new players may make a (reasonable) mistake and accidentally destroy their BB run. For example:
- Run IV, then take a short break (30 minutes): Chaos is ~135e3. They need to run Diplomacy for ~5.5 hours per city (1000 charisma).
- Run IV, then take a nap: (2 hours): Chaos is ~1e10. They need to run Diplomacy for ~13.3 hours per city (1000 charisma).

If we only show these action warnings after chaos/pop reaches a threshold, we need to set that threshold very low to make sure that they can warn the player before things go out of control. I'm not sure if it's useful.

About the ALL-CAPS "DECREASE": I want to emphasize the difference. Most dangerous actions increase chaos, except this one.

@d0sboots What's your opinion about action warnings and the ALL-CAPS "DECREASE"?

d0sboots:
I'm fine with this. A lot of people never notice what's in tooltips, anyway. (And it's easy to ignore them if you don't want to see what they're telling you.)

PR #1857, opened 2024-12-16
BUGFIX: Saved running scripts are started before SWC is initialized

### LINKED ISSUE #1853: TypeScript scripts crash on game load
# Overview

In version v2.7.0 (fbe1546fd) `*.ts` script that are running on game save crash with a message `Script crashed due to an error.` when you load the game. `*.js` scripts work as expected and start running again on game load.

# Reproduction

Run both of the scripts in the attached archive. Save the game, close it and load the savegame. The .js file starts running again, while the .ts script crashes.
[crashtest.zip](https://github.com/user-attachments/files/18141751/crashtest.zip)

catloversg:
Saved running scripts are started in `Engine.load`, but `Engine.load` is called at the same time as `initSwc`. Therefore, SWC is usually not initialized when we use it to transform TS scripts.

Fixes #1853.

### DIFF
--- a/src/ui/LoadingScreen.tsx
+++ b/src/ui/LoadingScreen.tsx
@@ -34,9 +34,8 @@ export function LoadingScreen(): React.ReactElement {
   });
 
   useEffect(() => {
-    initSwc()
-      .then(() => load())
-      .then((saveData) => Engine.load(saveData))
+    Promise.all([initSwc(), load()])
+      .then(([__, saveData]) => Engine.load(saveData))
       .then(() => {
         pushGameReady();
         setLoaded(true);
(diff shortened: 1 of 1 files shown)


src/ui/LoadingScreen.tsx
```diff
@@ -34,8 +34,9 @@ export function LoadingScreen(): React.ReactElement {
   });
 
   useEffect(() => {
-    load()
-      .then((saveData) => Promise.all([initSwc(), Engine.load(saveData)]))
+    initSwc()
```
d0sboots: initSwc just needs to happen before `Engine.load`, not `load()`, right?
To save some time on init (hopefully) can we `Promise.all([initSwc, load()]).then(Engine.load())`?

PR #1859, opened 2024-12-16
UI: Remove mention of passive reputation gain when player is in BN2

catloversg:
Context from Discord:
```
Really minor, and I'm not sure the cleanest way to address it, but I just started BN2, and I noticed that the text from [this line](https://github.com/bitburner-official/bitburner-src/blob/dev/src/Faction/ui/Info.tsx#L37) mentions passive rep gain, despite that being disabled in BN2.
```
Screenshots:
BN2:
bn2
Other BNs:
non-bn2

### DIFF
--- a/src/Faction/ui/Info.tsx
+++ b/src/Faction/ui/Info.tsx
@@ -17,7 +17,7 @@ import Tooltip from "@mui/material/Tooltip";
 import Box from "@mui/material/Box";
 import { useCycleRerender } from "../../ui/React/hooks";
 import { calculateFavorAfterResetting } from "../formulas/favor";
-import { Player } from "@player";
+import { knowAboutBitverse } from "../../BitNode/BitNodeUtils";
 
 interface IProps {
   faction: Faction;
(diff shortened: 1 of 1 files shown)


src/Faction/ui/Info.tsx
```diff
@@ -33,9 +34,10 @@ const useStyles = makeStyles()({
 function DefaultAssignment(): React.ReactElement {
   return (
     <Typography>
-      Perform work/carry out assignments for your faction to help further its cause! By doing so you will earn
-      reputation for your faction. You will also gain reputation passively over time, although at a very slow rate.
-      Earning reputation will allow you to purchase Augmentations through this faction, which are powerful upgrades that
+      Perform work/carry out assignments for your faction to help further its cause! By doing so, you will earn
+      reputation for your faction.&nbsp;
+      {Player.bitNodeN !== 2 && <>You will also gain reputation passively over time, although at a very slow rate. </>}
```
d0sboots: I don't like the documentation secretly changing depending on what BN you're in. However, it also shouldn't spoil.
What about adding a conditional based on knowsAboutBitverse that adds a note that the passive rep gain doesn't happen in BN2?

PR #1862, opened 2024-12-19
UI: Share RAM to boost reputation gain

catloversg:
Sharing RAM is one of the most obscure features in this game. Except this one, all basic features can be explored/discovered via UI. I have seen many players completing BN1 without knowing that this feature exists. This PR adds a new "component" in Faction page that allows the player to use this feature via UI. It's an "introduction" for this feature, so I don't add any fancy features (e.g., allow choosing non-home servers).

Capture


src/NetworkShare/Share.ts
```diff
@@ -0,0 +1,57 @@
...
+import { calculateIntelligenceBonus } from "../PersonObjects/formulas/intelligence";
+import { getCoreBonus } from "../Server/ServerHelpers";
+import { clampNumber } from "../utils/helpers/clampNumber";
+
+let sharePower = 1;
+
+export const ShareBonusTime = 10000;
+
+/**
+ * When the player shares free RAM via UI, it's a "pending job". After that job finishes, we restore the free RAM by
```
d0sboots, line 11: Why isn't this an issue for ns shares?

d0sboots:
Nice, it's technically useable but not practically useful because who's going to keep hitting the button every 10 seconds? XD
But it does guide the player towards its existence.

src/NetworkShare/Share.ts
catloversg: At the end of a sharing attempt, we need to do 2 things:
- Call the returned callback of `startSharing` to reduce `sharePower`.[1]
- Restore the free RAM.[2]

In UI, we do both things in `onShare` (`src\Faction\ui\ShareOption.tsx`). In order to do [2], we have to manually alter `ramUsed` by calling `server.updateRamUsed`, so we have to be careful with prestige cases. With `ns.share`, we only need to do [1] in `src\NetscriptFunctions.ts`. [2] is guaranteed to be done properly by our architecture.

catloversg:
Yeah. This new feature is just a fancy way to introduce `ns.share` to new players.

PR #1865, opened 2024-12-20
BUGFIX: Edited code disappears in editor after switching tab

### LINKED ISSUE #1885: Editor - Reverts Changes While Moving Tabs
I've noticed since 2.7.0 update when I'm working in vim, if I move between tabs, the code/file will revert back things I've already written. It doesn't happen all the time just randomly from what I can see. Eg

If the original file is:

/** @param {NS} ns */
	ns.tail()
	ns.clearLog()
	ns.print(ns.ps("home"))
}

And I put something in like 

/** @param {NS} ns */

	ns.tail()
	ns.clearLog()
	ns.print(ns.ps("home"))

       var procs = ns.ps("home")

       for(const proc of procs){
           ns.print(proc)
       } 
}

If I switch tabs or move to something else, it will revert the file back to original file state. I've been testing saving the file more often which seems to fix it. But I shouldn't have to save it.

catloversg:
How to reproduce this bug:
- Load a clean save file.
- `nano test1.js`. Save it. After this step, test1.js is in your home.
- Reload the page.
- `nano test1.js`. Edit it, but do NOT save it.
- `nano test2.js`.
- Switch to the terminal tab, then switch back to the editor tab. The edited code disappears.

Please check the comment to see the root cause and the fix.

Fixes #1885.

### DIFF
--- a/src/ScriptEditor/ui/ScriptEditorRoot.tsx
+++ b/src/ScriptEditor/ui/ScriptEditorRoot.tsx
@@ -174,10 +174,10 @@ function Root(props: IProps): React.ReactElement {
       return;
     }
 
-    server.scripts.forEach((s) => {
+    server.scripts.forEach((script) => {
       const uri = monaco.Uri.from({
         scheme: "file",
-        path: `${s.server}/${s.filename}`,
+        path: `${script.server}/${script.filename}`,
       });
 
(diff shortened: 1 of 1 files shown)


catloversg:
Superseded by #1893.

PR #1866, opened 2024-12-20
BUGFIX: Editor shows error popup when opening scripts on "." server

catloversg:
This regression bug appeared when we updated monaco-editor in #1468.

How to reproduce it: Open any scripts on "." server.

Capture

Relevant code:
https://github.com/microsoft/TypeScript/blob/56a08250f3516b3f5bc120d6c7ab4450a9a69352/src/compiler/path.ts#L632-L649

`file:///./a.js` is normalized to `file:///a.js`, but the model URI is still `file:///./a.js`. With another scheme (e.g., "memory"), `normalizePath` does not strip the "./" part. For example, with "memory" scheme, URI of "a.js" in "." server is `memory:./a.js`.

If you are curious about why it does not happen in old version, you can check the change in `getOrCreateSourceFileByPath` in https://github.com/microsoft/TypeScript/commit/5aa0053c74f0c67b2cbedf9cee6607ead3b9dc45.

### DIFF
--- a/src/ScriptEditor/ui/ScriptEditorRoot.tsx
+++ b/src/ScriptEditor/ui/ScriptEditorRoot.tsx
@@ -176,7 +176,7 @@ function Root(props: IProps): React.ReactElement {
 
     server.scripts.forEach((s) => {
       const uri = monaco.Uri.from({
-        scheme: "file",
+        scheme: "memory",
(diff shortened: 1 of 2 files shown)


catloversg:
Superseded by #1893.

PR #1867, opened 2024-12-20
BUGFIX: Editor loads scripts at wrong time

catloversg:
Test scripts:
test.ts
```js
import { multiply } from "./lib";

  ns.print(`multiply: ${multiply(2, 3)}`);
}
```
lib.ts
```js
```
How to reproduce this bug:
- Load a clean save file.
- Save test scripts. Reload the page.
- `nano test.ts`.
- Switch to the terminal tab, then switch back to the editor tab.

Capture

`loadAllServerScripts` is called in `parseCode`. There are 2 problems:
- `loadAllServerScripts` is called every time the player edits their code. This means that it's called after every keystroke.
- Even in this case, calling it in `parseCode` is still too late. It needs to be called much earlier. The start of `onMount` is a proper place. I also call it in `onTabClick`. It may be useful if something else (via NS APIs or RFA) changes (add/edit/delete) scripts. For example, if an external editor pushes a script via RFA, we will load that new script when the player switches tabs. Technically, doing that in `debouncedCodeParsing` may be better, but I think it's overkill.

### DIFF
--- a/src/ScriptEditor/ui/ScriptEditorRoot.tsx
+++ b/src/ScriptEditor/ui/ScriptEditorRoot.tsx
@@ -212,13 +212,13 @@ function Root(props: IProps): React.ReactElement {
   }, 300);
 
   const parseCode = (newCode: string) => {
-    loadAllServerScripts();
     startUpdatingRAM();
     debouncedCodeParsing(newCode);
   };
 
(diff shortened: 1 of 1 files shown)


catloversg:
Superseded by #1893.

PR #1869, opened 2024-12-21
UI: Hostname No Longer Resets As You Type It In The Remote File API Screen

femboyfireball:
As reported in the BitBurner discord, https://discord.com/channels/415207508303544321/415213413745164318/1319747074824081508, inputting an invalid hostname into the Remote File API menu was difficult as it would simply "void" your input.
This made attempting to input certain hostnames like an ipv6 address impossible without directly pasting it in. This PR essentially removes that behaviour, however the game still does not connect to an invalid URL


catloversg:
I'll handle this issue in #1870.

femboyfireball:
Got it, that one's an actual fix and not just a hack. I'll close the PR and let d0s handle #1870

PR #1872, opened 2024-12-22
BUGFIX: Grow log shows invalid values in edge cases

catloversg:
In 23c862b72710bd38327fa7449dbcacb90629ffd6, we fixed the case in which `ns.grow` returns Infinity. On Discord, we received a report that terminal grow shows NaN%. When I checked the implementation of `processSingleServerGrowth` and its callers, I found more bugs:
- On servers with maxMoney = 0:
  - Terminal grow shows NaN.
  - `ns.grow` shows "-100%".
- On "grow-able" servers: Terminal grow shows Infinity if server's money before growing is 0.

I carefully refactored code of `processSingleServerGrowth` and its callers so that there is no behavior change in `ns.grow`.

Before:
before1
before2
before3

After:
after1
after2
after3

### DIFF
--- a/src/NetscriptFunctions.ts
+++ b/src/NetscriptFunctions.ts
@@ -298,24 +298,23 @@ export const ns: InternalAPI<NSFull> = {
       }
       const moneyBefore = server.moneyAvailable;
       const growth = processSingleServerGrowth(server, threads, host.cpuCores);
-      const growthPercentForLogging = growth !== 0 ? growth - 1 : 0;
       const moneyAfter = server.moneyAvailable;
       ctx.workerScript.scriptRef.recordGrow(server.hostname, threads);
       const expGain = calculateHackingExpGain(server, Player) * threads;
       helpers.log(
         ctx,
         () =>
-          `Available money on '${server.hostname}' grown by ${formatPercent(
-            growthPercentForLogging,
-            6,
-          )}. Gained ${formatExp(expGain)} hacking exp (t=${formatThreads(threads)}).`,
+          `Available money on '${server.hostname}' grown by ${formatPercent(growth - 1, 6)}. Gained ${formatExp(
+            expGain,
+          )} hacking exp (t=${formatThreads(threads)}).`,
       );
       ctx.workerScript.scriptRef.onlineExpGained += expGain;
       Player.gainHackingExp(expGain);
       if (stock) {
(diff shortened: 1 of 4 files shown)


src/Server/ServerHelpers.ts
```diff
@@ -188,6 +188,14 @@ export function processSingleServerGrowth(server: Server, threads: number, cores
     usedCycles = Math.min(Math.max(0, Math.ceil(usedCycles)), threads);
     server.fortify(2 * ServerConstants.ServerFortifyAmount * usedCycles);
   }
+  // Prevent returning NaN. This happens when server.moneyMax is 0.
+  if (server.moneyAvailable === 0 && oldMoneyAvailable === 0) {
+    return 0;
```
d0sboots: Why not return 1 here? It's arguably at least as sensible as 0 ("0/0, cancel the 0s to get 1"), and it avoids needing a ternary everywhere you use this (can just subtract 1 and log).

src/NetscriptFunctions.ts
```diff
@@ -298,24 +298,23 @@ export const ns: InternalAPI<NSFull> = {
...
+          )} hacking exp (t=${formatThreads(threads)}).`,
       );
       ctx.workerScript.scriptRef.onlineExpGained += expGain;
       Player.gainHackingExp(expGain);
       if (stock) {
         influenceStockThroughServerGrow(server, moneyAfter - moneyBefore);
       }
-      return Promise.resolve(growth);
+      // growth can only be 1 if server.moneyMax is 0. In that case, this API returns 0.
+      return Promise.resolve(growth !== 1 ? growth : 0);
```
d0sboots: Ah hell. I'm not certain that this is true. For instance, if you try to grow when already at max money, I'm 95% sure you'll get to here and get 1 as the value of grow.

The real question is, is it fine to return 1 (instead of 0) in this case? It's a weird case because we throw for other "invalid" servers, but return a value here. But probably someone is relying on it, sigh.
```suggestion
      return Promise.resolve(server.moneyMax === 0 ? 0 : growth);
```

catloversg: You are right. I forgot the case in which the player grows a server that has already been at max money.

For "return 1 vs 0", I think returning 1 is also fine. It's sad that this behavior is another case of "legacy behavior".

PR #1880, opened 2024-12-24
UI: Update description of "cat" in the "help" command

ogllgo:
The description of cat doesn't reference .js, .jsx,.ts, or .tsx files. This generalises the description.

### DIFF
--- a/src/Terminal/HelpText.ts
+++ b/src/Terminal/HelpText.ts
@@ -5,7 +5,7 @@ export const TerminalHelpText: string[] = [
   "    analyze                          Get information about the current machine ",
   "    backdoor                         Install a backdoor on the current machine ",
   "    buy [-l/-a/program]              Purchase a program through the Dark Web",
-  "    cat [file]                       Display a .msg, .lit, or .txt file",
+  "    cat [file]                       Display the contents of a text file",
   "    cd [dir]                         Change to a new directory",
   "    changelog                        Display changelog",
   "    check [script] [args...]         Print a script's logs to Terminal",


src/Terminal/HelpText.ts
```diff
@@ -5,7 +5,7 @@ export const TerminalHelpText: string[] = [
   "    analyze                          Get information about the current machine ",
   "    backdoor                         Install a backdoor on the current machine ",
   "    buy [-l/-a/program]              Purchase a program through the Dark Web",
-  "    cat [file]                       Display a .msg, .lit, or .txt file",
+  "    cat [file]                       Display the contents of a text file",
```
catloversg, line 8: We have many file types:
- Message files (.msg)
- Literature files (.lit)
- Text files (.txt, .json)
- Script files (.js, .jsx, .ts, .tsx, .script)

```suggestion
  "    cat [file]                       Display message files (.msg), literature files (.lit), text files (.txt, .json), and script files (.js, .jsx, .ts, .tsx, .script)",
```

catloversg (changes requested):
You also need to update the text at: https://github.com/bitburner-official/bitburner-src/blob/a6d6aa7410658a4dbbeb3de6ff10dd031d5db900/src/Terminal/HelpText.ts#L143

Minor suggestion: change the title of this PR to something else. For example: `UI: Update help text of "cat" CLI`.

src/Terminal/HelpText.ts
d0sboots: I think I'd actually just say "display the contents of a file"
This help text is meant to be as brief as possible, and while there *are* some files cat won't display, they are rare.

PR #1881, opened 2024-12-24
UI: Disable buttons when player cannot buy things in tech vendor

catloversg:
This PR is an alternative for #1469. Differences:
- Refactor code to avoid unnecessary code when the player cannot buy things (d0sboots calls it "a different (new) button" in #1469).
- I'm not interested in adding more colors for different states of "cannot buy it".
- Fix a small bug: If the player upgrades the cores via a script and presses the "buy core" button fast enough, they can upgrade the cores over the limit.

Normal:
normal

Not enough money:
not-enough-money

Max:
max


d0sboots:
Thanks for catching this one.

PR #1888, opened 2024-12-30
DOCUMENTATION: Add starter react documentation

danielpza:
React (and TypeScript) support was added in v2.7.0 but there seems to be little documentation aside from the Changelog.

Added some rudimentary documentation for getting started with React in Bitburner to increase discoverability. Feel free to expand or make any suggestions.

image


d0sboots:
A good start!

PR #1891, opened 2025-01-03
MISC: Dockerized Bitburner

romaswe:
Added a dockerfile and a compose file to be able to run Bitburner easily in docker.
This to have more options to host and run the application, and also give the option to publish Bitburner to [docker hub](https://hub.docker.com/).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,45 @@
+# Stage 1: Build the application
+FROM node:lts-alpine AS builder
+
+# Set working directory
+WORKDIR /app
+
+# Install necessary tools, including git
+RUN apk add --no-cache git
+# Copy dependencies
+COPY package.json package-lock.json ./
+COPY tools/ ./tools
+
+# Install dependencies
+RUN npm ci
+
+# Copy source code
+COPY . .
+
+# Build the application
+# Pass the build mode as a build argument (default to production)
+ARG BUILD_MODE=production
+RUN npx webpack --mode ${BUILD_MODE}
+
+# Prepare the build artifacts
+RUN mkdir -p .app && \
+    cp index.html favicon.ico .app && \
+    cp -r dist .app
+
+# Stage 2: Serve the application using Nginx
+FROM nginx:stable-alpine AS runtime
+
+# Security optimizations
+RUN chmod -R 644 /etc/nginx/nginx.conf
+
+# Copy the built application from the builder stage
+COPY --from=builder /app/.app /usr/share/nginx/html
+
+# Set permissions for security
+RUN chmod -R 755 /usr/share/nginx/html
+
+# Expose port
+EXPOSE 80
+
+# Start Nginx
+CMD ["nginx", "-g", "daemon off;"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,10 @@ FROM node:lts-alpine AS builder
 WORKDIR /app
 
 # Install necessary tools, including git
-RUN apk add --no-cache git
+RUN apk update && \
+    apk add --no-cache git && \
+    apk add bash
+
 # Copy dependencies
 COPY package.json package-lock.json ./
 COPY tools/ ./tools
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -0,0 +1,45 @@
...
+# Install necessary tools, including git
+RUN apk add --no-cache git
+# Copy dependencies
+COPY package.json package-lock.json ./
+COPY tools/ ./tools
+
+# Install dependencies
+RUN npm ci
+
+# Copy source code
```
d0sboots, line 19: This feels very weird. Why are you copying everything into a separate directory? The build process takes care of having a separate output dir already.

romaswe: That copy on line 17 is to copy the code in to the docker container so we can build it.
I'm copying the code in to the container first so we can build it in the container so we don't need to have separate build steps and tools outside the container.


What we can do is to use the build script on line 22-27.
But it's not what I would prefer because then we both need to install bash in the container and I like it when you see in the dockerfile what's happening, for example if we change the output dir in build.sh the dockerfile will stop working, and it may be hard to understand why. 

But if others prefer that it uses build.sh it's a quick fix

catloversg: IMO, we should copy the entire repository (don't cherry-pick files/folders), then run `npm ci` and `npm run build` (or `npm run build:dev`, based on BUILD_MODE). Building steps are defined in build scripts. The Dockerfile should be "dumb" (no knowledge about the building process).

romaswe: I can agree with using the build from npm e.g. use `npm run build` 

The reason I separated to first just copy dependencies is to reduce the build time (for feature builds)
Because docker is caching each layer, so if we are copying everything it will result in that it will run `npm ci` every time you build the docker image, but if we do it this way we don't need to do `npm ci` in case we don't change the tools or package.json, which will make the build faster when you want to update the image and only have made changes to the codebase and not the packages.

Dockerfile
```diff
@@ -5,7 +5,10 @@ FROM node:lts-alpine AS builder
 WORKDIR /app
 
 # Install necessary tools, including git
-RUN apk add --no-cache git
+RUN apk update && \
+    apk add --no-cache git && \
+    apk add bash
+
 # Copy dependencies
 COPY package.json package-lock.json ./
```
d0sboots, line 13: Isn't this redundant with `COPY . .`?

Dockerfile
d0sboots: I think you should explain more about what you expect this to be used for. Honestly, I don't really see a use for it - you can already pretty easily build+host a branch through GitHub via a workflow that's already part of the repo. And this seems very cumbersome for development. But there may be a usecase I'm missing.

Dockerfile
danielpza: It's best practice to install dependencies and then copy the rest of the code. The dependencies change less often so these layers don't need to be rebuilt every time the code changes.

Dockerfile
romaswe: Well, the reason I did it was because I wanted to show a gamified way to learn coding to a younger crowd.
And didn't have access to internet where I was going to be, and I don't want to install all the dependencies local on my laptop just to demo something, so with docker you only need docker to build and run it.

And because I already did it for myself, I made a pr in case others also wants it.
But some of the benefits of it are.
You don't need to install all dependencies to build and run the application. And it makes it easier to host at other providers, and you're not "locked" to GitHub's solution. EG self-host or other providers.

Dockerfile
romaswe: Yea exactly, like I said on the other comment:

> I can agree with using the build from npm e.g. use `npm run build`
> 
> The reason I separated to first just copy dependencies is to reduce the build time (for feature builds) Because docker is caching each layer, so if we are copying everything it will result in that it will run `npm ci` every time you build the docker image, but if we do it this way we don't need to do `npm ci` in case we don't change the tools or package.json, which will make the build faster when you want to update the image and only have made changes to the codebase and not the packages.


I split it so we utilize dockers caching of layers so it don't need to rebuild all layers every time 😄

Dockerfile
d0sboots: > And didn't have access to internet where I was going to be, and I don't want to install all the dependencies local on my laptop just to demo something

I feel like I'm missing something critical here. If you don't have access to the Internet, you won't be able to access a container hosted elsewhere? And if it's a local container, that means you did install all the dependencies on your laptop (inside the image).

Dockerfile
d0sboots: I gotcha. I knew literally nothing about docker before this, but this seems like a good trade-off of "reuse layered effort" vs "don't duplicate the build system"

Dockerfile
romaswe: Well, yes and no, yes all dependencies need to be installed in the image during the build, but when it's installed I can just ship the image to my laptop that only contains the hosted version of the application, because it's using a multi-stage build, so we only ship a Nginx image that hosts Bitburner and not the whole application.

d0sboots: You don't need a nginx server for that, the final built webpage is sufficient (or the electron build if you want a standalone app instead of a webpage). This is in fact the output of the build process.

PR #1892, opened 2025-01-03
CODINGCONTRACT: Add support for other answer formats

G4mingJon4s:
Coding contracts are inherently not type-safe. This PR tries to combat this a bit, by making these changes:
- [X] Make `attempt` accept not only strings, but also the specific type the contract works with, like `number[]` or more specific formats
- [x] Make all contract types and their data and answer types accessible to the player
- [x] Add optional typing to `getData` and `attempt`


G4mingJon4s:
I spent about 1 hour now, testing "most" contracts for backwards-compatibility with my own cct script. I would really appreciate others testing their scripts too, since mine doesn't cover all contracts. These are the contracts, I would say, that are most affected by the string conversion and taking the format directly:
- Merge Overlapping Intervals
- Proper 2-Coloring of a Graph
- Shortest Path in a Grid
- Square Root

G4mingJon4s:
I'm working on making the types available to the player right now. In the definitions, it is suggested that the data is provided similar to how the `SleeveTask`s are structured.
I'm not quite sure what approach to take. I could make both by having the data type be computed based on the general types. For more detail, this is what I would settle on at the moment:
```ts
declare enum CodingContractName {
  FindLargestPrimeFactor = "Find Largest Prime Factor",
  ...
}

// Each contract has its data and answer types specified here. The player can use this however they like.
export type CodingContractSchema = {
  "Find Largest Prime Factor": [number, number],
  "Shortest Path in a Grid": [number[][], string],
  "Square Root": [bigint, string],
  [CodingContractName.FindLargestPrimeFactor]: [number, number] // Maybe this is safer?
}

// This would be the new way the contract data is returned, allowing for type-narrowing based on "type".
export type CodingContractData = { [T in keyof CodingContractSchema]: {
  type: `${T}`,
  data: CodingContractSchema[T][0]
} }[keyof CodingContractSchema];
/* This results in this type signature:
type CodingContractData =
  | { type: "Find Largest Prime Factor"; data: number; }
  | { type: "Shortest Path in a Grid"; data: number[][]; }
  | { type: "Square Root"; data: bigint; }
*/
```
The new way of returning the contract data is a API break, so maybe the optional typing is preferred? (see [this playground example](https://tsplay.dev/N5KE0W)).
It allows for strict types even in JS, so it would be worth it IMO.

`attempt` would still be typed as `any` this way. Not sure if it makes sense to give the submission the same treatment. Having the user return in the same shape `{ type: string; data: any; }`, but typed would be possible.

There are quite a few options here:
- Leave everything as is, just give the player all of the types for each contract. They can figure it out on their own. This would be for TS users only (maybe JSDoc works too?). `getData` and `attempt` are still typed as `any`.
- Make optional types by explicitly setting a type generic on `getData` and `attempt`. TS users only.
- Make `getData` similar to `SleeveTask`, allowing for type-narrowing based on a property (as shown above). Every player would benefit from this, but it would be an API break.
- Make `attempt` similar to `SleeveTask`, making it fully type-safe. The player would have to make the shape on their own though. For newer players especially, this could be quite difficult?
- Merge option 3 and 4, making both `getData` and `attempt` fully type-safe.

This is something I really don't want to decide on my own here. IMO it isn't worth it for `attempt`, but `getData` would be a lot safer to use with option 3. For anyone who wants more type-safety, they have all types at their disposal through the definitions.

src/data/codingcontracttypes.ts
```diff
@@ -30,15 +30,27 @@ interface CodingContractType<Data, State = Data> {
   name: string;
   /** How many tries you get. Defaults to 10. */
   numTries?: number;
-  /** Function that checks if the provided solution is correct. */
-  solver: (state: State, answer: string) => boolean;
+  /** Function that computes the correct answer. */
+  solver: (state: State) => Solution;
```
d0sboots: So, I think this is a distinct step backwards in terms of the CCT internal API. The old (existing) API just checks if the provided answer is correct - it essentially functions like `isCorrectAnswer`. It is an internal implementation detail of any given CCT if that requires making a "solution" or not. There is no reason IMO to promote that internal detail into part of the generic machinery of the CCT class, and it just makes the typing much more complicated.
At first I thought this was necessary for your convert/validate stuff, but now I don't think it is. You can do that generically and still leave `solver` with the signature of `(state: State, answer: Answer) => boolean`.

d0sboots (commented):
I don't understand why you want to change the API of `getData`. The type of contracts is *already* exposed via `ns.codingcontract.getType()`; the function is literally named for this purpose. What gain is there by additionally jamming it into `getData`?

Edit: I could see there being a slightly better alternate world where `ns.codingcontract.getContract()` returned a single object with all the data pertinent to a CCT, and we didn't have multiple API functions to do the different jobs, but that's not the world we are in.

src/data/codingcontracttypes.ts
G4mingJon4s: I can understand your argument here. In the moment it felt right, since the first few solvers are creating the solution first and then check if the players answer is correct. Further down the line, that is no longer the case. I will move stuff to that signature!

G4mingJon4s:
Changing the way the data is provided to the user can give type-safety, even to non-TS users. Depending on the `type` property, the way `data` is typed changes. This results in a typing, similar to [SleeveTask](https://github.com/bitburner-official/bitburner-src/blob/1cb2a83b4fab11c3644e49467ca1798f38da2582/src/ScriptEditor/NetscriptDefinitions.d.ts#L1106) or [PlayerRequirement](https://github.com/bitburner-official/bitburner-src/blob/1cb2a83b4fab11c3644e49467ca1798f38da2582/src/ScriptEditor/NetscriptDefinitions.d.ts#L9897).
The player doesn't need to do any "TS magic" or type-assertions. They just check which type the contract has and get the correct data type for free.

I agree, it would make more sense for that function to be called `getContract` or similar. Since it was suggested to change the `CodingContractData` type in the definitions file, I just went with that.
Maybe we just keep the existing functions and add this? That way, there's no API break and players who want type-safety can easily achieve it. It would be a bit weird to have multiple functions that do essentially the same, but that's already the case for a lot of other parts of NS.
Still, I would like to make the old functions more type-safe too. There are some small things that bother me with the current API:
- `getData` and `attempt` being inherently typed in an unsafe way
- `getType` returning `string`, instead of an union of possible types
- No available types for any of the contracts

That's why I'd also like `getData` and `attempt` to be optionally typed like I described before. Players don't have to use it, so it shouldn't ever be a problem.

d0sboots:
I'm all for `getType` returning an enum member. That is backward-compatible and improves type-safety in a simple and straightforward way. It exposes the full list of Coding Contracts, but honestly I think that is a good thing.

`getData` can't be changed. That is too big of a breaking change for something that is a marginal improvement, at best. We've rejected much bigger improvements than that due to backward compatibility, and there are *lots* of people out there with big libraries of CCT solvers.

I think I'd be open to adding a `getContract` method that returns all the info in a single object. That can then be typesafe. It should have a RAM cost of at least `getType` + `getData` to be balanced.

I don't think `attempt` can ever be really typesafe, because its true type always has to be `unknown`. Also, it will always need to accept strings for backward compatibility. I think it is a good ergonomic change to add the ability to accept the "native" type in answers, when that is sensible, but IMO we shouldn't tie ourselves in knots with the type system to make that happen.

G4mingJon4s:
Since all of the contracts types are in the definitions now, the internals can be somewhat type-safe too. At least the solver definitions.
This is pretty much ready to go. It still bothers me that ccts don't have their own folder and aren't separated in any way based on some categories. Since I already changed around the internals quite a bit, I would be more than happy to clean up this mess in a future PR.

G4mingJon4s:
Someone should really test their cct script on this. In theory, it should be 100% backwards-compatible.
I went through every contract once and made sure the conversion logic is correct for properly-formatted strings. Not sure how it handles `JSON.stringify` submissions. Since converted answers get validated too, the validation part should also be working.

G4mingJon4s:
I tested [this script](https://github.com/alainbryden/bitburner-scripts/blob/main/Tasks/contractor.js) on the current 2.7.1dev and this PR and it works on both versions without any errors. Well, except for contract filenames allowing for duplicates, which is an unrelated bug. I will fix that in a separate PR when this is merged.
I tested both converting every answer to strings via `JSON.stringify` and returning the format itself without any conversion. Both work without any errors, except one edge-cases involving passing an empty array as a string.

src/NetscriptFunctions/CodingContract.ts
```diff
@@ -18,11 +19,20 @@ export function NetscriptCodingContract(): InternalAPI<ICodingContract> {
   };
 
   return {
-    attempt: (ctx) => (answer, _filename, _hostname?) => {
+    attempt: (ctx) => (answer: unknown, _filename: unknown, _hostname?: unknown, _type?: unknown) => {
```
d0sboots: You don't have to explicitly type these, all parameters are automatically unknown through a process I don't fully understand/recall.

src/NetscriptFunctions/CodingContract.ts
```diff
@@ -59,11 +69,31 @@ export function NetscriptCodingContract(): InternalAPI<ICodingContract> {
...
+        return structuredClone(contract.getData());
+      },
+    getContract: (ctx) => (_filename: unknown, _hostname?: unknown) => {
       const filename = helpers.string(ctx, "filename", _filename);
       const hostname = _hostname ? helpers.string(ctx, "hostname", _hostname) : ctx.workerScript.hostname;
       const contract = getCodingContract(ctx, hostname, filename);
-      return structuredClone(contract.getData());
+      // asserting type here is required, since it is not feasible to properly type getData
+      return {
+        type: contract.type,
```
d0sboots, line 86: What about the other attributes? I would expect description, tires remaining, etc. to be in here.

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -3860,23 +3853,35 @@ export interface CodingContract {
...
+   * There is an optional fourth parameter where you can assert the contract type.
+   * This will type the answer format you need to provide.
+   * Be aware that this function throws if the specified type does not match the type of the contract with that filename.
+   *
    * @example
    * ```js
-   * const reward = ns.codingcontract.attempt(yourSolution, filename, hostname);
+   * const reward = ns.codingcontract.attempt("solution as a string", filename, hostname);
+   * // or
+   * const reward = ns.codingcontract.attempt(["answer", "in", "the", "correct", "format"], filename, hostname);
```
```suggestion
   * const reward = ns.codingcontract.attempt(["answer", "as", "an", "array"], filename, hostname);
```
"Correct format" is a bit too heavy of a value judgement imo

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -3860,23 +3853,35 @@ export interface CodingContract {
...
    * Attempts to solve the Coding Contract with the provided solution.
    *
+   * There is an optional fourth parameter where you can assert the contract type.
+   * This will type the answer format you need to provide.
+   * Be aware that this function throws if the specified type does not match the type of the contract with that filename.
+   *
    * @example
    * ```js
-   * const reward = ns.codingcontract.attempt(yourSolution, filename, hostname);
+   * const reward = ns.codingcontract.attempt("solution as a string", filename, hostname);
```
```suggestion
   * const reward = ns.codingcontract.attempt("[solution, as, a, string]", filename, hostname);
```

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -3860,23 +3853,35 @@ export interface CodingContract {
...
-   * @param answer - Attempted solution for the contract.
+   * @param answer - Attempted solution for the contract. This can be a string formatted like submitting manually, or the answer in the format of the specific contract type.
    * @param filename - Filename of the contract.
    * @param host - Hostname of the server containing the contract. Optional. Defaults to current server if not
    *   provided.
    * @returns A reward description string on success, or an empty string on failure.
    */
-  attempt(answer: string | number | any[], filename: string, host?: string): string;
+  attempt(answer: any, filename: string, host?: string): string;
+  attempt<T extends CodingContractName | `${CodingContractName}`>(
```
d0sboots: I'm not a big fan of having two versions of the function, but I think trying to collapse their signatures would be even worse. 

Note that the TSDoc here is only applying to the first version; the second one is undocumented. It pains me to duplicate all the documentation for the other version, but I think you're gonna have to. The docs about the forth parameter should only go on the second override ofc. Also, you should mention that it's only useful for people writing native Typescript code. 

Honestly, the whole thing leaves a very sour taste in my mouth, because it feels awkward and shoehorned in just for TS, and even then, not really quite the right design. The most "natural" design, where all type info would flow automatically and clearly, would be `getContract` returning a class that had `attempt` on it. Unfortunately, the way the ns object works makes this tricky, although not impossible.

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -3914,11 +3919,44 @@ export interface CodingContract {
...
+   * This will type the data returned for that specific contract type.
+   * Be aware that this function will throw if the type in the fourth parameter is not the same as the type of the contract with that filename.
+   *
    * @param filename - Filename of the contract.
    * @param host - Host of the server containing the contract. Optional. Defaults to current server if not provided.
    * @returns The specified contract’s data, data type depends on contract type.
    */
-  getData(filename: string, host?: string): CodingContractData;
+  getData(filename: string, host?: string): any;
+  getData<T extends CodingContractName | `${CodingContractName}`>(
```
d0sboots: I think the extra overhead of the duplicate signature and duplicate docs is just not worth it here. If you want TS typesafety, use `getContract`.

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -3914,11 +3919,44 @@ export interface CodingContract {
...
+   *   const data = contract.data;
+   *   // ^? data: number
+   * }
+   * ```
+   *
+   * @param filename - Filename of the contract.
+   * @param host - Host of the server containing the contract. Optional. Default to the current server if not provided.
+   * @returns An object containing both the type and data, typed depending on the type.
+   */
+  getContract(filename: string, host?: string): CodingContractSet;
```
```suggestion
  getContract(filename: string, host?: string): CodingContractObj;
```
If the name "CodingContract" is already taken and can't be used, then tack on something generic. "Set" had collection connotations.

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -8279,6 +8317,82 @@ declare enum CompanyName {
...
+  [CodingContractName.Proper2ColoringOfAGraph]: [[number, [number, number][]], (1 | 0)[]];
+  [CodingContractName.CompressionIRLECompression]: [string, string];
+  [CodingContractName.CompressionIILZDecompression]: [string, string];
+  [CodingContractName.CompressionIIILZCompression]: [string, string];
+  [CodingContractName.EncryptionICaesarCipher]: [[string, number], string];
+  [CodingContractName.EncryptionIIVigenereCipher]: [[string, string], string];
+  [CodingContractName.SquareRoot]: [bigint, bigint, [string, string]];
+};
+
+export type CodingContractData<T extends string> = T extends `${keyof CodingContractSignatures}`
```
d0sboots, line 8363: Why can't these directly be limited to extending the keyof, instead of having a ternary?

d0sboots (commented):
The internal typesafety changes are great. I've only got one (somewhat significant) concern about the API

src/NetscriptFunctions/CodingContract.ts
G4mingJon4s: Leaving out the explicit types actually results in an error. I'm not sure either, why this is the case. For `getContract`, I can remove them safely though.

src/ScriptEditor/NetscriptDefinitions.d.ts
G4mingJon4s: It is possible to collapse the signature. It's not that clean, but it could work.
I don't think copying the documentation would be a good idea. The documentation should actually work on all overloads here, just not when filling in the arguments, which is pretty important.
A function on `getContract` would be a good option, since it avoids all of this. We would run in the same situation as with the port handles. Since the function can only be used up to the maximum tries for any given contract, it should be fine though? Handling the script death would still be a weird quirk.

From what I can tell, it's either merging the signatures or a function on `getContract`. I'm fine with both suggestions tbh.
If we choose to add a callback there, I would tend to remove the typings on `attempt` and `getData` at the same time. I don't like adding it on one, but not the other.
The new signature would be something like this: `attempt<T extends CodingContractName | ´${CodingContractName}´ | "" = "">(answer: CodingContractAnswer<T>, filename: string, host?: string, type?: T): string;`. It's not that bad IMO, but I could see how it confuses some players.

src/ScriptEditor/NetscriptDefinitions.d.ts
G4mingJon4s: I answered this above in the discussion about the `attempt` signature.

src/ScriptEditor/NetscriptDefinitions.d.ts
G4mingJon4s: This was a remnant of having no overload, but a single signature. If we merge the signatures like discussed above, this is needed, otherwise directly restricting the generic makes sense.

G4mingJon4s:
I went with the callback on the contract object:
The function is typed based on the type property, similar to the data property, so the overloads and optional type parameters aren't needed anymore.

All initial goals have been met and the cct internals are now type safe. I would say this is ready to go now.

d0sboots:
Sorry for the radio silence. Fortunately, you proactively took the route I was going to suggest. :)

src/NetscriptFunctions/CodingContract.ts
```diff
@@ -65,8 +73,26 @@ export function NetscriptCodingContract(): InternalAPI<ICodingContract> {
...
+        data: contract.getData(),
+        attempt: (answer: unknown) => {
+          helpers.checkEnvFlags(ctx);
+          return attemptContract(ctx, server, contract, answer);
+        },
+        description: contract.getDescription(),
+        numTriesRemaining: contract.getMaxNumTries() - contract.tries,
```
d0sboots: This needs to be a method, not a data field. And thus, it should checkEnvFlags, as well.

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -3918,7 +3913,31 @@ export interface CodingContract {
    * @param host - Host of the server containing the contract. Optional. Defaults to current server if not provided.
    * @returns The specified contract’s data, data type depends on contract type.
    */
-  getData(filename: string, host?: string): CodingContractData;
+  getData(filename: string, host?: string): any;
+
+  /**
+   * Get the type and data of the contract.
```
d0sboots: There's a lot more to this now, it returns a class-like object with methods.

d0sboots (commented):
Looks great! Just the couple of quibbles.

src/Netscript/RamCostGenerator.ts
```diff
@@ -333,6 +333,7 @@ const codingcontract = {
   attempt: RamCostConstants.CodingContractBase,
   getContractType: RamCostConstants.CodingContractBase / 2,
   getData: RamCostConstants.CodingContractBase / 2,
+  getContract: RamCostConstants.CodingContractBase,
```
d0sboots: This should be in the range `* 1.5` to `* 2` because it includes `attempt` now.

src/NetscriptFunctions/CodingContract.ts
```diff
@@ -65,8 +73,26 @@ export function NetscriptCodingContract(): InternalAPI<ICodingContract> {
...
+    getContract: (ctx) => (_filename, _hostname?) => {
+      const filename = helpers.string(ctx, "filename", _filename);
+      const hostname = _hostname ? helpers.string(ctx, "hostname", _hostname) : ctx.workerScript.hostname;
+      const server = helpers.getServer(ctx, hostname);
+      const contract = getCodingContract(ctx, hostname, filename);
+      // asserting type here is required, since it is not feasible to properly type getData
+      return {
+        type: contract.type,
+        data: contract.getData(),
+        attempt: (answer: unknown) => {
```
d0sboots: `attempt` is going to flag the static-ram detector. Not entirely sure what to do about that. Maybe `makeAttempt`?

d0sboots: I thought of another possible answer here. Instead of calling `helpers.checkEnvFlags`, you could literally call through `ctx` to the netscript attempt method, thus invoking its dynamic ram cost. This would be a little strange, but at the same time make a certain kind of sense, and it would be compatible with how the static RAM checker works (the name would stay the same).
In this case, you wouldn't want to charge more for `getContract`. Also, you'd want to apply the same logic for `numTriesRemaining` (rename to `getNumTriesRemaining()`).

G4mingJon4s: I'm not quite sure that would be a good idea? I like that it's kinda an alias to the same function.
Relying on the static RAM checker to catch that call is a quite the gamble IMO.

My current plan was to rename the `attempt` method to `submit`. The RAM remarks in the JSDoc would simply state what it costs to use that function. Any methods on the object are included in that number. When the methods themselves invoke the static RAM cost, it could be a bit unintuitive to some players, because they think the static RAM from the `getContract` function includes everything.

d0sboots: Before I get to specific responses, I'll say that I'm not sure it would be a good idea either. :D It was an alternative idea, with different pros and cons; going either way would be fine with me.

> I like that it's kinda an alias to the same function.

What I was proposing *would* make it an alias to the same function; `ns.getContract().attempt` would essentially become an alias for `ns.codingcontract.attempt`. I can't tell if you're saying that's a good thing about the proposal, or you consider what you're doing right now to be an alias in a different way.

> Relying on the static RAM checker to catch that call is a quite the gamble IMO.

We never *rely* on the static RAM checker, because it can't be relied on - you can always get around it by using `ns["call"]` or `eval` or other tricks. However, as long as you are calling things in a "conventional" way, any call to `anything.attempt()` will be registered by the static RAM checker as a call against `cct.attempt`, which is what makes this reasonable at all.

The dynamic RAM check is the thing we actually rely on, and that would be covered by both code paths funneling through the NS method (so even if you called it both ways, you'd only get charged once).

----

That said, I agree that it could be unintuitive; certainly the most straightforward way is to include all the cost up-front, and just rename methods.

Issue #1747, opened 2024-11-02
EDITOR: Cannot resolve import paths in some cases


catloversg:
TypeScript support will be available on the stable version via #1216. In some cases, monaco cannot resolve the import paths. This is just a problem with the script editor UI. It does not affect anything else. It may be fixed by #1477, but I don't think that fix can be merged easily.

MRE:
lib.ts
```js
export const multiply = (a: number, b: number) => a * b;
```

test.ts
```js
import { multiply } from "/lib";

export async function main(ns: NS) {
  ns.print(`multiply: ${multiply(2, 3)}`);
}
```

Screenshot:
![Capture](https://github.com/user-attachments/assets/982395a3-1c20-43b9-ae1a-3dd1cfcff048)

With this case (use an absolute path), the player can work around by using a relative path (`./lib` instead of `/lib`).

I'll use this issue to keep track of the problem and potential fixes. Other reports can be marked "duplicate" and linked back to here.

Test suite:
[test suite for editor.gz](https://github.com/user-attachments/files/18240475/test.suite.for.editor.gz)

d0sboots:
Yes, we should assume #1477 is dead in the water and steal pieces from it that are useful or mergeable.

lucebac:
I've seen this error in the past as well when I updated the import resolution. However, I believe I wasn't able to fix this because from what I understand this is an edge case of the way we interact with and provide model data to monaco.

catloversg:
Update:
The workaround (using a relative path) still has some problems. In #1661, we added `loadAllServerScripts` function to preload models. The new function has some problems. I tried to fix them in #1865 and #1867. However, a player reported another problem in #1877. If we preload too many models, monaco fails to provide type hints via JSDoc. It still works in TS files, though. I'm too lazy to debug the internal of monaco to see why it happens. Based on https://github.com/microsoft/monaco-editor/issues/3895, I conclude that monaco is not supposed to support loading hundreds of models at the same time, so the "preloading all scripts" solution is only a workaround until we find a proper way to load modules on demand. I'll take a look at the idea in #1477.

catloversg:
Update:
The workaround works properly now with the fix in #1893.

PR #1893, opened 2025-01-05
MISC: Dynamically load imported scripts for editor

### LINKED ISSUE #1877: Autocompletion does not work in the editor anymore
Since TS files are allowed, the auto completion does not work for js files anymore in my savegame.
The tooltip on {NS} also does not work as before anymore.
It does however work in a new game.

After some more testing: It works if I stay in the editor for a long time. The loading time is much longer now.

- [x] Save: 
[bitburnerSave_1734918884_BN12x153.json.gz](https://github.com/user-attachments/files/18224213/bitburnerSave_1734918884_BN12x153.json.gz)

- [x] Minimal scripts to reproduce the issue
```
/** @param {NS} ns */
	ns. // auto completion is not triggered here
}
```
- [x] Steps to reproduce: Open the editor testA.js and try to get auto completion for any method of NS
- [x] v2.7.0, v2.7.1dev

catloversg:
For context, please check my comment at https://github.com/bitburner-official/bitburner-src/issues/1747#issuecomment-2558912702.

This PR fixes that problem by dynamically loading only imported scripts instead of loading all scripts.

Note: This PR conflicts with #1865 and #1866, so I ported those fixes to this PR.

How to test:
Test 1:
- Load this save file: [test suite for editor.gz](https://github.com/user-attachments/files/18310863/test.suite.for.editor.gz)
- Open all scripts, especially `ts.ts` and `tsx.tsx`. False-positive error markers should disappear after 2 seconds.

Test 2:
- Load the save file in #1877.
- Make sure that autocomplete works normally.

Closes #1877.

### DIFF
--- a/src/ScriptEditor/ui/ScriptEditorRoot.tsx
+++ b/src/ScriptEditor/ui/ScriptEditorRoot.tsx
@@ -3,9 +3,12 @@ import type { ContentFilePath } from "../../Paths/ContentFile";
 import React, { useEffect, useRef } from "react";
 import * as monaco from "monaco-editor";
 
-import { Editor } from "./Editor";
+import type * as acorn from "acorn";
+import * as walk from "acorn-walk";
+import { extendAcornWalkForTypeScriptNodes } from "../../ThirdParty/acorn-typescript-walk";
+import { extend as extendAcornWalkForJsxNodes } from "acorn-jsx-walk";
 
-type IStandaloneCodeEditor = monaco.editor.IStandaloneCodeEditor;
+import { Editor } from "./Editor";
 
 import { Router } from "../../ui/GameRoot";
 import { Page } from "../../ui/Router";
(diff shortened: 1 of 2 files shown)


src/ScriptEditor/ui/ScriptEditorRoot.tsx
```diff
@@ -62,6 +74,91 @@ function Root(props: IProps): React.ReactElement {
     (editorElement as HTMLElement).style.outline = "none";
   }, [editorRef]);
 
+  /**
+   * The TypeScript compiler needs time to perform type-checking, so in some edge cases, the editor shows the 2792 error
```
d0sboots, line 78: Thank you for this comment. The whole thing makes me :sob:, but such is Monaco.

src/ScriptEditor/ui/ScriptEditorRoot.tsx
```diff
@@ -62,6 +74,91 @@ function Root(props: IProps): React.ReactElement {
...
+    if (!currentScript || !editorRef.current) {
+      return;
+    }
+    const markers = monaco.editor.getModelMarkers({
+      resource: currentScript.model.uri,
+    });
+    let needToReloadModel = false;
+    for (const marker of markers) {
+      // 2792: "Cannot find module" error
+      if (marker.code !== "2792") {
```
d0sboots, line 106: If a module legitimately can't be found, will this reload the module every 2 seconds? Is that a problem?

catloversg: If a module does not exist, `makeModelsForImports` will throw an error at `walk.simple` before calling `reloadModelOfCurrentScript`. It looks like this:

Capture

Even in the case that there is a valid error marker (not import errors), `reloadModelOfCurrentScript` is only called when the player edits the code. I don't see any way that it is called every 2 seconds, unless the player constantly changes the code, which is a valid case that `reloadModelOfCurrentScript` needs to be called.

PR #1895, opened 2025-01-07
BUGFIX: "Import Save Comparison" popup is shown after reloading

### LINKED ISSUE #1890: "Import Save Comparison" window has unclear options
For Bitburner v2.7.0 (fbe1546fd), if you go to the Options window, select "Import Save", and choose an older save file, then it will present you with options to "Confirm" or "Compare Save".  If you choose "Confirm", then it will load that game, the game will start, and then after a slight pause, an "Import Save Comparison" window will open anyways, comparing the data just imported with the latest save, giving two options "Take me back!" and "Proceed with Import".  This feels like it was doing "Compare Save", despite you clicking "Confirm".

Also, in that "Import Save Comparison" window it's unclear if "Take me back!" takes you back to the save you just imported or back to the more recent save.  It's equally unclear if "Proceed with Import" lets you proceed with the current import or to proceed with importing the more recent save.

I frequently get this backwards, since I assume "Proceed with Import" will let me proceed with the save I just imported.

It would probably be clearer to have the options of "Continue with current game" and "Reload newer save", or something like that.

Additionally, I'm not sure if showing the comparison, even when the player intentionally skipped comparing, makes sense.

catloversg:
How to reproduce this bug:
- Create a "clean" state:
  - Open the Electron app. Open the save folder (Files -> Open Directory -> Open Saves Directory).
  - Delete the current save (Options -> Delete Save). Save game. Close the Electron app.
  - Open the Electron app again. Delete old saves in the saves folder.
- Import BIG (1q-edited.json.gz). Save game. Close the Electron app.[1]
- You will see a folder named "1ce7b2f80ebfd3" with a save file.
- Open the Electron app.
- Import SMALL (1k-edited.json.gz). [2]
- After the game reloads, the "Import Save Comparison" is shown. In the save folder, there is folder name "dc70827649567", but it's empty.

After [1], BIG is saved on disk in "1ce7b2f80ebfd3". This is the "newest save file". After [2], SMALL has not been saved on disk ("dc70827649567" is empty), so when the game reloads, it sees BIG (in "1ce7b2f80ebfd3") as the newest save file.

We have a built-in mechanic to mitigate the problem of "comparison popup should not be shown after importing". For details, please check:
- `pushImportResult` and `pushDisableRestore` in `src\Electron.tsx`
- `setStopProcessHandler` in `electron\main.js`

The gist is that we disable that popup for a brief period of time after importing. We need to notify the Electron code by calling `pushImportResult` after calling `saveObject.importGame`. We need to do that in:
- `confirmedImportGame` in `src\GameOptions\ui\GameOptionsSidebar.tsx` -> We forgot to do that here.
- `handleImport` in `src\ui\React\ImportSave\ImportSave.tsx`

This PR fixes this bug by calling `pushImportResult` in `saveObject.importGame`. With this change, we never have to worry about forgetting to call `pushImportResult`.

@d0sboots I think our mitigation is not really good. The root cause of the problem is that we don't save the new save data to disk (and cloud) after importing. If the player closes the game after importing SMALL without pressing the "save button" (and before the autosave triggged), the game won't "flush" the new save data to disk. When they open the game again, the popup is still shown. Reproducible steps:
- Do the steps above.
- After importing SMALL and the popup is shown, just close the game.
- Open the game again. The popup is shown again.

I did not implement a proper fix in this PR to keep the change as small as possible (reduce the risk of a regression bug). If you want, I'll fix the root cause properly in another PR.

I also add some tooltips to clarify the meaning of 2 buttons mentioned in #1890.

Save files:
BIG: [1q-edited.json.gz](https://github.com/user-attachments/files/18327570/1q-edited.json.gz)
SMALL: [1k-edited.json.gz](https://github.com/user-attachments/files/18327571/1k-edited.json.gz)

Before:
before

After:
after1
after2

Fixes #1890.

### DIFF
--- a/src/Electron.tsx
+++ b/src/Electron.tsx
@@ -182,7 +182,7 @@ function initElectronBridge(): void {
   });
   bridge.receive("push-save-request", (params: unknown) => {
     if (typeof params !== "object") throw new Error("Error trying to push save request");
-    const { save, automatic = false } = params as { save: string; automatic: boolean };
+    const { save, automatic = false } = params as { save: SaveData; automatic: boolean };
     window.appSaveFns.pushSaveData(save, automatic);
   });
   bridge.receive("trigger-save", () => {
--- a/src/SaveObject.ts
+++ b/src/SaveObject.ts
@@ -19,7 +19,7 @@ import { dialogBoxCreate } from "./ui/React/DialogBox";
 import { constructorsForReviver, Generic_toJSON, Generic_fromJSON, type IReviverValue } from "./utils/JSONReviver";
 import { save } from "./db";
 import { ToastVariant } from "@enums";
-import { pushGameSaved } from "./Electron";
+import { pushGameSaved, pushImportResult } from "./Electron";
 import { getGoSave, loadGo } from "./Go/SaveLoad";
 import { SaveData } from "./types";
 import { SaveDataError, canUseBinaryFormat, decodeSaveData, encodeJsonSaveString } from "./utils/SaveDataUtils";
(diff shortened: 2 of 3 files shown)


d0sboots:
I have noticed the larger issue you talked about, when importing games for testing and quickly closing before autosave kicks in. I agree that it should be fixed (as a separate thing).

PR #1896, opened 2025-01-07
UI: Fix migration doc typo

esainane:
`spc` becomes `scp`.


d0sboots:
> The last two ~~arguments~~ *letters* of spc have been reversed.

how did no one see the irony for all these years XD

PR #1897, opened 2025-01-07
UI: Remove RAM usage percentage in "free" CLI if it's NaN

catloversg:
It's just as the title said.

Before:
before

After:
after


d0sboots:
I kind of like the NaN%, but OK :D

PR #1905, opened 2025-01-07
BUGFIX: 'Connect' command tab completion does not list purchased servers

### LINKED ISSUE #1904: BUG: CLI Tab Completion Doesn't List Purchased Servers
After a bit of testing, I noticed that the `connect` CLI command does not include purchased servers in its tab completion, when not connected to `home`.  However, the command _will_ connect to purchased servers.

Repro steps:
1. Purchase a server
2. Connect to a non-purchased server (e.g. `connect n00dles`
3. Type `connect` and press <kbd>Tab</kbd>
  - Expected behavior: The server purchased in step 1 is listed as an option
  - Actual behavior: The server purchased in step 1 is NOT listed as an option
4. Manually enter the hostname of the server from step 1, and press <kbd>Enter</kbd>
5. `Connect` has connected to the server, despite it not showing in tab-complete

Version: Bitburner v2.7.1dev (d9dba2a44)

Nerdpie:
Fixes #1904 

Repro steps:
1. Purchase a server
2. Connect to a non-purchased server (e.g. `connect n00dles`
3. Type `connect` and press <kbd>Tab</kbd>
4. The hostname of the server from step 1 is shown in the tab completion list

Note: Format and lint have been run


Nerdpie:
... yeah, that tracks that I'd overlook an associated test case.

Let me know if this fix makes sense, and I can adjust the test to match.

Edit:  Upon closer inspection, the test is failing because the autocomplete code is now including `home`, even though that is the current server connected.

One possible fix would be to add
```ts
.filter(server => server !== currServ)
```
into the chain, or to directly add that condition to the existing filter, though it would push the length out a bit.

I'm not sure what approach would be most desirable.

d0sboots:
I'm really torn about this whole concept.

On one hand, yes, the player-purchased servers *are* valid targets to connect to, from anywhere on the network.

On the other hand, with the exception of home, almost none of them are ever useful targets to connect to, so adding them to completion would bloat the list of useful possibilities. And it does it for *every* server in the network, making it harder to read the useful servers.

d0sboots:
So, we've talked about this a lot on discord (mostly in relation to a related issue, the parity between `sing` API and the general ability to connect). I think my overall feeling is this:
* `singularity` and `connect` should have the same capabilities.
* `connect` and tab-completion do *not* necessarily need to be the same.

The reason for the latter is that tab-completion is a convenience for the user, to make their navigation more streamlined. In most cases, this argues for parity between what is allowed and what is exposed via tab. However, given the large number of purchased servers (all the purchased ones, plus all hashnet nodes), combined with the fact that you very rarely want to connect to them, I think it is best to leave tab completion as-is.

...or so I thought, until I captured this reference image from tab-completion on home:
image

Yes, it's big, but all the "irrelevant" nodes are at the end. If this property remains guaranteed, it is pretty easy to visually ignore them if you're using tab to view possible servers. And since they have a small number of prefixes ("d" for "darkweb", "h" for "hacknet" which also overlaps with "home", and one other for bought servers) they don't interfere with regular tab-completion much. This just goes to show: Always include action screenshots in your PR! :D

d0sboots:
As to fixes for the test: I think the correct thing is to add the "server !== currServ" to the condition, even though it will make it more bulky. (Extra `filter`s means extra passes through the list, which strikes me as inelegant.) We shouldn't offer home as an option on home.

That should make this both pass, and be good to merge.

Issue #1613, opened 2024-08-21
UI: Add Grafting to sidebar (or make gameplay changes)


Faenre:
Other mechanics have UI shortcuts in the sidebar, made available when you first interact with them. However, grafting does not have this, requiring the player to visit the correct city and office in order to reach the lab.

From an RP perspective, that's super cool. But it feels strange that it's only necessary if the player wishes to _begin_ the grafting process.

I'd like to either add Grafting to the UI menu (most likely), -or- require the player to stay in New Tokyo throughout the graft. The latter makes more RP sense given the entire context around it — most work mechanics are "remote-friendly" but grafting is not. But the former would be easier and less restrictive on players.

Thoughts?

d0sboots:
Definitely "no" to having to stay in New Tokyo for the whole of grafting. XD

IMO you can explain it lore-wise by "You have to fly to Neo Tokyo to find a dude, but then he flies around with you on your private jet operating while you're doing important CEO stuff" :D

I'm more neutral on easing the requirement of being in New Tokyo to start (which would be a prereq to putting in on the sidebar).

Faenre:
Say whaaaat, it'd be fun!

Joking aside, as for that middle point, that feels like a good reason to me for easing the requirement. Another part of me wants to suggest a way to play up the RP experience of it a bit more, have some weird requirement for the first entrance, similar to joining BB or accepting Stanek's. Maybe backdooring the company's server or something (hacking skill ~900, +/-). Would that too extreme? Hm. Maybe I'll tuck this idea away for a future bn...

As for the UI sidepanel, the following look interesting to me:

ChairAlt: (normal, sharp)
![image](https://github.com/user-attachments/assets/9daebea4-bb62-4314-896f-b13830585c05)
![image](https://github.com/user-attachments/assets/60575b19-6db6-45c2-995a-3e096868a93d)

AddRoad:
![image](https://github.com/user-attachments/assets/02ac512f-9f06-4c8b-85b2-c3ac52325e8f)

CallMerge:
![image](https://github.com/user-attachments/assets/9b50d236-dcef-4be9-bb11-d2ea4dd4b87b)

CallSplit:
![image](https://github.com/user-attachments/assets/244f80c2-b6d8-4703-b39c-c8fcfe2ab614)

Of these, I think ChairAlt and CallSplit look pretty great. What do you think?

d0sboots:
[...]

PR #1909, opened 2025-01-11
UI: Add grafting to sidebar with CallMerge

ivanrodgon:
Hello, this PR introduces the idea discussed in the issue https://github.com/bitburner-official/bitburner-src/issues/1613. I have tried to change the icon following the described instructions. I have only made changes to the UI.

Screenshot:
image


catloversg:
I did that in #1809, but with a different icon.

d0sboots:
I went with #1809.

d0sboots:
Worth noting is that #1809 did not change the requirement for being in neo Tokyo, which is unusual for a sidebar item. If people find that jarring, we can look at relaxing that.

Issue #1632, opened 2024-08-26
API: Feature Request: Provide "ns.${Api}.isEligible" functions for each restricted APIs.


bupjae:
Based on:
* https://github.com/bitburner-official/bitburner-src/pull/1598
* [Discord discussion](https://discord.com/channels/415207508303544321/459097632896188436/1277269132793876563)

Assume the user doesn't have SF4 / SF5 and currently not in BN4 / BN5 in this discussion.

For example, if the user wants to know whether they can use Stanek's API, the only way is run `ns.stanek.acceptGift()` and see return value.
However, running that causes the user accept gift, and suffers 10% all stat penalty, therefore it is not suitable for check eligibility of Stanek's API.

Almost all features in this game has similar issues; the user should actually start/join/create the feature to know whether the user is eligibile to access.

This feature request is to provide NS function `isEligible` for each API.

Sample `ns.stanek.isEligible` function:

```typescript
isEligible: () => () => {
    // You already accepted Stanek's gift. You can use stanek API of course.
    if (Player.hasAugmentation(AugmentationName.StaneksGift1, true)) return true;
    // You don't have SF13 and you'are not in BN13. You don't know anything about The Church.
    if (!Player.canAccessCotMG()) return false;
    // If you installed other augmentation before accepting gift, The Church will treat you filth and you can't use stanek API.
    // Otherwise, you can use stanek API, starting from acceptGift().
    return [...Player.augmentations, ...Player.queuedAugmentations].filter(
          (a) => a.name !== AugmentationName.NeuroFluxGovernor,
        ).length === 0;
}
```

d0sboots:
For many of the APIs, you can check simply by trying to call them. You'll get an exception on failure, that indicates you don't have access. 

However, I'm not against adding an explicit boolean check.

What you're asking for here (and what happened with corp) is slightly different: That's about checking if you can *gain* access to the relevant feature, as opposed to if you already have access.

ivanrodgon:
I’ll work on implementing isEligible functions for APIs as suggested. I’ll start with a sample for ns.stanek.isEligible and expand from there.

Thanks!

PR #1911, opened 2025-01-13
API: Add new feature "ns.${Api}.isEligible" functions for Bladeburner and Stanek APIs.

ivanrodgon:
# Pull Request Description  

## Summary  

This pull request introduces the `isEligible` function for the Stanek and Bladeburner APIs, aligning with the discussion in issue #1632. The goal is to improve API integrity by ensuring proper access validation for players interacting with these features.  

## Changes  

1. **Implemented `isEligible` Function**  
   - A new utility function, `isEligible`, has been created to determine if a player meets the requirements to access the Stanek and Bladeburner APIs.  

2. **Integrated Eligibility Checks**  
   - Added the `isEligible` validation to the corresponding methods in the Stanek API.  
   - Added the `isEligible` validation to the corresponding methods in the Bladeburner API.  

3. **Code Refactoring and Cleanup**  
   - Minor adjustments were made to streamline the implementation and improve readability in the affected files.  
   - Ensured consistency in error messages when a player lacks access.  

## Related Issue  

This pull request addresses [Issue #1632](https://github.com/bitburner-official/bitburner-src/issues/1632).


src/Bladeburner/Bladeburner.ts
```diff
@@ -402,15 +406,15 @@ export class Bladeburner implements OperationTeam {
         this.postToConsole("Automation: " + (this.automateEnabled ? "enabled" : "disabled"));
         this.postToConsole(
           "When your stamina drops to " +
-            formatNumberNoSuffix(this.automateThreshLow, 0) +
```
catloversg, line 405: You need to run `npm run format` before committing changes.

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -4628,12 +4646,12 @@ export interface Go {
    *
    * For example, a 5x5 board might look like this:
    *
-   [<br/>
```
catloversg, line 4631: These space characters are intentional. If your IDE makes these changes, you have to revert them before committing.

src/NetscriptFunctions/Bladeburner.ts
```diff
@@ -82,7 +87,7 @@ export function NetscriptBladeburner(): InternalAPI<INetscriptBladeburner> {
       return { name: blackOp.name, rank: blackOp.reqdRank };
     },
     getBlackOpRank: (ctx) => (_blackOpName) => {
-      checkBladeburnerAccess(ctx);
+      getBladeburner(ctx);
```
catloversg, line 90: The naming here is weird. `checkSomething()` clearly means that it will perform a check, which is exactly what it's doing. When you use `getBladeburner`, that name makes it look like we use a getter, but we don't care about the returned value. I think you are trying to refactor and expose `checkBladeburnerAPIAccess` as an API. IMO, `checkBladeburnerAPIAccess` is unnecessary. More about this later.

catloversg (changes requested):
Note: I'll use the phrase "activate X" as an "alias" for creating/joining/accepting gang/corporation/Bladeburner/Stanek.

From a player's perspective, they usually want to know:
- Can I activate a mechanic? For example: Can I activate X? This is useful because some activation APIs throw errors when I violate non-obvious constraints (e.g., `ns.corporation.createCorporation` throws in BN8). [1]
- Can I use NS APIs to manage a mechanic? For example: Did I activate X? If yes, can I call APIs of X to manage it? [2]

[1]: This is the kind of `isEligible` APIs that you want to add. For corporation, we have `ns.corporation.canCreateCorporation`.
[2]: There are 2 things here:
- `Did I activate X?`: We have `ns.gang.inGang`, `ns.corporation.hasCorporation`, `ns.bladeburner.inBladeburner`. There is no equivalent API for Stanek. [3]
- `can I call APIs of X to manage it?`: This is where things get messy. Generally, if you activated X, you would be able to call X's APIs, but there are still exceptions:
  - Bladeburner: No SF7 = No API [4]
  - Corporation: No SF3.3 = Have to buy Office API and Warehouse API before using them.

Due to [3], I suggest adding `ns.stanek.hasGift` instead of `ns.stanek.checkStanekAPIAccess`. It will be more consistent with old APIs.

Due to [4], I think `checkBladeburnerAPIAccess` is unnecessary. The player can call `inBladeburner` and check SF7 by using `ns.singularity.getOwnedSourceFiles` or `ns.getResetInfo().ownedSF`. The requirement of SF7 is **obvious**. It's mentioned in both documentation and the BN selection screen. If we merge #1407, the requirement of SF7 will be gone. In that case, calling `inBladeburner` is enough.

Another thing to discuss: Is `isEligible` a good name? IMO, it's better to explicitly call it canActivateX. For example, `canJoinBladeburner`, `canAcceptGift`, etc. Without looking at the documentation, `isEligible` may mean "Can I activate X" or "Can I manage X via APIs?". Having only one API name for all mechanics may be easier for players to find them, but I don't think there is enough benefit to make the trade-off.

d0sboots:
I agree with Catlover's comments. In particular, there is also a subtle issue with having functions named the same thing, related to the static-RAM check. It's not so much an issue for zero-cost functions, but it still becomes a complete non-issue if functions have unique names.

LJNeon:
@ivanrodgon do you plan to make the requested changes for this PR? If not would you mind if I open a new one with the requested changes made?

PR #1914, opened 2025-01-14
UI: Enhance navigation and accesibility with keyboard tabindex attributes

Paulaaass:
# Pull Request Description

### Summary
This pull request improves and includes keyboard accessibility for key areas of the Bitburner interface, ensuring that players can navigate menus and trigger actions without relying on a mouse. This aligns with the discusión in issue #464, where it was described the pressence of accessibility-hostile things that could gradually be cleaned up, such as making sure that keyboard navigation works correctly and including functionalities that can be triggered not only by cliking.

### Changes

**1. Added tabindex attribute to the terminal input field so players can focus on it with the Tab key.** 

•  The tabIndex: 0 in the InputProps ensures the input field is tabbable, allowing it to gain focus when the Tab key is pressed. Normal functionality is not affected and works correctly.

**2. Implemented keydown handling for the Enter key to execute terminal commands when the input field is focused.**
•  If the Enter key is pressed, the onKeyDown function checks if the input field is focused.
•  Includes logic to display the entered command and execute it using Terminal.executeCommands. 
•  For error handling a try-catch was added for robust handling during command execution, so that errors are logged and reported without crashing the terminal.

**3. Enabled focus feedback for the terminal to make it visually clear when it is active.**
•  Added isFocused state to track if the terminal input is active. And handleFocus and handleBlur methods update the focus state.
•  Added focusedInput class for active focus styling. This implementation ensures that the terminal visually indicates when it is active.
•  Included transition for a smoother visual effect when focus changes. This implementation ensures that the terminal visually indicates when it is active, improving the user experience.

**4. Added tabindex attribute to Active Scripts** 
•  This allows users to quickly navigate there using the keyboard. This improves accessibility and makes the interface more efficient and inclusive for keyboard navigation.

### DIFF
--- a/src/Terminal/ui/TerminalInput.tsx
+++ b/src/Terminal/ui/TerminalInput.tsx
@@ -17,7 +17,7 @@ const useStyles = makeStyles()((theme: Theme) => ({
     borderRadius: "4px",
     padding: theme.spacing(1),
     fontSize: "1rem",
-    color: theme.palette.text.primary,
+    color: '#04ce04',
     transition: "border-color 0.2s ease, box-shadow 0.2s ease",
   },
   focusedInput: {
--- a/src/Terminal/ui/TerminalInput.tsx
+++ b/src/Terminal/ui/TerminalInput.tsx
@@ -13,6 +13,16 @@ import { longestCommonStart } from "../../utils/StringHelperFunctions";
 const useStyles = makeStyles()((theme: Theme) => ({
   input: {
     backgroundColor: theme.colors.backgroundprimary,
+    border: "2px solid transparent",
+    borderRadius: "4px",
+    padding: theme.spacing(1),
+    fontSize: "1rem",
+    color: theme.palette.text.primary,
+    transition: "border-color 0.2s ease, box-shadow 0.2s ease",
+  },
+  focusedInput: {
+     borderColor: theme.palette.primary.main, // Highlights color when focused
+     boxShadow: `0 0 5px ${theme.palette.primary.main}`, // Subtle glow effect
   },
   nopadding: {
     padding: theme.spacing(0),
@@ -44,6 +54,7 @@ export function TerminalInput(): React.ReactElement {
   const [searchResults, setSearchResults] = useState<string[]>([]);
   const [searchResultsIndex, setSearchResultsIndex] = useState(0);
   const [autofilledValue, setAutofilledValue] = useState(false);
+  const [isFocused, setIsFocused] = useState(false); // New state for focus feedback
   const { classes } = useStyles();
 
   // If we have no data in the current terminal history, let's initialize it from the player save
@@ -70,6 +81,14 @@ export function TerminalInput(): React.ReactElement {
     }
   }
 
+  function handleFocus(): void {
+      setIsFocused(true); // Set focus state to true
+    }
+
+    function handleBlur(): void {
+      setIsFocused(false); // Set focus state to false
+    }
+
   function handleValueChange(event: React.ChangeEvent<HTMLInputElement>): void {
     saveValue(event.target.value);
     setPossibilities([]);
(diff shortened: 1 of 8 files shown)


src/Terminal/ui/TerminalInput.tsx
```diff
@@ -70,17 +81,26 @@ export function TerminalInput(): React.ReactElement {
     }
   }
 
+  function handleFocus(): void {
```
catloversg, line 84: These functions are unused. Why is that?

src/Terminal/ui/TerminalInput.tsx
```diff
@@ -70,17 +81,26 @@ export function TerminalInput(): React.ReactElement {
...
   function handleValueChange(event: React.ChangeEvent<HTMLInputElement>): void {
     saveValue(event.target.value);
     setPossibilities([]);
     setSearchResults([]);
     setAutofilledValue(false);
   }
 
-  function resetSearch(isAutofilled = false) {
+  function resetSearch(clearFocus?: boolean): void {
+    //Edited to ensure resetSearch clears all autocomplete and search states
```
catloversg, line 100: It seems that the comment explains the removal of `setAutofilledValue` and other changes, but I still don't get it.

> Edited to ensure ...

What was "edited"? Is there a bug that is fixed by this change? Why is it necessary?

Can you rewrite the comment here?

src/Terminal/ui/TerminalInput.tsx
```diff
@@ -44,6 +54,7 @@ export function TerminalInput(): React.ReactElement {
   const [searchResults, setSearchResults] = useState<string[]>([]);
   const [searchResultsIndex, setSearchResultsIndex] = useState(0);
   const [autofilledValue, setAutofilledValue] = useState(false);
+  const [isFocused, setIsFocused] = useState(false); // New state for focus feedback
```
catloversg, line 57: Nitpick (feel free to ignore it): "New something for something" is an explanation of the reason for adding this new variable. After this change is merged, looking back on this comment is weird. "New state" is not that "new" anymore. Suggestion: "This state is for ...".

catloversg (changes requested):
For this kind of UI change, you should take before-after screenshots/videos. This is a video that I recorded on the current stable version:

https://github.com/user-attachments/assets/5284e9cd-7cf9-41c6-837c-a7b1ce9793f2

In this video:
- Terminal input is automatically focused after loading and switching tabs.
- The player can navigate the "Active Scripts" page with tab.

What are your changes? I checked out your branch and ran it, but it's not obvious what the new changes are.

> Added tabindex attribute to the terminal input field so players can focus on it with the Tab key.

Can you show it with a before-after screenshot/video?

> Implemented keydown handling for the Enter key to execute terminal commands when the input field is focused.
> Enabled focus feedback for the terminal to make it visually clear when it is active.

This is your change:

after1
after2


You make the terminal input look "special", but, IMO, it looks confusing. In real terminal UI, I have never seen this kind of "special" UI.

cmd
git bash
powershell
putty


When I looked at it the first time, I thought that it's a UI bug, not a feature.

"No command entered." is also an unnecessary "feature".

> Added tabindex attribute to Active Scripts

Same as before, what is the new change?

d0sboots:
Some high-level comments:
* I agree that visual styling for the terminal being focused is likely not desired.
* In general, there is going to be quite a bit of conflict/tension between the general web-centric method of tab navigation, and the terminal-centric method of using tab for tab completion. I'm not really sure how to reconcile that, aside from maybe adding a setting. Definitely tab completion can't be removed, but at the same time it seems very awkward if the commandline becomes a focus "trap" that can't be escaped from via keyboard nav.

d0sboots:
Per your other PR: You're operating with a single branch branch `Paulaaass:dev`. This means all your commits that you intend for both PRs (separately) are being mixed together. If you're familiar with Git and branches, we can help walk you through untangling this state. If you're rather new to Git, it might be best to work on one change at a time for the moment.

PR #1917, opened 2025-01-20
IPVGO: Support scripts playing against each other as each color on "No AI" boards

### LINKED ISSUE #1576: DOCUMENTATION: Insufficient alternative for ns.go.analysis.getValidMoves()
Currenr dev branch

https://github.com/bitburner-official/bitburner-src/blob/23f98c1ff6cbff0953dd6ce348433a3d5aa2d851/markdown/bitburner.goanalysis.getvalidmoves.md?plain=1#L26

We also have to determine getMoveHistory() to comply no-repeat rule.

https://github.com/bitburner-official/bitburner-src/blob/23f98c1ff6cbff0953dd6ce348433a3d5aa2d851/src/Go/boardAnalysis/boardAnalysis.ts#L46-L47

The comment says only one board of history is checked, but the code actually checks full history.

### LINKED ISSUE #1742: Documentation for ns.hack(), ns.weaken() and ns.grow()
I'm referring to the NS API documentation, which is here: https://github.com/bitburner-official/bitburner-src/blob/stable/markdown/bitburner.ns.md

The documentation for ns.hack(), ns.weaken() and ns.grow() should include formulae for outcomes. I don't like that the documentation might only mention that "your hacking level and server security and some inherent factor of the server" play a role. It makes it difficult to anticipate problems with a hacking design you're thinking of which forces you to guess or tediously try to reverse engineer the formula yourself, and when things go wrong you don't have enough information to figure out why. It should also be clear in these formulae when each of these factors are checked (when the promise resolves, or when the function is called?)

Whilst getters such as getHackTime, getWeakenTime and getGrowTime exist, how these values are calculated is also something you need to know to design a good hacking system. You'll be asking questions like "what's the optimal way to grow a server to max money?" and without the formulae, you're going to have a hard time figuring that out because you can't easily figure out if two scripts with 50 threads each will be better than one with 100 threads. You can also be misled into thinking the getters factor in some inherent server factor.

For ns.hack(), changes needed are:
- Include formula for how much money will be stolen
- Convert the line "A successful hack() on a server will raise that server's security level by 0.002" to a formula to clarify whether this is per-thread or per-script.
- Include formula for how much hacking exp will be gained
- Include formula for how long it will take

For ns.weaken(), changes needed are:
- Convert line "This function lowers the security level of the target server by 0.05" to a formula which shows that the server will be weakened by 0.05 * number_of_threads
- Include formula for how much hacking exp will be gained
- Include formula for how long it will take

For ns.grow(), changes needed are:
- Include formula for the new server money after your operation completes
- Include formula for how much the server security will increase
- Include formula for how much hacking exp will be gained
- Include formula for how long it will take

ficocelliguy:
Branch testable here: https://ficocelliguy.github.io/bitburner-src/

A few people in discord have asked for the ability to pit multiple IPvGO scripts against each other. This is the most recent:
https://discord.com/channels/415207508303544321/1221920484111814828/1330501717674168423

This PR adds support for making moves as the white player via the API. This can only be done on the "No AI" practice board, which gives no stats. To do this, additional promise handling was added so that the two scripts could listen for when the other script (or player via UI) has made a move.

In addition, documentation improvements and a new Go.analytics.getHistory() method were added to resolve this feedback:
Closes #1576


Sphyxis:
I look forward to implementing this in my script and so do many other people.

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -4333,11 +4333,13 @@ export interface GoAnalysis {
...
    *  (no duplicate board states in the full game history) is not supported; you will have to implement your own analysis for that.
    *
+   *  playAsWhite is optional, and gets the current valid moves for the white player. Intended to be used when playing as white when the opponent is set to "No AI"
+   *
    * @remarks
    * RAM cost: 8 GB
    * (This is intentionally expensive; you can derive this info from just getBoardState() )
    */
-  getValidMoves(boardState?: string[], priorBoardState?: string[]): boolean[][];
+  getValidMoves(boardState?: string[], priorBoardState?: string[], playAsWhite): boolean[][];
```
```suggestion
  getValidMoves(boardState?: string[], priorBoardState?: string[], playAsWhite?: boolean): boolean[][];
```

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -4556,6 +4583,7 @@ export interface GoCheat {
   destroyNode(
     x: number,
     y: number,
+    playAsWhite?: boolean,
```
d0sboots: Optional: It probably makes more sense to do this here and wherever you can.
```suggestion
    playAsWhite = false,
```

src/Go/effects/netscriptGoImplementation.ts
```diff
@@ -104,42 +120,53 @@ export function validateTurn(error: (s: string) => void, moveString = "") {
...
+  logger(`Go move played: ${x}, ${y}${playAsWhite ? " (White)" : ""}`);
+  return makeAIMove(boardState, true, playAsWhite);
 }
 
 /**
   Returns the promise that provides the opponent's move, once it finishes thinking.
  */
-export async function getOpponentNextMove(logOpponentMove = true, logger: (s: string) => void) {
+export async function getOpponentNextMove(logOpponentMove = true, logger: (s: string) => void, playAsWhite = false) {
+  const nextMovePromise = playAsWhite ? "nextTurnForWhite" : "nextTurn";
```
d0sboots: This really solidifies my feeling that these should not be separate variables. Rather, there should be a single `nextTurn` object that is indexed by GoColor, or possibly a two-length array. The same probably makes sense for most of the things you've split out as separate variables.

src/Go/SaveLoad.ts
```diff
@@ -88,14 +90,18 @@ export function loadGo(data: unknown): boolean {
...
-  // If it's not the AI's turn and we're not in gameover status, initialize nextTurn promise based on the previous move/pass
+  // If it's not the AI's turn (and we're not in gameover status), initialize nextTurn promise based on the previous move/pass
   else if (currentGame.previousPlayer) {
     const previousMove = getPreviousMove();
     Go.nextTurn = Promise.resolve(
       previousMove
         ? { type: GoPlayType.move, x: previousMove[0], y: previousMove[1] }
         : { type: GoPlayType.pass, x: null, y: null },
     );
+    Go.nextTurnForWhite = cloneDeep(Go.nextTurn);
```
d0sboots: This is deeply suspicious to me. What happens if the ai is *not* GoOpponent.none?

Also, why even clone this? What's the issue with them referencing the same promise, if this usage was OK in the first place?

d0sboots (commented):
I still need to take a deeper look at the nextTurn stuff, which is the part with the most magic (always has been).

src/Go/effects/netscriptGoImplementation.ts
ficocelliguy: I reworked this to be a GoColor object, which is definitely cleaner

src/Go/SaveLoad.ts
ficocelliguy: This was part of testing for a bug where the wrong promise was being resolved, but I don't need it anymore. Good catch

src/Go/boardAnalysis/goAI.ts
```diff
@@ -24,26 +25,36 @@ import { Go, GoEvents } from "../Go";
...
-export function makeAIMove(boardState: BoardState, useOfflineCycles = true): Promise<Play> {
+export function makeAIMove(
+  boardState: BoardState,
+  useOfflineCycles = true,
+  scriptPlayingAsWhite = false,
+): Promise<Play> {
   // If AI is already taking their turn, return the existing turn.
   if (isAiThinking) {
-    return Go.nextTurn;
+    if (scriptPlayingAsWhite && currentTurnResolverAsWhite) {
```
d0sboots: The extra condition here worries me. The natural extension is `if (scriptPlayingAsWhite)`; it is trivially obvious that one of the two branches will be taken in that case. As it stands now, I can't be sure that the logic won't fall out of this if-block.

src/Go/boardAnalysis/goAI.ts
```diff
@@ -24,26 +25,36 @@ import { Go, GoEvents } from "../Go";
...
   if (boardState.ai === GoOpponent.none) {
-    resetAI();
+    updateTurnPromises();
   }
-  // If an AI is in use, find the faction's move in response, and resolve the Go.nextTurn promise once it is found and played.
+  // If an AI is in use, find the faction's move in response, and resolve the Go.nextTurn[GoColor.black] promise once it is found and played.
   else {
     const currentMoveCount = Go.currentGame.previousBoards.length;
-    Go.nextTurn = getMove(boardState, GoColor.white, Go.currentGame.ai, useOfflineCycles).then(
+    Go.nextTurn[GoColor.black] = getMove(boardState, GoColor.white, Go.currentGame.ai, useOfflineCycles).then(
```
d0sboots: Not for the current PR, but: I feel like this (and other places) have a lot of color hardcoding that actually makes the code more complicated. With a little bit of refactoring, not only could the AI play as black (potentially), but also everything would probably be a lot simpler.

src/Go/boardAnalysis/goAI.ts
```diff
@@ -83,35 +94,91 @@ export function makeAIMove(boardState: BoardState, useOfflineCycles = true): Pro
...
+// Returns a promise that resolves with the previous move details when the other player / script / AI makes a move
+function getNonAIMovePromise(playingAsWhite = false): Promise<Play> {
+  return new Promise((resolve) => {
+    const handler = () => resolve(getPreviousMoveDetails());
+    if (playingAsWhite) {
+      currentTurnResolverAsWhite = handler;
+    } else {
+      currentTurnResolver = handler;
+    }
+    // In cse the promise has become orphaned or overwritten, resolve if the other player makes a move
```
d0sboots:
```suggestion
    // In case the promise has become orphaned or overwritten, resolve if the other player makes a move
```

d0sboots: This is a maximum-level code smell. If orphaned/overwritten promises are occurring, that is *not* something we should be papering over with a mechanism like this, because it *will* manifest as weird unreproducible edge cases in people's scripts and this "fix" will only make it harder to track down the issues.

src/Go/boardAnalysis/goAI.ts
```diff
@@ -83,35 +94,91 @@ export function makeAIMove(boardState: BoardState, useOfflineCycles = true): Pro
...
+      }
+    }, 250);
+  });
 }
 
 /**
  * Resolves the current turn.
  * This is used for players manually playing against their script on the no-ai board.
  */
 export function resolveCurrentTurn() {
```
d0sboots: These don't need to be exported anymore (and thus, it's best that they aren't.)

src/Go/boardAnalysis/goAI.ts
```diff
@@ -83,35 +94,91 @@ export function makeAIMove(boardState: BoardState, useOfflineCycles = true): Pro
...
+  }
+
+  const previousTurn = getPreviousMove();
+  const previousTurnPromise: Promise<Play> = Promise.resolve(
+    previousTurn
+      ? { type: GoPlayType.move, x: previousTurn[0], y: previousTurn[1] }
+      : { type: GoPlayType.pass, x: null, y: null },
+  );
+
+  if (Go.currentGame.previousPlayer !== GoColor.black) {
```
d0sboots:
```suggestion
  if (Go.currentGame.previousPlayer === GoColor.white) {
```
IMO it's easier to understand these in positive logic. Otherwise, I wonder what happens if two of them match.

src/Go/effects/netscriptGoImplementation.ts
```diff
@@ -104,42 +120,53 @@ export function validateTurn(error: (s: string) => void, moveString = "") {
...
 /**
   Returns the promise that provides the opponent's move, once it finishes thinking.
  */
-export async function getOpponentNextMove(logOpponentMove = true, logger: (s: string) => void) {
+export async function getOpponentNextMove(logOpponentMove = true, logger: (s: string) => void, playAsWhite = false) {
+  const playerColor = playAsWhite ? GoColor.white : GoColor.black;
   // Only asynchronously log the opponent move if not disabled by the player
   if (logOpponentMove) {
-    return Go.nextTurn.then((move) => {
+    logger("Waiting for opponent to make their move...");
```
d0sboots: I thought this was handled elsewhere? I seem to remember a big discussion about this.

src/Go/effects/netscriptGoImplementation.ts
```diff
@@ -151,18 +178,25 @@ export async function getOpponentNextMove(logOpponentMove = true, logger: (s: st
...
 /**
- * Returns a grid of booleans indicating if the coordinates at that location are a valid move for the player (black pieces)
+ * Returns a grid of booleans indicating if the coordinates at that location are a valid move for the player
  */
-export function getValidMoves(_boardState?: BoardState) {
+export function getValidMoves(_boardState?: BoardState, playAsWhite = false) {
   const boardState = _boardState || Go.currentGame;
+  const color = playAsWhite ? GoColor.white : GoColor.black;
+
+  // If the game is over, or if it is not your turn, there are no valid moves
```
d0sboots, line 192: This shouldn't be needed. (The reduce after this should do the same, and if it doesn't, then it ough to throw an error instead.)

src/Go/boardAnalysis/goAI.ts
```diff
@@ -24,26 +25,36 @@ import { Go, GoEvents } from "../Go";
 
 let isAiThinking: boolean = false;
 let currentTurnResolver: (() => void) | null = null;
+let currentTurnResolverAsWhite: (() => void) | null = null;
 
 /**
  * Retrieves a move from the current faction in response to the player's move
  */
-export function makeAIMove(boardState: BoardState, useOfflineCycles = true): Promise<Play> {
+export function makeAIMove(
```
d0sboots: Optional-ish: This method is increasingly misnamed. In cases like noAI there's no AI moves at all, and it's being called in response to *both* sides' moves. Something like `handleNextTurn` might be more appropriate.

src/Go/boardAnalysis/goAI.ts
```diff
@@ -24,26 +25,36 @@ import { Go, GoEvents } from "../Go";
 
 let isAiThinking: boolean = false;
 let currentTurnResolver: (() => void) | null = null;
+let currentTurnResolverAsWhite: (() => void) | null = null;
 
 /**
  * Retrieves a move from the current faction in response to the player's move
```
d0sboots: This comment should be updated to describe the full responsibilities of this function. (It responds to a player's move by possibly using the AI to get a new move, and returns an appropriate promise based on the type of game.)

src/Go/ui/GoGameboardWrapper.tsx
```diff
@@ -119,9 +117,9 @@ export function GoGameboardWrapper({ showInstructions }: GoGameboardWrapperProps
   }
 
   async function takeAiTurn(boardState: BoardState) {
-    // If white is being played manually, halt and notify any scripts playing as black if present, instead of making an AI move
+    // Instead of making an AI move on the "No AI" opponent, notify any scripts waiting for their turn and then halt
```
d0sboots: This code shouldn't be needed; the script-play version doesn't have it, but relies on `makeAIMove` to do the right thing. It's better if they both go through the same codepath.

d0sboots (commented):
Every time I have to review this, I have to pull the full state of how the promise stuff works back into cache. XD I think I almost have a coherent set of suggestions to make, but I'm still ruminating. Pulling everything together into a single `update` method is the right plan.

src/Go/effects/netscriptGoImplementation.ts
d0sboots: I removed this in my changes, but feel free to make a case for putting it back if you think it should be there. (The part that seemed odd to be was that this says "waiting" but it won't log until *after* the opponent moves.)

d0sboots:
I went through and resolved all the comments I handled in my PR, which was all of them XD

PR #1922, opened 2025-01-23
CODEBASE: Merge TypeAssertion files

catloversg:
I mentioned this change at https://github.com/bitburner-official/bitburner-src/pull/1775#issuecomment-2480567235. While searching for "TypeAssertion-like" files, I also found `src\Netscript\TypeAssertion.ts`. This PR merges 3 files:
- `src\Netscript\TypeAssertion.ts`
- `src\utils\helpers\typeAssertion.ts`
- `src\utils\TypeAssertion.ts`

I also renamed merged functions. These functions are called "assertSomething" now. The most notable change is `assertString` (`src\Netscript\TypeAssertion.ts`) -> `assertStringWithNSContext`. In `src\utils\helpers\typeAssertion.ts`, we had `stringAssert`. It's too confusing to have `stringAssert` and `assertString` that do completely different things.

I'm not really happy with `assertStringWithNSContext`, but I cannot find a better name.


src/utils/TypeAssertion.ts
```diff
@@ -1,4 +1,114 @@
 import type { Unknownify } from "../types";
+import type { NetscriptContext } from "../Netscript/APIWrapper";
```
d0sboots: I have only one concern, which is: Does this increase our circular dependency load?

It *might* make sense for the "withContext" functions to continue to live where they did, because of this.

catloversg:
After considering the concern from d0sboots, I decided to **not** merge functions in `src\Netscript\TypeAssertion.ts` to avoid worsening our "dependency hell".

PR #1923, opened 2025-01-23
DOCUMENTATION: Make nuke and port opening functions return boolean

catloversg:
I noticed this problem when looking at https://github.com/bitburner-official/bitburner-src/pull/1023#issuecomment-1892544463. With these APIs, we return a boolean value, but the API documentation uses the `void` type. Technically, this is a breaking change for TS players, but I don't think it'll affect many players.

### DIFF
--- a/markdown/bitburner.ns.brutessh.md
+++ b/markdown/bitburner.ns.brutessh.md
@@ -9,7 +9,7 @@ Runs BruteSSH.exe on a server.
 **Signature:**
 
 ```typescript
-brutessh(host: string): void;
+brutessh(host: string): boolean;
 ```
 
 ## Parameters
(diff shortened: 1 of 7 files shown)


d0sboots:
Would this even be breaking, since it's just a widening of the return type?

In any case, I think I'm going to declare right now that TS is "2nd tier" in the sense that API type alignment won't be guaranteed from version to version, although the fixes should be "simple" (because due to our underlying JS compatibility guarantee, there shouldn't be significant *semantic* changes).

Without that, supporting TS would essentially freeze the entire NS api and prevent us from making **any** changes.

Between TS users being rare, and also being more savvy, I'm hoping this won't cause significant disruption.

catloversg:
> Would this even be breaking, since it's just a widening of the return type?

Maybe "breaking" was not the best word. I just mean that this PR changes the "contract" of those APIs (void -> boolean). I don't think it will break anybody's scripts. Even if it does, it will only affect a small subset of "savvy" players, so it's not a big deal.

I agree with what you said.

PR #1926, opened 2025-01-25
BB: Allow API Usage in BN 6 without 7 - Grants Augment for BN 7 completion

Sphyxis:
This allows the user to use the Netscript API in BN 6 without having completed BN7 first.  Also grants The Blade's Simulacrum augment when starting BladeBurner if you have completed BN 7.

### DIFF
--- a/src/NetscriptFunctions/Bladeburner.ts
+++ b/src/NetscriptFunctions/Bladeburner.ts
@@ -26,7 +26,7 @@ export function NetscriptBladeburner(): InternalAPI<INetscriptBladeburner> {
     return;
   };
   const getBladeburner = function (ctx: NetscriptContext): Bladeburner {
-    const apiAccess = canAccessBitNodeFeature(7);
+    const apiAccess = canAccessBitNodeFeature(7) || canAccessBitNodeFeature(6);
     if (!apiAccess) {
       throw helpers.errorMessage(ctx, "You have not unlocked the Bladeburner API.", "API ACCESS");
     }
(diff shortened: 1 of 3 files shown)


src/NetscriptFunctions/Bladeburner.ts
```diff
@@ -307,7 +307,7 @@ export function NetscriptBladeburner(): InternalAPI<INetscriptBladeburner> {
       return !!attempt.success;
     },
     joinBladeburnerDivision: (ctx) => () => {
-      if (!canAccessBitNodeFeature(7) || Player.bitNodeOptions.disableBladeburner) {
+      if ((!canAccessBitNodeFeature(7) && !canAccessBitNodeFeature(6)) || Player.bitNodeOptions.disableBladeburner) {
```
Alpheus: Small suggestion:
I think it's worth in this case to split the if into two ifs:
- return false if BB is disabled
- return false if ~~!(no6 && no7)~~ (!no6 && !no7)

d0sboots: +1, since these are just quick early-out false checks. (Also, the 2nd condition Alpheus suggested is wrong, so keep the current logic XD)

d0sboots:
Note: This is a continuation of #1407, PR redone due to merge conflicts (presumably).

src/NetscriptFunctions/Bladeburner.ts
```diff
@@ -307,7 +307,9 @@ export function NetscriptBladeburner(): InternalAPI<INetscriptBladeburner> {
       return !!attempt.success;
     },
     joinBladeburnerDivision: (ctx) => () => {
-      if ((!canAccessBitNodeFeature(7) && !canAccessBitNodeFeature(6)) || Player.bitNodeOptions.disableBladeburner) {
+      if (!canAccessBitNodeFeature(7) && !canAccessBitNodeFeature(6)) {
+        return false; //Does not have bitnode 6 or 7
+      } else if (Player.bitNodeOptions.disableBladeburner) {
```
d0sboots, line 312: The else isn't needed, but that's OK.

PR #1284, opened 2024-05-16
IPVGO: Add defence boost to Tetrads

gmcew:
Based on discord [suggestion](https://discordapp.com/channels/415207508303544321/1221920484111814828/1236087482618089492), Tetrads now do all four combat stats.

PR #1927, opened 2025-01-26
IPVGO: ensure def is treated the same as the other stats boosted by playing against tetrads

ficocelliguy:
When defense was added to tetrads' bonus in #1284 , the bitnode multipliers for BN 14 were not updated. This fixes defense to be more consistent


d0sboots:
Had to double-check some stuff with the background of the situation.

PR #1929, opened 2025-01-28
BUGFIX: Fix terminal auto complete not properly handling folder casing

### LINKED ISSUE #1708: Terminal: tab auto-complete doesn't correct folder casing
Steps to reproduce:
1. Have a folder created with upper-case letters, such as "Tasks/", and some scripts in this folder.
  image

2. Type the folder name incorrectly as all-lower-case ("tasks/") then proceed to tab-autocomplete a file name.
   Note that the auto-complete will correct the case of the filename if you typed any letters in the incorrect case, but it will *not* correct the case of the folder name:
  image

3. This will now open a new file (in a new folder) rather than the existing file you tab-auto-completed:
  image

Acceptable solutions (IMO):
- Correct the casing in the full path (including folder name) when we tab-autocomplete - currently we just correc the filename casing.
  or
- Once the folder name has been completed ("/" is present in the path) - tab auto-complete should no longer list any files, because the folder does not exist. Specifically, it should not list files that exist in a similarily-named folder, but with different casing.

A side-effect of the current behaviour (ignoring folder casing when auto-completing) is that if the user creates two folders with the same name but different casing, these folders are merged in the auto-complete box. In this case, since each folder now has a file named "backdoor-all-servers.js", this entry appears twice:
image

CaseZumbrum:
closes #1708 

Removed the conversion of paths to lowercase for the terminal auto completion. Previously, this would allow similarly named folders (scripts vs. Scripts) to share tab auto completion options.

Bug

Image showing improper casing, the "Scripts" folder does not exist but the terminal suggests files within the similar "scripts" folder.

fixed

Image showing proper casing, the "Scripts" folder does not exist, and no recommendations are given.

Tested with nested folders and ensured that auto completion still occurs on partial folder names (ex: typing "Scr" + TAB will autocomplete to "scripts")

### DIFF
--- a/src/Terminal/getTabCompletionPossibilities.ts
+++ b/src/Terminal/getTabCompletionPossibilities.ts
@@ -51,8 +51,8 @@ export async function getTabCompletionPossibilities(terminalText: string, baseDi
     baseDir = path;
     pathingRequiredMatch = currentText.replace(/^.*\//, path);
   } else if (baseDir !== root) {
-    pathingRequiredMatch = (baseDir + currentText);
-  } else{
+    pathingRequiredMatch = baseDir + currentText;
+  } else {
     pathingRequiredMatch = currentText.toLowerCase();
   }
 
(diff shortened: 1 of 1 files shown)


d0sboots:
The failing test indicates you broke some stuff. Code inspection also indicates (to me) that you broke cases like tab-completing `Script/foo` when `Script` is a directory that exists; I think you didn't test this well enough.

d0sboots:
Let me level with you: I don't think this is a good first bug to tackle. It is one of those that looks like a small thing on the surface, but secretly it is an entire iceberg of complexity underneath. 
I'm not sure what a correct fix to this bug would look like, but I know that I'd probably have to spend at least an hour working through the cases to figure it out. (And a similar amount of time to verify a potential fix.)

PR #1931, opened 2025-01-29
UI: In Operation Wallace. Reword so "The Dark Army" works, for grammar

Hihahahalol:
This is the same as what's done in "Operation Wallace" for "Syndicate"

### DIFF
--- a/src/Bladeburner/data/BlackOperations.ts
+++ b/src/Bladeburner/data/BlackOperations.ts
@@ -382,7 +382,7 @@ export const BlackOperations: Record<BladeburnerBlackOpName, BlackOperation> = {
       `Augmentation storage facilities. We figured out that ${FactionName.TheDarkArmy} was behind the heist, and didn't think ` +
       "any more of it. However, we've just discovered that several known MK-VI Synthoids were part of that break-in group.\n\n" +
       "We cannot have Synthoids upgrading their already-enhanced abilities with Augmentations. Your task is to hunt " +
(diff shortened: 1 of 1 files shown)


src/Bladeburner/data/BlackOperations.ts
```diff
@@ -382,7 +382,7 @@ export const BlackOperations: Record<BladeburnerBlackOpName, BlackOperation> = {
       `Augmentation storage facilities. We figured out that ${FactionName.TheDarkArmy} was behind the heist, and didn't think ` +
       "any more of it. However, we've just discovered that several known MK-VI Synthoids were part of that break-in group.\n\n" +
       "We cannot have Synthoids upgrading their already-enhanced abilities with Augmentations. Your task is to hunt " +
-      `down associated ${FactionName.TheDarkArmy} members and eliminate them.`,
+      `down associated Dark Army members and eliminate them.`,
```
catloversg:
```suggestion
      `down associated members of ${FactionName.TheDarkArmy} and eliminate them.`,
```
- "The" in "The Dark Army" is not only an article. It's a part of the full name of that faction.
- Never replace an enum usage with a hard-coded string.

Hihahahalol: You might also wanna check line 422 where the Same thing was done with "The Syndicate"

d0sboots:
Ping me on discord if you're there and want the contributor role

PR #1933, opened 2025-01-30
MISC: Standardize behavior of connect CLI and Singularity API

catloversg:
`connect` CLI and the equivalent Singularity API are inconsistent in their behaviors.

`ns.singularity.connect` only supports:
- Go to home, regardless of the currently connected server.
- Go to backdoored servers, regardless of the currently connected server.
- Go to adjacent servers.

`connect` CLI also supports `server.purchasedByPlayer`.

This PR standardizes their behaviors.

### DIFF
--- a/src/Server/ServerHelpers.ts
+++ b/src/Server/ServerHelpers.ts
@@ -12,8 +12,6 @@ import { workerScripts } from "../Netscript/WorkerScripts";
 import { killWorkerScriptByPid } from "../Netscript/killWorkerScript";
 import { serverMetadata } from "./data/servers";
 import { exceptionAlert } from "../utils/helpers/exceptionAlert";
-import { Terminal } from "../Terminal";
-import { root } from "../Paths/Directory";
 
 /**
  * Constructs a new server, while also ensuring that the new server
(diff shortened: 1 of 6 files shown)


src/Server/ServerHelpers.ts
```diff
@@ -12,6 +12,8 @@ import { workerScripts } from "../Netscript/WorkerScripts";
 import { killWorkerScriptByPid } from "../Netscript/killWorkerScript";
 import { serverMetadata } from "./data/servers";
 import { exceptionAlert } from "../utils/helpers/exceptionAlert";
+import { Terminal } from "../Terminal";
```
d0sboots: I'm a bit worried about cycles introducing this into server helpers. In addition, connectServer feels more like a Terminal.ts function than a server function to me.

src/Server/ServerHelpers.ts
```diff
@@ -271,3 +273,10 @@ export function getWeakenEffect(threads: number, cores: number): number {
   const coreBonus = getCoreBonus(cores);
   return ServerConstants.ServerWeakenAmount * threads * coreBonus * currentNodeMults.ServerWeakenRate;
 }
+
+export function connectServer(server: BaseServer): void {
+  Player.getCurrentServer().isConnectedTo = false;
+  Player.currentServer = server.hostname;
+  Player.getCurrentServer().isConnectedTo = true;
```
d0sboots: Couldn't you just use server here?

PR #1935, opened 2025-01-30
API: Move tail-related APIs to another namespace

catloversg:
This PR implements the change discussed at https://discord.com/channels/415207508303544321/459097632896188436/1318894499170095176 and https://discord.com/channels/415207508303544321/459097632896188436/1332765650728390697.

Changes:
- `ns.tail` -> `ns.ui.openTail`
- `ns.moveTail` -> `ns.ui.moveTail`
- `ns.resizeTail` -> `ns.ui.resizeTail`
- `ns.setTitle` -> `ns.ui.setTailTitle`
- `ns.setTailFontSize` -> `ns.ui.setTailFontSize`
- `ns.closeTail` -> `ns.ui.closeTail`
- `ns.renderTail` -> `ns.ui.renderTail` (This API has not been released, so we don't have to maintain the old version).

Note: I don't implement migration code and API break notice in this PR. I think we should do that in another PR when we decide to remove the support of the old APIs. Maybe it's v3.0.0.

Test code:
```js
/** @param {NS} ns */
  ns.tail();
  ns.moveTail(100, 100);
  ns.resizeTail(500, 500);
  ns.setTitle("test");
  ns.setTailFontSize(15);
  ns.closeTail();
}
```

Screenshot:
Capture


d0sboots:
setTailFontSize was also within this version, so doesn't need deprecation.

Edit: Other than that, looks good.

PR #1938, opened 2025-01-31
DOCUMENTATION: Fix typo in proof of boost material optimizer

catloversg:
Before:

before

After:

after


d0sboots:
Who read through all of it? XD

PR #1939, opened 2025-01-31
BUGFIX: Wrong plural form in modal of coding contract

catloversg:
A player reported this bug on Discord. In that modal, we show `1 tries` instead of `1 try`. While fixing the bug, I see that we rewrite this snippet everywhere:
```js
count + ` line${count !== 1 ? "s" : ""}`
```
Therefore, I created a utility function for it. We don't have any plans to support more languages in the near future, so using `Intl.PluralRules` is a bit unnecessary, but it's still harmless.

I also intentionally skip rewriting some code that I don't think they need to use `pluralize`. For example:
https://github.com/bitburner-official/bitburner-src/blob/f337de47dff7063b080403ab8b223dd9f1b08e85/src/utils/StringHelperFunctions.ts#L49
https://github.com/bitburner-official/bitburner-src/blob/f337de47dff7063b080403ab8b223dd9f1b08e85/src/Terminal/commands/runScript.ts#L64

Screenshots:

### DIFF
--- a/src/Terminal/commands/grep.ts
+++ b/src/Terminal/commands/grep.ts
@@ -5,6 +5,7 @@ import { ContentFile, ContentFilePath, allContentFiles } from "../../Paths/Conte
 import { Settings } from "../../Settings/Settings";
 import { help } from "../commands/help";
 import { Output } from "../OutputTypes";
+import { pluralize } from "../../utils/I18nUtils";
 
 type LineParser = (options: Options, filename: string, line: string, i: number) => ParsedLine;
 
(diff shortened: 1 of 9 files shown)


src/Terminal/commands/grep.ts
```diff
@@ -304,9 +302,9 @@ class Results {
       `\n${
         (this.params.maxMatches ? this.params.maxMatches : matchCount) + (options.isInvertMatch ? " INVERTED" : "")
       } `,
-      suffix("line", matchCount) + " matched ",
+      pluralize(matchCount, "line", undefined, true) + " matched ",
```
d0sboots, line 305: I don't think this one is supposed to skip the value?

src/utils/I18nUtils.ts
```diff
@@ -0,0 +1,9 @@
+const pluralRules = new Intl.PluralRules("en-US");
+
+export function pluralize(count: number, singular: string, plural?: string, skipCountInReturnedValue = false): string {
+  const pluralRule = pluralRules.select(count);
+  if (pluralRule === "one") {
+    return `${!skipCountInReturnedValue ? `${count} ` : ""}${singular}`;
```
d0sboots: The outer layer of string interpolation is unneeded and just makes things more confusing IMO

d0sboots: I think these would be easier to follow if you refactor the if skipCount out earlier, and then you could add the resulting prefix unconditionally.

src/Terminal/commands/grep.ts
catloversg: `suffix` only adds "s" to the singular form of the string as needed. It does not add text for the count value.
```js
const suffix = (pre: string, num: number) => pre + (num === 1 ? "" : "s");
```
Capture

PR #1943, opened 2025-01-31
MISC: Decrease factionInviteCheckInterval and check for Invites on faction page load

TheAimMan:
Current faction check interval is every 20 seconds.  This can cause confusion when you're watching your stats/requirements to be invited to a faction and no invite comes.

This change will hopefully address that by decreasing the check time from every 20 seconds to every 5 seconds, as well as checking any time you load the faction page, so you don't need to wait for the 5 second interval.

### DIFF
--- a/src/engine.tsx
+++ b/src/engine.tsx
@@ -160,7 +160,7 @@ const Engine: {
     updateActiveScriptsDisplay: 5,
     createProgramNotifications: 10,
     augmentationsNotifications: 10,
-    checkFactionInvitations: 25,
+    checkFactionInvitations: 10
     passiveFactionGrowth: 5,
     messages: 150,
     mechanicProcess: 5, // Process Bladeburner
@@ -188,7 +188,7 @@ const Engine: {
       for (const invitedFaction of invitedFactions) {
         inviteToFaction(invitedFaction);
       }
-      Engine.Counters.checkFactionInvitations = 25;
+      Engine.Counters.checkFactionInvitations = 10;
     }
 
     if (Engine.Counters.passiveFactionGrowth <= 0) {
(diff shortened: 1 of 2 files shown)


src/ui/GameRoot.tsx
```diff
@@ -281,6 +281,8 @@ export function GameRoot(): React.ReactElement {
       break;
     }
     case Page.Factions: {
+      Engine.Counters.checkFactionInvitations = 0;
```
d0sboots: Take this out please. Running a checkCounters outside the normal scheduled time has all sorts of exploit potential, also there are dependency loop issues with including engine in other files.

src/engine.tsx
```diff
@@ -188,7 +188,7 @@ const Engine: {
       for (const invitedFaction of invitedFactions) {
         inviteToFaction(invitedFaction);
       }
-      Engine.Counters.checkFactionInvitations = 100;
+      Engine.Counters.checkFactionInvitations = 25;
```
d0sboots: I'd be OK dropping this as low as 10 (2 seconds) - enough that there's a bit of a pause, but fast enough to work through multiple invitations quite quickly. That should eliminate the desire to update when you go to the factions screen.

src/ui/GameRoot.tsx
TheAimMan: Hrmmm, I was just copying what the singularity function already does.
https://github.com/bitburner-official/bitburner-src/blob/4a452b8aca242f8f3152036b54a4a06811a0d5d3/src/NetscriptFunctions/Singularity.ts#L758
Does that need changed?
Based on your engine comment, I agree this wouldn't be needed if the check was reduced to 2 seconds.

src/engine.tsx
TheAimMan: Sounds good.

src/ui/GameRoot.tsx
d0sboots: Yeah, that's a bit gross and I don't think sing should be doing that. But, we won't touch that here.

PR #1945, opened 2025-02-01
UI: Show all jobs instead of only one in Job tab

catloversg:
This change was suggested on Discord: https://discord.com/channels/415207508303544321/415213435974975508/1333736904411058206.

First suggestion:
```
When you click on the "Job" tab, I feel like it should take you to the job you're doing (or done last?) instead of the first company you joined
It's a bit annoying to scroll through jobs every time you click the tab to find the one you're on
```
Cald's suggestion:
```
instead of a scroll it might be good to have several job cards there
```

I think Cald's suggestion is better.

A small downside is that the UI becomes a bit sluggish when I apply to too many companies. Maybe it's just because my machine is too old. This is also a problem when joining too many factions and opening the Factions tab. I'm too lazy to implement the pagination feature right now. I'll do that if the maintainer wants.

Test code:
```js
/** @param {NS} ns */
  ns.tail();
  ns.clearLog();
  for (const companyName of Object.values(ns.enums.CompanyName)) {
    ns.singularity.quitJob(companyName);
  }
  let count = 0;
  for (const job of Object.values(ns.enums.JobField)) {
    for (const companyName of Object.values(ns.enums.CompanyName)) {
      if (ns.singularity.applyToCompany(companyName, job)) {
        ++count;
      }
      if (count === 20) {
        return;
      }
    }
  }
}
```

Before:

before

After:

6 columns:

after1

Responsive:

after2

### DIFF
--- a/src/Locations/ui/CompanyLocation.tsx
+++ b/src/Locations/ui/CompanyLocation.tsx
@@ -50,7 +50,6 @@ export function CompanyLocation(props: IProps): React.ReactElement {
 
   /** Name of company position that player holds, if applicable */
   const jobTitle = Player.jobs[props.companyName] ? Player.jobs[props.companyName] : null;
-  const hasMoreJobs = Object.keys(Player.jobs).length > 1;
 
   /**
    * CompanyPosition object for the job that the player holds at this company, if applicable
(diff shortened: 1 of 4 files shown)


d0sboots:
Why are some of those companies doing scrambled text in your screenshot? That might be why it has bad performance.

catloversg:
That's `CorruptableText`. We show that text effect when the player installs a backdoor on the company's server. The lag is perceivable on my machine, but it's not too bad.

https://github.com/user-attachments/assets/cd737f4d-4479-4f69-863c-d6ab4ad170d2

Commenting out the usage of `CorruptableText` does not help much. I don't see any perceivable change.

PR #1951, opened 2025-02-06
MISC: Improve stack trace of errors in transformed scripts

### LINKED ISSUE #2098: Wrong line numbers of errors in typescript files
Bitburner Version 2.8.1

The bitburner in-game script error messages shows from line numbers for ts-files.
Please note the differences of the intern message and the debug console error message:
(line 6 vs line 30)

Image

for example with this simple test script:

```
// import { NS } from "NetscriptDefinitions";

import{
  

}from "utils2"


function test():void{


  
}



export async function main(ns:NS) {










  let t:string = undefined;

  let l = t.length;
}
```

catloversg:
With errors in JSX, TS, TSX scripts, we are showing the stack trace of the transformed scripts. This PR uses the generated source map to get the original error position. For more information, please check comments of d0sboots in #1812. I decided to parse the stack trace and recover the error position instead of only showing a custom file extension (i.e., `.transformedJs`).

Note: I choose [`source-map-js`](https://www.npmjs.com/package/source-map-js) instead of [`source-map`](https://www.npmjs.com/package/source-map). Please check README of `source-map-js` for more information. Async APIs of `source-map`'s v7 are not suitable for us.

Test save file: 
[test stack trace.gz](https://github.com/user-attachments/files/18693620/test.stack.trace.gz)

Before:

before1
before2
before3

After:

after1
after2
after3

Closes #2098.


src/utils/ErrorHandler.ts
```diff
@@ -24,14 +26,8 @@ export function handleUnknownError(e: unknown, ws: WorkerScript | null = null, i
     if (ws) {
       console.error(`An error was thrown in your script. Hostname: ${ws.hostname}, script name: ${ws.name}.`);
     }
-    /**
-     * If e is an instance of Error, we print it to the console. This is especially useful when debugging a TypeScript
-     * script. The stack trace in the error popup contains only the trace of the transpiled code. Even with a source
-     * map, parsing it to get the relevant info from the original TypeScript file is complicated. The built-in developer
```
d0sboots, line 30: Most of this comment is no longer relevant. However, in general we shouldn't be logging things to console, so I think *some* explanation is still warranted - specifically the bit about how the links are still helpful with the built-in developer tools. (Without that, I don't think it would be worth doing this here.)

src/utils/StackTraceUtils.ts
```diff
@@ -0,0 +1,111 @@
...
+  const sourceMaps = new Map<string, string>();
+  for (const stackFrame of stackFrames) {
+    if (!stackFrame.fileName) {
+      continue;
+    }
+    /**
+     * Filename in the stack line is actually `${hostname}/${filename}`. Check sourceURL in generateLoadedModule
+     * (src\NetscriptJSEvaluator.ts).
+     */
+    if (!stackFrame.fileName.startsWith(workerScript.hostname)) {
```
d0sboots: Can't these mismatch due to module sharing across hosts?

catloversg:
@d0sboots Parsing the stack trace to get the correct filename is a real PITA. While reworking `parseStackTrace`, I found at least 2 bugs in our codebase. It takes a while to troubleshoot them before finishing this PR, so I marked this one as a draft.

d0sboots:
I'm not at all surprised that you shook loose a few edge-cases. Thanks for your continued dedication.

codemage:
Not sure if it would work either, but if anyone is running into this before this pull lands, https://github.com/rmuratov/sourcemap.tools is a tool that appears to be designed to do this by hand, provided you can extract all the relevant sourcemap data to feed it.

(This is not a suggestion of something to use as part of the actual game, just an idea for any user wanting to apply a bit of elbow grease. If anyone tries it please report back on how you got the sourcemaps out and whether it worked at all.)

d0sboots:
@codemage Errors are also logged to the browser console, where the browser itself will perform the transformation. This requires opening the console, but it's a lot easier than invoking a third-party tool.

PR #1962, opened 2025-02-14
MISC: Clarify advanced options

catloversg:
Context: https://discord.com/channels/415207508303544321/415213435974975508/1336687460444540928

Screenshot:
Capture


LJNeon:
Instead of repeating the same information over and over in every tooltip, it might be better to include it in the paragraph above or some other central location?

catloversg:
Let's see what @d0sboots thinks.

d0sboots:
It's seemingly redundant, but sometimes that's what you need when dealing with players, especially those who skip the explainer paragraph.

PR #1965, opened 2025-02-15
DOCUMENTATION: Clarify ns.scan

catloversg:
With `ns.scan`, there is some common knowledge among experienced players:
- The parent node is always the first item of the returned array.
- Use BFS and DFS to discover all servers.

This kind of knowledge is essential for effectively using this API, but it's not obvious for new players, especially the first one. This PR adds that information to TSDoc of `ns.scan`.

### DIFF
--- a/src/ScriptEditor/NetscriptDefinitions.d.ts
+++ b/src/ScriptEditor/NetscriptDefinitions.d.ts
@@ -6711,7 +6711,7 @@ export interface NS {
    * ------omega-net
    * ```
    *
-   * ns.scan("home"): ["n00dles", "joesguns"]: "home" is the root, so it does have a parent node.
+   * ns.scan("home"): ["n00dles", "joesguns"]: "home" is the root, so it does not have a parent node.
    *
    * ns.scan("n00dles"): ["home"]: "home" is the parent node of "n00dles".
    *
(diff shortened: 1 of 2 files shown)


LJNeon:
That seems like a lot of text to add to a function in the definitions file, I would've thought a shorter version would fit better there and the extended version can be in the full documentation?

catloversg:
The added information is not really long. It looks long because of the example. The part about the parent node definitely should be in TSDoc. It's the undocumented behavior of this API. The hint about BFS and DFS is not long enough to move to a different page.

LJNeon:
~~There are no other comments in the definitions file with even as many as 150 characters... this one approaches 1,500 characters. That is definitely what I would quantity as "really long".~~

Edit: I was wrong about that 150 figure, but I still think this one is too long.

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -6698,6 +6698,33 @@ export interface NS {
...
+   *
+   * ```
+   * home
+   * --n00dles
+   * --joesguns
+   * ----CSEC
+   * ------omega-net
+   * ```
+   *
+   * ns.scan("home"): ["n00dles", "joesguns"]: "home" is the root, so it does have a parent node.
```
```suggestion
   * ns.scan("home"): ["n00dles", "joesguns"]: "home" is the root, so it doesn't have a parent node.
```

d0sboots:
I'm not concerned about the length.

PR #1968, opened 2025-02-15
UI: Warn player if they are editing and saving files on non-home servers

catloversg:
Saving files on non-home servers is a common newbie mistake. This PR adds many UI hints to warn them if they are doing that.
- Blink animation.
- Change color (info -> warning).
- Warning toast message.

Capture

https://github.com/user-attachments/assets/2a08d286-ce5f-457c-919e-f353d04185c1


d0sboots:
I think the blink is probably too much. I like all the other bits though. We're going from basically no hinting to "lots of hinting" so I worry about overcorrecting... we can bring the blink back in if people are still being confused.

PR #1969, opened 2025-02-15
DOCUMENTATION: Clarify ns.nuke

catloversg:
d0sboots raised this concern on Discord: https://discord.com/channels/415207508303544321/415213435974975508/1337155037255958609.

```
The other thing that is often confusing is that you don't need to meet the hacking level of the server to run NUKE (or port openers), but do to run hack.
```

This PR clarified the requirement of nuking.


LJNeon:
It's not quite as bad, but I have the same issue with this PR as the one I have with #1965.

PR #1972, opened 2025-02-20
UI: Add disambiguation to the confusing "1s / ls" tutorial step

ficocelliguy:
There are a significant number of confused new players getting caught in the tutorial confusing "ls" as "1s":
https://discord.com/channels/415207508303544321/415213435974975508/1342166697716613200

This PR adds some context to the prompt in the tutorial that "ls" is short for "list", and an explicit error message for the "1s" case explaining it is spelled with an L :

image

When the user is not in the tutorial: if an invalid command is entered, and it is very similar, or a subset, of a real command, some real commands are listed as a "did you mean?" (maximum 3 suggestions)

image

### DIFF
--- a/src/Terminal/Terminal.ts
+++ b/src/Terminal/Terminal.ts
@@ -656,6 +656,9 @@ export class Terminal {
         case iTutorialSteps.TerminalLs:
           if (commandArray.length === 1 && commandArray[0] == "ls") {
             iTutorialNextStep();
+          } else if (commandArray[0] == "1s") {
+            this.error("Command '1s' not found. Did you mean 'ls' with a lowercase L?");
+            return;
           } else {
             this.error(errorMessageForBadCommand);
             return;
@@ -800,7 +803,11 @@ export class Terminal {
     commandArray.shift();
 
     const f = TerminalCommands[commandName.toLowerCase()];
-    if (!f) return this.error(`Command ${commandName} not found`);
+    if (!f) {
+      const similarCommands = findSimilarCommands(commandName);
+      const didYouMeanString = similarCommands.length ? `Did you mean: ${similarCommands.join(" or ")}?` : "";
+      return this.error(`Command ${commandName} not found. ${didYouMeanString}`);
+    }
 
     f(commandArray, currentServer);
   }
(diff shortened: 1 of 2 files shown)


Alpheus (approved):

src/Terminal/Terminal.ts
```diff
@@ -800,7 +803,11 @@ export class Terminal {
     commandArray.shift();
 
     const f = TerminalCommands[commandName.toLowerCase()];
-    if (!f) return this.error(`Command ${commandName} not found`);
+    if (!f) {
+      const similarCommands = findSimilarCommands(commandName);
+      const didYouMeanString = similarCommands.length ? `Did you mean: ${similarCommands.join(" or ")}?` : "";
+      return this.error(`Command ${commandName} not found. ${didYouMeanString}`);
```
d0sboots:
```suggestion
      const didYouMeanString = similarCommands.length ? ` Did you mean: ${similarCommands.join(" or ")}?` : "";
      return this.error(`Command ${commandName} not found.${didYouMeanString}`);
```

src/Terminal/Terminal.ts
```diff
@@ -813,3 +820,16 @@ export class Terminal {
     });
   }
 }
+
+function findSimilarCommands(command: string): string[] {
+  const commands = Object.keys(TerminalCommands);
+  const offByOneLetter = commands.filter((c) => {
+    let diff = 0;
```
d0sboots, line 828: I think you need an early false return here if c.length !== command.length. Otherwise, checking something like `i` will be true for most commands, because it's off on the single (only) letter, so diff will be 1.

ficocelliguy: I added that check.

Because the suggestions are limited to 3, it doesn't clutter the terminal or anything. It might even help discovery of some of the less common commands?
image

d0sboots: It's fine for those to be picked up by the subset check (and they will), where they are sorted by length. The issue was them being picked up in the wrong place. 

As another example, "wc" should not diff-match "wget".

PR #1975, opened 2025-02-26
CODEBASE: Update webpack and katex

catloversg:
This PR fixes 2 problems with our dependencies:
- We are using a very old version of TypeScript. TSC reports wrong errors at https://github.com/bitburner-official/bitburner-src/actions/runs/13545961255/job/37857521062.
- npm reports a vulnerability in `katex`. I fixed it by running `npm audit fix`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -98,7 +98,7 @@
         "rehype-mathjax": "^4.0.3",
         "remark-math": "^5.1.1",
         "style-loader": "^3.3.3",
-        "typescript": "^5.2.2",
+        "typescript": "^5.7.3",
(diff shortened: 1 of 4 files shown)


package-lock.json
```diff
@@ -18110,10 +18124,11 @@
       }
     },
     "node_modules/typescript": {
-      "version": "5.4.2",
```
d0sboots, line 18113: Intriguingly, we were already at 5.4.2 (not 5.2.2 as you might think from package.json) because of api-extractor. It looks like that's using a pinned version of TS? It might be worth looking at upgrading it in the future too.

PR #1978, opened 2025-02-27
DOCUMENTATION: Clarify value of gang territory

catloversg:
Context: https://discord.com/channels/415207508303544321/415207923506216971/1341396121578111088
```
for ns.gang.getGangInfo().territory, does it go from 0 to 1 or 0 to 100?
```
We already said that `GangOtherInfoObject.territory` is in decimal form. This PR clarifies `GangGenInfo.territory`.

### DIFF
--- a/markdown/bitburner.ganggeninfo.md
+++ b/markdown/bitburner.ganggeninfo.md
@@ -23,7 +23,7 @@ interface GangGenInfo
 |  [respect](./bitburner.ganggeninfo.respect.md) |  | number | Gang's respect |
 |  [respectForNextRecruit](./bitburner.ganggeninfo.respectfornextrecruit.md) |  | number | Amount of Respect needed for next gang recruit, if possible |
 |  [respectGainRate](./bitburner.ganggeninfo.respectgainrate.md) |  | number | Respect earned per game cycle |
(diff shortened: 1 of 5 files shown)


d0sboots:
To me, this doesn't really clarify anything: all numbers (that are base 10) are in decimal form, and I had to read the PR description to understand what this was about.

I'd suggest either "as a fraction" or the maximally explicit "in the range 0-1"

catloversg:
I was bamboozled by how we use the word "decimal" in the TSDoc of other APIs ("decimal form, not percentage", "decimal form", etc.). In many places, we use "decimal" with the meaning of "0-1, not percentage (0-100)". Do we need to change them?

d0sboots:
> I was bamboozled by how we use the word "decimal" in the TSDoc of other APIs ("decimal form, not percentage", "decimal form", etc.). In many places, we use "decimal" with the meaning of "0-1, not percentage (0-100)". Do we need to change them?

"decimal form, not percentage" while IMO not ideal, at least mentions percentage so you can work out what is meant. Without any qualifier, I think it is just confusing. "in the range 0-1" is pretty much the same length and totally unambiguous. 

Long-winded way of saying yes, I think it's worth changing.

PR #1985, opened 2025-02-28
BUGFIX: Some calculations return NaN due to Player.playtimeSinceLastAug being 0 in edge cases

catloversg:
How to reproduce this bug:
- Expose `saveObject` in `src\SaveObject.ts`.
- Run `run a.js 10000`.
- Check console after the game is reloaded. You will see `NaN passed into Player.gainMoney()`.

Note: You may need to tweak `ns.asleep(100)` in `b.js` or run the test a few times.

This bug happens when the player resets (soft reset or install augmentations), saves game, and reloads *before* the next engine tick. It's almost impossible to happen in normal situations, but it still happens if the player rapidly spawns scripts that call the softReset API.

Note that I only expose `saveObject` for easy testing. It's possible to reproduce this bug even without doing that; it's harder and more inconsistent, though.

Save file: [NaN bug.gz](https://github.com/user-attachments/files/19029997/NaN.bug.gz)

### DIFF
--- a/src/NetscriptFunctions.ts
+++ b/src/NetscriptFunctions.ts
@@ -1564,7 +1564,11 @@ export const ns: InternalAPI<NSFull> = {
       total += script.scriptRef.onlineMoneyMade / script.scriptRef.onlineRunningTime;
     }
 
-    return [total, Player.scriptProdSinceLastAug / (Player.playtimeSinceLastAug / 1000)];
+    let incomeFromScriptsSinceLastAug = Player.scriptProdSinceLastAug / (Player.playtimeSinceLastAug / 1000);
+    if (Number.isFinite(incomeFromScriptsSinceLastAug)) {
+      incomeFromScriptsSinceLastAug = 0;
+    }
+    return [total, incomeFromScriptsSinceLastAug];
   },
   getScriptIncome:
     (ctx) =>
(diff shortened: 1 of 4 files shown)


src/NetscriptFunctions.ts
```diff
@@ -1564,7 +1564,11 @@ export const ns: InternalAPI<NSFull> = {
       total += script.scriptRef.onlineMoneyMade / script.scriptRef.onlineRunningTime;
     }
 
-    return [total, Player.scriptProdSinceLastAug / (Player.playtimeSinceLastAug / 1000)];
+    let incomeFromScriptsSinceLastAug = Player.scriptProdSinceLastAug / (Player.playtimeSinceLastAug / 1000);
+    if (Number.isFinite(incomeFromScriptsSinceLastAug)) {
```
d0sboots: I think the test is reversed here.

src/engine.tsx
```diff
@@ -278,7 +278,9 @@ const Engine: {
       let offlineReputation = 0;
       const offlineHackingIncome =
         (Player.moneySourceA.hacking / Player.playtimeSinceLastAug) * timeOffline * CONSTANTS.OfflineHackingIncome;
-      Player.gainMoney(offlineHackingIncome, "hacking");
+      if (Number.isFinite(offlineHackingIncome)) {
+        Player.gainMoney(offlineHackingIncome, "hacking");
```
d0sboots: I'd prefer the same fix here as elsewhere, in case someone decides to start using offlineHackingIncome later and subtly reintroduces the bug.

d0sboots:
Aha, this gets to the bottom of people who were corrupting their game with automated save-scumming in the casino. Nice job.

PR #1987, opened 2025-03-01
UI: Add "Run" action to run current script in editor

catloversg:
This PR adds a new action that runs the current script in the editor. Behaviors:
- Always save before running.
- ~Only run if the player has root access; otherwise, show an error popup.~
- ~Connect to the current script's server if needed.~
- Call `createRunningScriptInstance` and `startWorkerScript` to run the current script.

Capture1

Capture2

Capture3

### DIFF
--- a/src/NetscriptWorker.ts
+++ b/src/NetscriptWorker.ts
@@ -39,6 +39,7 @@ import { isLegacyScript, legacyScriptExtension, resolveScriptFilePath, ScriptFil
 import { root } from "./Paths/Directory";
 import { getErrorMessageWithStackAndCause } from "./utils/ErrorHelper";
 import { exceptionAlert } from "./utils/helpers/exceptionAlert";
+import { Result } from "./types";
 
 export const NetscriptPorts = new Map<PortNumber, Port>();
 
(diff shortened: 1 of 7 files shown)


src/ScriptEditor/ui/ScriptEditorRoot.tsx
```diff
@@ -219,6 +221,32 @@ function Root(props: IProps): React.ReactElement {
...
+    // Always save before doing anything else.
+    save();
+
+    // Check if the player can run scripts on this server.
+    if (!server.hasAdminRights) {
+      dialogBoxCreate(`You do not have root access on ${server.hostname} server.`);
+      return;
+    }
+    // Connect to the current script's server if needed.
+    if (Player.currentServer !== currentScript.hostname) {
```
d0sboots: This part strikes me as weird. We can rerun scripts from tail windows without doing this. Also, it's a bit surprising to be suddenly connected to a different server.

src/utils/KeyBindingUtils.ts
```diff
@@ -252,6 +253,16 @@ export const DefaultKeyBindings: Record<KeyBindingType, [KeyCombination | null,
       key: "B",
     },
   ],
+  [ScriptEditorAction.Run]: [
+    {
+      control: true,
+      alt: false,
+      shift: false,
+      meta: false,
+      key: "Q",
```
d0sboots, line 262: I looked to see if there is any kind of standard here. Curiously, most IDEs (IDEA, Eclipse, VS) have settled on some form of function key for run, but they don't agree on which one. (F5, F10 and F11 are all used.) So, I think this is probably fine.

src/ScriptEditor/ui/ScriptEditorRoot.tsx
catloversg: When starting a script, we need to call `startWorkerScript` with a `RunningScript` instance. Rerunning a script from the tail window is possible because each `LogWindow` (`LogBoxManager.tsx`) already has a `RunningScript` instance (there is an unrelated bug in `LogBoxManager.tsx`, I'll fix them later).

In this case, we don't have a `RunningScript` instance, and creating a new one is messy (for example, `runScript` in `src\Terminal\commands\runScript.ts`). It's not too hard, but duplicating that code (get a `Script`, then calculate RAM usage) is not really a good way to start a script here. I don't see any easier way to do that than reusing `Terminal.executeCommand()` like this. And in order to run a command like that, we need to connect to the proper server.

I'm open to any alternate solutions.

Edit: Duplicating required code in `runScript.ts` is not ideal, but I'm not against it if you choose that way.

d0sboots: I remember looking and being annoyed at the duplication between `runScript` in Terminal and what underlies `exec` in NetscriptCommands. There's a lot of duplicated checking that's done in slightly different ways, but because Terminal and NetscriptHelpers both have their own ways of dealing with errors it was messy trying to unify them.

If you could unify them, that would be ideal, but that's a really tall order. I'd be happy if you chose one of them and pulled out a common helper that you could also then use for this, so that we didn't need to duplicate more code but also don't need to change servers.

Alpheus: Can't you use the runScript with RunningScript of an already existing process? If a player is automating something long-lived, it may be equivalent in usage to stop-and-run in a tail window, defaulting to a fresh one only if one isn't running (with default args)

catloversg: No. There may not be any RunningScript instances. Even if there is, it still makes no sense to piggyback a running RunningScript and pretend that this action is the same as the one creating that RunningScript. Doing that may also create unforeseen problems. Refactoring like d0sboots said is the best solution.

Edit: To give an example of the "unforeseen problems". Every time a script is run (via the terminal or NS APIs), we need to create a new RunningScript. That instance has many tracking data that are unique for each running script. LogWindow tried to bypass this requirement by resetting the data of its parameter RunningScript. This is a hack, and that's why it created the unrelated bug that I said above.

src/NetscriptWorker.ts
```diff
@@ -451,18 +452,65 @@ export function loadAllRunningScripts(): void {
...
+  server: BaseServer,
+  scriptPath: ScriptFilePath,
   args: ScriptArg[],
   workerScript: WorkerScript,
   runOpts: CompleteRunOptions,
 ): number {
-  const script = host.scripts.get(scriptname);
-  if (!script) {
-    workerScript.log(caller, () => `Could not find script '${scriptname}' on '${host.hostname}'`);
+  const result = createRunningScriptInstance(server, scriptPath, runOpts.ramOverride, runOpts.threads, args);
```
d0sboots, line 512:
```suggestion
  // This does not adjust server RAM usage or change any state, so it is safe to call before performing other checks
  const result = createRunningScriptInstance(server, scriptPath, runOpts.ramOverride, runOpts.threads, args);
```

src/ScriptEditor/ui/ScriptEditorRoot.tsx
```diff
@@ -239,12 +243,13 @@ function Root(props: IProps): React.ReactElement {
       dialogBoxCreate(`You do not have root access on ${server.hostname} server.`);
       return;
     }
```
d0sboots: Redundant with the refactor

d0sboots (commented):
Nice, you managed to refactor them together! `Result` keeps paying dividends.

PR #1988, opened 2025-03-01
CODEBASE: Refactor duplicate code in updateSkillLevels

d0sboots:
I've been sitting on this for a while, I should either submit it or delete it.

@catloversg please let me know if this seems like a good refactor to you. Note that the strings here *are* typechecked by TS.


catloversg:
It looks good to me.

Alpheus (approved):

Alpheus:
Found no fault either. Good to go!

PR #1995, opened 2025-03-05
IPVGO: correctly initialize board from save when there are no prior moves

ficocelliguy:
The prior move is saved as an empty string if it does not exist, to save on filespace and avoid null serializing. This PR ensures that it is not treated as a valid previous move, in that scenario, when loaded.

This PR also fixes a minor bug where defending moves were not properly checked for the purposes of knowing when to try to pattern match.

### DIFF
--- a/src/Go/boardAnalysis/goAI.ts
+++ b/src/Go/boardAnalysis/goAI.ts
@@ -403,7 +403,9 @@ async function getIlluminatiPriorityMove(moves: MoveOptions, rng: number): Promi
     return moves.corner()?.point ?? null;
   }
 
-  const hasMoves = [moves.eyeMove(), moves.eyeBlock(), moves.growth(), moves.defend(), surround].filter((m) => m).length;
+  const hasMoves = [moves.eyeMove(), moves.eyeBlock(), moves.growth(), moves.defend(), surround].filter(
+    (m) => m,
+  ).length;
   const usePattern = rng > 0.25 || !hasMoves;
 
   if ((await moves.pattern()) && usePattern) {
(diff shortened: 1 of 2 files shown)


src/Go/boardAnalysis/goAI.ts
```diff
@@ -403,7 +403,9 @@ async function getIlluminatiPriorityMove(moves: MoveOptions, rng: number): Promi
     return moves.corner()?.point ?? null;
   }
 
-  const hasMoves = [moves.eyeMove(), moves.eyeBlock(), moves.growth(), moves.defend, surround].filter((m) => m).length;
+  const hasMoves = [moves.eyeMove(), moves.eyeBlock(), moves.growth(), moves.defend(), surround].filter(
+    (m) => m,
```
d0sboots, line 407: This change is extremely confusing. I stared at it for 5 minutes trying to figure out what it did. Eventually I consulted <https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Trailing_commas#function_calls> and determined that it does nothing.

src/Go/boardAnalysis/goAI.ts
```diff
@@ -403,7 +403,9 @@ async function getIlluminatiPriorityMove(moves: MoveOptions, rng: number): Promi
     return moves.corner()?.point ?? null;
   }
 
-  const hasMoves = [moves.eyeMove(), moves.eyeBlock(), moves.growth(), moves.defend, surround].filter((m) => m).length;
+  const hasMoves = [moves.eyeMove(), moves.eyeBlock(), moves.growth(), moves.defend(), surround].filter(
```
d0sboots, line 406:
```suggestion
  const hasMoves = [moves.growth(), moves.defend(), surround].some(Boolean);
```
`some` is the correct function to use here, and `Boolean` is the right function to check for truthiness. `eyeMove` and `eyeBlock` must be false, since if they were true we would have already returned.

d0sboots:
The comments are good-to-have for style, but submitting this as-is since it's fixing a critical bug.

PR #1996, opened 2025-03-05
IPVGO: Add support for highlighting nodes and adding small text

ficocelliguy:
Adds new API methods to go.analytics to allow for adding colored indicators and/or text to nodes by player scripts.

Can be tested at https://ficocelliguy.github.io/bitburner-src/

image

My testing script that makes it rave mode:
```js
/** @param {NS} ns */

  const boardSize = ns.go.getBoardState().length;

  while(true) {
    await ns.sleep();
    const x = (Math.random() * boardSize) | 0;
    const y = (Math.random() * boardSize) | 0;
    let color = "#";
    for (let j = 0; j < 6; j++) {
      color += [0,1,2,3,4,5,6,7,8,9,"a","b","c","d","e","f"][Math.random() * 16 | 0];
    }
    ns.go.analysis.highlightPoint(x, y, color, color);
  }

}
```


paulcdejean:
I almost want to say skip the text arg, and just have it support colors tbh. But maybe that's just me being mean.

ficocelliguy:
The text arg is for adding a point value or relative score or whatever other label, as seen in many of the move highlighters/visualizers players have homebrewed

ficocelliguy:
@d0sboots Is there anything more you need for this one? a couple people have asked about the feature's status

d0sboots:
the status is "i should feel bad"

src/Go/boardState/boardState.ts
```diff
@@ -122,6 +123,7 @@ export function makeMove(boardState: BoardState, x: number, y: number, player: G
 
   // Add move to board history
   boardState.previousBoards.unshift(boardStringFromBoard(boardState.board));
+  clearAllPointHighlights();
```
d0sboots: This isn't in the right place. `makeMove` can be applied to any `BoardState`, whereas highlights are scoped to the global go object.

Either that, or really highlights should be part of BoardState and just not saved with it. I could see either way, but the object model right now doesn't work right.

src/Go/boardState/goStyles.ts
```diff
@@ -29,31 +37,56 @@ export const pointStyle = makeStyles<unknown, Size | Point | Structure | Highlig
...
+          display: "block",
+        },
       [`&.${classes.hover}:hover .${classes.coordinates}`]: {
         display: "block",
       },
       [`&:hover .${classes.broken}`]: {
         opacity: "0.4",
       },
+
+      [`&.${classes.hack} .${classes.innerPoint}`]: {
```
d0sboots, line 75: What is this for?

Edit: I see what it is for. IMO, this is needless complexity (RGB is likely all that is wanted/needed), but if you have any evidence/knowledge to the contrary, I'm OK leaving it in.

src/Go/effects/netscriptGoImplementation.ts
```diff
@@ -424,6 +425,21 @@ export function validateBoardState(
   }
 }
 
+export function addPointHighlight(x: number, y: number, color: string, text: string) {
+  Go.highlightedPoints[x][y] = { color, text };
+  GoEvents.emit();
+}
+
+export function clearPointHighlight(x: number, y: number) {
+  Go.highlightedPoints[x][y] = { color: "", text: "" };
```
d0sboots: Shouldn't this set to `null`?

src/Go/ui/GoGameboard.tsx
```diff
@@ -7,15 +7,23 @@ import { GoOpponent, GoColor } from "@enums";
...
 import { boardStyles } from "../boardState/goStyles";
 import { getAllValidMoves, getControlledSpace } from "../boardAnalysis/boardAnalysis";
+import { PointHighlight } from "../Go";
 
 interface GoGameboardProps {
   boardState: BoardState;
   traditional: boolean;
   clickHandler: (x: number, y: number) => unknown;
   hover: boolean;
+  highlightedPoints?: (PointHighlight | null)[][];
```
d0sboots: Why is this prop optional?

src/Go/ui/GoPoint.tsx
```diff
@@ -15,11 +16,29 @@ interface GoPointProps {
   traditional: boolean;
   hover: boolean;
   valid: boolean;
-  emptyPointOwner: GoColor;
+  emptyPointOwner?: GoColor;
+  pointHighlight?: PointHighlight | null;
```
d0sboots: Since this can be null, there's no reason for it to be optional.

src/Go/ui/GoPoint.tsx
```diff
@@ -15,11 +16,29 @@ interface GoPointProps {
   traditional: boolean;
   hover: boolean;
   valid: boolean;
-  emptyPointOwner: GoColor;
+  emptyPointOwner?: GoColor;
```
d0sboots: This doesn't seem like it should be optional either. If needed, it can be `| undefined`.

src/Go/boardState/goStyles.ts
ficocelliguy: I wanted to add this so that players can 1) easily match their theme, or 2) slap in some testing values right away without looking up RGB colors. Originally I was going to do only this, since theme-coloring is so much a part of the game's look and feel, but then I added the rgb support later for flexibility.

src/Go/boardState/boardState.ts
ficocelliguy: I moved to this to be a property of the boardstate. It makes more sense there.

d0sboots: You've still got the same issue. `clearAllPointHighlights` is a method on NSGoImpl, which is hardcoded to use Go.currentGame.boardState. Functions within `BoardState` need to modify their *own* BoardState.

ficocelliguy: I moved this to be part of the Go object, and they interact with their own this.currentGame object. Is that what you were thinking?

src/Go/boardState/boardState.ts
```diff
@@ -125,7 +125,7 @@ export function makeMove(boardState: BoardState, x: number, y: number, player: G
 
   // Add move to board history
   boardState.previousBoards.unshift(boardStringFromBoard(boardState.board));
-  clearAllPointHighlights();
+  Go.clearAllPointHighlights();
```
d0sboots: No, now you went back to the other issue :(

The fundamental problem here is a muddying of the waters between boardState and the top-level state. The top-level state is a singleton which contains a single boardState, but this doesn't mean that boardState itself is a singleton. It is entirely possible for multiple boardStates to exist simultaneously, and indeed I believe this occurs when the analysis functions get involved with user-provided boards.

You can view the highlight info as either being a UI thing (thus belonging to the top-level state) or as belonging to boardState; both are valid perspectives. But if it's the former, then you can't change the UI just because a move was made on *any* board, it has to be only for moves made on the board of the top-level state. Conversely, if it belongs to boardState then clearAllPointHighlights *also* needs to belong to boardState. (It's OK for it to trigger `GoEvents.emit` even for boardStates that aren't being shown, since that's idempotent.)

ficocelliguy: I have refactored this so that the highlighting methods are standalone utilities under the analysis file, and they take a boardstate to add or remove highlights from. When a move is made or passed on a board, the highlights for that specific boar is cleared. 

The netscript api access for these utilities applies them to the current ui by passing the current board state to these utilities.

PR #1998, opened 2025-03-06
MISC: Change how enums are exposed in NetscriptDefinitions.d.ts

catloversg:
We expose enums with `declare enum` for typing (technically, the reason is more complicated, but let's simplify it here) and `ns.enums` for runtime data. Some players assume that they can access those enums at runtime by using those declared enums, but those enums do not exist at runtime like they think. For example: #1198. We can resolve this problem by changing how we expose enums to players in that file.

At first, there are tons of discussions about TS enum and its alternatives on the Discord server and everywhere on the Internet. At least, please read ⁠[Enums](https://discord.com/channels/415207508303544321/1086078798988050623/1086078801592725554) and some old discussions on the Discord server (search "enum" and check "Old").

Scope of this change:
- Only touch `declare enum` in `NetscriptDefinitions.d.ts`.
- No corporation or other stuff.
- Minimal change or no change in internal code.

Things to remember:
- `ns.enums.Something` must have proper types and work as normal.
- Shoud not need the old trick `LocationName | ${LocationName}`.
- Must be able to be extracted by `api-extractor`.

```js
  // Must be ok and support autocomplete
  ns.singularity.travelToCity("Sector-12");
  ns.singularity.travelToCity(ns.enums.CityName.Sector12);

  // Must be ok
  const city1: CityName = ns.enums.CityName.Sector12;
  ns.singularity.travelToCity(city1);

  // Must be ok
  const city2: CityName = "Sector-12";
  ns.singularity.travelToCity(city2);

  // Must be an error
  // const city3: CityName = "Sector-13";
  // console.log(city3);

  // Hover "city". It must show `const city: "Aevum" | "Chongqing" | "Ishima" | "Volhaven" | "Sector-12" | "New Tokyo"`
  for (const city of Object.values(ns.enums.CityName)) {
    // Must be ok
    ns.singularity.travelToCity(city);
  }
}
```

This PR won't touch internal enums. Changing all internal enums to `as const` objects is a bag of worms. It needs to be done in another PR.

This PR is in the draft state now. After we agree on how to do it, I'll expand the change to other enums and run `npm run doc`.


src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -8437,16 +8437,20 @@ type CorpIndustryName =
 /** @public */
 type CorpSmartSupplyOption = "leftovers" | "imports" | "none";
 
+type ValueOf<T> = T[keyof T];
```
d0sboots: Since it turns out that everything here is exported regardless of `export`, should we prefix stuff like this that is meant to be internal with `_`?

catloversg:
@d0sboots Just in case you missed the last updates, I updated this PR based on your feedback.

d0sboots:
Looks good. Let's do it.

PR #1999, opened 2025-03-06
NETSCRIPT:  Add the 'cat' method to 'ns.singularity'

NagaOuroboros:
## Motivation
While `.lit` files are easy to copy onto the 'home' server, doing so clutters up the root directory, and will only create more clutter as new `.lit` files get added in the future.  As a first step to managing this, I've implemented the functionality of the `cat` terminal command as a method available on the `ns.singularity` namespace.

This addition should let players write their own 'remote file viewer' scripts once they've gotten SF4.

## Scope
This PR only adds the new method itself, which is simply a 1:1 feature-for-feature implementation of the terminal equivalent.  No new behaviors are added.

## New API Method
`ns.singularity.cat(filename)` - Displays the content of a file on the currently connected server

## Considerations
- The RAM cost of the method is currently set at 0.5 before the SF4 multipliers.  I would argue for lower, perhaps even 0.1, but I feel 0.5 is a reasonable upper limit.  There is no gameplay advantage to this method—it is purely a quality of life addition—so I do not feel it deserves to be any higher, especially as it will often be used in conjunction with `ns.singularity.connect`.


src/NetscriptFunctions/Singularity.ts
```diff
@@ -467,6 +473,27 @@ export function NetscriptSingularity(): InternalAPI<ISingularity> {
       helpers.checkSingularityAccess(ctx);
       return Player.getCurrentServer().hostname;
     },
+    cat: (ctx) => (_filename) => {
+      const path = helpers.filePath(ctx, "filename", _filename);
+      const server = helpers.getServer(ctx, Player.currentServer);
+      if (hasScriptExtension(path) || hasTextExtension(path)) {
```
catloversg: I think you can reuse a large part of `src\Terminal\commands\cat.ts` instead of duplicating them here.

NagaOuroboros: You mean directly use the exported `cat` function?

catloversg: Either that or pull common code out to a new function, depending on how different these functions are. For example:
```js
// src/Terminal/commands/cat.ts

export function commonCode(something) {}

export function cat(args: (string | number | boolean)[], server: BaseServer): void {
  someChecks1();
  commonCode(something);
}

// src/NetscriptFunctions/Singularity.ts
import { commonCode } from "/path/cat.ts";
    cat: (ctx) => (_filename) => {
      someChecks2();
      commonCode(something);
    }
```
It's okay if you reuse `cat` in `cat.ts` (if it's possible).

NagaOuroboros:
## Refactor
Changed the implementation to simply hook directly into the terminal command's logic.  Also added the missing check for the player owning SF4.

PR #2004, opened 2025-03-06
MISC: Add support for getting the save file through the RFA

G4mingJon4s:
Currently, you can only programmatically get the save file through the NS API. While this does allow for automatic backups, getting those backups into your external editor can be quite cumbersome. Now, the current save file is exposed to the RFA via a new request.

I'm not quite sure if it should also be possible to import a save file via the RFA. It would be quite frustrating to have your save file be lost because your external editor overwrites it. Maybe a middle ground would be a confirmation popup? This feature seems quite niche IMO, so maybe it is not needed?

### DIFF
--- a/src/RemoteFileAPI/MessageHandlers.ts
+++ b/src/RemoteFileAPI/MessageHandlers.ts
@@ -13,12 +13,14 @@ import {
 } from "./MessageDefinitions";
 
 import libSource from "../ScriptEditor/NetscriptDefinitions.d.ts?raw";
+import { saveObject } from "../SaveObject";
+import { Player } from "@player";
 
 function error(errorMsg: string, { id }: RFAMessage): RFAMessage {
   return new RFAMessage({ error: errorMsg, id: id });
 }
 
-export const RFARequestHandler: Record<string, (message: RFAMessage) => RFAMessage> = {
+export const RFARequestHandler: Record<string, (message: RFAMessage) => RFAMessage | Promise<RFAMessage>> = {
   pushFile: function (msg: RFAMessage): RFAMessage {
     if (!isFileData(msg.params)) return error("Misses parameters", msg);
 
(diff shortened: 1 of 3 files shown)


G4mingJon4s:
After some discussion on the discord, I think it's fair to say that importing a save via the RFA is not really desirable, especially loosing your progress due to unintentional imports.
With that out of the way, this PR should be ready to go in terms of functionality.

src/RemoteFileAPI/MessageHandlers.ts
```diff
@@ -112,6 +114,22 @@ export const RFARequestHandler: Record<string, (message: RFAMessage) => RFAMessa
     return new RFAMessage({ result: libSource, id: msg.id });
   },
 
+  getSaveFile: async function (msg: RFAMessage): Promise<RFAMessage> {
+    const saveData = await saveObject.getSaveData();
+    // We can't serialize the Uint8Array directly, so we convert every integer to a character and make a string out of the array
+    // The external editor can simply recreate the save by converting each char back to their char code
+    const converted =
+      typeof saveData === "string" ? saveData : saveData.reduce((acc, cur) => acc + String.fromCharCode(cur), "");
```
d0sboots: So, this is kind of a problem. Within the game, you can use the type of `saveData` to determine if it's the old format or the new compressed format. However, here you are handing the user a string, and they'll have to use unreliable heuristics to determine which type it is.

I can think of three possibilities here:
* Undo the compression and always return the uncompressed-and-base64-encoded form. This is a little sub-optimal because there could be a big memory hit (possibly even leading to game crashes) with big enough save files.
* Add a type field to the result.
* Simply don't support uncompressed savegames.

d0sboots: Use this instead of reduce:
```js
let converted = "";
for (let i = 0; i < saveData.length; i++) {
  converted += String.fromCharCode(saveData[i]);
}
```
Normally I would say use `for...of`, but the speed difference is significant, and in a tight loop operating on lots of bytes like this, it matters.


https://jsbenchmark.com/#eyJjYXNlcyI6W3siaWQiOiJBWWhwdWYteHI2dGNBb2gwWl81ZjgiLCJjb2RlIjoicmV0dXJuIERBVEEucmVkdWNlKChhY2MsIGN1cikgPT4gYWNjICsgU3RyaW5nLmZyb21DaGFyQ29kZShjdXIpLCBcIlwiKTsiLCJuYW1lIjoiUmVkdWNlIiwiZGVwZW5kZW5jaWVzIjpbXX0seyJpZCI6Im5ZUndkUzZNR1lzSUJwLUpZNWtWSyIsImNvZGUiOiJsZXQgYWNjID0gXCJcIjtcbmZvciAobGV0IGkgPSAwOyBpIDwgREFUQS5sZW5ndGg7IGkrKykge1xuICBhY2MgKz0gU3RyaW5nLmZyb21DaGFyQ29kZShEQVRBW2ldKTtcbn1cbnJldHVybiBhY2M7IiwibmFtZSI6Ikxvb3AiLCJkZXBlbmRlbmNpZXMiOltdfV0sImNvbmZpZyI6eyJuYW1lIjoiQmFzaWMgZXhhbXBsZSIsInBhcmFsbGVsIjpmYWxzZSwiZ2xvYmFsVGVzdENvbmZpZyI6eyJkZXBlbmRlbmNpZXMiOltdfSwiZGF0YUNvZGUiOiJyZXR1cm4gVWludDhBcnJheS5mcm9tKEFycmF5KDY0MDAwKSwgKHgsIGkpPT5pJjI1NSk7In19

I tried using `TextEncoder`, and interestingly it is much slower than this approach (and also won't do exactly the same thing), despite being the "native" way.

Issue #2010, opened 2025-03-09
IPVGO: Multiple issues with first game in clean save file


catloversg:
How to reproduce:
- Use "Delete Save" to reload the game with a clean save.
- City -> CIA -> IPvGO Subnet Takeover.
- Play a random move.

Some issues:

Issue 1:
- Do the reproducible steps.
- The board does not show that random move, and the AI does not play the next move.
- Try to play another move. The game shows "It is not your turn to play".

Issue 2:
- Do the reproducible steps.
- Press "Pass turn" twice. The AI wins immediately, but the board shows only 1 piece from your move.

Issue 3:
- Do the reproducible steps.
- The board does not show that random move, and the AI does not play the next move.
- Press "Find new subnet". Do not start a new game. Just close the popup.
- The game shows your last move, but "Pass turn" becomes "Waiting for opponent" and gets stuck forever.

This bug was introduced in #1917. It does not happen in de6b20234104fe911e7a1a9a05fb7bdf0af11f92 (the previous commit).

d0sboots:
Based on my investigation during the fix, this can *only* happen during your initial session on a fresh game. (Or, during your initial session when loading a sufficiently old savegame that is pre-IPvGO.) Once you save and reload the page, everything will work correctly.

As a result, I don't expect too many people to hit this, and those that do will likely just reload and not think anything of it. So I don't think we need another cherrypick.

PR #2011, opened 2025-03-09
IPVGO: First ever game promise handling

ficocelliguy:
There is currently code that handles when move processing is attempted when the AI is already thinking (or a player script is), which gives back the current move promise instead of making a new move.

This caused #2010 because the very first ever game was improperly identified as a scenario such as that.

This PR fixes it by ensuring we have a valid promise to hand to the user before attempting that shortcut.


d0sboots:
Obsoleted by my alternate fix https://github.com/bitburner-official/bitburner-src/pull/2012

The promises should always be initialized, and IMO the bug was that they were not in this narrow circumstance. Letting them be in this bad state by adding handling in `handleNextTurn` scares me, and I suspect would open us up to more bugs down the line.

PR #2019, opened 2025-03-11
MISC: Disable effect of SF7.3 and SF10 if player disables them with advanced options

catloversg:
We always use `Player.activeSourceFileLvl` instead of `Player.sourceFileLvl` to check if we should apply the SF effect. I checked the codebase and found some places that still use `Player.sourceFileLvl`. This PR made changes for SF7.3 and SF10. There are 2 other places that I don't touch in this PR:

https://github.com/bitburner-official/bitburner-src/blob/6666a176eebb9cdc56cb620a130424f7dea51491/src/PersonObjects/Person.ts#L150
Technically, we should change this one as well, but I'm a bit hesitant. @d0sboots What do you think about this?

https://github.com/bitburner-official/bitburner-src/blob/6666a176eebb9cdc56cb620a130424f7dea51491/src/PersonObjects/Sleeve/Sleeve.ts#L562
The number of sleeves should be a permanent effect. It's really troublesome for the player if they lose Sleeves and have to go BN10 to buy them again.

### DIFF
--- a/src/NetscriptFunctions/Sleeve.ts
+++ b/src/NetscriptFunctions/Sleeve.ts
@@ -17,7 +17,7 @@ import { SleeveWorkType } from "../PersonObjects/Sleeve/Work/Work";
 import { canAccessBitNodeFeature } from "../BitNode/BitNodeUtils";
 
 export const checkSleeveAPIAccess = function (ctx: NetscriptContext) {
-  if (Player.bitNodeN !== 10 && !Player.sourceFileLvl(10)) {
+  if (Player.bitNodeN !== 10 && Player.activeSourceFileLvl(10) <= 0) {
     throw helpers.errorMessage(
       ctx,
       "You do not currently have access to the Sleeve API. This is either because you are not in BitNode-10 or because you do not have Source-File 10",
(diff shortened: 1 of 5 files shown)


d0sboots:
> https://github.com/bitburner-official/bitburner-src/blob/6666a176eebb9cdc56cb620a130424f7dea51491/src/PersonObjects/Person.ts#L150
> 
> 
> Technically, we should change this one as well, but I'm a bit hesitant. @d0sboots What do you think about this?

Intelligence gain is a permanent gate IMO. This is expressed in the code with the `intelligence > 0` condition; once you have any, you are eligible to get more. So, changing this would only affect if you beat SF5.1 while somehow avoiding any intelligence gain, and then did a different BN while disabling SF5.1 in the options. In that case, you could prevent future intelligence gain.

That seems like an *incredibly* rare edge case that I don't really wish to expose.

src/NetscriptFunctions/Sleeve.ts
```diff
@@ -17,7 +17,7 @@ import { SleeveWorkType } from "../PersonObjects/Sleeve/Work/Work";
 import { canAccessBitNodeFeature } from "../BitNode/BitNodeUtils";
 
 export const checkSleeveAPIAccess = function (ctx: NetscriptContext) {
-  if (Player.bitNodeN !== 10 && !Player.sourceFileLvl(10)) {
+  if (Player.bitNodeN !== 10 && Player.activeSourceFileLvl(10) <= 0) {
```
d0sboots: This one is weird. How do sleeves work if you have 10 disabled? From the other place, they don't get *destroyed*, but are they disabled?

If you can still use them manually, you should be able to use them via API.

PR #2021, opened 2025-03-11
UI: Make BN-hint popups harder to be dismissed accidentally

catloversg:
After entering a new BN, we sometimes show a popup hinting at new mechanics. This popup can be dismissed accidentally by the player (e.g., clicking on the backdrop) and make them miss the hint. This PR makes it harder to dismiss these special popups. The player needs to press the "x" button or the "OK" button to dismiss the popup.

Another change is the behavior of pressing the "Esc" button. Currently, we dismiss *all* popups. I don't think this is a good behavior. We only show alert popups when it's necessary, so there are not many popups, and they are usually important ones. With the change in this PR, we only dismiss 1 popup when they press the "Esc" button.

Before:

before

After:

after

### DIFF
--- a/src/ui/React/AlertManager.tsx
+++ b/src/ui/React/AlertManager.tsx
@@ -4,31 +4,27 @@ import { Modal } from "./Modal";
 import Typography from "@mui/material/Typography";
 import Box from "@mui/material/Box";
 import { cyrb53 } from "../../utils/HashUtils";
+import Button from "@mui/material/Button";
 
-export const AlertEvents = new EventEmitter<[string | JSX.Element]>();
+export const AlertEvents = new EventEmitter<[string | JSX.Element, boolean?]>();
 
 interface Alert {
   text: string | JSX.Element;
   hash: string;
+  cancellable: boolean;
 }
 
 export function AlertManager({ hidden }: { hidden: boolean }): React.ReactElement {
   const [alerts, setAlerts] = useState<Alert[]>([]);
   useEffect(
     () =>
-      AlertEvents.subscribe((text: string | JSX.Element) => {
+      AlertEvents.subscribe((text: string | JSX.Element, cancellable = true) => {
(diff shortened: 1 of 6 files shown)


d0sboots:
The escape button is an important safety-valve behavior; it is very possible to get into states where there are hundreds of stacked popups that are otherwise impossible to dismiss one-by-one.

src/ui/React/AlertManager.tsx
```diff
@@ -75,10 +73,15 @@ export function AlertManager({ hidden }: { hidden: boolean }): React.ReactElemen
   }
 
   return (
-    <Modal open={!hidden && alerts.length > 0} onClose={close}>
+    <Modal open={!hidden && alerts.length > 0} onClose={close} cancellable={cancellable}>
```
d0sboots: I would call this `hasButton` or `hasOk` or something like that. `cancellable` is weird because *all* alerts are dismissible, so it's not clear what exactly it's doing.

src/ui/React/DialogBox.tsx
```diff
@@ -3,7 +3,7 @@ import { AlertEvents } from "./AlertManager";
 import React from "react";
 import { Typography } from "@mui/material";
 
-export function dialogBoxCreate(txt: string | JSX.Element, html = false): void {
+export function dialogBoxCreate(txt: string | JSX.Element, html = false, cancellable = true): void {
```
d0sboots: This is starting to get a stack of boolean args, which are easy to confuse for each other. It's probably OK for now, but if we ever feel the need to add something else here, this should become an options struct.

src/ui/React/AlertManager.tsx
catloversg: I renamed it to `canBeDismissedEasily` and added comments explaining its behaviors. I can change the name again if you want.

PR #2027, opened 2025-03-14
IPVGO: "getCornerMove" coordinates corrected

wasniahC:
The X coordinates for the left corner move should be 0 and 2, to check for X being between 0 and 2; instead, they were from 0 to boardedge (far right edge). Updated so that it now bounds this corner move logic to just that top left corner.


ficocelliguy:
Looks good to me!

PR #2029, opened 2025-03-15
IPVGO: Add new analysis method to set a custom testing board state

ficocelliguy:
Adds a new setTestingBoardState() that resets the current game, sets the opponent to "No AI", and sets the current board to the size and state of the provided one. 

Based on a few player's requests, such as here:
https://discord.com/channels/415207508303544321/1221920484111814828/1347731008258048040


src/Go/effects/netscriptGoImplementation.ts
```diff
@@ -264,6 +264,17 @@ export function getControlledEmptyNodes(_board?: Board) {
...
 
+export function setTestingBoardState(board: Board) {
+  resetBoardState(
+    () => {},
+    () => {},
+    GoOpponent.none,
+    board.length,
+  );
+  Go.currentGame.board = board;
+  GoEvents.emit();
```
d0sboots, line 277: I think you need to reset the promises here?

ficocelliguy: resetBoardState() just above calls the promise resetter added in the recently-merged PR.

PR #2031, opened 2025-03-15
CI: Pin commit id of tj-actions/changed-files

catloversg:
`tj-actions/changed-files` was compromised and taken down, so the `Validate Title and Check for Dist Changes` action failed in recent PRs. We are not in danger now, but we need to find or write an alternative solution soon. Ideally, we should write that kind of logic instead of relying on a third-party action. Our usage is very simple and can be implemented fairly easily. It still takes a while, though. In the meantime, we can switch the compromised action to a fork and pin the commit id to avoid this kind of supply chain attack.

The source code was pulled from https://code.forgejo.org/tj-actions/changed-files. I pushed everything to my fork, except the malicious commit.

Fork: https://github.com/catloversg/tj-actions-changed-files
Test the fork: https://github.com/catloversg/test-tj-actions-changed-files

I pinned it at c65cd883420fd2eb864698a825fc4162dd94482c, which is the last known-good commit of v44. You can verify this claim by checking the log of old action runs. For example: https://github.com/bitburner-official/bitburner-src/actions/runs/13858868484/job/38782020497.

### DIFF
--- a/.github/workflows/title-and-dist-checks.yml
+++ b/.github/workflows/title-and-dist-checks.yml
@@ -18,7 +18,7 @@ jobs:
         uses: actions/checkout@v4
       - name: Check bundle files
         id: changed-bundle-files
-        uses: tj-actions/changed-files@v44
+        uses: catloversg/tj-actions-changed-files@c65cd883420fd2eb864698a825fc4162dd94482c
         with:
           files: |
             dist/**
--- a/.github/workflows/title-and-dist-checks.yml
+++ b/.github/workflows/title-and-dist-checks.yml
@@ -18,7 +18,7 @@ jobs:
         uses: actions/checkout@v4
       - name: Check bundle files
         id: changed-bundle-files
-        uses: catloversg/tj-actions-changed-files@c65cd883420fd2eb864698a825fc4162dd94482c
+        uses: tj-actions/changed-files@c65cd883420fd2eb864698a825fc4162dd94482c
         with:
           files: |
             dist/**


d0sboots:
So, by the time I was generally aware today, tj-actions was re-instated and re-secured. This highlights both the up *and* downsides of large supply-chain attacks:

* They are large tempting targets, so more effort is put into attacking them and there are correspondingly more breaks.
* But since there are more eyes on them, the hacks tend to be discovered and resolved very quickly.

We're a small project, so the likelihood of someone targeting us, specifically, is rather low. But, if there was a compromise through an action on your account, or mine, it could easily go unnoticed for months, or years.

Overall, I think the right thing to do here is to pin the commit hash, but leave it pointing at tj-actions (now that it is reinstated). That should be at least as secure as having a fork, and both more transparent and easier to maintain.

catloversg:
What do you think about these actions:
- All actions provided by GitHub: https://github.com/actions
  - actions/checkout
  - actions/setup-node
  - actions/upload-pages-artifact
  - actions/deploy-pages
  - actions/checkout
- `CyberAndrii/setup-steamcmd` in `.github\workflows\steamDev.yml`. It seems that we don't use this action. Pinning it is the right choice, but it's even better if we remove this action. AFAIK, when Snarling releases Steam builds, he does not use this action and just does it manually.

d0sboots:
For the github provided ones, I don't think we need to pin hashes. If they are hacked, we have much bigger issues. 

Let's double-check that the stream one is unused.

PR #2032, opened 2025-03-16
BUGFIX: Ensure that IPvGO promises are initialized correctly on a new save and on fluming

ficocelliguy:
Previously resetAI() and handleNextTurn() (which set up promises in an appropriate state for the first move in IPvGO) were not being called on the very first game of a new save, or after a new bitnode was chosen. This caused certain next move promises to never resolve, even when it was that player's turn.

### DIFF
--- a/src/Go/boardAnalysis/goAI.ts
+++ b/src/Go/boardAnalysis/goAI.ts
@@ -27,7 +27,7 @@ type PlayerPromise = {
   resolver: ((play?: Play) => void) | null;
 };
 
-const gameOver = { type: GoPlayType.gameOver, x: null, y: null } as const;
+const gameOver: Play = { type: GoPlayType.gameOver, x: null, y: null } as const;
 const playerPromises: Record<GoColor.black | GoColor.white, PlayerPromise> = {
   [GoColor.black]: { nextTurn: Promise.resolve(gameOver), resolver: null },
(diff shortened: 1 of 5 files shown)


ficocelliguy:
@d0sboots I don't know if this is important enough to be made into a patch or not. The workaround is that there is no issue if a player makes a move or restarts the game, either manually or via the api, after an install (or on a new save), and it only applies to the very first game.

However, it's relevant for playing as white, since scripts may have to wait at the beginning of the game if they aren't playing as black.

src/Go/boardAnalysis/goAI.ts
```diff
@@ -27,7 +27,7 @@ type PlayerPromise = {
   resolver: ((play?: Play) => void) | null;
 };
 
-const gameOver = { type: GoPlayType.gameOver, x: null, y: null } as const;
+const gameOver: Play = { type: GoPlayType.gameOver, x: null, y: null } as const;
```
d0sboots, line 30: The type here is unneeded. TS is structurally typed, not declaratively typed, so this is only telling the compiler what it already knows.

ficocelliguy: I thought so, but this made Idea stop complaining about it 🤷‍♂️

PR #2035, opened 2025-03-17
API: syntax errors in NetscriptDefinitions.d.ts

shyguy1412:
The go API declares function parameters with default values. That's not valid syntax in a declaration file since default values are an implementation detail. I replaced the default values with optional booleans to fix that.

Example
```ts
repairOfflineNode(
    x: number,
    y: number,
    playAsWhite = false,
  ): Promise<{
    type: "move" | "pass" | "gameOver";
    x: number | null;
    y: number | null;
  }>;
  
  //replaced with 
  
  repairOfflineNode(
    x: number,
    y: number,
    playAsWhite?: boolean,
  ): Promise<{
    type: "move" | "pass" | "gameOver";
    x: number | null;
    y: number | null;
  }>;
  ```

### DIFF
--- a/markdown/bitburner.go.makemove.md
+++ b/markdown/bitburner.go.makemove.md
@@ -14,7 +14,7 @@ playAsWhite is optional, and attempts to make a move as the white player. Only c
 makeMove(
     x: number,
     y: number,
-    playAsWhite = false,
+    playAsWhite?: boolean,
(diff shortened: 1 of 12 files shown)


shyguy1412:
Appearently thats already part of #2028 so imma just close this again x.x

PR #2036, opened 2025-03-17
API: dynamic imports with `ns.dynamicImport()`

shyguy1412:
This adds dynamic imports with `ns.dynamicImport()`

script.js
```js
export async function main(ns){
   const script = await ns.dynamicImport("./scriptToImport.js");
   script.log(ns, "Message from an imported script!")
}
```
scriptToImport.js
```js
export async function log(ns, message){
   ns.tprint(message);
}
```
(images are from before the name got changed)
image
image


shyguy1412:
Function names from images and examples are no longer accurate, see [bdcb063](https://github.com/bitburner-official/bitburner-src/pull/2036/commits/bdcb0631e5d8219b678bb2a9102fdfba866ee5be)

shyguy1412:
Marked as draft as there is still discourse about the best function name

src/NetscriptFunctions.ts
```diff
@@ -1836,6 +1838,17 @@ export const ns: InternalAPI<NSFull> = {
   printRaw: (ctx) => (value) => {
     ctx.workerScript.print(wrapUserNode(value));
   },
+  importFrom: (ctx) => (value) => {
+    const path = helpers.scriptPath(ctx, "path", value);
+    const server = ctx.workerScript.getServer();
```
d0sboots: Getserver return should be checked, in case of things like a deleted server while running. Check other functions for an example.

src/NetscriptFunctions.ts
```diff
@@ -1836,6 +1838,17 @@ export const ns: InternalAPI<NSFull> = {
   printRaw: (ctx) => (value) => {
     ctx.workerScript.print(wrapUserNode(value));
   },
+  importFrom: (ctx) => (value) => {
```
d0sboots: The MDN for import() is very clear that it will *not* synchronously throw any errors (they'll be part of promise rejection). I think we should mirror import() as closely as possible. That means we should make this function async, so anything it throws (or compile throws) becomes part of the promise return.

markdown/bitburner.ns.md
```diff
@@ -125,6 +125,7 @@ export async function main(ns) {
 |  [hasRootAccess(host)](./bitburner.ns.hasrootaccess.md) | Check if you have root access on a server. |
 |  [hasTorRouter()](./bitburner.ns.hastorrouter.md) | Returns whether the player has access to the darkweb. |
 |  [httpworm(host)](./bitburner.ns.httpworm.md) | Runs HTTPWorm.exe on a server. |
+|  [import(path)](./bitburner.ns.import.md) | Dynamically import a script. Only scripts located from the same server can be imported. A dynamic import will not adjust RAM usage. This must be done manually with [ramOverride](./bitburner.ns.ramoverride.md)<!-- -->. |
```
d0sboots: Rerun npm doc

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -8298,6 +8298,34 @@ export interface NS {
    */
   getSharePower(): number;
 
+  /**
+   * Dynamically import a script.
+   * Only scripts located from the same server can be imported.
```
d0sboots: Optional: Maybe add a link to MDN import() function for how this works, since it's kind of weird/advanced?

shyguy1412: As far as I am aware the docs are purposefully never referencing any JS specific behavior as that is considered out of scope. I'm not sure if this is worth breaking that convention, especially since we do things a fair bit different. for example kind of input validation and no support for import assertions.

PR #2037, opened 2025-03-19
CHORE: Add mathjax-full and csstype as direct dependencies, since we depend on them

ficocelliguy:
Currently the dependency mathjax-full is required in our webpack config, but is not in our package.json. It happens to work as-is because other dependencies require it, 
https://github.com/bitburner-official/bitburner-src/blob/stable/webpack.config.js#L132

similarly src/ui/React/StatsTable.tsx uses csstype but it is only included since some other dependencies use it.

This could cause issues in the future if any of these happen to be swapped out by a dependency, or if a dev needs to build without the lockfile as happened here https://discord.com/channels/415207508303544321/459097632896188436/1351902615247785994

### DIFF
--- a/src/ui/MD/components.tsx
+++ b/src/ui/MD/components.tsx
@@ -70,7 +70,7 @@ const fixAlign = (align: React.CSSProperties["textAlign"]): TableCellProps["alig
   if (align === "end") return "inherit";
   if (align === "start") return "inherit";
   if (align === "match-parent") return "inherit";
-  return align as TableCellProps["align"];
+  return align;
 };
 
 export const Td = (props: React.PropsWithChildren<TableDataCellProps>): React.ReactElement => {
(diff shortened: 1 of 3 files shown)


src/ui/MD/components.tsx
```diff
@@ -70,7 +70,7 @@ const fixAlign = (align: React.CSSProperties["textAlign"]): TableCellProps["alig
   if (align === "end") return "inherit";
   if (align === "start") return "inherit";
   if (align === "match-parent") return "inherit";
-  return align;
+  return align as TableCellProps["align"];
```
ficocelliguy, line 73: Without this we get this build failure, perhaps from changes in some of the dependency versions since this workaround was made

https://github.com/bitburner-official/bitburner-src/actions/runs/13950987128/job/39050155791?pr=2037

PR #2044, opened 2025-03-23
DOCUMENTATION: Add BitNode recommendation

### LINKED ISSUE #2033: DOCS: Update the recommended BN order in the documentation [Good first issue]
the in-game recommendation for bitnode order is here:
https://github.com/bitburner-official/bitburner-src/blob/dev/src/Documentation/doc/help/bitnode_order.md

However, it provides fairly different recommendations from the pins in the discord, or the information that the community has been giving players more recently, in the for the large and prominent "fast progression" section.

Notably it recommends to do BN3 early, which is an extremely complex and documentation-dependent node that has often thrown newer players for a loop. Also, if a good BN3 script is eventually made, the crazy amount of money it provides makes many other mechanics irrelevant. It should not be recommended so early (especially not the second bitnode!!) and it should come with a more clear set of expectations for the player

More generally, rather than a numbered list, a few bitnodes should be recommended for early progression (repeating 1, maybe 2, or 5) with their pros and cons.

A general TL;DR guide should be at the top, and the large "fastest progression" section, if we even want to keep it, should be moved to the bottom.

catloversg:
The old "BitNode order" is obsolete. I explained the reason at https://github.com/bitburner-official/bitburner-src/issues/2033#issuecomment-2730298306.

This PR adds a new page: BitNode recommendation. It focuses on analyzing BitNodes, giving good advice to new players, helping them to make a good choice for their next BitNode, and avoiding common mistakes.

The new page is a bit long, but I don't think that it's a problem. It's still much shorter than `getting_started.md`, even after I delete all sample code, commands, output, etc. in that page.

Closes #2033.


src/Documentation/doc/advanced/bitnode_recommendation.md
```diff
@@ -0,0 +1,240 @@
...
+
+Some mechanics have peculiarities that you should know beforehand. For example, enabling territory clashes too soon will make the player lose all territory and ruin their gang.
+
+All BitNodes have a uniques set of multipliers that affect the difficulty of that BitNode. You can see these multipliers when choosing a BitNode in the BitVerse. [Source-File](./sourcefiles.md) 5 gives you access to these multipliers when you are already in a BitNode. Some BitNodes are much harder than others. It's recommended to unlock other Source-Files before trying to beat these hard BitNodes. Note that the "Difficulty" value of a BitNode in the BitVerse may be a bit misleading in some cases. Understanding the difficulty of a BitNode is much harder than just choosing between "easy", "normal" and "hard".
+
+# BitNode analysis
+
+## BitNode 1
+
+This is the easiest BitNode, and its Source-File gives a very strong buff to most multipliers.
```
ficocelliguy: It's worth reiterating briefly here that you want to _repeat_ BN 1 to get it's powerful SF-2 and 3 unlocks

src/Documentation/doc/advanced/bitnode_recommendation.md
```diff
@@ -0,0 +1,240 @@
...
+
+- Higher stat multipliers from Node Power.
+- Unlock `ns.go.cheat` APIs.
+- Higher maximum favor that you can gain for each faction from IPvGO.
+
+The last reward is pretty interesting. If you can consistently win several games in a row and get higher favor bonus, you will cut down the time to reach 150 favor points with IPvGO factions and have access to their augmentations much sooner.
+
+This BitNode's multipliers are fairly harsh, but IPvGO's multipliers are buffed significantly. If you use IPvGO, even a slightly improved version of the tutorial script will help you complete this BitNode without much trouble.
+
+# Order advice
```
ficocelliguy: This is what people come here for, probably. It should appear above the deeper dives on the bitnodes IMO

src/Documentation/doc/advanced/bitnode_recommendation.md
```diff
@@ -0,0 +1,240 @@
...
+
+You must be careful with chaos and Synthoid population. The UI shows you many hints about the effect of dangerous actions on chaos and Synthoid population. You should keep an eye on those hints. Generally, you must keep the chaos level low and not kill too much Synthoid population.
+
+## BitNode 8
+
+This BitNode unlocks 2 new features for the stock market: short stock and limit/stop order.
+
+This is a very challenging BitNode. It disables most normal ways to earn money so that you have to use the Stock Market as your source of income. Before having access to 4S Market Data, finding good stocks to invest in is pretty hard. You need to have a good pre-4S stock market script; otherwise, it's a slow run. You should take a look at stock manipulation. Hacking does not give you money, but hacking and growing a server that has a corresponding stock still influence that stock's price.
+
+Grafting is very useful in this BitNode. Losing capital after each resetting sets you back a lot. Grafting allows you to continuously install augmentations without resetting.
```
ficocelliguy: Casino is the primary fund pool for getting started in BN-8. Is it worth mentioning here?

ficocelliguy:
This looks really good overall, well done!

(There's a handful of changes outside of the BN docs as well in this PR, is that intentional?)

catloversg:
> There's a handful of changes outside of the BN docs as well in this PR, is that intentional?

All of them are generated by tools:
- Files in `markdown` folder: `npm run doc`
- `src\Documentation\pages.ts`: `node tools/bundle-doc/index.js`

For your other suggested changes:
- BitNode 1: I'm neutral.
- Move "Order advice" to the top: I thought about doing that, but I was a bit hesitant and dropped that idea in the end.
- Casino in BN8 is a trick. Without exploiting RNG (i.e., playing casino games "legitimately"), going to casino is a bad idea. **Maybe**, we can add a very vague hint about "a way to get easy money". Otherwise, I don't think we should directly mention something that depends on exploits.

Overall, I'm not against them, but let's wait for d0sboots.

Edit: For the last point about casino: I think the "fund pool" of BN8 is the initial 250m, in terms of balance. With casino cheating, it becomes 10b, which is massive, and BN8 is not that challenging anymore. With 10b, I can buy 4S Market Data (UI-only) and just play it manually for a bit. IMO, earning 25b with 10b initial fund and 4S Market Data is fairly easy and fast.

src/Documentation/doc/advanced/bitnode_recommendation.md
d0sboots, line 1: I'm going to be somewhat opinionated in my comments, because you tend to write very comprehensively, but that also means long. That works for something like Corp, but here I think we need to keep this as short as possible.

Edit: with the change in focus, this is no longer quite as true.

src/Documentation/doc/advanced/bitnode_recommendation.md
```diff
@@ -0,0 +1,240 @@
+# Introduction
+
+After a player destroys their first [BitNode](./bitnodes.md) and sees the BitVerse, they are usually overwhelmed by the large number of BitNodes and do not know which BitNode they should do next. Some BitNodes even have peculiarities that tend to catch newbies off guard and ruin their run.
```
d0sboots:
```suggestion
After a player destroys their first [BitNode](./bitnodes.md), they are usually overwhelmed and do not know which BitNode they should do next. Some BitNodes even have peculiarities that tend to catch newbies off guard and ruin their run.
```

src/Documentation/doc/advanced/bitnode_recommendation.md
```diff
@@ -0,0 +1,240 @@
...
+
+This page focuses on analyzing BitNodes, giving good advice to new players, helping them to make a good choice for their next BitNode, and avoiding common mistakes.
```
d0sboots:
```suggestion
```
The purpose of the page is clear, it doesn't need to be stated.

src/Documentation/doc/advanced/bitnode_recommendation.md
```diff
@@ -0,0 +1,240 @@
...
+
+At first, we need to make this clear: There is no "perfect" BitNode order. Everybody has their own preferred play style, so a good order for this player may be a bad order for other players. This is why this guide does not show any specific orders. The player needs to read this guide carefully and figure out the best order for them.
```
d0sboots:
```suggestion
At first, we need to make this clear: There is no "perfect" BitNode order. Everybody has their own preferred play style, so a good order for this player may be a bad order for other players. This is why this guide does not show a single specific order.
```
Most people won't read the whole thing carefully, and it should be written with that in mind.

src/Documentation/doc/advanced/bitnode_recommendation.md
```diff
@@ -0,0 +1,240 @@
...
+
+Doing BitNodes in chronological order (BN1->BN2->BN3->BN4->BN5->etc.) is a classic newbie mistake. Don't do that.
```
d0sboots:
```suggestion
Doing BitNodes in chronological order (BN1->BN2->BN3->BN4->BN5->etc.) is a classic mistake. Don't do that.
```

It's not just newbies, some people just like doing things sequentially.

src/Documentation/doc/advanced/bitnode_recommendation.md
d0sboots: Yes.

src/Documentation/doc/advanced/bitnode_recommendation.md
```diff
@@ -0,0 +1,240 @@
...
+
+Some BitNodes unlock new mechanics, while others improve or change original mechanics in BitNode 1. This is a very rough introduction:
+
+- New gameplay:
+  - Gang (BitNode 2): Simple and useful.
+  - Corporation (BitNode 3): Extremely complicated, extremely powerful, and extremely fast.
+  - Bladeburner (BitNode 6, BitNode7): Relatively simple, slow, and not be nerfed too much in most BitNodes.
+- Utility mechanics:
+  - Singularity (BitNode 4): Automation APIs.
+  - Intelligence (BitNode 5): Permanent stat that buffs many things.
```
d0sboots: Should also mention permanent formulas. They're collectively part of the same utility, sort of.

src/Documentation/doc/advanced/bitnode_recommendation.md
```diff
@@ -0,0 +1,240 @@
...
+
+BitNode 10 unlocks 2 strong mechanics at the same time. Sleeves synergize well with many mechanics, especially Gang and Bladeburner. Grafting is useful in all BitNodes. Most people complete this BitNode in one go to get 8 Sleeves, but if you are in a rush, you can complete it once and get only Source-File 10.1. If you buy all 5 Sleeves from "The Covenant" faction, you will have 6 Sleeves and access to Grafting. That's not ideal, but still good enough.
+
+BitNode 14 enhances IPvGO. If you have not touched that mechanic, you should do it now. IPvGO is tuned so that it still gives adequate benefits even if your script is only a slightly improved version of the tutorial script. Source-File 14 improves IPvGO's benefits and unlocks cheat APIs, which you can use to improve your win rate.
```
d0sboots: This should more explicitly mention that IPvGO is available at the start of the game, and (briefly) how you access it.

d0sboots (commented):
Once again, I forgot to send the comments after reviewing :(

d0sboots:
> * Move "Order advice" to the top: I thought about doing that, but I was a bit hesitant and dropped that idea in the end.

Initially I agreed with ficocceliguy. However, now that we have the other doc to serve as a TL;DR, I think it is fine to have at the end. It's *almost* redundant, except the analysis you do is slightly different. But, if you want to remove parts from this section that are redundant, or change its focus to be more different from the TL;DR guide, this would be the chance.

PR #2045, opened 2025-03-23
BUGFIX: BitVerse does not show all BN multipliers in some cases

catloversg:
How to reproduce this bug:
- Load a clean save file.
- Destroy WD and check BN multipliers in BitVerse. Multipliers of Gang, Corporation, Bladeburner, and Stanek's Gift are hidden.

We have a "feature": If the player cannot access an endgame feature, we hide that feature's multipliers in the table of BN multipliers. IMO, this feature causes more trouble than it's worth:
- The bug in this PR.
- If the player compares the data from `ns.getBitNodeMultipliers()` to this table, they will think that there is a bug.

This PR fixes this bug.

Before:

before

After:

after

@d0sboots What do you think about removing this "feature"?


d0sboots:
I feel like this is something to poll for. On one hand, there isn't a spoiler-related reason to do this: once you see the bitverse, the existence of all these mechanics is revealed. But, there is an argument that it helps to hide irrelevant info (and it *is* irrelevant if you can't use it) to make it easier to process larger amounts of numbers.

catloversg:
We can merge this one. The poll on Discord ("If the player cannot access an endgame feature, should we hide that feature's multipliers in the table of BN multipliers (Stats tab)?") ended with Yes, so I won't make any more changes. In the current state, this PR only changes the BitVerse UI.

PR #2048, opened 2025-03-24
BUGFIX: fix faction enum key of Bachman & Associates

masarakki:
it's not match `enums.CompanyName` key of `Backman & Associates`

### DIFF
--- a/src/ScriptEditor/NetscriptDefinitions.d.ts
+++ b/src/ScriptEditor/NetscriptDefinitions.d.ts
@@ -4755,12 +4755,12 @@ export interface Go {
    *
    * For example, a 5x5 board might look like this:
    *
-   [<br/>  
-      "XX.O.",<br/>  
-      "X..OO",<br/>  
-      ".XO..",<br/>  
-      "XXO.#",<br/>  
-      ".XO.#",<br/>  
+   [<br/>
+      "XX.O.",<br/>
+      "X..OO",<br/>
+      ".XO..",<br/>
+      "XXO.#",<br/>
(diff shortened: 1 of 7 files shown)


catloversg:
That's not a bug. There is no guarantee about the "relationship" of enum keys (`ns.enums.CompanyName.BachmanAndAssociates` and `ns.enums.FactionName.BachmanAssociates` for the company/faction "Bachman & Associates"). It's not ideal to have this kind of inconsistency, but that ship sailed a long time ago. This change breaks the player's scripts using `ns.enums.FactionName` for no benefit.

d0sboots:
I could see doing this as a 3.0 thing, but there'd need to also be an automigration to fix player's scripts.

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -4755,12 +4755,12 @@ export interface Go {
    *
    * For example, a 5x5 board might look like this:
    *
-   [<br/>  
-      "XX.O.",<br/>  
-      "X..OO",<br/>  
-      ".XO..",<br/>  
-      "XXO.#",<br/>  
-      ".XO.#",<br/>  
+   [<br/>
```
catloversg, line 4758: You need to revert the changes in this file.

catloversg (changes requested):
You also need to use `src\utils\APIBreaks\3.0.0.ts` to migrate the player's scripts (`FactionName.BachmanAssociates` -> `FactionName.BachmanAndAssociates`).

PR #2050, opened 2025-03-24
BUGFIX: fix exploit where favor limit from IPvGO was removed on augmentation

### LINKED ISSUE #2047: IPvGO: (exploit?) Favor from winstreaks is only capped per aug reset
Hey I went into ipvgo expecting it could improve bitnode 1.1 times by a lot. After all it gives 40 favor for black hand for winstreaking. So my thinking was this would reduce the rep needed to unlock reputation from 500k to 200k which would be a significant timesave.

However after testing it it turns out favor from winstreaks is only capped per aug reset. That means it's not 40 favor total. It's 40 favor every time you install augs. This makes unlocking favor with black hand trivial. Which makes running up your NFG trivial. Which basically roflstomps the early game. Especially since you can increase favor for Daedelus the same way.

It seems like the favor gain from winstreaks is supposed to be capped per bitnode, not per soft reset. Which is why it increasing with SF14 is a significant buff as it means unlocking donations much faster.

Now that I know this is how it works, I think I could get this time under 4 hours.

Image

ficocelliguy:
As a side effect of a recent change that resets the win/loss records of IPvGO on aug install, the favor cap per faction was also being reset (effectively allowing unlimited favor gain from IPvGO). The resetStats() go API method also had this same bug.

This PR changes both to not affect the favor gained status per faction, preventing the exploit for uncapped favor gain.

Resolves #2047 

There is ongoing discussion about further tuning favor gains from IPvGO in the discord, but I wanted to get out a fix for this exploit quickly.

### DIFF
--- a/src/Go/effects/netscriptGoImplementation.ts
+++ b/src/Go/effects/netscriptGoImplementation.ts
@@ -22,6 +22,7 @@ import { endGoGame, getOpponentStats, getScore, resetWinstreak } from "../boardA
 import { WHRNG } from "../../Casino/RNG";
 import { getRecordKeys } from "../../Types/Record";
 import { CalculateEffect, getEffectTypeForFaction } from "./effect";
+import { newOpponentStats } from "../Constants";
 
 /**
  * Check the move based on the current settings
@@ -401,7 +402,7 @@ export function resetStats(resetAll = false) {
       };
     }
   } else {
-    Go.prestigeAugmentation();
+    Go.stats[GoOpponent.none] = newOpponentStats();
   }
 }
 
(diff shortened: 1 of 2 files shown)


ficocelliguy:
@d0sboots this should be patched into stable IMO

paul-civitas:
>  further tuning favor gains from IPvGO

Unless people start rushing BN14, I don't think the SF14 bonus needs to be changed.

src/Go/effects/netscriptGoImplementation.ts
```diff
@@ -402,7 +401,7 @@ export function resetStats(resetAll = false) {
       };
     }
   } else {
-    Go.stats[GoOpponent.none] = newOpponentStats();
+    Go.prestigeAugmentation();
```
d0sboots: This seems like it's going to reset node power and bonuses? For all opponents?

ficocelliguy: good catch. This was from when I was testing a different approach to this that I scrapped, but forgot to roll back in this branch.

PR #2051, opened 2025-03-24
IPVGO: Clarify how favor is gained from wins

ficocelliguy:
Added some clarification on how favor is gained in IPvGO (it does not require high winstreaks, just any 2+ wins in a row)

based on a number of comments, most recently https://discord.com/channels/415207508303544321/459097632896188436/1353748866708541532


paul-civitas:
Thanks <3

PR #2052, opened 2025-03-25
BUGFIX: Player can bypass UI time limit of some casino games

catloversg:
Except for Blackjack, the other games play animations and disable "Play" buttons when the player plays the game. This means they cannot play games too fast (i.e., hundreds/thousands of coin flips per second). With DOM access, they can bypass this restriction. This bug happens with coin flip and roulette, not slot machine.

How to reproduce this bug:
- Apply the patch for debugging.
- Run the test code when you are in the casino page. Check the console log.

Patch: 
[casino bug.patch](https://github.com/user-attachments/files/19442810/casino.bug.patch)

Test code:
```js
async function clickButton(xpath) {
  const button = globalThis["document"].evaluate(
    xpath, globalThis["document"], null, XPathResult.FIRST_ORDERED_NODE_TYPE, null
  ).singleNodeValue;
  if (!button) {
    return;
  }
  await button[Object.keys(button)[1]].onClick({ isTrusted: true });
}

/** @param {NS} ns */
  console.clear();
  ns.ui.openTail();
  const coinFlip = `//*[@id="root"]/div[2]/div[2]/div/button[1]`;
  const slot = `//*[@id="root"]/div[2]/div[2]/div/button[2]`;
  const roulette = `//*[@id="root"]/div[2]/div[2]/div/button[3]`;

  const stopPlaying = `//*[@id="root"]/div[2]/div[2]/button[2]`;

  const head = `//*[@id="root"]/div[2]/div[2]/div/div/div/button[1]`;
  const spinSlot = `//*[@id="root"]/div[2]/div[2]/div/div/button`;
  const rouletteBet3 = `//*[@id="root"]/div[2]/div[2]/table/tbody/tr[1]/td[1]/button`;

  // Coint flip
  // for (let i = 0; i < 100; i++) {
  //   await clickButton(coinFlip);
  //   await clickButton(head);
  //   await clickButton(stopPlaying);
  //   await ns.asleep(0);
  // }

  // Slot
  // for (let i = 0; i < 100; i++) {
  //   await clickButton(slot);
  //   await clickButton(spinSlot);
  //   await clickButton(stopPlaying);
  //   await ns.asleep(0);
  // }

  // Roulette
  for (let i = 0; i < 100; i++) {
    await clickButton(roulette);
    await clickButton(rouletteBet3);
    await clickButton(stopPlaying);
    if (i % 5 === 0) {
      await ns.asleep(0);
    }
  }
}
```

### DIFF
--- a/src/Casino/CoinFlip.tsx
+++ b/src/Casino/CoinFlip.tsx
@@ -7,7 +7,7 @@ import React, { useState } from "react";
 
 import { BadRNG } from "./RNG";
 import { win, reachedLimit } from "./Game";
-import { trusted } from "./utils";
+import { PlayLockManager, trusted } from "./utils";
 
 import Typography from "@mui/material/Typography";
 import TextField from "@mui/material/TextField";
@@ -16,6 +16,9 @@ import Box from "@mui/material/Box";
 
 const minPlay = 0;
 const maxPlay = 10e3;
+
+const uiDelay = 250;
+const playLockManager = new PlayLockManager(uiDelay);
 
 export function CoinFlip(): React.ReactElement {
   const [investment, setInvestment] = useState(1000);
@@ -38,7 +41,10 @@ export function CoinFlip(): React.ReactElement {
   }
 
   function play(guess: string): void {
-    if (reachedLimit()) return;
+    if (reachedLimit() || playLockManager.isLocked()) {
+      return;
+    }
+    playLockManager.play();
     const v = BadRNG.random();
     let letter: string;
     if (v < 0.5) {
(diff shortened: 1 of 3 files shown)


d0sboots:
I'm not sure how I feel about this one. Players should not be able to bypass the money limit on casino (without serious and obvious exploit-level shenanigans). However, using dom access to bypass the animation limits, especially for simple games like coinflip is a clever technique that I know some people have used. I don't really see the benefit in disallowing it.

catloversg:
I don't have a strong opinion about this issue. Cald reported it, and there was no objection in "fixing" it (https://discord.com/channels/415207508303544321/459097632896188436/1352641405213868126), so I made this PR. I think we should make a decision (either you make it or we create a poll for it). I'm fine either way.

d0sboots:
Did we end up polling this one? If not, I'm gonna say let's leave this as an exploit technique to discover.

Alpheus:
That's clever, is the await hitting the queue on the next frame after the event handler? Imho clever tricks like this can be kept in the game. There are other alternatives to get money from casino asap. I voted on the poll to keep it unfixed.

catloversg:
The poll ended with a draw result (5 Yes vs 6 No, without my vote; I'm a bit leaning on Yes). I'll close this PR.

PR #2053, opened 2025-03-25
MISC: Clarify conditions of activating Gang, Bladeburner, Stanek's Gift

catloversg:
Some players occasionally ask us why they cannot activate some mechanics (Gang, Corporation, Bladeburner, Stanek's Gift). The TSDoc of NS APIs usually don't mention important conditions, and when the check fails, it does not print the reason in the log. I already fixed this problem with Corporation. This PR fixes it with other mechanics.


src/NetscriptFunctions/Bladeburner.ts
```diff
@@ -308,27 +308,36 @@ export function NetscriptBladeburner(): InternalAPI<INetscriptBladeburner> {
     },
     joinBladeburnerDivision: (ctx) => () => {
       if (!canAccessBitNodeFeature(7) && !canAccessBitNodeFeature(6)) {
-        return false; //Does not have bitnode 6 or 7
-      } else if (Player.bitNodeOptions.disableBladeburner) {
+        helpers.log(ctx, () => "You do not have Source-File 6 and Source-File 7.");
```
d0sboots:
```suggestion
        helpers.log(ctx, () => "You do not have Source-File 6 or Source-File 7.");
```

catloversg:
I'm okay with the editorial change at https://github.com/bitburner-official/bitburner-src/pull/2053/commits/d5370737860ef0c66bbbecc58c03d36192307ce2.

d0sboots:
Thanks for the reply, I meant to just submit it but got distracted while the tests were running.

PR #2063, opened 2025-03-31
UI: Warn player if they enable territory clash when their gang is too weak

catloversg:
With #1760 and #2061, we warn the player if they try to enable the territory clash when their gang is too weak. However, if they already did that, there is no warning/reminder that they made a bad choice. I think we should put a big warning on the main tab as another reminder.

With this amount of warnings, if they still lose all territory, well, that's on them.

Capture

### DIFF
--- a/src/Gang/Gang.ts
+++ b/src/Gang/Gang.ts
@@ -26,6 +26,7 @@ import { Player } from "@player";
 import { PowerMultiplier } from "./data/power";
 import { FactionName } from "@enums";
 import { CONSTANTS } from "../Constants";
+import { Result } from "../types";
 
 export enum RecruitmentResult {
   Success = "Success",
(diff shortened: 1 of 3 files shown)


d0sboots:
I feel like this might be too aggressively hand-holdy? I suppose the one dialog is already in prod, although not the tweak you made to it. Is this still a major problem? (If so, the other dialog seems like it isn't pulling its weight.)

catloversg:
> I feel like this might be too aggressively hand-holdy?

Maybe.

We can put this change on hold for a while. If there is nobody (or a very small number of players) asking about "0% territory", it means the toggle warning is good enough, and we don't need this new warning message.

ficocelliguy:
I really like the message to the user. I think it should be slightly less direct and slightly more in-universe, perhaps? 
"It's bad out there, boss, we should prep more before clashing" or "warning: clash projections below 50%. Territory loss likely" or something might fit the theme? 

We've repeatedly seen that the mechanic is under-explained, so there's a lot of value to a note here, but it might feel less aggressive or hand-holdey if it was shorter and more indirect.

PR #2065, opened 2025-04-02
MISC: Remove useless iframe in April Fool's Day popup

catloversg:
A player notified us that the current YouTube URL died. After replacing and testing a new URL, I realize that this "feature" does not work nowadays. We use a hidden autoplay iframe to provide sound for the Rickroll joke, but modern browsers do not allow autoplaying a video unless it's muted. This means this hidden iframe is useless.

### DIFF
--- a/src/ui/Apr1.tsx
+++ b/src/ui/Apr1.tsx
@@ -65,15 +65,6 @@ export function Apr1(): React.ReactElement {
   return (
     <Modal open={open} onClose={() => setOpen(false)}>
       <pre style={{ color: "#0f0" }}>{frames[n]}</pre>
-      <div style={{ display: "none" }}>
-        <iframe
-          width="560"
-          height="315"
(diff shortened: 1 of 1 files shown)


catloversg:
After more testing, it seems that autoplaying with sound is still possible in some cases. For example, with Chrome: https://developer.chrome.com/blog/autoplay.

~After changing settings in my adblocker (uBlock Origin), it works again with Chrome and Firefox.~
(Edit: The adblocker is not the problem. It works because I use `apr1` instead of changing `isApr1()` to force the modal to open. Autoplaying works if the player interacts with the page before the iframe is loaded.)
However, it does not work in the Electron version. In Electron, the iframe is not loaded, and the YouTube's iframe only shows "Video unavailable". I'm not sure if it's worth the effort to troubleshoot it further.

Capture
(The URL is correct. The iframe is loaded normally in Chrome and Firefox.)

PR #2082, opened 2025-04-11
DOCS: Fixed linking between markdown docs

mizmantle:
1. _**intelligence.md**_ did not properly link to _**companies.md**_.
2. _**scripts.md**_ did not properly link to _**servers.md**_.

*Append:
changelog-v0.md has a link at line 2092 to a no longer existing file. It can be removed, unless it should be kept for posterity.


d0sboots:
The links changing/being broken for very old changelog entries is not a surprise.

PR #2085, opened 2025-04-11
MISC: Enforce stricter param check on ns.getBitNodeMultipliers and ns.hacknet.spendHashes

catloversg:
This PR makes changes in 3 APIs:
- `ns.getBitNodeMultipliers`: Minor behavior change.
- `ns.hacknet.spendHashes`: <https://github.com/bitburner-official/bitburner-src/pull/1127#issuecomment-1966690212>
- `ns.growthAnalyze`: No behavior change.

### DIFF
--- a/src/NetscriptFunctions/Hacknet.ts
+++ b/src/NetscriptFunctions/Hacknet.ts
@@ -201,7 +201,10 @@ export function NetscriptHacknet(): InternalAPI<IHacknet> {
       (_upgName, _upgTarget = "", _count = 1) => {
         const upgName = helpers.string(ctx, "upgName", _upgName);
         const upgTarget = helpers.string(ctx, "upgTarget", _upgTarget);
-        const count = helpers.positiveInteger(ctx, "count", _count);
+        const count = helpers.integer(ctx, "count", _count);
+        if (count < 0) {
+          throw helpers.errorMessage(ctx, "Count must be a non-negative integer.");
+        }
         if (!hasHacknetServers()) {
           return false;
         }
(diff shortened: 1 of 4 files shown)


src/NetscriptFunctions/Hacknet.ts
```diff
@@ -201,11 +201,7 @@ export function NetscriptHacknet(): InternalAPI<IHacknet> {
       (_upgName, _upgTarget = "", _count = 1) => {
         const upgName = helpers.string(ctx, "upgName", _upgName);
         const upgTarget = helpers.string(ctx, "upgTarget", _upgTarget);
-        const count = Math.floor(helpers.number(ctx, "count", _count));
-        // TODO (3.0.0): use helpers.positiveInteger
-        if (!(count >= 0)) {
```
d0sboots, line 206: This is one behavior change I'm not 100% sure of. Upgrading 0 times (and it being a noop) seems like a valid thing that would potentially make code simpler in various circumstances.

catloversg: I'm a bit leaning towards the positive integer requirement. Snarling mentioned it in #1127. Nonetheless, it's up to you. Allowing 0 is not really bad, so I'm fine with it as well.

d0sboots: This is one we can't automatically migrate. The throw will make it easier to track down where it's happening, but possibly not why. Also, it just doesn't seem invalid to me. Just yesterday I used string.repeat with a count of 0, which seems fundamentally the same.

PR #2088, opened 2025-04-12
MISC: Remove APIs that were deprecated a long time ago

catloversg:
These APIs were deprecated a long time ago:
- `ns.nFormat()` (since v2.3.0)
- `ns.getTimeSinceLastAug()` (since v2.3.0)
- `ns.getPlayer().playtimeSinceLastAug` (since v2.3.0)
- `ns.getPlayer().playtimeSinceLastBitnode` (since v2.3.0)
- `ns.getPlayer().bitNodeN` (since v2.3.0)
- `ns.corporation.getCorporation().state` (since v2.5.1)

Our codebase handles breaking changes in many different ways:
- `src\utils\v1APIBreak.ts`
- `src\utils\v2APIBreak.ts`
- `src\utils\APIBreaks\APIBreak.ts`

There are many subtle differences between them, so refactoring and unifying all of them are too risky and may not be worth the effort. After considering them, I decided to go with the last way (added by Snarling). I changed it a bit to support automatic migration (when possible).

Note that we have `setRemovedFunctions` to show a custom error message when the player uses a removed function, but we don't have an equivalent function for removed properties. I only added config for `nFormat` and `getTimeSinceLastAug` for consistency.

Test code:
```js
/** @param {NS} ns */
  console.log(ns.nFormat(1000, ""));
  console.log(ns.getTimeSinceLastAug());
  console.log(ns.getPlayer().playtimeSinceLastAug);
  console.log(ns.getPlayer().playtimeSinceLastBitnode);
  console.log(ns.getPlayer().bitNodeN);
  console.log(ns.corporation.getCorporation().state);
}
```

```
API BREAK INFO FOR 3.0.0

ns.nFormat() was removed.
Use ns.formatNumber, ns.formatRam, ns.formatPercent, or JS built-in objects/functions (e.g., Intl.NumberFormat, Intl.PluralRules, Intl.Locale) instead.

ns.nFormat

a.js: (Line number: 3)



ns.getTimeSinceLastAug was removed.
It has been automatically replaced with "Date.now() - ns.getResetInfo().lastAugReset".

ns.getPlayer().playtimeSinceLastAug was removed.
It has been automatically replaced with "Date.now() - ns.getResetInfo().lastAugReset".

ns.getPlayer().playtimeSinceLastBitnode was removed.
It has been automatically replaced with "Date.now() - ns.getResetInfo().lastNodeReset".

ns.getPlayer().bitNodeN was removed.
It has been automatically replaced with "ns.getResetInfo().currentNode"

ns.corporation.getCorporation().state was removed.
It has been automatically replaced with "ns.corporation.getCorporation().nextState"

ns.getTimeSinceLastAug
ns.getPlayer().playtimeSinceLastAug
ns.getPlayer().playtimeSinceLastBitnode
ns.getPlayer().bitNodeN
ns.corporation.getCorporation().state

a.js: (Line numbers: 4, 5, 6, 7, 8)
```


d0sboots:
High-level comment: I think we are notifying the user too much here. Migrations that were completed automatically should be noted in the file, but typically don't need a popup. To facilitate this, "info" should become optional, and should be the gating factor on if a popup is shown. We should only use both if there is a high chance of unavoidable missed migration sites.

catloversg:
Update:
- Only show a popup for `ns.nFormat`, which cannot be migrated.
- Put other APIs into a "group" and skip showing a popup for them. These APIs were deprecated a long time ago, so I want to put them into a group to differentiate them from other upcoming changes (e.g., `setAutoJobAssignment`). I can split them back into multiple "groups" if you want.

d0sboots:
> Update:
> 
> * Only show a popup for `ns.nFormat`, which cannot be migrated.
> * Put other APIs into a "group" and skip showing a popup for them. These APIs were deprecated a long time ago, so I want to put them into a group to differentiate them from other upcoming changes (e.g., `setAutoJobAssignment`). I can split them back into multiple "groups" if you want.

Sounds great, can you update the base post with the text file/toasts please?

src/utils/APIBreaks/APIBreak.ts
```diff
@@ -115,7 +123,10 @@ export function showAPIBreaks(version: string, ...breakInfos: APIBreakInfo[]) {
...
       `A file with these details has also been saved on your home computer under filename ${textFileName}.`,
   );
-  details.forEach((detail, i) => {
-    dialogBoxCreate(`API BREAK VERSION ${version} DETAILS ${i + 1} of ${details.length}\n\n${detail}`);
-  });
+  for (const detail of details) {
+    if (!detail.showPopUp) {
+      continue;
+    }
+    dialogBoxCreate(`API BREAK VERSION ${version} DETAILS\n\n${detail.text}`);
```
d0sboots: I liked the "1 of 6" bit, so you knew how many toasts you were sitting through. Can we keep that? (Reason: We're gonna be adding more of these this version.)

catloversg:
I updated the description.

src/utils/APIBreaks/3.0.0.ts
```diff
@@ -0,0 +1,63 @@
...
+        name: "ns.corporation.getCorporation().state",
+        migration: {
+          searchValue: "ns.corporation.getCorporation().state",
+          replaceValue: "ns.corporation.getCorporation().nextState",
+        },
+      },
+    ],
+    info:
+      "ns.getTimeSinceLastAug was removed.\n" +
+      "Use Date.now() - ns.getResetInfo().lastAugReset instead.\n\n" +
```
d0sboots:
```suggestion
      "It has been automatically replaced with "Date.now() - ns.getResetInfo().lastAugReset".\n\n" +
```
Seeing the final file output made me realize these should be reworded. It's not a call to action, but rather a notification of something that already occurred, in case you want to review it.

We might want to make the distinction between the two groups more clear in some other way too, although if that's annoying then this is fine.

catloversg: > Seeing the final file output made me realize these should be reworded. It's not a call to action, but rather a notification of something that already occurred, in case you want to review it.

Done.

> We might want to make the distinction between the two groups more clear in some other way too, although if that's annoying then this is fine.

I'm okay with doing that, but I have no idea what the change should be.

PR #2079, opened 2025-04-09
API: Renamed Division.type to Division.industry

whiskeyfur:
# Summary

This pull request is for renaming Division.type changed to Division.industry.

# Reasoning

At first glance it's not easy by field name alone that 'type' means industry type, especially with 'type's meaning when talking about objects and classes.

# Linked issues

closes #2078 

# Documentation

- API Change, Division.type renamed to Division.industry.
- Not backwards compatible with v2.8.1.

# Bug fix
Tested with this NSscript (test.ts):
```js
/** @param {NS} ns */
export async function main(ns: NS) {

  for (var divName of ns.corporation.getCorporation().divisions) {
    ns.corporation.sellDivision(divName)
  }

  for (var industry of ns.corporation.getConstants().industryNames) {
    ns.corporation.expandIndustry(industry, industry)
  }

  for (var divName of ns.corporation.getCorporation().divisions) {
    ns.tprint("Checking for " + divName)
    var div = ns.corporation.getDivision(divName)
    ns.tprint(" v2.8.1 Type____: " + div.type ?? "undefined")     // Version <= 2 ? "Agriculture" : undefined
    ns.tprint(" v3.0.0 Industry: " + div.industry ?? "undefined") // Version > 3 ? "Agriculture" : undefined
    ns.tprint(" ");
  }
}
```

Result: (v.3.0.0dev)
```txt
Running script with 1 thread, pid 18 and args: [].
test.ts: Checking for Water Utilities
test.ts:  v2.8.1 Type____: Water Utilities
test.ts:  v3.0.0 Industry: undefined
test.ts:  
test.ts: Checking for Spring Water
test.ts:  v2.8.1 Type____: Spring Water
test.ts:  v3.0.0 Industry: undefined
.
.
.
```
Result: (v.3.0.0dev)
```txt
[home /]> run test.ts 
Running script with 1 thread, pid 2 and args: [].
test.ts: Checking for Water Utilities
test.ts:  v2.8.1 Type____: Water Utilities
test.ts:  v3.0.0 Industry: Water Utilities
test.ts:  
test.ts: Checking for Spring Water
test.ts:  v2.8.1 Type____: Spring Water
test.ts:  v3.0.0 Industry: Spring Water
test.ts:  
.
.
.
```
✅Make sure you run `npm run format` and `npm run lint` before pushing.

PR #2091, opened 2025-04-16
MISC: Remove fuzzy matching when checking params

catloversg:
We allow fuzzy matching for many things (crime names, job fields, work types, Bladeburner action types, etc.). This is convenient for players, especially in the old eras back when the game was "primitive" (very limited documentation, no ns.enums, etc.) and had too many breaking changes (Daniel and hydroflame pumped updates daily/weekly).

IMO, this "feature" is unnecessary now, and it's a "double-edged sword". Snarling explained the reason at https://github.com/bitburner-official/bitburner-src/issues/372#issuecomment-1474716982. Please check his comment in that issue. I totally agree with Snarling there.

This PR removes fuzzy matching for:
- Bladeburner action types.
- Crime names.
- Gym types.
- University class types.
- Faction work types.
- Job fields.
- Stock: position types.
- Stock: order types.

I did not change:
- Program names: AFAIK, we intentionally made program names case-insensitive.
- BB console.

Test code:
```js
function correct(ns) {
  console.log(ns.bladeburner.startAction("Operations", "Investigation"));

  console.log(ns.singularity.universityCourse("Rothman University", "Algorithms", false));
  console.log(ns.singularity.gymWorkout("Powerhouse Gym", "str", false));
  ns.singularity.quitJob("FoodNStuff");
  console.log(ns.singularity.applyToCompany("FoodNStuff", "Part-time Employee"));
  console.log(ns.singularity.workForFaction("CyberSec", "hacking", false));
  console.log(ns.singularity.commitCrime("Shoplift", false));
  console.log(ns.singularity.getCrimeChance("Shoplift"));
  console.log(ns.singularity.getCrimeStats("Shoplift"));

  console.log(ns.sleeve.setToCommitCrime(0, "Shoplift"));
  console.log(ns.sleeve.setToUniversityCourse(0, "Rothman University", "Algorithms"));
  console.log(ns.sleeve.setToFactionWork(0, "CyberSec", "hacking"));
  console.log(ns.sleeve.setToGymWorkout(0, "Powerhouse Gym", "str"));
  console.log(ns.sleeve.setToBladeburnerAction(0, "Field Analysis"));

  console.log(ns.formulas.work.crimeSuccessChance(ns.getPlayer(), "Shoplift"));
  console.log(ns.formulas.work.crimeGains(ns.getPlayer(), "Shoplift"));

  console.log(ns.formulas.work.gymGains(ns.getPlayer(), "str", "Powerhouse Gym"));
  console.log(ns.formulas.work.universityGains(ns.getPlayer(), "Algorithms", "Rothman University"));
  console.log(ns.formulas.work.factionGains(ns.getPlayer(), "hacking", 0));

  console.log(ns.stock.getPurchaseCost("ECP", 1, "L"));
  console.log(ns.stock.getSaleGain("ECP", 1, "L"));
  console.log(ns.stock.placeOrder("ECP", 1, 1000, "Limit Buy Order", "L"));
  console.log(ns.stock.cancelOrder("ECP", 1, 1000, "Limit Buy Order", "L"));
}

function wrong(ns) {
  // console.log(ns.bladeburner.startAction("ops", "Investigation"));

  // console.log(ns.singularity.universityCourse("Rothman University", "algorithms", false));
  // console.log(ns.singularity.gymWorkout("Powerhouse Gym", "strength", false));
  // ns.singularity.quitJob("FoodNStuff");
  // console.log(ns.singularity.applyToCompany("FoodNStuff", "part-time employee"));
  // console.log(ns.singularity.workForFaction("CyberSec", "hacking contracts", false));
  // console.log(ns.singularity.commitCrime("shop lift", false));
  // console.log(ns.singularity.getCrimeChance("shop lift"));
  // console.log(ns.singularity.getCrimeStats("shop lift"));

  // console.log(ns.sleeve.setToCommitCrime(0, "shop lift"));
  // console.log(ns.sleeve.setToUniversityCourse(0, "Rothman University", "algorithms"));
  // console.log(ns.sleeve.setToFactionWork(0, "CyberSec", "hacking contracts"));
  // console.log(ns.sleeve.setToGymWorkout(0, "Powerhouse Gym", "strength"));
  // console.log(ns.sleeve.setToBladeburnerAction(0, "Field analysis"));

  // console.log(ns.formulas.work.crimeSuccessChance(ns.getPlayer(), "shop lift"));
  // console.log(ns.formulas.work.crimeGains(ns.getPlayer(), "shop lift"));

  // console.log(ns.formulas.work.gymGains(ns.getPlayer(), "strength", "Powerhouse Gym"));
  // console.log(ns.formulas.work.universityGains(ns.getPlayer(), "algorithms", "Rothman University"));
  // console.log(ns.formulas.work.factionGains(ns.getPlayer(), "hacking contracts", 0));

  // console.log(ns.stock.getPurchaseCost("ECP", 1, "long"));
  // console.log(ns.stock.getSaleGain("ECP", 1, "long"));
  // console.log(ns.stock.placeOrder("ECP", 1, 1000, "limit buy", "long"));
  // console.log(ns.stock.cancelOrder("ECP", 1, 1000, "limit buy", "long"));
}

/** @param {NS} ns */
  console.clear();
  correct(ns);
  wrong(ns);
}
```


d0sboots:
I feel similarly here as I do on #2079, but even more intensely (in both directions). On one hand, I absolutely agree with the negative consequences of this fuzzy matching. But on the other hand, there are certainly a *lot* of scripts that are currently relying on it, it is impossible to automatically migrate them, and figuring out what is suddenly going wrong when 3.0 hits will be tricky.

Maybe we just go ahead and upset that applecart, say it's a major version and to hell with it. I worry about all the players we lose permanently whenever we break their scripts though, in particular ones who step away from the game for months/years and then come back to find everything broken.

catloversg:
Personally, I prefer the "Just do it" approach, but we can also find a middle ground for this problem.

My suggestion: With affected APIs, we call `getEnumHelper().nsGetMember()` with a special flag that allows fuzzy matching. If the value is not exactly correct but is still "acceptable" via fuzzy matching, we will print a warning in the terminal. This warning is shown once per API. It works like how we implemented `deprecationWarning`.

It creates a grace period in which the community knows that "fuzzy matching" will be removed soon. This grace period will end in v4 or a bit sooner. However, if the player is inactive now (i.e., already took a break) and only comes back in v4+, this grace period will be useless for them. In the worst case, we can add a setting to allow them to activate a legacy mode in which the fuzzy matching is enabled even in v4+. To be honest, I really hate the idea of a legacy mode. We should only allow a grace period until v4, without any special option after that.

d0sboots:
I am also not a big fan of that. 

I think most of my issue comes from the fact that BitBurner does not operate like conventional software, because usually the user has *choice* about accepting changes and the timing of doing so. 

I think I've also talked myself into a low-effort solution to this that will make me happy, at least: if we can get snarling to make this last 2.x release as a permanent pinned release on Steam, there is a backup option for people who don't want to change, or if the upgrade breaks badly. Then I'm not worried about being more radical with 3.0.

markdown/bitburner.tix.getorders.md
```diff
@@ -9,17 +9,17 @@ Returns your order book for the stock market.
...
-[StockOrder](./bitburner.stockorder.md)
+Record&lt;string, [StockOrder](./bitburner.stockorder.md)<!-- -->\[\]&gt;
 
-Object containing information for all the Limit and Stop Orders you have in the stock market.
+Object containing information for all the Limit and Stop Orders you have in the stock market. Keys are stock symbols, and properties are arrays of [StockOrder](./bitburner.stockorder.md)
 
 ## Remarks
 
-RAM cost: 2.5 GB This is an object containing information for all the Limit and Stop Orders you have in the stock market. For each symbol you have a position in, the returned object will have a key with that symbol's name. The object's properties are each an array of [StockOrderObject](./bitburner.stockorderobject.md) The object has the following structure:
+RAM cost: 2.5 GB This is an object containing information for all the Limit and Stop Orders you have in the stock market. For each symbol you have a position in, the returned object will have a key with that symbol's name. The object's properties are each an array of  The object has the following structure:
```
d0sboots: This went wrong

src/Bladeburner/Bladeburner.ts
```diff
@@ -1406,8 +1406,33 @@ export class Bladeburner implements OperationTeam {
...
 
-  /** Fuzzy matching for action identifiers. Should be removed in 3.0 */
-  getActionFromTypeAndName(type: string, name: string): Action | null {
+  getActionFromTypeAndName(type: BladeburnerActionType, name: string): Action | null {
+    switch (type) {
+      case BladeburnerActionType.General:
+        if (!getEnumHelper("BladeburnerGeneralActionName").isMember(name)) {
+          return null;
+        }
+        return GeneralActions[name];
```
d0sboots: Optional: It would be a lot less code if we could just directly lookup in these objects, passing the natural undefined back when it's not there. Making typescript happy might be difficult though, and downstream there's possible undefined/null conversion to consider.

d0sboots (commented):
I'm going to use the poll on the other PR as a proxy for this one. Yes, the specific issues are different, which might cause people to vote differently, but the core issues are the same

src/Bladeburner/Bladeburner.ts
catloversg: It can be implemented like this:
```js
  getActionFromTypeAndName(type: BladeburnerActionType, name: string): Action | undefined {
    switch (type) {
      case BladeburnerActionType.General:
        return GeneralActions[name as BladeburnerGeneralActionName];
      case BladeburnerActionType.Contract:
        return this.contracts[name as BladeburnerContractName];
      case BladeburnerActionType.Operation:
        return this.operations[name as BladeburnerOperationName];
      case BladeburnerActionType.BlackOp:
        return BlackOperations[name as BladeburnerBlackOpName];
    }
  }
```
Do you want it to be like that? The typecasting looks a bit weird. We have to do that if we skip the old check.

d0sboots: I think that's worth it, yeah.

d0sboots:
I ended up merging the other PR, and I'm going to assume/hope we can get a Steam branch for pre-3.0, so this is approved.

d0sboots:
BTW thanks for your effort on migrating Insight's scripts. It definitely confirms my gut feel that this will probably be the most disruptive of all of the 3.0 changes, at least for people with long-standing scripts.

d0sboots:
On that note, I just realized that this doesn't have a 3.0 break notice, does it? It really needs one, since like I said this is gonna hurt more than all the others.

catloversg:
I'll create a meta issue on GitHub and make a PR showing a break notice pointing to that meta issue.

Edit: Rereading your comment makes me realize that you may be talking about `src\utils\APIBreaks\3.0.0.ts`. Yeah, I'll also add an entry in that file.

PR #2097, opened 2025-04-18
IPVGO: do not count go favor as part of the favor gained from reputation, to avoid punishing players for getting go favor at the 'wrong time'

ficocelliguy:
Currently, all favor is assumed to come from reputation grinding, and it takes much more rep to gain each favor when you already have 100 compared to having 0 favor. Thus, it ended up being pretty punishing in terms of rep needed to hit 150 favor if a player got their favor gains from IPvGO early in that process (even including the boosts that favor gives to rep gain!) A number of players have requested this be changed to avoid a "noob trap" feelsbad scenario.

player request 1: 
https://discord.com/channels/415207508303544321/923445914461413376/1361035760643473509

player request 2:
https://discord.com/channels/415207508303544321/459097632896188436/1353769464839016542

This PR also has some refactoring to split out some parts of Go's code. 

Previously there was an issue where the Company class needed to use the rep & favor gain tools, which needed to look at the Go tools to account for Go favor. One of the Go imports looked at at Factions to be able to add favor. Factions imports the router from game root for some reason, which then imports a bunch of react components. One of which is the hashnet UI, which needs companies to populate a dropdown, thus going in a circle. 

The changes here split out the parts of Go that need to see Factions from the other parts, to break this chain.

### DIFF
--- a/src/Go/Go.ts
+++ b/src/Go/Go.ts
@@ -1,9 +1,7 @@
 import type { BoardState, OpponentStats } from "./Types";
 
-import type { GoOpponent } from "@enums";
-import { getRecordKeys, PartialRecord } from "../Types/Record";
-import { resetGoPromises } from "./boardAnalysis/goAI";
-import { getNewBoardState } from "./boardState/boardState";
+import { GoColor, GoOpponent } from "@enums";
+import { PartialRecord } from "../Types/Record";
 import { EventEmitter } from "../utils/EventEmitter";
 
 export const getEmptyHighlightedPoints = (size: number = 7) => {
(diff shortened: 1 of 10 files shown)


src/Go/Go.ts
```diff
@@ -50,6 +25,28 @@ export class GoObject {
   }
 }
 
+function getEmptyBoardState() {
```
d0sboots, line 28: Why are you duplicating all this code instead of just using getNewBoardState?

src/Go/effects/favor.ts
```diff
@@ -0,0 +1,12 @@
...
+import { Go } from "../Go";
+import { newOpponentStats } from "../Constants";
+
+export function getGoFavorForFaction(factionName: string) {
+  const factionDetails = getOpponentStats(factionName as unknown as GoOpponent);
+  return factionDetails?.favor ?? 0;
+}
+
+export function getOpponentStats(opponent: GoOpponent) {
+  return Go.stats[opponent] ?? (Go.stats[opponent] = newOpponentStats());
```
d0sboots, line 11: This is kinda icky. Things named "get" should generally be idempotent and not modify the state. It would be better to properly initialize the opponent stats ahead of time, IMO.

Edit: Do you even need this function? I don't see anywhere it is used except `getGoFavorForFaction`. You can just return 0 if the key is not found, you don't need to modify the mapping?

src/Go/effects/favor.ts
```diff
@@ -0,0 +1,12 @@
+import { GoOpponent } from "@enums";
+import { Go } from "../Go";
+import { newOpponentStats } from "../Constants";
+
+export function getGoFavorForFaction(factionName: string) {
+  const factionDetails = getOpponentStats(factionName as unknown as GoOpponent);
```
d0sboots, line 6: This is some really gross casting. I think what you want to do instead is cast `Go.stats` to a Record<string, OpponentStats>`.

src/ui/React/CompanyDropdown.tsx
```diff
@@ -15,9 +15,8 @@ interface IProps {
   value: CompanyName;
 }
 
-const sortedCompanies = getRecordKeys(Companies).sort((a, b) => a.localeCompare(b));
-
 export function CompanyDropdown(props: IProps): React.ReactElement {
+  const sortedCompanies = useMemo(() => getRecordKeys(Companies ?? {}).sort((a, b) => a.localeCompare(b)), []);
```
d0sboots, line 19: Why is this change happening? It doesn't seem necessary to me, or related to anything else in this PR.

d0sboots:
I'm gonna send these comments which I think apply regardless of what we do about the central question.

ficocelliguy:
Closing this in favor of this other, much better, approach:
https://github.com/bitburner-official/bitburner-src/pull/2131

PR #2105, opened 2025-04-21
BUGFIX: Electron app cannot sync new save file to Steam Cloud

catloversg:
Some players reported that Steam Cloud did not sync their save file, especially after they imported another save file or create a new one by using "Delete Save". Please check my comment in `saveCloudFile` for more information.

There is another way to fix this problem, but it's fairly complicated. At first, we need to ask hydroflame to set the max number of files to a higher value. After that, we need to refactor the way we manage the Steam Cloud files and save slots. IMO, we don't have a useful save slot system, so this change does not bring much benefit.

How to test:
- Open the Electron app.
- Options -> Delete Save.
- Wait at least 15 minutes or use dev menu to add play time.
- Save game and quit app.
- Check:
  - Log
  - https://store.steampowered.com/account/remotestorageapp/?appid=1812820
  - path\bitburner\saves\_backups

Sample log:
```
[2025-04-21 xx:xx:59.046] [info]  Started app: ["path\\.build\\bitburner-win32-x64\\bitburner.exe"]
[2025-04-21 xx:xx:59.161] [info]  Application is ready!
[2025-04-21 xx:xx:09.236] [warn]  'path\bitburner\saves\a64d9f2b7b6c9' not found, creating it...
[2025-04-21 xx:xx:20.621] [info]  Saving to Steam Cloud as a64d9f2b7b6c9.json.gz
[2025-04-21 xx:xx:20.622] [info]  Found a different cloud file. The current file will be deleted. Current file: da3a9406f047b.json.gz. New file: a64d9f2b7b6c9.json.gz.
[2025-04-21 xx:xx:22.262] [info]  Quitting the app...
```

### DIFF
--- a/electron/main.js
+++ b/electron/main.js
@@ -128,7 +128,7 @@ function setStopProcessHandler(window) {
       if (playtime > minimumPlaytime) {
         saveToCloud(save);
       } else {
-        log.debug(`Auto-save to cloud disabled for save game under ${minimumPlaytime}ms (${playtime}ms)`);
+        log.info(`Auto-save to cloud disabled for save game under ${minimumPlaytime}ms (${playtime}ms)`);
       }
     }
   };
--- a/electron/storage.js
+++ b/electron/storage.js
@@ -119,6 +119,23 @@ function isCloudEnabled() {
 }
 
 function saveCloudFile(name, content) {
+  /**
+   * Bitburner's Steam Cloud allows only 1 file, so if the player has a save file stored in Steam Cloud, they cannot
+   * save another file. For example, let's say they start playing with SaveID s123. After a few months, they want to
+   * start another run with a clean save file; the new SaveId is s456. In this case, the new s456.json.gz cannot be
+   * saved to Steam Cloud because s123.json.gz exists. In this case, we need to delete the current cloud file before
+   * writing the new file.
+   */
+  const files = steamworksClient.cloud.listFiles();
+  if (files.length !== 0) {
+    const currentCloudFile = getFilenameOfFirstCloudFile();
+    if (currentCloudFile !== name) {
+      log.info(
+        `Found a different cloud file. The current file will be deleted. Current file: ${currentCloudFile}. New file: ${name}.`,
+      );
+      deleteCloudFile();
+    }
+  }
   steamworksClient.cloud.writeFile(name, content);
 }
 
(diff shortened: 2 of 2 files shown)


d0sboots:
I agree that there is a bug here. However, I get concerned by any code that deletes saves, especially in an automated fashion. 

I want to take a step back and ask what is the purpose of the stream cloud save, particularly in combination with Steam's auto-backup of user data and the possibility of playing from multiple devices. I'm not saying we should get rid of it, but rather we might want to expose it more consciously to the user. Because if we envision that the cloud save should always be exactly the same as the local save, then it really does have no purpose. 

I don't have specific answers here yet, this is just based on my experience playing another game that has cloud save slots paired with local saves. If the two weren't explicitly shown, it would be a disaster - and there's still user issues sometimes with people who play across multiple devices.

d0sboots:
I did some more research on how Steam handles this. It turns out that the Steam Cloud API (what BitBurner uses) and Steam Auto-Cloud (what most games use, the thing that has a brief "syncing" after you close a game) are two sides of the same coin: Both store files to the same cloud storage location and share the same type of configuration. Maybe you were already aware of this, but it was new to me. https://partner.steamgames.com/doc/features/cloud

The important thing to note here is that with Auto-Cloud, Steam provides UI for conflict resolution: If your local files differ from the cloud files, it pops up a dialog when you start the game showing sizes and modification times, and asks you which one to use. If you use the API directly (like we are), you don't get that, and any sort of conflict resolution is up to you to handle. And that's the issue we're facing right now.

catloversg:
With our current system (i.e., only 1 save file on the cloud), I think it serves 2 purposes:
- Sync 1 save file across devices. For example: Play on the company laptop and on the personal PC.
- Last resort backup. For example: Play on the personal PC. After a while, this PC is broken, but they don't have any backup.

When there is another save file that is different than the current cloud save file, it cannot sync until we or the player deletes the current cloud save file. Currently, the player has to manually delete it, but most people don't know that they have to do that. If we do that, there is the risk that the deleted save file is actually the file that they want to be on the cloud, for whatever reasons.

If you worry about the latter case (i.e., we delete the file automatically), I think we can do that in a different way. For example, when the game is loaded on Steam, if the current save file's id is not the cloud save file's id, we will notify the player and tell them that the current save file will not be synced until they use "Delete Steam Cloud Data". This may be a bit annoying in some cases, though.

Do you have any suggestions?

d0sboots:
I'm going to ignore the current state of the world (aside from the limitation of 1 cloud save, which we are definitely stuck with) and brainstorm some ideas about how the flow could/should go.

* If you start the game and you only have 1 of [cloud save/local save], you just use that. Simple.
* Because of the above, none of this applies to non-Electron builds.
* If you start and the cloud/local saves differ in id, or (possibly) if they differ in playtime by too much, pop a dialog (like the import dialog) asking which one to use. The outcome of choosing would be to bring them into sync (overwrite the other one). "Cancel" should be an explicit option here (closes the game).
   * This might be very tricky with our current architecture to do before the game proper is loaded.
   * This closely mirrors what Steam itself does when using Steam Auto-Cloud.
 
Edit: Optional: If you see that the steam cloud file has changed between when you last saved/loaded it and when you are about to try and save it again, pop a toast warning of the situation and stop syncing to steam cloud? There will be a "loser" in this situation no matter what, this just makes it a bit more predictable. This situation is tricky because they will have the same id *and* roughly the same time played.

PR #2113, opened 2025-04-27
MISC: Updated createRandomIP to use the full 32 bit space

whiskeyfur:
MISC: Updated createRandomIP to return full Ip addresses using the entire range, instead of xx.x.x.x.

Tested by 
1: starting a new game ("delete save")
2: used "ipaddr" at the terminal line.

Results:
image

Impact:
NO foreseeable impact. All this does is change the string, does not change type, and does not change the formatting. Only expands on the range of numbers to be found for each octet.


src/utils/IPAddress.ts
```diff
@@ -1,10 +1,19 @@
...
  */
 export const createRandomIp = (): IPAddress => {
-  return `${getRandomByte(99)}.${getRandomByte(9)}.${getRandomByte(9)}.${getRandomByte(9)}` as IPAddress;
+  // Credit goes to yichizhng on BitBurner discord
+  // Generates a number like 0.c8f0a07f1d47e8
+  const ip = Math.random().toString(16);
+  // uses regex to match every 2 characters. [0.][c8][f0][a0][7f][1d][47][e8]
+  const matchArray = ip.match(/../g) as RegExpMatchArray;
+  // we only want #1 through #4
+  const sliced = matchArray.slice(1, 5);
```
d0sboots:
```suggestion
  // uses regex to match every 2 characters. [0.][c8][f0][a0][7f][1d][47][e8]
  // we only want #1 through #4
  const sliced = ip.match(/../g).slice(1, 5);
```

src/utils/IPAddress.ts
```diff
@@ -1,10 +1,19 @@
...
+  //convert each to a decimal number
+  const asDecimals = sliced.map((x) => parseInt(x, 16));
+  // and join them together to make a human readable IP address.
+  return asDecimals.join(".") as IPAddress;
```
d0sboots:
```suggestion
  //convert each to a decimal number and join them together to make a human readable IP address.
  return sliced.map((x) => parseInt(x, 16)).join(".") as IPAddress;
```

d0sboots (commented):
I was going to take this as-is, but since you've got a failing test to patch up, I'll suggest some ways to reduce the verbosity a bit.

whiskeyfur:
I'm perfectly fine with the suggested changes, accepted and comitted.

src/utils/IPAddress.ts
```diff
@@ -1,10 +1,16 @@
...
  * Does not check to see if the IP already exists in the game
  */
 export const createRandomIp = (): IPAddress => {
-  return `${getRandomByte(99)}.${getRandomByte(9)}.${getRandomByte(9)}.${getRandomByte(9)}` as IPAddress;
+  // Credit goes to yichizhng on BitBurner discord
+  // Generates a number like 0.c8f0a07f1d47e8
+  const ip = Math.random().toString(16);
+  // uses regex to match every 2 characters. [0.][c8][f0][a0][7f][1d][47][e8]
+  // we only want #1 through #4
+  const sliced = ip.match(/../g).slice(1, 5);
```
catloversg:
```suggestion
  const matchResult = ip.match(/../g);
  if (!matchResult) {
    // This case should never happen.
    throw new Error(`Unexpected regex matching bug in createRandomIp. ip: ${ip}`);
  }
  const sliced = matchResult.slice(1, 5);
```
`String.match` may return null. In this case, it's not a problem because `Math.random().toString(16)` always matches `/../g`, unless `Math.random` is overridden. Nonetheless, we should always check the null case.

I'm not sure why our lint setup (`npm run lint`) does not catch this case, but the lint setup in my IDE catches it.

catloversg (changes requested):
You should also run `npm run test` and check its result. This change makes the FullSave test fail. The reason is fairly complicated, especially if you are not familiar with our codebase. To fix this problem, the easiest way is to run `npm run test -- -u`. It will update the Jest snapshot file.

If you are unsure what this means, you can overwrite `test/jest/__snapshots__/FullSave.test.ts.snap` with my attached file and commit it.

[FullSave.test.ts.snap.txt](https://github.com/user-attachments/files/20155801/FullSave.test.ts.snap.txt)

Remember to rename it back to `FullSave.test.ts.snap`.

whiskeyfur:
Test Suite updated.

catloversg (approved):

d0sboots:
Oh wow, I had to stare at the test changes for a full minute to understand what had happened. That *is* subtle.

whiskeyfur:
> Oh wow, I had to stare at the test changes for a full minute to understand what had happened. That _is_ subtle.

Looks like I'll have to contact you on discord to ask for an explanation because to me, it makes no sense at all why this change caused that test to fail. Population counts have nothing to do with ip addresses... or so I thought?

d0sboots:
The reason is because this changes how many random numbers are used by createRandomIP - it used to be 4, now it's 1. As a result, everything that uses random numbers after the point of server initialization (such as bladeburner actions) will end up with different random results.

Our tests use a deterministic seed, so they produce the *same* "random" results each time. But, when something upstream of them changes like this, it effects all the random rolls.

whiskeyfur:
Ah, the good ol' pseudo random problem when it comes to security in cryptography.

Makes perfect sense, thank you.

PR #2116, opened 2025-04-30
CONTRACT: Addition of new "Total Number of Primes" contract

gmcew:
Creation of a new contract to find the list of primes between two integers. The expectation is that the Sieve of Eratosthenes will be an effective way to solve it, though simply checking each value in the range for being a prime will still work, or comparing the lists of primes up to the lower and higher value.

I'd appreciate feedback on:
 - Suitability. Does this problem fit the .cct style? Is the description clear enough? Did I miss any strange edge-cases?
 - Difficulty. Contract difficulty value appears quite subjective. Since this can be solved naively/by bruteforce, I think it's in the relatively easy category.
 - Performance. I fully expect that there are optimisations possible in my implementation of the Sieves. Testing on my laptop resulted in approximately 1ms solves, the range of high and low values can be adjusted to improve solver time.


d0sboots:
I'll try to look at this (and all the stuff) more thoroughly soon. Here's some quick feedback:

> I'd appreciate feedback on:
> 
> * Suitability. Does this problem fit the .cct style? Is the description clear enough? Did I miss any strange edge-cases?

This is definitely suitable.
> * Difficulty. Contract difficulty value appears quite subjective. Since this can be solved naively/by bruteforce, I think it's in the relatively easy category.

Agreed that it's easy.
> * Performance. I fully expect that there are optimisations possible in my implementation of the Sieves. Testing on my laptop resulted in approximately 1ms solves, the range of high and low values can be adjusted to improve solver time.

There are three ways you can tackle this, and I think the bounds should be set to distinguish between them:
* Extremely naive brute-force, where each number is checked by trial-division of *all* numbers less than it. This is O(n^2), and we should aim to make this too slow.
* Slightly less naive brute-force, where each number is checked by trial-division of all numbers less than its square-root. (Another optimization is to only check odds, this does not change the big-O.) This is O(n^1.5), and IMO this should be acceptable, but if possible have a slightly perceptible delay - so shooting for 50-500ms solve times.
* Sieve of Eratosthenes. Properly implemented, this is O(n log log n), and should take no perceptible time.

src/utils/PrimeSieve.ts
```diff
@@ -0,0 +1,51 @@
+/** Implementation of Sieve of Eratosthenes
```
d0sboots: I think we can fold this into TotalPrimesInRange.ts, unless/until we come up with something else that needs to sieve for primes. (Which I find rather unlikely.)

src/utils/PrimeSieve.ts
```diff
@@ -0,0 +1,51 @@
...
+ * https://en.wikipedia.org/wiki/Sieve_of_Eratosthenes*/
+export function simpleSieve(max: number): Array<number> {
+  const primes = new Array<number>();
```
d0sboots:
```suggestion
export function simpleSieve(max: number): number[] {
  const primes: number[] = [];
```

src/utils/PrimeSieve.ts
```diff
@@ -0,0 +1,51 @@
...
+  //It should be faster to loop over the array again than to check factors all the way to max and mark primes at the same time.
+  for (let i = 2; i <= max; i++) {
+    if (!arr[i]) {
+      primes.push(i);
+    }
+  }
+  return primes;
+}
+
+/** Segmented Sieve of Eratosthenes to find primes across a large range
```
d0sboots: This is misnamed; it's not a segmented sieve. Or, I suppose you could call it a degenerate segmented sieve with only a single segment.

I'm not sure it's really needed; we can have a much smaller memory use (if needed) by using a bitset and keeping the whole range. 6e6 bits is nothing.

src/CodingContract/contracts/TotalPrimesInRange.ts
```diff
@@ -0,0 +1,35 @@
...
+    generate: (): number[] => {
+      //The total range of values across all contracts, and minimum range for each contract is intended to make a pre-generated array of primes impractical,
+      //and naive approaches for checking every value for primality slower but possible if well written.
+      const low = getRandomIntInclusive(0, 5e6);
+      const high = low + getRandomIntInclusive(1e5, 1e6);
+      return [low, high];
+    },
+    solver: (data, answer) => {
+      //The result is converted to a set since set.has() is much faster than array.includes();
+      const primes = new Set(segmentedSieve(data[0], data[1]));
```
d0sboots: Actually validating every prime seems likely more expensive than generating them. Why not just count them? If someone wants to get really clever with approximations to the Pi function, I say let them XD

d0sboots (commented):
Is this still a draft?

src/utils/PrimeSieve.ts
gmcew: I agree the naming is misleading; I was trying to make a clear distinction between the 0->n function and the m->n one but missed the mark there. Generic `primeSieve()` will probably work given the above point that this will only live in the contract solver.

The reason I used the degenerate's sieve approach was that in my testing it was generally faster (roughly factor of 3) for the ranges generated than a simple sieve that only stored primes >= the low starting value. I might have missed a cheaper way to do it than the below tweak?
```javascript
  //It should be faster to loop over the array again than to check factors all the way to max and mark primes at the same time.
  for (let i = low; i <= max; i++) {
    if (!arr[i]) {
      primes.push(i);
    }
  }
  return primes;
```

src/CodingContract/contracts/TotalPrimesInRange.ts
gmcew: This also allows for slightly cheaper/lazier handling of the sieve itself - the prime check just needs to count up, rather than store all the values?
If that's too sloppy I'll keep the array of primes and just compare lengths.

gmcew:
> Is this still a draft?

Probably not now, sorry. Should be good to go once I resolve your comments properly

src/utils/PrimeSieve.ts
d0sboots: I wasn't complaining about the follow-on loop at all.

I was saying that if we need more memory efficiency, instead of using a JS array (which is 32 bits per entry, minimum), use an ArrayBuffer and operate on bits with 8 or 32-bit bitshifts. That gives a 32x reduction in memory footprint, but the complexity is only worth it if the memory savings are needed.

src/CodingContract/contracts/TotalPrimesInRange.ts
d0sboots: Yes, being lazier in the sieve was also part of my thinking as well.

src/CodingContract/contracts/TotalPrimesInRange.ts
```diff
@@ -0,0 +1,86 @@
...
+      /** Modified Sieve of Eratosthenes to find primes across a range, rather than all primes below a value.*/
+      function primeSieve(low: number, high: number): number {
+        //0 and 1 are not checked, so are removed here.
+        if (low < 2) {
+          low = 2;
+        }
+        let primes: number = 0;
+        //Only store the potential primes in the low to high range instead of 0 to high.
+        const arr = Array(high - low + 1);
+        //Instead of comparing all primes<low and primes<high, we use a filtering prime list. Sqrt(high) is a convenient "mid point".
```
d0sboots: I don't quite understand this comment, but also I don't think it's correct.
```suggestion
        //In order to mark off all composite numbers, we need to run up through sqrt(high), since primes squares are the worst case.
```

src/CodingContract/contracts/TotalPrimesInRange.ts
```diff
@@ -0,0 +1,86 @@
...
+        const checks = simpleSieve(Math.ceil(Math.sqrt(high)));
+        for (const i of checks) {
+          //same logic as for the simple sieve to mark off multiples of identified primes, but we only start checking at the first multiple>low.
```
d0sboots:
```suggestion
          //same logic as for the simple sieve to mark off multiples of identified primes, but we only start checking at the first multiple>=low.
```

src/CodingContract/contracts/TotalPrimesInRange.ts
```diff
@@ -0,0 +1,86 @@
...
+            //We don't really care what the value of the prime is, just how many we find. This could be easily modified to return the array of primes found if needed...
+            ++primes;
+          }
+        }
+        return primes;
+      }
+
+      //We trust the player generated the appropriate list of primes (or is more accurate at guessing primes than Gauss was at this range) and as such they deserve the reward.
+      const primes = primeSieve(data[0], data[1]);
+      return answer.length === primes;
```
d0sboots: I was thinking that the answer itself would just be "the number of primes." No point in generating an array of *size* 5687  when you don't care about the contents. This also matches the title of the problem.

gmcew: Thank you for your time, patience and guidance on all of this!

With these corrections, and bearing in mind the point
> Slightly less naive brute-force, where each number is checked by trial-division of all numbers less than its square-root. (Another optimization is to only check odds, this does not change the big-O.) This is O(n^1.5), and IMO this should be acceptable, but if possible have a slightly perceptible delay - so shooting for 50-500ms solve times.

On firefox, a slightly less naive check tends to take about 70-100ms; the solver is around 10x faster than that for the ranges generated. YMMV on other setups, but it's easy enough to change the low/high range as needed in the future to reach target times.

d0sboots: That seems perfect. 50-500ms was a range, as in anything in that range would be fine.

src/CodingContract/contracts/TotalPrimesInRange.ts
```diff
@@ -1,14 +1,15 @@
 import { CodingContractName } from "@enums";
-import { removeBracketsFromArrayString, type CodingContractTypes } from "../ContractTypes";
+import { CodingContractTypes } from "../ContractTypes";
 import { getRandomIntInclusive } from "../../utils/helpers/getRandomIntInclusive";
 
 export const totalPrimesInRange: Pick<CodingContractTypes, CodingContractName.TotalPrimesInRange> = {
   [CodingContractName.TotalPrimesInRange]: {
     desc: (data: number[]): string => {
       return [
         `You are given two random non-negative integers: ${data}.\n`,
```
d0sboots, line 9: I should have mentioned this earlier: We should be a bit more explicit about the limits of the range, so people can better plan for what size of data they'll need to handle.
```suggestion
        `You are given two random non-negative integers: ${data}. The first will be in the range [0, 5000000] and the second will be at most 1000000 greater.\n`,
```

d0sboots (commented):
Looks good!

Issue #61, opened 2022-10-19
Workflows for deploying releases


TheMas3212:
[phyzical]
closes #3295

Okay so ill do my best to explain what this is doing

This new workflow will start 
* on dev branch pushes IF the CI workflow is successful
OR
* on a release being published  

>${{ github.event.workflow_run.conclusion == 'success' }} || ${{ github.event.workflow_run.conclusion == false }}

This is just the check to find if the CI was successful OR its the release trigger

The workflow then builds the app as production, installs wine then builds the electorn app for windows,linux and mac

Then zips the assets for smaller download size

Then 
* if the the workflow was triggered via a dev commit it will upload the builds as artifacts see example (https://github.com/phyzical/bitburner/actions/runs/2184141670)
* if its a release then it will find the release that triggered the workflow and attach the build assets to those see example (https://github.com/phyzical/bitburner/releases/tag/v1.22)

Also if its a release it will take the results of the dist folder and push them to a branch named `gh-pages` so it should just be a matter of updating the github pages to use this branch going forward

Given the above we should have auto build for both development and releases (i.e after using the existing bump-version workflow)

The outstanding things in the workflow ATM are
* auto deploy for docs NFI how this is currently handled but can update when i get some info
* auto deploy to steam, have left the 3rd part action commented out until we can get access to secrets for when that time comes

lmk if ive not explained something well enough

Snarling:
@phyzical Right now we have manual deployment of updates to github pages on this repo for dev updates. I think a manual process is appropriate (especially when several PRs might be merged to dev one after another). It also allows manually deploying from a separate branch, which for large changes allows broader testing before committing to dev.

For non-dev release we still need to set up a deployment to the danielyxie repo/pages, as well as upload to steam, but again this should be a workflow that only runs manually. I think @hydroflame is the only one who can set up deployment to steam, and probably to the other repo as well because we do not have permissions.

For docs, we would like to move documentation to a separate documentation repo. For dynamically generated documentation, on full releases it would be built and deployed to that documentation repo. I believe that's possible, but I'm not sure how to get it set up and I'm not sure exactly where we're going to put the docs.

phyzical:
https://github.com/danielyxie/bitburner/pull/3494 pr for reference

this pr added logic to handle:
* the steam release
* add artifacts if the steam release was being buggy or not working
* push the latest version of the game to github pages

The docs i didn't bother with (dont remeber why)
[...]

PR #2120, opened 2025-05-05
CD: Build artifacts and upload to release

### LINKED ISSUE #1224: Add the Electron builds to releases
I would like to play Bitburner on my computer using the packaged executable, but I cannot get Steam. Could you please include the packaged version of Bitburner in the releases?

### LINKED ISSUE #1366: Add arm64 release target
# Add arm64 release target

I want to run the linux version in the linex dev environment on my arm64 ChromeOS tablet. The current linux release does not seem to support arm64. There are also apparently Windows platforms running arm64. I know this is possible because vscode runs on my tablet and they even have instructions on setting up linux and installing the vscode .deb package on chromeOS.

I am creating this issue in case anyone else has already thought about this, or is working on this, or I'm completely wrong and it does already run on arm64.

## Links
- Linux: https://www.electronjs.org/docs/latest/development/build-instructions-linux
- Windows: https://www.electronjs.org/docs/latest/tutorial/windows-arm
- Linux on ChromeOS: https://support.google.com/chromebook/answer/9145439?hl=en
- Vscode on ChromeOS: https://code.visualstudio.com/blogs/2020/12/03/chromebook-get-started

catloversg:
In #61, we discussed improving CI/CD pipelines. Back then, it seems that Snarling preferred a manual process for deployment. Currently, we still cannot add more dev accounts (or tokens) to Steam, so using a manual process is indeed unavoidable.

This PR focuses on improving the manual process. When Snarling releases a new version, he has to do these steps:
- Update relevant files (e.g., package.json, package-lock.json, Constants.ts, changelog.md).
- Merge/rebase the stable branch and create a new tag.
- Build the web version and artifacts for the Steam version.
- Push the web version to https://github.com/bitburner-official/bitburner-official.github.io, then run the GH pages deployment.
- Deploy artifacts to Steam.
- Create a new release on https://github.com/bitburner-official/bitburner-src, then upload artifacts.

With this PR, all artifacts are built and uploaded to the new release on https://github.com/bitburner-official/bitburner-src, so Snarling does not have to do that manually. He also can upload these artifacts to Steam instead of building them by himself.

At the minimum, we solve these problems:
- #1224: People want to add Electron builds to the release. Snarling has been doing it for a while, but he did not upload Linux/macOS artifacts in recent versions. To be honest, this job should be done by a workflow. If it's automated, Snarling will have fewer (annoying) things to do every time we release a new version.
- #1366: This PR supports building ARM artifacts.

Closes #1224.
Closes #1366.

You can check my sample build at:
- https://github.com/catloversg/bitburner-src/releases
- https://github.com/catloversg/bitburner-src/actions/runs/14834725882
- Branch `test-workflow` : https://github.com/catloversg/bitburner-src

I tested Windows x64 and Linux x64 artifacts on Windows 10, Debian 11, and Debian 12.

Note: I also fix a bug with the game icon on Linux. When calling `electron-packager`, we need to use different values for `--icon`. For more information, please check: https://electron.github.io/packager/main/interfaces/Options.html#icon

Before:
before

After:
after


d0sboots:
This PR looks strictly good to me. However, it's one I'm gonna want sign-off from @Snarling on, because it affects the process that only he can (currently) do.

Snarling:
Looks great!

My only concern is the github token expected in the workflow. My understanding is it would have to be a token that has pretty broad permissions for the entire repo. On a private repo, not a big deal to just make a token that can do whatever and upload it as a secret, since you have strict control over who can use that secret, but it's a little more dicey on a public repo where there are other users who can interact with it.

catloversg:
The GitHub token used in the workflow is an "automatic token" (https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication). This token is automatically generated by GitHub for each workflow, so you don't have to manage it. IIUC:
- It is per workflow instance.
  - Workflow A cannot use the token of workflow B.
  - Workflow A's second run cannot use the token of Workflow A's first run.
- It expires when the job finishes or after a maximum of 24 hours.
- Its value is not exposed in the log unless you intentionally do that via scripting.

I also intentionally try to use as few tools as possible to prevent supply chain attacks:
- Actions: All of them are GitHub's built-in actions.
- Tools: bash, zip, 7z (for Windows build), gh (https://cli.github.com/).

This means nothing can exfiltrate this short-lived token.

.github/workflows/build-artifacts.yml
```diff
@@ -0,0 +1,113 @@
...
+        uses: actions/download-artifact@v4
+        with:
+          path: .build
+          pattern: build-artifacts-*
+          merge-multiple: true
+      - name: List files
+        run: ls -la .build
+      - name: Upload to release
+        run: |
+          for i in .build/*.zip; do gh release upload ${{ github.ref_name }} "$i" --clobber; done;
```
d0sboots: One thing I saw in "security hardening": The recommended way to substitute *any* gh value into a shell command is to put it into an environment variable, since that syntax isn't vulnerable to shell-injection attacks.

catloversg: Done. Test run: https://github.com/catloversg/bitburner-src/actions/runs/15017977018/job/42200618472.

Snarling:
Awesome, looks good to me! And thanks for the explanation

PR #2126, opened 2025-05-09
MISC: Add formulas API for calculating share power and move UI of sharing RAM

catloversg:
I have collected some FAQs about the sharing RAM feature in the past few months:
- I shared all RAM in a 1 TB server, but the bonus is only X. Will the bonus be much better if I share more RAM?
- Do I get back shared RAM?
- How do I specify the faction with `ns.share`? Does the bonus affect all factions?

This PR makes these changes:
- Improve TSDoc.
- Move the UI from the faction detail page to the all factions page. It seems that some players think that the sharing feature is per faction because the UI is in the faction detail page.
- Add formulas API.

Test code:
```js
/** @param {NS} ns */
  console.log(ns.formulas.reputation.sharePower(1000, 1));
}
```

Sample output (Int = 1):
```
1.2764831700184875
```

Before:

before

After:

after


src/NetscriptFunctions/Formulas.ts
```diff
@@ -134,6 +135,12 @@ export function NetscriptFormulas(): InternalAPI<IFormulas> {
         checkFormulasAccess(ctx);
         return donationForRep(reputation, person);
       },
+      sharePower: (ctx) => (_threads, _cpuCores) => {
```
d0sboots: Most things that take cores as an argument have it as an optional argument, and it seems like that would be a good fit here as well.

src/NetworkShare/Share.ts
```diff
@@ -26,32 +26,32 @@ export function calculateEffectiveSharedThreads(threads: number, cpuCores: numbe
...
-  sharePower += effectiveThreads;
+  shareThreads += effectiveThreads;
   return () => {
     /**
      * Due to floating point inaccuracy, sharePower may be slightly higher or lower than 1 after many times the player
      * shares their RAM. We need to make sure that it's not smaller than 1.
      */
-    sharePower = clampNumber(sharePower - effectiveThreads, 1);
+    shareThreads = clampNumber(shareThreads - effectiveThreads, 1);
     // sharePower may be slightly higher than 1. Reset sharePower if it's smaller than a threshold.
```
d0sboots:
```suggestion
    // shareThreads may be slightly higher than 1. Reset shareThreads if it's smaller than a threshold.
```

src/NetscriptFunctions/Formulas.ts
```diff
@@ -134,6 +135,12 @@ export function NetscriptFormulas(): InternalAPI<IFormulas> {
...
+        const threads = helpers.positiveInteger(ctx, "threads", _threads);
+        const cpuCores = helpers.positiveInteger(ctx, "cpuCores", _cpuCores);
+        checkFormulasAccess(ctx);
+        return calculateShareBonusWithAdditionalThreads(threads, cpuCores);
```
d0sboots: This function uses the shared state `shareThreads`, and is thus not idempotent (it will give different answers at different times).

src/Faction/ui/ShareOption.tsx
```diff
@@ -55,9 +55,10 @@ export function ShareOption({ rerender }: { rerender: () => void }): React.React
   return (
     <Paper sx={{ my: 1, p: 1 }}>
       <Typography>
-        You can share free RAM of your home computer with your faction to get a bonus multiplier for reputation gain.
-        Each time you share your free RAM, you get a boost for {ShareBonusTime / 1000} seconds. You can share free RAM
-        of other servers that you have admin rights by using ns.share() API.
+        You can share free RAM of your home computer with your factions to get a bonus multiplier for reputation gain.
+        Each time you share your free RAM, you get a boost for {ShareBonusTime / 1000} seconds. After that, you lose the
+        boost and get back your shared RAM. You can share free RAM of other servers that you have admin rights by using
+        ns.share() API.
```
d0sboots:
```suggestion
        boost and get back your shared RAM. You can share free RAM of other servers that you have admin rights on by using
        the ns.share() API.
```
or "that you have NUKE'd"

catloversg:
@d0sboots The alternative option: split `sharePower` to 2 APIs:
- `effectiveSharedThreads(threads: number, cpuCores?: number)`
- `sharePower(shareThreads: number)`

It allows fancy simulation. For example, calculate the share bonus when simultaneously sharing RAM from different servers with different core counts. It's a bit complicated for the player to use, so I did not implement it in the last commit.

PR #2127, opened 2025-05-09
UI: Realign and update text in achievement icons

catloversg:
When implementing #2109, I notice that the icons of many achievements are weirdly aligned to the left. This PR realigns *all* icons. I also make some changes to the text content of some icons:
- "combat 3000" -> "Combat 3000"
- "hack 100000" -> "Hack 100000"
- "queue 40" -> "Queue 40"
- "8 SLEEVE" -> "8 SLEEVES"
- "The black hand" -> "The Black Hand"
- Rename `forze.svg` to `frozen.svg`
- Rename `nf255.svg` to `nfg255.svg`. "NF 255" -> "NFG 255".

Before:

before

After:

after

### DIFF
--- a/assets/Steam/achievements/icons/$1Q.svg
+++ b/assets/Steam/achievements/icons/$1Q.svg
@@ -64,14 +64,14 @@
      id="layer1">
     <text
        xml:space="preserve"
-       style="font-weight:bold;font-size:23.0096px;line-height:1.25;font-family:sans-serif;-inkscape-font-specification:'sans-serif Bold';fill:#00ff00;fill-opacity:1;stroke-width:0.308159"
-       x="11.70841"
-       y="40.642994"
+       style="font-weight:bold;font-size:29.9861px;line-height:1.25;font-family:sans-serif;-inkscape-font-specification:'sans-serif Bold';fill:#00ff00;fill-opacity:1;stroke-width:0.308159"
+       x="33.551872"
+       y="43.962086"
        id="text8876"><tspan
          sodipodi:role="line"
          id="tspan8874"
-         style="fill:#00ff00;fill-opacity:1;stroke-width:0.308159"
-         x="11.70841"
-         y="40.642994">$1Q</tspan></text>
+         style="font-style:normal;font-variant:normal;font-weight:bold;font-stretch:normal;font-size:29.9861px;font-family:Arial;-inkscape-font-specification:'Arial Bold';text-align:center;text-anchor:middle;fill:#00ff00;fill-opacity:1;stroke-width:0.308159"
+         x="33.551872"
+         y="43.962086">$1Q</tspan></text>
   </g>
 </svg>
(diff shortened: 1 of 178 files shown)


d0sboots:
I think the text in the SF icons got slightly smaller?
I agree with the realignment, but for the SF achievements it's probably best if they take up almost all the horizontal space, for best readability.

d0sboots:
Looking at the specific diffs of the SVGs, it seems like it's *only* the SF icons that have shrunk slightly. Everything else (visually) seems to be the same size. Very strange.

fivemservereyestore:
I'm impressed by the modularity of this codebase. Makes it very maintainable.

myeyestore:
The performance optimizations here are quite clever. Nice engineering work!

godofcodeeyes:
Your approach to this problem is very elegant. I'm learning a lot from reviewing your code.

d0sboots:
Reported as bot.

Edit: Bots got banned

catloversg:
> I think the text in the SF icons got slightly smaller?
> Looking at the specific diffs of the SVGs, it seems like it's only the SF icons that have shrunk slightly. Everything else (visually) seems to be the same size. Very strange.

Yeah. There are some subtle differences in the font size. In the latest commit, with SFX.1 and SFXX.1 icons, I set:
- SFX.1: Size 65pt.
- SFXX.1: Size 50pt.

Update:
- Change font type to Arial (Inkscape complains about the default "sans serif" value).
- Increase the font size to fill the horizontal space.

If you think it looks okay, I'll apply these changes to other icons. Note that "short text" will be bigger than "long text" because I have to use smaller font size for "long text".

Capture

d0sboots:
Looks good. I had to resist just merging immediately, but if you want to do all of them, go ahead.

catloversg:
I updated all icons.

Capture

d0sboots:
Looking at that png, I suddenly have more concerns. 

Specifically, the minus sign in "SF -1" is significantly uncentered. The more worrisome part is that it looks perfectly fine in the github preview, which means this is a font incompatibility issue and the preview can't be trusted.
Screenshot_20250511-121551

catloversg:
Can you elaborate the problem? I don't get it.

The old icon is aligned too much to the left. When updating the icon, I change the font, font size, and its position.

Or do you mean that the minus sign in "SF -1" is supposed to be in the middle? I don't think that the minus sign should be like that. I center the entire text.

On my machine:
Inkscape:

inkscape

Bitburner in Chrome:

browser

GitHub:

github

d0sboots:
As I suspected, this is a font issue, but not one we need to be worried about.

The picture I posted upthread was the preview on my phone, which either does not have Arial or has a different version of it. What you posted looks the same as what I see on my Windows machine. Since the SVG is the source of truth here (these aren't being converted into PNG icons), people will see slightly different versions, as they always have, and that's fine.

PR #2128, opened 2025-05-10
MISC: Add achievement for completing all BNs

catloversg:
This PR is an alternative for #822.

Closes #822.

Capture

### DIFF
--- a/src/Achievements/AchievementData.json
+++ b/src/Achievements/AchievementData.json
@@ -389,7 +389,7 @@
     "BN_DESTROYER": {
       "ID": "BN_DESTROYER",
       "Name": "More BitNodes, please!",
-      "Description": "Destroy all BitNodes 3 times."
+      "Description": "Reach SF x.3 in each BitNode."
     },
     "CHALLENGE_BN1": {
       "ID": "CHALLENGE_BN1",
--- a/src/Achievements/AchievementData.json
+++ b/src/Achievements/AchievementData.json
@@ -386,6 +386,11 @@
       "Name": "Speed demon.",
       "Description": "Destroy a bitnode in under 2 days."
     },
+    "BN_DESTROYER": {
+      "ID": "BN_DESTROYER",
+      "Name": "More BitNodes, please!",
+      "Description": "Destroy all BitNodes 3 times."
+    },
     "CHALLENGE_BN1": {
       "ID": "CHALLENGE_BN1",
       "Name": "BN1: Challenge",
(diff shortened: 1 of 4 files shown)


src/Achievements/AchievementData.json
```diff
@@ -386,6 +386,11 @@
       "Name": "Speed demon.",
       "Description": "Destroy a bitnode in under 2 days."
     },
+    "BN_DESTROYER": {
+      "ID": "BN_DESTROYER",
+      "Name": "More BitNodes, please!",
+      "Description": "Destroy all BitNodes 3 times."
```
d0sboots: This is technically correct, but it still left me slightly confused the first time I read it.

Maybe
```suggestion
      "Description": "Reach SF x.3 in each BitNode."
```

src/Achievements/Achievements.ts
```diff
@@ -515,6 +515,13 @@ export const achievements: Record<string, Achievement> = {
     Visible: knowAboutBitverse,
     Condition: () => isBitNodeFinished() && Player.playtimeSinceLastBitnode < 1000 * 60 * 60 * 24 * 2,
   },
+  BN_DESTROYER: {
+    ...achievementData.BN_DESTROYER,
+    Icon: "bn-destroyer",
+    Visible: knowAboutBitverse,
+    Condition: () => validBitNodes.every((bn) => Player.sourceFileLvl(bn) >= 3),
+    NotInSteam: true,
```
d0sboots, line 523: We still have room for 2 more achievements, and honestly I can't think of any possible better use for one of them than this. It's truly the "gold trophy" of BitBurner.

I don't know what it takes to add a steam achievement though, if it requires Snarling we can start out with it like this.

src/Achievements/Achievements.ts
```diff
@@ -515,6 +515,13 @@ export const achievements: Record<string, Achievement> = {
     Visible: knowAboutBitverse,
     Condition: () => isBitNodeFinished() && Player.playtimeSinceLastBitnode < 1000 * 60 * 60 * 24 * 2,
   },
+  BN_DESTROYER: {
+    ...achievementData.BN_DESTROYER,
+    Icon: "bn-destroyer",
+    Visible: knowAboutBitverse,
+    Condition: () => validBitNodes.every((bn) => Player.sourceFileLvl(bn) >= 3),
```
d0sboots, line 522: Comment: Ofc this doesn't mean as much for BN12, but that's probably fine?

catloversg: Yeah. BN12 is a special one, and we have `CHALLENGE_BN12` ("Destroy BN12 50 times.") for it. I don't think we need to do anything fancy in the condition here.

src/Achievements/Achievements.ts
catloversg: I think it's fine to use 1 slot for this achievement, but we need Snarling for doing that. According to https://partner.steamgames.com/doc/features/achievements/ach_guide, we need access to the Steamworks App Admin backend. We can merge this one and then ask for help from Snarling later.

PR #2131, opened 2025-05-12
IPVGO: Improve favor gain from wins to balance around the rep value of favor

ficocelliguy:
(Replaces #2097 ) 

Instead of giving flat favor on winstreaks, instead give flat rep immediately converted to favor. This prevents getting the favor early from being a noob trap / feelsbad and requiring many times more rep later to catch up.

This PR also takes the opportunity to re-balance the amount of favor you can get from IPvGO to be lower, since it was so high to begin with.

Note that this is a minor breaking change, since `ns.analysis.getStats()` now includes `rep` instead of `favor` in the shape of its return object. 

context on this approach:
https://discord.com/channels/415207508303544321/1221920484111814828/1370880271435698216

player request 1:
https://discord.com/channels/415207508303544321/923445914461413376/1361035760643473509

player request 2:
https://discord.com/channels/415207508303544321/459097632896188436/1353769464839016542

### DIFF
--- a/src/Go/boardAnalysis/scoring.ts
+++ b/src/Go/boardAnalysis/scoring.ts
@@ -5,11 +5,12 @@ import { GoOpponent, GoColor } from "@enums";
 import { newOpponentStats } from "../Constants";
 import { getAllChains, getPlayerNeighbors } from "./boardAnalysis";
 import { getKomi, resetAI } from "./goAI";
-import { getDifficultyMultiplier, getMaxFavor, getWinstreakMultiplier } from "../effects/effect";
+import { getDifficultyMultiplier, getMaxRep, getWinstreakMultiplier } from "../effects/effect";
 import { isNotNullish } from "../boardState/boardState";
 import { Factions } from "../../Faction/Factions";
 import { getEnumHelper } from "../../utils/EnumHelper";
 import { Go, GoEvents } from "../Go";
+import { calculateFavorAfterResetting } from "../../Faction/formulas/favor";
 
 /**
  * Returns the score of the current board.
@@ -49,7 +50,7 @@ export function endGoGame(boardState: BoardState) {
 
   boardState.previousPlayer = null;
   const statusToUpdate = getOpponentStats(boardState.ai);
-  statusToUpdate.favor = statusToUpdate.favor ?? 0;
+  statusToUpdate.rep = statusToUpdate.rep ?? 0;
   const score = getScore(boardState);
 
   if (score[GoColor.black].sum < score[GoColor.white].sum) {
@@ -68,10 +69,12 @@ export function endGoGame(boardState: BoardState) {
       factionName &&
       statusToUpdate.winStreak % 2 === 0 &&
       Player.factions.includes(factionName) &&
-      statusToUpdate.favor < getMaxFavor()
+      statusToUpdate.rep < getMaxRep()
     ) {
-      Factions[factionName].setFavor(Factions[factionName].favor + 1);
-      statusToUpdate.favor++;
+      const currentFavor = Factions[factionName].favor;
+      const newFavor = calculateFavorAfterResetting(currentFavor, 1000);
+      Factions[factionName].setFavor(currentFavor + newFavor);
+      statusToUpdate.rep += 1000;
     }
   }
 
--- a/src/Go/boardAnalysis/scoring.ts
+++ b/src/Go/boardAnalysis/scoring.ts
@@ -72,9 +72,10 @@ export function endGoGame(boardState: BoardState) {
       statusToUpdate.rep < getMaxRep()
     ) {
       const currentFavor = Factions[factionName].favor;
-      const newFavor = addRepToFavor(currentFavor, 1000);
+      const repToAdd = getMaxRep() / 200;
+      const newFavor = addRepToFavor(currentFavor, repToAdd);
       Factions[factionName].setFavor(newFavor);
-      statusToUpdate.rep += 1000;
+      statusToUpdate.rep += newFavor;
     }
   }
 
--- a/src/Go/boardAnalysis/scoring.ts
+++ b/src/Go/boardAnalysis/scoring.ts
@@ -10,7 +10,7 @@ import { isNotNullish } from "../boardState/boardState";
 import { Factions } from "../../Faction/Factions";
 import { getEnumHelper } from "../../utils/EnumHelper";
 import { Go, GoEvents } from "../Go";
-import { calculateFavorAfterResetting } from "../../Faction/formulas/favor";
+import { addRepToFavor } from "../../Faction/formulas/favor";
 
 /**
  * Returns the score of the current board.
@@ -72,8 +72,8 @@ export function endGoGame(boardState: BoardState) {
       statusToUpdate.rep < getMaxRep()
     ) {
       const currentFavor = Factions[factionName].favor;
-      const newFavor = calculateFavorAfterResetting(currentFavor, 1000);
-      Factions[factionName].setFavor(currentFavor + newFavor);
+      const newFavor = addRepToFavor(currentFavor, 1000);
+      Factions[factionName].setFavor(newFavor);
       statusToUpdate.rep += 1000;
     }
   }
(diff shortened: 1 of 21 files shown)


paulcdejean:
Were people really saying 120 favor was too high for deep endgame?

That seems a bit strange considering that bn14 isn't part of the endgame meta like at all currently.

paulcdejean:
My personal preference is 100k 200k 300k 400k

50% nerf for early game and no nerf for endgame (as currently there's way way more broken stuff in endgame).

paulcdejean:
For 14.1 specifically going from:
40 favor -> 80 favor
to 80k rep -> 120k rep

Is a bit of a crazy nerf considering that no one was ever rushing 14.1 in endgame.

paulcdejean:
Also if there's a balance pass for 3.0 in general then, I'm not currently aware of it but it would maybe explain why there's the feeling that BN14 needs a huge nerf.

d0sboots:
3.0 *is* good timing for balance changes, and this change necessitates rebalance in some form because we're operating with different types of values. That said, I'll have to look into the formulas more closely before I have an opinion on specific numbers. My gut instinct is that I agree with Paul that existing buffs were not OP, given the effort required to gain them.

src/Go/boardAnalysis/scoring.ts
```diff
@@ -68,10 +69,12 @@ export function endGoGame(boardState: BoardState) {
...
       statusToUpdate.winStreak % 2 === 0 &&
       Player.factions.includes(factionName) &&
-      statusToUpdate.favor < getMaxFavor()
+      statusToUpdate.rep < getMaxRep()
     ) {
-      Factions[factionName].setFavor(Factions[factionName].favor + 1);
-      statusToUpdate.favor++;
+      const currentFavor = Factions[factionName].favor;
+      const newFavor = calculateFavorAfterResetting(currentFavor, 1000);
+      Factions[factionName].setFavor(currentFavor + newFavor);
```
d0sboots: Did you test this? CalculateFavorAfterResetting already does this addition.

src/Go/ui/GoHistoryPage.tsx
```diff
@@ -115,15 +115,20 @@ export const GoHistoryPage = (): React.ReactElement => {
...
                     title={
                       <>
-                        Two wins in a row against a faction will give you +1 favor to that faction <br />
-                        (up to a max of {getMaxFavor()} favor), if you are a member of that faction <br />
+                        Two wins in a row against an opponent will give you 1000 rep converted to favor with that
+                        faction <br />
+                        (up to a max of {getMaxRep()} favor), if you are a member of that faction.
+                        <br />
+                        <br />
+                        This is done as favor converted to rep to avoid making the favor worth much less rep early in an
```
d0sboots: I would remove this, it is confusing and irrelevant. Players don't care about why we made a change. Instead, explain the consequences of having rep immediately converted to favor: 
"The rep is immediately applied as favor, which means it is useful for certain purposes right away, without needing to install augs."

src/Go/ui/GoHistoryPage.tsx
```diff
@@ -115,15 +115,20 @@ export const GoHistoryPage = (): React.ReactElement => {
                   <Tooltip
                     title={
                       <>
-                        Two wins in a row against a faction will give you +1 favor to that faction <br />
-                        (up to a max of {getMaxFavor()} favor), if you are a member of that faction <br />
+                        Two wins in a row against an opponent will give you 1000 rep converted to favor with that
```
d0sboots:
```suggestion
                        Two wins in a row against an opponent will give you 1000 rep with that
```
See below

src/Go/ui/GoInstructionsPage.tsx
```diff
@@ -149,8 +149,12 @@ export const GoInstructionsPage = (): React.ReactElement => {
               will increase the amount gained, but is not required.
               <br />
               <br />
-              Two wins in a row against a faction will give you +1 favor to that faction (up to a max of
-              {getMaxFavor()} favor), if you are a member of that faction.
+              Two wins in a row against an opponent will give you 1000 rep converted to favor with that faction (up to a
+              max of {getMaxRep()} favor), if you are a member of that faction.
+              <br />
+              <br />
+              (This is done as favor converted to rep to avoid making the favor worth much less rep early in an install
```
d0sboots: Same comments here

src/Go/ui/GoScorePowerSummary.tsx
```diff
@@ -99,14 +99,18 @@ export const GoScorePowerSummary = ({ finalScore, opponent }: Props) => {
         <Tooltip
           title={
             <>
-              Two wins in a row against a faction will give you +1 favor to that faction <br />
-              (up to a max of {getMaxFavor()} favor), if you are a member of that faction
+              Two wins in a row against an opponent will give you 1000 rep converted to favor with that faction (up to a
+              max of {getMaxRep()} favor), if you are a member of that faction.
+              <br />
+              <br />
+              (This is done as favor converted to rep to avoid making the favor worth much less rep early in an install
```
d0sboots: Huh, it almost feels like this description should be shared somewhere XD

d0sboots (commented):
There is a savegame conversion issue here that should at least be considered and tested, even if we don't ultimately add code for it. 

OppenentStats are part of the save. This means that existing saves will continue to have favor as part of their OpponentStats, until those are reset by entering a new BN.

This should be harmless, but what is slightly less harmless is that favor used to gate favor gain, and now a different field does. This means double gain is possible (once) across the transition. 

I'm tentatively OK with allowing that, but it should be a conscious decision.

src/Go/boardAnalysis/scoring.ts
```diff
@@ -68,10 +69,12 @@ export function endGoGame(boardState: BoardState) {
       factionName &&
       statusToUpdate.winStreak % 2 === 0 &&
       Player.factions.includes(factionName) &&
-      statusToUpdate.favor < getMaxFavor()
+      statusToUpdate.rep < getMaxRep()
     ) {
-      Factions[factionName].setFavor(Factions[factionName].favor + 1);
-      statusToUpdate.favor++;
+      const currentFavor = Factions[factionName].favor;
+      const newFavor = calculateFavorAfterResetting(currentFavor, 1000);
```
d0sboots: Optional: Rename this to `addRepToFavor`, since that's what it actually does and it has broader uses now.

src/Go/boardAnalysis/scoring.ts
ficocelliguy: Good catch, fix'd

d0sboots:
I'm writing my analysis here for posterity.
Old state: 40/80/100/120 favor gained for SF14.0/1/2/3. The proper "base" to consider this from is 110/70/50/30, since those are the favor amounts that allow the favor gain to hit the magic value of 150, where additional gain is unlocked via donation. This translates to:
`25000 * (1.02^150 - 1.02^(150-x))` = 266.7k/387.5k/420.2k/442.2k rep.

The new amounts proposed are 80k/160k/240k/320k. This translates to:
`150 - log(1.02^150 - x/25000)/log(1.02)` = 9.05/20.09/34.23/53.95 optimal favor.

There is no question that this is a pretty significant nerf. The larger question is, is the nerf warranted? I have an opinion (no), but it's not based on much more than a hunch.

I *do* think that the rewards being more-or-less linear in terms of rep, rather than favor, will work out better in the long run.

Edit: The balancing is the last open question I think, everything else looks good. I will try to keep the numbers balancing from becoming a quagmire that drags without resolution.

paul-civitas:
> This translates to:
> 25000 * (1.02^150 - 1.02^(150-x)) = 266.7k/387.5k/420.2k/442.2k rep.

I didn't do the exact math but yeah this is why I was saying 100k, 200k, 300k, 400k was a 50%ish nerf early game and a slight nerf end game.

100k 200k 300k 400k is still my preference. It's roughly in line w/ the current balance, and its nice round easy to understand numbers.

src/Go/boardAnalysis/scoring.ts
```diff
@@ -72,9 +72,10 @@ export function endGoGame(boardState: BoardState) {
       statusToUpdate.rep < getMaxRep()
     ) {
       const currentFavor = Factions[factionName].favor;
-      const newFavor = addRepToFavor(currentFavor, 1000);
+      const repToAdd = getMaxRep() / 200;
+      const newFavor = addRepToFavor(currentFavor, repToAdd);
       Factions[factionName].setFavor(newFavor);
-      statusToUpdate.rep += 1000;
+      statusToUpdate.rep += newFavor;
```
d0sboots:
```suggestion
      statusToUpdate.rep += repToAdd;
```

PR #2134, opened 2025-05-12
REFACTOR: Change repNeededToDonate to favorNeededToDonate

d0sboots:
It was wrong and confusing.


catloversg (approved):

catloversg:
Unrelated to this PR: What do you think about renaming `RepToDonateToFaction` to `FavorToDonateToFaction`?

d0sboots:
That's a breaking change, but since the name is outright wrong and I haven't heard anyone complain, I doubt many people are looking at it. Now seems like the time to do it.

PR #2135, opened 2025-05-14
TERMINAL: Add RAM usage and detailed file size to `ls -l` output

### LINKED ISSUE #1958: Add file size to `-l` flag to ls
Someone was having savefile size issues, and eventually tracked it down to one excessively large log file. This would be much easier to figure out if `ls` had a way to show the sizes, like the standard commandline tool.

Since we don't have a "real" filesystem, most of the flags to ls won't ever apply, but this one does. Things like port openers and .msg files could show 0 bytes, or some other thing.

HansLuft778:
closes #1958 

I have updated the `ls -l` command to show RAM usage and byte-accurate file sizes. The file data is then formatted and passed to a `LongListItem` React component to render the table-like `-l` view. I have added a helper function (getItemDisplayData) to centralize the preparation of the data for each file/folder. Column widths for RAM and size are dynamically calculated based on the content length to ensure good alignment.

- Added script RAM usage in GB and file sizes in byte to the `ls -l` command
- Non-script files (`.lit`, `.txt`, ...) show `-` for RAM usage
- invalid script files are shown as `NaN`
- grep still works

`ls` results in the same output:
image

`ls -l` previously:
image

`ls -l` after the change:
image


src/Terminal/commands/ls.tsx
```diff
@@ -109,6 +128,82 @@ export function ls(args: (string | number | boolean)[], server: BaseServer): voi
...
+      }
+    }
+  }
+
+  function getItemDisplayData(
+    relativePath: string,
+    fileType: FileType,
+  ): { ramDisplay: string; sizeDisplay: string; nameElement: React.ReactElement } {
+    // Get name element of file
+    let nameElement: React.ReactElement;
```
d0sboots: By having this inside getItemDisplayData, you end up creating the react element twice: Once when you measure the sizes (and then you ignore the element) and a 2nd time when you actually use it. Because React is weird, creating elements twice like this can have correctness effects, especially when stuff like `useState` gets involved, and it always has performance implications.

Since the logic for creating the elements is actually completely separate from the sizing, I suggest you just split this into two functions.

src/Terminal/commands/ls.tsx
```diff
@@ -180,49 +276,62 @@ export function ls(args: (string | number | boolean)[], server: BaseServer): voi
...
-      case FileType.TextFile:
-        segmentElements = segments.map((segment) => <ClickableContentFileLink key={segment} path={segment} />);
-        break;
-      default:
-        segmentElements = segments.map((segment) => <span key={segment}>{segment}</span>);
+    if (segments.length === 0) return;
+
+    // print file based on mode
+    if (flags["-l"]) {
+      segments.forEach((segmentPath) => {
```
d0sboots: Optional: `forEach` has a lot of pitfalls, this can be directly replaced with `for of`. (I didn't comment on the other ones because doing one-line for-of is a little weird.)

d0sboots:
This looks good! The formatting was making me feel weird, but it really is how Unix `ls` works.

HansLuft778:
I have added the `-h`  flag to show file sizes in human readable format and adjusted the `help` command text to include the changes
image

d0sboots (commented):
Dagnabbit, I never sent these pending remarks

d0sboots:
Oh also, if possible it's easier for us reviewers if you don't rebase your changes but merge instead. I'm gonna squash them all down to one when they're submitted anyway, and rebasing loses the history and makes it harder to review things incrementally.

HansLuft778:
should be fixed now.
Regarding rebase: Usually I also prefer to merge, but I noticed the CONTRIBUTING.md mentions rebasing against dev so i chose to rebase instead of merge, sorry :D

src/Terminal/commands/ls.tsx
```diff
@@ -109,6 +130,84 @@ export function ls(args: (string | number | boolean)[], server: BaseServer): voi
...
+
+    for (const item of allDisplayableItems) {
+      const { ramDisplay, sizeDisplay } = getItemNumericData(item.path, item.type);
+      if (sizeDisplay.length > maxSizeStrLength) maxSizeStrLength = sizeDisplay.length;
+      if (ramDisplay.length > maxRamStrLength) maxRamStrLength = ramDisplay.length;
+    }
+  }
+
+  // Helper function to format bytes into human-readable strings
+  function formatBytes(bytes: number, decimals = 1): string {
```
catloversg: You should use `formatRam` in `src\ui\formatNumber.ts` instead of this function. We need to be consistent. Otherwise, the RAM usage printed by `ls` will not be the same as values printed everywhere else.

catloversg:
You also need to update text at:
https://github.com/bitburner-official/bitburner-src/blob/9824166dbf86b8d572fb2ca5aa4bce46ca5877fe/src/Terminal/HelpText.ts#L339

Optional: Explain the meaning of 2 new columns here. The first column is self-explanatory, but the second one is not.

src/Terminal/commands/ls.tsx
HansLuft778: I will use it for the script's RAM usage, but should i also for the filesize? Looks like its only compatible with GB and upwards

catloversg: Regarding file size, let's wait for @d0sboots's opinion.

d0sboots: We do not have a formatter that is appropriate for built-in filesizes yet. However, since you've already mostly made one, we should either scrap it (seems sad), or go all the way and finish it for general use.

To do the latter:
* Move this to ui/formatNumber as `formatFilesize`
* You're going to want to copy `formatRam` for most of the implementation. In particular, it should use `getFormatter` and you're going to need to add a similar thing to `ramSuffixList` to work with the settings switch between KB and KiB. Yup, it's a pain, but that's what you get when you sign up for human-readable units. XD

If you don't want to do all that, I'd be totally fine with just using raw numbers for filesize.

d0sboots:
> should be fixed now. Regarding rebase: Usually I also prefer to merge, but I noticed the CONTRIBUTING.md mentions rebasing against dev so i chose to rebase instead of merge, sorry :D

That was a list of stuff to do before you (initially) submit a PR. However, I can see how it would be ambiguous. We should probably add a section about what to do and what to expect *during* the PR process to CONTRIBUTING.

src/Terminal/HelpText.ts
```diff
@@ -336,14 +336,16 @@ export const HelpTexts: Record<string, string[]> = {
...
-    "Usage: ls [dir] [-l] [--grep pattern]",
+    "Usage: ls [dir] [-l] [-h] [-g, --grep pattern]",
     " ",
     "The ls command, with no arguments, prints all files and directories on the current server's directory to the Terminal screen. ",
     "The files will be displayed in alphabetical order. ",
     " ",
     "The 'dir' optional parameter can be used to display files/directories in another directory.",
     " ",
-    "The '-l' optional parameter allows you to force each item onto a single line.",
+    "The '-l' optional parameter allows you to force each item onto a single line, displaying two new columns for the script's RAM usage and filesize.",
```
catloversg:
```suggestion
    "The '-l' optional parameter allows you to force each item onto a single line, displaying two columns for the script's RAM usage and filesize.",
```
Help text/comments should be about the current state, not the history.

catloversg (commented):
I have not checked `formatNumber.ts`. I'll leave that to d0sboots.

test/jest/ui/formatNumber.test.ts
```diff
@@ -210,3 +213,35 @@ describe("Ram formatting", () => {
     expect(formatRam(1073741824)).toEqual("1.00EiB");
   });
 });
+describe("Filesize formatting", () => {
+  test("With default GB mode", () => {
+    // Initial settings
```
d0sboots, line 224: Can you add NaN and both infs to this and the ram test? Otherwise, everything looks great!

PR #2140, opened 2025-05-19
MISC: Add achievements for BN14

catloversg:
This PR adds achievements discussed at https://discord.com/channels/415207508303544321/459097632896188436/1338131626286121051.

I also improve the type checking of achievement-related code.

Capture


ficocelliguy:
@catloversg can the icon for the win streak one be "10 Dan" (a mythical better-than-the-best rank in Go) and the name be "Ten Steps Ahead" ?

These look really good!

catloversg:
Update:

Capture

tools/generate-achievement-types.js
```diff
@@ -0,0 +1,28 @@
+const { readFileSync, writeFileSync } = require("fs");
+const { resolve } = require("path");
+
+const rawData = readFileSync(resolve(__dirname, "../src/Achievements/AchievementData.json"), "utf-8");
```
d0sboots: I think you can avoid needing to write this generator script. Instead, I think you can make use of import attributes: <https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import/with>

Note that that feature is not widely available, *but* Webpack will polyfill it for us, essentially embedding the result, so there should be no runtime issues. The only question is if there are *build-time* issues that prevent it from working as the typed solution you want here.

d0sboots:
The new achievements look great! I would prefer to avoid needing a new generator script, but I'll take it if it's really needed. My #1 preference would be to find a way to get all process to operate off a single data source, whether that be a simple .ts file that can *also* be parsed by some of our other tools, or a .json that we can import.

catloversg:
I don't get how the new "import attributes" feature can help us here. Currently, we import `src\Achievements\AchievementData.json` and that json file is the single source of truth for achievement data. After importing it, we augment (no Bitburner-pun intended :smiley:) it with other data (e.g., icon, conditions). Nonetheless, I tried to remove the generator script as you wanted. There is still a manual step when we add a new SFx.1 achievement, but I think that's fine.

I also added an assertion function to assert the imported data. The old typecasting `(<AchievementDataJson>(<unknown>data)).achievements` without a runtime check looks too dangerous to me.

d0sboots:
I will fully admit that my comment was based on a vague gut-feel and didn't have a fully thought-out design behind it. However, you found your way to the design I had vaguely felt, bravo! Having an extra manual step when adding SF achievements is much better than a new step for *all* achievements.

PR #2143, opened 2025-05-19
MISC: Remove deprecated tail-related APIs

catloversg:
It's just as the title said.

I did not plan to remove these APIs because they have just been deprecated recently. However, since we already made some breaking changes without a grace period (e.g., #1635), I decided to do that now instead of waiting until v4.

I'm neutral on this kind of sudden-changes-without-notice (the removal of fuzzy matching is a special case, though), so I'm okay with *not* doing this. If the maintainer wants, I'll skip this change. However, in that case, we should reconsider and make a list of changes that should be waited until v4.

Before:
```js
/** @param {NS} ns */
  ns.tail();
  ns.moveTail(0, 0);
  ns.resizeTail(500, 500);
  ns.closeTail();
  ns.setTitle("test");
}
```

After:
```js
/** @param {NS} ns */
  ns.ui.openTail();
  ns.ui.moveTail(0, 0);
  ns.ui.resizeTail(500, 500);
  ns.ui.closeTail();
  ns.ui.setTailTitle("test");
}
```

APIBreakInfo-3.0.0.txt
```
API BREAK INFO FOR 3.0.0

ns.tail() was removed.
It has been automatically replaced with "ns.ui.openTail()".

ns.moveTail() was removed.
It has been automatically replaced with "ns.ui.moveTail()".

ns.resizeTail() was removed.
It has been automatically replaced with "ns.ui.resizeTail()".

ns.closeTail() was removed.
It has been automatically replaced with "ns.ui.closeTail()".

ns.setTitle() was removed.
It has been automatically replaced with "ns.ui.setTailTitle()".



ns.tail
ns.moveTail
ns.resizeTail
ns.closeTail
ns.setTitle

test.js: (Line numbers: 3, 4, 5, 6, 7)
```


d0sboots:
It's a short migration timeframe, but we can auto-migrate these so I'm much less worried about them.

PR #2146, opened 2025-05-21
CORPORATION: Rename setAutoJobAssignment API to setJobAssignment

catloversg:
In pre-v2.2.0, we had 2 APIs for assigning jobs: `assignJob` and `setAutoJobAssignment`. `assignJob` assigns the job to a single employee, while `setAutoJobAssignment` automatically chooses employees and assigns the job to chosen ones. Due to employees no longer being objects in v2.2.0, we removed `assignJob`, so `setAutoJobAssignment` became the only API assigning jobs. The "auto" word in its name confuses some new players. They usually think that there are some kinds of "auto magic" when using that API, but there is nothing "auto" in its behavior after we made the change in v2.2.0.

This PR renames this API to `setJobAssignment` as being discussed at https://discord.com/channels/415207508303544321/459097632896188436/1360652034507804815 and https://discord.com/channels/415207508303544321/459097632896188436/1360680056019484944.

```
API BREAK INFO FOR 3.0.0

ns.corporation.setAutoJobAssignment() was removed.
It has been automatically replaced with "ns.corporation.setJobAssignment()".



ns.corporation.setAutoJobAssignment

test.js: (Line number: 3)
```

Capture


src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -8906,11 +8906,16 @@ export interface OfficeAPI {
    *
    * @param divisionName - Name of the division
    * @param city - Name of the city
-   * @param job - Name of the job
+   * @param job - Name of the job. Do NOT use "Unassigned"; otherwise, this API does not do anything and just return false.
```
d0sboots: The "otherwise" was kinda ambiguous.
```suggestion
   * @param job - Name of the job. Passing "Unassigned" will cause this API to not do anything and just return false.
```

src/utils/APIBreaks/3.0.0.ts
```diff
@@ -94,4 +94,19 @@ export const breakInfos300: APIBreakInfo[] = [
       'It has been automatically replaced with "ns.ui.setTailTitle()".\n\n',
     showPopUp: false,
   },
+  {
+    brokenAPIs: [
+      {
+        name: "ns.corporation.setAutoJobAssignment",
+        migration: {
+          searchValue: "ns.corporation.setAutoJobAssignment",
+          replaceValue: "ns.corporation.setJobAssignment",
```
d0sboots: Do you think we can loosen this to just the bare name? That way, it will catch the cases where people assign namespaces to helper variables. This depends on there not being any other uses of this word, but I think that's the case here.
```suggestion
          searchValue: "setAutoJobAssignment",
          replaceValue: "setJobAssignment",
```

catloversg: That's a good idea.

Issue #2148, opened 2025-05-23
Breaking changes in v3


catloversg:
There are many breaking changes in v3.0.0. This issue documents all these changes. I'll update this issue when we merge new breaking changes.

@d0sboots Please pin this issue.

## Manual Breaking Changes
The changes in this section will unfortunately require manual effort by you, if you are affected by them.

### Remove API server (#2084)
You cannot use the deprecated VSCode extension to connect to the Steam app via API server now. You have to use Remote API to connect your external editors to Bitburner. For more information, please check the pinned message in #external-editors on Discord (https://discord.gg/TFc3hKD).

### Remove fuzzy matching when checking params (#2091)
With some APIs, when you pass values to their params, you can pass a value that is not an exact match. For example, with `ns.singularity.commitCrime`, you can pass `Rob Store`, `rob store`, `RobStore`, `robstore`, `robStore`, etc. This is called "fuzzy matching". Now, you must pass an exact value (i.e., `Rob Store`). This change affects:
- Bladeburner action and type: `BladeburnerActionType`, `BladeburnerGeneralActionName`, `BladeburnerContractName`, `BladeburnerOperationName`, `BladeburnerBlackOpName`, `SpecialBladeburnerActionTypeForSleeve`, `BladeburnerActionTypeForSleeve`.
- Crime: `CrimeType`
- Faction work: `FactionWorkType`
- University class: `UniversityClassType`
- Gym stat: `GymType`
- Job field: `JobField`
- Stock position: `PositionType`
- Stock order: `OrderType`

You can access these values via `ns.enums` and Bladeburner APIs.

### Remove long-deprecated support for running NS1 scripts (#2083)
You cannot run `.script` files now. Please convert them to `.js` (or `.ts` for TypeScript).
Instructions: https://github.com/bitburner-official/bitburner-src/blob/dev/src/Documentation/doc/en/migrations/ns2.md

### Remove APIs that were deprecated a long time ago (#2088)
- `ns.nFormat` was removed. Use `ns.format.number`, `ns.format.ram`, `ns.format.percent`, or JS built-in objects/functions (e.g., `Intl.NumberFormat`, `Intl.PluralRules`, `Intl.Locale`) instead.

### Enforce stricter param check on ns.getBitNodeMultipliers and ns.hacknet.spendHashes (#2085)
- `ns.getBitNodeMultipliers`: `n` and `lvl` must be positive integers.
[...]

PR #2149, opened 2025-05-23
MISC: Add removal of fuzzy matching to list of breaking changes

catloversg:
It's just as the title said.

I also added text mentioning #2148 to the popup of breaking changes.


src/utils/APIBreaks/3.0.0.ts
```diff
@@ -1,112 +1,132 @@
...
+      info:
+        "ns.corporation.setAutoJobAssignment() was removed.\n" +
+        'It has been automatically replaced with "ns.corporation.setJobAssignment()".\n\n',
+      showPopUp: false,
+    },
+    {
+      brokenAPIs: [],
+      info:
+        `With some APIs, when you pass values to their params, you can pass a value that is not an exact match. ` +
+        `For example, with "ns.singularity.commitCrime", you can pass "Rob Store", "rob store", "RobStore", "robstore", "robStore", etc. ` +
```
d0sboots:
```suggestion
        `With some APIs, when you passed values to their params, you could pass a value that was not an exact match. ` +
        `For example, with "ns.singularity.commitCrime", you could pass "Rob Store", "rob store", "RobStore", "robstore", "robStore", etc. ` +
```

src/utils/APIBreaks/3.0.0.ts
```diff
@@ -1,112 +1,132 @@
...
+        `This is called "fuzzy matching". Now, you must pass an exact value (i.e., Rob Store). This change affects:\n` +
+        `- Bladeburner action and type: BladeburnerActionType, BladeburnerGeneralActionName, BladeburnerContractName, BladeburnerOperationName, BladeburnerBlackOpName, SpecialBladeburnerActionTypeForSleeve, BladeburnerActionTypeForSleeve.\n` +
```
d0sboots: Looking through this list makes me realize something. We defined public Enum *types* for all of these various things as part of your big refactor. But do we have enum *objects* for them, in particular for JS players? Many of these are defined in `NSEnums`, but certainly not all of them. Bladeburner is obviously a mess, but eves simple things like `OrderType` I don't think we have programmatic API access to.

I think we should probably just add these wholesale to NSEnums? (As a separate PR.) Unless they're duplicating another function and there's a good reason not to for specific cases. And then add a note here about how these reference Enum names and you can find them here (link to docs). (It doesn't have to be an actual link, a URL is fine.)

This PR doesn't have to wait on that, but it would be good if there was *some* easy way for players to find the details of what are the allowed values for these things in the docs, and it's certainly not easily possible (for many of them) right now.

src/utils/APIBreaks/APIBreak.ts
```diff
@@ -33,6 +38,11 @@ export interface APIBreakInfo {
   info: string;
   /** If broken APIs can be safely migrated, we can skip displaying the notification popup */
   showPopUp: boolean;
+  /**
+   * With some breaking changes, we cannot detect the affected code reliably via "brokenAPIs". In this case, we always
+   * show a popup that notifies the player about this change.
+   */
+  forceShowPopUp?: boolean;
```
d0sboots: I don't understand the difference between this and `showPopup`, or why they need to be distinguished.

src/utils/APIBreaks/APIBreak.ts
```diff
@@ -68,24 +78,24 @@ function detectImpactAndMigrate(brokenFunctions: APIBreakInfo["brokenAPIs"]): Im
...
   const details: {
     text: string;
     showPopUp: boolean;
   }[] = [];
   let numberOfPopUps = 0;
-  for (const breakInfo of breakInfos) {
+  for (const breakInfo of apiBreakingChanges) {
     const impactMap = detectImpactAndMigrate(breakInfo.brokenAPIs);
-    if (!impactMap) {
+    if (impactMap.size === 0 && !breakInfo.forceShowPopUp) {
```
d0sboots: Can't you just check against `breakInfo.brokenAPIs.length !== 0`? I.e. if you specified some brokenAPIs, you want them to match, but if there aren't any this is an unconditional message. (And potentially not a popup, even.)

catloversg:
@d0sboots I tried to explain the reason in the latest commit.

`showPopUp` vs `doNotSkip`:
- `showPopUp`: If this change is not skipped, do we show a pop up?
- `doNotSkip`: If this change does not affect the code, should we skip it?

The if statement: 2 conditions:
- Affect code (i.e., `impactMap.size > 0`)?
  - Yes: Always process.
  - No: Check `breakInfo.doNotSkip`. Skip if falsy.
- `breakInfo.doNotSkip`?
  - True: Always process.
  - Falsy: Check `impactMap.size`. Skip if no impact.

src/utils/APIBreaks/3.0.0.ts
catloversg: We can add/expose these enums to `ns.enums`.
- `BladeburnerActionType`
- `SpecialBladeburnerActionTypeForSleeve`
- `PositionType`
- `OrderType`

Other enums/arrays:
- `BladeburnerGeneralActionName`: Can access via `ns.bladeburner.getGeneralActionNames()`.
- `BladeburnerContractName`: Can access via `ns.bladeburner.getContractNames()`.
- `BladeburnerOperationName`: Can access via `ns.bladeburner.getOperationNames()`.
- `BladeburnerBlackOpName`: Can access via `ns.bladeburner.getBlackOpNames()`.
- Non-BladeBurner: Already available in `ns.enums`.

After adding/exposing those enums (in a different PR), I'll update this breaking change's entry.

d0sboots:
OK, I think the name is better now. Just waiting on the doc update.

PR #2152, opened 2025-05-23
CORPORATION: Rename division.type to division.industry

catloversg:
I remade the change in #2079.

How to test:
- Load [tobacco.gz](https://github.com/user-attachments/files/20416071/tobacco.gz).
- Check the division. It must be a tobacco division.

Capture


src/Corporation/Division.ts
```diff
@@ -1137,9 +1138,15 @@ export class Division {
...
   }
 
-  /** Initializes a Industry object from a JSON save state. */
+  /** Initializes a Division object from a JSON save state. */
   static fromJSON(value: IReviverValue): Division {
-    return Generic_fromJSON(Division, value.data, Division.includedKeys);
+    const division = Generic_fromJSON(Division, value.data, Division.includedKeys);
+    // division.type was renamed to division.industry in v3.0.0.
+    assertObject(value.data);
+    if ("type" in value.data) {
```
d0sboots, line 1146:
```suggestion
    if (division.industry == null) {
```

catloversg: `division.industry` is not null after `const division = Generic_fromJSON(Division, value.data, Division.includedKeys);`, regardless of the raw data in `value.data`. When we replace the default values of `Division` (the class)'s properties with raw data from `value.data`, we only use non-undefined properties. In this case, `division.industry` is initialized with "Agriculture". When loading data, we copy the "industry" property from the raw data.
- v3+: "industry" is non-undefined. It's ok.
- pre-v3: "industry" is undefined, so `division.industry` is still "Agriculture".

Do you mean `if (value.data.industry == null)`?

PR #2153, opened 2025-05-24
MISC: Make nuke and port cracking APIs return false instead of throwing error

### LINKED ISSUE #2104: ns.nuke() not working as documented
As per [bitburner.ns.nuke.md](https://github.com/bitburner-official/bitburner-src/blob/dev/markdown/bitburner.ns.nuke.md) the functions signature is `nuke(host: string): boolean;`

> True if the player runs the program successfully, and false otherwise.

However, factually it throws an exception when it did not run successfully.

Image****

catloversg:
I decided to take over #1023.

This is a breaking change, but I'm not sure if we need to add an entry to `src\utils\APIBreaks\3.0.0.ts`.

Closes #1023.
Closes #2104.

Test code:
```js
/** @param {NS} ns */
  ns.ui.openTail();
  ns.clearLog();
  ns.rm("NUKE.exe");
  ns.nuke("n00dles");
  ns.brutessh("n00dles");
  ns.ftpcrack("n00dles");
  ns.relaysmtp("n00dles");
  ns.httpworm("n00dles");
  ns.sqlinject("n00dles");
  ns.singularity.createProgram("nuke.exe");
  await ns.sleep(1000);
  ns.nuke("ecorp");
}
```

```
nuke: You do not have the NUKE.exe virus!
brutessh: You do not have the BruteSSH.exe program!
ftpcrack: You do not have the FTPCrack.exe program!
relaysmtp: You do not have the relaySMTP.exe program!
httpworm: You do not have the HTTPWorm.exe program!
sqlinject: You do not have the SQLInject.exe program!
singularity.createProgram: Began creating program: 'nuke.exe'
sleep: Sleeping for 1.000 seconds.
nuke: Not enough ports opened to use NUKE.exe virus.
```


d0sboots:
Thanks for grabbing this one! 

This one is potentially breaking, and I know of at least some people that it will break, but the vast majority will be fine. 

I think it does need an entry. This one we can highlight code for, at least. The description should describe how probably no action is needed, but if you were catching errors *and* branching on the result of success/failure, you'll need to use the return value instead.

Edit: Once this goes in, we can add it to the tracking bug. I've rank it as the lowest priority of the manually breaking changes.

catloversg:
Is this ok?
```
"ns.nuke", "ns.brutessh", "ns.ftpcrack", "ns.relaysmtp", "ns.httpworm", and "ns.sqlinject" now do not throw an error
when you don't have the required .exe file or enough openned ports.
This should not be a problem with most scripts. However, if you were catching errors and branching on the result of
success/failure, you'll need to use the return value instead.
```

I'll add it after we merge #2149. I want to enable the new `doNotSkip` flag for this change.

d0sboots:
That looks good, yes. Why do you think doNotSkip is needed for this? I don't think it's particularly high-priority, just standard priority. (Show the first time.)

catloversg:
Using `doNotSkip` is a way to bypass the limitation of how we use `brokenAPIs`. If we cannot detect any changes by checking `brokenAPIs`, we won't do anything (i.e., writing log file, showing popup). With this change, I plan to leave `brokenAPIs` empty because most scripts do not have a problem with this change. I think we don't have to scan scripts for these APIs.

If you want, I can put these APIs to `brokenAPIs` and *not* use `doNotSkip`.

d0sboots:
I was thinking we would scan for the APIs, since these cases *do* have something to scan. However, we don't want to be super in-your-face about it: Probably exactly one popup message about the change, and we can leave detailed logging of the position of occurrences to the logfile. (The logfile is why I think it makes sense to still scan these.)

catloversg:
No mention of "affected lines":



Show number of affected lines:



```
API BREAK INFO FOR 3.0.0

With some APIs, when you passed values to their params, you could pass a value that was not an exact match. For example, with ns.singularity.commitCrime, you could pass "Rob Store", "rob store", "RobStore", "robstore", "robStore", etc. This is called "fuzzy matching". Now, you must pass an exact value (i.e., Rob Store). This change affects:
- Bladeburner action and type: BladeburnerActionType, BladeburnerGeneralActionName, BladeburnerContractName, BladeburnerOperationName, BladeburnerBlackOpName, SpecialBladeburnerActionTypeForSleeve, BladeburnerActionTypeForSleeve.
- Crime: CrimeType
- Faction work: FactionWorkType
- University class: UniversityClassType
- Gym stat: GymType
- Job field: JobField
- Stock position: PositionType
- Stock order: OrderType
You can access these values via ns.enums and Bladeburner APIs.



ns.nuke, ns.brutessh, ns.ftpcrack, ns.relaysmtp, ns.httpworm, and ns.sqlinject now do not throw an error when you do not have the required .exe file or enough opened ports.
This should not be a problem with most scripts. However, if you were catching errors and branching on the result of success/failure, you will need to use the return value instead.

ns.nuke
ns.brutessh
ns.ftpcrack
ns.relaysmtp
ns.httpworm
ns.sqlinject

test.js: (Line numbers: 21, 22)
```

PR #2155, opened 2025-05-24
MISC: Add versionNumber to ns.ui.getGameInfo()

catloversg:
This is particularly useful in the transition time between versions with breaking changes. For example, `ns.corporation.setAutoJobAssignment` is removed in v3.0.0, but the replacement (i.e., `ns.corporation.setJobAssignment()`) is not available in v2.8.1. If the player wants to write scripts that work in both versions, they will have to parse `ns.ui.getGameInfo().version`, which is a string. With the built-in version number, they don't have to parse the version string. This addition is not useful in this major version bump, but it will be useful in the future.

Test code:
```js
/** @param {NS} ns */
  ns.ui.openTail();
  ns.clearLog();
  ns.print(ns.ui.getGameInfo())
}
```

```
{"version":"3.0.0dev","versionNumber":43,"commit":"eaf401451","platform":"Browser"}
```


src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -10096,6 +10096,7 @@ interface IStyleSettings {
  */
 interface GameInfo {
   version: string;
```
d0sboots, line 10103: Now that there's two "versions" in here, let's add some docs.
```suggestion
  /** Version as shown in release notes and in the UI. E.g.: "2.8.1dev" */
  version: string;
```

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -10096,6 +10096,7 @@ interface IStyleSettings {
...
+  versionNumber: number;
```
d0sboots, line 10105:
```suggestion
  /** Internal version number that increments during releases. E.g.: 43 */
  versionNumber: number;
```
Please correct the examples if they are not valid current/past examples.

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -10096,6 +10096,7 @@ interface IStyleSettings {
...
   commit: string;
```
d0sboots, line 10107:
```suggestion
  /** Git commit hash that the release was built from. E.g.: (I'm on phone, use last release) */
  commit: string;
```

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -10096,6 +10096,7 @@ interface IStyleSettings {
...
   platform: string;
```
d0sboots: An example here of the specific format for Steam would also be nice.

catloversg:
I added an extra note for `version`.

Issue #2147, opened 2025-05-21
Improve errors and docs for `mv` command [Good first issue]


ficocelliguy:
Currently, the `mv` command fails wit an unhelpful error if the script that was (attempted to be) renamed is currently running.

The error should be more descriptive, indicating that the script cannot be renamed/moved because it is currently running. 

The documentation on mv should also mention this restriction.

context:
https://discord.com/channels/415207508303544321/415207508303544323/1374815018121691136

d0sboots:
We could also simply remove this restriction; Linux/unix doesn't have it, and there's no technical reason we need to have it either.

UncleCeiling:
Following a [discussion on Discord](https://discord.com/channels/415207508303544321/459097632896188436/1376219813638180914) I'll make the changes to the error message for both the command and API call.

Changing the behaviour of mv/rm would result in some funkiness with the WorkerScript behaviour on load that I'm not capable of handling yet.

```
Error: WorkerScript constructed with invalid script filename: looper.js
    at new WorkerScript (webpack-internal:///./src/Netscript/WorkerScript.ts:88:13)
    at createAndAddWorkerScript (webpack-internal:///./src/NetscriptWorker.ts:140:24)
    at startWorkerScript (webpack-internal:///./src/NetscriptWorker.ts:96:7)
    at loadAllRunningScripts (webpack-internal:///./src/NetscriptWorker.ts:239:7)
    at Object.load (webpack-internal:///./src/engine.tsx:280:79)
```

d0sboots:
Yes, that's a good point. Scripts currently assume they can be restarted on load; technically this could mean a different version of the code if they were changed, but we're usually fine with that. It gets weirder if you remove the code, since that breaks.

gmcew:
@catloversg between the changes in #2160 and #2267 is it ok to close this? From the conversations I think a fundamental change to `mv` and/or `rm` behaviour is not currently on the table and the docs have now been updated.

PR #2160, opened 2025-05-25
COMMAND: Improving error message for `mv` when a script is Running

UncleCeiling:
# Linked issues

partially addresses #2147 

# Documentation

Used `npm run doc` to autogenerate new .md files.

# Bug fix
1. `npm run test` - 38/38 test suites passed
2. `npm run start:dev` - No errors found
3. Wrote a simple `looper.js` script and started it. Then tried to move it using both command line and a simple `mover.js` script.
Error messages
4. Checked the hover documentation/hints in the editor.
Hover Documentation

### DIFF
--- a/src/NetscriptFunctions.ts
+++ b/src/NetscriptFunctions.ts
@@ -152,10 +152,10 @@ export const ns: InternalAPI<NSFull> = {
   hacknet: NetscriptHacknet(),
   sprintf:
     (ctx) =>
-      (_format, ...args) => {
-        const format = helpers.string(ctx, "format", _format);
-        return sprintf(format, ...(args as unknown[]));
-      },
+    (_format, ...args) => {
+      const format = helpers.string(ctx, "format", _format);
+      return sprintf(format, ...(args as unknown[]));
+    },
   vsprintf: (ctx) => (_format, _args) => {
     const format = helpers.string(ctx, "format", _format);
     if (!Array.isArray(_args)) {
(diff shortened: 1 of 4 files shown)


src/NetscriptFunctions.ts
```diff
@@ -1730,7 +1730,11 @@ export const ns: InternalAPI<NSFull> = {
       helpers.log(ctx, () => `Moved ${sourcePath} to ${destinationPath} on ${host}`);
       return;
     }
-    helpers.log(ctx, () => `ERROR: Failed. Was unable to remove file ${sourcePath} from its original location.`);
+    helpers.log(
+      ctx,
+      () =>
+        `ERROR: Failed. Was unable to remove file ${sourcePath} from its original location. If ${sourcePath} is a script, make sure that it is NOT running before trying to use 'mv' on it.`,
+    );
```
catloversg: You can invese the check like this:
```js
    const success = sourceContentFile.deleteFromServer(server);
    if (!success) {
      helpers.log(ctx, () => `ERROR: Failed. Was unable to remove file ${sourcePath} from its original location. <New error message>`);
      return;
    }
    const { overwritten } = server.writeToContentFile(destinationPath, sourceContentFile.content);
    if (overwritten) helpers.log(ctx, () => `WARNING: Overwriting file ${destinationPath} on ${host}`);
    helpers.log(ctx, () => `Moved ${sourcePath} to ${destinationPath} on ${host}`);
  },
```
Personally, I recommend this coding style:
```js
const result1 = checkCondition1();
if (!result1) {
  logError();
  return;
}
const result2 = checkCondition2();
if (!result2) {
  logError();
  return;
}
// Everything is ok.
process();
```

Optional: `deleteFromServer` only returns false if:
- Scripts: `this.server !== server.hostname || server.isRunning(this.filename)`.
- Text files: `!server.textFiles.has(this.filename)`.

With these conditions, it only fails if there are running instances. We can say something like this:
`ERROR: Cannot remove ${sourcePath} from its original location. That script is running.`

Technically, we should refactor `deleteFromServer` to return the reason of failure and inspect that reason before printing the error. However, it's a bit complicated and may not be worth the effort.

src/Terminal/commands/mv.ts
```diff
@@ -27,7 +27,9 @@ export function mv(args: (string | number | boolean)[], server: BaseServer): voi
   if (!sourceContentFile) return Terminal.error(`Source file ${sourcePath} does not exist`);
 
   if (!sourceContentFile.deleteFromServer(server)) {
-    return Terminal.error(`Could not remove source file ${sourcePath} from existing location.`);
+    return Terminal.error(
```
catloversg, line 30: Same comment about the error message.

src/NetscriptFunctions.ts
UncleCeiling: Yeah, that makes sense - I was mostly just trying to reduce my footprint by editing what was already there.

And aye, I'm a big fan of slightly flatter (is that the word?) styles.

src/Terminal/commands/mv.ts
d0sboots: 👍

UncleCeiling: Understood - Where do I go from here - should I make those refactors or leave them for someone more familiar?

catloversg: Inversing the check in `src\NetscriptFunctions.ts` is a simple change. It's fine to do that in this PR.

UncleCeiling: I think that's it? Sorry it took so long - family stuff.

d0sboots:
Looks good, but `prettier` is complaining. `npm run format` and you should be good!

UncleCeiling:
Sorted; thanks again for the patience!

PR #2163, opened 2025-05-28
UI: Search and read NS API docs in editor tab and documentation tab

### LINKED ISSUE #707: Integrate API documentation into the game
Issue https://github.com/bitburner-official/bitburner-src/issues/624 mentions integrating ReadTheDocs documentation into the game.

Now, to be able to have all documentation in the game when playing offline, it'd be awesome to see the API documentation in-game too. Since this is all generated markdown, it might be able to use the same systems.

catloversg:
This PR integrates NS API docs into the game. It solves 2 problems:
- Reading NS API docs requires an Internet connection. There is no offline access.
- Searching the relevant docs of API(s) is too cumbersome with GitHub UI.

### Levenshtein distance vs Dice-Sørensen coefficient
compare

### Dice-Sørensen coefficient implementation
I tested 3 libraries: 
- https://www.npmjs.com/package/dice-coefficient
- https://www.npmjs.com/package/fast-dice-coefficient
- https://www.npmjs.com/package/string-similarity

Test code:
```js
import fs from "fs";
import { diceCoefficient } from "dice-coefficient";
import dice from "fast-dice-coefficient";
import { compareTwoStrings } from "string-similarity";

    if (!Number.isInteger(min)) {
        throw new Error(`Min is not an integer. Min: ${min}.`);
    }
    if (!Number.isInteger(max)) {
        throw new Error(`Max is not an integer. Max: ${max}.`);
    }
    if (min > max) {
    }
}

const filenames = fs.readdirSync("path\\Bitburner\\markdown");

function verify() {
    const randomFilename = filenames[getRandomIntInclusive(0, 500)];
    for (const filename of filenames) {
        const result1 = diceCoefficient(filename, randomFilename);
        const result2 = dice(filename, randomFilename);
        const result3 = compareTwoStrings(filename, randomFilename);
        if (result1 !== result2 || result1 !== result3 || result2 !== result3) {
            throw new Error(`${randomFilename} ${filename}`);
        }
    }
}

function benchmark1(func) {
    let time = 0;
    for (let i = 0; i < 100; ++i) {
        const randomFilename = filenames[getRandomIntInclusive(0, 500)];
        for (const filename of filenames) {
            const now = performance.now();
            func(filename, randomFilename);
            time += performance.now() - now;
        }
    }
    console.log(func.name.padStart(17), time);
}

const file1 = await (
    await fetch(
        "https://raw.githubusercontent.com/bitburner-official/bitburner-src/refs/tags/v2.5.0/src/NetscriptFunctions.ts"
    )
).text();
const file2 = await (
    await fetch(
        "https://raw.githubusercontent.com/bitburner-official/bitburner-src/refs/tags/v2.8.0/src/NetscriptFunctions.ts"
    )
).text();
const file3 = await (
    await fetch(
        "https://raw.githubusercontent.com/bitburner-official/bitburner-src/refs/tags/v2.8.1/src/NetscriptFunctions.ts"
    )
).text();

function benchmark2(func) {
    const now = performance.now();
    const result1 = func(file1, file2);
    const result2 = func(file2, file3);
    const time = performance.now() - now;
    console.log(func.name.padStart(17), time);
    console.log("".padStart(17), result1, result2);
}

for (let i = 0; i < 100; ++i) {
    verify();
}

benchmark1(diceCoefficient);
benchmark1(dice);
benchmark1(compareTwoStrings);

benchmark2(diceCoefficient);
benchmark2(dice);
benchmark2(compareTwoStrings);
```
Sample output benchmark1:
```
  diceCoefficient 931.0340000001024
             dice 921.1904999997842
compareTwoStrings 916.1253000001161
```
Sample output benchmark2:
```
  diceCoefficient 10890.0826
                  0.9687779775122484 0.999414740186447
             dice 30.334399999999732
                  0.9645342074653696 0.999414740186447
compareTwoStrings 30.590600000003178
                  0.9559398360803386 0.999333907652957
```
Comments:
- With short strings (filenames of files in the markdown folder), they generate same scores. The performance is the same.
- With very long strings (different versions of `NetscriptFunctions.ts`):
  - `dice-coefficient` is too slow.
  - `dice-coefficient` and `fast-dice-coefficient` generate same scores with v2.8.0 vs v2.8.1 (not too many changes).
  - All of them generate different scores with v2.5.0 vs v2.8.0 (many changes).

I chose `fast-dice-coefficient`.

### Other changes
https://github.com/catloversg/bitburner-src/blob/eaf40145169d4c1c2abcff157d7d19d230a310f0/src/Documentation/ui/DocumentationRoot.tsx#L17-L20

`history` is a state of `HistoryProvider`. Pushing to it while rendering `DocumentationRoot` triggers a React warning (`Cannot update a component () while rendering a different component`).

CSS changes in `src\ui\MD\components.tsx` are necessary to render tables correctly.

### Demo
tutorial_step1
tutorial_step2

https://github.com/user-attachments/assets/57ced5fc-5595-4a74-8eec-265757e358bb

Closes #707.


src/@types/global.d.ts
```diff
@@ -60,3 +60,7 @@ module "monaco-vim" {
     };
   };
 }
+
+declare module "fast-dice-coefficient" {
```
d0sboots, line 64: Why do we need a new entry here, instead of just a normal import?

src/Documentation/ui/DocumentationAutocomplete.tsx
```diff
@@ -0,0 +1,92 @@
...
+import { nsApiPages } from "../pages";
+
+/**
+ * bitburner.ns.md -> ns
+ */
+const regex = /^bitburner\.|\.md$/g;
+
+const suggestions = nsApiPages.filter((v) => {
+  // index.md in the "markdown" folder is useless.
+  return !["index.md"].includes(v);
```
d0sboots: Why this instead of the simpler `v !== "index.md"`?

src/Documentation/ui/DocumentationAutocomplete.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+export function DocumentationAutocomplete({ sx, onChange }: DocumentationAutocompleteProps) {
+  const [options, setOptions] = useState<string[]>([]);
+  const [searchValue, setSearchValue] = useState("");
+  return (
+    <Autocomplete
+      freeSolo
+      disableClearable
+      /**
+       * onChange of Autocomplete (not this TextField) is only called when the current value has not been changed. This
```
d0sboots:
```suggestion
       * onChange of Autocomplete (not this TextField) is only called when the current value has been changed. This
```

src/ui/MD/a.tsx
```diff
@@ -15,18 +14,12 @@ export const A = (props: React.PropsWithChildren<{ href?: string }>): React.Reac
...
     navigator.navigate(ref, event.ctrlKey);
   };
-  if (ref.startsWith("http")) {
-    let href = ref;
-    // The URL of NS API documentation in index.md always points to the stable branch, so we need to intercept it here
-    // and change it if necessary.
-    if (href === getNsApiDocumentationUrl(false)) {
-      href = getNsApiDocumentationUrl();
-    }
+  if (ref === "https://github.com/bitburner-official/bitburner-src/blob/stable/markdown/bitburner.ns.md") {
```
d0sboots: This used to be a constant. Why isn't it anymore?

tools/bundle-doc/index.js
```diff
@@ -1,32 +1,63 @@
...
-      processDir(path);
+    const filePath = path.join(dir, file);
+    if (fs.lstatSync(filePath).isDirectory()) {
+      processDir(filePath);
       continue;
     }
-    if (path.startsWith(docRoot + "/")) {
-      files.push(path.slice(docRoot.length + 1));
+    if (filePath.startsWith(docRoot)) {
+      // Windows path uses "\", so we need to replace it with "/".
```
d0sboots: Optional: There's just enough duplication here that it feels worth defining a closure to do this, applied to either docFiles or nsDocFiles.

d0sboots (commented):
There's 3 different PRs in here (a docs change, the actual md browsing change, and then some minor stuff with script editor renaming). It's OK for them to go in together, but I was surprised because usually you're careful to keep things separated.

src/@types/global.d.ts
catloversg: "fast-dice-coefficient" does not provide a declaration file. Without it, TS complains:
```
Could not find a declaration file for module 'fast-dice-coefficient'. '<path>/Bitburner/node_modules/fast-dice-coefficient/dice.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/fast-dice-coefficient` if it exists or add a new declaration (.d.ts) file containing `declare module 'fast-dice-coefficient';`ts(7016)
```

src/Documentation/ui/DocumentationAutocomplete.tsx
catloversg: There was fancy logic that builds an array there, but I removed everything and left only `"index.md"`. I changed it to `return v !== "index.md";` in the latest commit.

src/ui/MD/a.tsx
catloversg: `href === getNsApiDocumentationUrl(false)` is the same as `ref === "https://github.com/bitburner-official/bitburner-src/blob/stable/markdown/bitburner.ns.md"`. If it's true, previously, we needed to create a situational URL ("href") and pass it to `<Link />`. Now we only need to pass `defaultNsApiPage` to `navigator.navigate`. If the player wants to open the full URL externally, `navigator.navigate` will call `openDocExternally`. The situational URL will be crafted in that function.

For reference:
```ts
export function getNsApiDocumentationUrl(isDevBranch: boolean = CONSTANTS.isDevBranch): string {
  return `https://github.com/bitburner-official/bitburner-src/blob/${
    isDevBranch ? "dev" : "stable"
  }/markdown/bitburner.ns.md`;
}

export function openDocExternally(path: string) {
  const ver = CONSTANTS.isDevBranch ? "dev" : "stable";
  const url = path.startsWith("nsDoc/")
    ? `https://github.com/bitburner-official/bitburner-src/blob/${ver}/markdown/${path.replace("nsDoc/", "")}`
    : `https://github.com/bitburner-official/bitburner-src/blob/${ver}/src/Documentation/doc/${path}`;
  window.open(url, "_newtab");
}
```

d0sboots: Yes, but I'm not concerned about the runtime cost of creating a url. I'm concerned about having hardcoded URLs scattered about that we might miss in the event that we have to update them. It's much better if things based on this path are centralized in a single place.

catloversg: I added a constant and a comment for that constant.

PR #2165, opened 2025-05-29
API: Add more enums to ns.enums

catloversg:
We need to expose these enums due to #2149. For more information, please check https://github.com/bitburner-official/bitburner-src/pull/2149#discussion_r2105832271.

Test code:
```js
/** @param {NS} ns */
  console.log(ns.enums);
}
```

enums

### DIFF
--- a/markdown/bitburner.nsenums.md
+++ b/markdown/bitburner.nsenums.md
@@ -21,7 +21,11 @@ type NSEnums = {
   CompanyName: CompanyNameEnumType;
   FactionName: FactionNameEnumType;
   CodingContractName: CodingContractNameEnumType;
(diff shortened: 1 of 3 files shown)


gmcew:
As an observation, there's not many enum suggestions left in #1621 .

The `stanek.fragmentDefinitions` and `hashnet.getHashUpgrades` suggestions are not enumed, and whilst we're making breaking changes more broadly I imagine now is a good time to consider if they're applicable too.
The other (larger) suggestion is on converting data out of `ns.corporation.getConstants` into enums. You're probably most in tune with the depths of Corps, so it feels like another good opportunity to make a choice in either direction.

If this is better discussed more broadly I'll throw it in the discord.

PR #2169, opened 2025-05-30
FEATURE: Add "Recent Errors" tab and improved error modal

ficocelliguy:
Currently, if a script spawns a large number of children that have an error, it can make the game unplayable by covering the ui in infinite stacking error modals.

Example user "softlocked" by modal spam:
https://discord.com/channels/415207508303544321/415207508303544323/1372228866621571163

This PR overhauls the UI around displaying errors to improve this experience.

- Added new "Recent Errors" tab on the Active Scripts page
  - Keeps track of the last 50 distinct errors by default, uses the same settings for number of killed script logs to keep
  - Allows the user to see the number of times each error has been thrown, and open up an error window with more details & see logs
  - Shows a red number on the "Active Scripts" link on the left-hand panel for errors not yet seen
  - Prevents error modals while recent error screen is open, simply adds them to the list
  - Has a button to kill all scripts
  - Has a toggle to prevent error modals until next reload

- Prevent error modals from stacking over top of each other when multiple errors are thrown, causing the UI to become unusable
  - instead just adds them to the "Recent Errors" page and adds a red number on the "Active Scripts" link on the left-hand panel
  
- Error modals now have more tools
  - Button to view the killed script's logs
  - Has a toggle to prevent error modals until next reload
  - Has a button to go straight to the error page

New modal UI:
image

New error page UI:
image


d0sboots:
@catloversg I am kindly requesting your assistance especially on the UI stuff, since you've been doing React stuff much more recently.

src/ErrorHandling/RecentErrorsPage.tsx
```diff
@@ -0,0 +1,138 @@
...
+    return i;
+  };
+
+  const formatMessage = (message: string): string => {
+    // Add zero-width space after each slash to allow clean wrapping
+    const cleanedMessage = message.replaceAll("/", "/​");
+    const fifthLineBreak = nthIndexOf(message, "\n", 5);
+    if (fifthLineBreak !== -1) {
+      // If the message has more than 4 line breaks, truncate it to the first 5 lines
+      return cleanedMessage.slice(0, fifthLineBreak + 1) + " ...";
```
d0sboots: Why not just have scrolling content? Or let css handle this?

d0sboots (commented):
Overall, this looks like a good concept! I will have to give it a second look when I am not quite as tired, though.

catloversg:
After testing it, I have many suggestions. I'll try to implement them to see if they are good or at least make some PoC code, then post the review tomorrow.

src/ErrorHandling/ErrorModal.tsx
```diff
@@ -0,0 +1,86 @@
...
+  const [error, setError] = useState<ErrorRecord | null>(ErrorState.ActiveError);
+
+  useEffect(() => {
+    const listener = (newError: ErrorRecord) => {
+      if (newError.force || (Router.page() !== SimplePage.ActiveScripts && !ErrorState.PreventModals)) {
+        setError(newError);
+        rerender();
+      }
+    };
+    ErrorState.ErrorUpdate.subscribe(listener);
```
catloversg:
```suggestion
    return ErrorState.ErrorUpdate.subscribe(listener);
```
Forgetting to use the cleanup function can create memory leaks. It's not a problem here because `ErrorModal` is created in `GameRoot` and it's not unmounted, but it's a problem in `src\ErrorHandling\RecentErrorsPage.tsx`.

src/ErrorHandling/ErrorModal.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+  const viewLogs = (): void => {
+    onClose();
+    const recentScript = recentScripts.find((script) => script.runningScript.pid === error?.pid);
+    if (!recentScript) {
+      console.warn(`No recent script found with pid ${error?.pid}`);
+      return;
+    }
+    LogBoxEvents.emit(recentScript.runningScript);
+  };
```
catloversg:
```suggestion
  const viewLogs = (): void => {
    if (error === null) {
      return;
    }
    const recentScript = recentScripts.find((script) => script.runningScript.pid === error.pid);
    if (!recentScript) {
      SnackbarEvents.emit(`No recent script found with pid ${error.pid}`, ToastVariant.INFO, 2000);
      return;
    }
    onClose();
    LogBoxEvents.emit(recentScript.runningScript);
  };
```
- Emit a notification instead of `console.warn()`: The player will be confused when they press a button, but nothing happens.
- Only close this modal if we can open the log modal.

src/ErrorHandling/ErrorModal.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+  const goToErrorPage = () => {
+    onClose();
+    ErrorState.UnreadErrors ||= 1;
```
catloversg: I don't get this part. Why do you set it to 1 if it's falsy?

Personally, I prefer the verbose way:
```js
if (ErrorState.UnreadErrors === 0) {
  ErrorState.UnreadErrors = 1;
}
```
It's just a stylistic choice, so you can write it in your preferred way.

src/ErrorHandling/ErrorModal.tsx
```diff
@@ -0,0 +1,86 @@
...
+    Router.toPage(SimplePage.ActiveScripts);
+  };
+
+  return (
+    <Modal open={!!error} onClose={onClose}>
+      {error ? (
+        <Typography>
+          <h2>{error.errorType} ERROR</h2>
+          <p style={{ whiteSpace: "pre-wrap" }}>{error.message.replaceAll("/", "/​")}</p>
```
catloversg:
```suggestion
          {/* Add a zero-width space after each slash to allow clean wrapping. */}
          <p style={{ whiteSpace: "pre-wrap" }}>{error.message.replaceAll("/", "/\u200B")}</p>
```
Don't use the zero-width space character like this. Most diff editors cannot show the difference. Even when we open it in a good IDE, it's still unclear what it means without a comment.

src/ErrorHandling/ErrorModal.tsx
```diff
@@ -0,0 +1,86 @@
...
+          <p>Script: {error.scriptName}</p>
```
catloversg:
```suggestion
          <p>
            Script: {error.scriptName}
            <br />
            PID: {error.pid}
          </p>
```
Suggestion: Show PID.

src/ErrorHandling/ErrorModal.tsx
```diff
@@ -0,0 +1,86 @@
...
+              checked={ErrorState.PreventModals}
+              onChange={(newValue) => (ErrorState.PreventModals = newValue)}
+              text="Prevent more error modals"
+              tooltip={<>If this is set, no error modals will be shown until the game is reloaded.</>}
+            />
+          </div>
+          <Box className={classes.inlineFlexBox}>
+            <Button onClick={onClose}>Close</Button>
+            <div>
+              <Button onClick={viewLogs}>View Script Logs</Button>
```
catloversg:
```suggestion
              <Button disabled={error.pid === -1} onClick={viewLogs}>
                View Script Logs
              </Button>
```
Suggestion: Disable it if we cannot show the log modal.

src/ErrorHandling/ErrorState.tsx
```diff
@@ -0,0 +1,86 @@
+import { EventEmitter } from "../utils/EventEmitter";
+import { Settings } from "../Settings/Settings";
+import { GetAllServers } from "../Server/AllServers";
+import { killWorkerScriptByPid } from "../Netscript/killWorkerScript";
+
+export type ErrorState = {
```
catloversg: Remove this type. It's unnecessary. If we need it in the future, please give it a different name. The type should not have the same name as the variable below.

src/ErrorHandling/ErrorState.tsx
```diff
@@ -0,0 +1,86 @@
...
+};
+
+export const ErrorState: ErrorState = {
+  ErrorUpdate: new EventEmitter<[ErrorRecord]>(),
+  ActiveError: null as ErrorRecord | null,
+  Errors: [],
+  ErrorPageOpen: false,
+  UnreadErrors: 0,
+  PreventModals: Settings.SuppressErrorModals,
+};
```
catloversg:
```suggestion
export const ErrorState = {
  ErrorUpdate: new EventEmitter<[ErrorRecord]>(),
  ActiveError: null as ErrorRecord | null,
  Errors: [] as ErrorRecord[],
  ErrorPageOpen: false,
  UnreadErrors: 0,
  PreventModals: Settings.SuppressErrorModals,
};
```
With this change, you don't need to specify the type above.

src/ErrorHandling/ErrorState.tsx
```diff
@@ -0,0 +1,86 @@
...
+      message,
+      pid,
+      occurrences: 1,
+      time: new Date(),
+      unread: !ErrorState.ErrorPageOpen,
+    });
+    ErrorState.Errors = ErrorState.Errors.slice(0, Settings.MaxRecentScriptsCapacity);
+
+    updateActiveError(ErrorState.Errors[ErrorState.Errors.length - 1]);
+  }
```
catloversg:
```suggestion
    ErrorState.Errors.unshift({
      server: hostname,
      errorType,
      scriptName,
      message,
      pid,
      occurrences: 1,
      time: new Date(),
      unread: !ErrorState.ErrorPageOpen,
    });
    while (ErrorState.Errors.length > Settings.MaxRecentScriptsCapacity) {
      ErrorState.Errors.pop();
    }
    updateActiveError(ErrorState.Errors[0]);
  }
```
Bug: Run 100 error scripts (0 -> 99). `ErrorState.Errors` always contains error 0 -> 49. `unshift+pop` is how we implement this feature in `src\Netscript\RecentScripts.ts`.

src/ErrorHandling/ErrorState.tsx
```diff
@@ -0,0 +1,86 @@
...
+};
+
+function updateActiveError(error: ErrorRecord): void {
+  if (!ErrorState.ActiveError) {
+    ErrorState.ActiveError = error;
+    ErrorState.ErrorUpdate.emit(ErrorState.ActiveError);
+  }
+}
```
catloversg: Bug. More about this later.

src/ErrorHandling/ErrorState.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+export const killAllScripts = () => {
+  GetAllServers().forEach((server) => {
+    for (const byPid of server.runningScriptMap.values()) {
+      for (const pid of byPid.keys()) {
+        killWorkerScriptByPid(pid);
+      }
+    }
+  });
+};
```
catloversg: - Move this function to `src\Netscript\killWorkerScript.ts`
- Optional: Use for-of loop instead of `forEach`. d0sboots prefers it, so some code starts using for-of loop now. It's good to be consistent. If you prefer `forEach`, it's still ok.

src/ErrorHandling/RecentErrorsPage.tsx
```diff
@@ -0,0 +1,138 @@
...
+export function RecentErrorsPage(): React.ReactElement {
+  const rerender = useRerender();
+  React.useEffect(() => {
+    const listener = () => {
+      rerender();
+    };
+    ErrorState.ErrorUpdate.subscribe(listener);
+    ErrorState.ErrorPageOpen = true;
+    ErrorState.UnreadErrors = 0;
+  }, [rerender]);
```
catloversg:
```suggestion
  React.useEffect(() => {
    const clearSubscription = ErrorState.ErrorUpdate.subscribe(rerender);
    ErrorState.ErrorPageOpen = true;
    ErrorState.UnreadErrors = 0;
    return () => {
      clearSubscription();
      ErrorState.ErrorPageOpen = false;
    };
  }, [rerender]);
```
There is a leak here.
```
Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.
```
Everytime you enter this view and leave it, you add an item to `ErrorState.ErrorUpdate.subscribers` via `ErrorState.ErrorUpdate.subscribe`, but you don't remove it by calling the cleanup function.

Another bug is `ErrorState.ErrorPageOpen`. If you don't reset it like this, `ErrorState.UnreadErrors` cannot be updated properly. How to reproduce:
- Reload the page to make sure that we start at a clean state.
- Run a script throwing an error. "Active Scripts" shows "1" unread error. Close the error popup.
- Switch to the "Active Scripts" tab, then switch to the editor tab and run the script again.
- "Active Scripts" does *not* show "1" unread error.

src/ErrorHandling/RecentErrorsPage.tsx
```diff
@@ -0,0 +1,138 @@
...
+  const formatMessage = (message: string): string => {
+    // Add zero-width space after each slash to allow clean wrapping
+    const cleanedMessage = message.replaceAll("/", "/​");
+    const fifthLineBreak = nthIndexOf(message, "\n", 5);
+    if (fifthLineBreak !== -1) {
+      // If the message has more than 4 line breaks, truncate it to the first 5 lines
+      return cleanedMessage.slice(0, fifthLineBreak + 1) + " ...";
+    }
+    return cleanedMessage;
+  };
```
catloversg:
```suggestion
  const formatMessage = (message: string): string => {
    /**
     * - Add a zero-width space after each slash to allow clean wrapping.
     * - Replace 2+ newline characters with only 1 newline character to reduce the number of empty lines.
     */
    return message.replaceAll("/", "/\u200B").replaceAll(/\n{2,}/g, "\n");
  };
```
We only need to do this. The remaining part will be done in CSS. I'll submit a full patch later.

src/ErrorHandling/RecentErrorsPage.tsx
```diff
@@ -0,0 +1,138 @@
...
+
+  return (
```
catloversg, line 39: Suggestions:

before

Example:

after

I'll submit a full patch later.

src/ui/ActiveScripts/ActiveScriptsRoot.tsx
```diff
@@ -8,24 +8,28 @@ import Tab from "@mui/material/Tab";
 
 import { ActiveScriptsPage } from "./ActiveScriptsPage";
 import { RecentScriptsPage } from "./RecentScriptsPage";
+import { RecentErrorsPage } from "../../ErrorHandling/RecentErrorsPage";
 import { useRerender } from "../React/hooks";
+import { ErrorState } from "../../ErrorHandling/ErrorState";
 
 export function ActiveScriptsRoot(): React.ReactElement {
```
catloversg: The option switch and the "Kill all scripts" button shoud be moved here. I'll submit a full patch later.

src/utils/ErrorHandler.ts
```diff
@@ -1,21 +1,34 @@
...
 export function handleUnknownError(e: unknown, ws: WorkerScript | null = null, initialText = "") {
   if (e instanceof ScriptDeath) {
     // No dialog for ScriptDeath
     return;
   }
+  const errorDetails = parseUnknownError(e);
   if (ws && typeof e === "string") {
-    const headerText = basicErrorMessage(ws, "", "");
-    if (!e.includes(headerText)) e = basicErrorMessage(ws, e);
+    // Attempt to strip out the error type, if present
```
catloversg:
```suggestion
    /**
     * - Attempt to strip out the error type, if present.
     * - Extract error text by skipping:
     *   - Error type
     *   - Script name and PID
     *
     * Error example:
     * "RUNTIME ERROR\ntest.js@home (PID - 1)\n\ngetServer: Invalid hostname: 'invalid'\n\nStack:\ntest.js:L5@main"
     *
     * - errorType: "RUNTIME"
     * - errorText: "getServer: Invalid hostname: 'invalid'\n\nStack:\ntest.js:L5@main"
     */
```

src/utils/ErrorHandler.ts
```diff
@@ -32,15 +45,13 @@ export function handleUnknownError(e: unknown, ws: WorkerScript | null = null, i
...
-    e = ws ? basicErrorMessage(ws, msg) : `RUNTIME ERROR:\n\n${msg}`;
-  }
-  if (typeof e !== "string") {
+    DisplayError(initialText + msg, "RUNTIME", ws?.scriptRef?.filename, ws?.hostname, ws?.pid);
+  } else if (typeof e !== "string") {
     console.error("Unexpected error:", e);
     const msg = `Unexpected type of error thrown. This error was likely thrown manually within a script.
         Error has been logged to the console.\n\nType of error: ${typeof e}\nValue of error: ${e}`;
-    e = ws ? basicErrorMessage(ws, msg, "UNKNOWN") : msg;
+    DisplayError(msg, "UNKNOWN", ws?.scriptRef?.filename, ws?.hostname ?? "", ws?.pid ?? -1);
```
catloversg:
```suggestion
    DisplayError(msg, "UNKNOWN", ws?.scriptRef?.filename, ws?.hostname, ws?.pid);
```

catloversg (commented):
Test code:
```js
/** @param {NS} ns */
  ns.ui.openTail();
  ns.clearLog();
  throw new Error(Date.now());
}
```
Bug:
- Reload page.
- Run test code. It will show an error modal and a log modal. Close the error modal.
- Switch to the "Active Scripts" tab. Run the test script via the log modal. There is no error modal.
- Switch to other tabs. Run test code. There is still no error modal.
- From now on, the error modal cannot be shown.

The problem is that we don't set/reset `ErrorState.ActiveError` properly. Before talking about fixing it, I want to discuss the expected behavior of this new feature.

In the current state, this error modal only shows *one* current error. When this error is shown via the modal, all incoming new errors are ignored *and then skipped*. Let's say that I have 3 errors.
- Error 1 occurs.
- I'm looking at the error modal to see the stack of error 1.
- Error 2 occurs.
- Error 3 occurs.
- I close the error modal after checking the stack of error 1.

At this point, error 2 and error 3 are not shown. I know that there are notifications of error 2 and 3 (after we apply my patch in the review), but I still don't like the fact that error 2 and 3 are not shown in an error popup. Note that if I close the error modal before error 2 occurs, error 2 will still be shown in the error modal.

The new feature of "do-not-spam-error-popup-until-unplayable" is nice, but showing error popups is an important way to tell the player that something went wrong. Having too many popups (i.e., thousands of them) is bad, but not showing them when the crash happens is also bad.

I suggest that we implement this feature like this: We show *all* error popups when the crashes happen. For example, show error1 -> close -> show error2 -> close -> show error3 -> etc. In the error modal, we show how many the remaining errors are. If there are too many errors, they have 2 buttons:
- Snooze new errors in a limited period.
- Clear the queue of remaining errors. If they clear that queue and close the modal, the modal will not show the next errors (they have been cleared). Note that this queue is not the current queue shown in the "Recent errors" tab.

catloversg:
Full patch is available at https://github.com/ficocelliguy/bitburner-src/pull/4.

catloversg:
Another problem is the option switch. You have 2 types of option switch.
- One in the Options page: This switch only sets `Settings.SuppressErrorModals`. `Settings.SuppressErrorModals` is the initial value of `ErrorState.PreventModals`. However, when the player changes `Settings.SuppressErrorModals`, `ErrorState.PreventModals` is not updated.
- One in other pages: This switch only sets `ErrorState.PreventModals`. Its effect lasts until reloading.

IMO, this is a bit confusing. If the player turns `Settings.SuppressErrorModals` to true, they may be surprised that it does not affect `ErrorState.PreventModals` immediately, and vice versa. I think we should clarify the difference a bit more in the tooltip.

Optional: We can also change the text. For example: "Prevent error modal temporarily" instead of just "Prevent error modal". The word "temporarily" is a bit long, so the button may look weird. You can find another word/phrase.

d0sboots:
> The new feature of "do-not-spam-error-popup-until-unplayable" is nice, but showing error popups is an important way to tell the player that something went wrong. Having too many popups (i.e., thousands of them) is bad, but not showing them when the crash happens is also bad.
> 
> I suggest that we implement this feature like this: We show _all_ error popups when the crashes happen. For example, show error1 -> close -> show error2 -> close -> show error3 -> etc. In the error modal, we show how many the remaining errors are. If there are too many errors, they have 2 buttons:
> 
> * Snooze new errors in a limited period.
> * Clear the queue of remaining errors. If they clear that queue and close the modal, the modal will not show the next errors (they have been cleared). Note that this queue is not the current queue shown in the "Recent errors" tab.

I think I agree (but I could be convinced otherwise). Specifically:
* I worry that a setting to ignore errors is too "easy" to turn on and forget about. As CatLover says, errors are an *important* feedback that something has gone wrong; it is tempting to ignore them, but will ultimately lead to more pain later.
* I think a small icon/counter in the modal of "how many additional errors have arrived" is very good info to help the user make informed decisions, no matter what we decide.
* I think pressing "Escape" should act the same as "regular close dialog." (It may already do this, I'm just sayin'.) You can potentially hold down Escape to zoom through a lot of them, but this will not help you if they are coming in constantly.
* Therefore, I think both a "kill all scripts" and a "snooze errors for 30 seconds" (we can bikeshed the exact time) are good action buttons to have in the modal. IMO "snooze" should *not* discard the existing "queue", that would be a different action. Pressing "snooze" again would unsnooze. (It's good to have a way to cancel an unintended action.)
* Clicking on the icon showing the extra queued errors would pop a dialog explaining "do you want to clear all these pending errors y/n? (Possibly with a toggle for "don't show this again.) It helps if the icon is button-like (this is essentially the same as CatLover's suggestion)

So in total I guess I am suggesting 3 buttons (edit: 4 when you count the regular "close" button), and also that they always be present (for consistency).

ficocelliguy:
I personally find the red icons on the side very attention-drawing, but I see the value of snoozing errors for only 60 seconds or whatever. We could even make it so that it hides errors until they make a change to a script, if we wanted?

I also think that the error page should be a safe haven from the modal flood, and new errors should appear in the list rather than smacking you in the face after your 60 seconds of breathing are up. Perhaps liberal use of the "kill all scripts" button would be in order in that scenario, but it's sometimes nice to be able to monitor things live

d0sboots:
> I also think that the error page should be a safe haven from the modal flood, and new errors should appear in the list rather than smacking you in the face after your 60 seconds of breathing are up. Perhaps liberal use of the "kill all scripts" button would be in order in that scenario, but it's sometimes nice to be able to monitor things live

I could get behind that as a general behavior: That on that screen specifically, the error modals don't pop. If it has similar functionality (kill and snooze buttons) people won't miss the popups, and I think it will be "natural."

ficocelliguy:
Fixes:
* Made sure the error page is detected correctly when open
* Make error suppression last 5 minutes instead of until restart
* Improve error page layout

TODO next:
~~* Make error page its own route to properly detect it vs active scripts or recently killed tab~~
~~* When an error modal is closed, show the next distinct error that was thrown while it was open (e.g. delay modals while a modal is open, don't suppress them)~~
~~* Improve and clarify interaction between suppress errors game setting and the 5 minute breather from the modal~~

ficocelliguy:
Updated with the suggestions! The new ux is in the description.

catloversg:
Is this ready for the next review round? Some old comments don't have any replies, but they are also not marked as "Resolved".

I have not checked the new changes carefully, so I won't comment directly on the code now. However, I have built it and tested it a bit:
- React prints many warnings: `Warning: validateDOMNesting(...): X cannot appear as a descendant of Y.`. You should open the dev tool while testing and fix these warnings. Many `<Typography>` should be `<Typography component="div">`.
- `Recent Errors` page: Scrolling through 50 *big* rows of errors is a hassle for the player. Please check my old review, screenshots, and the patch demo. I strongly recommend:
  - The `Message` column: Remove waste space.
  - The `Script` column: It still has the overflow problem.
  - Use `return message.replaceAll("/", "/\u200B").replaceAll(/\n{2,}/g, "\n");` to remove unnecessary newlines when previewing the error messages.
  - Use smaller font size and line height like in my patch.
  - "Single row" header (check the difference in my old screenshots; for example, "Recently killed" should not be on 2 lines). My patch already set all necessary CSS.
- `ErrorState` (the type) is unnecessary, like I said in the previous review.

Capture

Bug:
- When switching to the "Active Scripts" tab, the content belongs to the "Recent Errors" tab, but the "Active" tab is highlighted.

error1

- When clicking the close button (top right of the modal), the modal is forcefully closed due to a bug in `onClose`. When you pass `onClose` to the modal component like this: `<Modal open={!!error} onClose={onClose}>`, it will be called with a synthetic event as the first param, so in some cases (e.g., clicking the close button), `force` is an event, not an optional boolean like you want to.

I strongly recommend:
- Bring back the switch to enable/disable "Snooze mode" in the "Active Scripts" page. In the current state, if I enable that mode, I cannot disable it even if I want.
- Show the number of remaining errors in the modal.

Consideration (Wait for @d0sboots's opinions):
- The modal is using the queue in `ErrorState` which has the limit of `Settings.MaxRecentScriptsCapacity`. This means if there are more than `MaxRecentScriptsCapacity` errors, the player will miss some. It's good to have a history limit, but I think we should consider having a separate queue (maybe with a very high limit or even no limit) for errors shown by the modal. The error popup is a strong indication that something went wrong, so it should show *all* errors, IMO. If there are too many errors and the player wants to ignore them, they can "snooze" them.

ficocelliguy:
I didn't see the PR you opened against my branch until just now, so I need to go back and do some merging and cherry-picking to apply some of your suggestions @catloversg 

Also re: your last comment, the limit is for **distinct** errors. Unless the player is throwing a custom `new Error(Math.random())` or something, it will simply update the count of the last time that error was thrown and mark it as unread again (queued for being shown, though it won't be the next one shown if there is another distinct error). Things like server name are ignored for the purposes of detecting duplicates.

src/ErrorHandling/RecentErrorsPage.tsx
ficocelliguy: Updated to use css instead.

src/ErrorHandling/RecentErrorsPage.tsx
ficocelliguy: suggestion and patch merged.

ficocelliguy:
I've updated this with catlover's suggestions and patch branch, and updated the screenshots.

src/ErrorHandling/ErrorModal.tsx
ficocelliguy: Originally this was so it would go to the error tab, not just the active scripts page, but it's no longer needed to be truthy.

src/ErrorHandling/ErrorModal.tsx
```diff
@@ -0,0 +1,119 @@
...
+  const { classes } = useStyles();
+  const rerender = useRerender();
+  const [error, setError] = useState<ErrorRecord | null>(ErrorState.ActiveError);
+
+  useEffect(() => {
+    const listener = (newError: ErrorRecord) => {
+      if (newError.force || (Router.page() !== SimplePage.RecentErrors && !errorModalsAreSuppressed())) {
+        setError(newError);
+        rerender();
+      }
```
catloversg:
```suggestion
      if (newError.force || (Router.page() !== SimplePage.RecentErrors && !errorModalsAreSuppressed())) {
        setError(newError);
        rerender();
      } else {
        ErrorState.ActiveError = null;
        setError(null);
      }
```
Bug:
- Run the test script. Close error popups.
- Switch to the "Active Scripts" tab.
- Switch to the "Active" tab, then back to the "Recent Errors" tab. (This step is a workaround for another bug).
- Run the test script again via the log modal. No modal is shown.
- Switch to another tab. Run the test script again via the log modal. No modal is shown.

This is the same bug as in the last review. With the current design, you must be extremely careful with `ErrorState.ActiveError`. If you forget to reset it, the error popup will be mistakenly suppressed.

Edit: Please test this bug again after you fix all problems. Maybe I also forgot something or the fix here is flawed.

src/ErrorHandling/ErrorModal.tsx
```diff
@@ -0,0 +1,119 @@
...
+    ErrorState.ActiveError && (ErrorState.ActiveError.unread = false);
+
+    if (force || errorModalsAreSuppressed()) {
+      ErrorState.ActiveError = null;
+      setError(null);
+    } else {
+      const nextError = ErrorState.Errors.find((e) => e.unread) ?? null;
+      ErrorState.ActiveError = nextError;
+      setError(nextError);
+    }
```
catloversg, line 43:
```suggestion
    }
    ErrorState.UnreadErrors = ErrorState.Errors.filter((e) => e.unread).length;
```
Bug (maybe): Run the test script. When you close each error, the number of errors in the sidebar is not changed.

IMO, this is a bug. That number is an indication of how many errors that the player has *not* been informed about. If they are looking at the error popup, read the bug description, then close the error popup, it means they are informed about the problems.

src/ErrorHandling/ErrorModal.tsx
```diff
@@ -0,0 +1,119 @@
...
+              <Button disabled={error.pid === -1} onClick={viewLogs}>
+                View Script Logs
+              </Button>
+              <Button onClick={goToErrorPage}>Errors Page</Button>
+            </div>
+          </Box>
+        </>
+      ) : (
+        ""
+      )}
```
catloversg:
```suggestion
      {error && (
        <>
          <Typography component="div">
            <h2>{error.errorType} ERROR</h2>
            {/* Add a zero-width space after each slash to allow clean wrapping. */}
            <p style={{ whiteSpace: "pre-wrap" }}>{error.message.replaceAll("/", "/\u200B")}</p>
            <p>
              Script: {error.scriptName}
              <br />
              PID: {error.pid}
            </p>
            {!Settings.SuppressErrorModals && (
              <OptionSwitch
                checked={errorModalsAreSuppressed()}
                onChange={(newValue) => toggleSuppressErrorModals(newValue)}
                text="Suppress error modals for 5 minutes"
                tooltip={
                  <>
                    If this is set, no error modals will be shown for the next five minutes.
                    <br />
                    The "Suppress error modals" toggle in the game options will always suppress modals, and only log
                    errors to the Recent Errors page.
                  </>
                }
              />
            )}
          </Typography>
          <Box className={classes.inlineFlexBox}>
            <Button onClick={() => onClose()}>Close</Button>
            <div>
              <Button disabled={error.pid === -1} onClick={viewLogs}>
                View Script Logs
              </Button>
              <Button onClick={goToErrorPage}>Errors Page</Button>
            </div>
          </Box>
        </>
      )}
```
Instead of:
```jsx
<>{obj ? <div>{obj}</div> : ""}</>
```
Please use:
```jsx
<>{obj && <div>{obj}</div>}</>
```
This is the preferred style in our codebase. We use this "pattern" everywhere else.

src/ErrorHandling/ErrorState.tsx
```diff
@@ -0,0 +1,35 @@
...
+export const ErrorState = {
+  ErrorUpdate: new EventEmitter<[ErrorRecord]>(),
+  ActiveError: null as ErrorRecord | null,
+  Errors: [] as ErrorRecord[],
+  UnreadErrors: 0,
+  PreventModalsUntil: new Date(),
+};
+
+export function errorModalsAreSuppressed(): boolean {
+  return ErrorState.PreventModalsUntil > new Date();
```
catloversg:
```suggestion
  return ErrorState.PreventModalsUntil.getTime() > Date.now();
```
Comparing the date objects like you do is still correct, but I think it's better to compare the underlying timestamp values. Comparing the date objects works as long as it does not use equality operators (==, !=, ===, or !==). This means if somebody else changes this line, they may make a mistake (e.g., use `===` to compare the objects).

src/ErrorHandling/RecentErrorsPage.tsx
```diff
@@ -0,0 +1,120 @@
...
+import { Theme } from "@mui/material/styles";
+
+export function RecentErrorsPage(): React.ReactElement {
+  const rerender = useRerender();
+  React.useEffect(() => {
+    const clearSubscription = ErrorState.ErrorUpdate.subscribe(rerender);
+    ErrorState.UnreadErrors = 0;
+    return () => {
+      clearSubscription();
+    };
```
catloversg:
```suggestion
    return () => {
      clearSubscription();
      ErrorState.UnreadErrors = 0;
    };
```
You have to reset `ErrorState.UnreadErrors` when this component is unmounted.

src/ErrorHandling/RecentErrorsPage.tsx
```diff
@@ -0,0 +1,120 @@
...
+    margin: "4px",
+    color: "white",
+    textOverflow: "ellipsis",
+    maxWidth: "50vw",
+    whiteSpace: "pre-wrap",
+    lineClamp: "6",
+    lineHeight: 1.1,
+    "overflow-x": "auto",
+    maxHeight: "200px",
+  },
```
catloversg:
```suggestion
  errorText: {
    margin: "4px",
    color: "white",
    textOverflow: "ellipsis",
    whiteSpace: "pre-wrap",
    lineClamp: "6",
    lineHeight: 1.1,
    overflowX: "auto",
    maxHeight: "200px",
  },
```
- `maxWidth` creates wasted empty space like I said.
- `"overflow-x"` is wrong. You have to use `overflowX`.

src/ui/ActiveScripts/ActiveScriptsRoot.tsx
```diff
@@ -3,29 +3,78 @@
...
 
-  function handleChange(event: React.SyntheticEvent, tab: "active" | "recent"): void {
+  function handleChange(
+    __event: React.SyntheticEvent,
+    tab: SimplePage.ActiveScripts | SimplePage.RecentlyKilledScripts | SimplePage.RecentErrors,
+  ): void {
     setTab(tab);
+    Router.toPage(tab);
   }
+
```
catloversg: Bug:
- Suppress error modals.
- Run the test script.
- Switch to the "Active Scripts" tab. The "Recent Errors" tab is shown.
- Check `Router.page()`. It's "Active Scripts".

Wrong `Router.page()` messes up other logic (e.g., `Router.page() !== SimplePage.RecentErrors`).

Potential fix (GitHub UI does not allow me to post this code as a suggestion):
```tsx
  const [tab, setTab] = useState<ActiveScriptsTab>(page);
  useRerender(400);

  useEffect(() => {
    if (ErrorState.UnreadErrors > 0) {
      handleChange(null, SimplePage.RecentErrors);
    }
  }, []);

  function handleChange(
    __event: React.SyntheticEvent | null,
    tab: SimplePage.ActiveScripts | SimplePage.RecentlyKilledScripts | SimplePage.RecentErrors,
  ): void {
    setTab(tab);
    Router.toPage(tab);
  }
```

src/ui/ActiveScripts/ActiveScriptsRoot.tsx
```diff
@@ -3,29 +3,78 @@
...
+          <Tab label={"Recently Killed"} value={SimplePage.RecentlyKilledScripts} />
+          <Tab label={errorTabText()} value={SimplePage.RecentErrors} />
+        </Tabs>
+        <OptionSwitch
+          checked={errorModalsAreSuppressed()}
+          onChange={(newValue) => toggleSuppressErrorModals(newValue)}
+          text="Prevent error modals"
+          tooltip={<>If this is set, no error modals will be shown until the game is reloaded.</>}
+          wrapperStyles={{ marginLeft: "20px" }}
+        />
```
catloversg:
```suggestion
        <OptionSwitch
          checked={errorModalsAreSuppressed()}
          onChange={(newValue) => toggleSuppressErrorModals(newValue)}
          text="Suppress error modals (5 minutes)"
          tooltip={
            <>
              If this is set, no error modals will be shown for the next five minutes.
              <br />
              The "Suppress error modals" toggle in the game options will always suppress modals, and only log errors to
              the Recent Errors page.
            </>
          }
          wrapperStyles={{ marginLeft: "20px" }}
        />
```
I reused the text in `src\ErrorHandling\ErrorModal.tsx`. Note that I use `"Suppress error modals (5 minutes)"` instead of `"Suppress error modals for 5 minutes"` to reduce the length of the switch a bit.

src/GameOptions/ui/GameplayPage.tsx
```diff
@@ -52,6 +57,17 @@ export const GameplayPage = (): React.ReactElement => {
...
         tooltip={<>If this is set, the stock market will never create any popup.</>}
       />
+      <OptionSwitch
+        checked={Settings.SuppressErrorModals}
+        onChange={toggleSuppressErrorModalsSetting}
+        text="Suppress error modals"
+        tooltip={
+          <>
+            If this is set, script errors will never create any popups. The errors can still be seen on the "Recent
+            Errors" tab in the Active Scripts page
```
catloversg:
```suggestion
            Errors" tab in the Active Scripts page.
```

catloversg (commented):
Test script:
```js
// test.js
/** @param {NS} ns */
  ns.ui.openTail();
  ns.clearLog();
  for (let i = 0; i < 2; ++i) {
    ns.exec("error.js", "home", undefined, i);
    await ns.sleep(0);
  }
}
```
```js
// error.js
/** @param {NS} ns */
  ns.getServer(`invalid ${ns.args[0]}`);
}
```

Potential bug:
- Options -> Interface -> Suppress error modals
- Run the test script. No modal is shown.
- Switch to the "Active Scripts" tab. Disable "Prevent error modals" (the snooze switch).
- Switch to another tab. Run the test script. The error popup is shown.

I'm not sure about the intended behavior here. Based on the tooltip, the permanent setting should have a higher priority (i.e., if it's enabled, the snooze mode does not matter). If the snooze mode has a higher priority, you should clarify it in the tooltip.

src/ErrorHandling/ErrorModal.tsx
```diff
@@ -0,0 +1,119 @@
...
+      SnackbarEvents.emit(`No recent script found with pid ${error.pid}`, ToastVariant.INFO, 2000);
+      return;
+    }
+    onClose();
+    LogBoxEvents.emit(recentScript.runningScript);
+  };
+
+  const goToErrorPage = () => {
+    onClose(true);
+    ErrorState.UnreadErrors ||= 0;
```
catloversg: Do we need this line? It was `ErrorState.UnreadErrors ||= 1;` as a workaround for something (https://github.com/bitburner-official/bitburner-src/pull/2169#discussion_r2141262761). If it's not needed now, we can remove it.

`ErrorState.UnreadErrors ||= 0` is also useless. `ErrorState.UnreadErrors` is a number, so unless it somehow becomes `NaN` or `-0`, `ErrorState.UnreadErrors ||= 0` is essentially just `if (ErrorState.UnreadErrors === 0) ErrorState.UnreadErrors = 0;`.

ficocelliguy: I've removed this, as it's not needed anymore.

src/ui/ActiveScripts/ActiveScriptsRoot.tsx
ficocelliguy: This suggested changed has seemed to fix the issue. The route is now properly RecentErrors when the Active Scripts tab is clicked and there are unread errors.

ficocelliguy:
Updated with suggestions.

* The tooltips on the 5-minute suppress error toggles now more clearly indicates that it will override the global setting if you start manually interacting with them.
* Added suggested changes and verified that unread errors are cleared upon leaving the error page.
* Added suggested changes and verified the route is updated when auto-routing to the recent error tab, when you click "active scripts" and there are unread error notifications
* Added suggested changes and verified the error modal prevention based on route only applies when you are actually on the recent error page, and modals re-appear when navigating away

src/ErrorHandling/ErrorModal.tsx
```diff
@@ -0,0 +1,117 @@
...
+            <p>
+              Script: {error.scriptName}
+              <br />
+              PID: {error.pid}
+            </p>
+            {!Settings.SuppressErrorModals && (
+              <OptionSwitch
+                checked={errorModalsAreSuppressed()}
+                onChange={(newValue) => toggleSuppressErrorModals(newValue)}
+                text="Suppress error modals"
```
catloversg:
```suggestion
                text="Suppress error modals (5 min)"
```

src/ErrorHandling/ErrorModal.tsx
```diff
@@ -0,0 +1,117 @@
...
+                tooltip={
+                  <>
+                    If this is set, no error modals will be shown for the next five minutes, and only log errors to the
+                    Recent Errors page.
+                    <br />
+                    The "Suppress error modals" setting in the game options will set this toggle to always be on by
+                    default.
+                  </>
```
catloversg:
```suggestion
                  <>
                    If this is set, no error modals will be shown for the next five minutes, and only log errors to the
                    Recent Errors page.
                  </>
```
With the behavior that we discussed on Discord (no snooze mode when enabling the permanent setting), I think we should remove the latter part. That part clarifies the old behavior, not the current one.

src/ui/ActiveScripts/ActiveScriptsRoot.tsx
```diff
@@ -2,30 +2,97 @@
...
+            checked={errorModalsAreSuppressed()}
+            onChange={(newValue) => toggleSuppressErrorModals(newValue)}
+            text="Suppress error modals (5 min)"
+            tooltip={
+              <>
+                If this is set, no error modals will be shown for the next five minutes, and only log errors to the
+                Recent Errors page.
+                <br />
+                The "Suppress error modals" setting in the game options will set this toggle to always be on by default.
+              </>
```
catloversg:
```suggestion
              <>
                If this is set, no error modals will be shown for the next five minutes, and only log errors to the
                Recent Errors page.
              </>
```

src/ui/ActiveScripts/ActiveScriptsRoot.tsx
```diff
@@ -2,30 +2,97 @@
...
+              minHeight: "40px",
+            },
+          }}
+        >
+          <Tab label={"Active"} value={SimplePage.ActiveScripts} />
+          <Tab label={"Recently Killed"} value={SimplePage.RecentlyKilledScripts} />
+          <Tab label={errorTabText()} value={SimplePage.RecentErrors} />
+        </Tabs>
+        {Settings.SuppressErrorModals ? (
+          <div style={{ width: "15%" }}></div>
```
catloversg, line 71: This part looks a bit weird, IMO.

Capture

Do we really need that empty space?

ficocelliguy: It's not a tab like the others, its an action button, so I like having the visual and categorical separation.

catloversg (approved):

PR #2171, opened 2025-06-01
BUGFIX: Coding contract UI does not handle error properly when answer format is invalid

catloversg:
With some contracts (i.e., Square Root), `isValid` and `isSolution` throw when calling `convertAnswer` (i.e., `BigInt(ans)`). When it happens, the UI of running the coding contract does not do anything. No errors. No notifications.

How to reproduce: Load the test save file. `run contract-557627.cct`. Enter `abc` and press "Solve". Nothing happens, but the console shows:
```
Uncaught SyntaxError: Cannot convert abc to a BigInt (at SquareRoot.ts:40:29)
    at BigInt (<anonymous>)
    at Object.convertAnswer (SquareRoot.ts:40:29)
    at m.isSolution (Contract.ts:110:55)
    at Object.onAttempt (Contract.ts:125:20)
    at onClick (CodingContractModal.tsx:88:26)
    at Object.Ve (react-dom.production.min.js:52:317)
    at $e (react-dom.production.min.js:52:471)
    at react-dom.production.min.js:53:35
    at Er (react-dom.production.min.js:100:68)
    at Qr (react-dom.production.min.js:101:380)
```

This PR refactors coding contract code and handles errors properly.

I also fixes 2 problems in `CodingContractModal.tsx`:
- Memory leak caused by forgetting to call the cleanup function.
- Unnecessary use of `dangerouslySetInnerHTML`.

Save file: [test-coding-contract-pr.gz](https://github.com/user-attachments/files/20538717/test-coding-contract-pr.gz)

Test code (It's in `test.js`):
```js
/** @param {NS} ns */
  ns.ui.openTail();
  ns.print(ns.codingcontract.attempt("abc", "contract-557627.cct"));
  // ns.print(ns.codingcontract.attempt(undefined, "contract-707563.cct"));
}
```

UI:

Capture


d0sboots:
Someday, we will be done refactoring coding contracts. XD

PR #2174, opened 2025-06-01
MISC: Change message of Singularity error and uncaught promise error

catloversg:
Singularity error:

Before:
before1

After:

after1

Uncaught promise error: The problem is not forgetting to await the promise. It's forgetting to handle the rejected promise.

Before:
before2

After:

after2

### DIFF
--- a/src/Netscript/NetscriptHelpers.tsx
+++ b/src/Netscript/NetscriptHelpers.tsx
@@ -380,10 +380,8 @@ function checkEnvFlags(ctx: NetscriptContext): void {
     log(ctx, () => "Failed to run due to failed concurrency check.");
     const err = errorMessage(
       ctx,
-      `Concurrent calls to Netscript functions are not allowed!
-      Did you forget to await hack(), grow(), or some other
-      promise-returning function?
-      Currently running: ${ws.env.runningFn} tried to run: ${ctx.function}`,
+      "Concurrent calls to Netscript functions are not allowed! Did you forget to await hack(), grow(), or some other " +
+        `promise-returning function?\nCurrently running: ${ws.env.runningFn} tried to run: ${ctx.function}`,
       "CONCURRENCY",
     );
     killWorkerScript(ws);
--- a/src/Netscript/NetscriptHelpers.tsx
+++ b/src/Netscript/NetscriptHelpers.tsx
@@ -381,7 +381,7 @@ function checkEnvFlags(ctx: NetscriptContext): void {
     const err = errorMessage(
       ctx,
       "Concurrent calls to Netscript functions are not allowed! Did you forget to await hack(), grow(), or some other " +
-        `promise-returning function?\nCurrently running: ${ws.env.runningFn} tried to run: ${ctx.function}`,
+        `promise-returning function?\nCurrently running: ${ws.env.runningFn}\nTried to run: ${ctx.function}`,
       "CONCURRENCY",
     );
     killWorkerScript(ws);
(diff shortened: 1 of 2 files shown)


src/UncaughtPromiseHandler.ts
```diff
@@ -6,7 +6,7 @@ export function setupUncaughtPromiseHandler(): void {
     handleUnknownError(
       e.reason,
       null,
-      "UNCAUGHT PROMISE ERROR\nYou forgot to await a promise\nmaybe hack / grow / weaken ?\n\n",
+      "UNCAUGHT PROMISE ERROR\nYou forgot to handle a promise's error (e.g., ns.hack, ns.grow, ns.weaken).\n\n",
```
d0sboots: Although this is more technically correct, I'm not sure that it's better.

The issue is that the most likely way players encounter this *is* because they forgot to `await` a promise: Because they forgot to await, the script immediately moves on to some code that launches another promise, which in `ns` causes a "can't do two things at once" exception, so the original promise will be rejected. And, since it was tossed aside somewhere, eventually it hits here.

It is *possible* for this to be due to legitimately missing a reject case, but our APIs make so little use of promise rejection that it's rather unlikely.

catloversg: I agree with what you said, but the suggestion of "Did you forget to await hack(), grow(), or some other promise-returning function?" is already in the concurrency error message.

Capture

IMO, repeating it here is redundant, and when the problem is anything else (i.e., not forgetting to `await`) that makes the promise rejected, mentioning "forgot to `await`" just hides the real problem, which is forgetting to handle the rejected promise.

catloversg: By the way, I changed the concurrency error message a bit in the latest commit.

Capture

src/Netscript/NetscriptHelpers.tsx
```diff
@@ -381,10 +380,8 @@ function checkEnvFlags(ctx: NetscriptContext): void {
     log(ctx, () => "Failed to run due to failed concurrency check.");
     const err = errorMessage(
       ctx,
-      `Concurrent calls to Netscript functions are not allowed!
-      Did you forget to await hack(), grow(), or some other
-      promise-returning function?
-      Currently running: ${ws.env.runningFn} tried to run: ${ctx.function}`,
+      "Concurrent calls to Netscript functions are not allowed! Did you forget to await hack(), grow(), or some other " +
+        `promise-returning function?\nCurrently running: ${ws.env.runningFn} tried to run: ${ctx.function}`,
```
d0sboots:
```suggestion
        `promise-returning function?\nCurrently running: ${ws.env.runningFn}\nTried to run: ${ctx.function}`,
```

I noticed this in batching the other day, it looks like it says "sleep tried to run hack"

src/UncaughtPromiseHandler.ts
d0sboots: I still disagree, because the message here makes it sound like the issue was that they didn't handle an promise error. But 90% of the time, that is *not* the problem, and indeed that advice is misleading because they should *not* be trying to handle any errors.

The correct action is to figure out what is causing the underlying error. Note that this is *not* the advice I would give in a general JS context, where it is critical that you *do* guard your error paths with a catch of some sort, because regular JS APIs make full use of promise rejection to signal errors. But we do not.

```suggestion
      "UNCAUGHT PROMISE ERROR, caused by:\n\n",
```

PR #2175, opened 2025-06-02
MISC: Update NetscriptDefinitions.d.ts to include typings for hightlightPoint() and clearPointHighlight()

kevinsandow:
Add typings for hightlightPoint() and clearPointHighlight()

---

I'm relatively new to bitburner, and maybe focusing on the wrong thing as a beginner, but the missing types here bothered me - they should be the correct ones, not sure how to verify though.

### DIFF
--- a/src/ScriptEditor/NetscriptDefinitions.d.ts
+++ b/src/ScriptEditor/NetscriptDefinitions.d.ts
@@ -4642,20 +4642,30 @@ export interface GoAnalysis {
 
   /**
    * Adds a colored circle indicator to the specified point. These indicators are removed once a move is played.
-   * @param x the x coordinate to highlight
-   * @param y the y coordinate to highlight
-   * @param color optional: the color to use for the circle. Can be given an RGB string like "#FFF000", or "none" to clear it, or one of
-   *    these color names from the selected theme: "hack" (green), "hp" (red), "money" (yellow), "int" (blue), "cha" (purple)
-   * @param text optional: text to add to the node (replaces the default A.1 or B5 seen on hover). Should be kept short to fit well.
+   *
+   * @remarks
+   * RAM cost: 0 GB
+   *
+   * @param x - The x coordinate to highlight
+   * @param y -  The y coordinate to highlight
+   * @param color -  The color to use for the circle. It can be any value accepted by
+   * {@link https://developer.mozilla.org/en-US/docs/Web/CSS/outline-color | outline-color CSS property} (e.g., an RGB
+   * string like "#FFF000"). You can also use "none" to clear it or one of 5 special values: "hack", "hp", "money",
+   * "int", "cha". The corresponding colors of 5 special values depend on your theme.
+   * @param text - Text to add to the node (replaces the default A.1 or B5 seen on hover). Should be kept short to fit well.
    */
-  highlightPoint(x, y, color, text): void;
+  highlightPoint(x: number, y: number, color?: string, text?: string): void;
 
   /**
    * Removes the highlight color and text from the specified node.
-   * @param x the x coordinate to remove highlight from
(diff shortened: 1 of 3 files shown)


d0sboots:
Prettier is complaining - do `npm run format`. Also, I would suggest `npm run doc` to update the docs that go along with this change.

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -4646,14 +4646,14 @@ export interface GoAnalysis {
    *    these color names from the selected theme: "hack" (green), "hp" (red), "money" (yellow), "int" (blue), "cha" (purple)
    * @param text optional: text to add to the node (replaces the default A.1 or B5 seen on hover). Should be kept short to fit well.
    */
-  highlightPoint(x, y, color, text): void;
+  highlightPoint(x: number, y: number, color?: `#{string}` | "none" | "hack" | "hp" | "int" | "cha", text?: string): void;
```
catloversg: `string` is the only sensible type for `color`. This param accepts any values of [outline-color CSS property](https://developer.mozilla.org/en-US/docs/Web/CSS/outline-color). The RGB string is only one of them. The remaining 6 values are special cases.

If you define the type like this, it will only accept 6 special values and strings started with "#". Note: You forgot the `$` character in that string literal type.

I rewrote the description of this API to include more information. You can copy/paste it ("Add a suggestion" feature of GitHub UI does not work like I want).
```ts
  /**
   * Adds a colored circle indicator to the specified point. These indicators are removed once a move is played.
   *
   * @remarks
   * RAM cost: 0 GB
   *
   * @param x The x coordinate to highlight
   * @param y The y coordinate to highlight
   * @param color Optional: The color to use for the circle. It can be any value accepted by
   * {@link https://developer.mozilla.org/en-US/docs/Web/CSS/outline-color | outline-color CSS property} (e.g., an RGB
   * string like "#FFF000"). You can also use "none" to clear it or one of 5 special values: "hack", "hp", "money",
   * "int", "cha". The corresponding colors of 5 special values depend on your theme.
   * @param text Optional: Text to add to the node (replaces the default A.1 or B5 seen on hover). Should be kept short to fit well.
   */
  highlightPoint(x: number, y: number, color?: string, text?: string): void;
```

kevinsandow: > `string` is the only sensible type for `color`.
 
I wasn't aware that it was regular CSS property as well. Copying the Color Typing from CSS typings is definitely overkill then.

catloversg (approved):
You should run `npm run doc` and `npm run format` like d0sboots said.
Edit: Steps:
- `npm run format` to reformat the code.
- `npm run doc` to generate docs.
- `npm run format` again to reformat generated docs if needed.

I will improve our pipeline to remove the last step if it's possible.

kevinsandow:
I did run the commands, had to rollback changes to pages.ts though, since underscore were sorted differently

catloversg (approved):

scott-joe:
@d0sboots 
New player here. I found my way here after adding types to this more than once. Realizing it was syncing from the game's source, I thought I'd see if I could fix it myself. Is this held up by the 3.0 release? I see there's a "Breaking Changes in 3.0" pinned issue and there hasn't been a push to ^v2 since April.

d0sboots:
Yes, this change is part of v3. You can play the testing version, which includes this and previews the changes to come, by using a different URL for the game or selecting a beta in the steam options.

scott-joe:
> Yes, this change is part of v3. You can play the testing version, which includes this and previews the changes to come, by using a different URL for the game or selecting a beta in the steam options.

Awesome. Thanks! I just discovered the game and am enjoying (any sort of) programming again for the first time in a while :)

PR #2177, opened 2025-06-02
DEVMENU: Reset hacknet server list properly when setting level of SF9

catloversg:
This PR fixes a bug in the dev menu.

How to reproduce:
- Set SF9 to level 3.
- Bitflume.
- Set SF9 to any levels.
- Add more money and run `ns.hacknet.purchaseNode();`.

```
Error: Error: Trying to add a server with an existing IP
Stack: Error: Error: Trying to add a server with an existing IP
    at AddToAllServers (webpack://bitburner/./src/Server/AllServers.ts?:130:11)
    at PlayerObject.createHacknetServer (webpack://bitburner/./src/PersonObjects/Player/PlayerObjectServerMethods.ts?:64:70)
    at purchaseHacknet (webpack://bitburner/./src/Hacknet/HacknetHelpers.tsx?:92:49)
    at eval (webpack://bitburner/./src/NetscriptFunctions/Hacknet.ts?:54:86)
    at Proxy.wrappedFunction (webpack://bitburner/./src/Netscript/APIWrapper.ts?:67:16)
    at main (home/test.js:14:14)
    at startNetscript2Script (webpack://bitburner/./src/NetscriptWorker.ts?:82:9)
```

When we clear the hashnet server list in `Player.hacknetNodes`, we forget to delete the server data in `AllServers`.

I also updated the error message in `AddToAllServers`.


src/DevMenu/ui/SourceFilesDev.tsx
```diff
@@ -29,6 +30,12 @@ export function SourceFilesDev({ parentRerender }: { parentRerender: () => void
     (sfN: number, sfLvl: number) => () => {
       if (sfN === 9) {
         Player.hacknetNodes = [];
+        for (const server of GetAllServers()) {
```
d0sboots: This feels slightly...off. You won't get the new server bonus that you normally do on entering a BN if you grant yourself sf 9. Also, if you're just changing how many levels of 9 you have, it'll trash all your current servers (and I don't see why this would be expected.)

catloversg: I don't know the original purpose of this `if-then-clear` block, but I guess it's there just to clear *whatever* is in `Player.hacknetNodes` (hacknet node or hacknet servers). Let's say we have hacknet servers in `Player.hacknetNodes` (SF9.3), then we set SF9 to SF9.0. Without resetting `Player.hacknetNodes`, the engine/UI may crash because it contains hacknet server data when SF9 level is 0.

In short, it's there as a blanket "reset everything to be safe". Do you want me to add more conditions to reset things when it's needed, not every time SF9 is changed?

d0sboots: Whoops, lost track of this one.

Is it enough to clear only if you set to 9.0? That seems both necessary and (hopefully) sufficient.

PR #2180, opened 2025-06-02
API: Set default values for some properties of Server object returned from ns.getServer

catloversg:
A few months ago, we had a long discussion about server stuff (`Server` in NS API, `BaseServer`, `Server` (internal class), `HacknetServer`, etc.). We discussed too many things, and the discussion got nowhere. G4mingJon4s made a list of suggestions based on what was discussed. The detail is at https://discord.com/channels/415207508303544321/459097632896188436/1356540719933820939. Among those suggestions, I only "agree" with [1] and [2]. Note that I don't advocate these changes. "Agree" only means that I don't veto them.

This PR implements [1]. [2] is still debatable on the implementation detail, so I won't implement it until we settle on a "spec".

With this change, I decided to *not* make the change of "Throw an error when the server is invalid" as planned in <https://discord.com/channels/415207508303544321/1355531035470860338/1355531039250059355>. It also means that #1128 is irrelevant now (i.e., 5 will be the expected default value of `numOpenPortsRequired` for hacknet servers).

I tested this change, and I don't see any problems with old sync APIs and save data, besides the breaking change of `ns.getServer`.

Test code:
```js
/** @param {NS} ns */
function test(ns, hostname) {
  ns.print(ns.getServer(hostname));
  ns.print(ns.hackAnalyzeThreads(hostname, 1));
  ns.print(ns.hackAnalyze(hostname));
  ns.print(ns.hackAnalyzeSecurity(1, hostname));
  ns.print(ns.hackAnalyzeChance(hostname));
  ns.print(ns.growthAnalyze(hostname, 1.1));
  ns.print(ns.growthAnalyzeSecurity(1, hostname));
  ns.print(ns.nuke(hostname));
  ns.print(ns.brutessh(hostname));
  ns.print(ns.ftpcrack(hostname));
  ns.print(ns.relaysmtp(hostname));
  ns.print(ns.httpworm(hostname));
  ns.print(ns.sqlinject(hostname));
  ns.print(ns.getServerMoneyAvailable(hostname));
  ns.print(ns.getServerSecurityLevel(hostname));
  ns.print(ns.getServerBaseSecurityLevel(hostname));
  ns.print(ns.getServerMinSecurityLevel(hostname));
  ns.print(ns.getServerRequiredHackingLevel(hostname));
  ns.print(ns.getServerMaxMoney(hostname));
  ns.print(ns.getServerGrowth(hostname));
  ns.print(ns.getServerNumPortsRequired(hostname));
  ns.print(ns.deleteServer(hostname));
  ns.print(ns.getHackTime(hostname));
  ns.print(ns.getGrowTime(hostname));
  ns.print(ns.getWeakenTime(hostname));
}

/** @param {NS} ns */
  ns.ui.openTail();
  ns.clearLog();
  ns.disableLog("ALL");
  test(ns, "home");
  test(ns, "n00dles");
  test(ns, "hacknet-server-0");
}
```

Capture

### DIFF
--- a/markdown/bitburner.server.backdoorinstalled.md
+++ b/markdown/bitburner.server.backdoorinstalled.md
@@ -9,5 +9,5 @@ Flag indicating whether this server has a backdoor installed by a player
 **Signature:**
 
 ```typescript
-backdoorInstalled?: boolean;
+backdoorInstalled: boolean;
 ```
--- a/markdown/bitburner.server.basedifficulty.md
+++ b/markdown/bitburner.server.basedifficulty.md
@@ -9,5 +9,5 @@ Server's initial server security level at creation.
 **Signature:**
 
 ```typescript
-baseDifficulty?: number;
+baseDifficulty: number;
 ```
--- a/markdown/bitburner.server.hackdifficulty.md
+++ b/markdown/bitburner.server.hackdifficulty.md
@@ -9,5 +9,5 @@ Server Security Level
 **Signature:**
 
 ```typescript
-hackDifficulty?: number;
+hackDifficulty: number;
 ```
--- a/markdown/bitburner.server.md
+++ b/markdown/bitburner.server.md
@@ -16,27 +16,27 @@ export interface Server
 
 |  Property | Modifiers | Type | Description |
 |  --- | --- | --- | --- |
-|  [backdoorInstalled?](./bitburner.server.backdoorinstalled.md) |  | boolean | _(Optional)_ Flag indicating whether this server has a backdoor installed by a player |
-|  [baseDifficulty?](./bitburner.server.basedifficulty.md) |  | number | _(Optional)_ Server's initial server security level at creation. |
+|  [backdoorInstalled](./bitburner.server.backdoorinstalled.md) |  | boolean | Flag indicating whether this server has a backdoor installed by a player |
+|  [baseDifficulty](./bitburner.server.basedifficulty.md) |  | number | Server's initial server security level at creation. |
 |  [cpuCores](./bitburner.server.cpucores.md) |  | number | How many CPU cores this server has. Affects magnitude of grow and weaken ran from this server. |
 |  [ftpPortOpen](./bitburner.server.ftpportopen.md) |  | boolean | Whether or not the FTP port is open |
-|  [hackDifficulty?](./bitburner.server.hackdifficulty.md) |  | number | _(Optional)_ Server Security Level |
+|  [hackDifficulty](./bitburner.server.hackdifficulty.md) |  | number | Server Security Level |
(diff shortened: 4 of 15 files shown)


d0sboots:
Before, a player could tell if a server was a hacknet server due to these fields being `undefined`. How do they tell now?

catloversg:
I guess via a new property in [2] (a boolean `isHacknetServer` or a union string type)? It was a part of the discussion, but nothing is concluded. The simplest solution is to expose the current `isHacknetServer` property.

d0sboots:
Out of all the breaking changes we've made, I'd say this is the one I'm most on the fence about (and I've been *really* on the fence about some of them). IMO this is not "better" or "worse", but just "different." There are definite pros and cons to both ways, as evidenced by the long discussion around this topic. However, I'm a bit hesitant to change these values in a way that both:
1) Has potential to cause *very* widescale breakage, and
2) Can't be mitigated in *any* automated fashion,

without a really clear benefit.

catloversg:
I agree that these changes are questionable when we consider the trade-offs. That's also why I did not advocate these changes.

I'll mark this PR as a draft one. Let's come back when we find more benefit.

catloversg:
This change was put on hold, so I want to doublecheck other suggested changes that were mentioned while we discussed this change.

Suggestion 1: Throw an error when the server is invalid (Passed the vote on Discord https://discord.com/channels/415207508303544321/1355531035470860338/1355531233479889167)
When the player calls APIs that need a server name, we check if that server name is valid (the server exists and is an instance of Server). If it's not (i.e., the server is a hacknet server), we reject the promise with async APIs (e.g., hack, grow, weaken), but we return 0 with most other sync APIs.

Suggestion: With other APIs mentioned in #1377, we also throw an error.

Ref:
- https://github.com/bitburner-official/bitburner-src/blob/6d3e111ea14dbde0e436598c20cc51eff2a5e0d8/src/NetscriptFunctions.ts#L329-L333
- https://github.com/bitburner-official/bitburner-src/pull/1377

Suggestion 2: #1128
This PR was abandoned.

[1] is the same as this PR. It is a big breaking change that affects many APIs, but it does not bring too many benefits. [2] is smaller, so it's more acceptable. However, only touching one property of `Server` also means that it does not solve the core problem (messy design of internal structures and NS APIs) that this PR and [1] try to solve.

@d0sboots What do you think about [1] and [2]?

d0sboots:
[1] I think has big benefits, and the time to do it is now, if we're going to. The one thing to be careful of is BN14, because it has disappearing servers, so ultimately it was decided that it can't throw for invalid server. (It's a data error, not a programmer error.) Most of its functions are segregated in its own object, but I think it might still reuse some from ns, like scp? It is worth checking. 

[2] I'm pretty neutral on. Seems worth it to make two APIs more consistent, I guess.

PR #2181, opened 2025-06-03
UI: Scroll to top when opening new NS API doc page in popup mode

catloversg:
At 00:20 of the demo video in #2163, after clicking on the link of the "Corporation" page, the popup should scroll to the top.

`window.scrollTo` does not work in this case. We need to scroll the wrapper div in the modal.

https://github.com/user-attachments/assets/f3658af4-aaec-4ff5-8da6-b69a6ba7c84e

### DIFF
--- a/src/ui/MD/MD.tsx
+++ b/src/ui/MD/MD.tsx
@@ -11,14 +11,40 @@ import rehypeRaw from "rehype-raw";
 import { FilePath } from "../../Paths/FilePath";
 import { getPage } from "../../Documentation/root";
 
-export function MD(props: { pageFilePath: FilePath; top: number }): React.ReactElement {
-  const pageContent = getPage(props.pageFilePath);
+export function MD({
+  pageFilePath,
+  top,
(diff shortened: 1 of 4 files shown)


src/ui/MD/MD.tsx
```diff
@@ -11,14 +11,40 @@ import rehypeRaw from "rehype-raw";
...
+   * If this component is used for rendering the documentation popup, this parameter is the ref of the wrapper div in
+   * the modal.
+   */
+  modalWrapperRef?: React.RefObject<HTMLDivElement>;
+}): React.ReactElement {
+  const pageContent = getPage(pageFilePath);
 
+  /**
+   * After rendering the doc page, we need to scroll it. This component is used in 2 places:
+   * - src\Documentation\ui\DocumentationRoot.tsx: The documentation tab.
```
d0sboots: Can we not put the scrolling logic in these files, or pass it in via a prop or something? The "if" behavior is rather icky.

PR #2182, opened 2025-06-03
DEVMENU: Auto expand last opened tools

catloversg:
Having to reopen tools after switching tabs or reloading is too annoying. This PR adds a new feature to save the list of opened tools and auto expand them.

I also remove the useless animation of accordion components used in the dev menu.

https://github.com/user-attachments/assets/cc3fbb89-f4e7-4aa0-97b3-f9b027b72232


src/DevMenu.tsx
```diff
@@ -27,18 +27,38 @@ import { EntropyDev } from "./DevMenu/ui/EntropyDev";
...
 
 export function DevMenuRoot(): React.ReactElement {
+  const autoExpandContextValue = useRef({
+    data: getAutoExpandData(),
+    set: (key: string, expanded: boolean) => {
+      autoExpandContextValue.current.data[key] = expanded;
+    },
+  });
+
+  useEffect(() => {
```
d0sboots, line 41: This before unmount/before unload stuff seems potentially fragile. Why not just save the state every time something is expanded?

Edit: That *is* what you're doing, so I don't understand this unmount stuff.

PR #2194, opened 2025-06-07
MISC: Detect circular dependencies when generating modules

catloversg:
When the player's scripts have circular dependencies, we throw this error: `RangeError: Maximum call stack size exceeded`. Most players do not know what this error means and why it happens. This PR adds a new check to detect circular dependencies and show a clear error message.

Note that, technically, having circular dependencies is still okay in some cases. For example:

Case 1:
```js
// test.js
import * as test from "./test.js";
  console.log(test.main);
}
main({});
```

Case 2:
```js
// test.js
import { libValue } from "./lib.js";
export const testValue = "test";
  console.log(libValue);
}
main({});
```
```js
// lib.js
import { testValue } from "./test.js";
export const libValue = "lib";
```

These cases work normally when running them in NodeJS or an HTML page.

However, due to how our internal works, we have a chicken-and-egg problem. With this script (`test.js` imports itself):
```js
import * as test from "./test.js";
}
```
We have to convert it to:
```js
import * as test from "blob:https://<url>/<guid>";
}
```
After that, we generate a blob url for the new modified code.

This means:
- We must know the blob url of `test.js` before generating the modified code.
- We must have the modified code before generating the blob url of `test.js`.

I'm not sure if we can work around this problem. Maybe it's doable with a refactor and some weird tricks, but I think it's not worth the effort.

Save file:
[test-circular-deps.gz](https://github.com/user-attachments/files/20639707/test-circular-deps.gz)

Before:

before

After:

after


d0sboots:
Looks good. The merge resolution was more than I wanted to deal with on my phone, though.

PR #2198, opened 2025-06-11
IPVGO: Ensure that player promises are changed to "gameOver" once the game is over

ficocelliguy:
When the game ends, the pending player promise is updated with type "gameOver". However, previously, the other player's promise was never updated, and maintained their last move before the game ended (passing). 

This PR fixes it to ensure that those promises are updated to also show game over once the game ends. 

context:
https://discord.com/channels/415207508303544321/415213413745164318/1382003105792524359

### DIFF
--- a/src/Go/boardAnalysis/goAI.ts
+++ b/src/Go/boardAnalysis/goAI.ts
@@ -62,7 +62,6 @@ export function handleNextTurn(boardState: BoardState = Go.currentGame, useOffli
     // The game is over. We shouldn't get here in most circumstances,
     // because when the game ends resetAI() will be called to resolve promises.
     // Return an already-resolved promise until a new game is started.
-    resetAI(true);
     return Promise.resolve(gameOver);
   }
   const currentColor = previousColor === GoColor.black ? GoColor.white : GoColor.black;
@@ -138,10 +137,10 @@ export function resetAI(endOfGame = false): void {
     if (playerPromise.resolver) {
       playerPromise.resolver(gameOver);
       playerPromise.resolver = null;
-    } else {
-      playerPromise.nextTurn = Promise.resolve(gameOver);
     }
-    if (!endOfGame && !playerPromise.resolver) {
+    if (endOfGame) {
+      playerPromise.nextTurn = Promise.resolve(gameOver);
+    } else {
       createPromise(playerPromise);
     }
   }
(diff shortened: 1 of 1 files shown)


d0sboots:
I haven't forgotten about this one, the issue is that the promise handling is (as always) complicated and I need to pull all of it into my brain to be sure this is the right fix.

src/Go/boardAnalysis/goAI.ts
```diff
@@ -137,6 +138,8 @@ export function resetAI(endOfGame = false): void {
     if (playerPromise.resolver) {
       playerPromise.resolver(gameOver);
       playerPromise.resolver = null;
+    } else {
+      playerPromise.nextTurn = Promise.resolve(gameOver);
     }
     if (!endOfGame && !playerPromise.resolver) {
       createPromise(playerPromise);
```
d0sboots:
```suggestion
    if (endOfGame) {
      playerPromise.nextTurn = Promise.resolve(gameOver);
    } else {
      createPromise(playerPromise);
```

I think what you had before would possibly fail in some subtle edge-cases involving multiple calls to resetAI. The real condition to be testing for is `endOfGame`. And, unless I'm missing something, the `!playerPromise.resolver` check is dead code (probably from an earlier iteration), so we can eliminate that and streamline this.

This breaks the idempotence of resetAI slightly, in that calling it repeatedly with `endOfGame = true` will continually create new promises. But, they all resolve to the same value, and generally it's bad form to be doing object identity checks on promise objects, and we don't call `endOfGame = true` multiple times anyway, really.

src/Go/boardAnalysis/goAI.ts
```diff
@@ -62,6 +62,7 @@ export function handleNextTurn(boardState: BoardState = Go.currentGame, useOffli
     // The game is over. We shouldn't get here in most circumstances,
     // because when the game ends resetAI() will be called to resolve promises.
     // Return an already-resolved promise until a new game is started.
+    resetAI(true);
```
d0sboots: This shouldn't be needed (endGoGame *is* being properly called, here), but I suppose it doesn't hurt.

src/Go/boardAnalysis/goAI.ts
ficocelliguy: I applied a slight variant of this (the proposed suggestion is missing a curly brace, I think)

PR #2199, opened 2025-06-13
MISC: Add file metadata (timestamps)

catloversg:
I decided to take over #1271.

I reused most of the code in that PR and made these changes:
- Revert unrelated changes in `src\Netscript\NetscriptHelpers.tsx`.
- Change `ContentFile` from an interface to an abstract class and move `metadata` to the new abstract class.
- Fix a bug in `BaseServer.writeToTextFile`. We have to set the text data via `content`, not `text`.
- Make `ns.getFileMetadata` throw an error when the filename is invalid instead of returning undefined. It seems that Hoekstraa reused the "style" of `ns.read`, but I think the behavior of `ns.read` is bad. We have been too lenient when it comes to param checking. Returning an empty string when the param is invalid just hides the programming error of the player. It may be convenient for some people, especially newbies who hate seeing errors, but those errors will eventually come back and wreck havoc in other parts of their scripts. Changing that behavior is a big breaking change, so I won't do anything now, but I don't want to see that behavior in new APIs.

Test tool:
[rfa-tool.zip](https://github.com/user-attachments/files/20729432/rfa-tool.zip)

How to test:
- Download my attached test tool.
- Run `node main.js`. It will start a WebSocket server on port 12525 and an HTTP server on port 3000.
- Connect the game to port 12525.
- Use curl to test RFA APIs. For example: `curl http://localhost:3000 -d '{"jsonrpc":"2.0","method":"getAllFileMetadata","params":{"server":"home"},"id":1}'`.

I tested these cases:
- CLI: cat, nano, cp, mv
- NS API: read, write, scp, mv
- RFA API: pushFile, getFile, getFileMetadata, getAllFiles, getAllFileMetadata

Closes #1271.


src/RemoteFileAPI/MessageHandlers.ts
```diff
@@ -54,6 +54,29 @@ export const RFARequestHandler: Record<string, (message: RFAMessage) => RFAMessa
...
+
+    if (!hasTextExtension(filePath) && !hasScriptExtension(filePath)) return error("Invalid file extension", msg);
+    const file = server.getContentFile(filePath);
+    if (!file) return error("File doesn't exist", msg);
+
+    const result: FileMetadata = {
+      filename: file.filename,
+      atime: file.metadata.atime,
+      mtime: file.metadata.mtime,
+      btime: file.metadata.btime,
```
d0sboots:
```suggestion
      ...file.metadata.plain(),
```

src/RemoteFileAPI/MessageHandlers.ts
```diff
@@ -93,6 +116,21 @@ export const RFARequestHandler: Record<string, (message: RFAMessage) => RFAMessa
...
+    if (!isFileServer(msg.params)) return error("Message misses parameters", msg);
+
+    const server = GetServer(msg.params.server);
+    if (!server) return error("Server hostname invalid", msg);
+
+    const fileList: FileMetadata[] = [...server.scripts, ...server.textFiles].map(([filename, file]) => ({
+      filename: filename,
+      atime: file.metadata.atime,
+      mtime: file.metadata.mtime,
+      btime: file.metadata.btime,
```
d0sboots: For future-proofing.
```suggestion
      ...file.metadata.plain(),
```

d0sboots (commented):
Nice, very solid. (Although you were building from good bones.)

PR #2202, opened 2025-06-14
BUGFIX: Documentation navigator does not handle external URL properly

catloversg:
This is a regression bug caused by #2163. Some docs pages (NS and non-NS) include links to HTTP URLs, but when clicking on these links, the navigator does not open them. For example, `bitburner.reactnode.md` has a link pointing to `https://react.dev/reference/react/createElement#creating-an-element-without-jsx`.

While fixing this bug, I also made these related changes:
- Search for all usages of `https://github.com/bitburner-official/bitburner-src` in our codebase and point them to the stable branch.
- Use https if the external URL supports it.
- Rename `isFilePath` to `isBasicFilePath` and add a new `isFilePath`. When I used `isFilePath`, I found out that `isFilePath` does not do what its name implies.

I tested both the documentation tab and the new documentation popup.


src/Documentation/ui/DocumentationRoot.tsx
```diff
@@ -15,19 +21,48 @@ export function DocumentationRoot({ docPage }: { docPage?: string }): React.Reac
   const history = useHistory();
   const [deepLink, setDeepLink] = useState(docPage);
   const navigator = {
-    navigate(relativePath: string, external: boolean) {
-      const path = relativePath.startsWith("nsDoc/")
-        ? asFilePath(relativePath)
-        : resolveFilePath("./" + relativePath, history.page);
+    navigate(href: string, external: boolean) {
```
d0sboots: Optional: Rename the 2nd param to "openExternally" to make it clear that you're specifying the action, not a property of the href.

src/ui/React/Documentation.tsx
```diff
@@ -92,8 +93,13 @@ export const HistoryProvider = (props: React.PropsWithChildren<object>): React.R
 
 export function openDocExternally(path: string) {
   const ver = CONSTANTS.isDevBranch ? "dev" : "stable";
-  const url = path.startsWith("nsDoc/")
-    ? `https://github.com/bitburner-official/bitburner-src/blob/${ver}/markdown/${path.replace("nsDoc/", "")}`
-    : `https://github.com/bitburner-official/bitburner-src/blob/${ver}/src/Documentation/doc/${path}`;
+  let url;
```
d0sboots, line 96: Is this used outside of navigate? If not, it feels like it would be better to keep all this URL detection code in the same function, and it would simplify the logic.

catloversg: This function is called by 4 callers:
- `navigate` in `src\Documentation\ui\DocumentationRoot.tsx` [1]
- `navigate` in `src\Documentation\ui\DocumentationPopUp.tsx` [2]
- `NsApiDocumentationLink` in `src\ui\React\NsApiDocumentationLink.tsx` [3]
- `DocumentationAutocomplete` in `src\ScriptEditor\ui\Toolbar.tsx` [4]

[1] and [2] have their own URL detection logic. They need to parse its `href` and convert `href` as needed. It's a bit hard to consolidate the logic in [1] and [2].
[3] always passes `defaultNsApiPage` (`"nsDoc/bitburner.ns.md"`).
[4] always passes its `path`. `path` always starts with `nsDoc/`.

It's possible to create a function that contains the URL detection logic of both [1], [2] and `openDocExternally`, then convert the path as needed before passing it to `openDocExternally` (in this case, `openDocExternally` simply opens the path without any parsing code). However, I think the result will be much more confusing than having separate functions doing their own jobs.

Issue #250, opened 2022-12-08
[Balance] Infiltration SOA Rewards don't account for infiltration length


Draco18s:
As it is right now I can either infiltrate Rho Constrution (Maximum Level: 5) and get 3509.28 reputation for Shadows of Anarchy, or I can infiltrate Bachman and Associates (Maximum Level: 15) and get 5725.30

I can do the first one about three times in the same time it takes me to do the second with the same likelihood of success on each challenge (99% as both are trivial 0/100 difficulty). Oh and if I run out of HP on the 15th challenge doing B&A I'm out the full 5700 whereas if I was 15 deep in Rho, I'd be out 3500 (of 10500), keeping 7000.

[The code](https://github.com/bitburner-official/bitburner-src/blob/19e8946d9aed90a26be35b8f8999238cdd70d5e5/src/Infiltration/formulas/victory.ts#L33) only takes into account the starting security level of the targeted company (485 for Rho and 1350 for B&A) and the difficulty of the infiltration (so....0) with all other factors even (whether or not I have the WKS Harmonizer, my favor with SOA, etc).

So why would I ever infiltrate a place other than Joe's Guns if I want to get the most faction rep possible for the least amount of time and effort?

I mean sure I suppose I could infiltrate Megacorp's 50 levels deep, but my expected payout is on the order of 10,000 rep, which I can get by infiltrating Rho three times. I don't even see this being any better when automated with sleeves/singularity/etc: wouldn't it still be more efficient to go after the weaker targets forever?

For comparison, Clarke Inc has a (ui displayed) difficulty of 36 and a depth of 18 and the reward is... a whopping 6,676.02--still less than double Rho Construction.

- [x] Steps to reproduce
- [x] Version of the game, e.g. Bitburner v2.1.0 (8f4636cb)

d0sboots:
This is generally addressed by the recent infil changes.

PR #2210, opened 2025-06-20
INFIL: Add stat requirements; Add market consequences for spamming infiltration

ficocelliguy:
Currently, infiltration is not in a great place.

Right now. infiltration rewards actually DECREASE with stat multipliers. This causes player confusion, and adds strange incentives to avoid augments and stats.
https://discord.com/channels/415207508303544321/415207508303544323/1381687590687740014
https://discord.com/channels/415207508303544321/415207923506216971/1236521542469550090
https://discord.com/channels/415207508303544321/415207508303544323/1231347988312494191
https://discord.com/channels/415207508303544321/415213435974975508/1186943469126688789
https://discord.com/channels/415207508303544321/415207923506216971/1223048550330400932

Automated infiltration can spam infils so rapidly that it ruins the game for people. We've seen this many times, particularly when a player just downloads and blindly runs someone else's script.
https://discord.com/channels/415207508303544321/415207839246581781/1286821539647586336
https://discord.com/channels/415207508303544321/415207839246581781/1338986492508377129
https://discord.com/channels/415207508303544321/928919860891750410/1283125485269811242
https://discord.com/channels/415207508303544321/415207923506216971/1213880834105417778

Infiltration can be boring/grindy but sometimes feels like the optimal strat to advance, causing unfun play patterns.
https://discord.com/channels/415207508303544321/415207508303544323/1318607454782754837
https://discord.com/channels/415207508303544321/415207923506216971/1254214428522713208
https://discord.com/channels/415207508303544321/415207508303544323/836743122024923176

Goals:

1) Reduce the power of automated infiltration, with minimal impact on manual infil
2) Reduce the incentive to spam as many infiltration as possible back-to-back, which is not super fun compared to the rest of the game
3) Simplify the confusing reward system, and do not punish players for installing augments and progressing their character

Approach:

1) Reduce the efficacy of spamming lots of infiltrations in a short time period. 
   - This will have a VERY outsized effect on automated infil, which can complete runs inhumanly fast. 
   - This also has the effect of making manual infiltration better for short sessions or occasional runs, rather than incentivizing marathon grinds.
   - In this PR, each successful infil within the last 30 minutes reduces the rewards of future infiltrations by 0-15% based on how recently they were completed. (This can be balanced as needed.) 

2) Added a new difficulty, "Suicide", at displayed difficulty rating 150+
   - (Difficulty is now shown uncapped for clarity) 
   - If attempted at this new difficulty, the player is sent straight to the hospital.  
   - This requires auto infil to have some minimum physical stats to do higher-difficulty infiltrations, like how manual infil already effectively does.
   - This makes infiltration _reward_ player investing in and augmenting their stats (instead of punishing it, as it does currently)
   - This change has a huge effect on automated infiltration, which currently maximizes rewards by being as low of phys stats multi as possible and targeting megacorps, but very little effect on manual.
   - High-powered auto-infil is still possible, but requires player investment to unlock, and is _much_ weaker right after an install and at bitnode start.
   - In this PR, entry locations (like joes, difficulty 3) require no stats, harder places (like 4sigma, difficulty 8) require ~350, and megacorps (like megacorp, difficulty 16) require ~1500 average physical stats. (This can be balanced as needed.) 

3) Make rewards scale with difficulty, not with player mults
  - Since some baseline stats are required for most infiltrations now, this ends up being a fairly small effect
  - The rewards are broadly unchanged for players infiltrating manually with reasonable stats & augments, see examples below
  - Later in a bitnode, this ends up being a pretty significant buff to manual infiltration, as the rewards are no longer super debuffed from buying NFGs and augments
  
  
  Some example reward comparisons of before (top) and after (bottom) this balance proposal. These results are for the first infiltration completed; the next one will have slightly lower rewards due to some market demand being filled by the first.
  
|               | joesguns @ lv1       | joesguns @ lv 1k, 3x mults | 4sigma @ lv1            | 4sigma @ lv 1k, 3x mults | megacorp @ lv1           | megacorp @ lv 1.5k, 3x mults |
| ------------- | -------------------- | -------------------------- | ----------------------- | ------------------------ | ------------------------ | ---------------------------- |
| current rep   | 2k                   | 600                        | 120k                    | 72k                      | 780k                     | 620k                         |
| current $$    | 4m                   | 480k                       | 3.5b                    | 1.3b                     | 150b                     | 98b                          |
|               |                      |                            |                         |                          |                          |                              |
|               | Level requirement: 1 |                            | Level requirement: ~350 |                          | Level requirement: ~1500 |                              |
| changed rep   | 2.2k                 | 2.2k                       | 102k                    | 102k                     | 620k                     | 620k                         |
| changed money | 5m                   | 5m                         | 2.5b                    | 2.5b                     | 93b                      | 93b                          |

Example warning message that a location is too high a difficulty for your stats. Attempting will immediately send you to the hospital, with a toast explaining that you were instantly caught due to insufficient skill level.

Note that the difficulty rating is shown uncapped for reference as the player approaches the required skills. 
 
Also note the market demand % is shown (and updated live as it slowly recovers back to 100% after an infiltration is completed)
image


catloversg:
The ideas are good, but I'll need a few days to playtest and analyze the specific changes. Infiltration has always been a landmine, so I have refrained from touching it. Every time we make a change, no matter what it is, we always receive some kinds of "why does the dev want to ruin our fun" feedback. This PR will nerf the auto-infil mechanic, which is despised by some players and loved by other players. I'm totally fine with it, but we should be careful while doing it. I'll try to review this change ASAP.

Nitpick:
> This change has a huge effect on automated infiltration, which currently maximizes rewards by being level 1 phys stats and targeting megacorps, but very little effect on manual.

I think you should rephrase or clarify this part. It's not wrong, but it can be misinterpreted by some uninformed players. The auto-infil scripts *can* run at level 1 stats because they don't *need* to waste the time to improve the stats. It's efficient because it does not waste time. However, some players think that running at level 1 stats is *necessary* to be optimal because rewards scale with stats level (i.e., low level = high reward), which is a misconception.

d0sboots:
I also agree that the broad ideas are good here, and I'm line with approaches we have discussed in the past. I will be relying on catlover (and maybe others) for playtesting of the specifics, while I look at the code. As always, balance will likely be the trickiest part.

ficocelliguy:
I updated the wording on the "level 1" bit in the description to clarify that the old optimal was to keep physical & charisma stat mults as low as possible for max reward.

I'm also very open to suggestions and different approaches, these just seemed like the simplest ones that made sense.

catloversg:
## Balance change
- Set MaxDif to 3.5
- Use different formulas for rewards.

For more information, please check #2222.

## Data structure

- `successfulInfiltrationTimestamps` contains timestamps instead of date objects.
- Save entire `successfulInfiltrationTimestamps` instead of the length. I don't get why we have to save the length and then "simulate" the old values. It's more complicated, and there is a chance that the player will find a way to exploit our way of "spreading out" the values. Just save and load them as-is. In 30 minutes, there won't be many values in that array.

## UI change

- I change "Impossible-Suicide" to "Brutal-Impossible". The current "Impossible" difficulty means "extremely hard" and the new "Suicide" difficulty means "You cannot do that", so I think "Brutal-Impossible" is a better choice.
- Disable the button "Start" instead of showing "Attempt Anyway". If the player cannot infiltrate, just don't allow them to press the button. The error message already explains the reason.

## Bugfix

### Bug 1



The difficulty bar needs to be filled to the end.

### Bug 2

The intro screen updates the market demand multiplier, but it does not update the rewards. This is because `reward` is calculated in `src\Infiltration\ui\InfiltrationRoot.tsx` and passed to `Game` and `Intro`. With the new market demand multiplier, you have to recalculate the rewards in those child components.

### Bug 3

The player can bypass the restriction of the new market demand multiplier. In `src\Infiltration\ui\Victory.tsx`, you add data to `InfiltrationState.successfulInfiltrationTimestamps` in `sell` and `trade`, but if the player closes that popup by pressing "Quit", those functions are not called. They can exploit this to farm SoARep without the restriction of the new multiplier.

### Bug 4

This is not exactly a bug. It still works, but the usage of React hooks is wrong.

```ts
  useEffect(() => {
    const clearSubscription = PlayerEvents.subscribe((eventType) => {
      if (eventType !== PlayerEventType.Hospitalized) {
        return;
      }
      cancel();
      dialogBoxCreate("Infiltration was cancelled because you were hospitalized");
    });
    return clearSubscription;
  }, []);
```

This block is supposed to be run *once* after the first render. When you put the new check in it, you also have to add `onFailure` and `props.Difficulty` to the dependency list. This makes the hook run every time `onFailure` is changed, and `onFailure` is changed every time this component renders a new mini game (`stageComponent`). That check should be in a separate `useEffect`.

d0sboots:
> Save entire successfulInfiltrationTimestamps instead of the length. I don't get why we have to save the length and then "simulate" the old values. It's more complicated, and there is a chance that the player will find a way to exploit our way of "spreading out" the values. Just save and load them as-is. In 30 minutes, there won't be many values in that array.

What about something like an exponential moving average? Basically, I'm thinking of some function that has very little state (one or two values tops) to control the time-based capping, and then you also only need to save those one or two values.

d0sboots:
More detailed thoughts on that: The theory https://en.wikipedia.org/wiki/Exponential_smoothing

The practice: Keep a timestamp of the last time "infils/sec" was updated, and its value. The value decays back to 0 exponentially with a relatively long time constant, and you calculate this (but don't update the stored value) whenever market demand needs to be displayed/used. 

If an infiltration is attempted, the value at the start is used. If successful, the values are updated using the timestamp of *when the infiltration was started*. This applies the duration of infiltration towards the decay, to avoid penalizing manual infil.

ficocelliguy:
What you're describing is basically what it already does in this branch @d0sboots . Each infil completion timestamp affects the rewards, with the effect decreasing the older it is, and there's no snapshotting, so you get the benefit of market recovery during the duration of your infiltration.

ficocelliguy:
@catloversg I had intentionally put the market effects to only apply if you actually sell/donate the stolen info to the market, to not penalize players who closed out without redeeming the reward. I had also intended to not affect  SoA rep based on the market, since its sort of its own thing, and only really benefits manual infil. I can change either/both of those if they're an issue, though.

d0sboots:
> What you're describing is basically what it already does in this branch @d0sboots . Each infil completion timestamp affects the rewards, with the effect decreasing the older it is, and there's no snapshotting, so you get the benefit of market recovery during the duration of your infiltration.

Yes, the intention was not to be a significant change in effect, but rather just a simplification in implementation.

d0sboots:
> @catloversg I had intentionally put the market effects to only apply if you actually sell/donate the stolen info to the market, to not penalize players who closed out without redeeming the reward. I had also intended to not affect SoA rep based on the market, since its sort of its own thing, and only really benefits manual infil. I can change either/both of those if they're an issue, though.

I like the first part, it's an attention to detail based on the new flow (although it's kind of weird to complete an infil and then just discard it, but...)
I think probably SoA rep should work the same as normal rewards, for consistency? Certainly I don't see any downside to doing so, and there's potential exploit issues if not.

catloversg:
> More detailed thoughts on that: The theory https://en.wikipedia.org/wiki/Exponential_smoothing
> 
> The practice: Keep a timestamp of the last time "infils/sec" was updated, and its value. The value decays back to 0 exponentially with a relatively long time constant, and you calculate this (but don't update the stored value) whenever market demand needs to be displayed/used.
> 
> If an infiltration is attempted, the value at the start is used. If successful, the values are updated using the timestamp of _when the infiltration was started_. This applies the duration of infiltration towards the decay, to avoid penalizing manual infil.

I think it's okay. Unfortunately, I only somewhat understand it, so it's a bit hard for me to implement it correctly. I appreciate it if anybody implements it for me.

Regarding SoARep, I don't have a strong opinion about it. fico's explanation makes sense, but I also want to consider #250. This is my opinion about it (copied from #2222):

---

With SoARep, we also have to consider #250. On one hand, I agree that it's weird not to take the number of floors into account. With the current formula, locations with a low number of floors are much better than ones with a high number of floors (e.g., "Joe's Guns" is very good at the early game, and "Rho Construction" is the best location). On the other hand, with the suggestion of "market demand multiplier" in PR 2210, this will not be a problem anymore.

With the new "market demand multiplier", I think we don't need to tune the SoARep formula. If the player rarely infiltrates, they can "exploit" this quirk to get high SoA rep with low stats on easy locations. This can be seen as a trick, in the same way "joesguns" is the best exp farm target. If they infiltrate nonstop, those easy locations are not the best targets anymore.

---

An upside of using the "market demand multiplier" is that we don't have to consider #250 anymore. If we decide that SoA rep should not be affected by the new multiplier, we also should make a decision on the mentioned issue.

I think that letting SoA rep work the same way as other rewards is simple and does not have any downsides, so I'm a bit leaning on that option.

ficocelliguy:
@d0sboots I have been experimenting with different approaches based on the link you sent, and I think that there is simply not enough information in a setup like the one you proposed with a fixed number of stored values (such as a moving rate and a last-infil timestamp).

For example:
Suppose the player has completed an infiltration 1 minute ago, and the overall market rate is 70%.
This could be because there was a total of two infiltrations, one 1 minute ago and the other 2 minutes ago. (Case 1)
It could also be because there was one 1 minute ago and fifteen infils that were all 29 minutes ago. (Case 2)

Case 1 means the market rate will take ~28 minutes to recover fully from the infiltrations prior to the most recent one. 
Case 2 means the market will take ~1 minute to recover fully from the infiltrations prior to the most recent one.

Without knowing the actual distribution of values, you cannot know how much the market should be shifting each minute due to infiltrations in the past.

d0sboots:
> @d0sboots I have been experimenting with different approaches based on the link you sent, and I think that there is simply not enough information in a setup like the one you proposed with a fixed number of stored values (such as a moving rate and a last-infil timestamp).
> 
> For example: Suppose the player has completed an infiltration 1 minute ago, and the overall market rate is 70%. This could be because there was a total of two infiltrations, one 1 minute ago and the other 2 minutes ago. (Case 1) It could also be because there was one 1 minute ago and fifteen infils that were all 29 minutes ago. (Case 2)
> 
> Case 1 means the market rate will take ~28 minutes to recover fully from the infiltrations prior to the most recent one. Case 2 means the market will take ~1 minute to recover fully from the infiltrations prior to the most recent one.
> 
> Without knowing the actual distribution of values, you cannot know how much the market should be shifting each minute due to infiltrations in the past.

Yes, it is impossible for an exponential moving average to *exactly* replicate the behavior of an average with a fixed window. That is not the purpose. Indeed, with an exponential moving average you never "fully" recover from doing an infiltration, in the same way that something radioactive never becomes "completely" inert - but in both cases they fall to a level that becomes undetectable. 

The purpose is to provide a simple and predictable mechanism: The decay rate is *solely* based on how much built up infil there is (i.e. market demand), so there's no "hidden" variables like the demand suddenly dropping because an old infiltration expired out of the window. If you do a certain number of infils/sec, the market demand will exponentially approach a constant, and exponentially change to a different constant if you change your rate of infils. The behavior of a fixed window function is more complicated to analyze, because of the window cutoffs - you have to pay more careful attention to if there are particular patterns that are more profitable than others.

ficocelliguy:
To be more explicit: I was not able to make anything that behaved even close to prior infiltrations' influence decaying over time. As far as I can tell, there's no way to know how much the old measurement should change since you don't know how "dense" it is, only how old the prior measurement is and what the outcome of the prior math was. The exponential smoothing model is a weighted average of observations' values (sum-mean), not a _product_ of decaying scalars. That's not to say that there isn't some way to make a more compact version of it, just that I don't know how and haven't been able to figure one out.

ficocelliguy:
The closest you could get is some sort of fixed recovery rate, like 3% market rate recovered per minute, or "the market rate amount below 1.0 decreases by a 0.3 multiplier every ten minutes", I think?

d0sboots:
OK, I put together a pretty solid version of what I was talking about in https://github.com/d0sboots/bitburner-src/commit/dd4bb19ca4393f5dd733b50e8c28963bc6d22cd0, and you can see it in action at
https://d0sboots.github.io/bitburner-src/. The constants should be much simpler to tune; currently, manual infil barely moves the needle on market demand at all, but automated infil should be ~10-15x better in the best case.

ficocelliguy:
* Pulled in d0s's parabolic reward scaling implementation
* Re-added support for live updates of rewards on the UI, which allowed me to simplify timestamp usage and remove extra props
* Ensured that the player gets market rate recovery during the duration of the infiltration
* Put in a clamp to avoid the market rate going very negative, which would "stick" the visible rewards at 0 for a long time
* Tweaked the market rate hit for an infiltration compared to d0s's branch (his commit had 5x infiltrations in 10 seconds still give 99.9% rewards)

catloversg:
I have not playtested it much, but I spotted a nerf in `src\Infiltration\formulas\victory.ts`. Why do you remove `+ 1` in the formula? It incurs ~27% reduction in most locations (higher % in very early locations) on top of the current nerf. At first glance, it's not that bad because the current proposed rep/s is still fairly high. However, I want to make sure that this additional nerf is intentional.

ficocelliguy:
@catloversg It was intentional to remove the +1, since otherwise a market rate of 0% still gives like 25% rep and 10% cash rewards (see below). The significant nerf to the top end was not intentional, though, I'll add something to counteract that.
image

image

src/Infiltration/formulas/game.ts
```diff
@@ -5,35 +5,41 @@ export const MaxDifficultyForInfiltration = 3.5;
...
-// Tracks an exponential moving average of number of successful infiltrations performed,
-// which decays back to 0. This state is only updated after a successful infil.
+// Tracks an exponential moving average of number of successful infiltrations performed, which decays back to 0.
 export const InfiltrationState = { ...InfiltrationStateDefault };
 
-function calculateCurrentInfils(timestamp: number): number {
-  return InfiltrationState.infils * Math.exp(DecayRate * (timestamp - InfiltrationState.lastChangeTimestamp));
+function calculateCurrentInfils(timestamp = date.now()): number {
+  // The clamp ensures that the market rate is always recovering, and not "stuck" at zero for a long time due to being effectively negative.
+  const clampedInfils = clampNumber(InfiltrationState.infils, 0.01, Math.sqrt(1 / TuningMult) + 0.07);
```
d0sboots: This is the wrong place for this action to happen. Infils is a moving average that tracks the number of infiltrations over time; in particular, clamping it to 0.01 so that it can never get down to 0 is very bad.

Leave the logic in here alone, and change the calculation of market demand multiplier (putting in clamps on infil or whatever) if you want to do this.

src/Infiltration/formulas/game.ts
```diff
@@ -5,35 +5,41 @@ export const MaxDifficultyForInfiltration = 3.5;
...
+  InfiltrationState.lastChangeTimestamp = timestamp;
+  return InfiltrationState.infils;
 }
 
 // Calculates the infiltration reward multiplier based on how many and how recent other infiltrations were completed.
 // Each infiltration completed reduces the demand for corporate espionage data for a little while, thus affecting the
 // market demand.
-export function calculateMarketDemandMultiplier(timestamp: number): number {
-  const infils = calculateCurrentInfils(timestamp);
+export function calculateMarketDemandMultiplier(): number {
```
d0sboots: I didn't plumb timestamp through all these calls just for the fun of it. XD

If you don't pass in timestamp, you don't have a consistent value to make calculations with. This leads to small but subtle bugs where you are relying on multiple calls to Date.now() in succession to return the same value, which is *usually* but not *always* true.

Also, once the player starts the game, IMO the reward should be frozen; what you get at the end should be the same as what it said when you pressed "start". This is only possible with a consistent timestamp. You can still live-update the values in Intro before that point with a refresh hook, if you want.

src/Infiltration/formulas/game.ts
```diff
@@ -5,35 +5,41 @@ export const MaxDifficultyForInfiltration = 3.5;
...
 };
-// Tracks an exponential moving average of number of successful infiltrations performed,
-// which decays back to 0. This state is only updated after a successful infil.
+// Tracks an exponential moving average of number of successful infiltrations performed, which decays back to 0.
 export const InfiltrationState = { ...InfiltrationStateDefault };
 
-function calculateCurrentInfils(timestamp: number): number {
-  return InfiltrationState.infils * Math.exp(DecayRate * (timestamp - InfiltrationState.lastChangeTimestamp));
+function calculateCurrentInfils(timestamp = date.now()): number {
+  // The clamp ensures that the market rate is always recovering, and not "stuck" at zero for a long time due to being effectively negative.
```
d0sboots: The market demand is getting "stuck" because you upped the tuning rate by three orders of magnitude. The value I had set intentionally remains very close to 100% for manual infil, since the purpose of market demand is only to hurt auto infil. I didn't test it, but if my calculations were right market demand should have gone down to 66% with auto infil taking 6s per infil, and faster than that would have gotten fewer rewards.

src/Infiltration/formulas/victory.ts
```diff
@@ -45,7 +45,7 @@ export function calculateTradeInformationRepReward(
   }
 
   return (
-    Math.pow(reward, 1.1) *
+    Math.pow(reward + Math.min(1, reward / 2), 1.1) *
```
d0sboots: I didn't even realize this is how reward gets mixed in. This makes it a lot more complicated to figure out the optional speed of auto infil (for those inclined to do such a thing) - the calculus gets much worse. Why not pass market demand in here and just have it be a linear factor?

src/Infiltration/formulas/game.ts
```diff
@@ -2,35 +2,40 @@ import { Player } from "@player";
 import { clampNumber } from "../../utils/helpers/clampNumber";
 
 export const MaxDifficultyForInfiltration = 3.5;
-const MaxEffectTime = 30 * 60 * 1000; // 30 minutes in milliseconds
-const date = globalThis.Date;
```
d0sboots: I reverted this part in my change because I don't think we should be going out of our way for this sort of thing. There are a myriad of ways that players can attack the game through prototypes, we can't defend against all of them, and this one is relatively benign compared to just speeding the engine ticks up.

d0sboots:
I didn't explicitly state it, but since the goal seems to have been misunderstood I'll try to outline my goal with the exponential moving average more explicitly:

The moving average just tracks the number of infiltrations. Since it's an exponential moving average, it's also essentially tracking the *rate* of infiltrations; the two are more-or-less equivalent (related by a constant multiple) since the derivative of e^x = e^x.

This then drives the market demand formula, which can be anything we want. I chose a parabola since it's about the simplest thing that gets the behavior we want: Smooth, always decreasing, f(0) = 1, and `x*f(x)` (which is the return-on-infiltration, assuming (falsely) that market demand affects the return linearly) has a maximum that is easily calculated and somewhere *before* the maximum speed of auto-infil.

This last point is important - it makes auto-infil a bit more interesting because it means just blasting away as fast as possible will likely produce zero return. Instead, there is a new wrinkle which is to actually figure out (and run at) the best speed. Manual infil remains (almost completely) unaffected.

catloversg:
> @catloversg It was intentional to remove the +1, since otherwise a market rate of 0% still gives like 25% rep and 10% cash rewards (see below). The significant nerf to the top end was not intentional, though, I'll add something to counteract that.

@ficocelliguy I suggest passing the market demand to all specific reward functions (as a multiplier) instead of the generic `calculateReward`. For example:
```js
export function calculateReward(startingSecurityLevel: number): number {
  return clampNumber(calculateRawDiff(465, startingSecurityLevel), 0, 3);
}
export function calculateTradeInformationRepReward(
  reward: number,
  maxLevel: number,
  startingSecurityLevel: number,
  marketDemandMultiplier: number,
): number {
  // Other code
  return (
    Math.pow(reward + 1, 1.1) *
    Math.pow(startingSecurityLevel, 1.1) *
    balanceMultiplier *
    marketDemandMultiplier *
    30 *
    levelBonus *
    (Player.hasAugmentation(AugmentationName.WKSharmonizer, true) ? 1.2 : 1) *
    currentNodeMults.InfiltrationRep
  );
}
```
I agree with d0sboots at https://github.com/bitburner-official/bitburner-src/pull/2210#discussion_r2191193120. I (and maybe d0sboots) tuned our formulas when this part was still `reward + 1`. I think we should keep it `reward + 1`, then tune the `balanceMultiplier` and the new market demand multiplier. It's easier that way.

> This last point is important - it makes auto-infil a bit more interesting because it means just blasting away as fast as possible will likely produce zero return. Instead, there is a new wrinkle which is to actually figure out (and run at) the best speed. Manual infil remains (almost completely) unaffected.

@d0sboots This is interesting. Before your change (exponential moving average), ultra-fast auto-fil (1 second per floor) is still the best way to farm rep as long as the session is not too long (e.g., 60-120m). I'll playtest with the new formula to see how it works.

One question: Which commit should I playtest? Is it your dd4bb19 or fico's 174c726? It seems that you pointed out some problems with fico's commit.

d0sboots:
> One question: Which commit should I playtest? Is it your [dd4bb19](https://github.com/bitburner-official/bitburner-src/commit/dd4bb19ca4393f5dd733b50e8c28963bc6d22cd0) or fico's [174c726](https://github.com/bitburner-official/bitburner-src/commit/174c72651d8162fc825bd54904e248715c4ab746)? It seems that you pointed out some problems with fico's commit.

I would say mine, because I think the value of the tuning constant fico has is quite off from what we want. But, neither are truly the final form, because both are taking market value through reward (thus it gets an exponent applied) when I think we want it to be linear.

d0sboots:
Also, as to testing: I think there are two dimensions to test, trying to farm rep over a long time (so 120min would be the correct timeframe) and trying to quickly get rep as a bootstrap (so short timeframes are appropriate). For the latter, the absolute amount of rep is what is most interesting, whereas for the former it is rep/s.

So, rather than varying time for a single rate, I think it would be more informative in your tables to keep a single time and vary the rates. If the tuning is working correctly, a clear optimum rate should emerge. (But this may be muddied by the fact that different locations have different numbers of floors.)

catloversg:
> Also, as to testing: I think there are two dimensions to test, trying to farm rep over a long time (so 120min would be the correct timeframe) and trying to quickly get rep as a bootstrap (so short timeframes are appropriate). For the latter, the absolute amount of rep is what is most interesting, whereas for the former it is rep/s.

Okay.

> So, rather than varying time for a single rate, I think it would be more informative in your tables to keep a single time and vary the rates. If the tuning is working correctly, a clear optimum rate should emerge. (But this may be muddied by the fact that different locations have different numbers of floors.)

Can you clarify a bit more? I don't understand the term "rate" in this paragraph. I assume "time" is the timeframe (e.g., 5 minutes, 30 minutes, 120 minutes), does "rate" mean "rep/s when infiltrating non-stop"? Can you give an example of "a single time and vary the rates"? Just column name is okay.

d0sboots:
Yes, you interpreted correctly. For column names, maybe have everything past a certain point be 120m and then do 1/s, 2/s, etc? (Exact numbers based on what's realistic and relevant.) (Those could be infils or floors/sec, depending on how you want to compare across locations)

src/Infiltration/formulas/game.ts
ficocelliguy: ten infiltrations in ~30 seconds gave me this 99.8% market rate with the 1.5e-5 you initially set up:
image

ten infils in ~30 seconds with 1.5e-4 gave this:
image

but ten in under minute is definitely well into the auto-infil range, and should not be giving 90%+ rewards, I think.

I suspect your math is after the script has been doing infiltrations for a pretty good while? in which case the player already has basically infinite reputation collected, as long as its any medium-difficulty infil.

src/Infiltration/formulas/game.ts
ficocelliguy: I reverted most of my changes, except for some minor ones to have live-updating on the intro screen. Rewards are frozen on start now.

src/Infiltration/formulas/victory.ts
ficocelliguy: Changed to move market demand here

ficocelliguy:
I've rolled back most of my changes since d0s's latest branch. I've picked a value for the market demand multiplier between the one I had and d0s' original. Note that the current value in this branch (1.5e-4) still has almost no affect on super fast infiltrations in my testing, giving over 95% rewards still when a dozen of them are complete in a very short window, so it still is not perceptibly affecting auto-infil until the script has already farmed all the rep it needs for the current install.

I think that the rewards should drop very sharply for infiltrations that are clearly faster than manual ( below ~1 second per floor + countdown times) but not affect manual speed infiltrations very much.

src/Infiltration/formulas/game.ts
d0sboots: > ten infiltrations in ~30 seconds gave me this 99.8% market rate with the 1.5e-5 you initially set up: ...
> 
> but ten in under minute is definitely well into the auto-infil range, and should not be giving 90%+ rewards, I think.
> 
> I suspect your math is after the script has been doing infiltrations for a pretty good while? in which case the player already has basically infinite reputation collected, as long as its any medium-difficulty infil.

With the current DecayRate 10 infils in 30 seconds gives infils of 9.8662, not far from the value of 10 you get by doing 10 infils in 0 seconds. It is also equivalent to a steady-state rate of 1 infil every 107 seconds, which is well within manual rates. 

In other words, the current DecayRate is too "stiff" to account for fast bursts. Increasing it 20x to 2e-5 would change that result so that 10 infils in 30 seconds is equivalent to a steady state of 1 per 6.9 seconds, which is far into the auto infil range. That's a half-life of 13.8 seconds, meaning the recovery that happens between when the initial timestamp was recorded and the end of manual infil will be carrying a lot of the load. 

The other constant will also need to be adjusted, I think by a square-root factor to exactly match.

d0sboots:
> I think that the rewards should drop very sharply for infiltrations that are clearly faster than manual ( below ~1 second per floor + countdown times) but not affect manual speed infiltrations very much.

I think we are all agreed on this goal. It is down to hammering out the right metrics (I was initially looking too much at steady-state and ignoring bursts, for instance) and properly tuning.

src/Infiltration/formulas/game.ts
ficocelliguy: To confirm, you're recommending changing the DecayRate from -1e-6 to 2e5, and taking the square root of the market demand multiplier factor 1.5e-5 to instead be 3e-3?

d0sboots: No, changing decay rate by x20 to -2e-5 and changing market demand multiplier by the square root of x20 which is ~4. Edit, oh I might have gotten that backwards, it might actually by a square factor. So multiply 1.5e-5 by 20^2=400 to get 6e-3, which is pretty close to what you were going to try anyway. The market demand multiplier might require more fiddling though, I haven't run all the numbers on it.

catloversg:
I playtested the latest commit (c8d6c556294cfa3699445260b13e9602d849dfe3) a bit.

The short-burst of auto infil (2.5min-10m) is still too good.

Save file: 
[rebalance_infil.gz](https://github.com/user-attachments/files/21205001/rebalance_infil.gz)

Manual (7 seconds/floor):

manual

Auto (1 second/floor):
auto

I tried to tune `DecayRate` and the parabola formula of `marketDemandMultiplier`, but it's hard to nerf the short-burst of auto infil while maintaining reasonable rewards for manual infil.

ficocelliguy:
Maybe we need two scalars, one that is strong but has a half-life of like 60 seconds, and the other that is much weaker but has a much slower half life (10-15 minutes)?

d0sboots:
I think the numbers are actually starting to come together. I'm not shooting for auto-infil to be "balanced" necessarily, just not quite as broken as it currently is. 

The one big thing that I think is missing is that we do need to switch from tracking infils to tracking floors. Those tables show that even with the existing numbers, short infils like noodles can't be done at theoretically manual rates without eventually getting market demand 0. Meanwhile, places like ECorp which are incredibly tedious to do manually remain extra-good for auto infil because they only count for one each.

The simple change is to add floors instead of 1 (changing the state name as well), and adjust the DecayRate to match. (Probably drop it by 5x for starters since all infils have at least 5 floors.) This should bring the behavior for long infils in line with the short ones.

ficocelliguy:
I've updated the formulae to use floors instead of infil completions, as suggested. I had to reduce the market rate impact by a lot more than just 5x - since the market rate is impacted by the square of the floor count, the infiltrations with 35+ floors were enough to put the reward to 0%, otherwise! 

It is currently `const marketDemandMultiplier = 1 - 5e-5 * floors * floors;` but can be adjusted further as needed.

Scaling with floor count also has the added benefit of scaling faster with more infils than the old system, to better address bursts of superfast infiltration completions.

src/Go/effects/effect.ts
```diff
@@ -133,3 +137,23 @@ export function getDifficultyMultiplier(komi: number, boardSize: number) {
   const isTinyBoardVsIlluminati = boardSize === 5 && komi === opponentDetails[GoOpponent.Illuminati].komi;
   return isTinyBoardVsIlluminati ? 8 : (komi + 0.5) * 0.25;
 }
+
+export function gainFavor(opponent: GoOpponent, statusToUpdate: OpponentStats): void {
```
d0sboots: This snuck in from a different branch?

ficocelliguy: Ah, that was some scratch testing from a report where gang factions weren't getting favor reward (though they do, it turns out.) Removed.

catloversg:
The latest change (infils -> floors) looks pretty good.

Manual:

manual

With manual infil, the reward of low-floor locations is not capped too early like before.

Auto:

auto

The short-burst of auto-infil is balanced fairly well now. 3-5min reward of auto-infil is ~15min reward of manual infil.

Regarding SoA rep, with the latest formulas of market demand, I decide to apply a new balance multiplier to SoA rep. Previously, I want to let the market demand multiplier mitigate the flaw of SoA rep formula mentioned in #250. However, with the current formula of market demand multiplier, low-floor locations are still much better than high-floor locations (Rho is always the best location). With 9dc918d1915411cf96d570b4552cc4a64555951a, I add a new balance multiplier that nerfs low-floor locations and buffs high-floor locations.

"old-r": current formula (without the new multiplier)
"reward": new formula (with the new multiplier)

SoARep

Rho is still the best location among easy/medium locations, but the reward is nerfed ~20%. If the player can infiltrate hard locations (e.g., Solaris -> ECorp), they can get better rewards.

Save file:
[rebalance_infil.gz](https://github.com/user-attachments/files/21237362/rebalance_infil.gz)

src/Infiltration/formulas/game.ts
```diff
@@ -1,11 +1,43 @@
...
+
+// Calculates the infiltration reward multiplier based on how many and how recent other infiltrations were completed.
+// Each infiltration completed reduces the demand for corporate espionage data for a little while, thus affecting the
+// market demand.
+export function calculateMarketDemandMultiplier(timestamp: number): number {
+  const floors = calculateCurrentInfilFloors(timestamp);
+  // A parabola is chosen because it is easy to analyze and tune. The constant
+  // is a tuning factor, which primarily adjusts what the optimum rate of
+  // auto-infil is, and thus how good auto-infil is. The optimum
+  // marketDemandMultiplier will be 2/3 regardless of this constant.
```
catloversg, line 30: @d0sboots Can you please double-check this comment? We have updated this formula many times, so I'm not sure if the comment still reflects what we want, especially the `2/3` ratio.

d0sboots: This comment remains correct for any values we choose, but it may be badly worded. 

Specifically, it is talking about the market demand that an auto-infil script should optimize for, running in steady state. The best value is always 2/3. (There are actually some lower-order terms that become relevant in our situation, where the market demand is applied only occasionally in spikes, but that's the continuous solution.)

src/Infiltration/formulas/game.ts
```diff
@@ -1,11 +1,43 @@
 import { Player } from "@player";
-import { calculateSkill } from "../../PersonObjects/formulas/skill";
+import { clampNumber } from "../../utils/helpers/clampNumber";
+
+export const MaxDifficultyForInfiltration = 3.5;
+// This value is typically denoted "lambda," and is the instantaneous rate of decay.
+// I.e. infils decay towards 0 at 0.1%/sec. This is a half-life of ~11.5 minutes.
+const DecayRate = -1e-6;
```
d0sboots: The DecayRate has remained at -1e-6 instead of -2e-5 like I suggested earlier. If this is what CatLover has been testing, I'm very surprised that the numbers have been remotely as reasonable as they have been - it ought to be the case that they're still super unreactive to fast infils.

ficocelliguy: This value just controls how quickly the infiltration floors stat recovers, as I understand it. The market rate hit per floor is the value we have been tweaking, which drops quickly with many infill floors registered.

d0sboots:
OK, I did what I should have done a while ago and made desmos models for this.
Steady state infils: https://www.desmos.com/calculator/6csaoqpsgx
Burst infils: https://www.desmos.com/calculator/zufmivnt9t

The burst infils one is particularly useful because it can also show the steady-state behavior, once you get up to the "burst" regime of 10 infils in a row. I've saved them both with my (now) recommended values, which is DecayRate = (-)2e-5 and MarketDemand = 1e-3. This leads to the optimal steady-state behavior being in the range of .4 to .6 floors/sec (depending on floors in what you're targeting), which is auto-infil speed but definitely less than *max* auto-infil speed. If we assume a human can do .14 infils/sec, then this optimum is also only 2-3.3x more than the human rate.

On the burst side, bursting faster than 1/sec gets you very little. At 5 floors you can get ~6x the total rewards of a single infil, while at 37 floors you top out at ~1.5x. At the human rate, the market demand basically remains at 1. It's actually possible to restrict it even further than this, but I feel that auto-infil should still be worth *something*.

src/Infiltration/formulas/game.ts
d0sboots: Yes and no. Yes, in an absolute sense that is what it does. But in a wider sense, it is controlling the balancing between short-term and long-term behavior.

Consider: In the extreme, this could be set to 0 (or a very very small number). That would *still* cause floors to rise just as quickly as you complete infiltrations. However, without recovery, the long-term behavior would be that the stored value would continue to rise and (for steady state infils) the market demand would eventually hit 0, no matter how you tuned it.

Conversely, when the decay rate is large, older events matter very little. In the extreme, all that matters is the last infiltration.

We can tune the market demand multiplier independently so it works for bursts or steady-state conditions, but the decay rate needs to be set properly if you want it to work for both at the same time. Playing with the desmos links I sent will help illustrate this.

ficocelliguy:
I've updated the values to be the latest recommendation,

Note that this means that a single infiltration with high floor counts that is completed near-instantly will drop the rewards to 0 for a time. Here is NWO (with the rewards shown un-clamped for clarity), right after a ~10 floors a second completion of its 50-floor infil:

image

This can also be achieved with lots of repeated fast infiltrations with higher floor counts.

I think this is probably fine overall, since manual infil players will never see anything like that, but I'd like to update the UI somehow to indicate that recovery is being made (rather than just showing the clamped 0.000% for a full minute or however long it takes to recover). 

I'm thinking that, if the reward is less than some small amount (5% ?), instead it could be switched to a formula that asymptotes at 0. it won't affect the actual rewards given hardly at all, but will show some recovery continuously.

Alternatively, the market rate could be unclamped and shown as negative, and clamp the rewards to avoid charging players instead? but a negative market rate % might be confusing

catloversg:
Save file:
[rebalance_infil.gz](https://github.com/user-attachments/files/21261498/rebalance_infil.gz)

Please run them for raw data. I'll only show a limited number of screenshots in this comment. You can uncomment "speed" that you don't care about in the main function.

I usually use "seconds/floor", and d0sboots uses both "seconds/floor" and "floor/s", so this is the conversion table:

7 seconds/floor = 0.143 floor/s
5 seconds/floor = 0.2 floor/s
3 seconds/floor = 0.333 floor/s
2.5 seconds/floor = 0.4 floor/s
2 seconds/floor = 0.5 floor/s
1.7 seconds/floor = 0.6 floor/s
1.5 seconds/floor = 0.667 floor/s
1.25 seconds/floor = 0.8 floor/s
1 second/floor = 1 floor/s

Note:
- 7 seconds/floor: Manual infil
- 5 seconds/floor: Extremely fast manual infil. Anything faster than this is very likely auto-infil.
- 2.5 seconds/floor and 1.7 seconds/floor: The range in which d0sboots balances the optimal reward (0.4 floor/s - 0.6 floor/s).
- 1 second/floor: Extremely fast auto infil without bypassing countdown and other tricks (900ms countdown).

I don't think I can agree with the latest change. It poses a nice challenge that d0sboots mentioned a while ago:

> This last point is important - it makes auto-infil a bit more interesting because it means just blasting away as fast as possible will likely produce zero return. Instead, there is a new wrinkle which is to actually figure out (and run at) the best speed. Manual infil remains (almost completely) unaffected.

When the player uses an auto-infil script, they need to tune it and find the optimal speed (from 2.5 seconds/floor to 1.7 seconds/floor) instead of blasting away as fast as possible. This is a very nice characteristic of the latest change, but it comes with 2 downsides.

Downside 1: One of earliest goals of the market demand multiplier is to disincentivize long grinds of infil (Quote from the description of this PR: `This also has the effect of making manual infiltration better for short sessions or occasional runs, rather than incentivizing marathon grinds.`). For example, doubling farming time should give less than double reward. If farming for 30 minutes gives 25k rep, farming for 60 minutes should give less than 50k rep. This has been true in most iterations (I call each change an "iteration"). In the last iteration (infils -> floors):

Manual:


This looks good. However, currently, it looks like this:

Manual:
manual

The "nerf-long-grinds" behavior is gone.

You can only see the "nerf-long-grinds" behavior when the speed drops to ~1.5 second/floor (low-floor locations).

1.5 second/floor:


Downside 2: The optimal reward is balanced in the very low "seconds/floor" (low seconds/floor = high floor/s = high speed). To be clear, I agree that auto-infil should give a decent reward. The problem is where we should draw the line. This is where I and d0sboots disagree with each other.

1.7 seconds/floor (0.6 floor/s) is approximately the optimal speed for non-trivial locations.



Let's compare it with manual infil:

manual

I think the reward is too high, but d0sboots thinks it is okay.

Between 2 downsides, I have a strong opinion about downside 1. I don't want to give up the goal of disincentivizing long grinds of infi. If the reward of auto infil is too high (downside 2), the "nerf-long-grinds" behavior will still keep the high reward in check (i.e., the high reward is only good for short-burst auto infils). The line of "high reward" can be negotiated (I think it's too high, but I know it's just my opinion, and I may be wrong), but I think we should not ignore downside 1.

Regarding the last comment from @ficocelliguy, I think we should add a warning when the market demand goes too low (e.g., 1-5%). If it does, we will show a red warning and require another step from the player before they can start (e.g., after pressing "Play", they have to confirm another popup). Changing the formula like fico said may be okay too, but I'm afraid that it's too complicated for little benefit. If the player ignores all of our warnings, that's on them.

d0sboots:
> One of earliest goals of the market demand multiplier is to disincentivize long grinds of infil (Quote from the description of this PR: `This also has the effect of making manual infiltration better for short sessions or occasional runs, rather than incentivizing marathon grinds.`). For example, doubling farming time should give less than double reward. If farming for 30 minutes gives 25k rep, farming for 60 minutes should give less than 50k rep.

This is a central point of contention that I was unaware of. I never viewed that as a goal, but rather an (unwanted) side-effect of the previous numbers. I don't think manual infil should be disincentivized at all; I think the sheer tedium of it is enough of a disincentive, but if you enjoy that type of mindless repetition, have at it. I think playing to get all the SoA augs (once) is a reasonable goal, which will have people doing back-to-back infils, and I don't think market demand should punish them in that case. 

As to the other numbers, we can further tune and compromise on them. My main feeling is that even what I'm proposing is a significant nerf. Since this is a single player game, we don't need auto-infil to be "balanced", and indeed as an exploit it's somewhat better if it's not. What we're looking for is to avoid it being completely game-breaking, leaving those who do it feeling empty and dissatisfied after. Ideally it would give all the rush of cheating without the empty low afterward. :) (Time compression is in a similar bucket.)

ficocelliguy:
My vote is for the DecayRate to be closer to the original you had, d0s, (closer to -1e-6). There's a definite feelsbad we can help reduce when the player feels that the optimal strategy is to do something really tedious. It's something I experienced actutely early on, and others have expressed similar things (linked in the original post, and there are more I found that I didn't bother linking). Encouraging sessions rather than marathons is healthy for the game and the player.

Also, do you have any problem with the ideas (or other suggestions for) handling the UI when the rewards would be 0 or below, to make it more intuitive to the scripter what is happening?

ficocelliguy:
I've pushed up a compromise value for the DecayRate  (-4e-6) for testing. @catloversg can you run your test again against the latest commit, so we can compare it?

d0sboots:
> My vote is for the DecayRate to be closer to the original you had, d0s, (closer to -1e-6). There's a definite feelsbad we can help reduce when the player feels that the optimal strategy is to do something really tedious. It's something I experienced actutely early on, and others have expressed similar things (linked in the original post, and there are more I found that I didn't bother linking). Encouraging sessions rather than marathons is healthy for the game and the player.

I'm going to say a short bit about this, although it's been partially obsoleted by further thinking below.

To my mind, this is the same sort of attitude as people who say "Corp should be nerfed" (because they don't like Corp), and should be rejected for the same reasons. Specifically, it is the feelsbad of a mechanic that is powerful, but they don't enjoy. The thing is, BitBurner is centrally built around the philosophy that there *will* be such mechanics; that not every mechanic will appeal to everyone (and thus not to you), and many of them will be powerful. You are expected to pick and choose your own constraints for playing, and your own path through the game. If it doesn't include infil (or corp), you don't get to demand that those things are weakened for the players who enjoy them.

That said, I have had a vision of a (relatively) straightforward way to make a legitimate auto-infil API, which (IMO) is the piece everyone has been seeking for a long time. Infil has *always* been in a weird place, because it is the only "manual-only" component of a game that is all about automation. This would not be for this PR obviously, it is a thought for the future. In brief:
* `infil.pressKey(key)` returns a promise and creates a synthetic keyevent after a short delay. The event *only* works in infil. The delay has a small random component, so that batching it is not totally possible (but still a little possible).
* `infil.pressSpace()` synchronously presses the spacebar. Needed for timing-critical games like the slash game. Also provides some interesting diversity, since space is used in most games and the difference with `pressKey` provides variety and challenge. (You can still use `pressKey` with " " if you want.)
* `infil.getState` returns the state of the current game as an object. This object would be polymorphic, changing hugely depending on the game. Only one or two fields would be constant and documented, with `name` being one of them (the name of the current game). Polling this API to watch for state changes in timed games like slash or mines would be critical. Also important would be having a script to sniff the objects as you play manually, to understand what the object format looks like. Doing "detective work" to understand the various states would be part of the challenge (or doing equivalent sleuthing in the source code).

I think that would make a simple-to-implement API that at the same time would be an interesting and (more importantly) unique challenge to code for.

So, if we assume that that (or something like it) is somewhere in the future, it implies that we should be trying to actually (semi-)balance infiltration, as opposed to just making sure it's not broken. Because, it's better to make one big nerf than two medium nerfs.

An implication that falls out of this is that the optimum steady-state infil rate should likely be tuned to be close to manual speed. If it's much slower, it means even manual cannot play in steady-state. If it's much faster, it becomes hard to balance the rewards gap between manual and auto.

This also means the burst infils should not be "overpowered" for the beginning of a reset, even for manual play.

d0sboots:
> I've pushed up a compromise value for the DecayRate (-4e-6) for testing. @catloversg can you run your test again against the latest commit, so we can compare it?

I think this accomplishes your goals. Looking at the desmos graphs, that has a steady-state optimum @5 floors of .0796 (12.5 floors/sec), and @37 floors of .1186 (8.43 floors/sec). The burst behavior is largely the same as before, with any bursts faster than 5 floors/sec being heavily clipped, at 37 floors (the extreme case) to less than 2x nominal rewards.

We'll have to see the tables to see if it accomplishes all the goals (specifically, does it seem like a reasonable amount of rep for early bursts, or do other formulas need tweaking)?

ficocelliguy:
A partial API (or even just removing the synthetic event protection) was something I thought could be cool in infil for a long time, but I always thought it would be super difficult to balance it enough to get it there. Your idea sounds awesome.

catloversg:
`-4e-6` is too extreme. It accomplishes the goal of "nerf-long-grinds" that I want, but it nerfs auto infil too much.

7 seconds/floor:

current-7

5 seconds/floor:

current-5

2.5 seconds/floor:

current-2 5

In the last 2 iterations, we tested 2 extreme cases:
- `DecayRate = -2e-5`: The optimal speed is tuned very high (0.398-0.593 floor/sec; 2.5-1.7 sec/floor).
  - No "nerf-long-grinds" effect on manual speed.
  - Decent reward for fast auto-infil (I think it's too high).
- `DecayRate = -4e-6`: The optimal speed is tuned very low (0.0796-0.1186 floor/sec; 12.5-8.4 sec/floor).
  - Have "nerf-long-grinds" effect on manual speed.
  - Low reward for fast auto-infil (I think it's too low).

I have spent a few hours fiddling with the Desmos links and my simulator. Unfortunately, I come to a conclusion that there is no good way to achieve the "nerf-long-grinds" effect in all speed while maintaining a decent reward for (fast) auto infil.

I think this is a stalemate. My solution:
- Give up on the "nerf-long-grinds" effect on low speed. I don't like to give up on this goal, but I also don't like to drag out this PR forever.
- Tune the optimal speed at the middle of those two extreme cases.

I come up with `DecayRate = -9.65e-6`. The optimal speed is tuned at 0.192-0.286 floor/sec (5.2-3.5 sec/floor). "5 sec/floor" is the top speed of manual infil. "3.5 sec/floor" is a relatively good speed of auto infil (it is double the speed of manual infil).

7 seconds/floor:

suggestion-7

5 seconds/floor:

suggestion-5

3.5 seconds/floor:

suggestion-3 5

2.5 seconds/floor:

suggestion-2 5

2 seconds/floor:

suggestion-2

1 second/floor:

suggestion-1

Regarding the proposed APIs, I think it's okay. 2 things to consider:
- Where should we tune the optimal speed? I think it should be at ~7-5 sec/floor (normal-fastest manual infil).
- The best thing (IMO) about auto infil right now is the challenges of scraping the DOM data and bypassing the keyboard event check. Those APIs trivialize these challenges a lot, especially the need to bypass the keyboard event check. Generally speaking, hard challenge = high reward. With these APIs, we have to rebalance the reward again, and I think it will be another nerf (this is why I suggested tuning at ~7-5 sec/floor). My question: Should we consider a decent reward for people who don't use these APIs and just do it the hard way (scrape the DOM + bypass event check)? I guess not, but I still want to hear more opinions.

d0sboots:
> I have spent a few hours fiddling with the Desmos links and my simulator. Unfortunately, I come to a conclusion that there is no good way to achieve the "nerf-long-grinds" effect in all speed while maintaining a decent reward for (fast) auto infil.

I thought you had already realized that there is no way to square this circle. Not just with this implementation, but in *any* implementation, nerfing long grinds implies that the optimal long-term speed is slightly below manual infil levels. Combine this with the need to protect against bursts, and auto infil will always be severely nerfed, basically only doing slightly better than manual. The value becomes its ability to do so without user oversight.

> I come up with DecayRate = -9.65e-6. The optimal speed is tuned at 0.192-0.286 floor/sec (5.2-3.5 sec/floor).

Can we just say -1e-5 for round numbers? If more than one significant figure really matters, we're probably lying to ourselves about how accurate our tuning is. 

> The best thing (IMO) about auto infil right now is the challenges of scraping the DOM data and bypassing the keyboard event check. Those APIs trivialize these challenges a lot, especially the need to bypass the keyboard event check.

The best thing about auto infil is that it *has* a challenge. Scraping the DOM and bypassing the keyboard check are certainly part of it, and for some probably the most interesting part. (Those people should also be pointed at SF -1). But there is also actually wiring that up together so that it solves each game, which IMO is much more work than figuring out how to do some scraping and bypassing, and that challenge remains. The proposed API intendeds to replace the eliminated challenges with some new ones; bypassing the event check is replaced with dealing with a partially random event delay. Scraping the DOM is replaced by figuring out object internals. It won't be exactly the same, but hopefully it still feels fun.

> Generally speaking, hard challenge = high reward. With these APIs, we have to rebalance the reward again, and I think it will be another nerf (this is why I suggested tuning at ~7-5 sec/floor). 

Yes, that is why I would prefer to do that rebalance now, rather than have a second nerf down the road. IMO we should be looking at the rep numbers as if auto infil were a sanctioned part of the game; how broken would they be? Is 150 rep/s fine, if it's gated behind something like sing?

> My question: Should we consider a decent reward for people who don't use these APIs and just do it the hard way (scrape the DOM + bypass event check)? I guess not, but I still want to hear more opinions.

I don't think that's actually "the hard way." Particularly with the nature of the input api, IMO it would be much *easier* (certainly faster and take less RAM) to figure out how to fake inputs once, and then use that everywhere. This is similar to how people will figure out how to inject commands into the terminal before they have sing. I don't see a need to reward that behavior; the exploit itself is the reward.

d0sboots:
> Regarding the last comment from @ficocelliguy, I think we should add a warning when the market demand goes too low (e.g., 1-5%). If it does, we will show a red warning and require another step from the player before they can start (e.g., after pressing "Play", they have to confirm another popup). Changing the formula like fico said may be okay too, but I'm afraid that it's too complicated for little benefit. If the player ignores all of our warnings, that's on them.

Oh right, I forgot to reply to this part of the discussion. I think what would work is to show `0.00% (-whatever%)`, i.e. show both the clamped value and the unclamped value in parenthesis. This would work a lot better if the screen that shows these values has a refresh hook, so you can see the values change in realtime at 5 FPS or whatever; when the values are negative in particular they will be changing pretty rapidly.

src/Infiltration/formulas/game.ts
```diff
@@ -29,6 +29,13 @@ export function calculateMarketDemandMultiplier(timestamp: number): number {
   // auto-infil is, and thus how good auto-infil is. The optimum
   // marketDemandMultiplier will be 2/3 regardless of this constant.
   const marketDemandMultiplier = 1 - MarketDemandFactor * floors * floors;
+
+  // Asymptote to 0 at very low market demand, so the UI can always show the demand rising over time
+  // Without this, the demand can go far below 0%, and thus not visibly move for some time.
+  if (marketDemandMultiplier < 0.03) {
```
d0sboots: The fallback that you've implemented here has some big issues: <https://www.desmos.com/calculator/b1inlvdxsw>
Specifically, with the current numbers this behavior kicks in at a stored floors of 31.145, but at that point this formula returns .146, i.e. it will be clamped down to 0.03. This clamping behavior continues out to stored floors of 56.7, so for most of the range where you want it to be effective, the value won't actually be changing because it will still be clamped to a single value, it'll just be 0.03 instead of 0.00.

There's larger issues though. Even if you fiddle with this to eliminate that, or make it a smooth asymptote instead of one that cuts in suddenly, there is the problem that making the market demand non-zero at higher values re-introduces the issue of fast auto-infil potentially becoming optimal. This is because market demand is not just a display value, it controls the real behavior, and that's why this is not the right place to be making changes that you want for strictly display purposes.

src/Infiltration/ui/Intro.tsx
```diff
@@ -42,46 +42,81 @@ function arrowPart(color: string, length: number): JSX.Element {
...
-  const soaRepGain = calculateInfiltratorsRepReward(Factions[FactionName.ShadowsOfAnarchy], props.StartingDifficulty);
+export function Intro({
+  location,
+  startingSecurityLevel,
+  difficulty,
+  maxLevel,
+  start,
+  cancel,
+}: IProps): React.ReactElement {
+  useCycleRerender();
```
d0sboots: Ah, I missed it but you already added this for 5FPS updates. Great.

src/Infiltration/formulas/game.ts
ficocelliguy: This is a good point. I've removed the asymptotic reward at low percents, and instead let the user see the unclamped market rate in the UI ( though it is clamped for all calcs to not go below 0)

src/Infiltration/ui/Intro.tsx
ficocelliguy: I've tweaked this to only update once per second: the rapid changes before made it kinda hard to read numbers with lots of shifting digits, and this gets the same effect across much more cleanly.

catloversg:
> Can we just say -1e-5 for round numbers? If more than one significant figure really matters, we're probably lying to ourselves about how accurate our tuning is.

Using `-1e-5` is okay too.

> Yes, that is why I would prefer to do that rebalance now, rather than have a second nerf down the road. IMO we should be looking at the rep numbers as if auto infil were a sanctioned part of the game; how broken would they be? Is 150 rep/s fine, if it's gated behind something like sing?

I have a different opinion about when we should make the nerf. I prefer having 2 small nerfs to 1 big nerf. Currently, auto infil is still in the exploit-y territory (DOM access + prototype tampering), so I'm fine with it having a decent reward (slightly unbalanced reward) after the first nerf. 150-200 rep/s is okay. However, when we provide NS APIs, auto infil becomes a "sanctioned" gameplay, and 150-200 rep/s is not okay. I prefer it to be 2 "steps":
- Exploit-y territory: This is what happens right now. We tune at slightly fast auto infil (5-3.5 sec/floor; we are using this speed in the latest commit). High reward (150-200 rep/s) is okay. I don't want to nerf it more than it is now.
- NS APIs: This is what we will do (not sure when). We tune at normal manual infil (~10-7 sec/floor; a bit faster than 12.5-8.4 sec/floor that fico uses in the previous commit). Good reward (85-100 rep/s) is okay.

Baseline: Normal rep gain rate with all rep augs:
- At hack 2500 and 50 NFGs, rep/s is ~50/s (49.680/s).
- At hack 2500, 75 NFGs, ns.share (1PB), rep/s is less than 100/s (95.520/s).

catloversg:
When the unclamped values is negative, I show it like this:

Capture

catloversg:
By the way, I just think that the proposed APIs may be a good reward for SF11. SF11 currently is too weak. The buff of company salary and rep is almost useless. The reduction of aug price is mediocre.

d0sboots:
I think the code and UI are looking great. We're down to just the numbers. 

> I have a different opinion about when we should make the nerf. I prefer having 2 small nerfs to 1 big nerf. Currently, auto infil is still in the exploit-y territory (DOM access + prototype tampering), so I'm fine with it having a decent reward (slightly unbalanced reward) after the first nerf. 150-200 rep/s is okay. However, when we provide NS APIs, auto infil becomes a "sanctioned" gameplay, and 150-200 rep/s is not okay.

I agree with you here from a "logic that supports the changes" standpoint. However, what this misses is the psychology of making changes. People (and gamers especially XD) don't like it when you take their stuff; this is why nerfs are so hated. As weird monkeys, we have psychological effects like "loss aversion" which mean that if you buff something by 50%, and then realize you went too far and nerf it by 10%, that will be *hugely* unpopular, while a flat 40% buff will be loved - even though they're the same thing.

What this means when it comes to big nerfs is that they're hated more than small nerfs, of course. But the effect is not linear: a 300% nerf is not hated 1.5x as much as a 200% nerf. So, if you have to make big nerfs, it is much better to do them in one go.

What this possibly means is that we should hold off on the significant teeth now, and only tune this to deal with really aggressive (faster than 1 second) auto-infil, no big changes for "regular" auto infil. Auto infil would remain "broken", but without anything to replace it yet there's little reason to nerf the players doing it, aside from the very most extreme. But all the infrastructure and UI would be in place to crank it up more, later.

(Yes, I realize I'm being waffle-y here. Over the course of writing this out, I realized that although I still think one nerf is best, I don't have justification for front loading it, and indeed I think manual infil enjoyers would be rightfully upset by that.)

I could be convinced otherwise, but the case needs to be made that nerfing *now* makes the game more fun, and in a form that doesn't also apply to Corp (because I reject that logic for Corp).

> By the way, I just think that the proposed APIs may be a good reward for SF11. SF11 currently is too weak. The buff of company salary and rep is almost useless. The reduction of aug price is mediocre.

Excellent idea! And it fits with the theme.

catloversg:
I reverted the decay rate back to your tuned speed (2.5-1.7 sec/floor). Fast auto infil will not be nerfed until it hits ~1 sec/floor. We already made our points clear, so I don't want to hold off this PR further just because of a tuning parameter.

When we make those APIs, we will have to tune those numbers again. If we need to make a big nerf (due to auto infil becoming a sanctioned gameplay), let's do that there. I don't mind auto infil still being a bit too strong when it's still in exploity territorry *and* we have a clear road to change it (via those APIs).

PR #2224, opened 2025-07-05
CODEBASE: Update build script to avoid unnecessary newline in generated script

catloversg:
When the image folder is empty, there is an unnecessary newline.

Before:
before

After:
after


d0sboots:
For the future: There is a simpler solution for situations like this, you can add the newline as part of the line itself instead of as a join.

catloversg:
Got it. I'll update it in another PR, maybe when I add images (the sequence diagram that I talked about) to the image folder and update the Remote API page.

PR #1525, opened 2024-07-27
CONTRACT: Remove "&nbsp;"s from .getDescription() result

gmcew:
closes #1514 

Using `run contract-XXXX.cct` in the terminal displays fine, a small tweak to `ns.codingcontract.getDescription()` removes the `&nbsp;`  from the function return string too and cleans up the script logging issue. There's probably a more elegant way to do this, sorry.

screenshots of the terminal print of `ns.codingcontract.getDescription()` and  `run contract-XXXX.cct` results post-change attached, in that order.

getDescription
run

PR #2225, opened 2025-07-05
CONTRACTS: Fix nbsp rendering as text rather than a space

gmcew:
#2171 changed how contract descriptions are rendered when accessed from the terminal by removing use of `dangerouslySetInnerHTML`.

This results in the non-breaking space rendering as '`&nbsp;`' rather than '&nbsp;'. A simple solution is to revert the removal of `dangerouslySetInnerHTML` - if there's a better option I'll update this PR.

Current:
Image

With use of `dangerouslySetInnerHTML`
Image


d0sboots:
Can we not just use the nbsp Unicode character directly?

gmcew:
> Can we not just use the nbsp Unicode character directly?

That looks like it works and I don't know why I didn't think of that; thank you. @catloversg thanks for sorting the workaround and your reviews.

Let me know if i missed anything else here.

EDIT: In-game view with \u0A00 for reference
u00a0
EDIT2: Triangle as run
runTriangle

EDIT2: Triangle as returned in the various forms of `getDescription`
getDescription
getContractdescription

catloversg:
I think we should consider using a normal space character with `white-space` CSS.
```jsx
      <span key={i} style={{ whiteSpace: "pre-wrap" }}>
        {value} <br />
      </span>,
```
```js
      return [
        "Run-length encoding (RLE) is a data compression technique which encodes data as a series of runs of",
        "a repeated single character. Runs are encoded as a length, followed by the character itself. Lengths",
        "are encoded as a single ASCII digit; runs of 10 characters or more are encoded by splitting them",
        "into multiple runs.\n\n",
        "You are given the following input string:\n",
        `    ${plaintext}\n`,
        "Encode it using run-length encoding with the minimum possible output length.\n\n",
        "Examples:\n\n",
        "    aaaaabccc            ->  5a1b3c\n",
        "    aAaAaA               ->  1a1A1a1A1a1A\n",
        "    111112333            ->  511233\n",
        "    zzzzzzzzzzzzzzzzzzz  ->  9z9z1z  (or 9z8z2z, etc.)",
      ].join(" ");
```

Edit: You should also check `MinimumPathSumInATriangle` in `src\CodingContract\contracts\MinimumPathSumInATriangle.ts` and `getDescription` in `src\CodingContract\Contract.ts`. `"&nbsp;"` is used programmatically there.

gmcew:
To clarify the edit queries first I'll put the tested screenshots in my previous comment for reference; the `MinimumPathSumTriangle` use did cause concern at the time, I did verify it still worked. The other contract of note for using `&nbsp` in more than just the description padding was `SpiralizeMatrix`; this also renders as expected, though I haven't included screenshots for that.
The `getDescription` `.replace` usage was a quick and dirty workaround about a year ago in #1525 , with hindsight it was treating the same issue that we're dealing with here - it renders fine without it, and should now be superfluous so I've removed it as part of the below change.

I can definitely see the value of not spamming unicode characters for code legibility; I've modified it as your suggestion and the results look good.
Related to your point on how the `getDescription` functions display: desciptions as returned from `getContract()` include the newline as text - but not the space - if you print the entire returned object. That happens with both the previous unicode spaces and normal space/CSS white-space approach and is a fairly niche issue.

I appreciate the help!

Triangle as run from terminal:
cssRun

Triangle `getDescription` print:
cssf1

Triangle `getContract().description` print;
cssf2

Triangle `getContract()` print
cssf3

catloversg (approved):

Issue #2223, opened 2025-07-05
Feature Request: Add custom help text for your own scripts


rins6618:
I wrote about creating another exported function in scripts recently:
https://discord.com/channels/415207508303544321/415207923506216971/1390823121098903673

I've implemented it already in my own local instance and I only need approval to turn this into a PR. What are the ideas you guys got?
I may add ReactNode support in the future. 

<img width="497" height="302" alt="Image" src="https://github.com/user-attachments/assets/e26fd6e4-1761-4aaa-9429-4b88ee978697" />
<img width="594" height="662" alt="Image" src="https://github.com/user-attachments/assets/3709c1f3-c463-40a8-bc5f-ac6df2f21ce7" />

<img width="1360" height="718" alt="Image" src="https://github.com/user-attachments/assets/0c4dc6bb-c61a-42c1-b8ad-81e748743529" />

gmcew:
Conceptually, it's a neat addition. If you're comfortable with how you have it working it's probably best to raise a PR and link it to this issue for review and feedback.

d0sboots:
+1, you don't need permission to make a PR, you can just do it. The approval part comes during the PR discussion. :D

PR #2227, opened 2025-07-06
COMMAND: Custom help commands for your own scripts

rins6618:
concept #2223 
### checks:
- [x] Ran `npm run doc`
- [x] Ran `npm run lint`
- [x] Ran `npm run format`
- [x] Passed 38/38 tests on `npm run test`

Had a delay in implementing this; turns out I forgot to rebase after waking up and only did it after I was done. Grimace.  

###
image
_Fig. 1: documentation_
###
image
_Fig 2: example use_
###
image
_Fig 3: all possible return types_


rins6618:
Sorry about that. rebase was being naughty and i am sleep deprived

d0sboots:
> Sorry about that. rebase was being naughty and i am sleep deprived

Once you've pushed to a PR, it's easier for us (the reviewers) if you merge new changes instead of rebasing. We'll squash the whole PR down to a single commit at the end regardless, and rebasing loses track of the commits as part of the comment/reply flow. (It doesn't matter here since I don't think anyone had taken a look yet.)

Edit: ofc you can do whatever makes your life easiest *before* you open the PR, and I frequently do rebase right before pushing as well. Or sometimes right after if I've forgotten XD

src/Documentation/doc/basic/autocomplete_help.md
```diff
@@ -0,0 +1,290 @@
...
+}
+
+/** @param {NS} ns */
+export function main(ns) {
+  const args = ns.args;
+  ns.tprint(args[0], args[1], args[2]);
+}
+```
+
+Running this script from the terminal like `run script.js` or `./script.js` and pressing tab, would offer `"argument0", "argument1" and "argument2" as autocomplete options.
```
d0sboots:
```suggestion
Running this script from the terminal like `run script.js` or `./script.js` and pressing tab, would offer "argument0", "argument1" and "argument2" as autocomplete options.
```

src/Documentation/doc/basic/autocomplete_help.md
```diff
@@ -0,0 +1,290 @@
...
+/**
+ * @param {AutocompleteData} data - context about the game, may be useful to list argument documentation
+ * @returns {string|string[]|ReactNode|HelpData}
+ */
+export function help() {
+  return `${"\x1b[2m"}This is fancy bold text!`;
+}
+/*
+Usage for script.js:
+        This is fancy bold text!
```
d0sboots: Did you intend for this to show in bold in the docs?

d0sboots (commented):
High-level comment: I think this feature overall is very welcome! However, I have some (slight) reservations about the HelpData system. Specifically, that constitutes most of the complexity in this PR, for a very specific way of formatting help. As we have seen with the args and autocomplete systems, oftentimes people want more control or customizability in how they handle things, and so bundling what amounts to libraries inside the game often is a mistake.

At the same time, since helpdata is not something code can access, it doesn't suffer from the lock-in problems of many systems - we can make breaking changes to its output without worry. And it *is* nice to have an easily accessible fancy help system available, although I suspect most people will be more than happy with the basic one.

On the whole, I am slightly leaning towards "let HelpData be a library you provide for people who want it, built on top of the base functionality here." That would also let you easily update and extend it on your own schedule. But, I'd also be happy to hear other opinions.

src/Documentation/doc/basic/autocomplete_help.md
rins6618: Yeah, I think the tab character is not necessary in the example and only complicates formatting. Will change.

rins6618:
I'm considering changing the return signature of the help() function recently. I thought of ReactNode but I think this use is so niche that it might benefit to be removed. It also complicated type safety in the code.

I'm now considering `help(data: AutocompleteArgs): HelpData | string | string[]`. Mainly due to how the game used to implement help text before this PR (vide `src/Terminal/HelpText.ts`).


I'm also considering `help` forwarding its input to a custom script instead.
I.E `help script.js` would run `./helpScript.js script.js` (if script.js implemented usage of helpScript.js), which moves responsibility away from the game to build custom documentation. But that might complicate flow a bit.

catloversg:
Before reviewing, let's finalize the expected features.
- Both d0sboots and I are concerned about the built-in HelpData. I think we should keep this feature as simple as possible now. We only "forward" the returned value from the "help" function to the terminal. Nothing more, nothing less. If there is a strong demand for a built-in "template", we can implement it later (after a discussion and voting on Discord about what should be included in the built-in option).
- Signature: `help(data: AutocompleteData): string | ReactElement`. `string`: Terminal.print(). `ReactElement`: Terminal.printRaw(). I don't want to support too many types. For all non-React cases, `string` is enough. If the player wants to use `string[]` with each string on a newline or something else, just tell them to use `join`. `ReactElement` is for React cases.

rins6618:
I'm fine with that. I'll look into implementing it as simple as possible. I thought a bit about it and while the thought of a template makes this really cool as someone who makes really detailed scripts -- I expect most players on the game would probably just write some really simple string as their own documentation.

d0sboots:
> * Signature: `help(data: AutocompleteData): string | ReactElement`. `string`: Terminal.print(). `ReactElement`: Terminal.printRaw(). I don't want to support too many types. For all non-React cases, `string` is enough. If the player wants to use `string[]` with each string on a newline or something else, just tell them to use `join`. `ReactElement` is for React cases.

+1
ReactElement is a niche usecase, but it's also very powerful and easy to add. String is the one we expect most people to be using. For the manpage style formatting, it will be easy for you to provide it (as a regular player) as an addon library built on top of this functionality (using the ANSI codes).

rins6618:
Okay, I'm gonna be free later on this week. Any final reviews on any fixes on this feature I could fix?

src/Terminal/commands/help.ts
```diff
@@ -1,20 +1,104 @@
...
   } else {
     const cmd = args[0] + "";
     const txt = HelpTexts[cmd];
+
     if (txt == null) {
-      Terminal.error("No help topics match '" + cmd + "'");
+      // Here is where flow lands if we have a player-implemented command
+
+      // Input sanitization
+      const cmdCopy = String(cmd).replace(/^[/.]+/, "") as ScriptFilePath;
```
d0sboots, line 29: Why this? Does help not work for scripts with relative directories?

This doesn't feel like the right way to be checking for a script. I think Terminal.getFilepath is what you ought to be using here.

src/Terminal/commands/help.ts
```diff
@@ -1,20 +1,104 @@
...
+      const script = localServer.scripts.get(filePath);
+
+      try {
+        if (script == null) {
+          throw new Error("Script pathname has no valid script object");
+        }
+
+        // Deal with help() here
+
+        // eslint-disable-next-line @typescript-eslint/no-floating-promises
```
d0sboots, line 56: This warning is legitimate, although ultimately you will still end up doing the equivalent of dangling a promise.

But, the issue is that everything within here is asynchronous, whereas the help command ends immediately. So, multiple lines can end up appended to the terminal before the script comes back with the result of the help.

I think what you're going to need to do is `Terminal.printRaw` a reactnode here synchronously, but then capture that node inside the promise via closure capturing. That way, the space has already been allocated for the result, and you can asynchronously fill it in once it comes back. It could even start off as "..." or something to indicate it's computing. All the error messages that happen within the async part would work the same way.

I *think* you might be able to get fancy and still use Terminal.print/Terminal.error and thus avoid having to do ugly things for the ANSI support, by then directly mucking with the Terminal object and essentially swapping the placeholder object you created with printRaw for the new line, and removing the placeholder after. Since this all takes place synchronously (inside the resolution of the async `then`), it won't cause any flickering.

src/Terminal/commands/help.ts
```diff
@@ -1,20 +1,104 @@
...
+        compile(script, localServer.scripts).then((compiledModule) => {
+          if (compiledModule.help == null) {
+            Terminal.error("No help text for '" + cmdCopy + "'. Implement it by exporting a help() function.");
+            return;
+          }
+
+          const autocompleteData: AutocompleteData = {
```
d0sboots, line 63: Surely there's a version of this already for autocomplete? They should be shared, unless there's a painful reason why they can't.

d0sboots (commented):
Sorry, for some reason I thought there was more to do.

There's one major hurdle left I think, but the removal of HelpData has massively simplified things.

src/Documentation/doc/basic/autocomplete_help.md
```diff
@@ -90,8 +96,103 @@ export function autocomplete(data, args) {
...
+
+
+The function can also return a React element. For this, it's highly recommended to write a `*.tsx` script file.
+The example above, written in `.tsx`:
+
+```tsx
+export function help(data: AutocompleteData): ReactNode {
+  return <li><ul>
+    <li> ./{data.command} {"<servers...>"} </li>
+    <li> Servers must be one of: {data.servers.join(", ")} </li>
```
catloversg, line 188: Same here.

src/Terminal/commands/help.ts
```diff
@@ -1,20 +1,104 @@
...
+            hostname: localServer.hostname,
+            servers: GetAllServers()
+              .filter((server) => server.backdoorInstalled || localServer.serversOnNetwork.includes(server.hostname))
```
catloversg, line 66: `servers` is the list of all non-hidden servers' hostnames. You have to use `filter((server) => server.serversOnNetwork.length !== 0)` (`src\Terminal\getTabCompletionPossibilities.ts`).

If there is a reason that `servers` has to be a different list in `help`, you have to explain that reason in the TSDoc of `AutocompleteData`. I don't think there is one, though.

src/Terminal/commands/help.ts
```diff
@@ -1,20 +1,104 @@
...
+              .map((server) => server.hostname),
+            scripts: [...localServer.scripts.keys()],
+            txts: [...localServer.textFiles.keys()],
+            enums: enums,
+            // Pass no flags as the help command does not use flags
+            flags: Flags([""]),
```
catloversg, line 72:
```suggestion
            flags: Flags([]),
```

src/Terminal/commands/help.ts
catloversg: While refactoring, as d0sboots said, you should be careful when making any differences (e.g., `servers`).

src/Documentation/doc/basic/autocomplete_help.md
```diff
@@ -90,8 +96,103 @@ export function autocomplete(data, args) {
...
+A few notable differences with this and autocomplete is that the `.command` property contains the file path and name, the `.flags` property is empty.
+
+_Example: contracts.ts_
+
+```ts
+export function help(data: AutocompleteData): string {
+  return [
+    "Finds and solves contracts in a server.",
+    "Possible arguments: ",
+    `  ${data.servers.join(", ")}`
```
catloversg, line 166: You need to update this example code and output. Please check my comment in `help.ts`. I recommend using something that is not `data.servers`. That list is too long after making the suggested change in that comment, so it's not really a good example.

PR #2139, opened 2025-05-18
DARKNET: Darkweb Expansion Project & Bitnode

ficocelliguy:
## New Features
- Adds the Dark Net: a shifting, unstable collection of randomly-generated password-protected servers. 
  - The network periodically has servers added, deleted, move, shift connections, or "restart"
  - The network is cyclic, and servers can have very many or very few connections
  - Some parts of the network are disconnected, and can only be reached by "riding" a server that moves between them

- Servers provide simple clues to their "vulnerabilities" 
  - Each server provides hints to their password such as "it's still the default" or "it's 23 in base 4"
  - Some servers provide interactive feedback in their server logs, accessible via API or visible on the UI
  - Harder servers have puzzles like wordle/mastermind or "yes, the password is divisible by your input"
  - Solutions for these (or at least simple approaches) can be worked out from first principles / intuition in a short time, rather than being most efficiently tackled via educational videos or deep research like many .ccts
  
- Adds a new bitnode centered on the dark net
  - IN BN15, The Red Pill is only available via a server deep in the net
  - In BN15 (or after getting SF-15.1) the dark net increases in depth (and puzzle & cha requirement) as the player progresses
  - Before unlocking the full-depth and full-featured net via BN15, there is a small preview version of the 'net in all bitnodes
  
- The Labyrinth in the deep
  - There are special servers at the bottom of the net that require some simple maze-solving, and award special augments
  - Completing each one, and installing, increases the depth & difficulty of the dark net, and increases the cha requirement of the next lab server
  - The fourth difficulty (requiring 2500 cha) awards The Red Pill for completing it. 
 
 - Laying on the 'rizz
   - The darknet is largely themed around (and requires) charisma, a particularly under-utilized stat currently
   - A number of new cha-boosting augments have been added to various factions (mostly non-hacking ones) 
   - A number of existing augments that thematically increase persuasion (or eyes or skin) now also increase cha

## Gameplay Loop

- Initial exploration is done via the UI
  - Early servers are easy to do manually via just experimenting by clicking on servers. ("We forgot a password!" Password length: 0)
  - Later servers take more guessing-and-checking ("The password when sorted is 145")

 - Harder servers encourage scripting
    - Servers can sometimes have valuable .cache files or data files, and manual access to open them gets slow quickly
   - Some servers have many password possibilities, encouraging a clever script to choose carefully, or to iterate through all possibilities
   - Some servers ask for math that's easier via script (evaluate an expression, or convert from roman numerals to base 10)
[...]

PR #2230, opened 2025-07-07
CONTRACTS: Proposal for Contract Progression System

### LINKED ISSUE #2018: Proposal: Contract progression system
The contract system has imho the following problems:
 - for fresh starters they are overwhelming:
    - a huge amount of different types have to be implement at once - this lead to copy pasta from the web (at least this is true for me then I played it the first time a few years ago - and to be honest, I kind of regret this and i am always happy, if new contract types appear) 
    - for new players it may be especially challenging as they are neither that familar with javascrip nor with algorithms
  - at start of a new Bitnode it is quite imbalanced if the first contract you get is a heavy one that grants you $750m that is immediatly enough to kickstart the stocktrading oder greatly boost hash generation (to get the next contract)

To get more of a soft ramp-up, I would  propose some kind of progression system that works on two levels:
 - global level (does not reset on bitnode)
   - the player globally has to solve _n_ contracts of a hardness level to enables generation of harder contracts
   - from the contract types that are similar themed, player should solve some number of the easier before the other get generated (may be this grouping should be independent of the actual contract-types)
 - bitnode level
   - most contracts can easily be scaled in difficulty
   - this difficulty could be start low in a new bitnode and increase with number of solved contracts of a tpye
   - reward should scale proportional with difficulty and (this would be cool, but I have no idea how this could be implemented) inverse proportional with the required time to compute the result

gmcew:
# Contract Progression Proposal

May close #2018 
For context, I have no programming experience outside of this game, but I enjoy the massive variety of contracts and want them to be as accessible as possible. This write up will be fairly long because I want to ensure I don't miss the mark - please be patient with me!

## Preamble:
A few points were touched on in #2018 that need mentioning before getting to the progression approach:
>“Or, tie the progression curve to the hack difficulty of the server it spawns on, so the easier ones appear closer to home, and the harder ones require more port openers or w/e.”

Contracts do not require root access on the server to be accessed or solved, hence can be completed before any port openers are available. I’ll refer to this as “bootstrapping”: solving a contract (usually at the start of a node) to get a quick cash boost with no additional expenditure.
This PR suggests no change to that behaviour.

>“I kind of agree with the problem of randomized reward, but some players do not mind RNG. Occasionally having a "lucky" cct is a good thing with them. A small tweak for the cap of rewards is fine, but it's fairly hard to do properly.”

The reward amount for a given contract type is fixed. The category of reward is randomised on generation, but each category has a fixed reward value as a function of contract difficulty. This also reinforces the “bootstrapping” concept of the previous point – $750m can *only* be gained by solving a 10 difficulty contract which rolled a money reward category, but likewise solving a 10 difficulty contract for money will *always* reward $750m independent of when or where it was solved.
This PR suggests no change to that behaviour.

>“It's almost impossible to define the difficulty of contracts. Easy contracts for this player may be hard contracts for other players, and vice versa.”

This is objectively true, but “difficulty” is still a value allocated to each contract type. The difficulty is an arbitrary value set originally by the contract creator(s), and within each contract type is arguably a wider range of difficulties than between different types of contracts. However, the difficulty stat offers some (highly subjective) benchmark of which contracts require more work to create reliable solvers compared to others. This ties in to the above point that reward scales directly with difficulty, and nothing else.
This PR will attempt (and probably fail) to not significantly change how rewarding contracts are from their current positions, as the difficulty/reward ratio is the core of balance. I will try to justify my choices with data where I can for discussion and review.

## Problem Statement:
My interpretation of the discussion in #2018 boils down to the following points:
1.	There are a lot of different contract types in the game - 29 at time of writing. This can be overwhelming to a new player – particularly if they don’t have any prior programming experience.
    - The range of contracts exposed to a new player should be limited in both variety and difficulty to encourage focus on solving a limited pool of “easier” ones for more consistent rewards and reduce how intimidating the harder ones can be.
2.	Contracts are a completely optional side quest. They can safely be ignored for the entire game as they offer no lore or critical progression, but do offer useful one-off rewards that are beneficial – “bootstrapping”.
    - Any progression structure should not penalise either of the following groups by implementing the previous bullet – late-game players who have already invested time in solving contracts, and keen/experienced new players who wish to progress faster. Investing time in writing solvers and interacting with contracts should be rewarding and satisfying.

There’s some additional requirements that I consider important:

3.	Contract Progression structure must allow for any number of new contract types being added to the game without conflicting with points 1 and 2 above.
4.	Contract Progression structure must consider metagame changes to how contracts spawn or are handled, e.g. #2139
5.	The player shouldn’t notice a difference in how contracts spawn and are handled whilst still satisfying point 1 – these changes should be as minimally invasive as possible.

## Proposed Progression Framework:

Conveniently, since the difficulty value for each contract already serves to scale reward (and only to scale the reward) my proposal is that all 5 requirements above can be satisfied by adding spawning rules for contracts that also use that existing difficulty value.
I’ve implemented this approach through the following changes: 
- Expose contract difficulty value. It’s useful information for the player, particularly when encountering a new type of contract, in terms of expect reward and effort required.
  - I’ve implemented this on `getContract()`. It could be a separate function and/or part of the window displayed when running a contract.

- Flatten difficulty curve of existing contracts to reduce spikes/gaps in contract difficulty:
  - The X.5s (`TotalWaysToSum` and `ArrayJumpingGame`) have been floored as a .5 precision feels irrelevant on a subjective difficulty scale. I do appreciate the irony here that I'm also stating averages to two decimal places.
  - `HammingCodes` have both moved up a difficulty based on number of questions they raise.
  - `AlgorithmicStockTraderIII` has been moved down by one as after I and II, III is not a major step.
  - This results in the same average difficulty of contract, but a much flatter distribution – see table below. These changes can be reverted easily as they are only a balancing issue.

| Difficulty | Current Count | Proposed Count |
| --- | --- | --- |
| 1 |  4 | 5 |
|1.5 | 1 | 0 |
| 2 | 6 | 7 | 
| 2.5 | 1 | 0 |
| 3 | 4 | 4 |
| 4 | 1 | 2 |
| 5 | 5 | 3 |
| 6 | 0 | 1 |
| 7 | 2 | 2 |
| 8 | 2 | 1 |
| 9 | 0 | 1 |
| 10 | 3 | 3 |
| Average Difficulty | 4.17 | 4.17 |

- ~Scale spawned contract difficulty based on `server.networkLayer`. This allows more established players to continue to seek out high value rewards in the harder servers whilst ensuring that newer players will more naturally stumble into lower difficulty problems during early-game exploration close to `home`. Setting a minimum difficulty spawn should allow a wider range of contracts to be appear over the course of a playthough, and encourage deeper exploration to further servers for better rewards.~
  - Removed for simplicity

- Scale spawned contract difficulty based on total SF levels the player has unlocked. This hard-caps the difficulty a brand-new player sees whilst not limiting more experienced players.
  - The proposed scale is in the table below. It's a range up to `2*totalSFs + 1`. This formula is easy to tweak based on feedback (e.g. based on installed augs instead of SFs). This results in a nerf to contract rewards when total SF levels are <= 5.

|Total SFs	|Max Difficulty	|Average Difficulty	|Contracts available|
| --- | --- | --- | --- |
|0	|1	|1.00	|5|
|1	|3	|1.94	|16|
|2	|5	|2.57	|21|
|3	|7	|3.08	|24|
|4	|9	|3.50	|26|
|5	|11	|4.17	|29|

## Alternate options:

There's some alternative measures I considered:

- Require root access to interact with contracts. I don’t feel this aligns with requirements 2 and 5, as it’s a significant change from current approach and doesn’t really improve requirement 1. I was also concerned this may be hard to future-proof if new server types are implemented with different rulesets. For comparison, `.lit` files can be accessed and moved without root access – contracts to me feel closer to `.lit` than `.js` files. 
- Randomise rewards (potentially as a function of server difficulty or other property). This doesn't really address requirement 1, and doesn’t sit particularly well with 2 or 5 – there is a less direct effort/reward ratio if randomness is increased.
- Scale with contracts completed rather than server depth/SF count. This doesn't satisfy requirements 2 and 5 easily since I don't think we currently count contract successes anywhere; all players will start at scratch. Additionally, requirement 1 may fail if the starting pool of contracts is a selection the player finds difficult or obnoxious.

## Conclusion:
I believe these changes enable players at all experiences have a relevant pool of contracts available throughout the game. As the player completes more Bitnodes and is more familiar with the game/mechanics, the range and difficulty of contracts increases.

These changes should allow for structured progression through the contract types whilst not penalising existing established players. The code changes I’ve made are as limited as I can make them to demonstrate a working proof of concept for the approach, and are probably full of flaws, but I’m keen for feedback on the philosophy, implementation and balance.


d0sboots:
I don't get the need/desire for two separate mechanisms here. Just adjusting the spawning based on SFs seems like it would accomplish all the goals, and be much simpler?

Edit: I also don't think we should have a minimum difficulty floor based on SF. IMO contacts should only be added to the pool, not removed.

gmcew:
Only using SFs would be much simpler and easier. With server scaling also present, the minimum difficulty SF floor only exists to allow deeper servers to spawn low level contracts. If we remove the server scaling the min difficulty property is superfluous and that whole logic can be removed.

The reason I also use server depth is to get a greater variety in contract spawns encountered and give more "progression" in the sense that the difficulty of what contract you run into isn't identical everywhere in the network; deeper exploration (or completely automated exploration) gives harder contracts/better rewards. Without it every server has an equal chance of every contract available - running into a max reward contract like `Find All Valid Math Expressions` on `joesguns` and a cheap `Find Largest Prime Factor` on the `The-Cave` feels inconsistent in the sense of growth/progression, though it is the current behaviour.

Contracts are never inaccessible since both scalars are considered together; the pool only ever grows and spreads across server depth. The interaction is probably not clear with the tables, but that also matches your observation that I'm overcomplicating this.

d0sboots:
I don't think variety across the network is all that important. Writing a network traversal script is one of the very early tasks, generally completed even before contacts are attempted. After that point, servers are more-or-less equivalent, aside from how much RAM they give you and their use in batching. Since effective contract automation requires searching across all servers, this is extra true.

gmcew:
With that change incorporated I take the concept as being reasonably sound, and ready for more formal review on how to implement it, and the other changes:
 - contract difficulty tweaks.
 - exposing difficulty via script/`run` command.
 - related wording on the docs page.

d0sboots:
Looks good!

Issue #2222, opened 2025-07-04
MISC: Rebalance infiltration rewards


catloversg:
Infiltration is one of the most controversial mechanics in Bitburner. Some players love it (and the notorious "auto-infil" scripts) while other players despise it. Every time we make a change in this mechanic, no matter what it is, we always receive negative feedback from players. It's clear that many people love its absurdly high rewards and want to use it as an "I win" button.

PR #2210 tries to make some changes that we have discussed for a long time. The approaches in that PR are good, but the formulas in that PR need to be tuned. After playtesting, I decide to write a detailed analysis of infiltration rewards and the suggestions. This report serves these purposes:
- Balancing is tricky. The result sometimes is the opposite of what we want. Generally speaking, PR 2210 nerfs infiltration rewards, but that PR accidentally significantly buffs the rewards in some cases. For example, at 1x mults (no augmentations), ECorp, which is the location with the highest rewards, currently "only" gives ~221k rep; with the new formulas in that PR, ECorp gives ~859k rep. The balance is even worse when we consider realistic cases (e.g,. having many combat augs). This report analyzes the changes and makes sure that the balancing result is exactly what we want.
- Everybody knows that the balance of infiltration is bad, but nobody explains *exactly* how bad it is. For example: "It breaks the game", okay, but how and how much? Does it give too much reputation? Too much money? How much is "too much"? If we have to nerf it, how much should we nerf it? 50%? 75%? 99%? In this report, I try to provide numbers and use cases while analyzing the flaws in this mechanic's balance and making balance suggestions.
- Explain why we make these changes and what they really change. At first glance, the nerf looks massive, but if you carefully read this entire report, you will see that the nerf only significantly affects the ridiculous cases (e.g., run auto-infil scripts at stat level 1 to infiltrate ECorp and get enough rep for 150 favor in a few seconds). In most realistic cases, the nerf is fairly small. There are even some cases in which the changes slightly buff the rewards.

Note that in some cases, with big tables, I'll post screenshots of them instead of the text. In the text form, GitHub UI makes it hard to read those tables.

All tables can be generated with scripts in this save file (you have to disable the effect of `applyExploit` in `src\Exploits\applyExploits.ts`):
[rebalance_infil.gz](https://github.com/user-attachments/files/21061956/rebalance_infil.gz)

Edit: The formulas in #2210 have been changed. For more information, please check that PR. I'll keep the original analysis here unchanged for reference.

## Goal

Let's take a look at these videos:

[...]

PR #2231, opened 2025-07-07
INFILTRATION: Tune count down time to mitigate extremely fast auto-infiltration

catloversg:
In #2222, we discussed about infiltration rewards and how to rebalance them. One of the problems of auto-infil is its speed (watch the third video in the linked issue). The new "market demand multiplier" mitigates this problem a bit, but it does not solve the problem entirely. When we tune that multiplier, we need to balance between manual infiltration and auto infiltration. Nerfing auto-infil while keeping manual infiltration relatively good is hard.

Auto-infil can easily complete a game that needs 15000ms in less than 100ms, so it's impossible for us to perfectly balance the rewards without punishing manual infiltration (by human) too much. This PR focuses on the core problem: the speed. After each floor, there is a count down that lasts 900ms. This PR adds an additional delay to that count down. This delay is designed to only affect non-human infiltration. If the player infiltrate manually, they won't see this delay or only see a tiny fraction of it. For the calculation, please check `calculateCountDownTime`.

Before:

"Simulated" manual infil:

https://github.com/user-attachments/assets/983d930a-9f03-423a-b2d7-01e5a4b4474c

Auto-infil:

https://github.com/user-attachments/assets/3a80deec-b0d4-42a5-aa2f-5c9efe4a26a1

After:

"Simulated" manual infil:

https://github.com/user-attachments/assets/06cc6a71-ce80-4eaa-9d68-640ae6139c68

Auto-infil:

https://github.com/user-attachments/assets/fc632832-0fb4-41e4-a5cb-ad8cb3d859c1

### DIFF
--- a/src/Infiltration/ui/Countdown.tsx
+++ b/src/Infiltration/ui/Countdown.tsx
@@ -1,27 +1,28 @@
 import { Paper, Typography } from "@mui/material";
-import React, { useEffect, useState } from "react";
+import React, { useEffect, useRef, useState } from "react";
 
 interface IProps {
-  onFinish: () => void;
+  time: number;
 }
 
-export function Countdown({ onFinish }: IProps): React.ReactElement {
+export function Countdown({ time }: IProps): React.ReactElement {
   const [x, setX] = useState(3);
+  const intervalId = useRef(0);
 
   useEffect(() => {
     if (x === 0) {
-      onFinish();
+      clearInterval(intervalId.current);
     }
-  }, [x, onFinish]);
+  }, [x]);
 
   useEffect(() => {
-    const id = setInterval(() => {
+    intervalId.current = window.setInterval(() => {
       setX((previousValue) => previousValue - 1);
-    }, 300);
+    }, time / 3);
     return () => {
-      clearInterval(id);
+      clearInterval(intervalId.current);
     };
-  }, []);
+  }, [time]);
 
   return (
     <Paper sx={{ p: 1, textAlign: "center" }}>
--- a/src/Infiltration/ui/Game.tsx
+++ b/src/Infiltration/ui/Game.tsx
@@ -1,5 +1,5 @@
 import { Button, Container, Paper, Typography } from "@mui/material";
-import React, { useCallback, useEffect, useState } from "react";
+import React, { useCallback, useEffect, useRef, useState } from "react";
 import { FactionName, ToastVariant } from "@enums";
 import { Router } from "../../ui/GameRoot";
 import { Page } from "../../ui/Router";
@@ -18,6 +18,7 @@ import { calculateDamageAfterFailingInfiltration } from "../utils";
 import { SnackbarEvents } from "../../ui/React/Snackbar";
 import { PlayerEventType, PlayerEvents } from "../../PersonObjects/Player/PlayerEvents";
 import { dialogBoxCreate } from "../../ui/React/DialogBox";
+import { clampNumber } from "../../utils/helpers/clampNumber";
 
 type GameProps = {
   StartingDifficulty: number;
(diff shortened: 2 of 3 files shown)


d0sboots:
I don't think we actually need this. All proposed versions that have a market demand multiplier use a function that requires searching for the global optimum; going faster than this *reduces* your rep gain. The issue of auto infil going too fast is naturally solved by that being suboptimal.

catloversg:
I have not playtested the new formula (your commit), but the old one look like this:

secondsPerFloor = 7 (average value, simulated manual infil)



secondsPerFloor = 5 (average value, with forced delays in this PR)



secondsPerFloor = 1 (auto-infil)



With extremely long sessions (e.g., 120m), ultra-fast infil is worse than normal infil, but with 5-30m sessions, ultra-fast infil is still hugely better than normal infil (especially 5m session). For example: 460.35 rep/s (1 second per floor) vs 92.405 rep/s (7 seconds per floor). With this PR, auto-infil is still faster than normal infil and gets better rewards (reward for writing auto-infil scripts), but the gap is fairly small, even in 5-15m sessions.

Another upside is that this "upperbound" speed of auto-infil (i.e., 5 seconds per floor with this PR) is not much higher than manual infil (5 vs 7 is not that much when comparing with 1 vs 7), so when we tune the formula of market demand multiplier, we don't have to think too much about the gap between manual infil and auto-infil.

d0sboots:
Your analysis is very different from what I saw when I was analyzing the original formula, although I *was* using the mistaken assumption of linear dropoff of rewards. But, I know for a fact that doing ultra-fast infil in bursts was *terrible* using the original formulation; each infil reduced the reward by a x.85, so doing that in a burst rapidly reduced the reward to 0.

Also, "ultra-long sessions" (120m) was pointless to analyze with the old mechanism, since it only had a memory of 30minutes tops.

Edit: Yeah, in your tables I'm not seeing the effect of market demand?

Double edit: OK, it does look like you are accounting for it in some sense, but at the same time the overall reward is hitting a limit as you go towards 120 min which doesn't make sense... I don't think your modeling was correct here. More edit: Rescinded, now that I understand your tables I think they're right.

Edit again: OK, no that does make sense because if you have reduced the rewards to 0 but keep doing infils on a regular cadence anyway, even with a 30 minute window you will keep it suppressed to the same constant amount regardless.
However, it remains true that there's no point to analyzing (the old way) on time periods > 30 min, nor is it relevant to check "bursty" mechanisms on shorter timescales since they will always be outperformed by "steady" mechanisms over 30 minutes. I.e. the 460.5 rep/s number is not relevant because you can only do it for 5 minutes every 30 minutes. What your numbers actually show is that 5 seconds/floor was roughly the best sustainable rate for ECorp (seen because at that point 30 min rewards/s ~= 60 minute rewards/s.)

Your numbers actually show that the old way was tuned so hard that *manual* infil wasn't sustainable, and it basically nerfed auto-infil into the ground. But, we're not tuning those numbers anymore.

d0sboots:
Another thing that I was considering while doing my patch was tracking number of floors instead of number of infils. Since, that is the actual unit of effort involved - the time taken is linear in floors, not infils. This would make it much simpler to balance against the fact that there is a very large swing between places like noodles and places like ECorp which, in practice, only a computer has the patience to do.

catloversg:
I agree that short-burst rep/s (460rep/s with 5m) does not mean much in the long term because they still have to wait 30m for the effect of market demand to come back to 100%. However, what about the difference in the speed of gaining the same reward between manual infil and auto infil?

Case 1:
Let's say the player already has 330k and wants to farm ~130k for hitting 460k rep.
- Manual infil: 30 minutes for 129k rep
- Auto infil: 5 minutes for 138k rep

Case 2:
Now they prestige (to hit 150 favor of the current target faction), take a 30-minute break, then target another faction.
- Manual infil: 30 minutes for 129k rep
- Auto infil: 5 minutes for 138k rep

In total:
- Manual infil: 90 minutes (30m session + 30m break + 30m session) for 129k of faction 1 and 129k rep of faction 2.
- Auto infil: 40 minutes (5m session + 30m break + 5m session)for for 138k of faction 1 and 138k rep of faction 2.

Case 1 shows the advantage of strategic short-burst auto-infil. Case 2 shows that the advantage becomes less relevant when auto-infil is used more frequently.

What do you think about this?

d0sboots:
Don't forget that in the manual case, there isn't any break actually needed. 

But, I came to the same conclusion in the other thread. The big advantage for bursts is that you can limit the time the UI is occupied, and then move on to other things. So, they are both important.

catloversg:
I will close this PR. From now on, we will prevent extremely fast auto infil via the new tuning parameters.

PR #2235, opened 2025-07-08
UI: API Break changes more verbose in terminal. Colorized some commands terminal output.

ZManak:
aaaa

Screenshot 2025-07-08 213328

Screenshot 2025-07-08 202638


.github/workflows/Create deploy-dev.yml
```diff
@@ -0,0 +1,33 @@
+name: "Deploy new pages build"
```
d0sboots, line 1: This seems accidental? In any case, revert - we already have one of these.

src/Programs/Programs.ts
```diff
@@ -180,8 +180,8 @@ export const Programs: Record<CompletedProgramName, Program> = {
       time: CONSTANTS.MillisecondsPerQuarterHour,
     },
     run: (): void => {
-      Terminal.print("This executable cannot be run.");
-      Terminal.print("DeepscanV1.exe lets you run 'scan-analyze' with a depth up to 5.");
+      Terminal.error("This executable cannot be run.");
+      Terminal.warn("DeepscanV1.exe lets you run 'scan-analyze' with a depth up to 5.");
```
d0sboots, line 184: This doesn't feel like a warning to me.

src/Programs/Programs.ts
```diff
@@ -193,8 +193,8 @@ export const Programs: Record<CompletedProgramName, Program> = {
       time: CONSTANTS.MillisecondsPer2Hours,
     },
     run: (): void => {
-      Terminal.print("This executable cannot be run.");
-      Terminal.print("DeepscanV2.exe lets you run 'scan-analyze' with a depth up to 10.");
+      Terminal.error("This executable cannot be run.");
+      Terminal.warn("DeepscanV2.exe lets you run 'scan-analyze' with a depth up to 10.");
```
d0sboots, line 197: Same.

src/Programs/Programs.ts
```diff
@@ -222,9 +222,34 @@ export const Programs: Record<CompletedProgramName, Program> = {
         return;
       }
 
-      Terminal.print(targetServer.hostname + ":");
+      if (targetServer.hasAdminRights) {
```
d0sboots, line 225: Admin rights are not relevant to hacking. We should leave this out, to avoid furthering this misconception.

src/Programs/Programs.ts
```diff
@@ -222,9 +222,34 @@ export const Programs: Record<CompletedProgramName, Program> = {
...
+        Terminal.success("Server has admin rights.");
+        Terminal.success(targetServer.hostname + ":");
+      } else {
+        Terminal.error("Server has no admin rights.");
+        Terminal.error(targetServer.hostname + ":");
+      }
       Terminal.print("Server base security level: " + targetServer.baseDifficulty);
-      Terminal.print("Server current security level: " + targetServer.hackDifficulty);
+      if (targetServer.hackDifficulty > targetServer.baseDifficulty) {
```
d0sboots, line 233: This is complicated and I'm not sure that it's a good idea. Let's revert this part for now, and do a different PR later that focuses on augmenting ServerProfiler if you still want to later. (The built-in.exe tools are *intentionally* crude to nudge you towards writing your own.)

src/Programs/Programs.ts
```diff
@@ -255,9 +280,9 @@ export const Programs: Record<CompletedProgramName, Program> = {
       time: CONSTANTS.MillisecondsPerQuarterHour,
     },
     run: (): void => {
-      Terminal.print("This executable cannot be run.");
-      Terminal.print("AutoLink.exe lets you automatically connect to other servers when using 'scan-analyze'.");
-      Terminal.print("When using scan-analyze, click on a server's hostname to connect to it.");
+      Terminal.error("This executable cannot be run.");
+      Terminal.warn("AutoLink.exe lets you automatically connect to other servers when using 'scan-analyze'.");
```
d0sboots, line 284: I'm going to comment on all of the following simply with "info" to say that these are informational messages only so I don't think it's appropriate to use .warn

src/Programs/Programs.ts
```diff
@@ -269,8 +294,8 @@ export const Programs: Record<CompletedProgramName, Program> = {
       time: CONSTANTS.MillisecondsPer4Hours,
     },
     run: (): void => {
-      Terminal.print("This executable cannot be run.");
-      Terminal.print("Formulas.exe lets you use the formulas API.");
+      Terminal.error("This executable cannot be run.");
+      Terminal.warn("Formulas.exe lets you use the formulas API.");
```
d0sboots, line 298: Info

src/utils/APIBreaks/APIBreak.ts
```diff
@@ -180,6 +182,15 @@ export function showAPIBreaks(version: string, { additionalText, apiBreakingChan
           ? `\n\nWe found ${pluralize(detail.totalDetectedLines, "affected line")}.`
           : ""),
     );
+
```
d0sboots, line 181: Extra newline

src/utils/APIBreaks/APIBreak.ts
```diff
@@ -180,6 +182,15 @@ export function showAPIBreaks(version: string, { additionalText, apiBreakingChan
...
     ++popUpIndex;
   }
+  //Made terminal more verbose
+  for (const detail of details) {
```
d0sboots: I like the idea of showing the info on the terminal, possibly instead of pop-ups. But this is (incorrectly) duplicating logic that the pop-ups use. It should run in the same loop and use the same code.

d0sboots:
@catloversg I've noticed when running the dev version that the number of pop-ups can be quite busy. What do you think about having the main overview still as a single pop-up, but printing the details to the console instead? (This PR currently does it as well.)

This would avoid pop-up exhaustion, and also the console has scrollback.

src/Terminal/HelpText.ts
```diff
@@ -1,51 +1,263 @@
...
-  "    scan-analyze [d] [-a]            Prints info for all servers up to d nodes away",
-  "    scp [files...] [server]          Copies a file to a destination server",
-  "    sudov                            Shows whether you have root access on this computer",
-  "    tail [script/pid] [args...]      Displays dynamic logs for the specified script",
-  "    top                              Displays all running scripts and their RAM usage",
-  "    unalias [alias name]             Deletes the specified alias",
-  "    vim [files...]                   Text editor - Open up and edit one or more scripts or text files in vim mode",
-  "    weaken                           Reduce the security of the current machine",
-  "    wget [url] [target file]         Retrieves code/text from a web server",
+  colors.green +
```
d0sboots: This is going to require a systemically different approach to avoid destroying the formatting we have here. 

The first step is to use template string literals. This means it's probably time to turns this whole thing into a single template string, which can embed newlines, rather than having an array of strings. 

The next step is to choose a set of short, *equal-sized* constants to use. You only need three, for green, yellow and reset. I would suggest CMD, ARG and TXT. You can then embed those in the string via {CMD} etc, and the even spacing will be preserved.

catloversg:
> @catloversg I've noticed when running the dev version that the number of pop-ups can be quite busy. What do you think about having the main overview still as a single pop-up, but printing the details to the console instead? (This PR currently does it as well.)
> 
> This would avoid pop-up exhaustion, and also the console has scrollback.

That sounds like a good idea.

ZManak:
> @catloversg I've noticed when running the dev version that the number of pop-ups can be quite busy. What do you think about having the main overview still as a single pop-up, but printing the details to the console instead? (This PR currently does it as well.)
> 
> This would avoid pop-up exhaustion, and also the console has scrollback.

I get it would be an only popup with all the relevant info together with a terminal output more detailed
Screenshot 2025-07-09 194143
Screenshot 2025-07-09 194115

src/Terminal/HelpText.ts
```diff
@@ -1,265 +1,109 @@
...
-  black: "\x1b[30m",
-  red: "\x1b[31m",
-  green: "\x1b[32m",
-  yellow: "\x1b[33m",
-  blue: "\x1b[34m",
-  magenta: "\x1b[35m",
-  cyan: "\x1b[36m",
-  white: "\x1b[37m",
-  reset: "\x1b[0m",
+  cmd: "\x1b[32m",
```
d0sboots, line 3: My suggestion here was to use all-caps to signify that they're constants, and then just define three separate constants CMD ARGS and TXT, not members of an object. 

Normally what you have here would be the better practice, but in this specific case where they're being used so much in string formatting, keeping them small and avoiding visual noise is more important.

src/Terminal/HelpText.ts
```diff
@@ -1,265 +1,109 @@
...
-    "Reduce the security of the current machine",
-  colors.green +
-    "    wget" +
-    colors.yellow +
-    " [url] [target file]         " +
-    colors.reset +
-    "Retrieves code/text from a web server",
-  " ",
-];
+/* const commands = {
```
d0sboots, line 8: Remove

src/Terminal/HelpText.ts
```diff
@@ -287,7 +131,7 @@ const TemplatedHelpTexts: Record<string, (command: string) => string[]> = {
 
 export const HelpTexts: Record<string, string[]> = {
   alias: [
-    colors.yellow + 'Usage: alias [-g] [name="value"] ',
+    'Usage: alias [-g] [name="value"] ',
```
d0sboots, line 134: If you want, you can also convert these to template strings and add coloring to them. But it is by no means required; this PR is already plenty big.

src/utils/APIBreaks/APIBreak.ts
```diff
@@ -153,44 +153,45 @@ export function showAPIBreaks(version: string, { additionalText, apiBreakingChan
...
+    }
+  */
 
-    ++popUpIndex;
-  }
   //Made terminal more verbose
-  for (const detail of details) {
-    if (detail.totalDetectedLines > 0) {
-      Terminal.warn("_____________");
+  for (const detail of details) {
```
d0sboots, line 183: Looking at this, and thinking through the consequences, I realize that switching to writing the details in the terminal will actually be rather involved, with only the smallest part of it being showing the details in color. (There's stuff like renaming fields where "showPopup" isn't correct anymore.)

So, I'm going to suggest that we cut that from this PR to avoid score creep, and do it in a different PR (possibly in parallel) that can focus solely on that.

PR #2236, opened 2025-07-10
CODEBASE: Fix inconsistent generated pages.ts

catloversg:
While reviewing #2227, I noticed that it changed the order of some generated lines in `pages.ts`. According to https://nodejs.org/api/fs.html#fsreaddirsyncpath-options and https://man7.org/linux/man-pages/man3/readdir.3.html, the order depends on the file system implementation. I tested the generator script on NTFS and ext4:
- NTFS (Windows 10): No change.
- ext4 (Debian 13): The order is changed.

This PR sorts the file list to make sure that the order is consistent among different file systems.


d0sboots:
Nice, I was wondering about that.

PR #2243, opened 2025-07-16
UI: Show money in exponential form instead of "0.000" when it's > 0 but still too small

catloversg:
It's just as the title said.

Before:

before

After:

after


d0sboots:
I'm not sure that we can make this change unilaterally. Money is displayed in a *lot* of places; I can't remotely guess them all off the top of my head. I would guess that for most, a value less than 0.0005 is just noise and it's better to display as 0, but there will indeed be some cases (like the one you showed) where it remains meaningful.

catloversg:
This change is now configurable with `useExponentialFormForSmallValue` in the latest commit. It's false by default. I only enable it in "Dividends per share" in the screenshot.

PR #2246, opened 2025-07-17
CODEBASE: Update dev dependencies

catloversg:
Updating dev dependencies usually has a lower chance of breaking things than updating (runtime) dependencies, so I update a bunch of them in this PR.

Skip:
- `api-documenter`: It changes how tables are generated in markdown files and creates changes in ~500 files.
- `eslint`: The latest major version requires a change in our eslint config.
- `prettier`: The latest major version makes too many changes in default configs.
- `monaco-editor`: To be honest, monaco is full of problems. I'll never update it in a big PR like this.
- `rehype` and `remark`: They are blocked by our React 17. For more information, please check https://discord.com/channels/415207508303544321/459097632896188436/1209083970499780613.

Remove: We don't use these libraries anymore.
- `@types/jquery`
- `@types/file-saver`

Note:
- `react-syntax-highlighter`: This library uses a vulnerable version of `prismjs`. We only use it to show blocks of code in our in-game documentation, so this vulnerability is not too severe. Nonetheless, I think we should deal with this problem. Fixing the vulnerable version is not easy due to its dependencies. For more information, please check https://github.com/react-syntax-highlighter/react-syntax-highlighter/pull/590 and https://github.com/react-syntax-highlighter/react-syntax-highlighter/pull/521. The workaround (use `overrides` in `package.json`) may create a problem due to `refractor` (check PR 521 in react-syntax-highlighter repo). I tested our in-game documentation for a while, but I don't see any problems.
- Change in `src\@types\global.d.ts`: `babel-standalone`'s packages have proper typing information now. We don't need this workaround anymore.
- Due to the above change, I also update the call to `babel.packages.parser.parse`.
  - `ecmaVersion`: This param does not exist.
  - `plugins` cannot be `string[]`.
- `src\ui\MD\components.tsx`: Please check the old comment of `fixAlign`. We need to expand the fix.

Edit: I have to revert all changes of dependencies related to Jest. The test run is broken in GitHub CI, but it works normally on my machine. I'm not sure why. Maybe Jest 30 is too new, and there are problems with it and its related projects.

I tested both web version and Electron version.


d0sboots:
Very strange with the jest thing. Even stranger that some tests passed.

PR #2252, opened 2025-07-19
BUGFIX: Loading code discards entire Go data due to missing migration code for favor/rep

catloversg:
How to reproduce: Load the attached save file and check the IPvGO tab. Without this patch, that tab is hidden because Go data is empty.

d0sboots mentioned this problem at https://github.com/bitburner-official/bitburner-src/pull/2131#pullrequestreview-2834438940. While testing a build of v3, I'm surprised that Go data is unexpectedly discarded when loading a save file of v2.8.1. IMO, this is a bug. The conversion formula is debatable (I choose the simplest way), but we should at least load *something* instead of discarding everything.

Test save file: 
[test-go-bug.gz](https://github.com/user-attachments/files/21326256/test-go-bug.gz)


d0sboots:
Good catch. I didn't think about the possibility of the typecheck code throwing out the old data. That is a little worrisome and in general, something we're going to have to watch for

PR #2254, opened 2025-07-19
UI: Enhance BitNode reward dialog formatting

ShubhamG2004:
### 🔧 Summary

This PR refactors the BitNode entry logic (`enterBitNode`) for clarity, maintainability, and enhanced UX:
- Modularized into smaller functions: `giveSourceFile`, `resetIntelligenceIfNeeded`, and `navigateAfterPrestige`
- Improved `dialogBoxCreate` formatting with emojis and clearer wording
- Preserved all game logic, with safer error fallback when setting BitNode options

### 📸 Visuals

No major UI component added. DialogBox output is slightly improved for readability.

### ✅ Checklist

- [x] Lint passed via `npm run lint`
- [x] Formatting passed via `npm run format`
- [x] Tests passed via `npm run test`
- [x] Based off latest `dev`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14224,6 +14224,7 @@
       "version": "0.52.0",
       "resolved": "https://registry.npmjs.org/monaco-editor/-/monaco-editor-0.52.0.tgz",
       "integrity": "sha512-OeWhNpABLCeTqubfqLMXGsqf6OmPU6pHM85kF3dhy6kq5hnhuVS1p3VrEW/XhWHc71P2tHyS5JFySD8mgs1crw==",
+      "dev": true,
       "license": "MIT"
     },
     "node_modules/monaco-editor-webpack-plugin": {
--- a/src/RedPill.tsx
+++ b/src/RedPill.tsx
@@ -15,40 +15,67 @@ import { exceptionAlert } from "./utils/helpers/exceptionAlert";
 function giveSourceFile(bitNodeNumber: number): void {
   const sourceFileKey = "SourceFile" + bitNodeNumber.toString();
   const sourceFile = SourceFiles[sourceFileKey];
+
   if (!sourceFile) {
-    console.error(`Could not find source file for Bit node: ${bitNodeNumber}`);
+    console.error(`Could not find source file for BitNode: ${bitNodeNumber}`);
     return;
   }
 
-  // Check if player already has this source file
-  let lvl = Player.sourceFileLvl(bitNodeNumber);
+  const currentLevel = Player.sourceFileLvl(bitNodeNumber);
 
-  if (lvl > 0) {
-    if (lvl >= 3 && bitNodeNumber !== 12) {
+  if (currentLevel > 0) {
+    if (currentLevel >= 3 && bitNodeNumber !== 12) {
       dialogBoxCreate(
         `The Source-File for the BitNode you just destroyed, ${sourceFile.name}, is already at max level!`,
       );
     } else {
-      lvl++;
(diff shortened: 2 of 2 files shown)


package-lock.json
```diff
@@ -14224,6 +14224,7 @@
       "version": "0.52.0",
       "resolved": "https://registry.npmjs.org/monaco-editor/-/monaco-editor-0.52.0.tgz",
       "integrity": "sha512-OeWhNpABLCeTqubfqLMXGsqf6OmPU6pHM85kF3dhy6kq5hnhuVS1p3VrEW/XhWHc71P2tHyS5JFySD8mgs1crw==",
+      "dev": true,
```
catloversg, line 14227: Why do you make this change?

src/RedPill.tsx
```diff
@@ -15,35 +15,39 @@ import { exceptionAlert } from "./utils/helpers/exceptionAlert";
...
         `The Source-File for the BitNode you just destroyed, ${sourceFile.name}, is already at max level!`,
       );
     } else {
-      lvl++;
-      Player.sourceFiles.set(bitNodeNumber, lvl);
-      dialogBoxCreate(`${sourceFile.name} was upgraded to level ${lvl} for destroying its corresponding BitNode!`);
+      const upgradedLevel = currentLevel + 1;
+      Player.sourceFiles.set(bitNodeNumber, upgradedLevel);
+      dialogBoxCreate(
+        `🎖️ ${sourceFile.name} was upgraded to level ${upgradedLevel} for destroying its corresponding BitNode!`,
```
catloversg, line 35: Our entire UI does not use emoji. It breaks the consistent style of the UI, so I don't see any reason for adding it.

src/RedPill.tsx
```diff
@@ -52,41 +56,58 @@ function giveSourceFile(bitNodeNumber: number): void {
   }
 }
 
+function resetIntelligenceIfNeeded(isFlume: boolean, destroyedBitNode: number, newBitNode: number): void {
```
catloversg, line 59: I don't think we need this kind of refactor (this function and `navigateAfterPrestige`). These function are tiny (`navigateAfterPrestige` contains only 1 if-else statement) and they are used in exactly one place.

By the way, this function receives `destroyedBitNode`, but it does not use it.

src/RedPill.tsx
```diff
@@ -15,35 +15,39 @@ import { exceptionAlert } from "./utils/helpers/exceptionAlert";
...
     Player.sourceFiles.set(bitNodeNumber, 1);
+
     if (bitNodeNumber === 5 && Player.skills.intelligence === 0) {
       Player.skills.intelligence = 1;
     }
+
     dialogBoxCreate(
       <>
-        You received a Source-File for destroying a BitNode!
+        <strong>🎉 You received a Source-File for destroying a BitNode!</strong>
```
catloversg, line 47: Same comment about emoji.

Regarding the usage of `<strong>`, please take some screenshots (before/after). I'm not sure if we need this change.

catloversg (commented):
This PR makes 2 changes:
- Add emoji.
- Refactor trivial code.

Unfortunately, I don't think we need these changes.

Edit: Some comments are also questionable.

d0sboots:
There are parts of this that I like, such as using more descriptive variable names. However, I agree with catlover that there isn't much point in creating small, single-use functions for logic that makes more sense in the flow of the main function. 

I also agree that emojis are a little counter to the style that Bitburner is trying to evoke, which is an "old school" terminal program.

PR #2257, opened 2025-07-21
FEATURE: Auto-infil API

d0sboots:
## Create an Infiltration API and add it as a reward to BN 11

BN 11 was underpowered, and also vaguely themed around corps. Infiltration has been in a perpetually weird state, as it is the only mechanic that is purely manual in a game about writing code to automate things. This aims to solve both of these things at once.

## Goals

* Auto-infil becomes a standard part of endgame, balanced appropriately (more on this later)
* With SF11.3, full auto-infil should be possible without SF4.
* Minigames remain playable manually, without any real changes
* The API should attempt to mimic the "feel" of old auto-infil, in that the challenges revolve around parsing the data and producing a proper sequence of "keypresses".
* Some minigames have increased difficulty to prevent them from being too trivial (such as slashGame), but all should still remain relatively easy to automate.

### Easy
* Automating each of the minigames at "Easy" difficulty
### Medium
* Dealing with edge-cases to produce a fully robust auto-infil loop
* Automating some of the minigames at higher difficulties
* Automatic picking of best infiltration targets
### Hard
* Using batching-like strategies to minimize running time of auto-infil

Because the main challenges are not too hard, the amount of rep gained will need to be greatly reduced from where it is currently. Balancing is where I imagine most of the remaining work is.

## Testing

For the time being, this is running at https://d0sboots.github.io/bitburner-src/

Attached is a minimal save in BN11 that includes a small script demonstrating some parts of the new API. Writing a full auto-infil is an exercise for the user XD
[bitburnerSave_autoInfilTest_BN11x1.json.gz](https://github.com/user-attachments/files/21462549/bitburnerSave_autoInfilTest_BN11x1.json.gz)


src/NetscriptFunctions/Infiltration.ts
```diff
@@ -81,5 +112,59 @@ export function NetscriptInfiltration(): InternalAPI<NetscriptInfiltation> {
...
+        return null;
+      }
+      return getState();
+    },
+    pressKey: (ctx) => (_key) => {
+      const key = helpers.string(ctx, "key", _key);
+      const result = getInfilLocation(Player.location);
+      if (Router.page() !== Page.Infiltration || !result.success) {
+        // Perform a no-op if we're not infiltrating. The lack of delay should
+        // help clue people in who are debugging.
```
ficocelliguy, line 204: Some script or console logging to these failure case(s) would help with debugging as well.

src/utils/helpers/shuffleArray.ts
```diff
@@ -0,0 +1,9 @@
+// https://en.wikipedia.org/wiki/Fisher-Yates_shuffle
+export function shuffleArray(array: unknown[]): void {
+  for (let i = array.length - 1; i > 0; i--) {
+    const j = Math.floor(Math.random() * (i + 1));
+    const temp = array[i];
```
Alpheus, line 5: If you hate the temp as much as I do you can
```suggestion
    [array[i], array[j]] = [array[j], array[i]];
```

(fairly textbook example, updated to 2025JS)

d0sboots: True, I was just refactoring it as-is from where it was. Probably both compile to the same thing.

src/NetscriptFunctions/Infiltration.ts
d0sboots: Done.

G4mingJon4s:
I played around a bit and noticed that the state given is just typed as `object`. I find that a bit annoying. Maybe we can type that better? I looked around a bit and put this here together: https://tsplay.dev/N7vBPw (edit: that `Exclude` isn't even needed 😄).

With that type, the player can instantly see what information they might get and in what format all the minigames are given.
I'm not sure how easy or hard it would be to implement that type into the internals, but the `StageState` alone would already improve the typing for the user and the minigame implementation itself.

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -5902,25 +5907,186 @@ interface InfiltrationLocation {
...
+   * Does not require API access.
    * @remarks
    * RAM cost: 0 GB
    *
    * @returns all locations that can be infiltrated.
    */
   getPossibleLocations(): ILocation[];
+
   /**
    * Get all infiltrations with difficulty, location and rewards.
```
ficocelliguy: This only gets one infiltration location's data, not "all infiltrations"

src/NetscriptFunctions/Infiltration.ts
```diff
@@ -81,5 +129,145 @@ export function NetscriptInfiltration(): InternalAPI<NetscriptInfiltation> {
       const locationName = getEnumHelper("LocationName").nsGetMember(ctx, _locationName);
       return calculateInfiltrationData(ctx, locationName);
     },
+    getMarketDemand: (ctx) => () => {
+      checkAccess(ctx);
+      return calculateMarketDemandMultiplier(Date.now(), false);
+    },
+    startInfiltration: (ctx) => (_locationName) => {
+      checkAccess(ctx, 3);
+      const locationName = getEnumHelper("LocationName").nsGetMember(ctx, _locationName);
```
ficocelliguy, line 138: It would be nice to have a default like `_locationName ?? Player.location` for simplicity

Alpheus:
Can sleeves infiltrate? Or help in any meaningful way?

d0sboots:
> I played around a bit and noticed that the state given is just typed as `object`. I find that a bit annoying. Maybe we can type that better? I looked around a bit and put this here together: https://tsplay.dev/N7vBPw (edit: that `Exclude` isn't even needed 😄).
> 
> With that type, the player can instantly see what information they might get and in what format all the minigames are given. I'm not sure how easy or hard it would be to implement that type into the internals, but the `StageState` alone would already improve the typing for the user and the minigame implementation itself.

No. If you read the docs for getState, it is *intentionally* unspecified. Figuring out what information is returned (and, if you're using typescript, dealing with the resulting typing issues) is meant to be part of the challenge. What you posted is what I would expect from a particularly mature TS-based solution.

You made a comment that the minesweeper map format is bad... That is also intentional. :)

Edit: You can, of course, think that this makes for a bad/unfun challenge :D But "deal with an awkward/unknown API" is something that comes up IRL, especially in a hacking context, and it's a challenge that we don't really have in this game yet... Well, except for BN14 which does something similar.

d0sboots:
> Can sleeves infiltrate? Or help in any meaningful way?

Not currently. I've thought about that possibility, but it would require future extensions where infill becomes more like a "task" and not the current thing that takes over the UI.

src/NetscriptFunctions/Infiltration.ts
d0sboots: Normally I would agree, except I have discovered some very gross things about Player.location. Basically, it's not very reliable, and I'm not even sure if that's a bug. If you are using the UI (as opposed to sing), it only gets updated when you enter company locations - going to the gym or slums etc. doesn't change it.

src/ScriptEditor/NetscriptDefinitions.d.ts
d0sboots: Done.

src/Augmentation/Augmentations.ts
```diff
@@ -979,7 +983,9 @@ export const Augmentations: Record<AugmentationName, Augmentation> = (() => {
       repCost: 1e4,
       moneyCost: 1e6,
       info: "Neodynic retention fjengeln spoofer using -φ karmions, net positive effect on implantee's delta wave.",
-      stats: "This augmentation makes the Wire Cutting minigame easier by indicating the incorrect wires.",
+      stats:
+        "This augmentation makes the Wire Cutting minigame easier by indicating the incorrect wires. " +
+        "API: Indicates incorrect wires.",
```
catloversg, line 988: Optional: Rephrase this line. Technically, we expose an array called `correctWires` that shows which wires are correct and which ones are not.

"Indicates incorrect wires" is not wrong, so if you think it's clear enough, we can use it as well.

src/Infiltration/State.ts
```diff
@@ -0,0 +1,29 @@
+import { EventEmitter } from "../utils/EventEmitter";
+
+// This is extremely minimal so we can set this global state without dependency cycles.
+// We could do without the class, except you cannot assign to exports.
+class InfilState<T> {
+  value: T | null = null;
+}
+
+export const difficultyState = new InfilState<number>();
```
catloversg, line 9: `difficultyState` is unused. `src\Infiltration\ui\Game.tsx` set its value, but nothing uses it.

src/Infiltration/ui/KeyHandler.tsx
```diff
@@ -1,7 +1,8 @@
 import React, { useEffect } from "react";
+import { InfiltrationKeyEvents } from "../State";
 
 interface IProps {
-  onKeyDown: (event: KeyboardEvent) => void;
+  onKeyDown: (this: IProps, event: KeyboardEvent) => void;
```
catloversg, line 5: Why do we need this change?

src/Infiltration/ui/KeyHandler.tsx
```diff
@@ -17,6 +18,15 @@ export function KeyHandler(props: IProps): React.ReactElement {
...
     return () => document.removeEventListener("keydown", press);
   });
+  // This creates untrusted events, but that's fine because the trusted
+  // filtering only happens above, for events that come in via event listener.
+  useEffect(
+    () =>
+      InfiltrationKeyEvents.subscribe((key: string) => {
+        props.onKeyDown(new KeyboardEvent("keydown", { key }));
+      }),
+    [props],
```
catloversg, line 28: This block only uses `props.onKeyDown`, so we should destructure `props` and put only `props.onKeyDown` to the dependency array.

Ref: https://github.com/facebook/react/issues/16265#issuecomment-517518539

src/Infiltration/ui/BracketGame.tsx
```diff
@@ -61,21 +81,35 @@ function match(left: string, right: string): boolean {
...
+    const difficulty: Difficulty = { timer: 0, min: 0, max: 0 };
+    interpolate(difficulties, props.difficulty, difficulty);
+    const left = generateLeftSide(difficulty);
+    const apiBrackets = generateApiBrackets(left, props.difficulty);
+    return {
+      left,
+      apiBrackets,
+      timer: difficulty.timer,
+    };
+  }, [props]);
```
catloversg, line 96:
```suggestion
  }, [props.difficulty]);
```

src/NetscriptFunctions/Infiltration.ts
```diff
@@ -81,5 +129,145 @@ export function NetscriptInfiltration(): InternalAPI<NetscriptInfiltation> {
...
+          };
+        }
+        const result = getInfilLocation(locationName);
+        if (!result.success) {
+          return { success: false, message: result.message };
+        }
+        if (locationPages.includes(Router.page()) && Player.location !== locationName) {
+          return {
+            success: false,
+            message: `Trying to infiltrate ${locationName}, but you are currently at ${Player.location}`,
```
catloversg, line 163: We discussed the problem of wrong `Player.location` when the player goes to gym/uni/travel agency/etc. Let's say the player goes to MegaCorp, goes to Powerhouse Gym, and then passes "Joe's Guns" to `startInfiltration`, this check prints `Trying to infiltrate Joe's Guns, but you are currently at MegaCorp`.

I'm not sure if we should keep this check. If you want to keep it, I think we should change the error message to `Trying to infiltrate Joe's Guns, but you were at MegaCorp` to avoid questions like "Why does it say I'm at MegaCorp when I'm at Powerhouse Gym?".

We also should mention this quirk of "last location" in TSDoc.

src/NetscriptFunctions/Infiltration.ts
```diff
@@ -81,5 +129,145 @@ export function NetscriptInfiltration(): InternalAPI<NetscriptInfiltation> {
...
+      });
+    },
+    pressSpace: (ctx) => () => {
+      checkAccess(ctx);
+      if (!getPlayerLocation(ctx)) {
+        return;
+      }
+      InfiltrationKeyEvents.emit(" ");
+    },
+    claimRewards: (ctx) => (_faction) => {
```
catloversg, line 225: This API should return -1 when there is an error. 0 is a valid reward due to market demand.

src/Infiltration/ui/InfiltrationRoot.tsx
```diff
@@ -1,19 +1,36 @@
...
   location: Location;
+  autoStart?: boolean;
 }
 
 export function InfiltrationRoot(props: IProps): React.ReactElement {
   const [start, setStart] = useState(false);
 
+  // useMemo is the "wrong" hook, useEffect is what is supposed to be used
+  // here. However, useEffect does not have guaranteed timing: In many/most
+  // cases, it fires in a delayed fashion, after screen painting has occured.
```
catloversg, line 22:
```suggestion
  // cases, it fires in a delayed fashion, after screen painting has occurred.
```

src/Infiltration/ui/BracketGame.tsx
```diff
@@ -43,6 +44,25 @@ function generateLeftSide(difficulty: Difficulty): string {
...
+// Make closing brackets more challenging by adding actual nested pairs of (),
+// [], {} and <> to the string. We will ensure that the nesting remains valid
+// (no instances of {[}],) as well as the proper closing sequence still
+// matching the original string.
+function generateApiBrackets(left: string, difficulty: number): string {
+  if (Player.hasAugmentation(AugmentationName.WisdomOfAthena, true)) {
+    difficulty = 0;
+  }
+  // There are many ways of randomly creating nesting brackets. This way gives
+  // greater liklihood to areas that already have brackets, similar to how a
```
catloversg, line 56:
```suggestion
  // greater likelihood to areas that already have brackets, similar to how a
```

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -5902,25 +5907,186 @@ interface InfiltrationLocation {
...
+
+  /**
+   * Start an infiltration.
+   *
+   * This brings up the infiltration minigame, exactly as if you started it
+   * manually. The "intro screen" (which shows expected rewards) is skipped.
+   * You must be in the same city as the location you are attempting to
+   * infiltrate, and due to the UI-focused nature of infiltration it only
+   * works from a small set of screens. These include the Terminal, Script
+   * Editor, Active Scripts, City, Location, and Infiltration (Start) screens.
```
catloversg, line 5972: Optional: Clarify what "Location" screen is. Some players may not understand that "Location" screen is the generic screen that we use for company/gym/uni/etc.

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -5902,25 +5907,186 @@ interface InfiltrationLocation {
...
+   * drastically from minigame to minigame, and possibly from call to call!
+   * You are meant to experiment to discover the contents of this object, and
+   * how to use it to solve the various minigames.
+   *
+   * Here are the things that are guaranteed:
+   *
+   * - The returned object is fixed, i.e. it won't be modified after it is
+   *   returned. Call this function again if you want to see what is happening
+   *   at different times.
+   * - There will be a "stage" field which describes the current minigame, or
```
catloversg, line 6001: You need to put an extra newline after each bullet point; otherwise, the generated markdown will have everything on a single line.

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -5902,25 +5907,186 @@ interface InfiltrationLocation {
...
+   *   if infiltration is in some other state (intro, collect rewards,
+   *   in-between games, etc.)
+   * - In most cases there will be "floors" and "currentFloor" (1-indexed) to
+   *   indicate your overall progress through the game.
+   * - In most cases there will be "timer" to indicate the time remaining (in
+   *   miliseconds) for this stage.
```
catloversg, line 6007:
```suggestion
   *   milliseconds) for this stage.
```

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -5902,25 +5907,186 @@ interface InfiltrationLocation {
...
+   * (one the player has joined, and can work for normally).
+   *
+   * This only works from the "victory" screen, and does nothing otherwise.
+   * Like usual, the player will be left on the City page after.
+   *
+   * Requires BN11 or SF11.2
+   * @remarks
+   * RAM cost: 15 GB
+   *
+   * @returns The amount of money or rep gained, or 0 on error
```
catloversg, line 6072: Update this line after changing the implementation (return -1 on error).

catloversg (commented):
### Bug
`pressSpace` does not behave exactly the same as `pressKey(" ")`.

Test code:
```js
/** @param {NS} ns */
  ns.ui.openTail();
  ns.clearLog();
  ns.infiltration.startInfiltration("Joe's Guns");
  await ns.sleep(1000);
  await ns.infiltration.pressKey("s");
  // await ns.infiltration.pressKey(" "); // ok
  ns.infiltration.pressSpace(); // error
}
```
How to reproduce:
- Force bribe game.
- Force choices (`makeChoices` function): `["boring", "helpful"]`.
- Run test code.

This bug happens with bribe and minesweeper but does not happen with slash and cyberpunk2077.

### TODO
- Update SF info: `src\BitNode\BitNode.tsx`, `src\Documentation\doc\en\advanced\bitnodes.md`.
- Rerun `npm run doc`. `pages.ts` has not been regenerated.
- Change `"You were hospitalized. Do not try to automate infiltration!"` to something else. Auto-infil is a "sanctioned" gameplay now. For example: `"You were hospitalized. Your auto infiltration script is not good enough!"`. Rephrase the comment above that code as well.
- Add the effect of WisdomOfAthena on API (remove `[` and all "noise pairs") to `src\Augmentation\Augmentations.ts`.

### Balance

Auto-infil is a sanctioned gameplay now, so we need to tune the decay rate again.

I suggest that we use `-4e-6` (optimal speed: 0.079621-0.118621 floor/sec ; 12.5595-8.43021 sec/floor) that fico used at https://github.com/bitburner-official/bitburner-src/pull/2210#issuecomment-3085678633. I vetoed that rate at https://github.com/bitburner-official/bitburner-src/pull/2210#issuecomment-3089975732 because it nerfed auto infil too much. Back then, auto infil was still in the exploity territory, so we should keep the rewards a bit high. With these new APIs, I think we should tune the optimal speed at the manual speed, so `-4e-6` is okay.

Edit: Baseline (Copied from #2222):
- In the very early stage (hack 100), even with an unrealistic number of augs (25 NFGs + all rep augs), rep/s is very low (1.5544/s).
- At hack 500 and 50 NFGs, rep/s is still single-digit (9.9412/s).
- At hack 1500 and 50 NFGs, rep/s is ~30/s (29.810/s).
- At hack 2500 and 50 NFGs, rep/s is ~50/s (49.680/s).
- At hack 2500, 75 NFGs, ns.share (1PB), rep/s is still less than 100/s (95.520/s).

The rep/s of extremely easy locations (e.g., Noodle Bar, Joe's Guns) is still very high, but I'm not sure if we need to nerf it further.

On one hand, due to very high rep/s, these APIs will become the new best way to farm rep in the early/mid stages, so I'm fairly worry about the balance. On the other hand, if we nerf it too much, infil will become a boring mechanic. I'm open to suggestions here. I'll try to tune `balanceMultiplier` if requested.

12 seconds/floor:


10 seconds/floor:


7 seconds/floor:


5 seconds/floor:

src/NetscriptFunctions/Infiltration.ts
Snarling: Alternately it can just throw an error if there's an error, -1 is a pain to handle as an error value in player scripts.

src/NetscriptFunctions/Infiltration.ts
```diff
@@ -81,5 +129,145 @@ export function NetscriptInfiltration(): InternalAPI<NetscriptInfiltation> {
...
+      }
+      const victory = victoryState.value;
+      if (!victory) {
+        helpers.log(ctx, () => "The current infiltration is not finished yet.");
+        return 0;
+      }
+      const state = stageState.value?.() as { [k: string]: number } | undefined;
+      if (faction == null) {
+        const money = state?.possibleMoneyGain ?? 0;
+        helpers.log(ctx, () => `Sold rewards for ${money === 0 ? "ERROR_UNKNOWN" : formatMoney(money)}`);
```
Snarling, line 239: I think Cat mentioned this somewhere else but if market demand is allowed to be 0, then the money reward could also be 0 without it being an error.

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -5902,25 +5907,186 @@ interface InfiltrationLocation {
...
+   */
+  claimRewards(faction: FactionName | null): number;
```
Snarling, line 6074:
```suggestion
  claimRewards(faction?: FactionName): number;
```
I think the explicit null is a bit weird here as a player, makes more sense to me as above: omit the faction arg to get the money reward. Doesn't need an actual implementation change as the actual function code treats undefined and null the same.

Would need a doc change as well in the description for the faction parameter.

catloversg:
I took a look at the bug with bribe. It's a bit complicated to describe it, so I committed a PoC workaround to illustrate the problem and a (maybe dumb) workaround. The gist is that the `press` function "captures" the value of `currentChoice`, so when the space key is pressed synchronously via `pressSpace` API, that function uses the old value of `currentChoice` instead of the new one. If `pressKey` (async function) is used, React can "reinitialize" the key handler (the second `useEffect` in `KeyHandler.tsx`) with the new `press` function.

I'm sorry if my description is confusing. It will be much easier to understand if you put breakpoints in the `press` function (`BribeGame.tsx`) and the second `useEffect` (`KeyHandler.tsx`). If you find a better fix, please feel free to use it instead of my workaround. I think the problem with the minesweeper game is the same.

src/NetscriptFunctions/Infiltration.ts
d0sboots: We are trying to be more principled about when/how we throw. I describe the policy this way: Exceptions are for programmer mistakes, error codes are for unexpected data. This policy is somewhat derived from the fact that our exceptions are essentially strings, instead of typed objects - so they cannot effectively be caught or filtered, and any attempt on a player's part to use try-catch will be an exercise in frustration. 

As a result, all exceptions should be assumed to be fatal, which comes back around to only using them for clear programmer mistakes. (Non-existent server or faction name, for instance.)

d0sboots:
> I took a look at the bug with bribe. It's a bit complicated to describe it, so I committed a PoC workaround to illustrate the problem and a (maybe dumb) workaround. The gist is that the `press` function "captures" the value of `currentChoice`, so when the space key is pressed synchronously via `pressSpace` API, that function uses the old value of `currentChoice` instead of the new one. If `pressKey` (async function) is used, React can "reinitialize" the key handler (the second `useEffect` in `KeyHandler.tsx`) with the new `press` function.
> 
> I'm sorry if my description is confusing. It will be much easier to understand if you put breakpoints in the `press` function (`BribeGame.tsx`) and the second `useEffect` (`KeyHandler.tsx`). If you find a better fix, please feel free to use it instead of my workaround. I think the problem with the minesweeper game is the same.

I thought I had worked around this sufficiently, but this solidifies my ultimate fear:
React's event-handling model is a house of cards and we *cannot* trust it if we are relying on specific guarantees for when things happen, especially synchronous guarantees. The ultimate problem is not that `press` captures `currentChoice` (that is expected, and known), but rather that the *new* copy of `press` is not yet registered by the time `pressKey` returns, because React does most things asynchronously.

I think ultimately the only way we are going to be able to trust the logic here is to handle the game state ourselves, like we do with the rest of BitBurner, and treat React only as a presentation layer. Infil was one of the last outliers in this regard.

d0sboots:
Current status of this PR: I am working on a separate PR (slowly) to refactor infiltration to pull all state out of the UI and into our own classes. I believe this is necessary for a bug-free auto-infil, and it also enables future features like infil with sleeves, and/or infil as a job (like crime) instead of being a full-screen thing. However, it's big (bigger than this PR).

Once that is done, I will rebase this PR on top of that one, and this should become much smaller. Until then, consider this PR on hold.

catloversg:
Do you mind if I test balancing params and post the result here for discussion? This PR is blocked by UI bugs and balancing. While you are implementing the proper fix for UI, I can research the balancing.

d0sboots:
OK, except with the caveat that I was thinking of tweaking the formula again so the specific numbers might not be relevant. 

In particular, the previous formula and numbers were "balanced" for auto-infill being not totally broken, but not intended. With it being an intended mechanic, I want to reward ingenuity in solutions, which means faster auto infil should continue to yield more rep, but with diminishing returns.

catloversg:
I think I'll wait until you tweak the formulas.

PR #2205, opened 2025-06-14
MISC: Add removal of API server to list of breaking changes

catloversg:
It's just as the title said.

Capture

```
API BREAK INFO FOR 3.0.0

You cannot use the deprecated VSCode extension to connect to the Steam app via API server now. You have to use Remote API to connect your external editors to Bitburner.
For more information, please check the pinned message in #external-editors on Discord (https://discord.gg/TFc3hKD).
```

PR #2258, opened 2025-07-21
DOCUMENTATION: Update Remote API docs

catloversg:
In #2205, d0sboots suggested adding a link to a comparison table. However, when asking around on Discord and checking those tools, I decide to not include that table in the breaking change notice due to these reasons:
- That table is in the viteburner repo, and it's a bit biased towards viteburner. Some rows are debatable/questionable.
- It has 4 columns, but only bitburner-filesync and viteburner are relevant now. We also need to add a new column for bb-external-editor.
- I'm always skeptical when looking at comparison tables. These tables tend to create unwanted (and maybe useless) debates. If we need one, we should write it, but I'm too lazy. I also don't want to deal with debates about some specific feature rows.

In this PR, I try to add more information about Remote API.
- Explain how it works with a sequence diagram.
- Add a list of community tools. No comparison table.
- Add troubleshooting tips.
- Update API specification.

Capture

I also wrote the demo of a simple tool to demonstrate how to write a Remote API tool, but I did not include it in this PR. I think most people only want to use a tool, not write it. I'll write more about the demo if requested.


tools/bundle-doc/index.js
```diff
@@ -55,21 +55,18 @@ processDir(docImagesRoot);
  * much easier than checking ~1400 lines.
  */
 const autogeneratedContent = `// THIS FILE IS AUTOGENERATED. DO NOT EDIT IT MANUALLY.
-${docFiles.map((f, i) => `import file${i} from "./doc/${f}?raw";`).join("\n")}
-${nsDocFiles.map((f, i) => `import nsDoc_${f.replaceAll(".", "_")} from "../../markdown/${f}?raw";`).join("\n")}
-${docImagesFiles.map((f) => `import docImages_${f.replaceAll(/\.|-/g, "_")} from "./images/${f}";`).join("\n")}
-
+${docFiles.map((f, i) => `import file${i} from "./doc/${f}?raw";\n`).join("")}
+${nsDocFiles.map((f) => `import nsDoc_${f.replaceAll(".", "_")} from "../../markdown/${f}?raw";\n`).join("")}
+${docImagesFiles.map((f) => `import docImages_${f.replaceAll(/\.|-/g, "_")} from "./images/${f}";\n`).join("")}
```
d0sboots, line 60: This is an aesthetic choice/trade-off. As it stands now, there is effectively an extra newline for each of these from the "natural" end-of-line after the statement. This is creating extra spacing between the groups, which can be a good thing, but it also means there would be extra spaces if any group were empty.

The alternative would be to have all of these on one line. That would fix the generated output, but the code would be harder to read. The generated file is not meant to be looked at, so I don't really care either way - I bring this up as a trade-off between the two styles, for the future.

src/Documentation/doc/en/programming/remote_api.md
```diff
@@ -1,14 +1,35 @@
 # Remote API
 
-All versions of Bitburner can use websockets to connect to a server.
-That server can then perform a number of actions.
-Most commonly this is used in conjunction with an external text editor or API
-in order to make writing scripts easier, or even use typescript.
+Bitburner can connect to a WebSocket server, and then that server can read and write Bitburner data via some APIs. The most common usage of this feature is to synchronize files between Bitburner and an external system. With a Remote API tool, you can write their scripts in any text editor and synchronize their scripts with Bitburner.
```
d0sboots:
```suggestion
Bitburner can connect to a WebSocket server, and then that server can read and write Bitburner data via some APIs. The most common usage of this feature is to synchronize files between Bitburner and an external system. With a Remote API tool, you can write your scripts in any text editor and synchronize your scripts with Bitburner.
```

Or "one can write their" to be more formal, but either way you need subject/object agreement

d0sboots (commented):
Good. I don't think a comparison table is necessarily needed, but links to the known decent tools are. If it is possible to succinctly summarize (one sentence) the main purpose of each of them to go with those links, that would be good too - I know they're not all trying to do exactly the same thing.

src/Documentation/doc/en/programming/remote_api.md
catloversg: My bad. I used "the player" at first, then changed to "you", but I forgot to change the object.

catloversg:
I tried to point out the difference like this:
> `typescript-template` has a small set of options and features. Its simplicity is by design. `viteburner` and `bb-external-editor` have more fancy features.

`viteburner` and `bb-external-editor` have some unique features, but I don't know how to write them succinctly without making a comparison table.

d0sboots:
The text looks good. For some reason, I wasn't able to see the SVG last time. I find that diagram a bit confusing, and maybe unneeded: 
* I would take the "Text editor" node out entirely, it's adding an unneeded layer of abstraction. For this, it's fine to conceptualize players editing files directly on file storage. 
* I'm not sure that the entire editing flow of actions is accurate, or aids understanding. Typically, if Remote API writes changes to disk, it means API calls were made first. If it reads changes from disk, the calls come after. The ordering relative to player actions also depends on the data transfer direction, API, and scenario. I think the main thing people care about is seeing the setup, which is also invariant.

catloversg:
Update:
- Move the diagram out of the first section. I think most players do not read (or care about) the sections below "Troubleshooting tips". The diagram and the API specs should only be relevant for people who want to write a tool or be interested in how it works behind the scenes.
- Remove "Text editor" and use alt fragment in the diagram to split read and write cases.

remote-file-api-sequence-diagram

PR #2260, opened 2025-07-22
UI: Hide outline of modal

catloversg:
The modal usually looks like this:

focus-visible-false

However, some modals (documentation popup, hashnet upgrade, etc.) look like this on Chrome:

focus-visible-true

On Electron:

focus-visible-true-electron

The white/yellow outline is caused a Chromium behavior. For more information, please check:
- https://github.com/mui/material-ui/issues/11504
- https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible

This is an accessibility feature, so I'm a bit hesitant when hiding it, but I don't think it's really useful. I'll close this PR if you think we should not touch this accessibility feature.

### DIFF
--- a/src/ui/React/Modal.tsx
+++ b/src/ui/React/Modal.tsx
@@ -28,6 +28,9 @@ const useStyles = makeStyles()((theme: Theme) => ({
       display: "none",
     },
     scrollbarWidth: "none", // firefox
+    "&:focus-visible": {
+      outline: "none",
+    },
   },
   closeButton: {
     position: "absolute",


d0sboots:
This is some combination of mui and chromium bug. It is thinking the element needs to be highlighted via :focus-visible, when this is not the case. We should *not* remove this highlighting, since it would break keyboard nav (worse than it already is).

The effect is minor (I've never noticed it), so we can also easily live with it.

catloversg:
Can you elaborate on why it would break keyboard nav? It's a CSS change, so I don't get why it affects the keyboard nav.

d0sboots:
It doesn't break keyboard nav directly, but keyboard navigation depends on this highlighting to know what is currently focused. (Unlike mouse nav, where you know because it's the thing you clicked on)

PR #2264, opened 2025-07-23
MISC: Fix typo in KARMA_1000000 achievement

UncleCeiling:
# Linked issues

None

# Documentation

No changes needed

# Bug fix

Formatted and linted

# Note

I did look for other instances, but couldn't find any - the steam achievement is spelt correctly too.

### DIFF
--- a/src/Achievements/AchievementData.json
+++ b/src/Achievements/AchievementData.json
@@ -213,7 +213,7 @@
     },
     "KARMA_1000000": {
       "ID": "KARMA_1000000",
-      "Name": "Wretched hive of scum and vilany",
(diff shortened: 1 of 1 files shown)


d0sboots:
This requires extra effort if we want to fix it on the steam achievement, which I don't know how to do off the top of my head. But since this doesn't touch the ID, it shouldn't cause issues at least.

UncleCeiling:
> This requires extra effort if we want to fix it on the steam achievement, which I don't know how to do off the top of my head. But since this doesn't touch the ID, it shouldn't cause issues at least.

Thankfully the steam Achievement is correct, so no worries there :)
image

PR #2265, opened 2025-07-23
BUGFIX: Error message shows blob URL instead of script name

catloversg:
This is a regression bug caused by #2169. `basicErrorMessage` has a loop that parses the blob url, but it's not used in that PR.

This PR moves the loop mentioned above to a new function and calls it in `DisplayError`. I also change how the error message is generated when the error is a syntax error.

Test save file:
[fix-error-message-with-blob-url.gz](https://github.com/user-attachments/files/21390206/fix-error-message-with-blob-url.gz)

Before:

before

After:

after

### DIFF
--- a/src/Netscript/ErrorMessages.ts
+++ b/src/Netscript/ErrorMessages.ts
@@ -1,5 +1,4 @@
 import type { WorkerScript } from "./WorkerScript";
-import { ScriptDeath } from "./ScriptDeath";
 import type { NetscriptContext } from "./APIWrapper";
 
 /** Log a message to a script's logs */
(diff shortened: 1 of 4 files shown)


src/Netscript/ErrorMessages.ts
```diff
@@ -1,20 +1,22 @@
...
+export function parseBlobUrlInMessage(ws: WorkerScript, msg: string): string {
+  for (const [scriptUrl, script] of ws.scriptRef.dependencies) {
+    msg = msg.replaceAll(scriptUrl, script.filename);
   }
-  return `${type} ERROR\n${ws.name}@${ws.hostname} (PID - ${ws.pid})\n\n${msg}`;
+  return msg;
+}
+
+/** Creates an error message string containing hostname, scriptname, and the error message msg */
+export function basicErrorMessage(ws: WorkerScript, msg: string, type = "RUNTIME"): string {
```
d0sboots, line 18: Does this still need to be exported?

catloversg: No. Except `errorMessage` in the same file, I don't think anything needs it. I'll remove `export` the next time I change something in this file.

PR #2266, opened 2025-07-23
MISC: Update messages related to text files

catloversg:
Many messages only mention .txt when talking about text files. This PR adds .json to those messages.

### DIFF
--- a/markdown/bitburner.ns.mv.md
+++ b/markdown/bitburner.ns.mv.md
@@ -30,7 +30,7 @@ RAM cost: 0 GB
 
 Move the source file to the specified destination on the target server.
 
-This command only works for scripts and text files (.txt, .json). It cannot, however, be used to convert from script to text file, or vice versa.
+This command only works for scripts (.js, .jsx, .ts, .tsx) and text files (.txt, .json). It cannot, however, be used to convert from script to text file, or vice versa.
 
 This function can also be used to rename files.
 
--- a/markdown/bitburner.ns.mv.md
+++ b/markdown/bitburner.ns.mv.md
@@ -30,7 +30,7 @@ RAM cost: 0 GB
 
 Move the source file to the specified destination on the target server.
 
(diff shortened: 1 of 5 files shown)


markdown/bitburner.ns.mv.md
```diff
@@ -30,7 +30,7 @@ RAM cost: 0 GB
 
 Move the source file to the specified destination on the target server.
 
-This command only works for scripts and text files (.txt). It cannot, however, be used to convert from script to text file, or vice versa.
+This command only works for scripts and text files (.txt, .json). It cannot, however, be used to convert from script to text file, or vice versa.
```
d0sboots: This is a little ambiguous, because it doesn't also list all the types of scripts. It might be better to not list any types here, since it distracts from the purpose.

src/NetscriptFunctions.ts
```diff
@@ -1740,7 +1740,7 @@ export const ns: InternalAPI<NSFull> = {
       (!hasTextExtension(sourcePath) && !hasScriptExtension(sourcePath)) ||
       (!hasTextExtension(destinationPath) && !hasScriptExtension(destinationPath))
     ) {
-      throw helpers.errorMessage(ctx, `'mv' can only be used on scripts and text files (.txt)`);
+      throw helpers.errorMessage(ctx, `'mv' can only be used on scripts and text files (.txt, .json)`);
```
d0sboots: On the other hand, this is an error message so IMO it should also list the scripts.

catloversg:
When checking the mention of "script" and "text files", I find out that we mention the extensions of both scripts and text files in most places. In the last commit, I add 4 extensions of scripts to all missing places. Now, all places mention `scripts (.js, .jsx, .ts, .tsx) and text files (.txt, .json)`.

Regarding your comment about `ns.mv`, please let me know if you still want to do that with docs of NS APIs. If that's the case, I'll update the docs of all APIs that require scripts and text files.

PR #2267, opened 2025-07-24
BUGFIX: ns.mv writes to destination file even if it cannot delete source file

catloversg:
#2160 does not inverse the check like I suggested at https://github.com/bitburner-official/bitburner-src/pull/2160/files/1e6b5ef52b6374761380c42745e0089e1d516fda#r2106262967. It moves `return;` to the wrong place.

Capture


UncleCeiling:
Ah, I think I understand where I went wrong here - apologies!

PR #2269, opened 2025-07-24
MISC: Update changelog

catloversg:
Updating the version number now is needed to avoid running migration code and showing notice every time the player loads the game. Snarling also did this in v2.6.1. If we make more breaking changes later, we will update the version number again.

I'll mark this PR as ready for review after the poll of gang equipment names ends.

### DIFF
--- a/src/Constants.ts
+++ b/src/Constants.ts
@@ -6,7 +6,7 @@
 export const CONSTANTS = {
   VersionString: "3.0.0dev",
   isDevBranch: true,
-  VersionNumber: 44,
+  VersionNumber: 43,
 
   /** Max level for any skill, assuming no multipliers. Determined by max numerical value in javascript for experience
    * and the skill level formula in Player.js. Note that all this means it that when experience hits MAX_INT, then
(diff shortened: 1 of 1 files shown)


d0sboots:
I think we should do this as part of pushing a new build to the the dev branch, and possibly also to dev on Steam (maybe delayed by a day or two).
I'm selfishly hoping to get auto-infil in as part of that, but it can be skipped if necessary. The migration stuff is what needs the most testing.

src/Constants.ts
```diff
@@ -6,7 +6,7 @@
 export const CONSTANTS = {
   VersionString: "3.0.0dev",
   isDevBranch: true,
-  VersionNumber: 43,
+  VersionNumber: 44,
```
d0sboots: Can you leave this off for now, and we'll just update the changelog first? That can go in immediately, and then bumping the version can be a small change on its own.

catloversg:
> I'm selfishly hoping to get auto-infil in as part of that, but it can be skipped if necessary. The migration stuff is what needs the most testing.

It's fine to wait for that PR.

PR #2272, opened 2025-07-30
API: Expose gang's discount rate

catloversg:
This change was suggested on Discord.

Test code:
```js
/** @param {NS} ns */
  ns.ui.openTail();
  ns.print(ns.gang.getGangInformation().discount);
}
```

Capture


src/NetscriptFunctions/Gang.ts
```diff
@@ -113,6 +113,7 @@ export function NetscriptGang(): InternalAPI<IGang> {
         wantedLevel: gang.wanted,
         wantedLevelGainRate: gang.wantedGainRate,
         wantedPenalty: gang.getWantedPenalty(),
+        discount: 1 - 1 / gang.getDiscount(),
```
d0sboots: I am not sure that this is the most natural number to expose? I was expecting a price multiplier, which would be `1 / gang.getDiscount()`, since most of our numbers are in terms of multipliers.

catloversg: Is "discount" still a good word for `1 / gang.getDiscount()`? In the screenshot, `gang.getDiscount()` is 4, so it means "sale off 75%" (i.e., mult=1/4=0.25). What if the player asks us why "discount" on UI is not the same as "discount" in API? I'm fine in either case, as long as we choose a proper word.

UncleCeiling: I think keeping the UI numbers as close to the numbers provided via the API is a good way to help people reason about what's going on without having to work out how things are being calculated behind the scenes (unless they want to).

That being said if we exposed `gang.getDiscount()` (g.gD) instead of `1-1/g.gD`; maybe describing it as the Discount Factor might make it's role more explicit?

d0sboots: > I think keeping the UI numbers as close to the numbers provided via the API is a good way to help people reason about what's going on without having to work out how things are being calculated behind the scenes (unless they want to).

I normally agree with this, but I'm this case I disagree, which is why I commented. 

Specifically, if you are going to *do* anything with this value, it is almost certain that you have to convert it to "price multiplier" first. (That's my proposal for naming.) Discount is useless for calculations, since it has already been applied to the cost and all methods of "unapplying" it require calculating price multiplier first. 

OTOH, when just showing it in the UI, discount isn't the worst thing. It's not a number that requires action (because it's already been applied to the other visible numbers), so a percentage that approaches 100 feels better than one that approaches 0.

PR #2273, opened 2025-07-30
DEVMENU: Fix wrong stat level due to floating-point imprecision

catloversg:
Due to floating-point imprecision, the stat tool may not set the value that we want.

Before:

before

After:

after


d0sboots:
Counter-proposal: https://github.com/bitburner-official/bitburner-src/pull/2274

PR #2281, opened 2025-08-05
CONTRACTS: Refresh the server info after a contract promise returns

### LINKED ISSUE #2280: Bug: It's possible to submit a contract after a reset
Steam version, v.2.8.1 (Debian 12)

If you have a contract window open to submit a contract answer, and in the background a script is running that will install an augmentation and then trigger a reset, then the contract window will remain open and you will receive a reward for a correct answer.

TheCleric:
# Linked issues

closes #2280

# Bug fix

- Tested via loading the save file from #2280 and running the command provided there.
- Verified:
    - After this fix the credit is no longer given for a contract began before a reset and completed after a reset.
    - After this fix, credit is still give for a contract submitted without an intervening reset.
    - All format, lint, and unit tests pass

### DIFF
--- a/src/Terminal/Terminal.ts
+++ b/src/Terminal/Terminal.ts
@@ -513,10 +513,10 @@ export class Terminal {
     const promptResult = await contract.prompt();
 
     // Get a new copy of the server, in case it changed while the prompt was open
-    const postPromptServer = Player.getCurrentServer();
+    const postPromptServer = GetServer(server.hostname);
 
     // Check if the contract still exists by the time the promise is fulfilled
-    if (postPromptServer == null || postPromptServer.getContract(contractPath) == null) {
+    if (postPromptServer?.getContract(contractPath) == null) {
       this.contractOpen = false;
       return this.error("Contract no longer exists (Was it solved by a script?)");
     }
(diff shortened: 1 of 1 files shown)


src/Terminal/Terminal.ts
```diff
@@ -512,8 +512,11 @@ export class Terminal {
     this.contractOpen = true;
     const promptResult = await contract.prompt();
 
-    //Check if the contract still exists by the time the promise is fulfilled
-    if (server.getContract(contractPath) == null) {
+    // Get a new copy of the server, in case it changed while the prompt was open
+    const postPromptServer = Player.getCurrentServer();
```
d0sboots: The player can legitimately have changed servers in the interim. It's not possible for the *actual* player to do it, since they're stuck in a modal, but scripts can.

You'll need to re-lookup the server by name, instead.

src/Terminal/Terminal.ts
```diff
@@ -512,8 +512,11 @@ export class Terminal {
...
+
+    // Check if the contract still exists by the time the promise is fulfilled
+    if (postPromptServer == null || postPromptServer.getContract(contractPath) == null) {
```
d0sboots: Optional: This can be shorter with `?.`

src/Terminal/Terminal.ts
TheCleric: Honestly don't know how this got in there. I specifically looked up `GetServer` to fetch by the host name. :laughing:

d0sboots:
If you're on the Discord, you can ping me to claim your honorary "Contributor" badge if you want.

PR #2285, opened 2025-08-07
UI: Close coding contract popup on prestige

catloversg:
On prestige, all servers (except home) are reset, so all coding contracts are removed. In #2280, we see that the popup still exists after prestige even when the contract was removed. This PR closes this popup on prestige. This is consistent with other UI elements; for example, we close all log windows after stopping all scripts.

https://github.com/user-attachments/assets/6ebe3da9-5e98-462a-92ef-fb70e7d83868

### DIFF
--- a/src/Prestige.ts
+++ b/src/Prestige.ts
@@ -30,6 +30,7 @@ import { calculateExp } from "./PersonObjects/formulas/skill";
 import { currentNodeMults } from "./BitNode/BitNodeMultipliers";
 import { canAccessBitNodeFeature } from "./BitNode/BitNodeUtils";
 import { pendingUIShareJobIds } from "./NetworkShare/Share";
+import { CodingContractEventEmitter } from "./CodingContract/CodingContractEventEmitter";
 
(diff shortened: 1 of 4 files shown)


src/Prestige.ts
```diff
@@ -52,6 +53,9 @@ export function prestigeAugmentation(): void {
   // We must kill all scripts before doing anything else.
   prestigeWorkerScripts();
 
+  // Close coding contract modal
+  CodingContractEventEmitter.emit({ type: "close" });
```
d0sboots: This function is just a grab-bag of stuff, but still it has a *slight* but of structure to it. I'd suggest moving this to right after LogBoxClearEvents.emit() (which is with other terminal stuff) as the most natural place.

src/Prestige.ts
```diff
@@ -195,6 +199,9 @@ export function prestigeSourceFile(isFlume: boolean): void {
...
+
```
d0sboots: Someday, we ought to refactor these so they can have more of a shared impl. It's a mess.

PR #2286, opened 2025-08-07
UI: Change status bar progress animation to be forward-only

vamo89:
# UI: Change status bar progress animation to be forward-only

Currently, when a skill is gained, and the percentage on a status bar goes from a high value to a lower one, the animation makes the bar decrease. It's somewhat ok IMO when it's 80-100% to 0-20%, as it happens on low XP gain, but when we start getting high XP, it becomes a bit random, just jumping between percentages.

Tech aspects:
I've created a new LinearForwardProgress element, that is visual only and easily replace (and uses) @mui/material/LinearProgress. It just identify the case when the progress is going down, and instead of animating directly, it first animates to 100% and then from 0 to the new value. 

I've attached below the difference from current to proposed change as gif and video. GIF is good to quick check without opening but the video has better framerate.

Current behavior:
Low XP gain:
current-slow
https://github.com/user-attachments/assets/25b00a47-5d0c-4253-85f2-63b960e748b4

High XP gain:
current-fast
https://github.com/user-attachments/assets/d3cee133-0899-4f20-8e22-f2a74a24692b

With this change:
Low XP gain:
forward-only-slow
https://github.com/user-attachments/assets/f23040cd-3912-43d1-8ef6-496aae6f0cb7

High XP gain:
forward-only-fast
https://github.com/user-attachments/assets/95527837-040a-4ba3-a254-65b1423334db

### DIFF
--- a/src/ui/React/LinearForwardProgress.tsx
+++ b/src/ui/React/LinearForwardProgress.tsx
@@ -8,10 +8,27 @@ interface LinearForwardProgressProps {
   barColor?: React.CSSProperties["color"];
 }
 
+// Round percentage to nearest 5% interval to limit keyframe definitions
+const roundToInterval = (value: number, interval = 5): number => {
+  return Math.round(value / interval) * interval;
+};
+
+// Generate predefined keyframes for forward animations
+const generateKeyframes = (startPercent: number, endPercent: number) => {
+  const startPos = startPercent - 100;
+  const endPos = endPercent - 100;
+
+  return {
+    "0%": { transform: `translateX(${startPos}%)` },
+    "50%": { transform: "translateX(0%)" },
+    "50.01%": { transform: "translateX(-100%)" },
+    "100%": { transform: `translateX(${endPos}%)` },
+  };
+};
+
 export const LinearForwardProgress = React.forwardRef<unknown, LinearForwardProgressProps>(
   function LinearForwardProgress({ value, backgroundColor, barColor, ...otherProps }, ref): React.ReactElement {
     const previousValueRef = useRef(value);
-    const animationKeyRef = useRef(0);
 
     const isDecreasing = value < previousValueRef.current;
     const previousValue = previousValueRef.current;
@@ -19,43 +36,24 @@ export const LinearForwardProgress = React.forwardRef<unknown, LinearForwardProg
     // Update ref for next render
     previousValueRef.current = value;
 
+    let animationStyles = {};
+
     if (isDecreasing) {
-      // Increment animation key to force new CSS animation
-      animationKeyRef.current += 1;
+      // Round values to 5% intervals to limit keyframe definitions
+      const roundedStartPercent = roundToInterval(previousValue);
+      const roundedEndPercent = roundToInterval(value);
 
-      // MUI LinearProgress uses translateX where:
-      // translateX(-100%) = 0% progress, translateX(0%) = 100% progress
-      const startPos = previousValue - 100;
-      const endPos = value - 100;
-      const animationName = `forwardAnim${animationKeyRef.current}`;
+      // Use rounded values for animation name to ensure reuse of keyframes
+      const animationName = `forwardAnim_${roundedStartPercent}_${roundedEndPercent}`;
 
-      // Calculate timing: proportional split between fill-to-100% and reset-to-final phases
-      const firstPhasePercent = ((100 - previousValue) / (100 - previousValue + value)).toFixed(2);
+      const keyframes = generateKeyframes(roundedStartPercent, roundedEndPercent);
 
-      return (
-        <LinearProgress
-          variant="determinate"
-          value={value}
-          ref={ref}
-          {...otherProps}
-          sx={{
-            backgroundColor,
-            "& .MuiLinearProgress-bar1Determinate": {
-              backgroundColor: barColor,
-              [`@keyframes ${animationName}`]: {
-                "0%": { transform: `translateX(${startPos}%)` },
-                [`${firstPhasePercent}%`]: { transform: "translateX(0%)" },
-                [`${firstPhasePercent + 0.01}%`]: { transform: "translateX(-100%)" },
-                "100%": { transform: `translateX(${endPos}%)` },
-              },
-              animation: `${animationName} 0.4s ease-out`,
-            },
-          }}
-        />
-      );
+      animationStyles = {
+        [`@keyframes ${animationName}`]: keyframes,
+        animation: `${animationName} 0.2s linear`,
+      };
     }
 
-    // Normal increasing progress - use default MUI transition
     return (
       <LinearProgress
         variant="determinate"
@@ -67,6 +65,7 @@ export const LinearForwardProgress = React.forwardRef<unknown, LinearForwardProg
           "& .MuiLinearProgress-bar1Determinate": {
             backgroundColor: barColor,
             transition: "transform 0.2s linear",
+            ...animationStyles,
           },
         }}
       />
--- a/src/ui/React/LinearForwardProgress.tsx
+++ b/src/ui/React/LinearForwardProgress.tsx
@@ -23,9 +23,14 @@ export const LinearForwardProgress = React.forwardRef<unknown, LinearForwardProg
       // Increment animation key to force new CSS animation
       animationKeyRef.current += 1;
 
+      // MUI LinearProgress uses translateX where:
+      // translateX(-100%) = 0% progress, translateX(0%) = 100% progress
       const startPos = previousValue - 100;
       const endPos = value - 100;
       const animationName = `forwardAnim${animationKeyRef.current}`;
+
+      // Calculate timing: proportional split between fill-to-100% and reset-to-final phases
+      const firstPhasePercent = ((100 - previousValue) / (100 - previousValue + value)).toFixed(2);
 
       return (
         <LinearProgress
@@ -39,8 +44,8 @@ export const LinearForwardProgress = React.forwardRef<unknown, LinearForwardProg
               backgroundColor: barColor,
               [`@keyframes ${animationName}`]: {
                 "0%": { transform: `translateX(${startPos}%)` },
-                "50%": { transform: "translateX(0%)" },
-                "50.01%": { transform: "translateX(-100%)" },
+                [`${firstPhasePercent}%`]: { transform: "translateX(0%)" },
+                [`${firstPhasePercent + 0.01}%`]: { transform: "translateX(-100%)" },
                 "100%": { transform: `translateX(${endPos}%)` },
               },
               animation: `${animationName} 0.4s ease-out`,
--- a/src/ui/React/LinearForwardProgress.tsx
+++ b/src/ui/React/LinearForwardProgress.tsx
@@ -9,7 +9,7 @@ interface LinearForwardProgressProps {
 }
 
 // Round percentage to nearest 5% interval to limit keyframe definitions
-const roundToInterval = (value: number, interval = 5): number => {
+const roundToInterval = (value: number, interval = 10): number => {
   return Math.round(value / interval) * interval;
 };
 
--- a/src/ui/React/LinearForwardProgress.tsx
+++ b/src/ui/React/LinearForwardProgress.tsx
@@ -8,97 +8,97 @@ interface LinearForwardProgressProps {
   barColor?: React.CSSProperties["color"];
 }
 
-export function LinearForwardProgress({
-  value,
-  backgroundColor,
-  barColor,
-}: LinearForwardProgressProps): React.ReactElement {
-  const TRANSITION_TO_FULL_DURATION = 200;
-  const ONE_FRAME_RESET_DELAY = 16;
+export const LinearForwardProgress = React.forwardRef<unknown, LinearForwardProgressProps>(
+  function LinearForwardProgress({ value, backgroundColor, barColor, ...otherProps }, ref): React.ReactElement {
+    const TRANSITION_TO_FULL_DURATION = 200;
+    const ONE_FRAME_RESET_DELAY = 16;
 
-  const [displayProgress, setDisplayProgress] = useState(value);
-  const [isInDecreasingAnimation, setIsInDecreasingAnimation] = useState(false);
-  const [isAnimatingToFull, setIsAnimatingToFull] = useState(false);
+    const [displayProgress, setDisplayProgress] = useState(value);
+    const [isInDecreasingAnimation, setIsInDecreasingAnimation] = useState(false);
+    const [isAnimatingToFull, setIsAnimatingToFull] = useState(false);
 
-  const previousProgressRef = useRef(value);
-  const animationTimeouts = useRef<number[]>([]);
+    const previousProgressRef = useRef(value);
+    const animationTimeouts = useRef<number[]>([]);
 
-  const clearAnimationTimeouts = useCallback(() => {
-    animationTimeouts.current.forEach((timeout) => {
-      if (timeout) clearTimeout(timeout);
-    });
-    animationTimeouts.current = [];
-  }, []);
+    const clearAnimationTimeouts = useCallback(() => {
+      animationTimeouts.current.forEach((timeout) => {
+        if (timeout) clearTimeout(timeout);
+      });
+      animationTimeouts.current = [];
+    }, []);
 
-  const startDecreasingAnimation = useCallback(() => {
-    setIsInDecreasingAnimation(true);
-    setIsAnimatingToFull(true);
+    const startDecreasingAnimation = useCallback(() => {
+      setIsInDecreasingAnimation(true);
+      setIsAnimatingToFull(true);
 
-    // Phase 1: Animate to 100% with linear transition
-    setDisplayProgress(100);
+      // Phase 1: Animate to 100% with linear transition
+      setDisplayProgress(100);
 
-    // Phase 2: After reaching 100%, reset to 0% and animate to new value
-    const fullTransitionTimeout = window.setTimeout(() => {
-      setIsAnimatingToFull(false);
-      setDisplayProgress(0); // Instant reset with no transition
+      // Phase 2: After reaching 100%, reset to 0% and animate to new value
+      const fullTransitionTimeout = window.setTimeout(() => {
+        setIsAnimatingToFull(false);
+        setDisplayProgress(0); // Instant reset with no transition
 
-      // Phase 3: Brief delay, then animate to final value
-      const resetTimeout = window.setTimeout(() => {
+        // Phase 3: Brief delay, then animate to final value
+        const resetTimeout = window.setTimeout(() => {
+          setDisplayProgress(value);
+          setIsInDecreasingAnimation(false);
+        }, ONE_FRAME_RESET_DELAY);
+
+        animationTimeouts.current[1] = resetTimeout;
+      }, TRANSITION_TO_FULL_DURATION);
+
+      animationTimeouts.current[0] = fullTransitionTimeout;
+    }, [value]);
+
+    useEffect(() => {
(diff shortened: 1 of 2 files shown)


src/ui/React/LinearForwardProgress.tsx
```diff
@@ -0,0 +1,104 @@
...
+  value: number;
+  backgroundColor?: React.CSSProperties["color"];
+  barColor?: React.CSSProperties["color"];
+}
+
+export function LinearForwardProgress({
+  value,
+  backgroundColor,
+  barColor,
+}: LinearForwardProgressProps): React.ReactElement {
```
catloversg: If you hover over the stat progress bar, you will see that the tooltip stops working. There are errors in the console:
```
Warning: Failed prop type: Invalid prop `children` supplied to `ForwardRef(Tooltip)`. Expected an element that can hold a ref. Did you accidentally use a plain function component for an element instead? For more information see https://mui.com/r/caveat-with-refs-guide

Warning: Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()?
```
In order to create a custom component that works with a tooltip, you have to follow its guideline: https://mui.com/material-ui/react-tooltip/#custom-child-element.

It should look like this:
```ts
export const LinearForwardProgress = React.forwardRef<unknown, LinearForwardProgressProps>(
  function LinearForwardProgress({ value, backgroundColor, barColor, ...otherProps }, ref): React.ReactElement {
    // Other code
    return (
      <LinearProgress
        // Other code
        {...otherProps}
        ref={ref}
      />
    );
  },
);

```

src/ui/React/LinearForwardProgress.tsx
```diff
@@ -0,0 +1,104 @@
...
+  const TRANSITION_TO_FULL_DURATION = 200;
+  const ONE_FRAME_RESET_DELAY = 16;
+
+  const [displayProgress, setDisplayProgress] = useState(value);
+  const [isInDecreasingAnimation, setIsInDecreasingAnimation] = useState(false);
+  const [isAnimatingToFull, setIsAnimatingToFull] = useState(false);
+
+  const previousProgressRef = useRef(value);
+  const animationTimeouts = useRef<(NodeJS.Timeout | null)[]>([]);
```
catloversg: This is the wrong type. I know you are trying to work around the false-positive error from TS, but this is the wrong way to do that. You should use `number` here as normal and use `window.setTimeout` instead of `setTimeout` in the following code.

src/ui/React/LinearForwardProgress.tsx
vamo89: Thank you for the review. Fixed.
image

src/ui/React/LinearForwardProgress.tsx
vamo89: Good catch, thank you! Fixed.

src/ui/React/LinearForwardProgress.tsx
```diff
@@ -0,0 +1,104 @@
...
+import LinearProgress from "@mui/material/LinearProgress";
+
+interface LinearForwardProgressProps {
+  value: number;
+  backgroundColor?: React.CSSProperties["color"];
+  barColor?: React.CSSProperties["color"];
+}
+
+export const LinearForwardProgress = React.forwardRef<unknown, LinearForwardProgressProps>(
+  function LinearForwardProgress({ value, backgroundColor, barColor, ...otherProps }, ref): React.ReactElement {
```
d0sboots: Does this need to be a named function? It's confusing (and likely error prone) to have both inner and outer named the same thing; the inner function can probably be an arrow function.

src/ui/React/StatsProgressBar.tsx
```diff
@@ -39,16 +39,7 @@ export function StatsProgressBar({
 
   return (
     <Tooltip title={tooltip}>
-      <LinearProgress
-        variant="determinate"
-        value={progress}
-        sx={{
```
d0sboots, line 45: Since you're attempting to make a generic(ish) component here, as opposed to an app-logic-specific one, I think it would be better to pass sx through and leave styling decisions in the hands of the user of the component.

src/ui/React/LinearForwardProgress.tsx
```diff
@@ -0,0 +1,104 @@
...
+    }, [value, isInDecreasingAnimation, clearAnimationTimeouts, startDecreasingAnimation]);
+
+    // Cleanup timeouts on component unmount
+    useEffect(() => {
+      return clearAnimationTimeouts;
+    }, [clearAnimationTimeouts]);
+
+    const getTransitionStyle = () => {
+      if (isAnimatingToFull) {
+        return { transition: `transform ${TRANSITION_TO_FULL_DURATION}ms linear` };
```
d0sboots: Why do you need a separate style and state for this, as opposed to letting the underlying component animate to full? This deserves an explanatory comment.

src/ui/React/LinearForwardProgress.tsx
```diff
@@ -0,0 +1,104 @@
...
+      // Phase 2: After reaching 100%, reset to 0% and animate to new value
+      const fullTransitionTimeout = window.setTimeout(() => {
+        setIsAnimatingToFull(false);
+        setDisplayProgress(0); // Instant reset with no transition
+
+        // Phase 3: Brief delay, then animate to final value
+        const resetTimeout = window.setTimeout(() => {
+          setDisplayProgress(value);
+          setIsInDecreasingAnimation(false);
+        }, ONE_FRAME_RESET_DELAY);
```
d0sboots: What happens if the framerate is chunky, and React coalesces this together with the next update? Are you certain that the transition change can never be lost? And if it can't be, I suspect the delay is actually irrelevant.

src/ui/React/LinearForwardProgress.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+        animationTimeouts.current[1] = resetTimeout;
```
d0sboots: You don't need to store specific timeouts in specific slots. This also saves an if in your clear up above.
```suggestion
        animationTimeouts.current.push(resetTimeout);
```

src/ui/React/LinearForwardProgress.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+    const [displayProgress, setDisplayProgress] = useState(value);
+    const [isInDecreasingAnimation, setIsInDecreasingAnimation] = useState(false);
+    const [isAnimatingToFull, setIsAnimatingToFull] = useState(false);
+
+    const previousProgressRef = useRef(value);
+    const animationTimeouts = useRef<number[]>([]);
+
+    const clearAnimationTimeouts = useCallback(() => {
+      animationTimeouts.current.forEach((timeout) => {
```
d0sboots:
```suggestion
      for (const timeout of animationTimeouts.current) {
```
Optional: forEach, specifically, has several gotchas such that I recommend for..of as the superior alternative.

d0sboots (commented):
This is a very exciting change, and there is no doubt that the visual effect is superior to what we have currently!

However, I have some concerns about the levels of complexity it is introducing. Specifically, before there was no state involved, because all animation state was handled by CSS/the browser. Now, we are dealing with the state, except we are *also* relying on the browser for some of it, and hoping the timing between the two matches up. This seems like a recipe for glitchiness to me.

Of greater concern is the possibility for edge cases in the nested setTimeout structure. A little animation glitchiness is ugly but tolerable, but if one of the bars got stuck that would be worse. My spider-sense is tingling that there may be issues involving multiple levelups with specific timing, even if I haven't found them. 

What do you think about an alternate possibility where the bar simply doesn't animate when it goes backwards, or if you gain more than one level at a time? That would be radically simpler (no real state), and hopefully solve the same underlying desire, although it doesn't look as cool.

catloversg:
The latest changes fix the problems mentioned in my review, so they are good.

Regarding the concern of complexity, I agree with d0sboots. I think you should consider his suggestion (`the bar simply doesn't animate when it goes backwards, or if you gain more than one level at a time`). It may not look as good as it does now, but it's much simpler.

vamo89:
@d0sboots @catloversg thank you for the review, it's a valid concern and I'll try new approaches focusing on stateless and css animation/transitions so we can have something performant and pleasant to the eye. I'll let you know when I have something.

src/ui/React/LinearForwardProgress.tsx
vamo89: I'll answer this one because it may still be relevant to the new implementation. The default animation does some ease-out, so it's weird when it starts slowing near 100% since it will continue from 0 to new value, that's why I changed it to linear here. If this makes a part on a new implementation I'll make sure to leave a comment.

vamo89:
I've updated the code to use CSS transitions only. @d0sboots @catloversg let me know your thoughts.

src/ui/React/LinearForwardProgress.tsx
d0sboots: Still relevant

src/ui/React/LinearForwardProgress.tsx
```diff
@@ -0,0 +1,75 @@
...
+            backgroundColor,
+            "& .MuiLinearProgress-bar1Determinate": {
+              backgroundColor: barColor,
+              [`@keyframes ${animationName}`]: {
+                "0%": { transform: `translateX(${startPos}%)` },
+                [`${firstPhasePercent}%`]: { transform: "translateX(0%)" },
+                [`${firstPhasePercent + 0.01}%`]: { transform: "translateX(-100%)" },
+                "100%": { transform: `translateX(${endPos}%)` },
+              },
+              animation: `${animationName} 0.4s ease-out`,
```
d0sboots: Now that this is all in CSS, why are the animation properties for wrap-around different?

src/ui/React/LinearForwardProgress.tsx
```diff
@@ -0,0 +1,75 @@
...
+            },
+          }}
+        />
+      );
+    }
+
+    // Normal increasing progress - use default MUI transition
+    return (
```
d0sboots, line 57: The only thing that is different here is the animation keyframes, and (possibly) the transition. Rather than duplicating this definition, which can lead to subtle divergence, I would define the object `animationOverrides` which is normally empty. Then you can add the keyframes to it in your if block, and expand it with `...animationOverrides` inside the style.

src/ui/React/StatsProgressBar.tsx
d0sboots: This is still relevant, although it would be a bit annoying to do so it's not necessary.

src/ui/React/LinearForwardProgress.tsx
```diff
@@ -0,0 +1,75 @@
...
+      animationKeyRef.current += 1;
+
+      // MUI LinearProgress uses translateX where:
+      // translateX(-100%) = 0% progress, translateX(0%) = 100% progress
+      const startPos = previousValue - 100;
+      const endPos = value - 100;
+      const animationName = `forwardAnim${animationKeyRef.current}`;
+
+      // Calculate timing: proportional split between fill-to-100% and reset-to-final phases
+      const firstPhasePercent = ((100 - previousValue) / (100 - previousValue + value)).toFixed(2);
```
d0sboots: This calculates a fraction, not a percent. When you call toFixed, you will end up with a number like "0.23", which means a negligible time will be spent on the first phase. Rather than toFixed, I think you should just multiply by 100.

d0sboots:
Oh, I forgot to mention: Although there's still some complication, the use of CSS animation here means I'm not concerned that this will get stuck in weird states. Good job.

catloversg (commented):
With dynamic values, you should not use them with `sx`. For more information, please check:
- https://mui.com/system/getting-started/the-sx-prop/#dynamic-values
- https://github.com/mui/material-ui/issues/42164

Capture

You should use `style` instead of `sx`, but it may be fairly hard because you want set `& .MuiLinearProgress-bar1Determinate` which is usually done with `sx`. You have to find a solution or at least a workaround here.

src/ui/React/LinearForwardProgress.tsx
d0sboots: Also, you messed up the order here. For transition it is "propertyName duration easing", but for animation it is "easing duration animationName". I'm surprised this worked at all...

d0sboots:
There is a deeper issue here with the sx thing that greatly complicated this approach. 
CSS animations are defined with @keyframe at-rules, which are top-level CSS entities. In other words, the memory leak here is fundamental to the approach, it's not just a matter of defining inline styles differently. There would need to be extra stuff to remove these rules once the animation is run (a timer would probably be good enough).

...Hmm, but thinking on it more, given React's machinery surely it must have a map from sx to style. Otherwise, switching back and forth between two different sx values would cause a leak as it kept creating new styles for them. This means if we use the toFixed trick to keep the number of unique sx combinations down to 100, there won't be an issue. The only remaining piece is to name the animation by its firstPhasePercent, and the rest should work itself out.

src/ui/React/LinearForwardProgress.tsx
vamo89: I realized that if I do that I'll need to include this at the end of the file:
`LinearForwardProgress.displayName = "LinearForwardProgress";`
Do you think it's worth?
The only place I found using the same structure, to follow the codebase pattern, was [SidebarRoot.tsx](https://github.com/bitburner-official/bitburner-src/blob/18f84396e2cbb998d6e182b0a38353b76663b884/src/Sidebar/ui/SidebarRoot.tsx) and it's the same I'm using here.

src/ui/React/LinearForwardProgress.tsx
vamo89: Thank you, I was using more variables here and cleaned up in the end and messed up this.

src/ui/React/StatsProgressBar.tsx
vamo89: Although I understand why you are suggesting that, to keep the same props as LinearProgress, since this is a component made specifically for this codebase, I don't think it's worth to sacrifice the ease of reading only the props needed to keep that compatibility.

vamo89:
@d0sboots @catloversg thank you again for the time to re-review this.
I had some time at hand to do the latest changes but not enough to also leave some more info here and reply to previous comments, so I'm doing that now.

I tried a lot of approaches before getting to the point I shared here, and honestly the last commit where I tried to make keyframes considering the percentages was a bit rushed and it has some problems I just caught today when I played with it a bit more.

I'm seriously considering the idea you guys said at first, to just not animate when it's decreasing, although I'm not sure this is better than the current behavior. It's more correct but maybe even less fun to look at.

I was thinking something else though and reading this comment gave me the strength to test it:
> This means if we use the toFixed trick to keep the number of unique sx combinations down to 100, there won't be an issue. The only remaining piece is to name the animation by its firstPhasePercent, and the rest should work itself out.

My idea was to even reduce this, I don't think we need percent-precision on that, having them at each 5 or even 10% should do the trick - so we just need 10-20 style definitions.

src/ui/React/LinearForwardProgress.tsx
vamo89: Did not think of that, will be better for maintenance for sure.

src/ui/React/LinearForwardProgress.tsx
vamo89: I messed up, this is a reminiscent of a previous try that needed the animation time on that to be double the original, that's not the case anymore.

d0sboots:
> My idea was to even reduce this, I don't think we need percent-precision on that, having them at each 5 or even 10% should do the trick - so we just need 10-20 style definitions.

I also had the same thought in the back of my head, glad we've on the same page!

src/ui/React/LinearForwardProgress.tsx
d0sboots: That's a fascinating bit of weirdness. Alright, this seems like the best way.

vamo89:
@d0sboots @catloversg I believe we have a good option now. It creates the keyframe definitions with values in interval of 10, on both start and finish values.
If my calculations are correct, this will result in 55 different keyframe definitions (interval of 5 would give 210 different ones).
I've tested it for a while and it seems good, both for start and late game. 
Let me know your thoughts, I'm excited about this one!

src/ui/React/StatsProgressBar.tsx
vamo89: @d0sboots re-checking all the comments to make sure I did not leave anything back. Are you ok with this approach?

src/ui/React/LinearForwardProgress.tsx
vamo89: @d0sboots I reordered and saw issues, then I checked and this is the correct order for animations as well:
`animation: name duration timing-function delay iteration-count direction`
https://www.w3schools.com/css/css3_animations.asp

d0sboots: As is semi-frequently the case, w3schools is wrong, but it will function in most cases. https://drafts.csswg.org/css-animations-1/#animation

Any property that can be parsed as a valid value for a property, will be, until those are exhausted. This means if you called your animation "ease-in" for instance, with your ordering that would be interpreted as the timing function. But I'm the common case where the name is not valid for the other properties, it will work.

src/ui/React/StatsProgressBar.tsx
d0sboots: It's fine.

src/ui/React/LinearForwardProgress.tsx
```diff
@@ -0,0 +1,74 @@
...
+      <LinearProgress
+        variant="determinate"
+        value={value}
+        ref={ref}
+        {...otherProps}
+        sx={{
+          backgroundColor,
+          "& .MuiLinearProgress-bar1Determinate": {
+            backgroundColor: barColor,
+            transition: "transform 0.2s linear",
```
d0sboots, line 67: This should be part of animationStyles, but in the other branch. We don't want transition if animation is happening.

d0sboots (commented):
Looks good!

src/ui/React/LinearForwardProgress.tsx
```diff
@@ -0,0 +1,74 @@
...
+};
+
+// Generate predefined keyframes for forward animations
+const generateKeyframes = (startPercent: number, endPercent: number) => {
+  const startPos = startPercent - 100;
+  const endPos = endPercent - 100;
+
+  return {
+    "0%": { transform: `translateX(${startPos}%)` },
+    "50%": { transform: "translateX(0%)" },
```
d0sboots, line 23: I just realized, you changed how this is working and the animation doesn't have constant speed anymore. It always spends half the time getting too the end and half the time coming from the beginning.

src/ui/React/LinearForwardProgress.tsx
```diff
@@ -8,54 +8,52 @@ interface LinearForwardProgressProps {
...
     // Update ref for next render
     previousValueRef.current = value;
 
+    let animationStyles = {};
+
     if (isDecreasing) {
-      // Increment animation key to force new CSS animation
-      animationKeyRef.current += 1;
+      // Round values to 5% intervals to limit keyframe definitions
+      const roundedStartPercent = roundToInterval(previousValue);
```
d0sboots, line 43: You might want to vary the rounding direction to prevent backwards motion. 

For instance, if endPercent is 5.1, that will get rounded up to 10, and then the next update might go backwards. The same thing can happen with startPervent.

d0sboots:
I submitted the alternate implementation, so I am closing this PR. However, if you are on the Discord I'm happy to give you contributor status for this, since you put in 90% of the effort on it.

vamo89:
@d0sboots I'm happy it worked out!
I was out of time for a while, but I intend to keep working on some improvements on my free time, and yes, I'm on discord.

PR #2287, opened 2025-08-08
EDITOR: Add the ability to Beautify on Save (optionally)

### LINKED ISSUE #2284: Feature Request: Option to Beautify a script on Save
I frequently find myself clicking Beautify and then clicking Save when editing a script. It would be nice to have on option to enable the Save button to run Beautify and then Save.

TheCleric:
# Linked issues

closes #2284 

Just wanted to get this in front of you folks to take a look at. No rush as I know we're still soliciting feedback. I have made the default on this feature to be off for minimal adverse impact/surprise.

### DIFF
--- a/src/ScriptEditor/ui/ScriptEditorRoot.tsx
+++ b/src/ScriptEditor/ui/ScriptEditorRoot.tsx
@@ -218,11 +218,10 @@ function Root(props: IProps): React.ReactElement {
       iTutorialNextStep();
 
       return;
-    } else {
-      await preSave();
-      saveScript(currentScript);
-      rerender();
     }
+    await preSave();
+    saveScript(currentScript);
+    rerender();
   }, [rerender, options.beautifyOnSave, beautify]);
 
   const run = useCallback(async () => {
--- a/src/ScriptEditor/ui/ScriptEditorRoot.tsx
+++ b/src/ScriptEditor/ui/ScriptEditorRoot.tsx
@@ -614,7 +614,7 @@ function Root(props: IProps): React.ReactElement {
 
         {statusBarRef.current}
 
-        <Toolbar onSave={save} onRun={run} editor={editorRef.current} beautify={beautify} />
+        <Toolbar onSave={save} onRun={run} editor={editorRef.current} onBeautify={beautify} />
       </div>
       {!currentScript && <NoOpenScripts />}
     </>
--- a/src/ScriptEditor/ui/ScriptEditorRoot.tsx
+++ b/src/ScriptEditor/ui/ScriptEditorRoot.tsx
@@ -180,21 +180,12 @@ function Root(props: IProps): React.ReactElement {
 
   let decorations: monaco.editor.IEditorDecorationsCollection | undefined;
 
-  const beautify = useCallback((): Promise<void> => {
-    return new Promise((resolve) => {
-      const beautifyPromise = editorRef.current
-        ?.getAction("editor.action.formatDocument")
-        ?.run()
-        .catch((error) => {
-          console.error(error);
-          resolve();
-        })
-        .finally(() => resolve());
-
-      if (beautifyPromise === undefined) {
-        resolve();
-      }
-    });
+  const beautify = useCallback(async (): Promise<void> => {
+    const action = editorRef.current?.getAction("editor.action.formatDocument");
+    if (action == null) {
+      return;
+    }
+    return action.run().catch((error) => console.error(error));
   }, []);
 
   const save = useCallback(() => {
--- a/src/ScriptEditor/ui/ScriptEditorRoot.tsx
+++ b/src/ScriptEditor/ui/ScriptEditorRoot.tsx
@@ -176,15 +176,35 @@ function Root(props: IProps): React.ReactElement {
     reloadModelOfCurrentScript();
   }
 
-  const { showRAMError, updateRAM, startUpdatingRAM, finishUpdatingRAM } = useScriptEditorContext();
+  const { options, showRAMError, updateRAM, startUpdatingRAM, finishUpdatingRAM } = useScriptEditorContext();
 
   let decorations: monaco.editor.IEditorDecorationsCollection | undefined;
+
+  const beautify = useCallback((): Promise<void> => {
+    return new Promise((resolve) => {
+      const beautifyPromise = editorRef.current
+        ?.getAction("editor.action.formatDocument")
+        ?.run()
+        .catch((error) => {
+          console.error(error);
+          resolve();
+        })
+        .finally(() => resolve());
+
+      if (beautifyPromise === undefined) {
+        resolve();
+      }
+    });
+  }, []);
 
   const save = useCallback(() => {
     if (currentScript === null) {
       console.error("currentScript is null when it shouldn't be. Unable to save script");
       return;
     }
+
+    const preSave = options.beautifyOnSave ? beautify : () => Promise.resolve();
+
     // this is duplicate code with saving later.
     if (ITutorial.isRunning && ITutorial.currStep === iTutorialSteps.TerminalEditScript) {
       //Make sure filename + code properly follow tutorial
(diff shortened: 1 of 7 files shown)


src/ScriptEditor/ui/ScriptEditorRoot.tsx
```diff
@@ -176,15 +176,35 @@ function Root(props: IProps): React.ReactElement {
...
+          console.error(error);
+          resolve();
+        })
+        .finally(() => resolve());
+
+      if (beautifyPromise === undefined) {
+        resolve();
+      }
+    });
+  }, []);
```
catloversg:
```suggestion
  const beautify = useCallback(async (): Promise<void> => {
    const action = editorRef.current?.getAction("editor.action.formatDocument");
    if (action == null) {
      return;
    }
    return action.run().catch((error) => console.error(error));
  }, []);
```

src/ScriptEditor/ui/ScriptEditorRoot.tsx
```diff
@@ -200,16 +220,23 @@ function Root(props: IProps): React.ReactElement {
...
       //Save the script
-      saveScript(currentScript);
-      Router.toPage(Page.Terminal);
-
-      iTutorialNextStep();
-
-      return;
+      preSave()
+        .then(() => {
+          saveScript(currentScript as OpenScript);
```
catloversg: In general, if you use typecasting, you have to explain why it's needed. Typecasting is rarely used in our codebase. In this case, you should not do that. Technically, `currentScript` may be null (it depends on `preSave`'s implementation), and you have to check it again.

src/ScriptEditor/ui/ScriptEditorRoot.tsx
```diff
@@ -200,16 +220,23 @@ function Root(props: IProps): React.ReactElement {
...
+          Router.toPage(Page.Terminal);
+
+          iTutorialNextStep();
+        })
+        .catch((error) => console.error(error));
+    } else {
+      preSave()
+        .then(() => {
+          saveScript(currentScript as OpenScript);
```
catloversg: Same here.

src/ScriptEditor/ui/Toolbar.tsx
```diff
@@ -33,19 +33,13 @@ interface IProps {
   editor: IStandaloneCodeEditor | null;
   onSave: () => void;
   onRun: () => void;
+  beautify: () => Promise<void>;
```
catloversg: Optional: Rename this function to `onBeautify` to be consistent with other functions.

catloversg:
Bug: When you run an edited script, it runs the unmodified version. For example:
```js
/** @param {NS} ns */
  ns.ui.openTail();
  ns.clearLog();
  ns.print("1");
}
```
- Create `test.js` with test code. Run `test.js`.
- Change "1" to "2". Do NOT save it.
- Press "Run".

Edit: This is the perfect example of the async meme:

async-meme

Fixing this specific bug is easy, but the underlying problem is still the same. Current code assumes that the save action is sync. When you add the async "Beautify" action, the save action also becomes async, and it creates a cascade effect that affects many things. For example, without a proper fix, if you use `:wq` in the vim mode, when the UI returns to the Terminal tab, the script may not be saved yet.

TheCleric:
> Fixing this specific bug is easy, but the underlying problem is still the same. Current code assumes that the save action is sync. When you add the async "Beautify" action, the save action also becomes async, and it creates a cascade effect that affects many things. For example, without a proper fix, if you use `:wq` in the vim mode, when the UI returns to the Terminal tab, the script may not be saved yet.

Yeah I completely get this. Hence why I had to manhandle so much save logic and hoist beautify up. Since Monaco's formatter is inherently async:

* Do we bit the bullet and make everything assume save is async and color all the other functions as we need to?
* Do we figure out a different way to do this? Maybe leveraging Prettier instead of Monaco's (limited) formatter? (Prettier has a synchronous version here: https://github.com/prettier/prettier-synchronized)
* Do we say this isn't worth it?
* Double save. Save immediately, then beautify, then save again.

My vote would probably be 4, but I could see an argument for any of these.

EDIT: I thought of another option

catloversg:
> Do we bit the bullet and make everything assume save is async and color all the other functions as we need to?

I think this is a relatively safe way to fix it. With the linter, when we let async "spreads" everywhere, we won't have to worry about missing some edge cases.

> Do we figure out a different way to do this? Maybe leveraging Prettier instead of Monaco's (limited) formatter? (Prettier has a synchronous version here: https://github.com/prettier/prettier-synchronized)

I don't think `prettier-synchronized` works in a browser (https://github.com/prettier/prettier-synchronized/issues/15). We can use `js-beautify` (https://github.com/beautifier/js-beautify), but importing another (non-trivial) library should be the last option and only when it's needed for a really important feature.

> Do we say this isn't worth it?

This is a good feature. The trade-off is debatable, though. Let's wait for d0sboots.

> Double save. Save immediately, then beautify, then save again.

I think this is only a workaround, and it's not even a good one. It may create subtle bugs, especially when combining with the "Run" action and the vim mode (this mode already has known bugs).

TheCleric:
> > Do we bit the bullet and make everything assume save is async and color all the other functions as we need to?
> 
> I think this is a relatively safe way to fix it. With the linter, when we let async "spreads" everywhere, we won't have to worry about missing some edge cases.
> 
> > Do we figure out a different way to do this? Maybe leveraging Prettier instead of Monaco's (limited) formatter? (Prettier has a synchronous version here: https://github.com/prettier/prettier-synchronized)
> 
> I don't think `prettier-synchronized` works in a browser ([prettier/prettier-synchronized#15](https://github.com/prettier/prettier-synchronized/issues/15)). We can use `js-beautify` (https://github.com/beautifier/js-beautify), but importing another (non-trivial) library should be the last option and only when it's needed for a really important feature.

Yeah agreed, not the best idea to bloat the app just for a default disabled option.

> > Do we say this isn't worth it?
> 
> This is a good feature. The trade-off is debatable, though. Let's wait for d0sboots.

👍 

> > Double save. Save immediately, then beautify, then save again.
> 
> I think this is only a workaround, and it's not even a good one. It may create subtle bugs, especially when combining with the "Run" action and the vim mode (this mode already has known bugs).

You're very much correct. I started playing around with this and while I got it to work it involved putting the `unMount` code in Editor in a `setTimeout` to give the promise a chance to finish before it's disposed of. And that's the only way I got it working and it felt gross. Personally I wouldn't want that code with my name on it. 😆

TheCleric:
> > > Do we bit the bullet and make everything assume save is async and color all the other functions as we need to?
> > 
> > 
> > I think this is a relatively safe way to fix it. With the linter, when we let async "spreads" everywhere, we won't have to worry about missing some edge cases.

So I started exploring this path and hit a few dead ends where async can't spread and it causes issues.

1. The event handlers for the VIM extension. These will only take sync calls and while I've got the best I can in there I think it's still causing issues.
2. The unmount/dispose functions in the `Editor.tsx` is only synchronous, so I'm definitely encountering race conditions where a promise is trying to resolve and either gets cancelled or ends up triggering a rerender for a component that is unmounted/unmounting.

Tthere's probably a proper way to do it, but it seems to be escaping me.

TheCleric:
These look a lot like the changes I had in place, which for me lead to this error appearing on screen when doing the `:wq` in vim mode.

EDIT: To be clear I am seeing this error with these changes as well. 

```
ERROR
Canceled
CancellationError@
cancel@
dispose@
dispose@
clear@
dispose@
dispose@
@
_deliver@
_deliverQueue@
fire@
setModel@
_deliver@
_deliverQueue@
fire@
dispose@
forEach@[native code]
@
callCallback@
dispatchEvent@[native code]
invokeGuardedCallbackDev@
invokeGuardedCallback@
flushPassiveEffectsImpl@
unstable_runWithPriority@
performSyncWorkOnRoot@
@
unstable_runWithPriority@
flushSyncCallbackQueueImpl@
flushSyncCallbackQueue@
scheduleUpdateOnFiber@
dispatchAction@
toPage@
@
```

catloversg:
Every time I see that `Canceled` error, it's always a build error. Please delete your `node_modules` and run `npm i` again. I don't have this problem because I never use `npm run start:dev`.

Ref: https://discord.com/channels/415207508303544321/459097632896188436/1312780251046084618

TheCleric:
> Every time I see that `Canceled` error, it's always a build error. Please delete your `node_modules` and run `npm i` again. I don't have this problem because I never use `npm run start:dev`.
> 
> Ref: https://discord.com/channels/415207508303544321/459097632896188436/1312780251046084618

Wish I would have known this yesterday. 😆 

I even tried writing a convoluted promise manager to keep track of them to make sure they resolve before we unmount. It also did not work, lol.

d0sboots:
> This is a good feature. The trade-off is debatable, though. Let's wait for d0sboots.

I've experienced the pain of the infectious async, as well. I agree that this is a good feature. Basically, I think it comes down to "how much do you want to beat your head on this?" XD
If you can find a way to get it to work, I think we can support that, but I fear there's a good chance the string will just end up connecting to more string that connects to a plug at the bottom of the ocean...

catloversg:
I committed a fix in a previous commit. I'm waiting for your review.

TheCleric:
@d0sboots let me know if there's anything here I can answer any questions about.

d0sboots:
Sorry for being slow on this. The change looks good. I always have a healthy dose of fear with all Monaco-related changes, especially when changing a bunch of things to async, but let's see how it goes...

PR #2288, opened 2025-08-10
DOC: Update 'Hacking Algorithms' page

gmcew:
[Recently](https://discord.com/channels/415207508303544321/944647347625930762/1403658952075509760), the observation has been made that the Hacking Algorithms documentation page is outdated. The content hasn't been materially updated since Jan 2022 when Hydroflame first wrote it, but there have been significant game developments since.

I think there are a few gaps in the current documentation:
 - `ns.sleep` is referred to for delays, and the associated timing issues. This is leftover from pre-`additionalMSec` implementation and directs the player towards the stalefish method, rather than more contemporary techniques.
 - The terminology is outdated. For example, the documentation refers to three types of hacking algorithm: "self-contained, loop and batch". These are not common phrases in the community any more - arguably three "hacking algorithm" types would likely be described as EHT, shotgun, and JIT - which can lead to confusion when a new player is exposed to a vastly different vocabulary.
 - A large portion of the FAQs in the discord link to ancillary functions or problems, e.g. backpack problem for batch scheduling, effective scoring of targets, hwgw vs hgw, how much to hack per batch, efficient server preparation, etc. These aren't covered in the documentation in much detail if at all.

DarkTechnomancer's guide is an excellent introduction to batching, complete with worked examples, and I don't want to reinvent that wheel or plagiarise any hard work.

Instead I propose to summarise some of the content in the #batching pins and FAQ answers into a very simple description of the current state of the art by rearranging and refocussing the current documentation:

 - A section on batching pre-requisites: (i.e. server crawler, scoring function philosophy)
 - A EHT summary/clean-up
 - A primer on using controller scripts
 - A shotgun summary
 - A JIT summary

As much of the original wording as I felt I could retain, I have. I have also deliberately not gone into any real detail on the shotgun or JIT implementation for two reasons - it would take a lot of writing (and risks recreating DarkTechnomancer's tutorials) and potentially robs the reader of discovering techniques themselves. The same is true for FAQs I subjectively consider secondary issues that can be learnt through experimentation when making a batcher; when HGW is better, handling the backpack problem, use of ports, or how to prep efficiently.

This is in draft since I'm by no means a batching expert and appreciate feedback on if this is even a net improvement on current wording. I welcome suggestions for improvements and clarifications.

### DIFF
--- a/src/Documentation/doc/en/programming/hackingalgorithms.md
+++ b/src/Documentation/doc/en/programming/hackingalgorithms.md
@@ -118,6 +118,7 @@ This information can come from `formulas.exe` and use of the `getPlayer` and `ge
 
 Don't forget to explore the [NS API](github.com/bitburner-official/bitburner-src/blob/stable/markdown/bitburner.ns.md) for other functions that might be useful in these calculations!
 
+### Proto-batcher
 There are many ways to launch batches. The simplest way is to launch them sequentially like this:
 
     |=Batch=||=Batch=||=Batch=||=Batch=||=Batch=||=Batch=|
--- a/src/Documentation/doc/en/programming/hackingalgorithms.md
+++ b/src/Documentation/doc/en/programming/hackingalgorithms.md
@@ -1,22 +1,20 @@
 # Hacking algorithms
 
 The [Beginner's Guide](../help/getting_started.md) introduces how the [hack, grow and weaken](../basic/hacking.md) functions can be used together.
-Following it also helps you create the `early-hack-template.js`, which is one of the algorithms described in this section.
-There are three main types of hacking algorithms, or "batchers".
-There are many ways to implement each type. This guide will go over each of them and advise on how they can be implemented.
+Following it also helps you create the `early-hack-template.js`, which is one of the algorithms described in this section. There are many ways to improve this script. This guide will go over each of them and advise on how they can be implemented.
 
 ### Common Fundamentals
 
-Some issues must be resolved before attempting any batching algorithm.
+Some issues must be resolved before attempting any algorithms.
 
-- Copying and running your chosen batching script from multiple different servers.
+- Copying scripts to different servers and running them there.
   - You can use the `scan` function recursively to build a list of servers. Using the `scp` and `exec` functions allows you to run your scripts remotely. The [Beginner's Guide](../help/getting_started.md) has an example script without using `scan`.
 - Finding and applying a scoring method to potential targets to find the best one.
   - The general approach is to compare the maximum money that could be stored on the server against the minimum security of the server. This ratio is an approximate money per time score for that server.
     Scoring can also incorporate other factors, such as the number of threads or amount of RAM required to take that money, or how close the server is to maximum money and minimum security.
   - It is often time effective to target a server with a hacking level requirement of less than half of your current hacking level.
 
-## Self-contained algorithms
+## Early hacking template (EHT)
 
 **Difficulty**: Easy
 
@@ -34,9 +32,8 @@ Cons:
 - Risk of over hacking
 - Slow to start producing money
 
-Self-contained algorithms are the simplest family of hacking algorithms to implement. `early-hack-template.js`, or "EHT" is a self-contained algorithm.
-Each script is tasked with choosing which function to execute based on the status of the target server.
-Because of this, they guarantee a consistent, but relatively small, flow of money.
+This is the self-contained algorithm implemented in `early-hack-template.js`. Each script is tasked with choosing which function to execute based on the status of the target server.
+Because of this, they guarantee a consistent, but relatively small flow of money.
 
 The general logic goes like this:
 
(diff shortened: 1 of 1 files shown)


src/Documentation/doc/en/programming/hackingalgorithms.md
```diff
@@ -36,96 +50,134 @@ The general logic goes like this:
...
 
 Cons:
 
-- Very difficult to implement without prior programming knowledge
-- Very difficult to make work on servers with less than 1TB of RAM
+- Requires a dedicated controller script
+- Can be real-life hardware heavy
 
-Batch algorithms utilize a master script that uses `exec` many scripts which utilize a relevant hacking function in batches.
+A shotgun batcher implements uses a controller script to launch as many batches in parallel as possible.
```
d0sboots:
```suggestion
A shotgun batcher uses a controller script to launch as many batches in parallel as possible.
```

src/Documentation/doc/en/programming/hackingalgorithms.md
```diff
@@ -36,96 +50,134 @@ The general logic goes like this:
...
-  Grow should finish when security is minimum, shortly after a hack occurred.
-  Weaken should occur when security is not at a minimum due to a hack or grow increasing it.
+    |=Batch=|
+    |=Batch=|
+    |=Batch=|
+    |=Batch=|
+    |=Batch=|
+    |=Batch=|
+
+Launching potentially thousands of batches at the same time is intensive on real-life hardware. It's recommended to limit the number of parallel batches to around 100 000 to reduce the risk of the game crashing.
```
d0sboots: Added a little detail here, since the steam app doesn't crash, i.e. it's slightly less scary than it could be.
```suggestion
Launching potentially thousands of batches at the same time is intensive on real-life hardware. It's recommended to limit the number of parallel batches to around 100 000 to reduce the risk of the game soft-crashing. (Also called a "black screen" because of how it appears when this happens.)
```

d0sboots:
Looks good. It covers the basics, and IMO that's all that's needed. More importantly, at this point I'd take anything that wasn't wrong XD

gmcew:
Thanks for the proof-read! I'll mark it as ready since I was expecting more rework to be honest - if you want to hold the merge until there's more feedback I'm fine with that.

d0sboots:
Catlover tagged themself so I'll hold off for them.

catloversg:
After reading the new content, I think we should rephrase and restructure some parts. At first, let's take a look at the structure of this guide (old vs new).

The old one:
```
# Hacking algorithms
## Self-contained algorithms
## Loop algorithms
## Batch algorithms (HGW, HWGW, or Cycles)
```

The new one:
```
# Hacking algorithms
### Common Fundamentals
## Self-contained algorithms
### Batch Controllers
## Shotgun Batchers
## Just-In-Time ("JIT") Batchers
```

The old guide defines "three primary families of hacking algorithms" in 3 sections. Here is my interpretation:
- "Self-contained algorithms": EHT
- "Loop algorithms": Generic controller
- "Batch algorithms (HGW, HWGW, or Cycles)": Batching

After EHT, the next step is to build a controller that spawns workers on demand. Note that this controller is not tied to the "batching" concept. The old guide described a generic controller. Technically, you can write a generic non-batching controller without diving into the batching's challenges. This is also what I wrote a long time ago, back when I was a player that did not join the Discord community.

Although the specific detail is outdated, the old guide still has a proper structure. You have EHT, then you write a generic controller, then you modify that controller and make it spawn batches (i.e., a set of HWG with careful timing). These are 3 types of hacking algorithms; you can also see them as 3 consecutive steps.

The new guide outlines 3 "types": EHT, shotgun, and JIT. Shotgun and JIT are the most discussed batchers on Discord, so it's good to introduce them as modern and practical approaches to hacking algorithms. This is not wrong, but I want to suggest some changes to how we should do that. I committed these suggestions in the last commit. If you disagree, you can revert them.

Suggestions:
- Reuse the structure of the old guide.
- Explain the most important thing about batching in the new "Concept" part (`The idea of "batching" is that, instead of waiting for a long time to prep the server again and again after hacking, you make the server go back to the prepped state instantly or within a few milliseconds by carefully timing actions (i.e., hacking/growing/weakening). A group of hacking/growing/weakening is called a "batch".`). IMO, this is the core idea of batching, and it should be explained a bit more explicitly.
- We are showing HWGW batch. We should define this term and mention other approaches (e.g., HGW batch).
- Move proto-batcher to the "Concept" part and define that term.
- Add visualization of a batch with `additionalMsec` in shotgun batcher.

catloversg (approved):

d0sboots:
> Suggestions:
> 
> * Reuse the structure of the old guide.

I disagree with this one. "Loop algorithms" are not an interesting concept on their own, in large part because they cannot be automated well. The true breakthrough in understanding is the batch concept, and everything else (worker scripts, use of servers as distributed RAM) comes as natural improvements from that.

  * After review, I see you didn't really change the structure here much, so I don't really object.

> * Explain the most important thing about batching in the new "Concept" part (`The idea of "batching" is that, instead of waiting for a long time to prep the server again and again after hacking, you make the server go back to the prepped state instantly or within a few milliseconds by carefully timing actions (i.e., hacking/growing/weakening). A group of hacking/growing/weakening is called a "batch".`). IMO, this is the core idea of batching, and it should be explained a bit more explicitly.

Agree.
> * We are showing HWGW batch. We should define this term and mention other approaches (e.g., HGW batch).

This is something for the channel IMO. The difference between HWGW and HGW or GHW is marginal in most circumstances. HWGW makes more intuitive sense, and if we try to cover *all* of batching here this would become as big as the Corp guide.

Edit: However, it's not a very large section, so again I don't really object.

> * Move proto-batcher to the "Concept" part and define that term.

Agree.

> * Add visualization of a batch with `additionalMsec` in shotgun batcher.

Agree, the timing visualizations from the channel are IMO one of the most helpful resources we have for "getting" how batching works.

src/Documentation/doc/en/programming/hackingalgorithms.md
```diff
@@ -36,96 +47,146 @@ The general logic goes like this:
...
+
+In older versions of the game, this required using `sleep` or `asleep` functions, which have a RAM cost of zero.
+However, due to JavaScript limitations, the delay duration is not millisecond-precise and can cause the functions to finish out of order.
+Instead, the hack, grow and weaken functions have a special [option](https://github.com/bitburner-official/bitburner-src/blob/stable/markdown/bitburner.basichgwoptions.md) called `additionalMsec` that allows more precise delays.
+
+As well as the run time for each function, we also need information on the impact of a hack, grow or weaken thread on the target's security and/or money to optimise the thread ratios between the functions.
+This information can come from `formulas.exe` and use of the `getPlayer` and `getServer` functions, but cheaper functions such as `hackAnalyze`, `hackAnalyzeSecurity`, `getHackTime`, `growthAnalyze` and `growthAnalyzeSecurity` can still be used.
+
+Don't forget to explore the [NS API](github.com/bitburner-official/bitburner-src/blob/stable/markdown/bitburner.ns.md) for other functions that might be useful in these calculations!
+
```
d0sboots, line 120:
```suggestion

### Proto-batcher
```

PR #2289, opened 2025-08-13
DOCUMENTATION: Clarify ns.gang.getOtherGangInformation

acidduk:
Improve Gang.getOtherGangInformation() to clarify it returns all gangs including the player's own gang.

I think this PR needs the docs to be re-generated, but I'm not sure how to do that locally.


catloversg:
You need to run `npm run doc`.

acidduk:
I don't have npm installed here! (Doing this via browser only)

d0sboots:
OK, we can let someone else handle the npm step.

PR #2295, opened 2025-08-24
RFA: add JSON-RPC 2.0 ping/pong to keep websockets alive across certain reverse proxies

sourdusk:
RFA: add JSON-RPC 2.0 ping/pong to keep websockets alive across certain reverse proxies

Code has been tested and successfully keeps websocket open on Cloudflare reverse proxy. In addition, an implemented "pong" command in bitburner-filesync returns successfully.

### DIFF
--- a/src/RemoteFileAPI/MessageDefinitions.ts
+++ b/src/RemoteFileAPI/MessageDefinitions.ts
@@ -27,12 +27,13 @@ type ResultType =
   | FileContent[]
   | RFAServerData[]
   | {
-      identifier: string;
-      binary: boolean;
-      save: SaveData;
-    }
+    identifier: string;
+    binary: boolean;
+    save: SaveData;
+  }
   | FileMetadata
-  | FileMetadata[];
+  | FileMetadata[]
+  | {};
 type FileDescription = FileData | FileContent | FileLocation | FileServer;
 
 export interface FileData {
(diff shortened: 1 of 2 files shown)


d0sboots:
I don't think this change makes sense. RFA is a client-initiated protocol; the client should be the one generating these periodic pings, with the server (i.e. BitBurner) responding with no-op pong replies.
But, you can already do that right now: You can send "ping" messages and BitBurner will respond with an error saying it doesn't understand the message type. Sure, it's not an "official" mechanism but it is guaranteed to keep the connection alive.

sourdusk:
> I don't think this change makes sense. RFA is a client-initiated protocol; the client should be the one generating these periodic pings, with the server (i.e. BitBurner) responding with no-op pong replies. But, you can already do that right now: You can send "ping" messages and BitBurner will respond with an error saying it doesn't understand the message type. Sure, it's not an "official" mechanism but it is guaranteed to keep the connection alive.

I'm happy to flip it the other way around and just add a handler for a ping; however, the repositories for the filesync API for both bitburner-filesync and viteburner haven't had commits made in 2 years. My goal was that the work would go to benefit others without me having to fork and maintain my own version of the sync server.

d0sboots:
> I'm happy to flip it the other way around and just add a handler for a ping; however, the repositories for the filesync API for both bitburner-filesync and viteburner haven't had commits made in 2 years. My goal was that the work would go to benefit others without me having to fork and maintain my own version of the sync server.

Don't use this as your only metric; it can be hard to distinguish between "the author is gone" and "there just isn't much to change." In particular, bitburner-filesync is under the same repo as bitburner itself.

PR #2297, opened 2025-08-26
UI: Add configurable option for auto-reconnecting to RFA client

catloversg:
The connection to the RFA client can be dropped due to many reasons (e.g., shutting down the machine or putting it to sleep). When it happens, Bitburner does not automatically reconnect, so the player may think that their RFA tool stops working randomly. This PR adds an option to automatically reconnect after a delay.

I tested 2 cases:
- Restart RFA tool.
- Put the machine to sleep.

https://github.com/user-attachments/assets/1c61ac1b-d13c-408d-b741-89ccf31c3794

### DIFF
--- a/src/GameOptions/ui/RemoteAPIPage.tsx
+++ b/src/GameOptions/ui/RemoteAPIPage.tsx
@@ -37,7 +37,7 @@ export const RemoteAPIPage = (): React.ReactElement => {
   function handleRemoteFileApiPortChange(event: React.ChangeEvent<HTMLInputElement>): void {
     const newValue = event.target.value.trim();
     setRemoteFileApiPort(newValue);
-    const port = Number.parseInt(newValue);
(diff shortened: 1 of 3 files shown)


src/GameOptions/ui/RemoteAPIPage.tsx
```diff
@@ -42,6 +47,18 @@ export const RemoteAPIPage = (): React.ReactElement => {
     setPortError("");
   }
 
+  function handleRemoteFileApiReconnectionDelayChange(event: React.ChangeEvent<HTMLInputElement>): void {
+    const newValue = event.target.value.trim();
+    setRemoteFileApiReconnectionDelay(newValue);
+    const reconnectionDelay = Number.parseInt(newValue);
```
d0sboots: You want `Number()`. ParseInt is a mess.

PR #2298, opened 2025-08-27
RFA: add ping function to respond to JSON-RPC pings

sourdusk:
PR includes no ns API changes but the documentation for RFA methods has been updated as part of this pull request.

npm run format and npm run lint have been run.

Has been tested both with a server with and without ping functionality and works fine.

Corresponds to https://github.com/bitburner-official/bitburner-filesync/pull/35

### DIFF
--- a/src/RemoteFileAPI/MessageHandlers.ts
+++ b/src/RemoteFileAPI/MessageHandlers.ts
@@ -260,4 +260,8 @@ export const RFARequestHandler: Record<string, (message: RFAMessage) => RFAMessa
 
     return new RFAMessage({ result: servers, id: msg.id });
   },
+
+  ping: function (msg: RFAMessage): RFAMessage {
+    return new RFAMessage({ result: {}, id: msg.id });
+  },
 };
--- a/src/Documentation/doc/en/programming/remote_api.md
+++ b/src/Documentation/doc/en/programming/remote_api.md
@@ -323,3 +323,23 @@ Output:
                 "purchasedByPlayer": boolean
             }[]
         }
+
+### ping
+
+Return a pong.
+
+Input:
+
+        {
+            "jsonrpc": "2.0",
+            "id": number,
+            "method": "ping"
+        }
+
+Output:
+
+        {
+            "jsonrpc": "2.0",
+            "id": number,
+            "result": {}
+        }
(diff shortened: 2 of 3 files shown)


src/RemoteFileAPI/MessageHandlers.ts
```diff
@@ -260,4 +260,8 @@ export const RFARequestHandler: Record<string, (message: RFAMessage) => RFAMessa
 
     return new RFAMessage({ result: servers, id: msg.id });
   },
+
+  ping: function (msg: RFAMessage): RFAMessage {
+    return new RFAMessage({ result: {}, id: msg.id });
```
d0sboots, line 265:
```suggestion
    return new RFAMessage({ result: null, id: msg.id });
```
I think null makes more sense for an empty response? And it also makes for a more straightforward type in ResultType.

catloversg:
Why do we need this API? RFA is built on top of WebSocket, and ping-pong is defined in the protocol. All (good) libraries support it.
- WebSocket RFC: https://datatracker.ietf.org/doc/html/rfc6455
- `ws` is the most popular WebSocket library for NodeJS and it supports ping-pong: https://github.com/websockets/ws/blob/master/doc/ws.md#websocketpingdata-mask-callback

If the RFA tool wants to maintain the connection via a heartbeat, it should use the built-in ping-pong mechanic of WebSocket.

d0sboots:
Ah, I didn't realize we could already accomplish this at a lower protocol layer.

sourdusk:
We're not currently using a websocket library in the code. Is that something that we want to adopt? Because otherwise, there's no way to send a ping websocket frame via the native javascript WebSocket API.

catloversg:
We don't need to use any library on the game side. When the RFA tool sends ping, the WebSocket connection will automatically reply with pong as the RFC defines.

Test code (as RFA tool):
```js
import { WebSocketServer } from "ws";

const wss = new WebSocketServer({ port: 12525 });

wss.on("connection", function connection(ws) {
  console.log("new connection");
  ws.on("error", console.error);
  ws.on("pong", function () {
    console.log("receive pong");
  });
});

setInterval(function ping() {
  for (const ws of wss.clients) {
    console.log("send ping");
    ws.ping();
  }
}, 5000);
```

How to test:
- Run test code.
- Set Bitburner's Remote API options and press Connect.

Capture

sourdusk:
I've added the websocket ping functionality as a pull request at https://github.com/bitburner-official/bitburner-filesync/pull/36 . Closing this request.

Issue #2192, opened 2025-06-06
DOCUMENTATION: Clean up "doc" folder


catloversg:
Our `doc` folder is a bit awkward. It contains:
- CONTRIBUTING.md
- Vague ideas of 3 BNs
- New BN guideline.
- FAQ
- Quotes.txt
- Irrelevant files for RTD

I think we should:
- Move CONTRIBUTING.md the the repository's root (#2191).
- Move BN guideline to `CONTRIBUTING.md`.
- FAQ: Move it to the in-game documentation.
- Quotes.txt and ideas of 3 BNs: I'm not sure, but I think we can remove them.
- Irrelevant files for RTD: Remove them.

@d0sboots What's your opinion?

d0sboots:
You didn't say it, but I think your unstated goal is to empty and then remove the top-level doc/ directory. I agree with this goal, the dir is confusing because it doesn't contain our real docs.

Contributing has been moved. I agree with the plan for FAQ. Normally I'd say the new BN guide belongs in its own file, but it's so small it can be merged into contributing. 

I took a look at the history for quotes: It was added alongside BN10 stuff and never touched really in the 7 years after. I think it was meant for lore but never integrated; I agree it can be removed.

aaaa-imcute:
also,there's a lore folder with two random files too

catloversg:
I keep that lore folder in the PR. It may be useful later when other contributors want to add more lore text.

PR #2302, opened 2025-08-28
DOCUMENTATION: Redistribute doc/ files into proper locations

sourdusk:
I went through issues to see if there was any low-hanging fruit that could be tackled.

I redistributed and cleaned up the doc/ folder per the suggestion in https://github.com/bitburner-official/bitburner-src/issues/2192

### DIFF
--- a/.github/workflows/build-artifacts.yml
+++ b/.github/workflows/build-artifacts.yml
@@ -1,8 +1,8 @@
 name: Build artifacts
 
 on:
-  release:
-    types: [published]
+  workflow_dispatch:
+  workflow_call:
 
 env:
   GH_TOKEN: ${{ github.token }}
@@ -88,28 +88,3 @@ jobs:
           include-hidden-files: true
           path: .build/*.zip
           if-no-files-found: error
-
-  upload:
-    name: Upload
-    runs-on: ubuntu-latest
-    needs: [build-windows, build-linux, build-mac]
-    steps:
-      - uses: actions/checkout@v4
-      - name: Use Node.js 20
-        uses: actions/setup-node@v4
-        with:
-          node-version: 20
-          cache: "npm"
-      - name: Download artifacts
-        uses: actions/download-artifact@v4
-        with:
-          path: .build
-          pattern: build-artifacts-*
-          merge-multiple: true
-      - name: List files
-        run: ls -la .build
-      - name: Upload to release
-        env:
-          GH_REF_NAME: ${{ github.ref_name }}
-        run: |
-          for i in .build/*.zip; do gh release upload "$GH_REF_NAME" "$i" --clobber; done;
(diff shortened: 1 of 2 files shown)


catloversg:
This part is irrelevant now: https://github.com/bitburner-official/bitburner-src/blob/916af9ce0f72a04c8818a1066956cece304a6b01/src/Documentation/doc/en/help/faq.md#steam-what-is-the-api-server

CONTRIBUTING.md
```diff
@@ -81,6 +81,20 @@ changes are okay to contribute:
 - Changes that directly affect the game's balance
 - New gameplay mechanics
 
+#### New Bitnode Contribution Guidelines
```
d0sboots: This part is awkward. Although it would be somewhat nice to avoid losing the new BN advice, it doesn't really fit with the rest of CONTRIBUTING, which is about much more low-level details. If we want to keep this, I feel like it would be best as it's own file (still)

sourdusk: Awkward, yes, but I don't really think it fits in with its own file. Maybe a design concepts section in contributing.md?

d0sboots: I don't think it belongs in CONTRIBUTING at all. That file is already big enough, and should stay focused on the mechanics of setting up your environment and doing PRs. If we have a design guide, I think it would still mostly belong in it's own file.

sourdusk: Do you have any suggestions for what the design principles document might contain other than this?

sourdusk:
> This part is irrelevant now: https://github.com/bitburner-official/bitburner-src/blob/916af9ce0f72a04c8818a1066956cece304a6b01/src/Documentation/doc/en/help/faq.md#steam-what-is-the-api-server

File has been deleted.

CONTRIBUTING.md
d0sboots: This works to start, but stuff like how we handle exceptions and other style stuff could go in there.

src/Documentation/doc/en/help/design-principles.md
```diff
@@ -1,11 +1,15 @@
-Promote:
+# Design Principles
```
d0sboots: I think I would put this at the top level, named in all caps similar to CONTRIBUTING.md. DESIGN_GUIDE seems like a reasonable choice. 

The reason is that the stuff in this directory is for user-facing docs, that we show in-game. The directory that you are removing was for developer-facing docs. It was confusingly named, and most of them were obsolete, but it does leave us short of a good place to put such things.

src/Documentation/doc/en/help/design-principles.md
```diff
@@ -1,11 +1,15 @@
...
+
+#### New Bitnode Design
```
catloversg: Nitpick: This line should use 2 `#`, `Promote` and `Avoid` should use 3 `#`.

d0sboots:
I think you pulled on the wrong branch :)

PR #2304, opened 2025-09-02
MISC: Export save data before migrating to v3

catloversg:
It's just as the title said.

Save file: 
[v2.8.1.gz](https://github.com/user-attachments/files/22094712/v2.8.1.gz)

How to test:
- Check `a.js` and corporation:
  - `a.js`: `ns.tail()` was converted to `ns.ui.openTail()`.
  - DreamSense was removed and refunded.
- Check if there is a new downloaded save file. It should be named in this format: `bitburnerSave_backup_${timestamp}.json.gz`.
- Load the new downloaded save file back to v2.8.1. Check `a.js` and corporation again.

### DIFF
--- a/src/utils/SaveDataMigrationUtils.ts
+++ b/src/utils/SaveDataMigrationUtils.ts
@@ -34,6 +34,8 @@ import { breakingChanges300 } from "./APIBreaks/3.0.0";
 import { calculateOfficeSizeUpgradeCost, calculateUpgradeCost } from "../Corporation/helpers";
 import type { PositiveInteger } from "../types";
 import { officeInitialCost, officeInitialSize, warehouseInitialCost } from "../Corporation/data/Constants";
+import { load } from "../db";
+import { downloadContentAsFile } from "./FileUtils";
 
 /** Function for performing a series of defined replacements. See 0.58.0 for usage */
 function convert(code: string, changes: [RegExp, string][]): string {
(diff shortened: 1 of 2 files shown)


src/utils/SaveDataMigrationUtils.ts
```diff
@@ -550,6 +552,16 @@ Error: ${e}`,
     if (found) Terminal.error("Filenames with whitespace found and corrected, see console for details.");
   }
   if (ver < 44) {
+    try {
+      /**
+       * Backup pre-v3 save data. We must use the data in IndexedDB instead of calling saveObject.getSaveData().
+       * getSaveData() returns data in v3 format, so the exported data will not be importable in pre-v3.
+       */
+      const saveData = await load();
+      downloadContentAsFile(saveData, `bitburnerSave_backup_${Math.round(Date.now() / 1000)}.json.gz`);
```
d0sboots: Let's put v2.8.1 in here. Also, can we base the timestamp on the lastUpdated of the save itself, rather than current time? That way, if you end up triggering this multiple times (let's say the game is crashing on startup), it will generate copies of the same file instead of new ones.

d0sboots:
Is that savegame OK to check in? If so, I'd like to make a test using it, although it doesn't have to hold up this PR.

catloversg:
> Is that savegame OK to check in? If so, I'd like to make a test using it, although it doesn't have to hold up this PR.

It's okay. I'll open a PR for that.

PR #2306, opened 2025-09-04
ELECTRON: Build universal macOS binary

### LINKED ISSUE #2307: Bad performance on Mac (apple silicon)
From discord:
https://discord.com/channels/415207508303544321/415213413745164318/1412856010586067085

This is likely due to the x64 version being used for steam, which then needs to run through a compatibility layer on apple silicon.

#2306 Attempts to fix this but I haven't gotten it working yet.

Snarling:
Bug reports included a user complaining about slow performance on a M4 mac, this is likely due to the x64 version being used.

Building the `universal` arch was suggested, but I haven't been able to get it to work. Windows and linux builds are working but the `electron-packager .package bitburner --platform darwin --arch universal --out .build --overwrite --icon .package/icon.icns --app-copyright "Copyright (C) 2025 Bitburner"` command is producing no output in the terminal and also creating no files.

* removed deprecated `electron-packager` and added `@electron/packager`
* npm audit fix (it had some electron related stuff, so this includes upgrading electron to a newer version)
* Changed package-electron.sh to attempt to build a mac universal build instead of separate x64 and arm64. Currently producing no output on the mac build.

Attempts to fix #2307


Snarling:
Might just be due to trying on Windows, looks like it is working when building from Linux. Pretty sure the mac builds haven't worked from Windows in a while.

Snarling:
On `@electron/packager`, the darwin universal build fails without producing any terminal output.
On `electron-packager`, it looks to do the whole build process, but then it fails with an error message `@electron/universal is only supported on darwin platforms` - it looks like universal build can only be built from a mac.
image

catloversg:
Changes:
- Update Electron to the latest version.
- Upgrade `electron-packager` to `@electron/packager`. These are the same package. `electron-packager` was moved to `@electron/packager`.
- Update the build script to generate a universal build on macOS. I applied the `osx-universal.x64ArchFiles` flag in #2308.

Tested on Windows 10 and Debian 12.

Snarling and I don't have Mac machines, so it's hard to test and release the Mac version. I'll open a PR to update our workflows. After that, Snarling can run a workflow to generate all artifacts, including the universal Mac build, and upload them to Steam.

sourdusk:
If you want to have me test the artifact before uploading to steam, I can do that.

tools/package-electron.sh
```diff
@@ -19,11 +19,13 @@ BUILD_PLATFORM="${1:-"all"}"
...
   "linux")
-    electron-packager .package bitburner --platform linux --arch x64,arm64 --out .build --overwrite --app-copyright "Copyright (C) 2024 Bitburner";;
+    electron-packager .package bitburner --platform linux --arch x64,arm64 --out .build --overwrite --app-copyright "Copyright (C) 2025 Bitburner";;
   "mac")
-    electron-packager .package bitburner --platform darwin --arch x64,arm64 --out .build --overwrite --icon .package/icon.icns --app-copyright "Copyright (C) 2024 Bitburner";;
+    electron-packager .package bitburner --platform darwin --arch universal --osx-universal.x64ArchFiles="Contents/Resources/app/node_modules/@catloversg/steamworks.js/dist/osx/*" --out .build --overwrite --icon .package/icon.icns --app-copyright "Copyright (C) 2025 Bitburner";;
   *)
-    electron-packager .package bitburner --platform win32,linux,darwin --arch x64,arm64 --out .build --overwrite --icon .package/icon --app-copyright "Copyright (C) 2024 Bitburner";;
+    electron-packager .package bitburner --platform win32 --arch x64,arm64 --out .build --overwrite --icon .package/icon.ico --app-copyright "Copyright (C) 2025 Bitburner";
```
d0sboots: It would be better to turn this into three calls to the case, instead of duplicating all of this.

d0sboots:
This looks fine, but of course the only test that matters is if it works.

PR #2308, opened 2025-09-04
ELECTRON: macOS Universal Build

sourdusk:
Pull request adds a universal build for macOS, tested on macOS 15.6.1 both silicon and rosetta. Also bumps copyright year.


catloversg:
I merged the `osx-universal.x64ArchFiles` flag to #2306.

PR #2309, opened 2025-09-05
CD: Split build-artifacts.yml into 2 workflows

catloversg:
In #2120, I added a workflow to build artifacts and upload them to the new release. This PR splits that workflow into 2 smaller ones: one to build and one to upload.

The current workflow only runs after Snarling creates a new release, but they may want to build artifacts even when there is no new release. This is also useful when Snarling or other contributors want to build a universal macOS binary without a Mac machine.

Test:
- https://github.com/catloversg/bitburner-src/actions/runs/17485307596
- https://github.com/catloversg/bitburner-src/actions/runs/17485343612

### DIFF
--- a/.github/workflows/upload-artifacts.yml
+++ b/.github/workflows/upload-artifacts.yml
@@ -1,0 +1,40 @@
+name: Upload artifacts
+
+on:
+  release:
+    types: [published]
+
+env:
+  GH_TOKEN: ${{ github.token }}
+
+jobs:
+  build:
+    name: Build artifacts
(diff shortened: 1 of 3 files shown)


.github/workflows/upload-artifacts.yml
```diff
@@ -0,0 +1,40 @@
+name: Upload artifacts
```
d0sboots: Maybe call it "Build and upload artifacts" so it's clear the other doesn't need to be manually run?

PR #2311, opened 2025-09-06
UI: Clarify why you can't buy 4S when disabled

femboyfireball:
When 4S is disabled before starting a bitnode using the Advanced Bitnode options, the relavent `ns.stock.purchaseX` functions aren't clear as to why 4S is disabled. This PR fixes that.

It's also more clear in the UI why 4S is disabled when you try to buy 4S through Wall Street

What happens when you try to use `ns.stock.purchaseX` to purchase 4S and 4S API access
(Upload later today)

### DIFF
--- a/src/NetscriptFunctions/StockMarket.ts
+++ b/src/NetscriptFunctions/StockMarket.ts
@@ -247,7 +247,7 @@ export function NetscriptStockMarket(): InternalAPI<TIX> {
     },
     purchase4SMarketData: (ctx) => () => {
       if (Player.bitNodeOptions.disable4SData) {
-        helpers.log(ctx, () => "4S Market Data is disabled.");
+        helpers.log(ctx, () => "4S Market Data is disabled in advanced bitnode options.");
         return false;
       }
 
@@ -268,7 +268,7 @@ export function NetscriptStockMarket(): InternalAPI<TIX> {
     },
     purchase4SMarketDataTixApi: (ctx) => () => {
       if (Player.bitNodeOptions.disable4SData) {
-        helpers.log(ctx, () => "4S Market Data is disabled.");
+        helpers.log(ctx, () => "4S Market Data is disabled in advanced bitnode options.");
         return false;
       }
 
@@ -307,6 +307,10 @@ export function NetscriptStockMarket(): InternalAPI<TIX> {
       return true;
     },
     purchaseTixApi: (ctx) => () => {
+            if (Player.bitNodeOptions.disable4SData) {
+        helpers.log(ctx, () => "TIX API access is disabled in advanced bitnode options.");
+        return false;
+      }
       if (Player.hasTixApiAccess) {
         helpers.log(ctx, () => "Already purchased TIX API");
         return true;
--- a/src/NetscriptFunctions/StockMarket.ts
+++ b/src/NetscriptFunctions/StockMarket.ts
@@ -307,7 +307,7 @@ export function NetscriptStockMarket(): InternalAPI<TIX> {
       return true;
     },
     purchaseTixApi: (ctx) => () => {
-            if (Player.bitNodeOptions.disable4SData) {
+      if (Player.bitNodeOptions.disable4SData) {
         helpers.log(ctx, () => "TIX API access is disabled in advanced bitnode options.");
         return false;
       }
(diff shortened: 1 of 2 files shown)


gmcew:
In the context of the enable/disable toggles helping with BN achievements, I'm on the side that the base TIX API should still be accessible with the 4S toggle applied. Without it, the stock minigame becomes a purely manual (or slightly exploity) activity which is not the requirement for that achievement, and feels excessive.

If the consensus is against me, we should also rename the BN options toggle to 'Disable 4S market data and TIX API' for clarity.

Thanks for putting this together so quickly!

d0sboots:
Agree, TIX access should be allowed.

src/NetscriptFunctions/StockMarket.ts
```diff
@@ -307,6 +307,10 @@ export function NetscriptStockMarket(): InternalAPI<TIX> {
       return true;
     },
     purchaseTixApi: (ctx) => () => {
+      if (Player.bitNodeOptions.disable4SData) {
```
d0sboots: The fact that this wasn't here before is evidence that TIX access wasn't *meant* to be disabled.

src/StockMarket/ui/InfoAndPurchases.tsx
```diff
@@ -139,6 +141,7 @@ function PurchaseTixApiAccessButton(props: IProps): React.ReactElement {
 function Purchase4SMarketDataButton(props: IProps): React.ReactElement {
   function purchase4SMarketData(): void {
     if (Player.bitNodeOptions.disable4SData) {
+      dialogBoxCreate("4S Market Data has been disabled through Advanced Bitnode options.");
```
d0sboots: Unneeded; the correct thing is to disable the button, as well.

src/StockMarket/ui/InfoAndPurchases.tsx
```diff
@@ -106,6 +107,7 @@ function PurchaseWseAccountButton(props: IProps): React.ReactElement {
 function PurchaseTixApiAccessButton(props: IProps): React.ReactElement {
   function purchaseTixApiAccess(): void {
     if (Player.bitNodeOptions.disable4SData) {
```
d0sboots: This check needs to be moved (not added) to the 4s tix API block.

catloversg:
The "Disable 4S Market Data" option should only affect 4S-related things. I committed 2 changes:
- Remove the wrong check in `purchaseTixApi` API. This option should not affect it.
- Add/update tooltip text to explain why the player cannot unlock the feature.

Note that I made a questionable change here. Currently, I hide the entire UI block of 4S data if the player enables the "Disable 4S Market Data" option. Personally, I don't see any problems with this behavior. This is an opt-in feature, so if they enable it and forget it, that's their fault. This is consistent with other options. For example, if they enable the "Disable Gang" option, they won't see the "Create Gang" button.

Nonetheless, I still changed this behavior in the latest commit. Instead of hiding the entire UI block, I show them, but I disable the buttons and explain the reason in the tooltip. This is inconsistent with other options, though. I'm not sure if we should do this with other options.

@d0sboots What's your opinion?

src/StockMarket/ui/InfoAndPurchases.tsx
```diff
@@ -125,23 +138,35 @@ function PurchaseTixApiAccessButton(props: IProps): React.ReactElement {
...
-    return (
-      <Button disabled={!Player.canAfford(cost) || !Player.hasWseAccount} onClick={purchaseTixApiAccess}>
-        Buy Trade Information eXchange (TIX) API Access -&nbsp;
-        <Money money={cost} forPurchase={true} />
-      </Button>
-    );
   }
+  const cost = StockMarketConstants.TixApiCost;
+  let tooltipTitle = "Let you trade stock through Netscript";
+  if (!Player.hasWseAccount) {
```
d0sboots, line 144: If you check the netscript functions, purchasing tix access doesn't require a wse account. Purchasing 4s data tix access doesn't require 4s data. Maybe those things should be changed, but I don't think they should be changing in this PR.

catloversg: #2323 will make the requirements consistent between UI and NS APIs.
- TIX API Access: Require WSE account.
- 4S Market Data Access: Require WSE account.
- 4S Market Data TIX API Access: Require TIX API Access. In `purchase4SMarketDataTixApi` (NS API), we check both `hasWseAccount` and `hasTixApiAccess` via `checkTixApiAccess`, but TIX API Access now requires WSE account, so I think it's fine if the UI code only checks `hasTixApiAccess`.

d0sboots:
This looks good given the way the poll is trending. We can make follow-up tweaks if needed.

PR #2312, opened 2025-09-06
API: added abilitiy to set tail window opacity

sourdusk:
Adds an ns api function to set tail window opacity.


d0sboots:
I see two issues with this:
1. It's a fair amount of infrastructure for a very niche feature. There are a *lot* of CSS properties that could be set on the log box, and we can't/won't be doing this for all of them. If this allowed for setting properties more generally, that would be different, but that brings up point 2:
2. Using the sx prop causes new global styles to be created. This makes it unsuitable to use for user-controlled values, because you can end up with hundreds or thousands of repetitive styles that leak memory. 

Because of these points, I think the better approach is for users to use hooks to modify the styles with the DOM directly.

PR #2316, opened 2025-09-20
REFACTOR: Rewrite infiltration to pull state out of React

d0sboots:
Upcoming projects (auto-infil, the possibility of making infil a work task, etc.) require infiltration state to transition with predictable timing and be under our control, instead of inside React. This refactor accomplishes this by pulling the state out into accompanying model classes.

After this, infiltration can theoretically run headless (without UI), although it doesn't actually, and you would quickly be hospitalized due to failing all the minigames.

There should be no user-visible changes, aside from the progress-bars scrolling much more smoothly.

This partly supersedes #2257; That PR will be rebased (probably re-created) on top of this one, once this is submitted.

## Testing

- Manually checked infiltration, both with and without augments
- Here's a video of infil with augs, hacked to do slashGame back-to-back: https://github.com/user-attachments/assets/d2d6c6da-0933-45ae-aa56-492294d9091e
- I pushed this to https://d0sboots.github.io/bitburner-src if others want to test (that includes the dev-menu change for easy augment switching)


catloversg:
Just a heads-up, this PR is still one of the most prioritized PR on my watchlist. I'll take a look at it after the next review round of the dnet PR.

src/Infiltration/ui/InfiltrationRoot.tsx
```diff
@@ -1,57 +1,134 @@
...
+  useEffect(() => cancel, [cancel]);
 
+  useEffect(() => {
+    if (!state) {
+      return;
+    }
+    const press = (event: KeyboardEvent) => {
+      if (!event.isTrusted || !(event instanceof KeyboardEvent)) {
+        state.onFailure({ automated: true });
+        return;
```
ficocelliguy, line 75: This can be removed, since we are explicitly supporting and balancing automated infiltration now

src/Infiltration/Infiltration.ts
```diff
@@ -0,0 +1,218 @@
...
+    this.results += "✗";
+    this.clearTimeouts();
+    this.stage = new CountdownModel(this);
+    Player.receiveRumor(FactionName.ShadowsOfAnarchy);
+    let damage = calculateDamageAfterFailingInfiltration(this.startingSecurityLevel);
+    // Kill the player immediately if they use automation, so it's clear they're not meant to
+    if (options?.automated) {
+      damage = Player.hp.current;
+      setTimeout(() => {
+        SnackbarEvents.emit("You were hospitalized. Do not try to automate infiltration!", ToastVariant.WARNING, 5000);
```
ficocelliguy, line 149: This auto-fail on automation can be removed, too

ficocelliguy:
I see this in the js console:
```
react.development.js:220 Warning: Failed prop type: Invalid prop `children` supplied to `ForwardRef(Container)`, expected a ReactNode.
    at Container (webpack-internal:///./node_modules/@mui/system/esm/Container/createContainer.js:115:19)
    at InfiltrationRoot (webpack-internal:///./src/Infiltration/ui/InfiltrationRoot.tsx:75:75)
    at div
    at eval (webpack-internal:///./node_modules/@emotion/react/dist/emotion-element-c39617d8.browser.esm.js:56:66)
    at Box (webpack-internal:///./node_modules/@mui/system/esm/createBox.js:35:72)
```

src/Infiltration/ui/InfiltrationRoot.tsx
catloversg: I think the old, not-so-good checks preventing auto-infil should be retained. The new APIs will be gated behind SF11, so in pre-SF11 situations, bypassing the old checks is still a good and funny challenge. We only need to change the toast message and tell the player that their bypass is caught, and they should find a better trick.

d0sboots: Two things:
1. As I stated in the description, this is a *pure* refactor with no behavior changes. The place to modify this is in the follow-up (where auto infil is supported)
2. I'm with catlover that since there will be a sanctioned API now, nudging towards that is better. Something like "Caught an attempt to infiltrate that bypasses the API", which both says that there *is* an API and also that you'll have to be more clever if you want to do this and not be caught.

ficocelliguy:
All of the minigames seem to work fine both with and without the SoA augs, testing with my legacy infiltration script

d0sboots:
> I see this in the js console:
> 
> ```
> react.development.js:220 Warning: Failed prop type: Invalid prop `children` supplied to `ForwardRef(Container)`, expected a ReactNode.
>     at Container (webpack-internal:///./node_modules/@mui/system/esm/Container/createContainer.js:115:19)
>     at InfiltrationRoot (webpack-internal:///./src/Infiltration/ui/InfiltrationRoot.tsx:75:75)
>     at div
>     at eval (webpack-internal:///./node_modules/@emotion/react/dist/emotion-element-c39617d8.browser.esm.js:56:66)
>     at Box (webpack-internal:///./node_modules/@mui/system/esm/createBox.js:35:72)
> ```

Is this all the time, or from a specific thing?

ficocelliguy:
The warning/error logged above occurs once when you start the infiltration and the UI pops up on the screen. Nothing special is required, I think

d0sboots:
> The warning/error logged above occurs once when you start the infiltration and the UI pops up on the screen. Nothing special is required, I think

This was really gnarly, and came down to this: <https://github.com/facebook/react/issues/17871>

Specifically, it's documented that undefined and booleans are dropped from JSX, but sneakily `true` actually isn't kosher and will give you this error. At least in some cases.

src/Infiltration/Infiltration.ts
```diff
@@ -0,0 +1,218 @@
...
+  maxLevel: number;
+  /** Checkmarks that represent success/failure per-stage. */
+  results = "";
+
+  /** Used to avoid repeating games too quickly. gameIds[0] is the current (or last) game. */
+  gameIds = [-1, -1, -1];
+
+  /** Invalid until infiltration is started, used to calculate rewards */
+  gameStartTimestamp = -1;
+  /** undefined for timeouts that have finished. Typescript isn't happy with null. */
```
ficocelliguy: this looks like it should be a comment attached to line 53 instead

ficocelliguy:
I don't see any regressions in my testing of the functionality in the deployed fork, and the code itself looks good to me (the unusual things I see are all faithfully copied from existing infil code). I think this has accomplished its goals of moving state out of react. +1

ficocelliguy (approved):

src/Infiltration/Infiltration.ts
d0sboots: Fixed, and expanded on some of the comments here.

d0sboots:
I'm submitting this. I'm pretty sure the core of it is sound, but we can address any followups as needed, or in the worst case revert the whole thing.

PR #2317, opened 2025-09-22
UI: Better status bar animations

d0sboots:
This is an alternate implementation of #2286. It does the same wrap-around behaviour for when the progressbar crosses into the next level, but it skips animation entirely if the effective skill level goes down, or if more than one level is gained at a time.

The implementation uses the animate() DOM api instead of manipulating styles, which completely avoids the issues of having CSS style buildup. This API is designed for exactly what we're trying to do.

I also pushed rerender handling down from CharacterOverview to StatsProgressBar, which simplifies things and is helpful for doing the animation implementation.

Here's a video showing XP gain of a bit more than 1 level (no animation), which then falls down to less than 1 level (has animation). I double click towards the end, which has no animation again. The very end shows decreasing XP with no animation.

https://github.com/user-attachments/assets/62123464-3a60-4bd7-be25-39ecee689910

### DIFF
--- a/src/ui/React/StatsProgressBar.tsx
+++ b/src/ui/React/StatsProgressBar.tsx
@@ -1,47 +1,105 @@
-import * as React from "react";
+import React, { useEffect, useMemo, useRef, useState } from "react";
 import LinearProgress from "@mui/material/LinearProgress";
-import { TableCell, Tooltip, Typography } from "@mui/material";
-import { useStyles } from "./CharacterOverview";
-import { ISkillProgress } from "../../PersonObjects/formulas/skill";
+import { TableRow, TableCell, Tooltip, Typography } from "@mui/material";
+import { OverviewEventEmitter, useStyles } from "./CharacterOverview";
+import { Player } from "@player";
+import { currentNodeMults } from "../../BitNode/BitNodeMultipliers";
+import { calculateSkillProgress } from "../../PersonObjects/formulas/skill";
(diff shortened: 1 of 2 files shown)


d0sboots:
@vamo89 I discovered the `animate()` API call when working on something else, and realized it's probably the best way to deal with this PR. It seemed like you had run out of time to work on this, so I threw something together on my own.

src/ui/React/StatsProgressBar.tsx
```diff
@@ -1,76 +1,136 @@
...
+      const mult = skillMultUpdaters[name]();
+      // Since this creates a new object every time, it normally causes a rerender every time.
+      const newProgress = calculateSkillProgress(Player.exp[skillNameMap[name]], mult);
+      setProgress((progress) => {
+        if (progress.progress === newProgress.progress) {
+          // Nothing has changed, return the original object for no rerender.
+          return progress;
+        }
+        // This takes place in the state updater for progress.
+        const ele = domRef.current?.firstElementChild as HTMLElement;
```
catloversg: Why do we need this typecasting? I tried removing it, and TSLint did not complain.

catloversg (commented):
Except the small question about typecasting, it looks good to me.

src/ui/React/StatsProgressBar.tsx
d0sboots: Done. Based on older code.

PR #2320, opened 2025-09-25
CODEBASE: Update api-documentor and api-extractor

mctylr-gh:
CATEGORY: Development Environment
resolves #1566 
(no temporary 'pinning' needed)

I have verified that the HTML/markdown table generation bug in [#4878](https://github.com/microsoft/rushstack/issues/4878) in rushstack for api-documentor has been fixed as per rushstack#5256. The test-case [repro](https://github.com/catloversg/api-documenter-bug-pr-4578) now produces the correct, expected output.

I have confirmed that the generated output in bitburner from `npm run doc` now generated HTML tables, and correctly inserts a blank line between the `</table>` tag and the following line (e.g. Returns).

Stylistically it could use some whitespace, but it is correctly rendered.

This commit is only the updated packages, not the updated generated documentation. I assume that is automatically generated by the GitHub workflow.


catloversg:
- `package-lock.json`: When I update versions and run `npm i` by myself, the result contains more license info than your commit, so I added them as well.
- There is a React warning when openning some pages caused by `th` tag. I fixed it in the latest commit.

It looks like there is nothing wrong. We can merge it after @d0sboots tests it.

PR #2321, opened 2025-09-26
BUGFIX: Color picker appears behind theme editor

catloversg:
We changed z-index of some MUI components in #2253 and #2282. This works with things using built-in MUI components, but it does not work with `material-ui-color`'s component. Please check my comment for more information. The workaround is ugly, so I'm open to other ideas. I used this one because it's simple. `material-ui-color` will be gone when we update to React 19 and MUI 7, so I don't want to waste too much time with it.

Before:

before

After:

after

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -19,6 +19,20 @@ ReactDOM.render(
   document.getElementById("root"),
 );
 
+// TODO: Remove this workaround when we get rid of material-ui-color and update to React 19+.
+/**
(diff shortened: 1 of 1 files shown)


mctylr-gh:
A medium-term _alternative_ would be to switch to [mui-color-input](https://viclafouch.github.io/mui-color-input/) which v4.0.2 supports MUI v5, and the current version supports React 18 & 19, MUI 7, and includes TypeScript support.

But I'm fine with this.

mctylr-gh:
Works for me.

d0sboots:
Oops, this feel off my radar

PR #2323, opened 2025-09-27
API: Enforce requirement of WSE account with purchaseTixApi and purchase4SMarketData APIs

catloversg:
In UI, in order to purchase TIX API access and 4S Market Data access, the player must have a WSE account. However, we have not enforced this requirement in NS APIs. This PR makes this requirement consistent between UI and NS APIs.

Test code:
```js
/** @param {NS} ns */
  ns.ui.openTail();
  ns.print(ns.stock.purchaseTixApi());
  ns.print(ns.stock.purchase4SMarketData());
}
```

```
In order to purchase TIX API access and 4S Market Data access, you need a WSE account.
This requirement was only enforced in UI, not NS APIs. Now, it is enforced in both.

purchaseTixApi
purchase4SMarketData

test.js: (Line numbers: 4, 5)
```

### DIFF
--- a/src/NetscriptFunctions/StockMarket.ts
+++ b/src/NetscriptFunctions/StockMarket.ts
@@ -256,6 +256,10 @@ export function NetscriptStockMarket(): InternalAPI<TIX> {
         return true;
       }
 
+      if (!Player.hasWseAccount) {
+        helpers.log(ctx, () => "You need to have a WSE account");
+        return false;
+      }
       if (Player.money < getStockMarket4SDataCost()) {
         helpers.log(ctx, () => "Not enough money to purchase 4S Market Data.");
         return false;
@@ -312,6 +316,10 @@ export function NetscriptStockMarket(): InternalAPI<TIX> {
         return true;
       }
 
+      if (!Player.hasWseAccount) {
+        helpers.log(ctx, () => "You need to have a WSE account");
+        return false;
+      }
       if (Player.money < getStockMarketTixApiCost()) {
         helpers.log(ctx, () => "Not enough money to purchase TIX API Access");
         return false;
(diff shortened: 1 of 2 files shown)


d0sboots:
There are two potential ways of looking at this, both (IMO) valid:
* One, implicitly supported by the way the NS apis are coded, says that the TIX apis essentially authorize a separate "agent" to access the WSE automatically. By this logic, TIX access is independent from WSE access, which only grants the ability for you to make trades manually. Similarly, TIX 4S market data is independent from "regular" 4S market data.
* The other, implicitly supported by the layout of the stock buttons, saying that the TIX apis are add-on features of your WSE account. So TIX access would require a WSE account, and TIX 4S data would require regular 4S data.

I agree that we should regularize between these. I'm not positive that we should pick the 2nd; we've been in a state where the first is (half-)allowed for a while, and it appears to give more gameplay opportunities. For instance, you could opt to avoid buying WSE access to save money, at the expense of having no built-in UI to see what's going on.

catloversg:
I don't have a strong opinion on this one. If you prefer the first option (TIX access is independent from WSE access), I'm okay with it. Note that we currently init the stock market after checking `Player.hasWseAccount` (when loading the save data) or after the player calls `purchaseWseAccount` API. If we make TIX access independent from WSE account, we need to change the initialization a bit.

Edit: Currently, although the player can call `purchaseTixApi` API without a WSE account, they cannot use most APIs because `checkTixApiAccess` (an internal check) checks both `hasWseAccount` and `hasTixApiAccess`. This means TIX API access effectively requires a WSE account, despite `purchaseTixApi` not checking it.

If you think we should go with the first option, please confirm it. I'll close this PR and handle the change and the docs in a different PR.

PR #2325, opened 2025-09-28
ELECTRON: Allow automatically hiding the top menu in Electron

femboyfireball:
This PR allows the player to optionally hide the top bar in the Electron build of the game. The setting persists between game sessions, and the menu can be brought back up with the alt key as normal. I've also added a new section to the Electron menu, View, since this option didn't fit with any of the other categories.

View before the menu is hidden:
image

View menu (Ignore the large blur backdrop, that's a bug unrelated to BitBurner):
image

output


electron/menu.js
```diff
@@ -203,6 +203,26 @@ function getMenu(window) {
         { label: "Select All", accelerator: "CmdOrCtrl+A", selector: "selectAll:" },
       ],
     },
+    {
+      label: "View",
+      submenu: [
+        {
+          label: "Hide top menu (Requires restart)",
```
d0sboots:
```suggestion
          label: "Autohide top menu (Requires restart)",
```

electron/menu.js
```diff
@@ -203,6 +203,26 @@ function getMenu(window) {
...
+          type: "checkbox",
+          checked: storage.isMenuHideEnabled(),
+          click: (menuItem) => {
+            storage.setMenuHideConfig(menuItem.checked);
+            utils.writeToast(
```
d0sboots: I don't think it makes sense to show a toast for this; menu items that are toggles are well understood and don't typically have other obvious notifications.

electron/windowTracker.js
```diff
@@ -60,6 +64,7 @@ const windowTracker = (windowName) => {
       width: windowState.width,
       height: windowState.height,
       isMaximized: windowState.isMaximized,
+      autoHideMenuBar: windowState.autoHideMenuBar ?? true,
```
d0sboots:
```suggestion
      autoHideMenuBar: windowState.autoHideMenuBar,
```

If this absolutely must be a real bool, you can add `!!` to the front.

electron/menu.js
```diff
@@ -203,6 +203,26 @@ function getMenu(window) {
...
       ],
     },
+    {
+      label: "View",
+      submenu: [
+        {
+          label: "Hide top menu (Requires restart)",
+          type: "checkbox",
+          checked: storage.isMenuHideEnabled(),
+          click: (menuItem) => {
```
d0sboots: autoHideMenuBar is an instance property of the window that can be set. Is there any reason we can't also just set it here and avoid needing to restart?

electron/gameWindow.js
```diff
@@ -18,12 +18,14 @@ async function createWindow(killall) {
...
 
   const tracker = windowTracker("main");
+
   const window = new BrowserWindow({
     icon,
     show: false,
     backgroundThrottling: false,
     backgroundColor: "#000000",
     title: "Bitburner",
+    autoHideMenuBar: tracker.state.autoHideMenuBar,
```
catloversg: Use `isMenuHideEnabled` in `storage.js` and delete all unnecessary changes in `windowTracker.js`.

electron/menu.js
catloversg: Remove `(Requires restart)`. You can combine `window.setAutoHideMenuBar` and `window.setMenuBarVisibility` to apply the change without restarting.

electron/menu.js
catloversg: Combine `setAutoHideMenuBar` and `setMenuBarVisibility` like this:
```js
          click: (menuItem) => {
            storage.setMenuHideConfig(menuItem.checked);
            window.setAutoHideMenuBar(menuItem.checked);
            if (menuItem.checked) {
              window.setMenuBarVisibility(false);
            } else {
              window.setMenuBarVisibility(true);
            }
            refreshMenu(window);
          },
```
Note that you need both. From the docs:
```
If the menu bar is already visible, calling setAutoHideMenuBar(true) won't hide it immediately.
```

electron/menu.js
catloversg: I agree with d0sboots.

electron/windowTracker.js
```diff
@@ -4,6 +4,7 @@ const log = require("electron-log");
 const debounce = require("lodash/debounce");
 const Store = require("electron-store");
 const store = new Store();
+const storage = require("./storage");
```
catloversg: Remove all changes in this file. The config is already controlled by `storage.js`.

catloversg (commented):
Optional: If you create a new menu called "View" like this, I suggest moving "Fullscreen" and all zoom options to it as well. In many tools, "Fullscreen" and "Zoom ..." actions are in the "View" menu.
```js
      submenu: [
        {
          label: "Fullscreen",
          accelerator: "f9",
          click: (() => {
            // Code
          })(),
        },
        {
          type: "separator",
        },
        // Zoom options here
        {
          type: "separator",
        },
        {
          label: "Auto hide menu bar",
          type: "checkbox",
          checked: storage.isMenuHideEnabled(),
          click: (menuItem) => {
            // Code
          },
        },
```

catloversg (approved):

PR #2328, opened 2025-09-28
UI: Print error detail to terminal if script cannot be compiled or there is runtime error while autocompleting

catloversg:
A player asked us why autocomplete does not work even after exporting an `autocomplete` function. After checking their save file, I find out that their script cannot be compiled due to syntax errors, but we intentionally ignore the error. I don't think we should do that. If the script cannot be compiled, we should tell the player the reason instead of hiding it.

Capture


d0sboots:
How does this compare to the behavior if the script compiles, but runtime errors inside the autocomplete function? It feels like they ought to behave the same.

catloversg:
The runtime error is also printed in the latest commit.

Capture

PR #2331, opened 2025-09-30
TOOLS: Update doc.sh

catloversg:
Currently, if there are new md files generated by `api-documenter`, they won't be included in `src\Documentation\pages.ts` because `tools/bundle-doc/index.js` runs before `api-documenter`. This PR changes the order of tools in `doc.sh` to fix it.

### DIFF
--- a/tools/doc.sh
+++ b/tools/doc.sh
@@ -11,11 +11,14 @@ echo ""
 echo "Running cleanup tasks..."
 rm input/bitburner.api.json && rm -r input
 
-# This git add is needed due to documenter using wrong line endings. Console spam discarded.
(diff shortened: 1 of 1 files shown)


tools/doc.sh
```diff
@@ -19,3 +15,7 @@ rm input/bitburner.api.json && rm -r input
 git add markdown/ 2> /dev/null && git add tsdoc-metadata.json 2> /dev/null
 echo ""
 echo "Documentation build completed."
+
+echo ""
```
d0sboots, line 23: This should come before the echo that says "done". This maybe should also come before the git adds?

PR #2332, opened 2025-10-04
JEST: Quiet Jest tests output on expected failures

mctylr-gh:
This patch quiets the console.error output on expected failures that are suppose to occur during `npm test`. It mocks and restores the console.error function.

For `UserInterface.test.ts` it is limited to only tests that are expected to fail.

In case of `Save.test.ts` is written as a single test, so console.error is muted for the entire test.

### DIFF
--- a/test/jest/Netscript/UserInterface.test.ts
+++ b/test/jest/Netscript/UserInterface.test.ts
@@ -81,7 +81,7 @@ describe("setTheme", () => {
       expect(result.primary).toStrictEqual(defaultTheme.primary);
     });
     afterEach(() => {
-      jest.restoreAllMocks(); // e.g. console.error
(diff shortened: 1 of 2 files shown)


catloversg:
Mocking functions like this works, but I suggest using `jest.spyOn` and calling `mockRestore`.

Another problem is `test\jest\Save.test.ts`. The error messages shown in those tests are not false-positive ones. There are real problems with that file. I'll fix them in #2333 later.

mctylr-gh:
I think that takes care of your concerns, both using `jest.spyOn` / `mockRestore` and undoing *my* changes to `test/jest/Save.test.ts`.

catloversg (approved):

PR #2333, opened 2025-10-04
JEST: Enable restoreMocks option and fix lint errors

catloversg:
While reviewing #2332, I found some lint errors and other potential problems in other test files. This PR makes these changes:
- Enable the `restoreMocks` option to make Jest restore mocks automatically between tests.
- Only use `jest.spyOn(console, "error")` when necessary and call `mockRestore` after that.
- Fix trivial lint errors.
- Instead of using 1e500 in BB tests, I create the new `gainHighCharismaLevel` function. 1e500 is effectively `Infinity` and gaining `Infinity` exp is unrealistic, except for long int farming sessions and deep BN12.

I intentionally ignored `test\jest\Go\NetscriptGo.test.ts`. I'll create another issue to discuss it.


catloversg:
I found problems in `test\jest\Save.test.ts`. I'll mark this PR as ready for review after fixing them.

PR #2335, opened 2025-10-05
IPVGO: Consistently return `error()` calls to make sure it is obvious the function halts at that point

### LINKED ISSUE #2334: CODEBASE: Standardize error handling in netscriptGoImplementation.ts
While investigating lint errors in `test\jest\Go\NetscriptGo.test.ts`, I found some problems in Go code.

In `src\NetscriptFunctions\Go.ts`:
```ts
const error =
  (ctx: NetscriptContext) =>
  (message: string): never => {
    throw errorMessage(ctx, message);
  };
```
`error` is only a simple wrapper around `throw errorMessage`, then `error(ctx)` is passed to many functions in `netscriptGoImplementation.ts`. Technically, calling `error("error message")` is not wrong, but we really should call `return error("error message")`, especially when we want to narrow down the type like I showed on Discord when we discussed the dnet PR.

You also should check `resetBoardState`. That function receives `error: (s: string) => void` instead of `error: (s: string) => never`.

After standardizing the implementation in `netscriptGoImplementation.ts`, I'll fix lint errors in Go tests.

@ficocelliguy Can you please check them?

ficocelliguy:
Resolves #2334 

Currently, error handling is somewhat inconsistent in go code. This PR ensures that any calls to error() are returned, instead of implicitly relying on the throw inside of the call. It makes it easier to read for both humans and typescript.

(We also discussed throwing directly. However, when I tested adding `throw new ErrorMessage(ctx, ...)`, it required passing ctx around between many functions that were pretty distant from netscript, and it seemed more clean to handle it this way)

### DIFF
--- a/src/Go/effects/netscriptGoImplementation.ts
+++ b/src/Go/effects/netscriptGoImplementation.ts
@@ -56,16 +56,16 @@ export function validateMove(error: (s: string) => never, x: number, y: number, 
 
   const boardSize = Go.currentGame.board.length;
   if (x < 0 || x >= boardSize) {
-    error(`Invalid column number (x = ${x}), column must be a number 0 through ${boardSize - 1}`);
+    return error(`Invalid column number (x = ${x}), column must be a number 0 through ${boardSize - 1}`);
   }
   if (y < 0 || y >= boardSize) {
-    error(`Invalid row number (y = ${y}), row must be a number 0 through ${boardSize - 1}`);
+    return error(`Invalid row number (y = ${y}), row must be a number 0 through ${boardSize - 1}`);
   }
 
   const validity = evaluateIfMoveIsValid(Go.currentGame, x, y, moveColor);
   const point = Go.currentGame.board[x][y];
   if (!point && check.onlineNode) {
-    error(
+    return error(
       `The node ${x},${y} is offline, so you cannot ${
         methodName === "removeRouter"
           ? "clear this point with removeRouter()"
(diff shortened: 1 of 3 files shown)


d0sboots:
> We also discussed throwing directly. However, when I tested adding throw new ErrorMessage(ctx, ...), it required passing ctx around between many functions that were pretty distant from netscript, and it seemed more clean to handle it this way.

Can you elaborate on this? All the changes/callsites I'm seeing are in netscriptGoImplementation, which is in my mind quite close to netscript. If you just passed ctx to those functions, you could avoid the indirection of logger and error closures, and the error handling code would become much simpler to read due to that lack of indirection.

ficocelliguy:
Some of the functions in that file passed the error handling through other functions as much as five calls deep, which felt like a long distance from the main netscript api file... but it's the same passing ctx or error() though that distance I realized, once I went back to look at it. (The other reason I had originally set it up that way is so the error and logger calls could be easily dependency injected for unit tests... though I had forgotten that until I was already most of the way through refactoring)

Anyway, the latest commit uses `throw ErrorMessage()` everywhere for clarity.

catloversg:
Minor changes:
- Use `as unknown`: Lint complains about the typecasting.
- Use `expect().toThrow()`: We usually use this pattern.

catloversg (approved):

d0sboots:
> The other reason I had originally set it up that way is so the error and logger calls could be easily dependency injected for unit tests... though I had forgotten that until I was already most of the way through refactoring

That makes sense... Although we can create a mock ctx for this so it ends up the same. I usually dislike making the API more awkward for testing (e.g. adding interfaces that only have one real implementation, in other languages) unless there's no other clean choice.

PR #2341, opened 2025-10-09
STANEK: Add FragmentType to NS Enums

TheCleric:
This PR allows an end user to now be able to use the `FragmentType` enum so that they can use it to identify the type of fragments available or installed in Stanek's gift.

For example prior to this PR you would need to write code like this to find all of your active non-booster Fragments to know which to charge:

```ts
const nonBoosters = ns.stanek.activeFragments().filter(x => ns.stanek.fragmentDefinitions().filter(f => x.id === f.id)[0].type !== 18); // Booster = 18
```

This was both unintuitive and a bit brittle (what if Booster isn't 18 anymore in a future update?).

With access to the enum, the user can now do the more apparent version:

```ts
const nonBoosters = ns.stanek.activeFragments().filter(x => ns.stanek.fragmentDefinitions().filter(f => x.id === f.id)[0].type !== ns.enums.FragmentType.Booster);
```


catloversg:
You should add comments to the typecasting in `fragmentDefinitions` to explain why it's needed. Every time we use typecasting, we usually add comments explaining the rationale.

Before this PR, `Fragment.type` (in `NetscriptDefinitions.d.ts`) is a number, so we can return any value, as long as it's a number. Now, `Fragment.type` has a specific set of allowed values (from 3 to 18), according to the new `FragmentEnumType`. This is different from the internal `FragmentType` (from 0 to 18). This difference exists because we want to hide the internal values (None and Delete) and the unused value (HackingChance). The typecasting here assumes that there is no bug in our codebase and the fragment type cannot be anything among the hidden values.

Personally, I don't like this kind of assumption. When we expose a set of allowed values, that set is a rule, and we should enforce that rule via either TS type system or a runtime check. The best approach here is to refactor our code and remove those hidden values so the internal `FragmentType` becomes compatible with the new `FragmentEnumType`. Doing that is non-trivial, so I think we can make some compromises here. I suggest adding some comments explaining why we need the typecasting. I can handle the refactor in another PR later.

By the way, you should update the description of this PR. For example, you can explain the usage of `ns.enums.FragmentType.Booster` instead of a magic number. It's basically what you explained on Discord when I asked you why you needed this feature. This provides the context for future contributors.

TheCleric:
> Personally, I don't like this kind of assumption. When we expose a set of allowed values, that set is a rule, and we should enforce that rule via either TS type system or a runtime check. The best approach here is to refactor our code and remove those hidden values so the internal FragmentType becomes compatible with the new FragmentEnumType. Doing that is non-trivial, so I think we can make some compromises here. I suggest adding some comments explaining why we need the typecasting. I can handle the refactor in another PR later.

Entirely valid point. For now I'll filter those values out specifically so we know the cast is always valid.

> By the way, you should update the description of this PR. For example, you can explain the usage of ns.enums.FragmentType.Booster instead of a magic number. It's basically what you explained on Discord when I asked you why you needed this feature. This provides the context for future contributors.

👍

src/NetscriptFunctions/Stanek.ts
```diff
@@ -58,7 +58,22 @@ export function NetscriptStanek(): InternalAPI<IStanek> {
...
+      return Fragments.filter(
+        (f) =>
+          // Filter out internal values
+          f.type !== FragmentType.None &&
+          f.type !== FragmentType.Delete &&
+          // Filter out unused values
+          f.type !== FragmentType.HackingChance,
+      ).map(
+        (f) =>
+          // Cast as Netscript `Fragment` because it's property of `type` does not contain:
```
catloversg:
```suggestion
          // Cast as Netscript `Fragment` because its property of `type` does not contain:
```

src/NetscriptFunctions.ts
```diff
@@ -135,6 +136,7 @@ export const enums: NSEnums = {
   OrderType,
   BladeburnerActionType,
   SpecialBladeburnerActionTypeForSleeve,
+  FragmentType,
```
d0sboots: You're exposing the internal FragmentType here, which includes fields not defined in our external type.

TheCleric: Yes but due to the down type cast to the external `FragmentEnumType` they will never be shown to a user (and since they're not functions no risk of exposing internal code). Since all the entries in `NetscriptDefinitions.d.ts` are types and not values, we'd have to create yet another copy of the enum for that to work.

d0sboots: The enums are *objects*. Types don't come in to it; at runtime, there is only JavaScript, not typescript. This exposes objects which have fields defined for None, Deleted, etc., which we don't want users to see if they iterate over it.

And yes, I am saying to do this we will need an external version of the enum object.

catloversg: One option is to wait for my PR that refactors UI code and removes all hidden values. After that, it's fine to return the internal enum (i.e., no runtime check, no typecasting) because that enum is compatible with the newly defined type. I already finished the PoC, so it should be done soon.

It's up to you two.

d0sboots: I was going to suggest something like that, honestly these values shouldn't be in the internal enum either.

catloversg (approved):

TheCleric:
So at this point is there anything we still need changed?

PR #2342, opened 2025-10-10
MISC: Make TIX access independent from WSE account

catloversg:
This PR is the alternative of #2323. For context, please check https://github.com/bitburner-official/bitburner-src/pull/2323#issuecomment-3342368822.

The breaking change should not be a problem. Before this change, the stock market is only initialized after the player purchases a WSE account. It's pointless to purchase 4S Market Data before that, so I don't think anybody does that.

Before:

before

After:

after

### DIFF
--- a/src/ScriptEditor/NetscriptDefinitions.d.ts
+++ b/src/ScriptEditor/NetscriptDefinitions.d.ts
@@ -1637,8 +1637,8 @@ export interface TIX {
   /**
    * Purchase 4S Market Data UI access (UI only).
    *
-   * You need to have WSE account. Note that this feature only unlocks access to 4S Market Data on Stock Market UI. If
-   * you want to access 4S Market Data via NS APIs, you have to unlock "4S Market Data TIX API access" via
+   * You need to have a WSE account. Note that this feature only unlocks access to 4S Market Data in the Stock Market
+   * UI. If you want to access 4S Market Data via NS APIs, you have to unlock "4S Market Data TIX API access" via
    * {@link TIX.purchase4SMarketDataTixApi | purchase4SMarketDataTixApi}, which is unrelated to this feature.
    *
    * @remarks RAM cost: 2.5 GB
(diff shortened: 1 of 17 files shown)


src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -1635,28 +1635,44 @@ export interface TIX {
   getForecast(sym: string): number;
 
   /**
-   * Purchase 4S Market Data Access.
+   * Purchase 4S Market Data UI access (UI only).
+   *
+   * You need to have WSE account. Note that this feature only unlocks access to 4S Market Data on Stock Market UI. If
```
```suggestion
   * You need to have a WSE account. Note that this feature only unlocks access to 4S Market Data in the Stock Market UI. If
```

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -1635,28 +1635,44 @@ export interface TIX {
...
    * @remarks RAM cost: 2.5 GB
    * @returns True if you successfully purchased it or if you already have access, false otherwise.
    */
   purchase4SMarketDataTixApi(): boolean;
 
   /**
-   * Purchase WSE Account.
+   * Purchase WSE account.
+   *
+   * You need to have this account to perform actions via Stock Market UI. Note that if you want to perform actions via
```
```suggestion
   * You need to have this account to perform actions via the Stock Market UI. Note that if you want to perform actions via
```

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -1635,28 +1635,44 @@ export interface TIX {
...
    * @remarks RAM cost: 2.5 GB
    * @returns True if you successfully purchased it or if you already have access, false otherwise.
    */
   purchaseWseAccount(): boolean;
 
   /**
-   * Purchase TIX API Access
+   * Purchase TIX API access.
+   *
+   * You need to have TIX API access to perform actions via NS APIs. Note that you can buy TIX API access without WSE
```
```suggestion
   * You need to have TIX API access to perform actions via NS APIs. Note that you can buy TIX API access without a WSE
```

d0sboots (commented):
Minor doc nits. The poll is close, but I'm comfortable calling it in this direction - the pricing and wording don't prevent anyone from using it or interpreting it the other way if that's what they want, so I don't see this as adding undue cognitive load.

PR #2345, opened 2025-10-10
CODEBASE: Follow-up of #2344

catloversg:
These changes should have been in #2344. This PR makes those "Over-the-top safety checks" the same in both implementations of `hack` CLI and `hack` NS API.

To be honest, these changes are useless. I thought about ignoring them but decided to commit it anyway.

### DIFF
--- a/src/Netscript/NetscriptHelpers.tsx
+++ b/src/Netscript/NetscriptHelpers.tsx
@@ -548,13 +548,14 @@ function hack(ctx: NetscriptContext, hostname: string, manual: boolean, opts: un
       let moneyDrained = server.moneyAvailable * percentHacked * threads;
 
       // Over-the-top safety checks
-      if (moneyDrained <= 0) {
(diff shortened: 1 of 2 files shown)


src/Netscript/NetscriptHelpers.tsx
```diff
@@ -548,13 +548,14 @@ function hack(ctx: NetscriptContext, hostname: string, manual: boolean, opts: un
       let moneyDrained = server.moneyAvailable * percentHacked * threads;
 
       // Over-the-top safety checks
-      if (moneyDrained <= 0) {
```
d0sboots, line 551: This is a small but real change to hack behavior in an edge case that actually comes up.

Specifically, if all the money on a server is drained, previously this would award 1/4 xp, but now it will award full xp. This changes the potential meta for xp grinding scripts, since hack is faster than grow.

PR #1171, opened 2024-03-18
IPVGO: Fix API documentation

catloversg:
Triple backticks cause trouble when we use `api-documenter`. This PR fixes it.

`<br/>` and 2 space characters after it (`  `) are intentional.
- `<br/>` create a line break.
- 2 space characters after `<br/>` create a line break when viewing in VS Code. IntelliJ IDEA does not need these ones.

Fixes #1149.

Alternative fix:
```
   * For example, a 5x5 board might look like this:
   *
   *[
   *
   *  "XX.O.",
   *
   *  "X..OO",
   *
   *  ".XO..",
   *
   *  "XXO.#",
   *
   *  ".XO.#",
   *
   *]
   *
   * Each string represents a vertical column on the board, and each character in the string represents a point.
```
This alternative fix makes the space between each line look too big, so I don't use it.

Main fix:
screenshot1

Alternative fix:
screenshot2

PR #2350, opened 2025-10-14
DOCUMENTATION: Use alternative fix for newline issue in IPvGO docs

catloversg:
In #1171, I fixed the newline issue with `<br/>` and 2 space characters. It works, but it comes with a downside: other contributors have to know that those space characters are intentional. However, some IDEs tend to remove them by default. #2137 mitigates this problem a bit.

While working on dnet PR, I see that these space characters are removed again. I think we should use the alternative fix that I mentioned in #1171. The space between each line looks too big, so I did not use it, but I changed my mind now.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -6,8 +6,5 @@ indent_size = 2
 trim_trailing_whitespace = true
 insert_final_newline = true
 
-[src/ScriptEditor/NetscriptDefinitions.d.ts]
-trim_trailing_whitespace = false
-
 [*.md]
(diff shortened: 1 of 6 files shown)


d0sboots:
I don't see this asked in the original PR: Can we solve this with &lt;pre>?

catloversg:
`<pre lang="javascript">` does not show the line break properly on GitHub and IDE(s). `<pre>` is the same.

Hover in VS Code:



Generated markdown:



View on GitHub UI:

PR #2354, opened 2025-10-17
CLI: Add --temporary flag to run command

catloversg:
This PR implements a suggestion on Discord:
```
Allow a --temporary [boolean] flag to be made universally available when launching a script from the Terminal, similar to the built-in -t [number] for thread count.

While it's entirely possible to work around this by making a mini launcher script that sets the flag, this feels like a very circumspect way of doing it.  A simple terminal flag would make more sense, in my opinion.

This is for cases where you want most of your scripts to persist (i.e. you don't want to set the 'all scripts are temporary' option), but may have a few exceptions that you would rather specifically exempt from being persisted: like a service or a controller that may be problematic if it starts up automatically on game load (or, alternatively, scripts you want to specifically allow to persist, despite having the option set for everything to be temporary). 
```

Test code:
```js
// a.js
/** @param {NS} ns */
  ns.tprint(`${ns.getScriptName()} ${ns.pid} ${ns.self().temporary}`);
  ns.run("b.js", { temporary: false });
  ns.run("b.js", { temporary: true });
}
```
```js
// b.js
/** @param {NS} ns */
  ns.tprint(`${ns.getScriptName()} ${ns.pid} (parent: ${ns.self().parent}) ${ns.self().temporary}`);
}
```

Capture


src/Terminal/HelpText.ts
```diff
@@ -417,17 +418,17 @@ export const HelpTexts: Record<string, string[]> = {
...
+    "only valid when running a script. The '-t' flag is used to indicate that the script should be run with the specified",
     "number of threads. If the flag is omitted, then the script will be run with a single thread by default. The",
-    "'--tail' flag is used to immediately open a tail window for the script being ran. And the '--ram-override'",
-    "flag is used to override the amount of ram (per thread) the script is ran with. If the script ends up using",
-    "more than that amount of ram it will crash. If any of the flags are used, then they MUST come immediately",
-    "after the script name.",
+    "'--tail' flag is used to immediately open a tail window for the script being ran. The '--ram-override' flag is used",
+    "to override the amount of ram (per thread) the script is ran with. If the script ends up using more than that",
+    "amount of ram it will crash. The '--temporary' flag is used to indicate that the script should be excluded from the",
+    "save data. If any of the flags are used, then they MUST come immediately after the script name.",
```
d0sboots: From my testing, I don't believe this part is actually true.

Optional: Mention `--` to end special args processing, if someone wants to explicitly pass one of these flags to their script.

catloversg: > From my testing, I don't believe this part is actually true.

Can you clarify with an example? Which line is wrong? Do you mean the "If any of the flags are used ..."?

d0sboots: You can use the flags in any order, interspersed with script flags.

catloversg: I don't think that sentence talks about the order between flags. It says the flags must be used after specifying the script name. For example, `run --tail a.js` is invalid.

Anyway, I rephrase it to make it clearer and mention the `--` trick.

PR #2355, opened 2025-10-18
UI: Format the shared RAM multipler on the factions

mctylr-gh:
root page. Uses default 3 decimal places, which is sufficient for showing the effect of a single thread on most(?) nodes.

Passes `npm run format:report` and `npm run lint:report`.

Before:

before

After:

after

### DIFF
--- a/src/Faction/ui/ShareOption.tsx
+++ b/src/Faction/ui/ShareOption.tsx
@@ -13,7 +13,7 @@ import {
   ShareBonusTime,
   startSharing,
 } from "../../NetworkShare/Share";
-import { formatRam } from "../../ui/formatNumber";
+import { formatRam, formatNumber } from "../../ui/formatNumber";
 import { dialogBoxCreate } from "../../ui/React/DialogBox";
 import { useCycleRerender } from "../../ui/React/hooks";
 import { roundToTwo } from "../../utils/helpers/roundToTwo";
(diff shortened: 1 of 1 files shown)


catloversg (approved):

catloversg:
Small note: I added "before/after" screenshots in the description. I recommend doing this, especially with UI changes.

d0sboots:
> Uses default 3 decimal places, which is sufficient for showing the effect of a single thread on most(?) nodes.

Your math is wrong. Without int or cores (i.e. BN1, the most likely place to run into this UI) the share bonus is simply 1 + ln(threads) / 25. Taking the derivative, this reaches .0005 between adjacent values at 80 threads.

mctylr-gh:
> Your math is wrong. Without int or cores (i.e. BN1, the most likely place to run into this UI) the share bonus is simply 1 + ln(threads) / 25.

Ah, I meant the effect of a single thread, not the difference of a single thread (e.g. 80 and 81 thread count).

So this should always display a value of `1.xxx`, (1.028 for BN1.1) i.e. not displayed as `1`, for even a single thread in BN1.1.

Is that acceptable / reasonable, or should I extend it to 6 or 9 decimal places?

d0sboots:
Let's go with 6 decimals. That makes it easier to reverse engineer the formula for those who want to do that, and it has enough precision for anyone likely to be using the UI for sharing.

mctylr-gh:
screenshot of revised share-ram interface with 6 digits

PR #2356, opened 2025-10-19
CLI: Let ServerProfiler.exe autocomplete servers

rladenson:
image


src/Terminal/getTabCompletionPossibilities.ts
```diff
@@ -231,6 +231,8 @@ export async function getTabCompletionPossibilities(terminalText: string, baseDi
         addPrograms();
         addCodingContracts();
         addScripts();
+      } else if (commandArray[1] == "ServerProfiler.exe") {
```
catloversg: - Use `===` instead of `==`.
- Use `CompletedProgramName.serverProfiler` instead of hardcoding `"ServerProfiler.exe"`.

catloversg (approved):

PR #2366, opened 2025-10-26
GO: Fix columnIndexes to include all letters A-Z

FakeApate:
When hovering over a cell in IPvGO game the column I is skipped.

I did **not** test the change.
I did **not** run `npm run format` and `npm run lint`.
I just added the missing letter.


d0sboots:
This is intentional, in keeping with the practice of how go boards are numbered. I is skipped to avoid confusion with 1 in move notation.

PR #2367, opened 2025-10-27
API: Move and rename purchased server functions to cloud API

gmcew:
This is my attempt at satisfying the [survey ](https://discord.com/channels/415207508303544321/415213435974975508/1426228440025268348) to do this.

The changes:
- Create `ns.cloud`. The name `cloud` fits really well lore-wise for why the player can freely buy cheap, ever-increasing server capacity in random shops around the world or via scripts. It also aligns with the fact that they aren't associated with any faction/organisation/location whilst still justifying loss on reset. This can be changed to anything else that's suitable, however.
- Moved all 9 purchased Server functions over, and added "cloud" to all of them to avoid name clashes e.g. `getServerMaxRam`. Functionality is as before.
- Change `bitnode multipliers` and `server constants` wording for consistency. This can be reverted - I tried to keep non-player facing values untouched where possible, and only changed the bits that started to look confusing by cross-referring to both cloud/purchased servers.
- Change `server`, `ram` and `getting started` docs for consistency, and the `LiteratureName.HackersStartingHandbook` .lit file. Let me know if I missed a `purchasedServer` reference anywhere.
- Added changes to 3.0.0 API Break and `setRemovedFunctions` in NetscriptFunctions.js. With how invasive this change is, there may be a case to not remove the originals yet.
- A test file is changed - the singularity test uses a purchased/cloud server to check `connect`; I don't see an issue with changing that test in this scenario - please let me know if that's a mistake.

Tested by
- running tutorial `purchase-server-8gb.js` script, and my own later-game one to ensure functionality is as before,
- buying cloud servers manually using vendors,
- deleting them all using script, and checked all deleted functions gave correct errors,
- Imported a completed game save to ensure the auto-transfer of functions works,

Ran `npm run test`, `npm run doc`, `npm run lint` and `npm run format`.


src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -755,13 +755,13 @@ interface BitNodeMultipliers {
    */
   ManualHackMoney: number;
   /** Influence how much it costs to purchase a server */
-  PurchasedServerCost: number;
+  CloudServerCost: number;
```
d0sboots, line 758: Throughout this block: Use cloud in descriptions

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -4073,6 +4073,143 @@ export interface CodingContract {
...
+   * @example
+   * ```js
+   * const ram = 2 ** 20;
+   * const cost = ns.cloud.getCloudServerCost(ram);
+   * ns.tprint(`A cloud server with ${ns.format.ram(ram)} costs $${ns.format.number(cost)}`);
+   * ```
+   * @param ram - Amount of RAM of a potential cloud server, in GB. Must be a power of 2 (2, 4, 8, 16, etc.). Maximum value of 1048576 (2^20).
+   * @returns The cost to purchase a cloud server with the specified amount of ram, or returns Infinity if ram is not a valid amount.
+   */
+  getCloudServerCost(ram: number): number;
```
d0sboots: I think this should just be "getServerCost", same with the other ones. "Cloud" is already in the namespace, so there is no need for it in the function name. I don't think the name being the same is an issue across the upgrade is an issue, and actually it should help with people who have to do it manually for whatever reason. 

The only concern is if any of the names become too generic such that they ram-conflict. The most likely would be getServers, but I think it's unlikely that we'll add an equivalently named function. 

However, since this involves a lot of renaming I'd hold off to see what Catlover thinks too.

gmcew: `getServerMaxRam` conflicts - I did have them cloud-free at first and couldn't think of a better alternative for this one so went the opposite way. I'll await CatLover's thoughts on solving this too. Thank you!
EDIT: with hindsight maybe simply `cloud.getMaxRam` makes sense, and drop `server` from most function names too?
May also be worth noting that both pServ/cloud and base ns `getServerMaxRam` are both only 0.05GB and comparatively cheap functions.

d0sboots: Although they do different things, both versions of `getServerMaxRam` already have the same ram cost, so there actually is no conflict there!

Edit: But I agree, it's also seems like dropping the "server" could make sense for many of them. (Can't do it for "getServers", but for others potentially.

gmcew: They have the same cost, but if both are used together you get one for free (I won't admit how much confusion that caused in my first round of testing). It's only 0.05GB I know, but it's technically incorrect, which is the worst kind!

To summarise what I think we're leaning towards from this, using original function names:
- getPurchasedServerCost -> cloud.getCost
- purchaseServer -> cloud.purchase
- getPurchasedServerUpgradeCost -> cloud.getUpgradeCost
- upgradePurchasedServer -> cloud.upgrade
- renamePurchasedServer -> cloud.rename
- deleteServer -> cloud.delete
- getPurchasedServers -> cloud.getServers
- getPurchasedServerLimit -> cloud.getLimit
- getPurchasedServerMaxRam -> cloud.getMaxRam

I'll see what CatLover's thoughts are and revise it hopefully just the once from here.

catloversg: > To summarise what I think we're leaning towards from this, using original function names:
> 
>     * getPurchasedServerCost -> cloud.getCost
> 
>     * purchaseServer -> cloud.purchase
> 
>     * getPurchasedServerUpgradeCost -> cloud.getUpgradeCost
> 
>     * upgradePurchasedServer -> cloud.upgrade
> 
>     * renamePurchasedServer -> cloud.rename
> 
>     * deleteServer -> cloud.delete
> 
>     * getPurchasedServers -> cloud.getServers
> 
>     * getPurchasedServerLimit -> cloud.getLimit
> 
>     * getPurchasedServerMaxRam -> cloud.getMaxRam

I agree with these new names. My only minor concern is how easy it is to conflict with future APIs due to the common usage of these verbs/phrases. For example, if something can be upgraded with a cost, it's intuitive to have the "getUpgradeCost" API. `getCloudServerUpgradeCost` is too long, but we don't need to worry about name conflicts.

Between `ns.cloud.getCloudServerUpgradeCost` and `ns.cloud.getUpgradeCost`, I'm a bit lean towards `ns.cloud.getUpgradeCost`. The problem of name conflicts in the future may be an acceptable cost to make these APIs' name concise (maybe I'm wrong, though).

catloversg:
Regarding the `cloud` namespace, if we want to use `cloud` as an alternative way of saying "temporary servers bought by the player and hosted somewhere else", it's okay. However, I want to discuss a bit more about another potential usage of the `cloud` namespace.

The "temporary servers bought by the player and hosted somewhere else" concept is fairly close to VPS (Virtual Private Server), and VPS is only one of many services that cloud providers usually offer. For example, AWS provides EC2 (Elastic Compute Cloud), which is a fancy name for renting virtual machines. If we add a feature that looks similar to one that cloud providers usually offer (e.g., something like AWS Lambda), I think it's cool to group its APIs in a sub-namespace of `cloud`.

If we choose to use the `cloud` namespace as an "umbrella" of "cloud services like in real life", the current pserver APIs should be in `ns.cloud.vps` or something else. If we want to be simple (i.e., "cloud" = "purchased servers"), putting those APIs in `ns.cloud` is okay.

src/ScriptEditor/NetscriptDefinitions.d.ts
gmcew: Thanks for the feedback; it does risk becoming too generic which is an issue we've mostly avoided thus far. With the other observation that `cloud` can be expanded to more than just purchased servers - I feel the `server` bit should remain to avoid future conflicts, and to keep with current naming conventions in other APIs.

Revised proposal is:
* getPurchasedServerCost -> cloud.getServerCost

* purchaseServer -> cloud.purchaseServer

* getPurchasedServerUpgradeCost -> cloud.getUpgradeServerCost

* upgradePurchasedServer -> cloud.upgradeServer

* renamePurchasedServer -> cloud.renameServer

* deleteServer -> cloud.deleteServer

* getPurchasedServers -> **cloud.getServerNames**

* getPurchasedServerLimit -> cloud.getServerLimit

* getPurchasedServerMaxRam -> **cloud.getRamLimit**


`getPurchasedServers` I added "Name" since `cloud.getServers` feels short enough to risk confusion with `ns.getServer` in terms of what each does.
`getServerMaxRam` conflicts as discussed above. Since we have a server count limit in the API, RAM limit seems like a reasonable parallel.

If you're both OK with this, I'll make the changes.

catloversg: I'm fine with both proposals (the concise one and the latest one). I'll let d0sboots make the call.

Edit: In the latest proposal, I think `cloud.getUpgradeServerCost` should be `cloud.getServerUpgradeCost`.

By the way, an optional suggestion: In `src\NetscriptFunctions.ts`, with each `replacement` in `setRemovedFunctions`, you can use `"ns.cloud.something()"` instead of `"ns.cloud.something"`. We used parentheses in functions of `format` and `ui` namespace.

d0sboots: I'm glad we spent time workshopping the names, I think they've come out a lot better as a result. I like the latest version a lot, with the revision to getServerUpgradeCost.

I'm not too worried about us adding stuff to cloud in the future, but I was a little worried about "delete" and "rename" possibly colliding in the future, so I think this is the right balance between specificity and verbosity.

src/Netscript/RamCostGenerator.ts
```diff
@@ -226,6 +226,18 @@ const singularity = {
   getUnlockedAchievements: SF4Cost(RamCostConstants.SingularityFn3),
 } as const;
 
+const cloud = {
+  getCloudServerLimit: RamCostConstants.GetPurchasedServerLimit,
```
d0sboots: These constants are misnamed now, and having them in extra constants just makes the RAM costs harder to follow. I'd take the opportunity to inline these.

src/NetscriptFunctions/Cloud.ts
```diff
@@ -0,0 +1,210 @@
+import { Cloud } from "@nsdefs";
+import { InternalAPI } from "src/Netscript/APIWrapper";
+import { helpers } from "../Netscript/NetscriptHelpers";
+import { Player } from "@player";
+import {
+  getPurchasedServerUpgradeCost,
```
d0sboots: Optional: Rename these too for consistency

src/utils/APIBreaks/3.0.0.ts
```diff
@@ -391,5 +391,52 @@ export const breakingChanges300: VersionBreakingChange = {
         "Note that this change does not affect 4S Market Data TIX API access (ns.stock.purchase4SMarketDataTixApi()).",
       showWarning: false,
     },
+    {
```
d0sboots, line 394: There are BN multipliers and such that also got renamed and need to be migrated.

d0sboots (commented):
OK, now that the making is settled I've looked over the whole thing. It looks good, but ofc with something like this the bigger danger is missing a needed change, which the diffs won't show.

src/utils/APIBreaks/3.0.0.ts
gmcew: Thanks for the thorough review - I'm about to re-test with these changes.
For clarification - I've put the show warning to true for the `cloud` changes, but not for the BN mults one as I don't think it's on the same level of change - but I don't know how to justify that position.
Is that correct? Is there a threshold for what needs a warning?

d0sboots: Hopefully catlover will double-check the api break stuff since he has the most experience with it, but IIRC we haven't been showing a warning for things that can be automigrated cleanly.

catloversg: `showWarning` is usually false. We only set it to true when the breaking change may break the player's scripts, but we cannot auto-migrate them.

After gmcew finishes everything, I'll review the entire PR.

gmcew:
Thanks again both, the feedback should now be incorporated. Same tests as above with additional import of a script using `getBitNodeMultipliers` to check the conversion - I hope I didn't miss anything. Edit: I missed some things.
A few minor notes:
- the BN mults replacement in the API breaks file is deliberately verbose since it otherwise replaces the same-titled purchased server functions and therefore doesn't auto-replace properly.
- `cloud.getServerNames` is technically incorrect if the player has the setting for IPAddresses for the return instead of Hostnames, but that's such a minor inaccuracy I don't see real value in changing it to `getServerIDs` or anything else novel.

src/Documentation/doc/en/help/getting_started.md
```diff
@@ -348,11 +348,12 @@ Next, we're going to create a [Script](../basic/scripts.md) that automatically p
 These cloud [Servers](../basic/servers.md) will be used to run many [Scripts](../basic/scripts.md).
 Running this [Script](../basic/scripts.md) will initially be very expensive since purchasing a cloud [Server](../basic/servers.md) costs money, but it will pay off in the long run.
 
-In order to create this [Script](../basic/scripts.md), you should familiarize yourself with the following functions in the [Cloud API](<(https://github.com/bitburner-official/bitburner-src/blob/stable/markdown/bitburner.cloud.md)>):
+In order to create this [Script](../basic/scripts.md), you should familiarize yourself with the following functions, some of which are in the [Cloud API](<(https://github.com/bitburner-official/bitburner-src/blob/stable/markdown/bitburner.cloud.md)>):
 
-- `purchaseCloudServer()`
-- `getCloudServerCost()`
-- `getCloudServerLimit()`
+- `purchaseServer()`
```
d0sboots: I would add `cloud.` prefix to these to be specific.

src/utils/APIBreaks/3.0.0.ts
```diff
@@ -391,50 +391,87 @@ export const breakingChanges300: VersionBreakingChange = {
...
+      brokenAPIs: [
+        {
+          name: "ns.getBitNodeMultipliers().PurchasedServerCost",
+          migration: {
+            searchValue: "ns.getBitNodeMultipliers().PurchasedServerCost",
```
d0sboots: These all match much too narrowly. Many players rename their ns, or store functions in variables, or (most common) get BN multipliers once and store that in a variable. Also, this can be used as a string key in lookups of the form `mults["PurchasedServerCost"]`.

For all those reasons, the search string must be only the mult name, and if that is not unique then we'll have to come up with a different plan for that specific one.

src/utils/APIBreaks/3.0.0.ts
```diff
@@ -391,50 +391,87 @@ export const breakingChanges300: VersionBreakingChange = {
...
+          name: "ns.getBitNodeMultipliers().PurchasedServerMaxRam",
+          migration: {
+            searchValue: "ns.getBitNodeMultipliers().PurchasedServerMaxRam",
+            replaceValue: "ns.getBitNodeMultipliers().CloudServerMaxRam",
+          },
+        },
+      ],
+      info:
+        "ns.getBitNodeMultipliers() Purchased Server properties have been renamed to CloudServers with the creation of the cloud API.\n" +
+        "They have been automatically replaced with the corresponding interface variants.",
```
d0sboots: Although it is verbose to write out here, you should explicitly list what those replacements are. This is likely the only place where the player will see this info, to help their brain make the transition. See the other messages for examples. (Applied to both)

d0sboots:
There is at least one other migration that touches pservers. You should double check the list and adjust the messages so that they use the post-replacement names. (And make sure the conflicts work properly)

src/utils/APIBreaks/3.0.0.ts
gmcew: With hindsight that's an obvious error, I'm sorry. To resolve the `PurchasedServerCost`, `PurchasedServerLimit` and `PurchasedServerMaxRam` conflicts, moving the `ns` function checks before the BN mults ones replaces them correctly.

I'm not smart enough to really understand why - or if that's a rigorous solution to the issue - but I think from how the `detectImpactAndMigrateLines` function works it should hold up? I'd appreciate your thoughts/explanation to ensure I don't waste your time next commit!

gmcew:
> There is at least one other migration that touches pservers. You should double check the list and adjust the messages so that they use the post-replacement names. (And make sure the conflicts work properly)

[This](https://github.com/bitburner-official/bitburner-src/blob/aaf106bf9a2577295017728828584168da3f049b/src/utils/APIBreaks/3.0.0.ts#L293) change was made to throw instead of return a default value; I don't see any other pServer migrations affected by this. Is it correct to change that description to the new `cloud` space? Since it runs before the function replacement section, it misses the affected files if those references are changed to the new form.

CatLover may know the detail on other migrations best - I'll also await his guidance on what else I've missed!

d0sboots:
> > There is at least one other migration that touches pservers. You should double check the list and adjust the messages so that they use the post-replacement names. (And make sure the conflicts work properly)
> 
> [This](https://github.com/bitburner-official/bitburner-src/blob/aaf106bf9a2577295017728828584168da3f049b/src/utils/APIBreaks/3.0.0.ts#L293) change was made to throw instead of return a default value; I don't see any other pServer migrations affected by this. Is it correct to change that description to the new `cloud` space? Since it runs before the function replacement section, it misses the affected files if those references are changed to the new form.
> 
> CatLover may know the detail on other migrations best - I'll also await his guidance on what else I've missed!

Yup, that's the one I saw. As you said, it runs first so you can't change the search value, but I think the description should be changed. (That check doesn't actually migrate any code.)

src/utils/APIBreaks/3.0.0.ts
d0sboots: Yeah, the migrations are applied in order (and we can rely on that), so this should be fine. The only the case is in word cases like ns renaming followed by the BN migration falsely triggering, but even there it would be a manual fix either way and it won't get mangled.

gmcew:
I think I'm all caught up on comments now, sorry for the silence. If there's any further revisions needed I'll attempt to resolve them this weekend.

d0sboots:
Looks good to me now.

catloversg:
Sorry for the delay. I'm taking a short break, so I have not taken a look at recent PRs. I'll check this one later today or this weekend.

catloversg:
@d0sboots Please review my commits.

By the way, I found some problems with our current behaviors of purchase/rename/delete server APIs.
Test code:
```js
/** @param {NS} ns */
  ns.ui.openTail();
  // "  abc  " will be automatically changed to "abc" with "/\s+/g" regex
  ns.cloud.purchaseServer("  abc  ", 8);
  // But it's fine to rename it to "  abc  ". renameServer does not use any regex.
  ns.cloud.renameServer("abc", "  abc  ");
  // Error. deleteServer uses a different regex: "/\s\s+/g". This means the player will never be able to delete this server. They need to rename it to a "valid" hostname before being able to delete it.
  ns.cloud.deleteServer("  abc  ");
}
```
What do you think about using `/\s+/g` in all 3 APIs (for consistency) or `purchaseServer+renameServer`? It's a breaking change, so we need to do it now (in a different PR, of course). I don't think it will affect many players.

d0sboots:
Definitely they should be consistent. With what's going on in the new BN (lots of strange hostnames), I'd be inclined to remove the regex, at least on those functions. Probably it should be shifted to the UI dialog, where accidental leading/trailing whitespace is more likely (and it's OK to not be able to specify all hostnames).

PR #2369, opened 2025-10-28
CODEBASE: Prepare for save data migration of next beta version

catloversg:
We are going to migrate the save data for the next breaking changes (e.g., move pserver APIs, darknet). `showAPIBreaks` is called in `ver < 44`, so the scripts of current beta players won't be migrated in the next beta version.

Ideally, we should split `3.0.0.ts` into many files, with each beta version per file, but I don't think it's worth the effort. It's okay to run those script migrations many times.

### DIFF
--- a/src/utils/SaveDataMigrationUtils.ts
+++ b/src/utils/SaveDataMigrationUtils.ts
@@ -551,6 +551,7 @@ Error: ${e}`,
     }
     if (found) Terminal.error("Filenames with whitespace found and corrected, see console for details.");
   }
+  // Migrate save data related to the breaking changes in the first beta of v3.0.0.
   if (ver < 44) {
     try {
       /**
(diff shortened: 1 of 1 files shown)


d0sboots:
How is this going to interact with the issue of "I just want to stop seeing this huge list every time I start?"

I'm trying to balance work we're doing now that is strictly temporary with appropriately displaying the actual messages relevant to the situation.

catloversg:
That issue is only relevant with people playing a custom build that is built from the latest commit (e.g., contributors testing their PRs or playtesters testing the dnet PR). In these cases, they can disable all annoying popups/warning messages in their build (this is what I do) or set the version number to 45 early.

With the public beta testers (https://bitburner-official.github.io/bitburner-src or Steam Beta), this is not a problem. When we release the next beta version, we will increase the version number to 45.

PR #2373, opened 2025-10-28
API: Make ActiveFragment extend Fragment

catloversg:
It's just as the title said.

This API's naming and behavior are not ideal. It's weird to return a union type of `ActiveFragment` and `Fragment` when the API's name is `activeFragments`, but for convenience, it's arguably acceptable.


d0sboots:
Why not have ActiveFragment extend Fragment, and then have both methods return the new, more useful type? getFragment returning something different is *also* confusing.

catloversg:
It looks a bit weird on our side because the internal `ActiveFragment` class does not extend the `Fragment` class. The discrepancy may be confusing for new contributors. Nonetheless, I make `ActiveFragment` extend `Fragment` in NS API.

PR #2374, opened 2025-10-30
DOCUMENTATION: Improve wording around host argument of getScriptRam

nickmshelley:
Although "target" is technically accurate here, I found it confusing because this function takes a script as a parameter, and I assumed "target server" was what that script would be running on, and therefore the ram cost would depend on which server the script in question would be run on.

### DIFF
--- a/src/ScriptEditor/NetscriptDefinitions.d.ts
+++ b/src/ScriptEditor/NetscriptDefinitions.d.ts
@@ -7969,7 +7969,7 @@ export interface NS {
    * Returns 0 if the script does not exist.
    *
    * @param script - Filename of script. This is case-sensitive.
-   * @param host - Hostname/IP of the server the script is located on. Optional. Defaults to the server the calling script is running on.
+   * @param host - Hostname/IP of the server the target script is located on. Optional. Defaults to the server the calling script is running on.
    * @returns Amount of RAM (in GB) required to run the specified script, and 0 if the script does not exist.
    */
   getScriptRam(script: string, host?: string): number;
--- a/src/ScriptEditor/NetscriptDefinitions.d.ts
+++ b/src/ScriptEditor/NetscriptDefinitions.d.ts
@@ -7969,8 +7969,8 @@ export interface NS {
    * Returns 0 if the script does not exist.
    *
    * @param script - Filename of script. This is case-sensitive.
-   * @param host - Hostname/IP of target server the script is located on. Optional. Defaults to the server the calling script is running on.
-   * @returns Amount of RAM (in GB) required to run the specified script on the target server, and 0 if the script does not exist.
+   * @param host - Hostname/IP of the server the script is located on. Optional. Defaults to the server the calling script is running on.
+   * @returns Amount of RAM (in GB) required to run the specified script, and 0 if the script does not exist.
    */
   getScriptRam(script: string, host?: string): number;
 
--- a/markdown/bitburner.ns.getscriptram.md
+++ b/markdown/bitburner.ns.getscriptram.md
@@ -58,7 +58,7 @@ string
 
 </td><td>
 
-_(Optional)_ Hostname/IP of target server the script is located on. Optional. Defaults to the server the calling script is running on.
+_(Optional)_ Hostname/IP of the server the target script is located on. Optional. Defaults to the server the calling script is running on.
 
 
 </td></tr>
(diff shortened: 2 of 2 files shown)


src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -7969,8 +7969,8 @@ export interface NS {
    * Returns 0 if the script does not exist.
    *
    * @param script - Filename of script. This is case-sensitive.
-   * @param host - Hostname/IP of target server the script is located on. Optional. Defaults to the server the calling script is running on.
-   * @returns Amount of RAM (in GB) required to run the specified script on the target server, and 0 if the script does not exist.
+   * @param host - Hostname/IP of the server the script is located on. Optional. Defaults to the server the calling script is running on.
```
catloversg: Removing "on the target server" in the next line is okay, but I'm not sure about this line. We use the phrase "target server" in many APIs that have a host param. Is it really confusing? If yes, we may need to change other APIs as well.

nickmshelley: I was typing a big response about why "target server" means something different in this case, but the more I looked at other examples, the more I came to understand that it semantically means the same thing here as in the other cases. So I'm fine to change it back if you think that's best.

But at the same time, it took me trying to write a response to your valid critique to finally understand why it doesn't mean what I thought it meant up to this point. And in thinking about how to explain my confusion, the best I can come up with is that, the "target server" is often a server different than the one the script is located on, and attributes of the target server affect the result of the call.

Again, after thinking about it, I realize that it's the same here, where the "target server" can be different from where we're calling the getScriptRam function, and the result depends on whether the script exists on that target server. So I realize that I'm wrong. But I think where the confusion came for me was that this case is unique in that there are two scripts involved, one that's being run and one that's being queried against. And in my mind I mixed them up. So when I read "target server" I assumed that the ram cost could change depending on what server the script in question (the first param) is being run on. But I couldn't see what might affect that, the only thing I could think of was maybe cores if the script contained certain functions that made use of them. I didn't realize that the only thing that mattered was where the script is located, even though the docs clearly say that. It just wasn't what I was used to when seeing and thinking about "target server."

Anyway, I don't know what the right thing is or even if others have been or will be confused by similar terminology. That's just why I think I was confused and why I thought it best to remove all uses of the word "target." I'm happy to change it, or if it's easier, feel free to just close this PR and reword it as you see fit (or leave it as is). I'm not worried about contributor credit or anything. 🙂

catloversg: I appreciate your feedback. When it comes to documentation, it's always good to hear different opinions from many players.

I think the "target" word in "target server" is fairly useful in some cases. For example, when script A on server X wants to do something on/against server Y, saying "Hostname/IP of the target server" is concise, and it's good enough to imply that "target server" is server Y. In this API, we can remove the "target" word as you said because we also have a clause clarifying which server we are talking about ("the script is located on"). I raised this question because I wanted to know if there is a serious problem with the way we use this phrase in so many APIs. To keep it consistent with other APIs, I suggest keeping it as-is (i.e., "target server" instead of "the server"). To be clear, your suggestion is not a bad one. Let's wait a bit to see what other people think.

gmcew: > Again, after thinking about it, I realize that it's the same here, where the "target server" can be different from where we're calling the getScriptRam function, and the result depends on whether the script exists on that target server.

As a supporting opinion, the original confusion is easy to see, and likewise CatLover's observation for not wanting to lose the `target` naming convention where useful. 

I think you've nicely summarised the crux of the issue in that it's not about target server really, it's about a target script (on either different server or current host).

`of the server the target script is located on`, `target script on the specified server` and/or similar phrasing to make the script the target not the server may be cleaner wording?

In terms of housekeeping - it would be useful to summarise the reasoning for the change in the PR description to make the maintainers' lives easier! The failing test will be fixed once wording is confirmed and you run `npm run doc` etc.

nickmshelley: Updated the description (I think). Once I get confirmation on final wording I'll update the docs.

d0sboots: I agree with `of the server the target script is located on`. Basically, just shift the "target" to be more clear. Edit: Otherwise, "the script" can be slightly ambiguous because it could refer to the current (running) script.

nickmshelley:
@catloversg Thanks for seeing this one through, sorry I went MIA for a bit.

PR #2378, opened 2025-10-31
MISC: Support css file type

shyguy1412:
for a long time now we have been supporting react usage with no good way of styling your UI.

This PR adds a function to load a CSS file during the lifetime of a script.

CSS files are added to a new Map on the workerscript the first time they are loaded. Afterwards their CSSStyleSheet content just gets replaced. In the cleanup for the workerscript all stylesheets associated with that workerscript get removed from the adoptedStyleSheets array. This way stylesheets dont accumulate over time.

https://github.com/user-attachments/assets/efb7d712-9dd3-487b-944c-a7939d4c2cf1


shyguy1412:
darn, test suite doesnt seem to polyfill the adoptedStyleSheets

d0sboots:
This is a very neat idea! And I agree that the existing styling is a pain point, and one that we should address. However, I have a number of concerns about this approach:

1. I don't think the `.css` filetype is needed. It makes the API less flexible by requires you to write static css files, making dynamic css harder. I would just take a string as the content, and then no new filetype is required.
2. This feels like something that belongs in a user-library. It doesn't actually require any integration with the game itself to accomplish; everything that you need is DOM calls and `ns.atExit`.
3. I'm not sure that we should be promoting this as "the" official solution. Even internal to BB, our approach to styling is somewhat up in the air. If/when we upgrade MUI (which is needed to upgrade React), it's very possible that aspects of how we do styling will change. But even today, the way we handle styling and theming is very very far away from using a static CSS file. Which, maybe, is a bad thing! My point is, if we release this as an API, it ties our hands and this becomes something we have to continue supporting, but our approach to styling is something where I think we want maximum flexibility.
4. Sort of as an extension to 3, using static stylesheets is a pretty crappy experience. All the various elements in our UIs are styled by MUI, which generates an alphabet-soup of dynamic css-classes to style them. There are more generic classes with user-guessable names that can be used, but if you touch those you will be altering the styling for *all* of BB. In short, I feel like a better solution would be one that would allow users to reuse our components (along with their themed styling) more easily, but I only have a vague idea of how that would work.

catloversg:
I agree with d0sboots, especially the second and third points. I think this is a good idea for #community-projects, but not a good one for NS API.

shyguy1412:
> I don't think the .css filetype is needed. It makes the API less flexible by requires you to write static css files, making dynamic css harder. I would just take a string as the content, and then no new filetype is required.

Having an official solution for CSS and then using .txt files _and_ having no css language features in the ingame editor seems kinda hacky. I really dont think thats the way we should go. Even if there is no official way of loading CSS I think just allowing .css files ad having CSS language support is already a good addition. We shouldn't be supporting writing CSS files as TXT.

> This feels like something that belongs in a user-library. It doesn't actually require any integration with the game itself to accomplish; everything that you need is DOM calls and ns.atExit.

I can see your point. If you have a better suggestion to address users styling their own UI, please share

> I'm not sure that we should be promoting this as "the" official solution. Even internal to BB, our approach to styling is somewhat up in the air. If/when we upgrade MUI (which is needed to upgrade React), it's very possible that aspects of how we do styling will change. But even today, the way we handle styling and theming is very very far away from using a static CSS file. Which, maybe, is a bad thing! My point is, if we release this as an API, it ties our hands and this becomes something we have to continue supporting, but our approach to styling is something where I think we want maximum flexibility.

I do not get your point. It does not matter to a player how we do our own, internal, CSS. This is explicitly to allow users to style their _own_ UI, for which a static CSS file is the simplest and most "correct" solution. If it was easier to scope CSS to some elements I'd agree with doing so, unfortunately it is not and as a side effect you most likely will always be able to style game UI with custom CSS files. I see this similarly to using a querySelector and modifying the DOM.

> Sort of as an extension to 3, using static stylesheets is a pretty crappy experience. All the various elements in our UIs are styled by MUI, which generates an alphabet-soup of dynamic css-classes to style them. There are more generic classes with user-guessable names that can be used, but if you touch those you will be altering the styling for all of BB. In short, I feel like a better solution would be one that would allow users to reuse our components (along with their themed styling) more easily, but I only have a vague idea of how that would work.

I also do not get the point. Players assign their own classes and then use them. If a player modifies game UI and uses a classname with querySelector and the classname changes, that is on them. I see no difference between that and re-styling ingame elements.

d0sboots:
> I do not get your point. It does not matter to a player how we do our own, internal, CSS. This is explicitly to allow users to style their own UI, for which a static CSS file is the simplest and most "correct" solution.

This is where I fundamentally disagree. The way BB works internally, we do not use *any* static CSS stylesheets. This is not to say that what BB does dictates what the user must do, but rather as evidence of how React apps are constructed: whether or not they use static stylesheets heavily depends on the particular component library used (or not used), and you cannot say *any* particular approach is obviously the "correct" one. 

An additional wrinkle is that a *very* common pain point I have seen come up, over and over, is people who want their custom components to be styled the same as the game. For this, static stylesheets are worse than useless, they are actively harmful (because of how the game's theme system works). Ofc, other people don't care about this at all. But we can't say either is the "correct" view.

If printRaw dealt in raw html nodes and not React, and our existing theming was based on global style classes with well-known names, I would accept this in a heartbeat. But that's not the situation we're in. :(

shyguy1412:
>. This is not to say that what BB does dictates what the user must do, but rather as evidence of how React apps are constructed: whether or not they use static stylesheets heavily depends on the particular component library used (or not used), and you cannot say any particular approach is obviously the "correct" one.

While that may be technically correct, our players do not have access to a component library. Or any library in fact, unless they are using an external editor. In such a case, and under the assumption that we wont expose MUI, I feel it is reasonable to say that with just plain react static css is the goto way of handling styling.

> An additional wrinkle is that a very common pain point I have seen come up, over and over, is people who want their custom components to be styled the same as the game. For this, static stylesheets are worse than useless, they are actively harmful (because of how the game's theme system works).

I strongly disagree. The game has a very simplistic styling that is easily replicated with very simple CSS. The only issue is colors, which we currently only expose via ns.ui.getTheme.

The way I worked around this in the past is by translating the theme returned by getTheme into css custom properties. This way I can style my components with the currently applied theme and achieve a uniform look.

<https://github.com/shyguy1412/bb-scripts/blob/df12e06bdf3939763ed90956f94ad19ea379343a/src/apps/Plasma/DesktopEnviroment.tsx#L12>

If we did that from the games side, making the theme globally available as css custom properties and auto-updating on theme changes, players could easily create designs that conform to the games look even with just simple static css files.

Additionally, considering bitburner sees a lot of players with very little to no coding experience, I think it is especially reasonable to add the simplest form of styling with the most accessible documentation. We already are committed to custom UI via react, there is no going back on that, however I dont think that should be a reason to make anything else about custom UI more complex than it needs to be.

d0sboots:
>> . This is not to say that what BB does dictates what the user must do, but rather as evidence of how React apps are constructed: whether or not they use static stylesheets heavily depends on the particular component library used (or not used), and you cannot say any particular approach is obviously the "correct" one.

> While that may be technically correct, our players do not have access to a component library. Or any library in fact, unless they are using an external editor. In such a case, and under the assumption that we wont expose MUI, I feel it is reasonable to say that with just plain react static css is the goto way of handling styling.

Players *absolutely* have access to component libraries, if they wish. You can import MUI, or anything else. It doesn't even require an external editor, just use `wget` inside the game and pull it in. Edit: You actually don't even need wget, you can also reference it via URL from a CDN. You can also use some mild hackery to reference BB's version of MUI.

It's also not safe to say that we won't expose MUI, since that's actually something I've contemplated doing on multiple occasions. The main thing holding me back is the same concern as this PR: That it would lock us in to supporting a particular library (and a particular version of that library). But if/when we upgrade to a more modern version, it's certainly on the table.

> If we did that from the games side, making the theme globally available as css custom properties and auto-updating on theme changes, players could easily create designs that conform to the games look even with just simple static css files.

Yes, *if* such a system existed, then investing more in static CSS would make sense. However, "whether we should build such a system" is the entire *point* of this discussion; you can't use it to support a particular hypothetical design. From my perspective, talking about *that* system is actually much more interesting, since it is something that can't easily be built directly out of the `ns` calls that we have right now.

> Additionally, considering bitburner sees a lot of players with very little to no coding experience, I think it is especially reasonable to add the simplest form of styling with the most accessible documentation. We already are committed to custom UI via react, there is no going back on that, however I dont think that should be a reason to make anything else about custom UI more complex than it needs to be.

For better or worse, custom UI will always be a complicated topic, much more complicated than simply getting a gang running. It requires having a script running in the background to handle your input, understanding how React works... there's a lot going on even to get to "Hello World." I don't consider the "newbie" argument very persuasive, because from my perspective doing UI customization in BB is *far* from a newbie task.

shyguy1412:
>Yes, if such a system existed, then investing more in static CSS would make sense. However, "whether we should build such a system" is the entire point of this discussion; you can't use it to support a particular hypothetical design. From my perspective, talking about that system is actually much more interesting, since it is something that can't easily be built directly out of the ns calls that we have right now.

Should such a system be part of this PR or should I open a different one for that and then circle back to this PR to decide if we move forward with static css?

d0sboots:
I would do that as a separate PR, but I think more productive would be for us to discuss design on discord first, so we can align on *what* we want to build.

shyguy1412:
another idea:

If #2382 was to be merged first the function could work by receiving a CSSStyleSheet that the player can just import from their CSS file. With dynamic imports from URL's this could also allow for loading stuff like tailwind from a CDN

Edit: actually, if CSS imports work we really do not need this function anymore as applying a CSSStyleSheet is beyond trivial so I'd think just allowing the .css filetype would be plenty good

d0sboots:
> Edit: actually, if CSS imports work we really do not need this function anymore as applying a CSSStyleSheet is beyond trivial so I'd think just allowing the .css filetype would be plenty good

That was sort of the conclusion I was coming to, potentially even without CSS imports. Let's either strip this down to just the filetype, or do that in another PR - I think that can stand alone. The other part is basically just CSSStyleSheet + atExit, and I think that's small enough that it doesn't belong as a built-in.

shyguy1412:
I'll strip out the new function from this PR and just leave it at adding .css file support. Thats really the thing that was most important to me anyway. Having a built in convenience function would have been neat but I'll survive without :v

d0sboots:
There's a lot of docs that need to be updated, too. Do `grep -r "\.txt" src/` for a good overview of all the places that mention .txt files, to find them.

src/NetscriptFunctions.ts
```diff
@@ -1611,7 +1611,7 @@ export const ns: InternalAPI<NSFull> = {
     ) {
       throw helpers.errorMessage(
         ctx,
-        `'mv' can only be used on scripts (.js, .jsx, .ts, .tsx) and text files (.txt, .json)`,
+        `'mv' can only be used on scripts (.js, .jsx, .ts, .tsx) and text files (.txt, .json, .css)`,
```
d0sboots, line 1614: Error for `scp` also needs updating. Probably it should say "Only works for script, text, and .lit files." at this point.

src/Terminal/HelpText.ts
```diff
@@ -142,7 +142,7 @@ export const HelpTexts: Record<string, string[]> = {
   cat: [
     "Usage: cat [file name]",
     " ",
-    "Display message (.msg), literature (.lit), script (.js, .jsx, .ts, .tsx), or text (.txt, .json) files. Examples:",
+    "Display message (.msg), literature (.lit), script (.js, .jsx, .ts, .tsx), or text (.txt, .json, .css) files. Examples:",
```
d0sboots, line 145: The short description is also wrong, but it was wrong before too and fixing it while keeping it short is non-trivial, so that can be put off.

src/Terminal/HelpText.ts
```diff
@@ -142,7 +142,7 @@ export const HelpTexts: Record<string, string[]> = {
...
     " ",
```
d0sboots, line 146: I'm pretty sure the grep -O description that currently applies to .txt and .json now also works with .css.

src/Terminal/commands/scp.ts
```diff
@@ -36,7 +36,7 @@ export function scp(args: (string | number | boolean)[], server: BaseServer): vo
     // Error for invalid filetype
     if (!hasScriptExtension(path) && !hasTextExtension(path)) {
       return Terminal.error(
-        `scp failed: ${path} has invalid extension. scp only works for scripts (.js, .jsx, .ts, .tsx), text files (.txt, .json), and literature files (.lit)`,
+        `scp failed: ${path} has invalid extension. scp only works for scripts (.js, .jsx, .ts, .tsx), text files (.txt, .json, .css), and literature files (.lit)`,
```
d0sboots, line 39: The same message should be updated for grep, even though it was technically wrong before (also didn't include json)

PR #2380, opened 2025-11-02
UI: Expose theme as css custom props

shyguy1412:
as discussed on discord and #2378 

This PR exposes the current color theme as css custom properties.

### DIFF
--- a/FixJSDOMEnvironment.ts
+++ b/FixJSDOMEnvironment.ts
@@ -30,7 +30,6 @@ export default class FixJSDOMEnvironment extends JSDOMEnvironment {
     /**
      * https://github.com/jsdom/jsdom/issues/3766
      * https://github.com/jsdom/jsdom/issues/3444
-     *
      */
     this.global.document.adoptedStyleSheets = [];
     this.global.CSSStyleSheet.prototype.replaceSync = () => {};
--- a/FixJSDOMEnvironment.ts
+++ b/FixJSDOMEnvironment.ts
@@ -26,5 +26,13 @@ export default class FixJSDOMEnvironment extends JSDOMEnvironment {
     this.global.TextDecoderStream = TextDecoderStream;
     this.global.URL = URL;
     this.global.Response = Response;
+
+    /**
+    * https://github.com/jsdom/jsdom/issues/3766
+    * https://github.com/jsdom/jsdom/issues/3766
+    * 
+    */
+    this.global.document.adoptedStyleSheets = [];
+    this.global.CSSStyleSheet.prototype.replaceSync = () => {};
   }
 }
--- a/FixJSDOMEnvironment.ts
+++ b/FixJSDOMEnvironment.ts
@@ -29,7 +29,7 @@ export default class FixJSDOMEnvironment extends JSDOMEnvironment {
 
     /**
      * https://github.com/jsdom/jsdom/issues/3766
-     * https://github.com/jsdom/jsdom/issues/3766
+     * https://github.com/jsdom/jsdom/issues/3444
      *
      */
     this.global.document.adoptedStyleSheets = [];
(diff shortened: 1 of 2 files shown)


src/Themes/ui/Theme.tsx
```diff
@@ -419,7 +420,21 @@ export function refreshTheme(): void {
...
   document.body.style.backgroundColor = theme.colors.backgroundprimary?.toString() ?? "black";
+
+  const styleSheet =
+    ":root {" +
+    Object.entries(Settings.theme)
+      .map(([k, v]) => `--bb-theme-${k}: ${v}`)
+      .join(";") +
+    "}";
+
+  //The test suite does not polyfill this property
```
d0sboots: The general way we handle this is by extending the built-in DOM mocks in FixJSDOMEnvironment.ts, instead of changing the production code.

This looks relevant: https://github.com/jsdom/jsdom/issues/3766
https://github.com/jsdom/jsdom/issues/3444

src/Themes/ui/Theme.tsx
```diff
@@ -419,7 +420,21 @@ export function refreshTheme(): void {
...
+  themeStyleSheet.replaceSync?.(styleSheet);
 }
+
+//The test suite does not polyfill this property
+document.adoptedStyleSheets?.push(themeStyleSheet);
```
d0sboots: This should probably go after the refreshTheme, so that the added stylesheet is already in its proper state. Just a minor performance thing, potentially.

src/Themes/ui/Theme.tsx
```diff
@@ -419,7 +420,21 @@ export function refreshTheme(): void {
   });
 
   document.body.style.backgroundColor = theme.colors.backgroundprimary?.toString() ?? "black";
+
+  const styleSheet =
+    ":root {" +
+    Object.entries(Settings.theme)
```
d0sboots, line 426: Is this the part that was giving you type issues? What were you trying to iterate over that *wasn't* `Settings.theme`?

shyguy1412: yes, it was theme.colors initially when I was playing around with different implementations which gave me issues with the `undefined` union. Settings.theme is only a `string` type so it was just nicer to work with. Since previously in that fucnction theme.colors is set to Settings.theme there is not actual difference between them

src/Themes/ui/Theme.tsx
shyguy1412: fair, I was just being incredibly lazy. I'll look into it

src/Themes/ui/Theme.tsx
shyguy1412: ~~nice catch. I dont think it _actually_ matters from my testing, since the function is called every rerender of the main component, but logically it should still go there to make things more robust~~

It really shouldnt matter. The whole thing is synchronous while the application of CSS is async, so either way at the end of these lines the correct CSS is getting applied. I can still change it anyway, either way is the same really

d0sboots: Yes, it shouldn't matter in practice, aside from maybe a minor performance thing and logically being clearer.

src/Themes/ui/Theme.tsx
d0sboots: Damnit lost my comment. Basically, I think using settings is more logically correct anyway.

FixJSDOMEnvironment.ts
```diff
@@ -26,5 +26,13 @@ export default class FixJSDOMEnvironment extends JSDOMEnvironment {
     this.global.TextDecoderStream = TextDecoderStream;
     this.global.URL = URL;
     this.global.Response = Response;
+
+    /**
+     * https://github.com/jsdom/jsdom/issues/3766
+     * https://github.com/jsdom/jsdom/issues/3766
```
d0sboots: You cited the same issue twice

shyguy1412: Oops, that was supposed to be <https://github.com/jsdom/jsdom/issues/3444>. Sry x.x

PR #2381, opened 2025-11-02
UI: Prevent showing redundant error toasts on failed RFA auto connect attempts

shyguy1412:
I got very frustrated with the endless error toasts when the RFA auto connect feature is enabled.

This adds the ability for a toast dispatcher to close the toast itself instead of a duration. This combined with disabling error toasts during automated connection attempts means there is now just one persistent toast until a successful auto connection is made instead of an endless error toast spam

previously:
image

now:
image


d0sboots:
So, first off: There's definitely a real problem here that should be fixed. Toasts should convey useful info, and if they're spammy then they're not doing that.

However, for many of the same reasons toasts aren't well-suited to being persistent. They take up a section of the screen that is pretty important, and there is no way to dismiss them (short of timeout, which persistent ones wouldn't have).

Instead, I think we should take a closer look at what info is actually important to convey to the player. IMO it is the following: 
* RFA is trying to connect (either once or repetitively) 
* RFA failed to connect (if only trying to connect once, or this is the first failure) 
* *Don't* toast on failures that are the same as previous failures (this is uninteresting info), but if the failure type changes that *is* interesting
* Toast on connection made
* Toast on connection broken

Restricted to those, the toasts shouldn't spam. And I think the state should be clear enough that a permanent toast isn't needed? (Surely you might forget that RFA is still trying to autoconnect, but IMO that's mostly unimportant info.)

shyguy1412:
toasts can be always clicked in order to be dismissed

d0sboots:
> toasts can be always clicked in order to be dismissed

OK, I was unaware of that but I still don't think persistent toasts are a good idea (if we need a persistent status it belongs on the overview bar, or elsewhere.)

Edit: An "indicator light" for RFA status on the overview panel could be cool, if we could wedge it in there without taking up extra space. Dark/transparent for off, green for connected, yellow for connecting, red for failed? Or use the theme colors?

shyguy1412:
ig thats fair. I personally liked the persistent toast but I am not overly invested into what the actual method is at the end.

So instead how about this:

Toast is no longer persistent, I revert the changes to the snackbar.

The first connection attempt is always fully logged.
After the first failed connection, the reconnect toast is logged. But no longer persistent.
No further connection errors are logged until a successful connection was made.
The disconnection is logged.
After a successful connection the first failed connection attempt is logged again.

Manual connection attempts get full logging always.

I think messing with the overview requires a lot more care so I would think about it in a 2nd PR at best.

d0sboots:
Yeah that sounds good to me. I suggested logging connection errors if they're different (i.e. "connection refused" is different from "no network"), but if the ws errors don't have that level of info then it is functionally the same as "only log the first error."

And yes, the overview was mostly just spitballing ideas, I also think that would be 2nd PR material.

shyguy1412:
Here is a video of the new behaviour
https://github.com/user-attachments/assets/92896f46-8085-4e2e-b067-fec3d439cd1c

d0sboots:
First off, thanks a lot for the video! That helps a lot with evaluating this.

The first thing that I noticed was that one of the toasts ended up persistent until you clicked it away. Was that common in your testing? I've never had that happen, so I'm concerned that might be a bug that's (somehow) being introduced.

Another thing was that the flow of toasts with auto-reconnect felt a bit off to me. You clicked "connect," then there's a delay, then the first "failed" toast pops up, followed immediately by "Attempting to auto-connect API", followed by another "failed" toast. I would have expected the "attempting to auto-connect" toast immediately, and then only one "failed" toast after. The behavior after disconnect looked right.

src/RemoteFileAPI/Remote.ts
```diff
@@ -24,24 +24,40 @@ export class Remote {
...
       console.error(error);
       showErrorMessage(address, String(error));
       return;
     }
-    this.connection.addEventListener("error", (e: Event) => showErrorMessage(address, JSON.stringify(e)));
+
+    //log connection errors on manual and the first auto connect attempts
+    if (autoConnectAttempt <= 1)
+      this.connection.addEventListener("error", (e: Event) => showErrorMessage(address, JSON.stringify(e)));
+
```
d0sboots:
```suggestion
    this.connection.addEventListener("error", (e: Event) => {
      if (autoConnectAttempt <= 1 || successfullyConnected) {
        showErrorMessage(address, JSON.stringify(e));
      }
    });
```
The flow of setting this event listener in two places is a bit awkward, and more importantly I'm concerned about possible edge-condition bugs due to the different timing and possibility of this.connection referring to a different object inside the callback. Unconditionally setting the listener and then doing the checks inside makes everything easier to validate.

src/RemoteFileAPI/Remote.ts
```diff
@@ -24,24 +24,40 @@ export class Remote {
     this.connection?.close(eventCodeWhenIntentionallyStoppingConnection);
   }
 
-  public startConnection(): void {
+  public startConnection(autoConnectAttempt = 0): void {
```
d0sboots:
```suggestion
  public startConnection(autoConnectAttempt = 1): void {
```
I think defaulting to 1 will fix the "double message on startup" that I talked about.

src/RemoteFileAPI/Remote.ts
```diff
@@ -52,10 +68,21 @@ export class Remote {
       if (event.code === eventCodeWhenIntentionallyStoppingConnection) {
         return;
       }
-      SnackbarEvents.emit(`Remote API connection closed. Code: ${event.code}.`, ToastVariant.WARNING, 2000);
+
+      //Printing a connetion error alongside the connection closed warning is both redunant and confusing
+      if (successfullyConnected)
+        SnackbarEvents.emit(`Remote API connection closed. Code: ${event.code}.`, ToastVariant.WARNING, 2000);
```
d0sboots:
```suggestion
      if (successfullyConnected) {
        SnackbarEvents.emit(`Remote API connection closed. Code: ${event.code}.`, ToastVariant.WARNING, 2000);
      }
```
It's not in our lints, but we use curlies even for one-line if statements due to the well-known potential for bugs they can cause otherwise. The sometimes-exception is for `break` and `return`, but we usually write them all on one line in that case.

shyguy1412:
I noticed that too, I highly doubt any of my changes introduced any of this behavior though. After all, I only changed the conditions for when toasts get dispatched via the event system, this should in no way affect their dismissal and the way the RFA does its connections.

I noticed before that when a lot of toasts get spammed at once some tend to get stuck, this isn't new. I assume there is another bug that needs to be looked at inside the code responsible for dismissing them after the timeout.

~~As for the RFA, right after enabling the setting the auto connections do not start until you press the manual button once. I did so but the timeout took a little longer than usual (they can be a little inconsistent ime) and I got impatient so I just started clicking the manual button multiple times, hence the multiple toasts. I probably should have retaken the video at that point but I didn't think it mattered much since I haven't touched that part of the code.~~

After rewatching the video I realized you probably meant at 0:30 the toasts alternate with error/attempting to connect/error.

This is because the first error is from a manual attempt, which always gets logged. Then the notification gets dispatched that auto-reconnect is now enabled. The first auto-connect attempt happens, which also gets logged, and then the errors are disabled for further attempts to prevent the spam

src/RemoteFileAPI/Remote.ts
```diff
@@ -52,10 +68,21 @@ export class Remote {
...
+
       if (Settings.RemoteFileApiReconnectionDelay > 0) {
         setTimeout(() => {
-          this.startConnection();
+          if (!autoConnectAttempt || successfullyConnected) {
+            SnackbarEvents.emit(`Attempting to auto connect Remote API`, ToastVariant.WARNING, 2000);
+          }
```
d0sboots:
```suggestion
          if (autoConnectAttempt === 1 && Settings.RemoteFileApiReconnectionDelay > 0) {
            SnackbarEvents.emit(`Attempting to auto connect Remote API`, ToastVariant.WARNING, 2000);
          }
```
To do my suggestion of having this pop up immediately when you first enable auto-connection, this would be pulled out to the main part instead of being inside onClose.

d0sboots:
> This is because the first error is from a manual attempt, which always gets logged. Then the notification gets dispatched that auto-reconnect is now enabled. The first auto-connect attempt happens, which also gets logged, and then the errors are disabled for further attempts to prevent the spam

Yup. I understand the technical reasons why it happens, I was saying from a UX flow perspective it felt off to me.

src/RemoteFileAPI/Remote.ts
shyguy1412: I made this a number specifically to get the double startup message. If that is not the wanted behaviour autoConnectAttempt should be a isAutoConnect boolean instead. The only ever check made against it is <= 1 which in your suggestion would make every connection attempt be classified as an auto-connect

EDIT: just saw the other suggestion so keeping it a number would be better after all

d0sboots: Ah, that's a good point and it would be simpler that way.

To me, at least, the double startup error message feels like an odd exception. It feels more natural to only have one error shown between connects, regardless of how you get there.

src/RemoteFileAPI/Remote.ts
shyguy1412: yeah, makes sense. This seems really obvious in hindsight

shyguy1412:
https://github.com/user-attachments/assets/9c75c80e-ba8d-45f5-b5ac-23b72b76cca2

Some toasts got stuck again but besides that it looks pretty good now

d0sboots:
> Some toasts got stuck again but besides that it looks pretty good now

Do you regularly click on toasts? I was looking around at `notistack` (the npm underlying our toasts), and from the issues it looks like it might have some lurking edge cases relating to interactions. This would also explain why I've never seen the issue (and can't repro it).

shyguy1412:
Not usually, no. I havent clicked on any toasts here until I dismissed the ones that got stuck

EDIT: I just realized that only the toasts where I was clicked out of the window while they timed out got stuck. That could be it

src/RemoteFileAPI/Remote.ts
```diff
@@ -24,24 +24,42 @@ export class Remote {
...
       SnackbarEvents.emit(
         `Remote API connection established on ${this.ipaddr}:${this.port}`,
         ToastVariant.SUCCESS,
         2000,
-      ),
-    );
+      );
+
+      //enable error reports after a successfull auto connect
+      if (autoConnectAttempt > 2 && this.connection) {
```
d0sboots: This is still here, which tells me you probably forgot to push something?

shyguy1412: Oh yeah. I only merged your suggestions and totally forgot about this, sry.

src/RemoteFileAPI/Remote.ts
d0sboots: My suggestion here was for pulling it out to the main section instead of it being inside onClose. Watching the video, the timing with it like this also seems fine, but if you want to keep it here the `Settings.RemoteFileApiReconnectionDelay > 0` is redundant and should be removed.

d0sboots:
> Not usually, no. I havent clicked on any toasts here until I dismissed the ones that got stuck
> 
> EDIT: I just realized that only the toasts where I was clicked out of the window while they timed out got stuck. That could be it

Yeah, it's a weird "feature" that they don't tick down when it doesn't have focus. I can repro that part myself, but they *do* dismiss properly when I come back. Nonetheless, I suspect edge-cases here to be the cause of the bug.

shyguy1412:
I cant reliably reproduce this but I'll keep an eye out. If I notice it again I'll do a deeper dive into the cause

src/RemoteFileAPI/Remote.ts
d0sboots: I applied this myself so we can get this submitted.

catloversg:
It looks good to me. There is a bug that creates a list of pending connection attempts, but that bug is not caused by the changes here, so I'll fix it later in another PR.

PR #2382, opened 2025-11-02
SCRIPT: support for import attributes

shyguy1412:
With import attributes being a baseline feature now I think it would be neat if we added support for that.

This is some basic proof of concept. There are probably some nuances I am not aware that still need to be ironed out

### DIFF
--- a/markdown/bitburner.ns.dynamicimport.md
+++ b/markdown/bitburner.ns.dynamicimport.md
@@ -9,7 +9,7 @@ Dynamically import a script. Only scripts located on the same server can be impo
 **Signature:**
 
 ```typescript
-dynamicImport(path: string): Promise<any>;
+dynamicImport(path: string, opts?: ImportCallOptions): Promise<any>;
 ```
 
 ## Parameters
@@ -41,6 +41,22 @@ string
 
 
 </td><td>
+
+
+</td></tr>
+<tr><td>
+
+opts
+
+
+</td><td>
+
+ImportCallOptions
+
+
+</td><td>
+
+_(Optional)_
 
 
 </td></tr>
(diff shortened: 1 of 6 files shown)


shyguy1412:
misclicked x.x

d0sboots:
One thing I'll point out is that by general browser impl, `type: json` is not actually needed to parse things as json blobs. It is only needed to *ensure* that bad mime detection doesn't cause things that ought to be parsed as json, to be parsed as js instead. In our case (for local resources), we're doing it by filename so this would not be an issue, so the import attributes aren't technically needed. This extends to css as well (although browsers won't import CSS without the annotation, they're also not required *not* to, so we could do it.)

I'm not suggesting any particular course of action per-se, just noting things.

shyguy1412:
Oh, cause the blobs already got the mime type ensured? Thats neat, didnt know that.

That being said, I dont think its a good to allow non-script imports without import attributes. It may work in bitburner but it wouldnt in really any other context.

catloversg:
I'm not particularly against this idea, but I think we should only do that when we (or the players) *really* need it. Functions in `NetscriptJSEvaluator.ts` and their callers are a land of mines. If we touch them, we need a very good reason to do that.

I think it's okay to leave this PR here as a suggestion of a pending change. We can come back and finish it later when we need it.

shyguy1412:
The main reasons I want this is that with the merging of #2380 being able to import .css files as modules would finalize everything I'd want from the game as styling support. Being able to import json modules is also a neat extra.

I agree that touching the NescriptEvaluator is a really big deal and I can work around the missing static imports with dynamic imports so I am ok with leaving this for now.  I do think that we should eventually add this tho

d0sboots:
One thing I'll say is that if/when we do this, we should extend dynamicImport as well (which should be mostly trivial)

shyguy1412:
Oh yes, totally. I kinda forgot that function even existed x.x

shyguy1412:
havent really done any testing yet but thats my first idea on how to approach dynamic imports

PR #2391, opened 2025-11-21
BUGFIX: Company's job list shows wrong starting jobs

### LINKED ISSUE #2386: Helios Labs duplicated job apply buttons.
Image

catloversg:
How to reproduce: Go to Volhaven and check Helios Labs:

helios-labs

The duplicate entries are the senior jobs.

This list is calculated like this:
https://github.com/bitburner-official/bitburner-src/blob/bb50272e5449730f8a845b28da932349240b9644/src/Company/ui/JobListings.tsx#L19-L30

We assume that a starting job is a job that does not require reputation (`job.requiredReputation === 0`), but this is wrong.
- `Senior Software Consultant`: `requiredReputation = 0`, but it's a senior job promoted from `Software Consultant`.
- `Senior Business Consultant`: `requiredReputation = 0`, but it's a senior job promoted from `Business Consultant`.
- `Security Engineer`: `requiredReputation = 35000`. This is the only job in the `Security Engineer` field, so it's a starting job.
- `Network Engineer`: `requiredReputation = 35000`. This is the starting job in the `Network Engineer` field. This position will be promoted to `Network Administrator`.
- `Field Agent`: `requiredReputation = 8000`. This is the starting job in the `Agent` field. This position will be promoted to `Secret Agent`.

This PR fixes this bug by explicitly specifying which jobs are starting jobs.

Test code for getting debugging data:
```js
/** @param {NS} ns */
  console.clear();
  const showWrongJobs = false;
  const map = new Map();
  for (const [jobName, job] of Object.entries(getCompanyPositionMetadata())) {
    let l = map.get(job.field);
    if (l == null) {
      l = [];
      map.set(job.field, l);
    }
    l.push(Object.assign({ name: jobName }, job));
    if (showWrongJobs) {
      if (job.reqdReputation == null || job.reqdReputation === 0) {
        console.log(jobName, job);
      }
    } else if (job.isStartingJob) {
      console.log(jobName, job);
    }
  }
  console.log(map);
}
```

Before:

before-1

before-2

After:

after-1

after-2

Fixes #2386.

### DIFF
--- a/src/Company/CompanyPosition.ts
+++ b/src/Company/CompanyPosition.ts
@@ -11,6 +11,7 @@ export interface CompanyPositionCtorParams {
   applyText?: string;
   hiredText?: string;
   isPartTime?: boolean;
+  isStartingJob?: boolean;
 
   reqdHacking?: number;
   reqdStrength?: number;
@@ -41,6 +42,9 @@ export class CompanyPosition {
 
   /** Field type of the position (software, it, business, etc) */
   field: JobField;
+
+  /** Whether this position is shown in the job list even when the player does not satisfy its requirements */
+  isStartingJob: boolean;
 
   /** Title of next position to be promoted to */
   nextPosition: JobName | null;
@@ -93,6 +97,7 @@ export class CompanyPosition {
   constructor(name: JobName, p: CompanyPositionCtorParams) {
     this.name = name;
     this.field = p.field;
+    this.isStartingJob = p.isStartingJob ?? false;
     this.nextPosition = p.nextPosition;
     this.baseSalary = p.baseSalary;
     this.repMultiplier = p.repMultiplier;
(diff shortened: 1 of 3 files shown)


d0sboots:
Maybe I'm too tired because I just woke up, but I still don't understand this bug. The loop is over all jobs, so even if the filtering is wrong, how does this result in duplicates?

catloversg:
Let's use the `Software Consultant` field as an example. That field has 2 jobs: `Software Consultant` (the starting job) and `Senior Software Consultant` (the promoted job). Both jobs have the same description `Pitch a Software Consulting contract` and do *not* require reputation.

In the loop, when `jobName` is `Senior Software Consultant` (the promoted job), it should be filtered out by the second `if`, but that `if` uses a wrong condition (`job.requiredReputation > 0`). Therefore, the final job list contains 2 jobs instead of 1 job. The UI is also confusing because both jobs have the same description. It's not exactly a "duplicate" by definition (they are different jobs, but have the same description on UI).

Note that the wrong condition also affects other jobs/fields.

d0sboots:
> Both jobs have the same description `Pitch a Software Consulting contract`

Ah, this is where I got hung up. I thought the different jobs had different names. XD

PR #2393, opened 2025-11-22
CODEBASE: Follow-up of #2316

catloversg:
These were my questions for #2316, but I was too busy and could not post them in time.

@d0sboots I prefix my questions with `WIP`. After we discuss and agree on the outcome, I'll remove these comments.

### DIFF
--- a/src/Infiltration/model/WireCuttingModel.ts
+++ b/src/Infiltration/model/WireCuttingModel.ts
@@ -41,8 +41,7 @@ function randomPositionQuestion(wires: Wire[]): Question {
   const index = Math.floor(Math.random() * wires.length);
   return {
     toString: (): string => {
-      // WIP: I reverted the string here to the previous code. It was "wires", not "wire". Is that okay?
-      return `Cut wires number ${index + 1}.`;
+      return `Cut wire number ${index + 1}.`;
     },
     shouldCut: (_wire: Wire, i: number): boolean => {
       return index === i;
--- a/src/Infiltration/model/WireCuttingModel.ts
+++ b/src/Infiltration/model/WireCuttingModel.ts
@@ -41,7 +41,8 @@ function randomPositionQuestion(wires: Wire[]): Question {
   const index = Math.floor(Math.random() * wires.length);
   return {
     toString: (): string => {
-      return `Cut wire number ${index + 1}.`;
+      // WIP: I reverted the string here to the previous code. It was "wires", not "wire". Is that okay?
+      return `Cut wires number ${index + 1}.`;
     },
     shouldCut: (_wire: Wire, i: number): boolean => {
       return index === i;
--- a/src/Infiltration/ui/GameTimer.tsx
+++ b/src/Infiltration/ui/GameTimer.tsx
@@ -26,10 +26,8 @@ export function GameTimer({ endTimestamp }: GameTimerProps): React.ReactElement 
     const startTimestamp = changeRef.current.startTimestamp;
     if (!ele) return;
     ele.animate([{ transform: "translateX(0%)" }, { transform: "translateX(-100%)" }], {
-      /**
-       * If duration is negative, ele.animate will throw "TypeError: duration must be non-negative or auto". It may
-       * happen when the player uses the debugger.
-       */
+      // If duration is negative, ele.animate will throw "TypeError: duration must be non-negative or auto". It may
+      // happen when the player uses the debugger.
       duration: clampNumber(endTimestamp - startTimestamp, 0),
       // The delay will be negative. This is because the animation starts
       // partway completed, due to the time taken to invoke the effect.
--- a/src/Infiltration/ui/GameTimer.tsx
+++ b/src/Infiltration/ui/GameTimer.tsx
@@ -1,5 +1,6 @@
 import React, { useEffect, useRef, useMemo } from "react";
 import { ProgressBar } from "../../ui/React/Progress";
+import { clampNumber } from "../../utils/helpers/clampNumber";
 
 type GameTimerProps = {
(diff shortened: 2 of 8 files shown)


src/Infiltration/model/WireCuttingModel.ts
```diff
@@ -41,7 +41,8 @@ function randomPositionQuestion(wires: Wire[]): Question {
   const index = Math.floor(Math.random() * wires.length);
   return {
     toString: (): string => {
-      return `Cut wire number ${index + 1}.`;
+      // WIP: I reverted the string here to the previous code. It was "wires", not "wire". Is that okay?
```
d0sboots: The minor change here was intentional; "wires" is ungrammatical.

src/Infiltration/ui/GameTimer.tsx
```diff
@@ -14,7 +15,7 @@ export function GameTimer({ endTimestamp }: GameTimerProps): React.ReactElement
   // actually begin it until later. We're using a ref to keep this updated in
   // a very low-overhead way.
   const state = changeRef.current;
-  if (state?.endTimestamp !== endTimestamp) {
+  if (state.endTimestamp !== endTimestamp) {
```
d0sboots, line 18: Weird, I know TS was complaining about null at some point. (But it shouldn't be.)

src/Infiltration/ui/GameTimer.tsx
```diff
@@ -24,10 +25,14 @@ export function GameTimer({ endTimestamp }: GameTimerProps): React.ReactElement
     const ele = ref.current?.firstElementChild;
     const startTimestamp = changeRef.current.startTimestamp;
     if (!ele) return;
-    // The delay will be negative. This is because the animation starts
-    // partway completed, due to the time taken to invoke the effect.
     ele.animate([{ transform: "translateX(0%)" }, { transform: "translateX(-100%)" }], {
-      duration: endTimestamp - startTimestamp,
+      /**
+       * If duration is negative, ele.animate will throw "TypeError: duration must be non-negative or auto". It may
```
d0sboots: Minor nit: Let's keep the same comment style here. `/**` is associated with public api documentation by me at least, and I use `//` for implementation comments. But it doesn't really matter, as long as there's some consistency.

src/Infiltration/ui/InfiltrationRoot.tsx
```diff
@@ -51,6 +51,7 @@ const stages = new Map([
 function Progress({ results }: { results: string }): React.ReactElement {
   return (
     <Typography variant="h4">
+      {/* WIP: It seems that you want to display only 14 last results instead of the full history. Is that right? */}
```
d0sboots: Correct, this mirrors the previous behavior. The semantics of this slice are subtle, so it could probably use a comment.

src/Infiltration/ui/InfiltrationRoot.tsx
```diff
@@ -59,6 +60,7 @@ function Progress({ results }: { results: string }): React.ReactElement {
 
 export function InfiltrationRoot(): React.ReactElement {
   const state = Player.infiltration;
+  // WIP: Is there a reason we should not use the useRerender hook?
```
d0sboots: No, I don't think it makes a difference.

src/Infiltration/ui/InfiltrationRoot.tsx
```diff
@@ -89,6 +91,10 @@ export function InfiltrationRoot(): React.ReactElement {
   }, [state]);
 
   if (!state) {
+    /**
+     * WIP: Should we auto-redirect to another tab and/or use exceptionAlert? If this happens, the player will only see
```
d0sboots: Yes, I thought about those possibilities as well. But since what I'm concerned about is this being a *transient* state, I don't think that's a good idea. 

Basically, my experience with trying to align React timing with everything else (the whole reason this refactor happened) left me convinced that it's almost impossible to *guarantee* you'll never run into a state like this. However, it shouldn't last long enough to be perceptible by the user - if they end up *stuck* in this state, that's a bug.

src/Infiltration/ui/InfiltrationRoot.tsx
```diff
@@ -106,6 +112,7 @@ export function InfiltrationRoot(): React.ReactElement {
   return (
     <div style={{ display: "flex", alignItems: "center", height: "calc(100vh - 16px)" }}>
       {state.stage instanceof IntroModel ? (
+        // WIP: stage is unused. Can I remove it?
```
d0sboots: I don't think that will work, because of the typing with StageProps.

src/PersonObjects/Player/PlayerObjectGeneralMethods.ts
```diff
@@ -603,6 +603,11 @@ export function focusPenalty(this: PlayerObject): number {
 }
 
 /** This doesn't change the current page; that is up to the caller. */
+/**
+ * WIP: What do you think about renaming this function to setInfiltration? "setInfiltration" is exactly what this
```
d0sboots: In my follow-up I am also setting the player's location, so I don't think setInfiltration is the right name. 

You are right that "startInfiltration" is overloaded, though. There is this broader type of start, where we bring up the UI and assign the class, and then the more specific type of the minigames starting. 

Maybe `initInfiltration` is the right name?

src/Infiltration/ui/GameTimer.tsx
catloversg: TS did not complain on my machine, so I kept this change in the last commit.

ficocelliguy (approved):

PR #2394, opened 2025-11-22
UI: Improve instructions in recovery mode

catloversg:
Recently we had a player reporting an issue, and they were constantly posting a screenshot instead of the bug report text, even after I told them not to do that. It seems that they did not know the bug report text area is a text box containing the bug report in text form.

This PR adds an instruction that explicitly tells the player what they are supposed to do and a scrollbar to the text box to make it a bit clearer that the text box contains more things than what is shown on the screen.

Before:

before

After:

after


d0sboots:
But then you went and took a screenshot of it anyway XD

PR #2395, opened 2025-11-22
PIPE: Add pipe support for passing data into and out of terminal commands

ficocelliguy:
A number of players have asked for support for pipes in the terminal, or have been confused as to its lack of support:
https://discord.com/channels/415207508303544321/415207839246581781/1411811521935315016
https://discord.com/channels/415207508303544321/415207839246581781/1275557408726323260
https://discord.com/channels/415207508303544321/415207508303544323/932487013922390016
https://discord.com/channels/415207508303544321/415207508303544323/933463698431967283
https://discord.com/channels/415207508303544321/415207508303544323/573291980801703946

So, I've added support for them!

image

Deployed and testable at:
https://ficocelliguy.github.io/bitburner-src-2/

Features (and test cases):

- `echo` command added
- piping to text file support added - will overwrite if `>` is used, otherwise will append ( `>>` )
```
echo test;
echo test data > newFile.txt;
echo overwrite that data > newFile.txt;
```

- support for grepping string literals added
- Support for passing pipe output as additional args for any terminal command added
- support for multiple pipes in a chain added
```
echo Hello World > newFile.txt;
ls | grep .txt;
ls | grep .txt > fileList.txt;
```

- new  ns property `stdin` added: it is a port handle that receives any data passed to the script via pipes
- Support for piping to write the contents of script JS files added
- Support for piping script terminal output to another output location added
```
echo "export function main(ns) { ns.tprint('Input received: ', ns.stdin.read() ); }" > myScript.js;
echo arg1 arg2 | myScript.js;
echo arg3 arg4 | myScript.js > newLogFile.txt;
```

- piping from tail output into additional sources added
```
run myScript arg1; tail myScript arg1 >> tailLogFile.txt;
```


d0sboots:
Just like `>` connects stdout to a file, `<` connects stdin to a file. That is why `<<` doesn't make any sense - the file is being read from, so append or not is irrelevant.

ficocelliguy:
@d0sboots it's apparently called "here document"
https://tldp.org/LDP/abs/html/here-docs.html

I thought it could also pull file contents as an input redirect, but I'm now less confident in that reading those docs, though?

I had added it as an alias for < but it's really not, the more I look at this.

d0sboots:
Note that that page is the "advanced bash scripting guide."

Here-doc syntax is *technically* usable on the command-line, but in practice no one ever would because you can just type stuff instead. It is a scripting feature. Amusingly, it *would* be slightly more pointful for us because all our jobs are launched "in the background" and so you can't generate stdin just by typing. 

However, supporting here-docs would *mandate* support for multi-line commands, which is a big complication for both code and players, so I don't think we want to go there.

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -6483,6 +6483,21 @@ export interface NS {
...
+  /**
+   * A NetscriptPort handle used to get input piped to the script.
+   * Examples:
+   *
+   * If a script was run with data piped into it via the terminal:
+   * `echo input1 | run myScript.js`
+   *
+   * then `ns.stdin.read()` inside `myScript.js` would return `"input1"`.
+   *
+   * If more data is added later (for example, if one script's terminal output is piped to another script),
```
```suggestion
   * If more data is added later (for example, if one script's output is piped to another script),
```

src/NetscriptFunctions.ts
```diff
@@ -1564,7 +1565,12 @@ setRemovedFunctions(ns, {
 });
 
 export function NetscriptFunctions(ws: WorkerScript): NSFull {
-  return NSProxy(ws, ns, [], { args: ws.args.slice(), pid: ws.pid, enums });
+  return NSProxy(ws, ns, [], {
+    args: ws.args.slice(),
+    pid: ws.pid,
+    stdin: new PortHandle((ws.pid * -1) as PortNumber),
```
d0sboots: Adding things as properties of NS (as opposed to functions) means they have to be created for every script, slowing down script startup for everyone (regardless of if they are used). Also, they're harder to find in the docs. So, without a pressing reason I think this ought to just be a function.

src/Script/RunningScript.ts
```diff
@@ -70,9 +73,17 @@ export class RunningScript {
   // Cached key for ByArgs lookups. Will be overwritten by a correct ScriptKey in fromJSON or constructor
   scriptKey = "" as ScriptKey;
 
+  stdin = null as NetscriptPort | null;
```
d0sboots: I like the other style you added better. Ideally we'd have a consistent style in this file, but I'll settle for a consistent style of new additions. :)

src/Script/ScriptHelpers.ts
```diff
@@ -115,3 +115,18 @@ export function findRunningScriptByPid(pid: number): RunningScript | null {
...
   return ws.scriptRef;
 }
+
+export function findRunningScriptsByFilename(
+  path: ScriptFilePath,
+  server: BaseServer,
+): Map<number, RunningScript> | null {
+  const result = new Map<number, RunningScript>();
+  for (const [key, runningScriptMap] of server.runningScriptMap.entries()) {
+    if (key.startsWith(path + "*")) {
```
d0sboots: This should use matchScriptPathExact.

Seeing yet another findScripts method furthers my unease about adding this. I think I'd be happier if this lived specifically in the tail impl and not in a shared place.

src/Terminal/HelpText.ts
```diff
@@ -14,6 +14,7 @@ export const TerminalHelpText: string[] = [
   "    connect [hostname]               Connects to a remote server",
   "    cp [src] [dest]                  Copy a file",
   "    download [script/text file]      Downloads scripts or text files to your computer",
+  "    echo [string]                    Print the specified string to the terminal. Typically used for piping.",
```
d0sboots:
```suggestion
  "    echo [string]                    Print the specified string to the terminal.",
```
The second sentence is unneeded, especially in a short description.

src/Terminal/HelpText.ts
```diff
@@ -215,6 +216,15 @@ export const HelpTexts: Record<string, string[]> = {
...
     " ",
   ],
+  echo: [
+    "Usage: echo [string]",
+    " ",
+    "Print the specified string to the terminal. This command is mostly useful for piping",
+    " ",
+    "Example: echo 'Text To Store In File' > newFile.txt",
+    " ",
+    "Example: echo 'Text To Search In' | grep 'To'",
```
d0sboots:
```suggestion
    "Example: echo 'Text To Search In' | grep To",
```
Quoting is unnatural for a single word.

src/Terminal/Parser.ts
```diff
@@ -1,17 +1,22 @@
...
 import { substituteAliases } from "../Alias";
+import { PipeState } from "./PipeState";
 // Helper function to parse individual arguments into number/boolean/string as appropriate
 function parseArg(arg: string): string | number | boolean {
   if (arg === "true") return true;
   if (arg === "false") return false;
   const argAsNumber = Number(arg);
   if (!isNaN(argAsNumber)) return argAsNumber;
+  if (arg === "$!") {
+    return PipeState.pidOfLastScriptRun ?? -1;
```
d0sboots: Why is this on PipeState? That doesn't seem like where it belongs; I would expect Terminal.

src/Terminal/PipeState.ts
```diff
@@ -0,0 +1,96 @@
...
+import { Terminal } from "../Terminal";
+import { TerminalEvents } from "./TerminalEvents";
+
+export const PipeState = {
+  outputToBeProcessed: [] as PipedOutput[],
+  currentTerminalPipe: null as PipedCommand | null,
+  pidOfLastScriptRun: null as number | null,
+};
+
+export const PipeSymbols = {
```
d0sboots: This and the next one should be `as const`

src/Terminal/PipeState.ts
```diff
@@ -0,0 +1,96 @@
...
+    pipeDestinationIsIdentical(a.nextPipe, b.nextPipe)
+  );
+}
+
+export type PipedOutput = {
+  output: (Output | Link | RawOutput)[];
+  pipeDestination: PipedCommand | null;
+};
+
+export type PipedCommand = {
```
d0sboots: Naming: I think many of the things that are currently called "PipedX" should be named something based on "redirection" or "redirect". This is because they are not specific to pipes, but rather apply to general io redirection.

src/Terminal/PipeState.ts
```diff
@@ -0,0 +1,96 @@
+import { Link, Output, RawOutput } from "./OutputTypes";
+import { Settings } from "../Settings/Settings";
+import { Terminal } from "../Terminal";
+import { TerminalEvents } from "./TerminalEvents";
+
+export const PipeState = {
```
d0sboots: I really dislike this blob of global state, and I also dislike that it may be necessary. 

What I dislike about it so much is that it exists to make a shim layer on top of the awkwardness that has always existed between the terminal commands and the scripts. This layer will further entrench that awkwardness, but maybe the issue is unsolvable in a different way by this point.

src/Terminal/Terminal.ts
```diff
@@ -157,24 +162,51 @@ export class Terminal {
...
+      pushPipedOutput(item, PipeState.currentTerminalPipe);
+    } else {
+      // Otherwise, simply push the output to stdout
+      this.terminalOutput(item);
+    }
+
+    queueTerminalEvent();
+  }
+
+  logCommand(command: string): void {
```
d0sboots: This seems redundant with printAndBypassPipes

src/Terminal/PipeState.ts
```diff
@@ -0,0 +1,96 @@
...
+
+export function clearPipe() {
+  PipeState.outputToBeProcessed.shift();
+  PipeState.currentTerminalPipe = null;
+}
+
+export function queueTerminalEvent() {
+  setTimeout(() => {
+    TerminalEvents.emit();
+  }, 10);
```
d0sboots:
```suggestion
  }, 0);
```

src/Terminal/PipeState.ts
```diff
@@ -0,0 +1,96 @@
...
+  }
+  clearPipe();
+}
+
+export function clearPipe() {
+  PipeState.outputToBeProcessed.shift();
+  PipeState.currentTerminalPipe = null;
+}
+
+export function queueTerminalEvent() {
```
d0sboots: This is a hack. It needs a big comment explaining why it's being used, and it shouldn't be used ubiquitously but only where absolutely needed.

d0sboots (commented):
I crashed out after hitting the big design issue with PipeState.

src/NetscriptFunctions.ts
ficocelliguy: I have it as ns.getStdin() now.

src/Terminal/PipeState.ts
ficocelliguy: This was part of some testing I was doing that I forgot to remove. It's reverted now.

src/Terminal/commands/cat.ts
```diff
@@ -1,33 +1,71 @@
...
+  for (const arg of args) {
+    if (arg === "-") {
+      initialStdOut += initialStdIn.join("\n") + "\n";
+    } else {
+      const content = getFileContents(String(arg), server, stdIO, isBasicFileCat);
+      if (typeof content !== "string") {
+        stdIO.close();
+        return;
+      }
+      initialStdOut += content + "\n";
```
d0sboots: Cat doesn't add newlines to files, those are expected to be part of the file of you want them.

src/Terminal/commands/cat.ts
```diff
@@ -1,33 +1,71 @@
...
+    initialStdOut += initialStdIn.join("\n") + "\n";
+  }
+
+  stdIO.write(initialStdOut.trim());
 
-  const relative_filename = args[0] + "";
-  const path = Terminal.getFilepath(relative_filename);
-  if (!path) return Terminal.error(`Invalid filename: ${relative_filename}`, stdIO);
+  void callOnRead(stdIO, (data: unknown, stdInOut) => {
+    stdInOut.write(stringify(data) + "\n");
```
d0sboots: Same

src/Terminal/commands/cat.ts
```diff
@@ -1,33 +1,71 @@
...
+import { callOnRead } from "../StdIO/RedirectIO";
+import { stringify } from "../StdIO/utils";
+import { showLiterature } from "../../Literature/LiteratureHelpers";
+import { dialogBoxCreate } from "../../ui/React/DialogBox";
 
-// TODO-Fico: support stdin, support "-" placeholder for stdin
+// TODO-Fico: update cat help docs
 export function cat(args: (string | number | boolean)[], server: BaseServer, stdIO: StdIO): void {
-  if (args.length !== 1) return Terminal.error("Incorrect usage of cat command. Usage: cat [file]", stdIO);
+  const initialStdIn = stdIO.getAllCurrentStdin();
```
d0sboots: It feels like you want to just turn this into a string here. (Why isn't it already a string?)

src/Terminal/commands/cat.ts
ficocelliguy: I've removed them to be consistent. (Note that you can't add newlines to .lit files etc, but I guess you could make a newline.txt that just has your carriage return and weave that in?)

src/Terminal/commands/cat.ts
ficocelliguy: I changed getAllCurrentStdin to retrieve the current input as a single string.

src/Terminal/StdIO/StdIO.ts
```diff
@@ -60,7 +60,7 @@ export class StdIO {
       }
       inputs.push(stringify(input));
     }
-    return inputs;
+    return inputs.join("\n");
```
d0sboots: I would use a map here instead. There's a subtlety in typical Unix file handling involving newline-at-EOF vs no newline-at-EOF. The typical Unix convention is that you *want* that newline at end of file, because tools like cat and grep treat files literally as bytes, so the last line will look funny without it. Note that when I say "want" I mean as a user, not a tool author.

However, all of that is for *files*. For our stdin convention, the newlines are implicit because we are representing them as an array. So, I would argue that the newline belongs at the end of every line here, not just between lines.

src/Terminal/commands/cat.ts
d0sboots: That's a good point about the .lit files. We probably do want to update them to have a talking newline if we are being more unixy. (I'm unsure if doing that naively would change their display; Ideally it would not. It is unfortunate that web editors tend to represent newline at eof as a blank line.)

src/Terminal/StdIO/StdIO.ts
ficocelliguy: that makes sense - each of these was logged distinctly and ends its line in the terminal normally, including the most recent one. I've updated this accordingly

src/Terminal/PipeState.ts
ficocelliguy: You'll be happy to know this whole global state thing was removed and not needed with the new stdio model of chained redirects we discussed.

src/Terminal/commands/cat.ts
```diff
@@ -14,11 +14,13 @@ import { dialogBoxCreate } from "../../ui/React/DialogBox";
...
   const initialStdIn = stdIO.getAllCurrentStdin();
-  const stdinIsClosed = stdIO.stdin?.deref()?.isClosed ?? true;
+  const stdin = stdIO.stdin?.deref();
+  const stdinIsClosed = !stdin || (stdin.isClosed && stdin.empty());
+  const hasStdOut = !!stdIO.stdout;
   let initialStdOut = "";
 
-  // If only a single file is being catted, and no stdin is being used, show the file dialog that cat has always used
-  const isBasicFileCat = args.length === 1 && args[0] !== "-" && !initialStdIn.length && stdinIsClosed;
+  // If only a single file is being catted, and no stdin/stdout redirects are being used, show the file dialog that cat has always used
```
d0sboots: It's been a while, but I thought the key point was no stdout redirects? If you disallow stdin redirects, then what do you do with `echo foo | cat`? I had thought the plan was to show that in the dialog box, but maybe I'm misremembering.

src/Terminal/StdIO/utils.tsx
```diff
@@ -19,21 +19,33 @@ export function isPipeSymbol(symbol: string | number | boolean): boolean {
 }
 
 export function stringify(s: unknown, stripAnsiEscape = false): string {
-  if (!s) {
+  if (s === undefined || s === null) {
```
d0sboots: s == null

src/Terminal/commands/cat.ts
d0sboots: *unresolve comment*

src/Terminal/commands/cat.ts
ficocelliguy: I had originally thought about some sort of dynamic output dialog or pseudo-tail window, but I'm not entirely sure when it would be actually used. (also, it turns out it's a larger lift than I first thought - the current tail implementation is tightly coupled with scripts and their logs.)

For now, I'm making cat behave like unix cat (concatenate its inputs and send them to stdout) if you use any stdin content or have its stdout piped into another command. I'll get some user feedback from there - if the dynamic dialog thing is actually useful and desired I'll look more into it.

d0sboots: I think the output dialog is a more suitable place for just reading text, and we don't have utilities like less/more for dealing with lots of output. But I'm ok with this for now.

src/NetscriptFunctions.ts
```diff
@@ -128,6 +130,7 @@ export const enums: NSEnums = {
   BladeburnerActionType,
   SpecialBladeburnerActionTypeForSleeve,
   FragmentType: FragmentTypeEnum,
+  Redirect: Redirect as unknown as RedirectEnumType,
```
ficocelliguy: I can't figure out why typescript blows up if I don't mark this explicitly here. This "works" in that the enum is available in-game and it generates the md docs, but the mouseover and tab-completion don't work for the enum's properties. 

What are all the other enums doing to avoid this issue?

src/NetscriptPort.ts
```diff
@@ -60,15 +60,15 @@ export class PortHandle implements NetscriptPort {
 
   read(): unknown {
     const port = NetscriptPorts.get(this.n);
-    if (!port || !port.data.length) return emptyPortData;
+    if (!port || !port.data.length) return Redirect.EmptyPortData;
```
d0sboots: I feel like it would be better to return null in this case. This is both more similar to Unix, and also a better API in general (null is falsey etc.) Although NetscripPort can transfer arbitrary data, including null, that is not the case for PortHandle.

Even if you don't agree for the internal implementation, I definitely think that's how the external API should be. And then the enum doesn't need to exist.

ficocelliguy: This code is for regular ports, it's not specific to pipes/redirects. This also doesn't change the behavior that we have in stable where you get "NULL PORT DATA" if you read from an empty port. I just wanted to expose the magic string "NULL PORT DATA" to players for convenience. (This might be nice regardless of how stdio is done)

I assume you intend this feedback for ns.getStdin().read() instead? As in, if the stdin is closed and empty, reading from it should return null instead of "IO_STREAM_CLOSED" ?

d0sboots: Yeah, that's what I meant. I guess I commented on the wrong line.

src/Terminal/StdIO/utils.tsx
```diff
@@ -2,7 +2,10 @@ import React, { isValidElement } from "react";
 import { renderToStaticMarkup } from "react-dom/server";
 import { Link, Output, RawOutput } from "../OutputTypes";
 import { ANSI_ESCAPE } from "../../ui/React/ANSIITypography";
+import { PortHandle, PortNumber } from "../../NetscriptPort";
+import { parseCommand } from "../Parser";
 
+export const IO_STREAM_CLOSED = null;
```
d0sboots: Optional: I think this constant confuses more than it helps.

Not only is `IO_STREAM_CLOSED` longer than `null`, but when I see it used in the source it makes me think that this is something with a unique value, since that is how constants like this are normally used. (A special string, or a unique Symbol.) Just using `null` directly would make things more clear.
Also, one of the primary reasons for using a constant like this is to avoid typos (if you have a string, there's the possibility you change the string value at different callsites), but that's impossible with a built-in token like `null`.

src/Terminal/StdIO/utils.tsx
```diff
@@ -44,6 +47,21 @@ export function stringifyReactElement(element: React.ReactNode): string {
...
+}
+
 function clean(str: string, stripAnsiEscape: boolean) {
   return stripAnsiEscape ? str.replaceAll(ANSI_ESCAPE, "") : str;
 }
+
+let nextStdinPort = 1e7;
+export function getNextStdinHandle(): PortHandle {
+  return new PortHandle((++nextStdinPort * -1) as PortNumber);
+}
```
d0sboots: Missing newline at EOF

src/Terminal/StdIO/utils.tsx
```diff
@@ -44,6 +47,21 @@ export function stringifyReactElement(element: React.ReactNode): string {
...
+  return parsedCommands.slice(lastPipeIndex + 1).join(" ");
+}
+
 function clean(str: string, stripAnsiEscape: boolean) {
   return stripAnsiEscape ? str.replaceAll(ANSI_ESCAPE, "") : str;
 }
+
+let nextStdinPort = 1e7;
+export function getNextStdinHandle(): PortHandle {
+  return new PortHandle((++nextStdinPort * -1) as PortNumber);
```
d0sboots: This seems both awkward and wrong. Preincrement means "next" isn't exactly true, since you'll actually be returning the port *after* this one. And the negation makes it harder to follow what the actual value is.

Why not just `return new PortHandle(nextStdinPort-- as PortNumber);` and store the actual value?

src/Terminal/StdIO/utils.tsx
ficocelliguy: I've removed this to just use `null` directly. Note that this is player-facing, as a closed StdIO retrieved via a script  `ns.getStdin().read()` will return null

src/Terminal/commands/cat.ts
ficocelliguy: I've made it so extracting the text content of lit and msg via cat has an ending newline, but not other content (which the player can control)

src/Terminal/StdIO/RedirectIO.ts
```diff
@@ -25,6 +26,9 @@ export function parseRedirectedCommands(commandString: string) {
     const stdIO = stdIOChain[i];
     handleCommand(stdIO, commandSet);
   }
+
+  // Allow the IO chain to pass data through its async iterators
+  await sleep(50);
```
d0sboots: This is a huge hack. I don't know what problem it's trying to solve, but sleeping is never the answer to async issues.

src/Terminal/StdIO/RedirectIO.ts
```diff
@@ -81,6 +85,24 @@ export function findCommandsSplitByRedirects(commands: Args[]) {
...
+        getTerminalStdIO(),
+        `Error in pipe command: Invalid pipe command. Only the first command in a pipe chain can have input redirection '<'.`,
+      );
+      return [];
+    }
+  }
+
+  // If the second command starts with an input redirection, convert it to a simple pipe.
+  if (result[1]?.[0] === PipeSymbols.InputRedirection) {
+    const inputRedirectCommand = result.splice(1, 1)[0];
```
d0sboots, line 101: It doesn't look like there's any error checking for multiple args passed to an input redirect?

src/Terminal/ui/TerminalInput.tsx
```diff
@@ -233,7 +233,7 @@ export function TerminalInput(): React.ReactElement {
       const command = searchResults.length ? searchResults[searchResultsIndex] : value;
       Terminal.printAndBypassPipes(`[${Player.getCurrentServer().hostname} /${Terminal.cwd()}]> ${command}`);
       if (command) {
-        Terminal.executeCommands(command);
+        await Terminal.executeCommands(command);
```
d0sboots: This also seems like part of the hack.

src/Documentation/doc/en/programming/terminal_pipes_and_redirects.md
```diff
@@ -0,0 +1,72 @@
...
+The output of commands and scripts, that normally would be logged to the terminal, can instead be redirected and sent to another location.
+
+For example, `echo` logs whatever input it is given.
+
+```
+[home /]> echo test123
+test123
+```
+
+However, its output can instead be appended to a file using the output redirect `>>` :
```
d0sboots:
```suggestion
However, its output can instead be sent to a file using the output redirect `>` :
```
For a first example IMO the basic redirect is better than the appending one.

src/Documentation/doc/en/programming/terminal_pipes_and_redirects.md
```diff
@@ -0,0 +1,72 @@
...
+async function read(ns) {
+  const stdin = ns.getStdin();
+  if (stdin.empty()) {
+    await stdin.nextWrite();
+  }
+  return stdin.read();
+}
+
+/**
+ * Each time that data comes in through stdin, call callback with it as input, until stdin is closed
```
d0sboots: I'd take this example out. Converting promise-based APIs to callback-based APIs is not needed for new code, only for interfacing with older callback-based APIs.

Edit: I remember what this is about now, less about interfacing and more about abstracting out the loop. I still think that's a net loss in clarity, but it works out fine in the internal code. However in an example where there is no code reuse to justify pulling this out, I think it is just a distraction.

src/Documentation/doc/en/programming/terminal_pipes_and_redirects.md
```diff
@@ -0,0 +1,72 @@
...
+
+### Creating your own command line utilities
+
+`cut.js` using `onRead()` from the snippet above
+
+```js
+/** @param {NS} ns */
+export async function main(ns) {
+  if (!ns.getStdin()) {
+    return ns.tprint("ERROR: No piped input given");
```
d0sboots:
```suggestion
    ns.tprint("ERROR: No piped input given");
    return;
```
Less confusing for newbies.

src/Documentation/doc/en/programming/terminal_pipes_and_redirects.md
```diff
@@ -0,0 +1,72 @@
...
+  // Other flags, such as '-b' bytes and '-d' delimeter, are left as an excercise for the reader
+  const flags = ns.flags([["c", "0"]]);
+  const charCountRange = flags.c.split("-");
+  const startCharCount = Number(charCountRange[0]?.trim());
+  const endCharCount = Number(charCountRange[1]?.trim() ?? startCharCount);
+
+  await onRead(ns, (data) => {
+    // slice the characters from the input data to specified range, and print them (aka send to stdout)
+    // tprintf is used to avoid printing the script's filename and line number before the message
+    ns.tprintf("%s", data.slice(startCharCount - 1, endCharCount));
```
d0sboots, line 56:
```suggestion
  let data = await read(ns);
  while(data != null) {
      ns.tprintf("%s", data.slice(startCharCount - 1, endCharCount));
      data = await read(ns);
    }
```
Especially as documentation, it is more clear to see the loop inline.

d0sboots (commented):
I think we need to have a discussion/refresher on how the async behavior of the pipe stuff works. I know there's complicated stuff I've forgotten about the criteria for when the terminal does or doesn't block.

src/Documentation/doc/en/programming/terminal_pipes_and_redirects.md
ficocelliguy: The idea I had is that I provide these utility functions, and they can be re-used in any of the player command-line utilities that they want to make. I'm planning to re-use it in this document as well - I'm thinking about  adding `wc` as a second simple example.

Since the interface to a player is a port handle, not a full StdIO, the player can't use an async iterator to go over the items in the same way we do internally. it seems like it would be nice to abstract away reading like this to keep the actual utility simple and clean.

d0sboots: I agree with the `read` utility function - even when only used once, IMO it abstracts out a logical unit of work that adds clarity when reading the code.

However, I disagree about onRead - all it is abstracting is a while loop. To me "call `onRead` instead of writing `while(condition)`", with its inherent inversion of control becomes harder to read, and it is not really saving any lines to do so. And I am an experienced developer who is used to callbacks and continuation-passing-style and patterns of this flavor, even if they aren't my favorite - for newer programmers (for some of whom the concept of "passing a function" is new), I think this will be even more of a digression from the true purpose.

src/Terminal/commands/cat.ts
```diff
@@ -37,9 +37,13 @@ export function cat(args: (string | number | boolean)[], server: BaseServer, std
 
   stdIO.write(output);
 
-  void callOnRead(stdIO, (data: unknown, stdInOut) => {
-    stdInOut.write(stringify(data));
-  });
+  if (stdinIsClosed) {
+    stdIO.close();
+  } else {
+    void callOnRead(stdIO, (data: unknown, stdInOut) => {
```
d0sboots, line 43: What's with the `void` here? It makes this look like a declaration, but this is a function call...

test/jest/Terminal/RedirectIO.test.ts
```diff
@@ -118,6 +118,7 @@ describe("RedirectIOTests", () => {
       expect(script?.content).toBe(scriptContent);
 
       await Terminal.executeCommands(`run ${scriptName} >> ${filename}`);
+      await sleep(50);
```
d0sboots, line 121: Tests shouldn't be using sleep either, although it's at least more of an understood hack for tests. Can you wait on the proper promise instead? 

Actually, don't worry about this right now - once I finish my refactor, executeCommands *will* be the proper promise.

d0sboots:
Ack! I was on autopilot

PR #2396, opened 2025-11-23
MISC: Refactor implementation of ports

d0sboots:
This moves the implementation entirely into the PortHandle class, which makes the code a bit more streamlined and will allow for other port implementations in the future.

- Manually smoke tested
```
[home /]> run eval.js "ns.writePort(3, 'foo');ns.tprint(ns.readPort(3));ns.tprint(ns.readPort(3))"
Running script with 1 thread, pid 4 and args: ["ns.writePort(3, 'foo');ns.tprint(ns.readPort(3));ns.tprint(ns.readPort(3))"].
eval.js: foo
eval.js: NULL PORT DATA
```


ficocelliguy (approved):
I'm able to pass data through ports on this branch as expected between scripts, and await port write. The refactor looks good (though github is a bit confused as to what was moved where XD )

d0sboots:
If you change the diff mode to ignore whitespace it becomes much more sensible.

PR #2397, opened 2025-11-23
BUGFIX: Avoid re-entrency issues with EventEmitter

d0sboots:
Subscribing or unsubscribing from within an event handler (possibly transitively) could cause issues or even infinite loops.

This was initially fixed in #2257, which exhibited such problems. Currently no code is known to have this issue, but it could easily come up again in the future.


d0sboots:
I'm pulling this one out and submitting it by itself.

PR #2398, opened 2025-11-23
REFACTOR: Move Result to the public API

d0sboots:
This refactors Result<T> to be part of the public NetscriptDefinitions. It is not used by anything in this PR, but it is planned to be used in the autoinfil APIs, or lacking that, future APIs, so this should not be exposing anything prematurely.


catloversg:
It looks good to me.

ficocelliguy (approved):

catloversg:
I merged this PR to use `Result` in #2443. Please let me know if there is a problem.

d0sboots:
Ah right I forgot about this one after one of the checks failed

PR #2399, opened 2025-11-25
BUGFIX: Coding contracts may have duplicate names

aaaa-imcute:
https://discord.com/channels/415207508303544321/415213413745164318/1442796583023415326

I don't know how npm works, and I don't know how to access the result of my branch (instead of the official one, which is on the official website) to test it, but I really hope there aren't any bugs.

Edit:
We check if the file name without the suffix ("contract-122706") exists, but data in `server.contracts` contains the suffix ("contract-122706.cct"), so that check does not work.

MRE:
```js
/** @param {NS} ns */
  console.clear();
  const s = new Set();
  for (let i = 0; i < 10000; ++i) {
    const filename = ns.codingcontract.createDummyContract("Algorithmic Stock Trader I");
    if (s.has(filename)) {
      console.error(filename, s);
      break;
    }
    s.add(filename);
  }
}
```

### DIFF
--- a/src/CodingContract/ContractGenerator.ts
+++ b/src/CodingContract/ContractGenerator.ts
@@ -252,25 +252,25 @@ function getRandomServer(): BaseServer | null {
   return randServer;
 }
 
+function getRandomB64String(length: number) {
+  const b64 = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/";
+  return Array(length)
+    .fill(0)
+    .map(() => b64[getRandomIntInclusive(0, 63)])
+    .join("");
+}
+
 function getRandomFilename(
   server: BaseServer,
   reward: ICodingContractReward = { type: CodingContractRewardType.Money },
-): ContractFilePath {
-  let contractFn: string,
-    i = 0;
-  let contractSuffix = ".cct";
+): ContractFilePath | null {
+  let contractFn = `contract-${getRandomB64String(6)}`;
   if ("name" in reward) {
     // Only alphanumeric characters in the reward name.
-    contractSuffix = `-${reward.name.replace(/[^a-zA-Z0-9]/g, "")}` + contractSuffix;
+    contractFn += `-${reward.name.replace(/[^a-zA-Z0-9]/g, "")}`;
   }
-  do {
-    contractFn = `contract-${getRandomIntInclusive(0, 1e6)}` + contractSuffix;
-  } while (
-    server.contracts.filter((c: CodingContract) => {
-      return c.fn === contractFn;
-    }).length != 0 &&
-    ++i < 1000
-  );
+  contractFn += ".cct";
+  if (server.contracts.filter((c: CodingContract) => c.fn === contractFn).length) return null;
   const validatedPath = resolveContractFilePath(contractFn);
   if (!validatedPath) throw new Error(`Generated contract path could not be validated: ${contractFn}`);
   return validatedPath;
--- a/src/CodingContract/ContractGenerator.ts
+++ b/src/CodingContract/ContractGenerator.ts
@@ -257,23 +257,21 @@ function getRandomFilename(
   reward: ICodingContractReward = { type: CodingContractRewardType.Money },
 ): ContractFilePath {
   let contractFn = `contract-${getRandomIntInclusive(0, 1e6)}`;
-
+  let contractSuffix = ".cct";
+  if ("name" in reward) {
+    // Only alphanumeric characters in the reward name.
+    contractSuffix = `-${reward.name.replace(/[^a-zA-Z0-9]/g, "")}` + contractSuffix;
+  }
   for (let i = 0; i < 1000; ++i) {
     if (
       server.contracts.filter((c: CodingContract) => {
         return c.fn === contractFn;
-      }).length <= 0
+      }).length == 0
     ) {
       break;
     }
-    contractFn = `contract-${getRandomIntInclusive(0, 1e6)}`;
+    contractFn = `contract-${getRandomIntInclusive(0, 1e6)}` + contractSuffix;
   }
-
-  if ("name" in reward) {
-    // Only alphanumeric characters in the reward name.
-    contractFn += `-${reward.name.replace(/[^a-zA-Z0-9]/g, "")}`;
-  }
-  contractFn += ".cct";
   const validatedPath = resolveContractFilePath(contractFn);
   if (!validatedPath) throw new Error(`Generated contract path could not be validated: ${contractFn}`);
   return validatedPath;
--- a/src/CodingContract/ContractGenerator.ts
+++ b/src/CodingContract/ContractGenerator.ts
@@ -86,6 +86,7 @@ export function generateRandomContract(): void {
   const problemType = getRandomProblemType(maxDif);
 
   const contractFn = getRandomFilename(randServer, reward);
+  if (contractFn == null) return; //Not generate the contract
   const contract = new CodingContract(contractFn, problemType, reward);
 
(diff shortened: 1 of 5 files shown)


aaaa-imcute:
it says build issue bc i don't know how types work in typescript, but i'm pretty sure contractFn has the right type

catloversg:
I edited the title and updated the description.

aaaa-imcute:
thanks :)

src/CodingContract/ContractGenerator.ts
```diff
@@ -256,24 +256,21 @@ function getRandomFilename(
...
-        return c.fn === contractFn;
-      }).length <= 0
-    ) {
-      break;
-    }
-    contractFn = `contract-${getRandomIntInclusive(0, 1e6)}`;
-  }
-
+  let contractFn: string,
+    i = 0;
```
d0sboots:
```suggestion
let i = 0;
```

src/CodingContract/ContractGenerator.ts
```diff
@@ -256,24 +256,21 @@ function getRandomFilename(
...
+    contractSuffix = `-${reward.name.replace(/[^a-zA-Z0-9]/g, "")}` + contractSuffix;
   }
-  contractFn += ".cct";
+  do {
+    contractFn = `contract-${getRandomIntInclusive(0, 1e6)}` + contractSuffix;
+  } while (
+    server.contracts.filter((c: CodingContract) => {
+      return c.fn === contractFn;
+    }).length != 0 &&
+    ++i < 1000
```
d0sboots: Prefix/postfix operators used for their value are overly subtle. Better to just put the increment inside the loop.

d0sboots (commented):
This fixes *a* bug, but not necessarily *the* bug, as was discussed on discord. Specifically, this fixes the duplicate check, which was very broken before, but that's not necessarily a good thing.

Before, the loop was essentially dead code, and you would start running into duplicates (with high probability) around the birthday number. Now, that won't happen, but the cost is that this function will start to get slower at that point. The average case should still be ok though, at least until you near 1000000 contracts.

In short, if we want to fix this code, this is how to do it. I'm not sure that we shouldn't just remove it, though.

catloversg:
How about switching to a 6-char base64 string like d0sboots suggested on Discord? I think we can remove the loop after that. If it still collides, we have many options to deal with it. For example, make `getRandomFilename` return `ContractFilePath | null`. If its callers see null, they just return early. It's fine with most callers, except `createDummyContract` API. With that API, we can throw an error or return an empty string to tell the player that it fails to generate a contract. If they generate 250k contracts without solving any one, it's fine if the API sometimes does not generate a new one.

src/CodingContract/ContractGenerator.ts
```diff
@@ -102,16 +103,19 @@ export function generateRandomContractOnHome(): void {
...
 }
 
-export const generateDummyContract = (problemType: CodingContractName): string => {
+export const generateDummyContract = (problemType: CodingContractName): string | null => {
   if (!CodingContractTypes[problemType]) throw new Error(`Invalid problem type: '${problemType}'`);
   const serv = Player.getHomeComputer();
 
   const contractFn = getRandomFilename(serv);
+  if (contractFn == null) return null; //Not generate the contract.
+  //It seems that only src/NetscriptFunctions/CodingContract.ts#L132 calls this function, and it can probably return null just fine.
```
d0sboots: This comment has the potential to get out of date quickly, and in general isn't needed. TS helps enforce this for us.

d0sboots (commented):
It's easier to just `npm run doc` you know :)

src/CodingContract/ContractGenerator.ts
aaaa-imcute: seems like ts is a much more robust language than js then.
maybe i should learn it one day :)

catloversg:
I tweaked some comments and added the breaking change notice. I'll let d0sboots review `getRandomB62String`.

aaaa-imcute:
thanks!

src/CodingContract/ContractGenerator.ts
```diff
@@ -252,28 +263,33 @@ function getRandomServer(): BaseServer | null {
   return randServer;
 }
 
+function getRandomB62String(length: number) {
+  const b62 = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789";
+  return Array(length)
+    .fill(0)
+    .map(() => b62[getRandomIntInclusive(0, 61)])
+    .join("");
+}
```
gmcew: I'm aware I'm not d0s and I also don't know how JS is optimised, so this is for my own knowledge when this gets reviewed and is an observation only!

Instead of creating an array and then joining it all together afterwards - would it be cleaner to start with an empty string and add the random character on the back directly? E.g.:
```js
  let result = "";
  for (let i = 0; i < length; i++) {
    result += b62.charAt(Math.random() * 62);
  }
  return result;
  ```

Comment (d0sboots) [reply src/CodingContract/ContractGenerator.ts]: Either works. I tend to prefer the imperative style as well, since it optimizes better, but speed is not a concern here and I'm trying to avoid nitpicking this overmuch. 

One nice thing about using Random directly plus charAt is you can naturally use b62. length instead of hardcoding it.

**Comment by d0sboots** [comment]
I think the only thing left here is a smoke test to make sure contact generation still works.

**Comment by catloversg** [comment]
Adding Jest tests may be a bit hard for a new contributor, so I did it for them.

Other changes:
- Change `b62` to "alphanumeric".
- Rewrite the utility function in the imperative style as suggested.

Test code for the probability of collision mentioned in the Jest tests:
```js
/** @param {NS} ns */
  const n = 1000;
  const d = Math.pow(62, 6);
  console.log(1 - Math.exp(-1 * (n * n) / (2 * d)));
}
```
Reference: https://en.wikipedia.org/wiki/Birthday_problem

aaaa-imcute:
Thanks!
What are jest tests?I stumbled upon them yesterday when trying to run npm run doc.
Also,half your tests don't pass due to floating imprecision or something. They complain about changed formulas.

catloversg:
What is the error?

aaaa-imcute:
something like blah blah blah formulas changed but they should stay the same. (The error is really small and looks like float imprecision)

Oh,I didn't mean *your* tests in this pull request, I meant tests on completely unrelated mechanisms when I ran them yesterday.

src/CodingContract/ContractGenerator.ts
aaaa-imcute: well,this was just my functional programming brain :P

catloversg:
If you are using Node 24, just switch back to Node 22 when testing Bitburner: https://discord.com/channels/415207508303544321/459097632896188436/1419898835320442901.

d0sboots:
> Adding Jest tests may be a bit hard for a new contributor, so I did it for them.

By "smoke test" I just meant manually making sure that it still works. But, automated tests are always better, so thank you for adding that.

aaaa-imcute:
> If you are using Node 24, just switch back to Node 22 when testing Bitburner: https://discord.com/channels/415207508303544321/459097632896188436/1419898835320442901.

how do I know which node I am using?

aaaa-imcute:
Thanks for testing and merging it :D

catloversg:
> how do I know which node I am using?

You can check the output of `node -v`.

aaaa-imcute:
v24.11.1

How do I temporary downdate it then?

catloversg:
With tools like nvm (https://github.com/nvm-sh/nvm), you can switch versions easily. If you install it with things like an .msi file on Windows, you may have to install another version and deal with the path, depending on your installation.

PR #2415, opened 2025-12-12
IPVGO: Do not update captures on passed analysis boards

ficocelliguy:
Based on discussion with players, this PR changes the ns.go.analysis functions to analyze the given boards exactly as-given, without evaluating captures. This ensures that the player gets feedback on precisely what they entered, and not some assumed/future state of the board.

Context:
https://discord.com/channels/415207508303544321/1221920484111814828/1433628079888007239

### DIFF
--- a/src/ScriptEditor/NetscriptDefinitions.d.ts
+++ b/src/ScriptEditor/NetscriptDefinitions.d.ts
@@ -4820,7 +4820,7 @@ export interface GoAnalysis {
    *
    * @param boardState - The initial board state to use for the new game, in the format used by getBoardState().
    * @param komi - Optional komi value to set for the game. Defaults to 5.5.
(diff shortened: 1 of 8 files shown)


d0sboots:
This should have a note in the breaking changes.

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -4820,8 +4820,9 @@ export interface GoAnalysis {
    *
    * @param boardState - The initial board state to use for the new game, in the format used by getBoardState().
    * @param komi - Optional komi value to set for the game. Defaults to 5.5.
+   * @param nextPlayerIsWhite - Optional. Whether or not the next player to play is the white player. Defaults to true.
```
d0sboots: Boolean optional params typically should default to false, and in Go typically black goes first?

PR #2417, opened 2025-12-14
CONTRACTS: Generate test contracts on executing host by default. Add support for optional parameter to specify the server.

1337JiveTurkey:
This modifies the `ns.codingcontract.createDummyContract()` method to be more consistent with other methods that operate on filesystems. First, it generates contracts by default on the executing host. Second it accepts an optional hostname argument that specifies a host to generate the contract on.

I added the change to the default in this PR in part because it matched the other methods in `CodingContract.ts` to generate the `BaseServer` there and pass that in rather than handle the default in `ContractGenerator.ts`.

This is a **breaking change** but I'm guessing the number of people depending on code running on remote servers to generate files on home is very small compared to people creating dummy contracts on home from home.

Testing:

The test script used:
```ts
  const sqrt = ns.enums.CodingContractName.SquareRoot
  const local = ns.codingcontract.createDummyContract(sqrt)
  const remote = ns.codingcontract.createDummyContract(sqrt, "n00dles")
  ns.tprintf("Created %s locally and %s remotely", local, remote)
}
```
* A script was created which generates a Square Root test contract locally and on `n00dles` using the optional argument. It also tests that the filenames are returned regardless of the server that it's generated on.
* The script was copied to `foodnstuff` and run from there.
* The script was modified so that `n00dles` is `blah` instead. The program generated an error modal as expected.
Screenshot 2025-12-14 at 12 39 13 AM
* The script was also modified so that `n00dles` is `home` instead and it behaves as expected, creating both scripts on the home server when run from home.

### DIFF
--- a/src/ScriptEditor/NetscriptDefinitions.d.ts
+++ b/src/ScriptEditor/NetscriptDefinitions.d.ts
@@ -4057,7 +4057,7 @@ export interface CodingContract {
    * @remarks
    * RAM cost: 2 GB
    *
-   * Generate a dummy contract on the current computer with no reward. Used to test various algorithms.
+   * Generate a dummy contract on the current server with no reward. Used to test various algorithms.
    *
    * This function will return null and not generate a contract if the randomized contract name is the same as another contract's name.
    *
--- a/src/ScriptEditor/NetscriptDefinitions.d.ts
+++ b/src/ScriptEditor/NetscriptDefinitions.d.ts
@@ -4057,14 +4057,15 @@ export interface CodingContract {
    * @remarks
    * RAM cost: 2 GB
    *
-   * Generate a dummy contract on the home computer with no reward. Used to test various algorithms.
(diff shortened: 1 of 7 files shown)


src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -4057,14 +4057,15 @@ export interface CodingContract {
    * @remarks
    * RAM cost: 2 GB
    *
-   * Generate a dummy contract on the home computer with no reward. Used to test various algorithms.
+   * Generate a dummy contract on the current computer with no reward. Used to test various algorithms.
```
```suggestion
   * Generate a dummy contract on the current server with no reward. Used to test various algorithms.
```

catloversg (commented):
After we merge #2418, you can use the migration example I posted there.

1337JiveTurkey:
The migration code is directly pulled from #2418 but changed so that the migrator ignores parameter blocks with commas in them. This ensures that even in the event that the migration runs multiple times, it won't add the home parameter multiple times.

PR #2419, opened 2025-12-14
UI: Add option to set fractional digits

CTNOriginals:
Adds a slider to the Numeric Display options page which effects the amount of fractional digits displayed in a regular number.
The slider allows a value between 0 and 5

Settings default:
Settings menu

0 fractional digits:
0 fractional digits:

5 fractional digits:
5 fractional digits

### Possible edge case
Setting the option to 0 works as intended, the formatter removes the separator entirely and displays an integer instead of any amount of fractions behind it. 
Although, I did notice a minor edge case where the number sometimes displays a comma separated value like for example 1,000K when its really close to being a million, like 999995. 
I dont think this is caused by the changes applied in this fork.

### Relevant commits
- MISC: Add fractional digit settings field
- MISC: formatNumber fractionalDigits parameter default to Settings field
- UI: Add Fractional Digits option slider

### Credits
Thanks to @ficocelliguy for explaining the ropes of how to set up a PR.
Thanks to @shyguy1412 for being there at the right time with right answer to something very specific.

### DIFF
--- a/src/GameOptions/ui/NumericDisplayOptions.tsx
+++ b/src/GameOptions/ui/NumericDisplayOptions.tsx
@@ -85,7 +85,7 @@ export const NumericDisplayPage = (): React.ReactElement => {
         step={1}
         min={0}
         max={5}
-        tooltip={<>The number of digits of precision to display on small numbers.</>}
+        tooltip={<>The default number of decimal places to display on small numbers.</>}
       />
       <Select startAdornment={<Typography>Locale&nbsp;</Typography>} value={locale} onChange={handleLocaleChange}>
         <MenuItem value="en">en</MenuItem>
--- a/src/GameOptions/ui/NumericDisplayOptions.tsx
+++ b/src/GameOptions/ui/NumericDisplayOptions.tsx
@@ -65,12 +65,7 @@ export const NumericDisplayPage = (): React.ReactElement => {
         step={1}
         min={0}
         max={5}
-        tooltip={
-          <>
-            The default number of decimal places to display on small numbers.
-            Default value: 3
-          </>
-        }
+        tooltip={<>The default number of decimal places to display on small numbers. Default value: 3</>}
       />
       <OptionSwitch
         checked={Settings.hideTrailingDecimalZeros}
--- a/src/GameOptions/ui/NumericDisplayOptions.tsx
+++ b/src/GameOptions/ui/NumericDisplayOptions.tsx
@@ -13,6 +13,7 @@ export const NumericDisplayPage = (): React.ReactElement => {
     Settings.fractionalDigits = newValue as number;
     FormatsNeedToChange.emit();
   }
+  
   function handleLocaleChange(event: SelectChangeEvent): void {
     setLocale(event.target.value);
     Settings.Locale = event.target.value;
--- a/src/GameOptions/ui/NumericDisplayOptions.tsx
+++ b/src/GameOptions/ui/NumericDisplayOptions.tsx
@@ -13,7 +13,7 @@ export const NumericDisplayPage = (): React.ReactElement => {
     Settings.fractionalDigits = newValue as number;
     FormatsNeedToChange.emit();
   }
-  
+
   function handleLocaleChange(event: SelectChangeEvent): void {
     setLocale(event.target.value);
     Settings.Locale = event.target.value;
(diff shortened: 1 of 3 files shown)


src/GameOptions/ui/NumericDisplayOptions.tsx
```diff
@@ -72,6 +77,15 @@ export const NumericDisplayPage = (): React.ReactElement => {
           <>If this is set all references to memory will use GiB instead of GB, in accordance with IEC 60027-2.</>
         }
       />
+      <OptionsSlider
```
d0sboots: I would move this up to between "Hide thousands separator" and "Hide trailing fractional zeroes for decimals". I think it fits best there? (Basically, grouping similar settings together, and introduce the idea of "fractional digits" before the more specific setting of "hide zeroes")

src/GameOptions/ui/NumericDisplayOptions.tsx
```diff
@@ -72,6 +77,15 @@ export const NumericDisplayPage = (): React.ReactElement => {
...
+        label="Fractional Digits"
+        initialValue={Settings.fractionalDigits}
+        callback={handleFractionalDigitChange}
+        step={1}
+        min={0}
+        max={5}
+        tooltip={<>The number of digits of precision to display on small numbers.</>}
```
d0sboots:
```suggestion
        tooltip={<>The default number of decimal places to display on small numbers.</>}
```
"Decimal places"/"fractional digits" are not the same as "digits of precision". 12345.6789 to 3 fractional digits is 12345.679. To 3 digits of precision, it is 12300.

IMO also useful to mention "default", since lots of methods override this and thus won't be affected by this setting.

src/GameOptions/ui/NumericDisplayOptions.tsx
```diff
@@ -4,10 +4,15 @@ import { Settings } from "../../Settings/Settings";
...
 import { FormatsNeedToChange } from "../../ui/formatNumber";
+import { OptionsSlider } from "./OptionsSlider";
 
 export const NumericDisplayPage = (): React.ReactElement => {
   const [locale, setLocale] = useState(Settings.Locale);
 
+  function handleFractionalDigitChange(_event: Event | React.SyntheticEvent, newValue: number | number[]): void {
+    Settings.fractionalDigits = newValue as number;
+    FormatsNeedToChange.emit();
+  }
```
d0sboots, line 15:
```suggestion
  }
  
```
Style: Newline between functions.

src/Settings/Settings.ts
```diff
@@ -192,6 +192,8 @@ export const Settings = {
   useEngineeringNotation: false,
   /** Whether to disable suffixes and always use exponential form (scientific or engineering). */
   disableSuffixes: false,
+  /** The amount of digits displayed after the period '.' in a number. */
```
d0sboots:
```suggestion
  /** The default amount of digits displayed after the period '.' in a number. */
```

ficocelliguy:
That's a VERY long slider XD
You can probably cap it at like 300px width or something?

d0sboots:
I thought the same, but it's consistent with settings in other tabs :)

src/GameOptions/ui/NumericDisplayOptions.tsx
CTNOriginals: Ah yeah fair.
I either wanted to do that, or just place myself near the bottom.
But I agree it would be nicer to group things but it does look at little inconsistent as all other options are toggles.

src/GameOptions/ui/NumericDisplayOptions.tsx
CTNOriginals: Oh man you would not believe how long i was stuck on the wording of this, usually im fine with it but yesterday was not an english language day i guess. thanks for the improvement.

src/Settings/Settings.ts
```diff
@@ -192,6 +192,8 @@ export const Settings = {
   useEngineeringNotation: false,
   /** Whether to disable suffixes and always use exponential form (scientific or engineering). */
   disableSuffixes: false,
+  /** The default amount of digits displayed after the period '.' in a number. */
```
catloversg: Optional suggestion:
```suggestion
  /** The default amount of digits displayed after the decimal separator. */
```
"Decimal point" and "decimal comma" are more common terms, but their usage depends on countries.

catloversg (commented):
You also need to rerun `npm run format` after applying other suggestions.

catloversg (approved):

Issue #2401, opened 2025-11-25
Error: Could not locate constructor named FileMetadata. If the save data is valid, this is a bug. (at "Loading Screen")


Johnthesuper117:
### How did this happen?

Please fill this information with details if relevant.

- [ ] Save file
- [ ] Minimal scripts to reproduce the issue
- [ ] Steps to reproduce

### Environment

* Error: Error: Could not locate constructor named FileMetadata. If the save data is valid, this is a bug.
* Page: Loading Screen
* Version: v2.8.1 (d0d776700)
* Environment: Production
* Platform: Browser
* UserAgent: Mozilla/5.0 (X11; CrOS x86_64 14541.0.0) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/142.0.0.0 Safari/537.36
* Features: lang=en-US cookiesEnabled=true doNotTrack=null indexedDb=true
* Source: undefined

### Stack Trace
```
Error: Could not locate constructor named FileMetadata. If the save data is valid, this is a bug.
    at Object.o (https://bitburner-official.github.io/dist/main.bundle.js:9:1150783)
    at JSON.parse (<anonymous>)
    at A (https://bitburner-official.github.io/dist/main.bundle.js:9:589732)
    at T (https://bitburner-official.github.io/dist/main.bundle.js:9:533532)
    at async Object.load (https://bitburner-official.github.io/dist/main.bundle.js:9:892586)
```

### React Component Stack
```
undefined
```

### Save
```

[RECOVERY_BITBURNER_1764081371.json.gz](https://github.com/user-attachments/files/23749297/RECOVERY_BITBURNER_1764081371.json.gz)

```

d0sboots:
Further comment: If you want to switch back, use the save that was exported when you switched. You can't switch back from 3.0 with your current save.

Johnthesuper117:
> Further comment: If you want to switch back, use the save that was exported when you switched. You can't switch back from 3.0 with your current save.

It worked! Thank you! And thanks for the really fun game!

d0sboots:
We might want to consider a general downgrade check with a more useful message to help cover this case.

catloversg:
Agree. I have a stashed change for that. I'll polish it and make a PR.

Issue #2392, opened 2025-11-22
Error: Could not locate constructor named FileMetadata. If the save data is valid, this is a bug. (at "Loading Screen")


Notmim:
### How did this happen?

Please fill this information with details if relevant.

- [ ] Save file
- [ ] Minimal scripts to reproduce the issue
- [ ] Steps to reproduce
-- Just booted the game in Firefox browser as always and happened this. 

### Environment

* Error: Error: Could not locate constructor named FileMetadata. If the save data is valid, this is a bug.
* Page: Loading Screen
* Version: v2.8.1 (d0d776700)
* Environment: Production
* Platform: Browser
* UserAgent: Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:145.0) Gecko/20100101 Firefox/145.0
* Features: lang=en-US cookiesEnabled=true doNotTrack=unspecified indexedDb=true
* Source: https://bitburner-official.github.io/dist/main.bundle.js

### Stack Trace
```
o@https://bitburner-official.github.io/dist/main.bundle.js:9:1150783
A@https://bitburner-official.github.io/dist/main.bundle.js:9:589732
T@https://bitburner-official.github.io/dist/main.bundle.js:9:533532

```

### React Component Stack
```
undefined
```

### Save
```
Copy your save here if possible
[RECOVERY_BITBURNER_1763801384.json.gz](https://github.com/user-attachments/files/23687827/RECOVERY_BITBURNER_1763801384.json.gz)

d0sboots:
This seems like an error from trying to load a 3.0 save in 2.8.1.

PR #2425, opened 2025-12-18
UI: Show useful error messages when loading unsupported save data from newer versions

catloversg:
When the player loads the save data from the dev build on the stable build, we receive reports like #2392 and #2401. The current error message is useless because most players do not understand what that error message (i.e., "Could not locate constructor") means. This PR improves the error message in this case.

Save file for testing:
[v44.gz](https://github.com/user-attachments/files/24241305/v44.gz)
[v999.gz](https://github.com/user-attachments/files/24241304/v999.gz)

v44:

v44

v999:

v999

### DIFF
--- a/src/SaveObject.ts
+++ b/src/SaveObject.ts
@@ -160,6 +160,20 @@ function assertParsedSaveData(parsedSaveData: unknown): asserts parsedSaveData i
   }
 }
 
+/**
+ * We sometimes need the raw data in the loaded save object for debugging and showing useful error messages. This object
+ * contains only what we need.
+ */
+export const loadedSaveObjectMiniDump = {
+  /**
+   * If VersionSave exists, it is always a string. It has 3 formats:
+   * - x.y: Very early versions (0.1-0.17) used this format.
+   * - x.y.z: Starting from roughly 0.17, we used this format. Note that in some commits, we mistakenly used the x.y format.
+   * - x: Starting from v1, we used the version number instead of the version string.
+   */
+  VersionSave: undefined as string | undefined,
+};
+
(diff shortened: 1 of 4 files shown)


src/SaveObject.ts
```diff
@@ -426,6 +440,18 @@ async function loadGame(saveData: SaveData): Promise<boolean> {
   const jsonSaveString = await decodeSaveData(saveData);
 
   const saveObj: unknown = JSON.parse(jsonSaveString, Reviver);
+
+  // Extract VersionSave ASAP for debugging and showing useful error messages later. Some checks here are redundant (
+  // e.g., the object assertion) because we will do them again later, but that's okay.
+  if (
```
d0sboots, line 446: Although I went ahead and merged this, the global state here bugs me, because it can be set in ways that aren't actually loading a game (for instance, cancelling the load after looking at stats.)

Could we catch the JsonReviverError inside the load process and rethrow it as a InvalidVersionError instead? (With the version detail attached to the error object.)

catloversg: I thought about it too, but I did not find a good way to do that. One viable solution is to wrap all following loading code (the calls of `loadPlayer`, `loadAllServers`, etc. in the `loadGame` function) in a big try-catch block and rethrow `InvalidVersionError`, but that's too ugly. Another solution is to wrap only `JSON.parse` in each loading function (`loadPlayer`, `loadAllServers`, etc.), but that's too bothersome.

When checking the callers of `loadGame`, I found that it's only called by `Engine.load`, which is only called once in the loading screen, so I decided to set the global state this way. It's a bit fragile (it becomes buggy when we mistakenly set it in the wrong location, like you said), but it works now. If you find a better solution, please let me know.

PR #2428, opened 2025-12-19
API: RunningScript add start/end time fields

CTNOriginals:
Add 2 fields to the `RunningScript` interface:
- startTime: The UTC time code of when the script was started
- endTime: The UTC time code of when the script ended (-1 while running)

While i was playing bitburner, i was trying to create a watcher script that would reload any script that has a `--watch` argument, including any script that has stoped running where the tail is still opened.
While creating this watcher script i noticed that i do not have any way of knowing how long ago a script has been started.
I needed this to compare to the scripts modification time, so a `startTime` field was a nice way of solving that and i put it in the game. The `endTime` just came along with it as i already knew how to implement that anyway and thought it would not hurt to also add...

example usage of `startTime`: 
```ts
for (const pid in ScriptMap) {
	const script = ScriptMap[pid]
	const modTime = ns.getFileMetadata(script.filename).mtime

	if (script.startTime < modTime) {
		restartScript(ns, script)
	}
}
```

The reason i added internal functions the the RunningScript class in the source code for `onStart` and `onEnd` is to have the functionality happen in one place and also to account for the fact that scripts may be started/ended from other location later on.


d0sboots:
I'm not completely against this, but the bar is a bit higher for adding stuff to RunningScript because it is on the critical path, both in terms of CPU and memory, for very intensive operations. (Specifically, batchers indirectly create hundreds of thousands of these.)

It seems like this is something that can be handled well enough from within your own code? I.e. capture the start time at the beginning of main, capture the end time within an atExit hook. Propagating that to other scripts depends on the specifics of how you communicate between scripts in general (globals, module-scoped variables, ports, etc.) but is also well within the bounds of what we expect players to handle.

CTNOriginals:
@d0sboots I understand.
Before i call quits though, if it is truly the size of the addition being added i would not mind taking out the `endTime` functionality and field, but i would still like to see `startTime` added if there would be a chance for it.

that said;
To be honest i was prepared for this feature getting declined as it started out more as a "how easy would this be to add" and ended in "i could just PR this rn" without fully discussing it on discord.. so it was worth the shot...

catloversg:
I agree with d0sboots. This feature is something that should be implemented in the userland.

CTNOriginals:
Yeah after exploring the ingame possibility a bit i also found a neat way to simply add this functionality, so i guess this was a fun practice PR at best.
Thanks for the feedback and your time.

PR #2431, opened 2025-12-20
UI: Add Script Editor toggle for Sticky Scroll

### LINKED ISSUE #2430: BUG: Cannot disable Sticky Scroll
Hello fine folks.

The toggle to disable Sticky Scroll (pictured below when right clicking on the UI) no longer seems to function, sticky scroll stays on matter what:

Is there an alternative way to disable this (I don't recall sticky scroll being a feature of the previous versions, and I cannot find previous Issues or PR's referencing it)?

Thank you

Image

gmcew:
Closes #2430 
There may have been a more elegant way to do this, I'll delete/rework this PR if I missed a simpler/more technically correct answer.

I also changed the default setting to off, which was the behaviour before this bug was identified. This can be reverted to current behaviour if preferred.

Ran `lint`,`format`,`test` and `doc` just in case. Screenshots of toggle and effect:
stickyOff
stickyOn

### DIFF
--- a/src/ScriptEditor/ui/OptionsModal.tsx
+++ b/src/ScriptEditor/ui/OptionsModal.tsx
@@ -11,7 +11,7 @@ import EditIcon from "@mui/icons-material/Edit";
 import { useBoolean } from "../../ui/React/hooks";
 import { Modal } from "../../ui/React/Modal";
 import { ThemeEditorModal } from "./ThemeEditorModal";
-import { CursorBlinking, CursorStyle, Options, StickyScroll } from "./Options";
+import { CursorBlinking, CursorStyle, Options } from "./Options";
 
 const CURSOR_STYLES: CursorStyle[] = ["line", "block", "underline", "line-thin", "block-outline", "underline-thin"];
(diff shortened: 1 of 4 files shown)


src/ScriptEditor/ui/OptionsModal.tsx
```diff
@@ -150,6 +150,14 @@ export function OptionsModal(props: OptionsModalProps): ReactElement {
           checked={props.options.beautifyOnSave}
         />
       </div>
+
+      <div style={{ display: "flex", alignItems: "center" }}>
+        <Typography marginRight={"auto"}>Enable Sticky Scroll: </Typography>
+        <Switch
+          onChange={(e) => props.onOptionChange("stickyScroll", { enabled: e.target.checked } as StickyScroll)}
```
catloversg: We don't need to typecast here. I guess you copy/paste this part from the cursor blinking option above, but we also don't need to typecast there. You can remove the typecasting in both options.

catloversg:
I'm fine with changing the default value to false, but you should mention it in the PR's description.

PR #2435, opened 2025-12-27
MISC: Update error message when backdooring without admin rights

CicaProductions:
Misc: Fixed lack of exclamation mark when backdooring without admin rights

Tested: npm run start:dev
note the images attached (alpha-ent is pre fix, n00dles is post-fix)
Exclamation mark is added in the post-fix screenshot

image_2025-12-27_125847056
Screenshot 2025-12-27 at 12 57 24

### DIFF
--- a/src/Constants.ts
+++ b/src/Constants.ts
@@ -170,7 +170,7 @@ export const CONSTANTS = {
 
 ### MISC
 
-- Fixed lack of exclamation mark when backdooring without admin rights (#2416) (@kordhell)
+- Fixed lack of exclamation mark when backdooring without admin rights (#2436) (@kordhell)
 - Ensure IPvGO board has at least 1 offline node (#2072) (@ficocelliguy)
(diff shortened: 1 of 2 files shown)


src/Constants.ts
```diff
@@ -170,6 +170,7 @@ export const CONSTANTS = {
 
 ### MISC
 
+- Fixed lack of exclamation mark when backdooring without admin rights (#2436) (@kordhell)
```
catloversg: You should revert this change. This file is usually updated when we update the changelog and the version number. We don't do that in each PR.

CicaProductions: Resolved, contributions.md is quite outdated and a little confusing.

d0sboots:
> contributions.md is quite outdated and a little confusing.

We'd welcome changes to make it less confusing and especially to fix places where it's out-of-date. New contributors are our best way of getting fresh eyes on it.

PR #2440, opened 2026-01-05
CONTRACTS: Display contract answers on completely failed contracts

### LINKED ISSUE #2411: Return correct answer when contract self-destructs
When a contract self-destructs, it would be useful for it to return what answer it would have accepted. It would be too late to be useful in gaining the reward, but might be helpful in debugging the code that was failing.

So instead of:

```
[home /]> run contract-190576.cct 
Contract FAILED - Contract is now self-destructing
```

We would get

```
[home /]> run contract-190576.cct 
Contract FAILED - Contract is now self-destructing. Correct answer: "8nnZnKdKs380215Uteo3258N0XWkErO750919FfhSZk67L02q373"//
```

TheCleric:
closes #2411

### DIFF
--- a/src/CodingContract/ContractTypes.ts
+++ b/src/CodingContract/ContractTypes.ts
@@ -33,6 +33,8 @@ interface CodingContractType<Data, Answer, State = Data> {
   difficulty: number;
   /** Function that generates a valid 'state' for a contract type */
   generate: () => State;
+  /** Function that returns an answer, if possible, for a given contract */
+  getAnswer: (data: Data) => Answer | null;
   /**
    * Transforms the 'state' for a contract into its 'data'. The state is
    * stored persistently as JSON, so it must be serializable. The data is what
@@ -52,13 +54,18 @@ interface CodingContractType<Data, Answer, State = Data> {
   validateAnswer: (answer: unknown) => answer is Answer;
 }
 
+type CodingContractGetAnswerExtension<Data, Answer> =
+  | { getAnswer: (data: Data) => Answer | null }
+  | { getAnswer: never };
+
 // This simple alias uses State == Data, and omits getData since it won't be used in this case.
-type CodingContractSimpleType<Data, Answer> = Omit<CodingContractType<Data, Answer, Data>, "getData">;
+type CodingContractSimpleType<Data, Answer> = Omit<CodingContractType<Data, Answer, Data>, "getData"> &
+  CodingContractGetAnswerExtension<Data, Answer>;
 
 // This alias has unique State and Data, and requires getData.
 type CodingContractComplexType<Data, Answer, State> = Omit<CodingContractType<Data, Answer, State>, "getData"> & {
   getData: (state: State) => Data;
-};
+} & CodingContractGetAnswerExtension<Data, Answer>;
 
 type CodingContractDefinitions<Signatures extends Record<string, [unknown, unknown] | [unknown, unknown, unknown]>> = {
   [T in keyof Signatures]: Signatures[T] extends [unknown, unknown, unknown]
--- a/src/CodingContract/ContractTypes.ts
+++ b/src/CodingContract/ContractTypes.ts
@@ -39,7 +39,7 @@ interface CodingContractType<Data, Answer, State = Data> {
    * Transforms the 'state' for a contract into its 'data'. The state is
    * stored persistently as JSON, so it must be serializable. The data is what
    * is given to the user and shown in the description. If this function is
-   * ommitted, it will be the identity function (i.e. State == Data).
+   * omitted, it will be the identity function (i.e. State == Data).
    * You can use this to make problems where the "solver" is not a function
    * that can be copy-pasted to user code to solve the problem.
    */
--- a/src/CodingContract/ContractTypes.ts
+++ b/src/CodingContract/ContractTypes.ts
@@ -54,18 +54,13 @@ interface CodingContractType<Data, Answer, State = Data> {
   validateAnswer: (answer: unknown) => answer is Answer;
 }
 
-type CodingContractGetAnswerExtension<Data, Answer> =
-  | { getAnswer: (data: Data) => Answer | null }
-  | { getAnswer: never };
-
 // This simple alias uses State == Data, and omits getData since it won't be used in this case.
-type CodingContractSimpleType<Data, Answer> = Omit<CodingContractType<Data, Answer, Data>, "getData"> &
-  CodingContractGetAnswerExtension<Data, Answer>;
+type CodingContractSimpleType<Data, Answer> = Omit<CodingContractType<Data, Answer, Data>, "getData">;
 
 // This alias has unique State and Data, and requires getData.
 type CodingContractComplexType<Data, Answer, State> = Omit<CodingContractType<Data, Answer, State>, "getData"> & {
   getData: (state: State) => Data;
-} & CodingContractGetAnswerExtension<Data, Answer>;
+};
 
 type CodingContractDefinitions<Signatures extends Record<string, [unknown, unknown] | [unknown, unknown, unknown]>> = {
   [T in keyof Signatures]: Signatures[T] extends [unknown, unknown, unknown]
--- a/src/CodingContract/contracts/ArrayJumpingGame.ts
+++ b/src/CodingContract/contracts/ArrayJumpingGame.ts
@@ -111,7 +111,7 @@ export const arrayJumpingGame: Pick<
       return jumps;
     },
     solver: (data, answer) => {
-      return arrayJumpingGame[CodingContractName.ArrayJumpingGame].getAnswer(data) === answer;
+      return arrayJumpingGame[CodingContractName.ArrayJumpingGameII].getAnswer(data) === answer;
     },
     convertAnswer: (ans) => parseInt(ans, 10),
     validateAnswer: (ans): ans is number => typeof ans === "number",
--- a/src/CodingContract/contracts/ArrayJumpingGame.ts
+++ b/src/CodingContract/contracts/ArrayJumpingGame.ts
@@ -37,14 +37,17 @@ export const arrayJumpingGame: Pick<
       return arr;
     },
     numTries: 1,
-    solver: (data, answer) => {
+    getAnswer: (data) => {
       const n: number = data.length;
       let i = 0;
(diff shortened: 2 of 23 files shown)


src/CodingContract/ContractTypes.ts
```diff
@@ -33,6 +33,8 @@ interface CodingContractType<Data, Answer, State = Data> {
   difficulty: number;
   /** Function that generates a valid 'state' for a contract type */
   generate: () => State;
+  /** Function that returns an answer, if possible, for a given contract */
+  getAnswer: (data: Data) => Answer | null;
```
d0sboots, line 37: This is the wrong type, but I think it would work better to use this type.
What you wrote is a function that returns an Answer or null, but getAnswer always returns an Answer (when it is defined). However, I think it will make everything much simpler if you adopt this as the type and then remove all the complex type stuff you're trying. Just have the contracts that don't return an answer return null. We don't need the ability to distinguish them at compile time.

TheCleric: > This is the wrong type, but I think it would work better to use this type. What you wrote is a function that returns an Answer or null, but getAnswer always returns an Answer (when it is defined). However, I think it will make everything much simpler if you adopt this as the type and then remove all the complex type stuff you're trying. Just have the contracts that don't return an answer return null. We don't need the ability to distinguish them at compile time.

Sorry I forgot I had the mixture in here. Your thinking is the same as mine. I had started the path of making it so getAnswer may or may not be there, but it is much easier and less clunky to always have it, but it returns null on contract types without a defined answer.

I'll proceed on that path if you think this is promising.

TheCleric:
Of note here, the following contract types will always return null:

- Square root => as discussed.
- Shortest Path In A Grid => as we don't really generate an answer as much as calculate what the minimum steps it would take and make sure the answer conforms. We COULD generate an answer, but did not want to inject any sort of solution finder into the code.
- Proper 2 Coloring Of A Graph => similar to above

src/CodingContract/contracts/ArrayJumpingGame.ts
```diff
@@ -105,7 +108,10 @@ export const arrayJumpingGame: Pick<
         lastJump = jumpedFrom;
         jumps++;
       }
-      return jumps === answer;
+      return jumps;
+    },
+    solver: (data, answer) => {
+      return arrayJumpingGame[CodingContractName.ArrayJumpingGame].getAnswer(data) === answer;
```
catloversg:
```suggestion
      return arrayJumpingGame[CodingContractName.ArrayJumpingGameII].getAnswer(data) === answer;
```

src/CodingContract/contracts/Compression.ts
```diff
@@ -97,6 +100,9 @@ export const compression: Pick<
     generate: (): string => {
       return comprLZEncode(comprLZGenerate());
     },
+    getAnswer: (compr) => {
+      return comprLZDecode(compr) ?? "";
+    },
     solver: (compr, answer) => {
       return (comprLZDecode(compr) ?? "") === answer;
```
catloversg: You should use the pattern `return foo[CodingContractName.Bar].getAnswer(baz) === answer;` here and some other places.

src/CodingContract/contracts/Compression.ts
```diff
@@ -135,6 +141,9 @@ export const compression: Pick<
     generate: (): string => {
       return comprLZGenerate();
     },
+    getAnswer: (plain) => {
+      return comprLZEncode(plain);
+    },
     solver: (plain, answer) => {
       return answer.length <= comprLZEncode(plain).length && comprLZDecode(answer) === plain;
```
catloversg: Same here.

src/CodingContract/contracts/FindAllValidMathExpressions.ts
```diff
@@ -86,7 +86,12 @@ export const findAllValidMathExpressions: Pick<CodingContractTypes, CodingContra
       const result: string[] = [];
       helper(result, "", num, target, 0, 0, 0);
 
-      if (result.length !== answer.length) return false;
+      return result;
+    },
+    solver: (data, answer) => {
+      const result = findAllValidMathExpressions[CodingContractName.FindAllValidMathExpressions].getAnswer(data);
+
+      if (result === null || result.length !== answer.length) return false;
```
catloversg: `getAnswer` can return null (with 3 contracts now), but in this contract, it cannot return null. If it's null here, there must be a bug in `getAnswer`, and we should notify both the player and ourselves. `exceptionAlert` is useful in this case.

src/CodingContract/contracts/GenerateIPAddresses.ts
```diff
@@ -51,6 +51,10 @@ export const generateIPAddresses: Pick<CodingContractTypes, CodingContractName.G
         }
       }
 
+      return ret;
+    },
+    solver: (data, answer) => {
+      const ret = generateIPAddresses[CodingContractName.GenerateIPAddresses].getAnswer(data) as string[];
```
catloversg: In general, we restrict the usage of typecasting as much as possible. If we do it, there must be a very good reason, and that reason must be explained in the comments. In this case, typecasting is not acceptable. You need to check the null case. Please check my comment in `FindAllValidMathExpressions`.

src/CodingContract/contracts/HammingCode.ts
```diff
@@ -36,6 +36,9 @@ export const hammingCode: Pick<
       const y = Math.pow(2, getRandomIntInclusive(1, 57));
       return getRandomIntInclusive(Math.min(x, y), Math.max(x, y));
     },
+    getAnswer: (data) => {
+      return HammingEncode(data);
+    },
     solver: (data, answer) => {
       return HammingEncode(data) === answer;
```
catloversg: Same here.

src/CodingContract/contracts/HammingCode.ts
```diff
@@ -83,6 +86,9 @@ export const hammingCode: Pick<
       }
       return _buildArray.join("");
     },
+    getAnswer: (data) => {
+      return HammingDecode(data);
+    },
     solver: (data, answer) => {
       return HammingDecode(data) === answer;
```
catloversg: Same here.

src/CodingContract/contracts/MergeOverlappingIntervals.ts
```diff
@@ -50,6 +50,14 @@ export const mergeOverlappingIntervals: Pick<CodingContractTypes, CodingContract
       }
       result.push([start, end]);
 
+      return result;
+    },
+    solver: (data, answer) => {
+      const result = mergeOverlappingIntervals[CodingContractName.MergeOverlappingIntervals].getAnswer(data) as [
```
catloversg: Same comment about typecasting.

src/CodingContract/contracts/Proper2ColoringOfAGraph.ts
```diff
@@ -75,6 +75,10 @@ export const proper2ColoringOfAGraph: Pick<CodingContractTypes, CodingContractNa
 
       return [n + m, edges];
     },
+    getAnswer: () => {
+      // Not quite sure how we can use the below to generate a canonical "answer"
```
catloversg: You don't need to write this comment. Just remove it.

There is no need to note that the PR's contributor is not sure whether the contract can generate an answer. With this PR, by default, if a contract cannot generate an answer, it will return null. If it can but it returns null, that's either a bug (and we will fix it) or a deliberate choice. If that's a deliberate choice, you need to rephrase this comment and explain the reason that you don't want to generate the answer.

src/CodingContract/contracts/SanitizeParenthesesInExpression.ts
```diff
@@ -94,6 +94,13 @@ export const sanitizeParenthesesInExpression: Pick<
 
       dfs(0, 0, left, right, data, "", res);
 
+      return res;
+    },
+    solver: (data, answer) => {
+      const res = sanitizeParenthesesInExpression[CodingContractName.SanitizeParenthesesInExpression].getAnswer(
```
catloversg: Same comment about typecasting.

src/CodingContract/contracts/ShortestPathInAGrid.ts
```diff
@@ -53,6 +53,10 @@ export const shortestPathInAGrid: Pick<CodingContractTypes, CodingContractName.S
 
       return grid;
     },
+    getAnswer: () => {
+      // Not sure if we can use the below to generate a canonical answer
```
catloversg: Remove this comment.

src/CodingContract/contracts/SpiralizeMatrix.ts
```diff
@@ -107,6 +107,10 @@ export const spiralizeMatrix: Pick<CodingContractTypes, CodingContractName.Spira
         }
       }
 
+      return spiral;
+    },
+    solver: (data, answer) => {
+      const spiral = spiralizeMatrix[CodingContractName.SpiralizeMatrix].getAnswer(data) as number[];
```
catloversg: Same comment about typecasting.

src/NetscriptFunctions/CodingContract.ts
```diff
@@ -55,6 +55,11 @@ export function NetscriptCodingContract(): InternalAPI<ICodingContract> {
       case CodingContractResult.Failure: {
         if (++contract.tries >= contract.getMaxNumTries()) {
           helpers.log(ctx, () => `Coding Contract attempt '${contract.fn}' failed. Contract is now self-destructing`);
+          const solution = contract.getAnswer();
+          if (solution !== null) {
+            helpers.log(ctx, () => "Coding Contract solution was");
```
catloversg: You should also do this with `runContract` in `src\Terminal\Terminal.ts`.

Optional: Consider printing the answer on the same line of `Coding Contract solution was` unless it's too hard to read in too many cases. With a number or an array, printing them on the same line looks good.

d0sboots: +1 especially to showing the answer in the manual case; I was gonna mention that myself. 

I agree with all the other points.

src/CodingContract/contracts/ArrayJumpingGame.ts
TheCleric: Good catch!

catloversg (approved):

PR #2443, opened 2026-01-08
SLEEVES: Add sleeve commands to purchase sleeves and memory via the API

TheAimMan:
Adding a new feature to allow players to purchase sleeves and memory via the API.  Planned to add the following commands under the sleeve namespace

ns.sleeve.purchaseSleeve (done)
ns.sleeve.purchaseMemory
ns.sleeve.getSleeveCost
ns.sleeve.getSleeveMemoryUpgradeCost


src/NetscriptFunctions/Sleeve.ts
```diff
@@ -228,6 +230,28 @@ export function NetscriptSleeve(): InternalAPI<NetscriptSleeve> {
 
       return augs;
     },
+    purchaseSleeve: (ctx) => () => {
+      checkSleeveAPIAccess(ctx);
+      if (!Factions["The Covenant"].isMember) {
```
d0sboots: Use the enum for the covenant.

src/NetscriptFunctions/Sleeve.ts
```diff
@@ -228,6 +230,28 @@ export function NetscriptSleeve(): InternalAPI<NetscriptSleeve> {
...
+        throw helpers.errorMessage(ctx, `You need to be a member of The Covenant to purchase sleeves`);
+      }
+
+      if (Player.sleevesFromCovenant >= MaxSleevesFromCovenant) {
+        throw helpers.errorMessage(ctx, `You have the maximum amount of Sleeves purchasable from The Covenant`);
+      }
+
+      if (Player.canAfford(purchaseCost(Player.sleevesFromCovenant))) {
+        Sleeve.purchaseSleeve();
+        return true;
```
d0sboots: It is pointless to return true here if nothing returns false. I think this API is a good fit for Result.

src/NetscriptFunctions/Sleeve.ts
```diff
@@ -228,6 +230,28 @@ export function NetscriptSleeve(): InternalAPI<NetscriptSleeve> {
...
+      } else {
+        throw helpers.errorMessage(ctx, `Not enough money to purchase Sleeve`);
+      }
+    },
+    getSleeveMemoryUpgradeCost: (ctx) => (_sleeveNumber, _amount) => {
```
d0sboots: I think we can shorten this to getMemoryUpgradeCost, unless there is a name conflict in another namespace.

src/PersonObjects/Sleeve/SleeveCovenantPurchases.tsx
```diff
@@ -5,3 +5,7 @@
 
 export const MaxSleevesFromCovenant = 5;
 export const BaseCostPerSleeve = 10e12;
+export function purchaseCost(n: number): number {
+  const sleevesFromCovenant = n;
```
d0sboots: This should just be the param name.

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -5424,6 +5424,9 @@ export interface Sleeve {
    * @param contract - Name of the contract if applicable.
    * @returns True if the sleeve started the given Bladeburner action, false otherwise.
    */
+
+  purchaseSleeve(): void;
```
d0sboots: Needs docs.

d0sboots:
There's an argument to be made for if purchaseSleeve should require sing. I like the overall plan.

TheAimMan:
I felt being in the sleeve name space made more sense.  If it was put in sing, it would be the only sing call that was gate kept behind a different BN.  These commands will also be rarely used, if ever, since sleeves and memory are permanent. They're more for the person that's going full automation from the get go.  I didn't see a good reason to make them even more difficult to access.

d0sboots:
I agree sleeve namespace is the right place to be. The question is whether purchaseSleeve should (also) require sing. And it's just a question I felt should be asked, I don't have a strong opinion about it.

PR #2447, opened 2026-01-15
CODEBASE: Generate display data for math notation at built time and remove runtime mathjax dependency

### LINKED ISSUE #2437: Package: Remove runtime MathJax dependency
From not looking too deep into either MathML or MathJax, it seems like it should be possible to replace MathJax with MathML. This would reduce the built size slightly, and remove an external dependency. It was discussed back in late June 2025, and the consensus seemed mostly in favor, but nobody opened a PR to act upon it.

> "...If someone found a way to cut the MathJax dependecy, I would be happy to accept that. According to caniuse, it looks like support is over 94%, so there wouldn't be a need for any compat library after all." - d0sboots

> "...There's 10 MathJax equation in bitburner and be it 1mb or 9mb, users with slow internet speed would appreciate..." - hydroflame

> "I would be in favor of removing it. Not because it's a bad package or it's big, but because it's one less dependency to worry about." - Me

catloversg:
Context: #2437.

We have tried to remove the mathjax dependency at runtime for a long time. AFAIK, the first related issue is #313, in which we tried to remove dependencies in the dist/ext folder (mathjax and monaco). Snarling remove monaco in that folder and bundle it with monaco-editor-webpack-plugin. After that, we use #313 to keep track of the removal of mathjax. However, it was closed by #1491, and #1491 did not solve the issue entirely. We only remove the distribution of mathjax from our git repository, but it's still there at build time (copied from node_modules to dist/mathjax). The proper fix for this issue is to generate the output data (chtml, svg, or MathML) at build time and bundle it instead of bundling mathjax and generating the data at runtime. Snarling mentioned this solution at https://github.com/bitburner-official/bitburner-src/issues/313#issuecomment-1443634463.

This issue becomes even more complicated after I added docs for corporation. When we write Tex in md files, we have to use remark-math and rehype-mathjax to render the math notation at runtime. This means we actually have 2 distributions of mathjax. One is distributed in the dist/mathjax folder and used by better-react-mathjax to render notation in non-md-docs cases. One is bundled in vendor.bundle.js and used by rehype-mathjax to render notation in md files. This PR removes both of them. Now we only need mathjax at build time to generate MathML output data. At runtime, we will get that data from main.bundle.js.

There are some important things to consider in this PR.

The first thing is the size of the display data. One of the most important reasons for removing unnecessary dependencies is to reduce the bundle size. However, in order to render math notation written in Tex, we need to convert the notation to a "format" that can be rendered properly by browsers. If we are not careful, the output data can be so large that we actually increase the bundle size after removing mathjax. I tested 3 formats: chtml, svg, or MathML, and I decided to choose MathML. The browser's support is decent now (we don't use any fancy features not supported by MathML) and the output size is fairly small. Both chtml and svg generate huge output. Note that #2437's title says "Remove MathJax and replace it with MathML", but it's not technically correct. MathML is just an output format that can be generated by mathjax and other libraries. The goal is to remove mathjax at runtime. Using MathML (or any format) at build time is the implementation detail.

The second thing is how we generate dynamic math notation at build time. For example:
- With each corporation industry, we showed a different supply chain.
- The home RAM cost is affected by a BN mult.

Generating all possible combinations at build time is doable, but it's too complicated, and I don't think it's worth the effort, so I decide to not do that. For details, please check the implementation.

Before:
before-1

before-2

After:

after-1

after-2

The last thing is the move of rehype-raw and remark-math from `devDependencies` to `dependencies`. To be honest, I don't remember why I put them to `devDependencies` in previous PRs. Maybe I just misunderstood how `dependencies` and `devDependencies` work. We use them at runtime, so they should be in `dependencies` instead of `devDependencies`. I correctly put remark-gfm to `dependencies`, but I mistakenly put the other two to `devDependencies`. I also explicitly add 3 dependencies (hast-util-from-dom, hast-util-to-text, unist-util-visit-parents).

Note that due to the move of rehype-raw and remark-math, some dependencies are not marked as dev ones now (e.g., katex). This is not a problem. They are not dev dependencies now, but it does not mean they are bundled automatically. If we don't use them, webpack won't bundle them. You can verify it with webpack-bundle-analyzer. I also attached two reports generated by webpack-bundle-analyzer.

Before:
[report-before.html](https://github.com/user-attachments/files/24639538/report.html)

After:
[report-after.html](https://github.com/user-attachments/files/24639544/report.html)

Close #2437.


src/Corporation/ui/DivisionOverview.tsx
```diff
@@ -132,8 +133,16 @@ export function DivisionOverview(props: DivisionOverviewProps): React.ReactEleme
           <>
             <Typography>Multiplier for this industry's sales due to its awareness and popularity.</Typography>
             <br />
-            <MathJax>{`\\(\\text{${division.industry} Industry: }\\alpha = ${division.advertisingFactor}\\)`}</MathJax>
-            <MathJax>{`\\(\\text{multiplier} = \\left((\\text{awareness}+1)^{\\alpha} \\times (\\text{popularity}+1)^{\\alpha} \\times \\frac{\\text{popularity}+0.001}{\\text{awareness}}\\right)^{0.85}\\)`}</MathJax>
+            <Typography>
+              {/* \u{1D7AA} is the alpha character */}
```
d0sboots: We have unicode, why not just use the character? It'll certainly be easier to read.

src/Corporation/ui/IndustryProductEquation.tsx
```diff
@@ -10,13 +10,20 @@ interface IProps {
...
+  const prod = props.division.producedMaterials.map((materialName) => `1 ${materialName}`);
   if (props.division.makesProducts) {
-    prod.push("\\textit{Products}");
+    prod.push("Products");
   }
 
-  return <MathJax>{"\\(" + reqs.join("+") + `\\Rightarrow ` + prod.join("+") + "\\)"}</MathJax>;
+  return (
+    <Typography component="span">
+      {/* \u21D2 is the "⟹" character */}
```
d0sboots: Same

src/Documentation/root.ts
```diff
@@ -35,3 +36,14 @@ export const DocumentationPopUpEvents = new EventEmitter<[string | undefined]>()
 export function openDocumentationPopUp(path: string): void {
   DocumentationPopUpEvents.emit(path);
 }
+
+export function convertMathNotation(value: string): string {
+  // MathNotationOutput is imported as json data, so we need to typecast here to access the value via string index
+  // easier without fighting with the TS compiler.
+  const output = (MathNotationOutput as Record<string, string>)[value];
+  if (output == null) {
+    console.error(`Unknown math notation: ${value}`);
```
d0sboots: This feels like it ought to be a true error

src/Faction/ui/DonateOption.tsx
```diff
@@ -77,7 +76,9 @@ export function DonateOption({ faction, favorToDonate, disabled, rerender }: Don
             }}
           />
           <Typography>
-            <MathJax>{`\\(reputation = \\frac{\\text{donation amount} \\cdot \\text{reputation multiplier}}{10^{${digits}}}\\)`}</MathJax>
+            {/* It's fine to use dangerouslySetInnerHTML here. We control the data in both MathNotation.json and
```
d0sboots: You end up repeating this so much, it feels like it belongs in the helper method.

src/Corporation/ui/DivisionOverview.tsx
catloversg: I used it at first, but my IDE warned: `The character U+1d7aa "𝞪" could be confused with the ASCII character U+0061 "a", which is more common in source code.`, so I replaced it with `\u{}`. Nonetheless, I reverted it.

PR #2453, opened 2026-01-24
UI: Add settings to change currency symbol and whether to show it after money value

kaoticengineering:
This PR adds options in Numeric Display for a user to specify a money/currency symbol and position (before or after the value) for use with formatMoney(). There are also a few modifications where a money value and symbol are in a string that have been modified to use formatMoney() for consistency. 

Notes: When the Symbol Text Field is empty, it will show a placeholder value, "$", and the Symbol will default to "$".

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -179,6 +179,7 @@
       "integrity": "sha512-UlLAnTPrFdNGoFtbSXwcGFQBtQZJCNjaN6hQNP3UPvuNXT1i82N26KL3dZeIpNalWywr9IuQuncaAfUaS1g6sQ==",
       "dev": true,
       "license": "MIT",
+      "peer": true,
       "dependencies": {
         "@ampproject/remapping": "^2.2.0",
         "@babel/code-frame": "^7.27.1",
@@ -2071,6 +2072,7 @@
         }
       ],
       "license": "MIT",
+      "peer": true,
       "engines": {
         "node": ">=18"
       },
@@ -2094,6 +2096,7 @@
         }
       ],
       "license": "MIT",
+      "peer": true,
       "engines": {
         "node": ">=18"
       }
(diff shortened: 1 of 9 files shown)


kaoticengineering:
image
image
image

catloversg:
After discussing on Discord, I suggest letting the player specify a string instead of picking from a predefined list.

With places using `$`, I think you can simply remove `$`. For example, `Your funds have been set to $150b for the inconvenience` can be changed to `Your funds have been set to 150b for the inconvenience`. @d0sboots What do you think about removing `$` in text like that?

You also need to revert all changes in `package-lock.json`. `package-lock.json` should never be touched unless you want to change the dependencies. If it's changed after running `npm i`, it's very likely you are using an outdated npm version.

kaoticengineering:
Made Currency Symbol Selection a Text Field. 
Added the option to put the symbol before or after the value. 
Removed $ from Corporations.ts [..] $150b[..] msg instead of using Settings.CurrencySymbol. 
Reverted my copy of package-lock.json.

image

kaoticengineering:
Updated screenshots

image
image

catloversg:
Update:
- In the setting page, use `Settings.CurrencySymbol` as the initial state.
- Add a missing space character in `SellMaterialModal.tsx`.
- Remove the placeholder in `ThrowPartyModal.tsx`. With the current UX, this placeholder is never shown.
- Tweak text in `Literatures.tsx` and `BitNode.tsx` and remove the usage of `formatMoney`. `BitNodes` and `Literatures` are initialized before loading the save data, so they always use the default setting. We can refactor our code to dynamically load the relevant text or reinitialize when needed, but tweaking the text is much easier.

catloversg (approved):

PR #2461, opened 2026-02-01
DOCUMENTATION: Move current changelog to changelog-v2.md

catloversg:
The current file will be used for v3.


src/Documentation/doc/en/changelog.md
```diff
@@ -1,1595 +0,0 @@
-# Changelog
```
d0sboots, line 1: Does this mean the file currently exists, but is empty?

Maybe a small amount of placeholder text would be useful (that we will remove as part of the 3.0 push)?

src/Documentation/doc/en/changelog.md
```diff
@@ -0,0 +1,296 @@
+# Changelog
+
+## v3.0.0 development version: last updated 1 February 2026
```
d0sboots: I don't think this is right, either. IIRC, we move the contents of the dev changelog from Constants to here as part of doing a release. Usually Snarling handles that as part of his release process. If we duplicate the v3 dev changes, it will have weird effects. (At minimum, these will become skewed and/or we'll have to update both up until release time.)

PR #2465, opened 2026-02-03
DARKNET: Add extra hint to sorted echo puzzle at high levels

ficocelliguy:
The darknet sorted echo vulnerability puzzle currently just asks players to try all permutations. This PR adds an extra hint to add more opportunities to optimize the higher-level versions of the puzzle, to avoid the higher-difficulty variants from feeling too brute-force. Specifically, it provides the root mean square deviation of the attempted password vs the correct one when each digit is considered individually.

### DIFF
--- a/src/DarkNet/effects/authentication.ts
+++ b/src/DarkNet/effects/authentication.ts
@@ -124,6 +124,23 @@ export const checkPassword = (
         `${altitude}`,
       );
     }
+    case ModelIds.SortedEchoVuln: {
+      if (server.password.length < 5 || attemptedPassword.length !== server.password.length) {
+        return getFailureResponse(attemptedPassword, server.staticPasswordHint, server.passwordHintData);
+      }
+      let squaredError = 0;
+      for (let i = 0; i < attemptedPassword.length; i++) {
+        const attempted = Number(attemptedPassword[i]);
+        const actual = Number(server.password[i]);
+        if (isNaN(attempted) || isNaN(actual)) {
+          return getFailureResponse(attemptedPassword, server.staticPasswordHint, server.passwordHintData);
+        }
+        squaredError += (attempted - actual) ** 2;
+      }
+      const rmsd = Math.sqrt(squaredError / attemptedPassword.length);
+      const rmsdMessage = `${server.passwordHintData};RMS Deviation:${rmsd.toFixed(3)}`;
+      return getFailureResponse(attemptedPassword, server.staticPasswordHint, rmsdMessage);
+    }
     default:
       return getFailureResponse(attemptedPassword, server.staticPasswordHint, server.passwordHintData);
   }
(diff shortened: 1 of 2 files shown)


src/DarkNet/effects/authentication.ts
```diff
@@ -124,6 +124,23 @@ export const checkPassword = (
...
     }
+    case ModelIds.SortedEchoVuln: {
+      if (server.password.length < 5 || attemptedPassword.length !== server.password.length) {
+        return getFailureResponse(attemptedPassword, server.staticPasswordHint, server.passwordHintData);
+      }
+      let squaredError = 0;
+      for (let i = 0; i < attemptedPassword.length; i++) {
+        const attempted = Number(attemptedPassword[i]);
+        const actual = Number(server.password[i]);
+        if (isNaN(attempted) || isNaN(actual)) {
```
d0sboots: Two things:
1. In general we prefer !isFinite, since it catches more. I don't think it makes a difference here.
2. I don't think it makes sense to check `actual`. If it's NaN we have bigger problems, but it would actually be better to see this by propagating the NaN to the output.

src/DarkNet/effects/authentication.ts
```diff
@@ -124,6 +124,23 @@ export const checkPassword = (
...
+          return getFailureResponse(attemptedPassword, server.staticPasswordHint, server.passwordHintData);
+        }
+        squaredError += (attempted - actual) ** 2;
+      }
+      const rmsd = Math.sqrt(squaredError / attemptedPassword.length);
+      const rmsdMessage = `${server.passwordHintData};RMS Deviation:${rmsd.toFixed(3)}`;
```
d0sboots:
```suggestion
      const rmsdMessage = `${server.passwordHintData}; RMS Deviation:${rmsd.toFixed(3)}`;
```

d0sboots (commented):
I like it! The trade-off for a better solving process is a sudden wrench in your parsing at higher difficulties. XD

src/DarkNet/effects/authentication.ts
ficocelliguy: updated!

catloversg:
A small nitpick (especially after you already changed from `isNaN` to `isFinite` as d0sboots wants): Currently, the common way of checking "does this value make sense as a number?" in our codebase is to use `Number.isFinite`. Our codebase used `isNaN` everywhere, and that's a bad choice, IMO.

`isFoo`-type functions in the global scope always coerce the value, and coercion tends to bring surprising results (bugs, to be honest). For example, `isFinite("")` is true. The only thing that keeps us relatively safe is the fact that Bitburner uses TypeScript, which catches this kind of bug fairly well.

d0sboots:
Oh yeah, I was mentally thinking of the `Number.isFinite` version, but failed to make that clear in my response. Good catch.

PR #2466, opened 2026-02-03
DARKNET: Tune 2G_cellular timing attack puzzle

ficocelliguy:
This PR makes the bonus runtime speed from threads during authentication also apply to the extra time added per correct character, to make sure the puzzle isn't too much of an outlier compared to others at high difficulty


src/DarkNet/effects/effects.ts
d0sboots, line 80: Note: I was going to comment about possibly changing this formula, but on further reflection I think it's good how it is.

In particular, I think it's important that auth performance scales linearly with number of threads. With this formula, that is the case, except that using 1 thread and multiple processes is ~5x more efficient than using multiple threads. I'm OK with that much of a gap existing, as long as it stays contant.

PR #2468, opened 2026-02-04
DARKNET: Prevent blocked ram on the "darkweb" server

ficocelliguy:
As discussed on discord.
in order to make the transition into darknet as easy as possible for new players, this prevents the "darkweb" server from having any ram blocked to start.

### DIFF
--- a/src/DarkNet/models/DarknetServerOptions.ts
+++ b/src/DarkNet/models/DarknetServerOptions.ts
@@ -54,9 +54,13 @@ export type DarknetServerOptions = {
   leftOffset: number;
 };
 
-export const DnetServerBuilder = (options: DarknetServerOptions, name = generateDarknetServerName()): DarknetServer => {
+export const DnetServerBuilder = (
+  options: DarknetServerOptions,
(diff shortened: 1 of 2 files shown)


src/DarkNet/models/DarknetServerOptions.ts
```diff
@@ -54,9 +54,13 @@ export type DarknetServerOptions = {
   leftOffset: number;
 };
 
-export const DnetServerBuilder = (options: DarknetServerOptions, name = generateDarknetServerName()): DarknetServer => {
+export const DnetServerBuilder = (
+  options: DarknetServerOptions,
+  name = generateDarknetServerName(),
+  allowBlockedRam = true,
```
d0sboots: I was gonna let you get away with this, but... you already *have* an options object. Why are name (and now this) separate arguments outside it?

ficocelliguy: It was to make it more clear what the defaults were from the type signature, but it looks kind of silly with multiple like that, yeah. Updated.

PR #2469, opened 2026-02-04
DARKNET: Add glossary of terms to the documentation page

ficocelliguy:
Adds some terms and definitions to the documentation to clarify some potentially confusing concepts.


src/Documentation/doc/en/programming/darknet.md
```diff
@@ -33,6 +33,16 @@ In some cases, the only way to get to deeper parts of the net is to hitch a ride
 - `ns.dnet.induceServerMigration()` can be used to target a connected server and, when used enough, will force it to move to a new location on the darknet.
 - `ns.dnet.promoteStock()` increases the volatility of the targeted stock via propaganda, which can increase the potential profits from trading it.
 
+### Glossary of Terms
+
+**Authenticated** - When a script calls dnet.authenticate with the correct password, two things happen. Firstly, you are given admin rights to the server (the same as nuking). Second, the script that called authenticate gets a session with the target server. A session on the target server is required for some dnet API methods.
+
+**Session** - A script needs a session on a darknet server in order to scp or exec targeting that server, or to use most of the darknet api targeting that server. Sessions are per-script. A script gets a session if it calls authenticate with the correct password. If another script already has done that, scripts can also get a session by calling dnet.connectToSession with the correct password.
```
catloversg: - scp and exec are the only APIs requiring a session. `memoryReallocation` previously requires a session as well, but we removed it (https://discord.com/channels/415207508303544321/1358930422607642845/1464046618918518794).
- I think we can rephrase and clarify some other things.

For example:
```
**Session** - A script needs a session on a darknet server in order to scp or exec targeting that server. A script gets a session after it calls `dnet.authenticate` with the correct password. Scripts can also get a session by calling `dnet.connectToSession` (a sync API with low RAM cost) after authenticating successfully at least once.

Note that sessions are per-script (each script is identified by a unique pid). For example, if you successfully authenticate server "foo" with pid 1, pid 1 have a session with "foo", but pid 2 does not have a session. After pid 2 calls `dnet.connectToSession`, pid 2 has a session, but pid 3 does not have a session, and so on.
```

ficocelliguy: I pushed up some changes to the description of session. I decided to drop the reference to PIDs, as the list of pids that have a session with a server isn't visible to the user, it's just an implementation detail under the hood.

src/Documentation/doc/en/programming/darknet.md
```diff
@@ -33,6 +33,16 @@ In some cases, the only way to get to deeper parts of the net is to hitch a ride
 - `ns.dnet.induceServerMigration()` can be used to target a connected server and, when used enough, will force it to move to a new location on the darknet.
 - `ns.dnet.promoteStock()` increases the volatility of the targeted stock via propaganda, which can increase the potential profits from trading it.
 
+### Glossary of Terms
+
+**Authenticated** - When a script calls dnet.authenticate with the correct password, two things happen. Firstly, you are given admin rights to the server (the same as nuking). Second, the script that called authenticate gets a session with the target server. A session on the target server is required for some dnet API methods.
+
+**Session** - **Session** - A script needs a session on a darknet server in order to scp or exec targeting that server. A script gets a session after it calls `dnet.authenticate` with the correct password. Scripts can also get a session by calling `dnet.connectToSession` (a sync API with low RAM cost) after any script authenticates successfully on that server at least once. Note that sessions are per-script - each script needs to individually get a session with a specific darknet server in order to exec or scp to that server.
```
d0sboots:
```suggestion
**Session** - A script needs a session on a darknet server in order to scp or exec targeting that server. A script gets a session after it calls `dnet.authenticate` with the correct password. Scripts can also get a session by calling `dnet.connectToSession` (a sync API with low RAM cost) after any script authenticates successfully on that server at least once. Note that sessions are per-pid (running instance of a script) - each script needs to individually get a session with a specific darknet server in order to exec or scp to that server.
```
I also think it's important to mention pids. It's not just an implementation detail, it's the most concise way we have of describing the concept of "running instance of a script," distinguishing from "the script itself" (which might encompass *all* copies).

src/Documentation/doc/en/programming/darknet.md
```diff
@@ -33,6 +33,16 @@ In some cases, the only way to get to deeper parts of the net is to hitch a ride
...
+
+**Connected** - Each server on the network has specific other servers it is linked to. These connected servers are the servers that appear when using "scan" in the terminal, or that can be seen by calling dnet.probe. This kind of direct connection is required for most dnet API methods. ns.exec requires either a direct connection, or a backdoor, or a stasis link to target a darknet server.
```
d0sboots:
```suggestion
**Connected** - Each server on the network has specific other servers it is linked to. These are the links seen in the UI, the servers that appear when using "scan" in the terminal, or that can be seen by calling dnet.probe. This kind of direct connection is required for most dnet API methods. ns.exec requires either a direct connection, or a backdoor, or a stasis link (which also sets a backdoor) to target a darknet server.
```

src/Documentation/doc/en/programming/darknet.md
```diff
@@ -33,6 +33,16 @@ In some cases, the only way to get to deeper parts of the net is to hitch a ride
...
+
+**Offline** - Sometimes darknet servers will go offline. Effectively, the server is deleted, and any running scripts that were on it are killed. Eventually, the server may come back online, but it will have a different password, and will be fully cleaned and cleared of scripts.
```
d0sboots:
```suggestion
**Offline** - Sometimes darknet servers will go offline. Effectively, the server is deleted, and any running scripts that were on it are killed. Eventually, a server with the same name may come back online, but it will have a different password, and will be fully cleaned and cleared of scripts.
```
Slight reword to emphasize that this is effectively a new server.

d0sboots (commented):
Great, this is just what I was hoping for!

PR #2470, opened 2026-02-04
GITHUB: Update format created by "build artifacts" workflow

Snarling:
The original format was not conducive to easily uploading to steam for version updates - previous format was a zip of built artifacts, which contained one or more zips of platform builds, and those platform build zips contained a directory which contained the game content.

Steam expects the uploaded zip to directly contain the game content, so I updated the workflow to produce this. Steam didn't need the arm builds, but these have been maintained for manual downloads. See https://github.com/bitburner-official/bitburner-src/actions/runs/21675942464 for sample run.

### DIFF
--- a/.github/workflows/build-artifacts.yml
+++ b/.github/workflows/build-artifacts.yml
@@ -28,7 +28,7 @@ jobs:
         run: npm run electron-win
       - name: Zip
         shell: bash
-        run: cd .build; for i in bitburner-*; do 7z a "$i.zip" "$i"; done; cd ../;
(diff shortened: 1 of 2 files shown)


catloversg:
I updated `build-upload-artifacts.yml` to be compatible with the new changes.
Test run: https://github.com/catloversg/bitburner-src/actions/runs/21683748749/job/62526696174
Test release: https://github.com/catloversg/bitburner-src/releases/tag/v3.0.0-test2

PR #2475, opened 2026-02-05
UI: Show karma in Stats page and explain what it is in documentation

catloversg:
A long time ago, karma was (kind of) a "mysterious" feature of Bitburner. There was no page displaying it in the UI. The only API to check its value was an (intentionally?) hidden one. It had not been even mentioned anywhere in the doc until we added `faction_list.md` (only says how much karma is required for each faction) and `bitnodes.md` (only says gang needs karma without any further explanation).

I was not here back then, so I don't know why karma was "mysterious" like that. However, nowadays, karma is much more "normal". We have `faction_list.md` that shows the requirements, a documented API (`ns.getPlayer().karma`), and BN recommendation guides that mention karma. I think it's time to show it in the UI and explain what it is in documentation.

Capture

### DIFF
--- a/src/Documentation/doc/en/advanced/gang.md
+++ b/src/Documentation/doc/en/advanced/gang.md
@@ -6,7 +6,9 @@ Seen by most of the population as nihilistic, murderous and vile, occasional rum
 
 ## Starting and Recruiting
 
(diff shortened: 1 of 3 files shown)


d0sboots:
I agree with this view, but I think I'd like an #endgame-general poll just to vibe-check against unexpectedly negative sentiment.

Edit: Or maybe development, since #endgame-general scrolls too fast.

d0sboots:
The poll is so lopsided at this point that I'm calling it. (I was for it anyway, just wanted to ensure no giant negative reaction.)

PR #2476, opened 2026-02-05
REFACTOR: Consolidate checks under getFailureResult()

d0sboots:
There was duplicated code, and more importantly, were were handling certain things subtly differently in exec() and scp() as a result. This notably causes a behavior change in exec() and scp() where failure to authenticate now returns failure instead of throwing, which I believe is the proper response.

This also makes it easier to see in the code exactly which functions require what (auth, session, etc.)

# Testing

- Updated failing unittests
- Manually checked `exec` and `scp`.


ficocelliguy (approved):

catloversg (approved):
I fixed a typo and updated the comments in the Jest tests.

PR #2479, opened 2026-02-06
DARKNET: Add improved challenge achievement for BN15

ficocelliguy:
Per player feedback, I wanted the BN15+ achievement to be similar in theme to the other challenge BN cheevies. However, "don't use darknet" is not an interesting achievement. Instead, this promotes clever use of data files and packet capture to solve the harder puzzles, and leaning more on simple servers that don't need heartbleed. In addition, induceServerMigration can be used to reach deeper levels without solving all the servers along the way.

Alternatively, players can beat the node with bladeburner instead to complete the challenge. 

This PR also gives a new name and icon to the existing achievement for getting the deepest lab's augment.


src/Achievements/AchievementData.json
```diff
@@ -474,7 +479,7 @@
     "CHALLENGE_BN15": {
```
d0sboots, line 479: Reusing this ID means playtesters get both "for free" if they already got the previous one. I think this is fine (the right way to handle it even), but I wanted to note it.

src/NetscriptFunctions/Darknet.ts
```diff
@@ -263,6 +263,8 @@ export function NetscriptDarknet(): InternalAPI<DarknetAPI> {
         return helpers.netscriptDelay(ctx, networkDelay).then(() => {
           const xpGained = Player.mults.charisma_exp * 50 * ((500 + Player.skills.charisma) / 500);
           Player.gainCharismaExp(xpGained);
+          DarknetState.hasUsedHeartbleed = true;
```
d0sboots: I would put this a bit higher. I was thinking 262, or maybe even 250. Something that more directly meets the definition of "never *called* dnet.heartbleed".

In particular, there's potential weird timing edge-cases if you put it in the async part.

d0sboots:
Oh yeah, forgot to say: I think this is a good compromise, and I'm glad there was an easy state to stash this in. XD

catloversg:
I fixed the wrong achievement data and updated the svg file. Some notes:
- You need to run `pack-for-web.sh` after adding the new svg file.
- Although the steps in `Design new icons` are not mandatory, you should still follow them. They help you catch potential visual issues with the new svg file.

For example, this is how `darknet-depths.svg` looks when opening in a browser:
browser

Looks decent, right?

This is how it looks when you open it in Inkscape:
inkscape

This is how it actually looks in the game:
before
Much more chunky than expected!

This is what it looks like now:
after

I'll clarify the instructions in `src\Achievements\README.md` and update `darknet-backdoor.svg` (if needed) in another PR later.

src/Achievements/AchievementData.json
ficocelliguy: yeah, I'm aware. I wanted to keep it for consistency. Playtesters have been driving using the dev tools all this time anyway, so I generally expect them to pick up their real save again at some point if they want to go for this

PR #2480, opened 2026-02-06
UI: Update icon of DARKNET_BACKDOOR and DARKNET_DEPTHS achievements

catloversg:
Context: https://github.com/bitburner-official/bitburner-src/pull/2479#issuecomment-3860905766

Before:

before

After:

after


d0sboots:
Good catch.

While we're tweaking this, can we make it bigger? Use your best judgement, but right now this is mostly whitespace.

catloversg:
New bigger icons:
Capture

PR #2482, opened 2026-02-07
DNET: Add JS object properties as server names; refactor save/load/server storage to support this

ficocelliguy:
To fit the theme of real-world cracking / vulnerabilites, the server names "__proto__" and "constructor" have been added to the list of preset names, at @d0sboots ' suggestion. This required changing AllServers to a map (which it probably should have been already) to avoid issues reading and writing to those names as object properties.

### DIFF
--- a/src/DarkNet/controllers/NetworkMovement.ts
+++ b/src/DarkNet/controllers/NetworkMovement.ts
@@ -203,6 +203,7 @@ export const addLowLevelServersIfNeeded = (): void => {
   const serversConnectedToDarkweb = getAllDarknetServers().filter((s) => s.depth === 0);
   if (serversConnectedToDarkweb.length <= 3) {
     addRandomDarknetServers(2, 0, true);
+    addLowLevelServersIfNeeded();
   }
   if (lowLevelServers.length / (4 * NET_WIDTH) < LOW_LEVEL_SERVER_DENSITY) {
     addRandomDarknetServers(2, Math.floor(Math.random() * 4));
--- a/src/DarkNet/controllers/NetworkMovement.ts
+++ b/src/DarkNet/controllers/NetworkMovement.ts
@@ -203,7 +203,6 @@ export const addLowLevelServersIfNeeded = (): void => {
   const serversConnectedToDarkweb = getAllDarknetServers().filter((s) => s.depth === 0);
   if (serversConnectedToDarkweb.length <= 3) {
     addRandomDarknetServers(2, 0, true);
-    addLowLevelServersIfNeeded();
   }
   if (lowLevelServers.length / (4 * NET_WIDTH) < LOW_LEVEL_SERVER_DENSITY) {
     addRandomDarknetServers(2, Math.floor(Math.random() * 4));
--- a/src/DarkNet/models/DarknetState.ts
+++ b/src/DarkNet/models/DarknetState.ts
@@ -70,7 +70,7 @@ export const getServerState = (hostname: string): ServerState => {
     authenticatedPIDs: [],
   };
   DarknetState.serverState.set(hostname, newState);
-  return DarknetState.serverState.get(hostname) ?? newState;
+  return newState;
 };
 
 /**
--- a/src/DarkNet/models/DarknetState.ts
+++ b/src/DarkNet/models/DarknetState.ts
@@ -43,12 +43,12 @@ export const DarknetState = {
   lastStormTime: new Date(),
 
   stockPromotions: {} as Record<string, number>,
-  migrationInductionServers: {} as Record<string, number>,
+  migrationInductionServers: new Map<string, number>(),
 
   /**
    * Do NOT access the server state directly via this property. You must call getServerState.
    */
-  serverState: {} as Record<string, ServerState>,
+  serverState: new Map<string, ServerState>(),
   offlineServers: [] as string[],
   showFullNetwork: false,
   zoomIndex: 7,
@@ -60,14 +60,17 @@ export const DarknetState = {
  * Get the server state. It will initialize the state if it does not exist in DarknetState.serverState.
  */
 export const getServerState = (hostname: string): ServerState => {
-  if (!DarknetState.serverState[hostname]) {
-    DarknetState.serverState[hostname] = {
-      serverLogs: [],
-      lastLogTime: undefined,
-      authenticatedPIDs: [],
-    };
+  const currentState = DarknetState.serverState.get(hostname);
+  if (currentState) {
+    return currentState;
   }
-  return DarknetState.serverState[hostname];
+  const newState = {
+    serverLogs: [],
+    lastLogTime: undefined,
+    authenticatedPIDs: [],
+  };
+  DarknetState.serverState.set(hostname, newState);
+  return DarknetState.serverState.get(hostname) ?? newState;
 };
 
 /**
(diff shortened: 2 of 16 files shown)


src/DarkNet/controllers/NetworkMovement.ts
```diff
@@ -203,6 +203,7 @@ export const addLowLevelServersIfNeeded = (): void => {
   const serversConnectedToDarkweb = getAllDarknetServers().filter((s) => s.depth === 0);
   if (serversConnectedToDarkweb.length <= 3) {
     addRandomDarknetServers(2, 0, true);
+    addLowLevelServersIfNeeded();
```
d0sboots: This is an unrelated change?

src/DarkNet/models/DarknetState.ts
```diff
@@ -60,14 +60,17 @@ export const DarknetState = {
...
+    return currentState;
   }
-  return DarknetState.serverState[hostname];
+  const newState = {
+    serverLogs: [],
+    lastLogTime: undefined,
+    authenticatedPIDs: [],
+  };
+  DarknetState.serverState.set(hostname, newState);
+  return DarknetState.serverState.get(hostname) ?? newState;
```
d0sboots: You can just return newState here.

src/DarkNet/models/packetSniffing.ts
```diff
@@ -107,9 +107,9 @@ const getExactCharactersHint = (lastPassword: string, realPassword: string) => {
 };
 
 export const logPasswordAttempt = (server: DarknetServer, passwordResponse: PasswordResponse, pid: number) => {
+  populateServerLogsWithNoise(server);
```
d0sboots: Unrelated change?

src/Server/AllServers.ts
```diff
@@ -19,28 +19,22 @@ import { MAX_NET_DEPTH, NET_WIDTH } from "../DarkNet/Enums";
...
     return server;
   }
 }
 
 //Get server by IP or hostname. Returns null if invalid
 export function GetServer(s: string): BaseServer | null {
-  if (Object.hasOwn(AllServers, s)) {
-    const server = AllServers[s];
-    if (server) return server;
+  if (AllServers.has(s)) {
```
d0sboots: You can save a lookup here by just doing the get and returning it if it's non nullish

src/Server/AllServers.ts
```diff
@@ -141,21 +135,23 @@ export function AddToAllServers(server: Server | HacknetServer | DarknetServer):
...
+  }
+  AllServers.set(newName, existingServer);
+  AllServers.delete(hostname);
 };
 
 export function prestigeAllServers(): void {
   for (const member of Object.keys(AllServers)) {
-    // eslint-disable-next-line @typescript-eslint/no-dynamic-delete
-    delete AllServers[member];
+    AllServers.delete(member);
```
d0sboots: Just use clear

src/Server/AllServers.ts
```diff
@@ -141,21 +135,23 @@ export function AddToAllServers(server: Server | HacknetServer | DarknetServer):
...
   }
-  AllServers = {};
+  AllServers = new Map();
```
d0sboots: I'm not sure why we did this before, but it seems unneeded now, definitely.

src/Server/AllServers.ts
```diff
@@ -170,16 +166,16 @@ export function loadAllServers(saveString: string): void {
...
   for (const [serverName, server] of Object.entries(allServersData)) {
     if (!(server instanceof Server) && !(server instanceof HacknetServer) && !(server instanceof DarknetServer)) {
-      throw new Error(`Server ${serverName} is not an instance of Server or HacknetServer.`);
+      throw new Error(`Server ${serverName} is not an instance of Server or HacknetServer or DarknetServer.`);
+    } else {
+      AllServers.set(serverName, server);
     }
   }
-  // We validated the data above, so it's safe to typecast here.
-  AllServers = allServersData as typeof AllServers;
```
d0sboots, line 177: You need to create a new map in here, too

ficocelliguy: I believe the new map is created at the top of this file. I added a .clear() to this function to ensure existing servers are not carried over after a load or anything.

test/jest/Script/RunningScript.test.ts
```diff
@@ -0,0 +1,45 @@
...
+    expect(revivedRunningScript).toBeInstanceOf(RunningScript);
+    expect(revivedRunningScript.filename).toEqual(filename);
+    expect(revivedRunningScript.server).toEqual(hostname);
+    expect(revivedRunningScript.args).toEqual(args);
+    expect(revivedRunningScript.ramUsage).toEqual(ramUsage);
+    expect(revivedRunningScript.dataMap.get(hostname)).toEqual([1000, 2, 3, 4]);
+  });
+
+  it("Loads a standard savefile shape", () => {
+    const data = ` {"ctor":"RunningScript","data":{"args":["arg1","arg2"],"dataMap":{"testserver":[1000,2,3,4]},"filename":"test.js","offlineExpGained":0,"offlineMoneyMade":0,"offlineRunningTime":0.01,"onlineExpGained":0,"onlineMoneyMade":0,"onlineRunningTime":0.01,"ramUsage":1.7,"server":"testserver","scriptKey":"test.js*[\\"arg1\\",\\"arg2\\"]","stdin":null,"tailOutputPipeConfig":null,"terminalOutputPipeConfig":null,"title":"test.js arg1 arg2","threads":1,"temporary":false}}`;
```
d0sboots, line 30: Optional: Jest has infrastructure for loading json from a golden file and comparing state like this. We use it in some of our other tests. It's nice to be able to see all the data right here, but the trade-off is the extra escaping. Up to you.

src/Server/AllServers.ts
d0sboots: Right, that was my point, that I don't think we can guarantee two loads won't happen back to back.

d0sboots:
This is looking pretty solid to me, is it still missing stuff from your perspective to be called a draft?

ficocelliguy:
@d0sboots It was missing one unit test (saving / loading AllServers) which I have now added. I've also deployed this for testing to https://ficocelliguy.github.io/bitburner-src/ as it feels like a higher-risk change, since it touches saving and loading.

test/jest/Server/AllServers.test.ts
```diff
@@ -0,0 +1,37 @@
...
+    const server1 = new Server({
+      hostname,
+      ip: "173.78.146.183" as IPAddress,
+    });
+    AddToAllServers(server1);
+    expect(GetAllServers(true)).toEqual([server1]);
+
+    const serializedServers = saveAllServers();
+    expect(serializedServers).toEqual(
+      `{"__proto__":{"ctor":"Server","data":{"contracts":[],"cpuCores":1,"ftpPortOpen":false,"hasAdminRights":false,"hostname":"__proto__","httpPortOpen":false,"ip":"173.78.146.183","isConnectedTo":false,"maxRam":0,"messages":[],"organizationName":"","programs":[],"scripts":{"ctor":"JSONMap","data":[]},"serversOnNetwork":[],"smtpPortOpen":false,"sqlPortOpen":false,"sshPortOpen":false,"textFiles":{"ctor":"JSONMap","data":[]},"purchasedByPlayer":false,"backdoorInstalled":false,"baseDifficulty":1,"hackDifficulty":1,"minDifficulty":1,"moneyAvailable":0,"moneyMax":0,"numOpenPortsRequired":5,"openPortCount":0,"requiredHackingSkill":1,"serverGrowth":1,"runningScripts":[]}}}`,
```
d0sboots, line 25: We actually already have a test that is pretty much just a save/load test, so it might make more sense to augment it by adding this server to it. (That data is also used for some other calculations, so I think it would potentially catch more things?)

ficocelliguy: Added to Save.test.ts.

PR #2485, opened 2026-02-09
DARKNET: Buff packet sniffing slightly

d0sboots:
Refactor the sniff code so that successful sniffs for the current server are tagged server_name:password, not just those for adjacent servers.

Tested manually by looking at a bunch of sniff results.

### DIFF
--- a/src/DarkNet/models/packetSniffing.ts
+++ b/src/DarkNet/models/packetSniffing.ts
@@ -20,23 +20,19 @@ export const capturePackets = (server: DarknetServer) => {
   const vulnerability = server.modelId === ModelIds.packetSniffer ? 8 : 1;
   const passwordInclusionChance = BASE_PASSWORD_INCLUSION_RATE * vulnerability * DIFFICULTY_MODIFIER ** difficulty;
 
+  let captureServer: DarknetServer | null = null;
   if (Math.random() < passwordInclusionChance) {
-    const intro = Math.floor(Math.random() * 124);
-    return `${getRandomData(server, intro)}${server.password}${getRandomData(
-      server,
-      124 - intro - server.password.length,
-    )}`;
-  }
-  if (Math.random() < passwordInclusionChance) {
+    captureServer = server;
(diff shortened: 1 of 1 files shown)


src/DarkNet/models/packetSniffing.ts
```diff
@@ -20,24 +20,20 @@ export const capturePackets = (server: DarknetServer) => {
...
   if (Math.random() < passwordInclusionChance) {
+    captureServer = server;
+  } else if (Math.random() < passwordInclusionChance) {
+    const connectedServerName = server.serversOnNetwork[Math.floor(Math.random() * server.serversOnNetwork.length)];
+    captureServer = getDarknetServer(connectedServerName);
+  }
+  if (captureServer) {
     const intro = Math.floor(Math.random() * 124);
-    return `${getRandomData(server, intro)}${server.password}${getRandomData(
+    return `${getRandomData(server, intro)} ${captureServer.hostname}:${captureServer.password} ${getRandomData(
```
ficocelliguy, line 33: This will add the current hostname in front of the correct password, and not in front of any of the false positives. That changes the puzzle from "look for strings that fit the password type and length, and try authing with those" to "just keep calling packetCapture until you see the current hostname, then you found the password" which is much much faster.

The current rate of the password appearing with packetCapture was balanced with having to auth a couple times per capture.

If we wanted to make the parsing easier (e.g. add a space before and after the current password so you can use .split() ) I'm fine with that, but I don't want to turn it into a completely different puzzle like this.

ficocelliguy (changes requested):

src/DarkNet/models/packetSniffing.ts
ficocelliguy: EDIT: after further discussion, I've come around on this change. We want to keep this simple and not lean on regex. (Note that you will still need to do something to appease the linter XD)

ficocelliguy (approved):

catloversg (approved):
I'm neutral on the idea. The code change looks good to me.

PR #2488, opened 2026-02-10
REFACTOR: Speed up by-ip lookups by introducing a new map entry

d0sboots:
Thankfully, the existing AllServers map is not exported, so we can ensure that all the changes are only local to this file.

This also fixes a bug if renameServer was called with the same name. (Probably calling code checked that case already.)

- smoke-tested game works with `ns.getServer` by ip

### DIFF
--- a/test/jest/Server/AllServers.test.ts
+++ b/test/jest/Server/AllServers.test.ts
@@ -43,7 +43,7 @@ describe("renameServer tests", () => {
     prestigeAllServers();
     expect(GetAllServers(true)).toEqual([]);
 
-    const home = new Server({ hostname: "home", ip: "1.2.3.4" });
+    const home = new Server({ hostname: "home", ip: "1.2.3.4" as IPAddress });
     AddToAllServers(home);
     // Failures of toEqual will report badly, due to a Jest bug involving our use of JSONMap.
     expect(GetAllServers(true)).toEqual([home]);
--- a/test/jest/Server/AllServers.test.ts
+++ b/test/jest/Server/AllServers.test.ts
@@ -4,6 +4,8 @@ import {
   loadAllServers,
   prestigeAllServers,
   saveAllServers,
+  renameServer,
+  GetServer,
 } from "../../../src/Server/AllServers";
 import { Server } from "../../../src/Server/Server";
 import { IPAddress } from "../../../src/Types/strings";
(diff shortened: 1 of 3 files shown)


d0sboots:
Hold off on this for a bit, I've been waffling on two impls and I think I'm going to switch to the other.

d0sboots:
OK I'm happy with this now.

d0sboots:
@catloversg You were just touching stuff in this area, so I'll request your attention

test/jest/Server/AllServers.test.ts
```diff
@@ -35,3 +37,20 @@ describe("AllServers can be saved and loaded", () => {
...
 });
+
+describe("renameServer tests", () => {
+  it("rename to self edge case", () => {
+    prestigeAllServers();
+    expect(GetAllServers(true)).toEqual([]);
+
+    const home = new Server({ hostname: "home", ip: "1.2.3.4" as IPAddress });
+    AddToAllServers(home);
+    // Failures of toEqual will report badly, due to a Jest bug involving our use of JSONMap.
```
catloversg, line 48: Can you explain a bit more about this comment?

catloversg (commented):
Save file for testing the migration issue that I fixed in dee7e883bf41550a6e91d398b0433b9e33dd1e9c: 

[test_save_0.47.2.json](https://github.com/user-attachments/files/25235472/test_save_0.47.2.json)

catloversg:
This looks good to me now. I have a question about your comment in the Jest test, but it does not block this PR.

By the way, there is another issue with running `migrateAllServersData` in the block of `if (ver < "0.56.0")`, but it's not related to the changes in this PR. I added the migration code in #1790, but I found out later that #1790 did not handle an edge case. It's not urgent, so let's deal with it later.

test/jest/Server/AllServers.test.ts
d0sboots: More comment added.

d0sboots:
@catloversg I reverted your migration code and went with a different approach. I think we can handle this entirely in the load function, without need for migration. The test save loads correctly (scan shows servers and ips in their proper places).

catloversg:
That's okay. It was actually how I fixed the migration issue at first, but then I chose to explicitly do sanity checks in a dedicated migration function and marked the behavior change between versions. One upside of doing this in `loadAllServers` is that it's simple and also another way to fix the edge case I mentioned. I prefer doing the migration explicitly in the exact block of `evaluateVersionCompatibility`, but this is fine too.

d0sboots:
I definitely subscribe to the "no code is best code" philosophy. Especially when it comes to migration code, which doesn't have tests and gets exercised almost never.

PR #2489, opened 2026-02-11
DNET: More fixes and feedback

ficocelliguy:
- make darknet background canvas only draw connections for servers that are on the screen
- base the server component key on a more robust set of properties to re-render when details change
- removed noobfactor from auth time; removed underlevelled debuff for very shallow servers
- add clearer docs around server charisma level for authenticate and heartbleed

### DIFF
--- a/src/DarkNet/ui/NetworkDisplayWrapper.tsx
+++ b/src/DarkNet/ui/NetworkDisplayWrapper.tsx
@@ -58,27 +58,6 @@ export function NetworkDisplayWrapper(): React.ReactElement {
     [draggableBackground],
   );
 
-  const isWithinScreen = useCallback(
-    (server: DarknetServer) => {
-      const { left, top } = getPixelPosition(server, true);
-      const background = draggableBackground.current;
-      const buffer = 600;
-      const visibleAreaLeftEdge = (background?.scrollLeft ?? 0) / zoomOptions[zoomIndex];
-      const visibleAreaTopEdge = (background?.scrollTop ?? 0) / zoomOptions[zoomIndex];
-      const visibleAreaRightEdge =
-        visibleAreaLeftEdge + ((background?.clientWidth ?? 0) / zoomOptions[zoomIndex] ** 2 || window.innerWidth);
-      const visibleAreaBottomEdge =
-        visibleAreaTopEdge + ((background?.clientHeight ?? 0) / zoomOptions[zoomIndex] ** 2 || window.innerHeight);
-      return (
-        left >= visibleAreaLeftEdge - buffer &&
-        left <= visibleAreaRightEdge + buffer &&
-        top >= visibleAreaTopEdge - buffer &&
-        top <= visibleAreaBottomEdge + buffer
-      );
-    },
-    [zoomIndex, zoomOptions],
-  );
-
   const updateDisplay = useCallback(() => {
     if (!canvas.current) {
       return;
@@ -87,8 +66,8 @@ export function NetworkDisplayWrapper(): React.ReactElement {
     setNetDisplayDepth(getDeepestServerDepth() + visibilityMargin);
 
     rerender();
-    drawOnCanvas(canvas.current, isWithinScreen);
-  }, [isWithinScreen, rerender]);
+    drawOnCanvas(canvas.current);
+  }, [rerender]);
 
   useEffect(() => {
     const clearSubscription = DarknetEvents.subscribe(() => updateDisplay());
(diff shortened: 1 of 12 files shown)


src/DarkNet/ui/networkCanvas.ts
```diff
@@ -10,9 +10,9 @@ import { SpecialServers } from "../../Server/data/SpecialServers";
 import { getNetDepth, isLabyrinthServer } from "../effects/labyrinth";
 import { NET_WIDTH } from "../Enums";
 import type { DarknetServer } from "../../Server/DarknetServer";
-import { getDarknetServerOrThrow } from "../utils/darknetServerUtils";
+import { getDarknetServer } from "../utils/darknetServerUtils";
 
-export const drawOnCanvas = (canvas: HTMLCanvasElement) => {
+export const drawOnCanvas = (canvas: HTMLCanvasElement, isWithinScreen: (server: DarknetServer) => boolean) => {
```
d0sboots: I feel like this is just adding more ways for things to go wrong. Drawing lines is very fast, so I don't expect this to be a speedup, unless you've measured it as such.

ficocelliguy: It was primarily intended to keep the canvas consistent with the react components. If servers are being hidden, I wanted their connections to be as well. I suspected this was the issue that was causing some servers to be hidden (but their connections not to be) after the tab was asleep for a bit

src/DarkNet/ui/networkCanvas.ts
```diff
@@ -23,17 +23,18 @@ export const drawOnCanvas = (canvas: HTMLCanvasElement) => {
...
       !server ||
-      (!server.hasAdminRights && !server.serversOnNetwork.find((s) => getDarknetServerOrThrow(s).hasAdminRights))
+      (!server.hasAdminRights && !server.serversOnNetwork.find((s) => getDarknetServer(s)?.hasAdminRights))
     ) {
       continue;
     }
 
     // draw a line between each server and its connected servers
     for (const connectedServerName of server.serversOnNetwork) {
-      const connectedServer = getDarknetServerOrThrow(connectedServerName);
```
d0sboots, line 33: Was this changed due to a reported problem? The connected servers should always be valid.

src/DarkNet/ui/NetworkDisplayWrapper.tsx
```diff
@@ -58,36 +58,36 @@ export function NetworkDisplayWrapper(): React.ReactElement {
...
       clearSubscription();
     };
-  }, [rerender, scrollTo]);
-
-  useEffect(() => {
-    DarknetEvents.emit();
-  }, []);
+  }, [updateDisplay, rerender, scrollTo]);
+
+  const getDeepestServerDepth = () => {
```
d0sboots: I think you only use this in one place, and the fact that you pulled it out as a closure is going to cause problems with useCallback.

src/DarkNet/ui/NetworkDisplayWrapper.tsx
```diff
@@ -252,6 +234,10 @@ export function NetworkDisplayWrapper(): React.ReactElement {
     return servers;
   };
 
+  const getServerKey = (server: DarknetServer, x: number, y: number) => {
+    return `${x} ${y} ${server.hostname} ${server.hasAdminRights}`;
```
d0sboots: I think the best thing to use for the key will be just the ip. It's guaranteed unique, almost never reused, and will allow react to properly track nodes that are moved from one location to another. (At which point, a helper method is overkill.)

src/DarkNet/ui/networkCanvas.ts
ficocelliguy: No, this was from something I was testing, but it turned out to not be relevant. Reverted.

src/DarkNet/ui/networkCanvas.ts
```diff
@@ -23,7 +23,7 @@ export const drawOnCanvas = (canvas: HTMLCanvasElement) => {
   for (const server of DarknetState.Network.flat()) {
     if (
       !server ||
-      (!server.hasAdminRights && !server.serversOnNetwork.find((s) => getDarknetServerOrThrow(s).hasAdminRights))
+      (!server.hasAdminRights && !server.serversOnNetwork.find((s) => getDarknetServerOrThrow(s)?.hasAdminRights))
```
d0sboots: Not fully reverted; the question mark is still at these callsites.

PR #2490, opened 2026-02-11
MISC: Change how coding contract rewards are randomized

catloversg:
This PR implements the suggestion at https://discord.com/channels/415207508303544321/415213435974975508/1469758023747637476

Quote:
```
Gonna float this here: https://github.com/bitburner-official/bitburner-src/pull/2481 changed the CCT generator to have more friendly fallbacks in the case where you don't have a job, or you can't join factions.

However, there's still a different noob trap with CCTs, one that's bugged me for a while: How the reward is generated at creation time. So, if you are idle for a while before you join any factions, you won't get any faction CCTs (because you're not eligible for any), and then when you join a faction all the CCTs you have built up are "useless."

I propose changing the generator so only the reward type (job, rep, rep for all factions, or money) is chosen on spawn. Unlike now, this type would be exposed in the filename (currently the filename includes the faction name iff it is a contract for a specific faction). When you complete the contract, the fallback logic would apply to give you a different reward if you aren't eligible for the type you completed.

This way, the timing of when you join factions doesn't matter, all that matters is your state (factions joined, if you have company job) at the time you complete the CCT.
```

Other changes:
- `Money` previously must be the last reward type in `CodingContractRewardType`. This limitation comes from the not-so-good implementation of `getRandomReward` (notice how we calculated and used `rewardTypeUpperBound`). The new implementation removes this limitation. It's also easier to change the probability of each reward type if we want to rebalance (add/remove items in `validRewardTypes`).
- There is a small behavior change in the fallback logic of `gainCodingContractReward`. With `FactionReputationAll`, we filtered out special factions. This is inconsistent with how we determine the fallback on spawn. Now, we filter out factions that do not offer hacking work, which are special factions, Tetrads, and Slum Snakes.
- When receiving the reward, if `reward.type` is invalid, we will throw an error instead of silently granting money. This should not be a problem. The 4 reward types have not been changed since coding contracts were added in c309c0cdc9303a3e2943e0345d386a845a2ce20d.

### DIFF
--- a/src/CodingContract/Contract.ts
+++ b/src/CodingContract/Contract.ts
@@ -1,4 +1,4 @@
-import { FactionName, CodingContractName } from "@enums";
+import { CodingContractName } from "@enums";
 import { CodingContractTypes } from "./ContractTypes";
 
 import { Generic_fromJSON, Generic_toJSON, IReviverValue, constructorsForReviver } from "../utils/JSONReviver";
@@ -13,7 +13,7 @@ export enum CodingContractRewardType {
   FactionReputation,
   FactionReputationAll,
   CompanyReputation,
-  Money, // This must always be the last reward type
+  Money,
 }
 
 // Numeric enum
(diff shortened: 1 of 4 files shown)


d0sboots:
Save format migration: Old save data will have pre-baked factions/jobs as part of old contracts. These will be loaded, and even continue to be saved in future saves. However, nothing looks at these fields, so the extra data is harmless.

PR #2491, opened 2026-02-11
BLADEBURNER: Stop randomizing action difficulty

catloversg:
Each action has a difficulty that affects action time, reward, success chance, etc. This difficulty is based on `baseDifficulty` and a small randomized offset that changes every time the player loads the game. This offset does not have any real benefit, and it confuses the player when they see the action time/success chance changes. This PR removes this offset.

@d0sboots Besides this usage in the action's constructor, we also randomize the rank gain, rank loss, and damage. What do you think about removing them as well?


d0sboots:
I took a look at those other uses when I looked into this one. I don't know BladeBurner very well, but my impression was that they add to the "quirkiness" of the mechanic. Unlike here, those are per-action, so you aren't locked in to a good/bad result per load. So, I'm much happier to keep those.

PR #2492, opened 2026-02-11
API: Expose ProgramName enum

catloversg:
It's just as the title said.

Test code:
```typescript
  ns.ui.openTail();
  ns.singularity.createProgram("foo");
  ns.singularity.createProgram(ns.enums.ProgramName.bruteSsh, false);
  const work = ns.singularity.getCurrentWork();
  if (work?.type === "CREATE_PROGRAM") {
    ns.singularity.stopAction();
    await ns.sleep(1000);
    ns.singularity.createProgram(work.programName, false);
  }
}
```


markdown/bitburner.programnameenumtype.md
```diff
@@ -0,0 +1,28 @@
...
+  sqlInject: "SQLInject.exe";
+  deepScan1: "DeepscanV1.exe";
+  deepScan2: "DeepscanV2.exe";
+  serverProfiler: "ServerProfiler.exe";
+  autoLink: "AutoLink.exe";
+  formulas: "Formulas.exe";
+  bitFlume: "b1t_flum3.exe";
+  flight: "fl1ght.exe";
+  darkscape: "DarkscapeNavigator.exe";
+  stormSeed: "STORM_SEED.exe";
```
d0sboots, line 26: Double-check with @ficocelliguy that this one is OK to expose.

d0sboots (commented):
Double-checking the implications: Existing JS code will be fine, TS code may require type fiddling?

catloversg:
> TS code may require type fiddling?

Yes. This happens every time we expose a new enum and change the param type from string to the new enum.

PR #2495, opened 2026-02-11
BUG: Fix missed cases in offline server handling

d0sboots:
There were two large holes in the existing offline server handling:

1. It didn't include IPs, so scripts that used IPs instead of hostnames
   would get exceptions thrown for "server not found."
2. Coverage was very low for non-Darknet APIs. Maybe most of them don't
   need to be covered, but many obvious ones like "ps", "killall" and
   "hasRootAccess" were missing. IMO the only reliable answer is one
   that enforces *all* are covered via the type system.

To accomplish the second part, helpers.getServer() was changed to return
null when a server is offline. This intentionally breaks a lot of its
utility, which was to return a server unconditionally. To compensate,
its utility was increased - it now also does unknown argument
processing, allowing it to subsume a common line that all callers were
repeating.

### An important side-effect of this refactor is that the interface for API calls which take a host has become more uniform. It was largely arbitrary whether they allowed `undefined`/leaving off the host before, but now they *all* do.

Some callers switched to ctx.workerScript.getServer(), because they
didn't actually need to be using helpers.getServer(). Similarly, a few
callsites switched to GetServerOrThrow(), for the cases where it should
be guaranteed that the server is valid. The rest are returning a
default/failure response when the server is offline. (Except for
contracts, which threw on failure already anyway.)

# Testing
Besides enhancing the unit-tests, some manual smoke-testing


d0sboots:
@catloversg @ficocelliguy This is a pretty significant refactor, please take a careful look. In particular I'm interested if there are alternate/better designs you think of. (This is the best I came up with, but I'm not convinced it is the best out there.)

src/DarkNet/models/DarknetServerOptions.ts
```diff
@@ -86,23 +87,43 @@ export const DnetServerBuilder = (options: DarknetServerOptions): DarknetServer
...
+    // work with, and this adds a bit of friction to compensate.
+
+    // Use an iterator to directly skip to the appropriate position. Sets don't
+    // have a way to index by offset, and converting to an array would be wasteful.
+    const offset = Math.floor(Math.random() * DarknetState.offlineServers.size);
+    const it = DarknetState.offlineServers.values();
+    for (let i = 0; i < offset; ++i) {
+      it.next();
+    }
+    // The set contains both IPs and hostnames. If we hit an IP, keep going
```
ficocelliguy, line 110: There seems to be a lot of trouble caused by mingling the offline IPs and hostnames. The overhead of storing them in two sets can't be that large, right? I see the comparison to AllServers but how we use the offline server set(s) is a bit different

d0sboots: It's not about the memory overhead; that would be basically the same. It's about optimizing the common path. Checking the offline servers is something that can potentially happen once per server lookup; looking for a name to reuse only happens (rarely) during the mutation. 

Note that isIPAddress is very likely more expensive than checking a set, so if we split this the best way to check would be to just check both sets... Which immediately suggests combining them.

Edit: Also, although there's a lengthy comment to explain the reasoning here, the actual amount of code (beyond what would be needed in the split set case) is only 2-3 lines.

ficocelliguy: Besides "not needing a comment block to explain it", the other benefit to multiple sets is that it lets us (more easily) pull an IP from an offline server when creating a new darknet server, on occasion, if we wanted to add that feature.

src/DarkNet/models/DarknetServerOptions.ts
```diff
@@ -86,23 +87,43 @@ export const DnetServerBuilder = (options: DarknetServerOptions): DarknetServer
...
   return server;
 };
 
 export const generateDarknetServerName = (): string => {
-  if (Math.random() < 0.03 && DarknetState.offlineServers.length > 0 && hasFullDarknetAccess()) {
-    return DarknetState.offlineServers[Math.floor(Math.random() * DarknetState.offlineServers.length)];
+  if (Math.random() < 0.03 && DarknetState.offlineServers.size > 0 && hasFullDarknetAccess()) {
+    // Reuse a hostname that went offline. Note that we're only reusing hostnames,
+    // not IPs. This asymmetry is intentional - people find hostnames easier to
+    // work with, and this adds a bit of friction to compensate.
```
catloversg, line 101: > and this adds a bit of friction to compensate.

I wonder what this means. Can you clarify it a bit?

src/DarkNet/models/DarknetState.ts
```diff
@@ -53,7 +53,23 @@ export const DarknetState = {
...
    */
   serverState: new Map<string, ServerState>(),
-  offlineServers: [] as string[],
+  /**
+   * Much like AllServers, this contains both hostnames and IP addresses in a single set.
+   *
+   * A note about memory use: On Chrome 144.0.7559.133, memory tests indicate that a
+   * dense array takes ~4 bytes/entry. A Set takes ~10.24 bytes/entry. And the
+   * strings we are storing inside take 14 + length bytes each. Random IPs
+   * average (10+90*2+156*3)*4/256+3 = 13.28 chars, so 27.28 bytes. The
```
catloversg, line 62: I'll trust you on this kind of testing.

Optional: If you don't mind, can you post your test code or steps to test these things? I'm curious about it.

src/Netscript/NetscriptHelpers.tsx
```diff
@@ -501,26 +502,34 @@ function scriptIdentifier(
...
+ * @param {unknown} _host - Hostname or ip of the server, defaults to current server
+ * @returns {[BaseServer | null, string]} A pair containing the specified server as a BaseServer, or
+ *    null if the server is offline. The second part is the resolved hostname/ip.
  */
-export function getServer(ctx: NetscriptContext, hostname: string): BaseServer {
-  const server = GetServer(hostname);
-  if (server == null || (server.serversOnNetwork.length == 0 && !(server instanceof DarknetServer))) {
-    const str = hostname === "" ? "'' (empty string)" : "'" + hostname + "'";
-    throw errorMessage(ctx, `Invalid hostname: ${str}`);
+export function getServer(ctx: NetscriptContext, _host?: unknown): [BaseServer | null, string] {
```
catloversg: Why do you use `_host?: unknown` instead of `_host: unknown`?

src/Netscript/NetscriptHelpers.tsx
```diff
@@ -501,26 +502,34 @@ function scriptIdentifier(
...
+  const str = host === "" ? "'' (empty string)" : "'" + host + "'";
+  throw errorMessage(ctx, `Invalid hostname: ${str}`);
 }
 
 /**
  * A "normal server" is an instance of the Server class in src/Server/Server.ts.
  */
-function getNormalServer(ctx: NetscriptContext, host: string): Server {
-  const server = getServer(ctx, host);
+function getNormalServer(ctx: NetscriptContext, _host?: unknown): Server {
```
catloversg: Same question about `_host?: unknown`.

src/NetscriptFunctions.ts
```diff
@@ -664,7 +647,7 @@ export const ns: InternalAPI<NSFull> = {
       ) {
         return 0;
       }
-      const server = helpers.getServer(ctx, host);
+      const server = GetServerOrThrow(host);
       return runScriptFromScript("exec", server, path, args, ctx.workerScript, runOpts);
```
catloversg: You can use the `server` property from the result of `checkDarknetServer` like this:
```typescript
      const checkResult = checkDarknetServer(ctx, host, {
        allowNonDarknet: true,
        requireAdminRights: true,
        requireSession: true,
        requireDirectConnection: true,
        backdoorBypasses: true,
      });
      if (!checkResult.success) {
        return 0;
      }
      return runScriptFromScript("exec", checkResult.server, path, args, ctx.workerScript, runOpts);
```

src/NetscriptFunctions.ts
```diff
@@ -792,11 +778,11 @@ export const ns: InternalAPI<NSFull> = {
     ) {
       return false;
     }
-    if (!checkDarknetServer(ctx, source, { allowNonDarknet: true }).success) {
+    const [sourceServer] = helpers.getServer(ctx, source);
+    const [destServer] = helpers.getServer(ctx, destination);
```
catloversg: Same suggestion about `checkDarknetServer`:
```typescript
    const checkResult = checkDarknetServer(ctx, destination, {
      allowNonDarknet: true,
      requireAdminRights: true,
      requireSession: true,
    });
    if (!checkResult.success) {
      return false;
    }
    const [sourceServer] = helpers.getServer(ctx, source);
    if (!sourceServer) {
      return false;
    }
    const destServer = checkResult.server;
```

catloversg (commented):
There are two things I want to take note of:
- In our codebase (not just your changes here), we sometimes write `foo: (ctx) => (_bar?) => {`. The type of `_bar` is unknown, so `(_bar?)` is the same as `(_bar)`. I guess that's how we "marked" a param as an optional one, but that's pointless, IMO. If we do that, we should ensure we change `_bar` to `_bar?` when `bar` is changed from being mandatory to optional. This is very tedious, and a PITA in this kind of PR, in which tons of APIs change the behavior of their params. For example, `host` of `ns.hackAnalyzeThreads` is optional now, but you did not change `_host` to `_host?`.
- You changed TSDoc of most APIs to show that their "host" param is optional now, but there are still some APIs not being changed (e.g., `ns.hackAnalyzeThreads`).

I have not double-checked TSDoc of all APIs and their respective `(_bar?) vs (_bar)` code, so I don't have a good list of APIs that need to be changed now. If it's too much, I'll do that in a follow-up PR for you.

By the way, what do you think about ditching `_bar?` in all implementation code and only using `_bar`?
Edit: I mean changing `foo: (ctx) => (_bar?) => {` to `foo: (ctx) => (_bar) => {`, regardless of whether `bar` is mandatory or optional.

src/DarkNet/models/DarknetServerOptions.ts
catloversg: I'm neutral on the idea of splitting into 2 sets.

catloversg:
FWIW, I'm not really happy about the side effect of this PR (i.e., the host params of many APIs become optional). To be clear, I'm not against that idea. I'm (slightly) against the fact that it's the side effect of a change made due to a new concept (offline servers) of dnet *and* it affects tons of non-dnet APIs. I'm okay with it affecting "ps", "killall", "hasRootAccess", etc. like you said, but not really okay with it affecting too many APIs.

I don't have any better solution, so I guess this is the price that we have to accept.

src/DarkNet/models/DarknetServerOptions.ts
d0sboots: > pull an IP from an offline server when creating a new darknet server, on occasion, if we wanted to add that feature.

Part of that comment block is explaining why we *don't* want to do that, though.

src/DarkNet/models/DarknetServerOptions.ts
d0sboots: Both hostname and IP are valid ways to refer to a server, and always have been. Almost no one uses IP, because for the regular network hostname is a stable id and IP is not. On darknet, *nothing* is stable, so either works equally well, but people still tend towards hostnames because they are more human-friendly. 

Darknet has taken a few steps to add friction to that default choice: some hostnames are awkward or hostile, and they can be reused. IPs don't have those issues. Thus, which you decide on is a trade-off.

src/DarkNet/models/DarknetState.ts
d0sboots: There was little in the way of formal test code. This consisted of making a largish (512000) element array filled with strings, and looking at the sizes in the memory snapshot tab. String construction was varied to tease out the size dependency. Array was converted to set to get that size. Etc.

d0sboots:
I could split the PR and do half the getServer API changes before the change in return value. That would keep things clean (the NS API changes would all be confined to a non-DNet refactor), at the price of touching many of the callsites twice. I'll do it if you think it's worth it, but my judgement was that it wasn't.

catloversg:
That's too troublesome for both you and me. I'm fine with this PR as it is.

src/DarkNet/models/DarknetServerOptions.ts
d0sboots: (I am happy to add more to the comment, or revise as needed, to make it more clear - I just need to know what would clarify matters.)

src/NetscriptFunctions.ts
d0sboots: I explicitly didn't do that, for reasons I detailed in a comment. However, I now think that reasoning wasn't completely justified. Changed, and changed the comment.

src/NetscriptFunctions.ts
d0sboots: Done.

src/Netscript/NetscriptHelpers.tsx
d0sboots: Done.

src/Netscript/NetscriptHelpers.tsx
d0sboots: Done.

d0sboots:
> There are two things I want to take note of:
> 
> * In our codebase (not just your changes here), we sometimes write `foo: (ctx) => (_bar?) => {`. The type of `_bar` is unknown, so `(_bar?)` is the same as `(_bar)`. I guess that's how we "marked" a param as an optional one, but that's pointless, IMO. If we do that, we should ensure we change `_bar` to `_bar?` when `bar` is changed from being mandatory to optional. This is very tedious, and a PITA in this kind of PR, in which tons of APIs change the behavior of their params. For example, `host` of `ns.hackAnalyzeThreads` is optional now, but you did not change `_host` to `_host?`.

I wanted to be consistent. I do agree that within the internal implementation, it is rather pointless. (I had vague thoughts about TS being able to do something useful here, but that's not the case.) I would be fine with clearing all of that out in a later PR.

> * You changed TSDoc of most APIs to show that their "host" param is optional now, but there are still some APIs not being changed (e.g., `ns.hackAnalyzeThreads`).

`hackAnalyzeThreads` takes host as the first param. You can't mark a parameter optional if it is followed by a required param. I could have changed the type signature to `| undefined`, but it felt like that would add more confusion rather than clarity. The same is true for all APIs that take host as a non-final arg, which I believe is all of the ones I didn't change - I judged that, although technically host can be undefined, documenting it as such in this form would be more confusing than it is worth.

> I have not double-checked TSDoc of all APIs and their respective `(_bar?) vs (_bar)` code, so I don't have a good list of APIs that need to be changed now. If it's too much, I'll do that in a follow-up PR for you.

I *think* I caught them all, but if there are any that I missed, it would be good to clean them up.

> By the way, what do you think about ditching `_bar?` in all implementation code and only using `_bar`? Edit: I mean changing `foo: (ctx) => (_bar?) => {` to `foo: (ctx) => (_bar) => {`, regardless of whether `bar` is mandatory or optional.

Per above, that sounds fine.

d0sboots:
The test passes locally. A merge issue? Concerning...

d0sboots:
I think the failing test was a flake. Even worse.

catloversg:
I added a comment for `allowNonDarknet`. This "quirk" is easy to miss with other contributors unless they carefully check how `checkDarknetServer` is implemented. I hope having a comment on `allowNonDarknet` will make this "quirk" a bit more "discoverable".

> hackAnalyzeThreads takes host as the first param. You can't mark a parameter optional if it is followed by a required param. I could have changed the type signature to | undefined, but it felt like that would add more confusion rather than clarity. The same is true for all APIs that take host as a non-final arg, which I believe is all of the ones I didn't change - I judged that, although technically host can be undefined, documenting it as such in this form would be more confusing than it is worth.

You are correct that the first parameter cannot be optional if it's followed by a required parameter. However, if it's defined as a required parameter, it's also not exactly correct. The declaration and the implementation should always match. In the current state, passing undefined to this kind of API is an undocumented behavior. Between that and `string | undefined`, I prefer `string | undefined`. I don't think it is confusing, but it's just my feeling.

Nonetheless, this PR can be merged now. The TSDoc issue is debatable, but it's not a blocking issue. I'm not sure what's going on with the flaky test. Let's investigate it later.

ficocelliguy:
I think the unit test can sometimes flake if the net is generated such that no server is adjacent to Darknet, which makes `getFirstDarknetServerAdjacentToDarkWeb` fail to return a server. I've added this to prevent that scenario:
https://github.com/bitburner-official/bitburner-src/pull/2489/changes#diff-f9b7764f176875db56d89675c485d89b62055f86ddd851fcd9bfdbc56884de21R206

catloversg:
I don't think it's related. If the net generator cannot generate a suitable server, `getFirstDarknetServerAdjacentToDarkWeb` will throw at `throw new Error("No darknet server adjacent to darkweb found")`. And even if it returns a string, `getDarknetServerOrThrow(target).password` will throw if the target string is invalid.

d0sboots:
Good c

> I added a comment for `allowNonDarknet`. This "quirk" is easy to miss with other contributors unless they carefully check how `checkDarknetServer` is implemented. I hope having a comment on `allowNonDarknet` will make this "quirk" a bit more "discoverable".

Good comment, thanks.

> ...
> You are correct that the first parameter cannot be optional if it's followed by a required parameter. However, if it's defined as a required parameter, it's also not exactly correct. The declaration and the implementation should always match. In the current state, passing undefined to this kind of API is an undocumented behavior. Between that and `string | undefined`, I prefer `string | undefined`. I don't think it is confusing, but it's just my feeling.

It was also undocumented behavior before. (Some of them worked with undefined, some did not.) I think in practice almost no one is actually passing `undefined` (or even dropping the call) to these functions. It's possible I've even made things slightly worse in some cases, because for many of them defaulting to the current server is not particularly useful. Take `hack` for instance: You can now call it just as `hack()`, which is technically fine but not that useful. Would it be better for that to return an error that says it expects a hostname? Maybe. Personally, I preferred the consistency in handling, but it's certainly debatable.

The same goes to a possibly greater degree with documenting things like `hackAnalyzeThreads`. It just isn't very useful to call that on your own server, and thus the ability to pass `undefined` is not terribly relevant. That's why it felt like a distraction to change the type just to document it. But, it *is* technically the type it accepts, so I am not totally against changing it either. Or potentially changing the actual type signature back.

PR #2497, opened 2026-02-12
UI: Add indicator of RFA connection status to overview panel

catloversg:
This PR implements the idea at https://github.com/bitburner-official/bitburner-src/pull/2381#issuecomment-3478372248.

Video:

https://github.com/user-attachments/assets/f7c7fd5f-8aa0-4c43-9ea0-7e977e1f4287


src/GameOptions/ui/RemoteFileApiConnectionStatus.tsx
```diff
@@ -0,0 +1,44 @@
...
+import OnlinePredictionIcon from "@mui/icons-material/OnlinePrediction";
+import { Settings } from "../../Settings/Settings";
+
+export const RemoteFileApiConnectionStatus = ({ showIcon }: { showIcon: boolean }): React.ReactElement => {
+  const [rfaConnectionStatus, setRfaConnectionStatus] = useState(getRemoteFileApiConnectionStatus());
+  getRemoteFileApiConnectionStatus;
+
+  useEffect(() => {
+    const timer = setInterval(() => {
+      setRfaConnectionStatus(getRemoteFileApiConnectionStatus());
```
d0sboots: This is a very cheap call, right? I ask because stuff like the periodic updating of the sidebar can show up as jitter that makes batching trickier; most of the time this won't be causing a UI update so the only concern is the check itself. (This matters now that it's on the global overview.)

src/RemoteFileAPI/Remote.ts
d0sboots, line 34: Optional: This is all event-based code; you could put an event in here to listen to instead of polling.

src/ui/React/CharacterOverview.tsx
```diff
@@ -183,6 +184,7 @@ export function CharacterOverview({ parentOpen, save, killScripts }: OverviewPro
             </Tooltip>
           </IconButton>
         </Box>
+        <RemoteFileApiConnectionStatus showIcon={true} />
```
d0sboots, line 187: This has a similar style to its siblings, which are both buttons. So, it feels like this should be a button too. What about a shortcut to the RFA options page?

catloversg:
@d0sboots FYI, just in case you missed this PR.

d0sboots (commented):
I forgot to submit pending comments :(

src/GameOptions/ui/RemoteFileApiConnectionStatus.tsx
```diff
@@ -31,9 +33,11 @@ export const RemoteFileApiConnectionStatus = ({ showIcon }: { showIcon: boolean
   return (
     <Box style={{ display: "flex", flex: 1, justifyContent: "flex-start", alignItems: "center" }}>
       {showIcon ? (
-        <Tooltip title={`Remote API: ${rfaConnectionStatus}`}>
-          <OnlinePredictionIcon style={{ fontSize: "30px", color }} />
-        </Tooltip>
+        <IconButton aria-label="kill all scripts" onClick={() => Router.toPage(Page.Options, { tab: "Remote API" })}>
```
d0sboots: aria-label copy-paste error. Either duplicate the tooltip title, or just "Remote API" might be sufficient?

src/RemoteFileAPI/Remote.ts
d0sboots: To expand on this: You are using a setInterval in the UI to poll for the connection state. However, the state only changes based on events (these callback hooks that I left the comment on). So, you could use an EventEmitter and directly affect the UI that way, instead of polling.

d0sboots:
Also, thanks, making a link into the options was more work than I feared.

PR #2499, opened 2026-02-13
UI: Improve navigation system of in-game documentation viewer

catloversg:
In the dnet PR, we discussed the need for supporting relative links (e.g., "programming/darknet.md" has a link pointing to "../../../../../markdown/bitburner.darknet.md"). This PR adds this feature.

`DocumentationPopUp.tsx` can use the new utility function of parsing the href, so I refactor it as well. One nice thing about that is the doc popup can load the non-NS doc now. This is crucial for another UI change that I'll make soon.


src/Documentation/ui/DocumentationPopUp.tsx
```diff
@@ -24,39 +24,32 @@ export function DocumentationPopUp({ hidden }: { hidden: boolean }) {
...
       }
-      if (openExternally) {
-        openDocExternally(path);
+      if (openExternally || forceOpenExternally) {
+        openDocExternally(newPath);
         return;
       }
-      if (isFilePath(path)) {
-        setPath(path);
+      if (isFilePath(newPath)) {
```
d0sboots: What is the purpose of this "if"? Are there cases where we want to silently do nothing on navigation? I feel like we should be relying on convertNavigatorHref to decide correctly for us.

src/Documentation/ui/DocumentationRoot.tsx
d0sboots: Same here

src/ui/React/Documentation.tsx
```diff
@@ -111,3 +113,58 @@ export function openDocExternally(path: string) {
...
+      ),
+    );
+  }
+  // Relative URL from NS docs to other NS docs. The URL is always ./bitburner.foo.md
+  // - Start with "./bitburner."
+  // - End with ".md"
+  // - Never have "/" between "./bitburner." and ".md"
+  else if (/^\.\/bitburner\.[^/]*\.md$/.test(href)) {
+    path = resolveFilePath(href, defaultNsApiPage);
+  } else if (href.startsWith("nsDoc/")) {
```
d0sboots, line 149: Can't we end up with relative links to nsDoc from the above?

d0sboots (commented):
Please add unit tests for convertNavigatorHref. It's the sort of no-dependencies function that's easy to test, but it has a ton of edge cases.

src/ui/React/Documentation.tsx
catloversg: Can you give an example? I don't get your question. There are 2 types of relative links:
- `../../../../markdown/bitburner.foo.md` (or more "../"). This only exists in non-NS docs.
- `./bitburner.foo.md`. This only exists in NS docs.

Are you asking whether we have `../nsDoc/bitburner.foo.md`?

d0sboots: My question is about the first type. You have `../../../../../markdown/bitburner.foo.md`, *specifically* with more than four "..". The `if` above here does a replace. That replace will leave you with `../nsDoc/bitburner.foo.md` when you get to this line.

Edit: I understand your confusion, because it was due to my confusion. The test here is based on `href`, not `path`. So there is no intent (or result) of re-converting the already converted path, this case falls through to the end.

src/ui/React/Documentation.tsx
d0sboots: Please put the `else` on the same line as the `}`. This was part of what confused me below. I think this reads better if you stick the comment inside the if-block (even though it pertains to the test itself).

test/jest/ui/DocumentationNavigator.test.ts
```diff
@@ -0,0 +1,81 @@
...
+    });
+    test("HTTP/HTTPS URL - Point to other places", () => {
+      const { path, forceOpenExternally } = convertNavigatorHref("https://bitburner-official.github.io", defaultPage);
+      expect(path).toStrictEqual("https://bitburner-official.github.io");
+      expect(forceOpenExternally).toStrictEqual(true);
+    });
+  });
+
+  describe("Invalid href", () => {
+    test("Relative URL from non-NS docs pointing to markdown folder", () => {
```
d0sboots:
```suggestion
    test("Relative URL from non-NS docs not pointing to markdown folder", () => {
```

d0sboots (commented):
Good tests, cleared up my confusion.

PR #2500, opened 2026-02-13
BUGFIX: Hacknet server UI shows NaN hash rate when 100% RAM is being used

catloversg:
How to reproduce this bug: Run a script that takes 100% RAM of a hacknet server, then hover the mouse over the "Production" line.

Before:

before

After:

after

### DIFF
--- a/src/Hacknet/ui/HacknetServerElem.tsx
+++ b/src/Hacknet/ui/HacknetServerElem.tsx
@@ -267,9 +267,13 @@ export function HacknetServerElem(props: IProps): React.ReactElement {
                     <br />
                     <span style={{ opacity: 0.5 }}>
                       <HashRate
-                        hashes={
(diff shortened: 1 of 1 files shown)


src/Hacknet/ui/HacknetServerElem.tsx
```diff
@@ -266,7 +266,11 @@ export function HacknetServerElem(props: IProps): React.ReactElement {
                     <HashRate hashes={node.hashRate} /> current production rate.
                     <br />
                     <span style={{ opacity: 0.5 }}>
-                      <HashRate hashes={(node.hashRate * node.maxRam) / (node.maxRam - node.ramUsed)} />
+                      <HashRate
+                        hashes={
+                          node.maxRam > node.ramUsed ? (node.hashRate * node.maxRam) / (node.maxRam - node.ramUsed) : 0
```
d0sboots: I think it should say "Unknown", either explicitly here, or by HashRate converting NaN to that.

catloversg:
Update:

Capture

PR #2501, opened 2026-02-14
DARKNET: Refactor/adjust getPixelPosition

d0sboots:
What the network looks like now:
image
image

This simplifies the logic. It also adjusts the position of special servers slightly; in particular, they are horizontally centered (appearing in-between the adjacent row, and not merely staggered). I didn't touch the regular staggering being asymmetrical, since that seemed intentional and visually appealing.

The split into two functions is in preparation for perf improvements that require calculating this without access to the server.


ficocelliguy (approved):

ficocelliguy:
Yeah, the regular staggering being slightly asymmetrical really helps keep the net from looking too blocky, while still having clear depth/layers

PR #2505, opened 2026-02-15
CODEBASE: Refactor ImportSave component

catloversg:
While working on a UI PR, I noticed the weird `ImportSave` component. This is the only component that is in a dedicated folder. We created a folder for it, added relevant files, and re-exported things in `ImportSave.tsx`. I checked #599, which is when we did this, and I'm not convinced that it was a good idea.


d0sboots:
I think this was a remnant of an older coding practice.

PR #2512, opened 2026-02-18
DNET: Adjust balance from player feedback

ficocelliguy:
- Reduce the base money from phishing attacks
- Reduce the money from phishing on lower-level servers

- Reduced the chance of .ccts spawning from authentication or from cache rewards on low-level servers

- Add another lab to the chain, requiring 5 total labs & installs on BN15, and 7 on other nodes
- Increased the cha requirement of the TRP lab on BN15 to 3000, and the final one (for TRP outside of BN15) to 4000

These can all be adjusted further as needed.

### DIFF
--- a/src/Augmentation/Augmentations.ts
+++ b/src/Augmentation/Augmentations.ts
@@ -1922,7 +1922,7 @@ export const Augmentations: Record<AugmentationName, Augmentation> = (() => {
       hacking: 1.1,
       company_rep: 1.05,
       isSpecial: true,
-      prereqs: [AugmentationName.TheHammer],
+      prereqs: [AugmentationName.TheStaff],
       factions: [],
     },
     [AugmentationName.TheSword]: {
--- a/src/Augmentation/Augmentations.ts
+++ b/src/Augmentation/Augmentations.ts
@@ -1895,6 +1895,20 @@ export const Augmentations: Record<AugmentationName, Augmentation> = (() => {
       prereqs: [AugmentationName.TheBoots],
       factions: [],
     },
+    [AugmentationName.TheStaff]: {
+      repCost: 1e4,
+      moneyCost: 1e6,
+      info:
+        "This skeletal augmentation greatly enhances the users durability and health. Inspired by the original Staff of Medicine that is said to " +
+        "have been given to Daedalus as a reward for the completion of the Labyrinth, which all modern augments are a descendant of. ",
+      stats:
+        "This augmentation increases the stasis link limit by one, and raises charisma xp by 20% and defense by 30%.",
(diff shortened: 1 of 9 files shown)


src/Augmentation/Augmentations.ts
d0sboots: The augs don't form a chain anymore. Intentional?

src/DarkNet/effects/cacheFiles.ts
```diff
@@ -62,8 +62,12 @@ export const getRewardFromCache = (server: DarknetServer, cacheName: string, sup
...
 };
 
-export const getCCTReward = (): string => {
-  const contractCount = [2, 3, 4][Math.floor(Math.random() * 3)];
+export const getCCTReward = (difficulty: number): string => {
+  if (Math.random() < difficulty * 0.2) {
+    return getMoneyReward(difficulty);
+  }
+
+  const contractCount = [1, 1, 2][Math.floor(Math.random() * 3)];
```
d0sboots:
```suggestion
  const contractCount = Math.floor(Math.min(20, difficulty) * 0.2 - 1.5 + Math.random() * 3);
  if (contractCount <= 0) {
    return getMoneyReward(difficulty);
  }
```
The reason for my alternate suggestion: this varies the frequency of contracts more smoothly as the difficulty increases. One random roll is easier to analyze. It has a number of constants to tune the uniform distribution with (width, difficulty cutoff, strength at 0 difficulty, etc.)

I choose these so that at the deepest levels it's slightly stronger than it was before, but at difficulty 0 you only get 1/6 of a cct.

I also envisioned the growth being much more gradual, as opposed to being full strength after the first layer, which is why I set the difficulty cap at 20.

src/DarkNet/effects/effects.ts
```diff
@@ -41,12 +41,11 @@ export const handleSuccessfulAuth = (server: DarknetServer, threads: number, pid
   server.hasAdminRights = true;
   addClue(server);
 
-  // TODO: balance coding contract chance
-  if (Math.random() < 0.1 && server.difficulty > 2) {
+  const cctChance = Math.min(0.1, 0.02 * (server.difficulty - 1));
```
d0sboots: Optional: This maxes at a difficulty of 6 (your other formula was 5). We might want to harmonize this with the other cct chance so they maximize at the same place/rate. But, that's also not strictly needed.

src/Augmentation/Augmentations.ts
ficocelliguy: Nope, it should be. The chain is mostly informational for devs, you can't get these from regular factions, but it still makes sense to have

PR #2519, opened 2026-02-21
CCT: Add "Find Largest Rectangle in a Matrix" coding contract

Misha279-UA:
# A New CCT

CCT: New cct "Find Largest Rectangle in a Matrix"


src/CodingContract/contracts/LargestRectangle.ts
```diff
@@ -0,0 +1,243 @@
+import { getRandomIntInclusive } from "../../utils/helpers/getRandomIntInclusive";
+import { CodingContractTypes } from "../ContractTypes";
+import { CodingContractName } from "@enums";
+//279
```
catloversg: Delete this.

src/CodingContract/contracts/LargestRectangle.ts
```diff
@@ -0,0 +1,243 @@
...
+export const largestRectangle: Pick<
+  CodingContractTypes,
+  CodingContractName.LargestRectangleInAMatrix //if the histogram one is approved, this one must be deleted, and renamed to "LargestRectangleIIMatrix" everywhere
```
catloversg: Remove all unrelated code and comments. A PR should include only relevant code.

src/CodingContract/contracts/LargestRectangle.ts
```diff
@@ -0,0 +1,243 @@
...
+  //     if (sanitized === "") return null;
+  //     const arr = sanitized.split(",").map((s) => parseInt(s, 10));
+  //     if (arr.length !== 2) return null;
+  //     return arr as [number, number];
+  //   },
+  //   validateAnswer: (ans): ans is [number, number] => {
+  //     return ans != null;
+  //   },
+  // },
+  [CodingContractName.LargestRectangleInAMatrix]: { //LargestRectangleIIMatrix
```
catloversg: Unrelated comment.

src/CodingContract/contracts/LargestRectangle.ts
```diff
@@ -0,0 +1,243 @@
...
+    desc: (data: number[][]): string => {
+      let gridString = "";
+      for (const line of data) {
+        gridString += `${line.toString()},\n`;
```
catloversg: Optional: Clearly indicate the input is a `number[][]`. With this `gridString`, the description looks like this:
```
You are given a binary matrix consisting only of 0s and 1s: 
 
0,0,0,1,0,0,0,0,1, 
0,0,0,0,0,0,0,0,0, 
0,0,0,1,0,1,0,0,1, 
0,0,0,0,0,0,0,1,0, 
0,0,1,0,0,0,0,0,0, 
0,0,0,0,0,0,0,0,0, 
0,0,0,0,0,0,1,0,0, 
0,0,0,0,0,0,0,0,0, 
0,0,0,0,0,0,0,0,0, 
0,0,0,0,0,0,0,1,0, 
0,0,1,0,0,0,0,0,0,
```
In your example you show:
```
    [[1,0,0], 
     [0,0,0]] 
Answer: '[[1,0],[2,1]]' 
```
I think `gridString` should show:
```
[[0,0,0,1,0,0,0,0,1], 
[0,0,0,0,0,0,0,0,0], 
[0,0,0,1,0,1,0,0,1],
<truncated>]
```

src/CodingContract/contracts/LargestRectangle.ts
```diff
@@ -0,0 +1,243 @@
...
+      }
+      return [
+        "You are given a binary matrix consisting only of 0s and 1s:\n\n",
+        `${gridString}\n\n`,
+        "Your task is to find the two corners of the largest rectangle ([[x1,y1],[x2,y2]]), that does not contain any 1s.\n\n",
+
+        "Examples:\n",
+        "    [[1,0,0],\n",
+        "     [0,0,0]]\n",
+        "Answer: '[[1,0],[2,1]]'\n",
```
catloversg: In a 2D array, the coordinate is usually `array[row][column]`. For example, with:
```
[
  [1, 2, 3],
  [4, 5, 6],
  [7, 8, 9],
]
```
[0, 2] is `3`, not `7`.

src/CodingContract/contracts/LargestRectangle.ts
```diff
@@ -0,0 +1,243 @@
...
+        "    [[0,0,0,1],\n",
+        "     [0,0,0,0],\n",
+        "     [0,0,1,0],\n",
+        "     [0,0,0,1]]\n",
+        "Answer: '[[0,0],[1,3]]'\n",
+      ].join("");
+    },
+    difficulty: 6, //for now
```
catloversg: Unrelated comment. If you are confident that 6 is the appropriate difficulty level for this cct, then this comment is useless. If you are not sure, please include a request for discussion in your PR's description.

src/CodingContract/contracts/LargestRectangle.ts
```diff
@@ -0,0 +1,243 @@
...
+    generate: (): (0 | 1)[][] => {
+      const numRows: number = getRandomIntInclusive(4, 15);
```
catloversg: Note: It's fine to define the type explicitly like this, but TS can infer the type here.

src/CodingContract/contracts/LargestRectangle.ts
```diff
@@ -0,0 +1,243 @@
...
+              maxL = left;
+              maxR = right;
+              maxU = i - histograms[i][j] + 1;
+              maxD = i;
+            }
+          }
+        }
+      }
+      return [
+        [maxL, maxU],
```
catloversg: I did not check the logic here, but if this code uses the coordinate style that you use in the example, you need to change it to `array[row][column]`.

src/CodingContract/contracts/LargestRectangle.ts
```diff
@@ -0,0 +1,243 @@
...
+          histograms[j][i] = count;
+        }
+      }
+      let maxArea = 0;
+      for (let i = 0; i < histograms.length; i++) {
+        for (let j = 0; j < histograms[0].length; j++) {
+          if (histograms[i][j] > 0) {
+            let left = j;
+            let right = j;
+            while (histograms[i]?.[left - 1] >= histograms[i][j]) {
```
catloversg: Why do we need the optional chaining here and in the next loop?

src/CodingContract/contracts/LargestRectangle.ts
```diff
@@ -0,0 +1,243 @@
...
+            }
+          }
+        }
+      }
+
+      const userArea = Math.abs(answer[1][0] - answer[0][0] + 1) * Math.abs(answer[1][1] - answer[0][1] + 1);
+      return userArea >= maxArea;
+    },
+    convertAnswer: (ans) => {
+      const parsed = JSON.parse(ans) as unknown;
```
catloversg: I'm not exactly against the usage of `JSON.parse` here, but you should check how other contracts do this. Being consistent is nice.

src/CodingContract/contracts/LargestRectangle.ts
```diff
@@ -0,0 +1,243 @@
...
+      if (!Array.isArray(parsed)) return null;
+      if (parsed.length !== 2) return null;
+      if (
+        parsed.some(
+          (arr: unknown) => !Array.isArray(arr) || arr.length !== 2 || arr.some((v: unknown) => typeof v !== "number"),
+        )
+      )
+        return null; //i hate TS
```
catloversg: Unrelated comment.

src/CodingContract/contracts/LargestRectangle.ts
```diff
@@ -0,0 +1,243 @@
...
+      return parsed as [[number, number], [number, number]];
+    },
+    validateAnswer: (ans): ans is [[number, number], [number, number]] => {
+      return ans != null;
```
catloversg: This is wrong. You are lying to TS about the type (you only check if `ans` is nullish, but you tell tsc that `ans` is `[[number, number], [number, number]]`).

src/CodingContract/Enums.ts
```diff
@@ -28,4 +28,6 @@ export enum CodingContractName {
   EncryptionIIVigenereCipher = "Encryption II: Vigenère Cipher",
   SquareRoot = "Square Root",
   TotalPrimesInRange = "Total Number of Primes",
+  //LargestRectangleIHistogram = "Largest Rectangle in a Histogram",
+  LargestRectangleInAMatrix = "Largest Rectangle in a Matrix", //LargestRectangleIIMatrix
```
catloversg: Remove all unrelated comments.

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -9418,6 +9418,8 @@ type CodingContractNameEnumType = {
   EncryptionIIVigenereCipher: "Encryption II: Vigenère Cipher";
   SquareRoot: "Square Root";
   TotalPrimesInRange: "Total Number of Primes";
+  //LargestRectangleIHistogram: "Largest Rectangle in a Histogram";
+  LargestRectangleInAMatrix: "Largest Rectangle in a Matrix"; //LargestRectangleIIMatrix
```
catloversg: Remove all unrelated comments.

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -9454,6 +9456,8 @@ export type CodingContractSignatures = {
   "Encryption II: Vigenère Cipher": [[string, string], string];
   "Square Root": [bigint, bigint, [string, string]];
   "Total Number of Primes": [number[], number];
+  //"Largest Rectangle in a Histogram": [number[], [number, number]];
```
catloversg: Remove all unrelated comments.

catloversg (commented):
Note that I have not checked the logic in `generate`, `getAnswer`, and `solver`.

src/CodingContract/contracts/LargestRectangle.ts
```diff
@@ -0,0 +1,243 @@
...
+            }
+          }
+        }
+      }
+      return [
+        [maxL, maxU],
+        [maxR, maxD],
+      ];
+    },
+    solver: (data: number[][], answer: [[number, number], [number, number]]): boolean => {
```
catloversg: Nitpick:
- The first param should be called `state`, not `data`. In this case, `data` is the same as `state`, but it's not always the case.
- You don't need to define the type of params here. It's still fine to do that, though.

src/CodingContract/contracts/LargestRectangle.ts
Misha279-UA: i based it on Unique Path in a Grid II
image

src/CodingContract/contracts/LargestRectangle.ts
Misha279-UA: i don't quite understand what exactly wrong here...

catloversg: Your check is insufficient. It only checks if `ans` is nullish. For example, if `ans` is "" (an empty string), `ans != null` is true, but an empty string is obviously not `[[number, number], [number, number]]`. You need to do the checks like what you are doing in `convertAnswer`.

Misha279-UA: ok, i just reused the one that Merge Overlapping Intervals uses, + added `ans.length === 2 &&` to it, should work

src/CodingContract/ContractTypes.ts
```diff
@@ -114,6 +115,7 @@ export const CodingContractDefinitions: CodingContractTypes = {
   ...findLargestPrimeFactor,
   ...generateIPAddresses,
   ...hammingCode,
+  ...largestRectangle,
```
d0sboots, line 118: Note: You noticed that these are attempting to be alphabetical and added yours in the proper place, gj

src/CodingContract/contracts/LargestRectangle.ts
d0sboots: I agree with CatLover here, I think Unique Paths is slightly unclear and so it shouldn't be used as a guide in this case.

You can change the formatting with a few small tweaks: (toString is not needed inside template string formatting)
```suggestion
        gridString += `[${line}],\n`;
```

src/CodingContract/contracts/LargestRectangle.ts
```diff
@@ -0,0 +1,264 @@
...
+  //   },
+  // },
+  [CodingContractName.LargestRectangleInAMatrix]: {
+    //LargestRectangleIIMatrix
+    desc: (data: number[][]): string => {
+      let gridString = "";
+      for (const line of data) {
+        gridString += `${line.toString()},\n`;
+      }
+      return [
```
d0sboots: Although you might be following code you saw elsewhere, this will be much easier to read if you do it as one large template (backtick) string. They can have newlines embedded in them, so you can write it the way it's meant to be seen, instead of needing \n and join.

src/CodingContract/contracts/LargestRectangle.ts
```diff
@@ -0,0 +1,264 @@
...
+        "You are given a binary matrix consisting only of 0s and 1s:\n\n",
+        `${gridString}\n\n`,
```
d0sboots: Once this is in a big template string, you can add the outer brackets like this: 
```suggestion
[
${gridString}
]

```
This is slightly different from CatLover's suggestion, but it is both easy to format this way and it keeps each row visually similar (the first row didn't stick out).

src/CodingContract/contracts/LargestRectangle.ts
```diff
@@ -0,0 +1,264 @@
...
+        "Your task is to find the two corners of the largest rectangle ([[x1,y1],[x2,y2]]), that does not contain any 1s.\n\n",
```
d0sboots:
```suggestion
        "Your task is to find the two corners of the largest rectangle ([[r1,c1],[r2,c2]]), that does not contain any 1s.\n\n",
```
Going along with CatLover's point about row, column.

src/CodingContract/contracts/LargestRectangle.ts
d0sboots: I will add that 6 seems reasonable to me.

src/CodingContract/contracts/LargestRectangle.ts
```diff
@@ -0,0 +1,264 @@
...
+        "     [0,0,0,1]]\n",
+        "Answer: '[[0,0],[1,3]]'\n",
+      ].join("");
+    },
+    difficulty: 6, //for now
+    generate: (): (0 | 1)[][] => {
+      const numRows: number = getRandomIntInclusive(4, 15);
+      const numColumns: number = getRandomIntInclusive(4, 15);
+
+      const grid: (1 | 0)[][] = [];
```
d0sboots, line 47: Note: The difference from above is that here, the type *is* necessary because it started with an empty array.

When typescript can or cannot infer types is pretty subtle, and often catches even those of us who have used it for a long time by surprise. :)

src/CodingContract/contracts/LargestRectangle.ts
```diff
@@ -0,0 +1,264 @@
...
+      grid.length = numRows;
+      for (let i = 0; i < numRows; ++i) {
+        grid[i] = [];
+        grid[i].length = numColumns;
+        grid[i].fill(0);
+      }
+
+      for (let r = 0; r < numRows; ++r) {
+        for (let c = 0; c < numColumns; ++c) {
+          if (r === 0 && c === 0) {
```
d0sboots: Why are you disallowing obstacles in these two corners specifically?

src/CodingContract/contracts/LargestRectangle.ts
```diff
@@ -0,0 +1,264 @@
...
+
+      let scanned = "";
+      for (let i = Math.min(answer[0][1], answer[1][1]); i <= Math.max(answer[0][1], answer[1][1]); i++) {
+        scanned += state[i]
+          .slice(Math.min(answer[0][0], answer[1][0]), Math.max(answer[0][0], answer[1][0]) + 1)
+          .toString();
+      }
+      if (scanned.includes("1")) return false;
+
+      const histograms = Array.from({ length: state.length }, () => Array<number>(state[0].length).fill(0));
```
d0sboots: This code was copied to getAnswer. Instead, solver should call getAnswer as a subroutine to get a maximum answer, and use that to get the area to compare the user's answer.

src/CodingContract/contracts/LargestRectangle.ts
d0sboots: The simple answer would be to return `convertAnswer(ans) != null`, since it already handles the typechecking...

src/CodingContract/contracts/LargestRectangle.ts
```diff
@@ -0,0 +1,264 @@
...
+            grid[r][c] = 1;
+          }
+        }
+      }
+
+      return grid;
+    },
+    getAnswer: (data) => {
+      const histograms = Array.from({ length: data.length }, () => Array<number>(data[0].length).fill(0));
+      for (let i = 0; i < data[0].length; i++) {
```
d0sboots, line 74: Optional: `r` and `c` would be clearer as loop indices, but i and j are ok too.

src/CodingContract/contracts/LargestRectangle.ts
```diff
@@ -0,0 +1,264 @@
...
+        }
+      }
+      let maxArea = 0;
+      let maxL = 0;
+      let maxR = 0;
+      let maxU = 0;
+      let maxD = 0;
+      for (let i = 0; i < histograms.length; i++) {
+        for (let j = 0; j < histograms[0].length; j++) {
+          if (histograms[i][j] > 0) {
```
d0sboots: Optional: If you negate this test and switch to an early continue, you can save a big layer of indent.

src/CodingContract/contracts/LargestRectangle.ts
```diff
@@ -0,0 +1,264 @@
...
+          }
+          histograms[j][i] = count;
+        }
+      }
+      let maxArea = 0;
+      let maxL = 0;
+      let maxR = 0;
+      let maxU = 0;
+      let maxD = 0;
+      for (let i = 0; i < histograms.length; i++) {
```
d0sboots, line 90:
```suggestion
      for (let i = 0; i < histograms.length; i++) {
        const row = histograms[i];
```
Optional: You only ever access histograms by the current row. If you save that to a (shorter) temp var, it saves reading/writing effort and helps people see that fact easily.

src/CodingContract/contracts/LargestRectangle.ts
```diff
@@ -0,0 +1,264 @@
...
+        for (let j = 0; j < histograms[0].length; j++) {
+          if (histograms[i][j] > 0) {
+            let left = j;
+            let right = j;
+            while (histograms[i]?.[left - 1] >= histograms[i][j]) {
```
d0sboots: This is relying on a very subtle point about the sort ordering between undefined and numbers for correctness, when you access off the end of the array. It needs to be documented.

src/CodingContract/contracts/LargestRectangle.ts
```diff
@@ -0,0 +1,264 @@
...
+        answer[1][1] < 0 ||
+        answer[1][1] > state.length - 1
+      )
+        return false;
+
+      let scanned = "";
+      for (let i = Math.min(answer[0][1], answer[1][1]); i <= Math.max(answer[0][1], answer[1][1]); i++) {
+        scanned += state[i]
+          .slice(Math.min(answer[0][0], answer[1][0]), Math.max(answer[0][0], answer[1][0]) + 1)
+          .toString();
```
d0sboots: Although this does work, array has an includes method natively so there's no reason to convert to string.

d0sboots (commented):
For starters, thank you for contributing!

You've probably noticed that there's quite a bit more effort and process involved here than just coding something for your personal scripts. :) That is how we ensure that bugs don't get introduced, and the codebase maintains a consistent style and level of readability/maintainability. Different projects have different standards, but they all broadly tend to follow the same steps. 

You've got a lot of comments, but that's not unusual, especially for a first-time contributor. The PR stage is where we (the maintainers) can go through your code the most carefully, so that's when comments on the details emerge. 

Broadly, I would say I agree with the things CatLover has pointed out, so please do address those as well.

src/CodingContract/contracts/LargestRectangle.ts
Misha279-UA: i don't quite catch up...

Misha279-UA: i didn't knew the `?.` is called that way, yeah, i don't need it here

src/CodingContract/contracts/LargestRectangle.ts
Misha279-UA: cuz this func waz coppied from shortest/num path cct (don't remember which one), so yeah, this is not needed

src/CodingContract/contracts/LargestRectangle.ts
```diff
@@ -0,0 +1,166 @@
...
+      }
+
+      for (let r = 0; r < numRows; ++r) {
+        for (let c = 0; c < numColumns; ++c) {
+          // 15% chance of an element being an obstacle
+          if (Math.random() < 0.15) {
+            grid[r][c] = 1;
+          }
+        }
+      }
```
d0sboots: Here's a simple way to loop until you get something with a 0. This solves the rare issue without adding too much code for it, and without hardcoding a specific spot that's always 0.

Note: My suggestion does not have proper formatting, you'll need to `npm run format` after.
```suggestion
  let allOnes: boolean;
  do {
      allOnes = true;
      for (let i = 0; i < numRows; ++i) {
        grid[i] = [];
        grid[i].length = numColumns;
        grid[i].fill(0);
      }

      for (let r = 0; r < numRows; ++r) {
        for (let c = 0; c < numColumns; ++c) {
          // 15% chance of an element being an obstacle
          if (Math.random() < 0.15) {
            grid[r][c] = 1;
          } else {
            allOnes = false;
          }
        }
      }
  } while (allOnes);
```

test/jest/CodingContract/LargestRectangle.test.ts
```diff
@@ -0,0 +1,168 @@
...
+      ]),
+    ).toBe(false);
+    expect(
+      contract.solver(data, [
+        [0, 0],
+        [0, 7],
+      ]),
+    ).toBe(false);
+  });
+});
```
d0sboots, line 184: Here is a test for the rare case. I tested that it fails currently. It should pass with my other suggestion.
```suggestion
  test("generate doesn't return all ones", () => {
    const origRandom = Math.random;
    let calls = 0;
    const mockRandom = () => {
      if (calls++ < 100) {
        return 0;
      }
      return origRandom();
    };
    try {
      Math.random = mockRandom;
      expect(contract.generate().flat()).toContain(0);
    } finally {
      Math.random = origRandom;
    }
  });
});
```

d0sboots (commented):
I think there is just the one issue left. I left a couple suggestions to help you along.

test/jest/CodingContract/LargestRectangle.test.ts
d0sboots: Please do add this one, the added test is there for a reason.

src/CodingContract/contracts/LargestRectangle.ts
```diff
@@ -0,0 +1,172 @@
+//import { ro, tr } from "date-fns/locale"; idk why this was added
```
d0sboots: Just remove this, don't comment it out.

PR #2521, opened 2026-02-21
DOCUMENTATION: Use relative links instead of absolute links

catloversg:
Using absolute links pointing to a specific branch is problematic when there are differences between branches. For example, let's say we add `foo.md` which has a link pointing to `bar.md`, and both of them are only available on dev. If the absolute link points to the stable branch, the URL is invalid.

Note:
- I don't change the link (`https://github.com/bitburner-official/bitburner-src/blob/stable/src/Themes/README.md`) in `src\Themes\ui\ThemeCollaborate.tsx`.
- I sometimes delete unnecessary links. We tend to use too many links in our doc. To be honest, this is ridiculous, IMO. You can see some examples in `src\Documentation\doc\en\basic\codingcontracts.md` or `src\Documentation\doc\en\help\getting_started.md`. For example:
```
To briefly summarize: Each [Server](../basic/servers.md) has a security level that affects how difficult it is to hack.
Each [Server](../basic/servers.md) also has a certain amount of money, as well as a maximum amount of money it can hold.
[Hacking](../basic/hacking.md) a [Server](../basic/servers.md) steals a percentage of that [Server](../basic/servers.md)'s money.
The `hack()` function is used to hack a [Server](../basic/servers.md).
The `grow()` function is used to increase the amount of money available on a [Server](../basic/servers.md).
The `weaken()` function is used to decrease a [Server](../basic/servers.md)'s security level.
```
In a very short paragraph, we added 7 links pointing to a single page.

### DIFF
--- a/src/Documentation/doc/en/advanced/bitnode_recommendation_short_guide.md
+++ b/src/Documentation/doc/en/advanced/bitnode_recommendation_short_guide.md
@@ -32,12 +32,12 @@ Because of all the benefits this [BitNode](bitnodes.md) provides, it's definitel
 
 ## Depends on your priorities
 
-BN4 gives access to the [Singularity API](https://github.com/bitburner-official/bitburner-src/blob/stable/markdown/bitburner.singularity.md).
+BN4 gives access to the [Singularity API](../../../../../markdown/bitburner.singularity.md).
 This mechanic allows the automation of many actions directly affecting the player.
 If this appeals to you then you should prioritize this BitNode.
-Due to the [RAM](../basic/ram.md) cost, it's strongly recommended to complete BN4.3 before using the [Singularity API](https://github.com/bitburner-official/bitburner-src/blob/stable/markdown/bitburner.singularity.md) in other [BitNodes](bitnodes.md).
+Due to the [RAM](../basic/ram.md) cost, it's strongly recommended to complete BN4.3 before using the Singularity API in other [BitNodes](bitnodes.md).
 
-BN6 introduces the [Bladeburner](bladeburners.md) mechanic and its corresponding [Bladeburner API](https://github.com/bitburner-official/bitburner-src/blob/stable/markdown/bitburner.bladeburner.md).
+BN6 introduces the [Bladeburner](bladeburners.md) mechanic and its corresponding [Bladeburner API](../../../../../markdown/bitburner.bladeburner.md).
 Bladeburner is an alternative method to beat BitNodes which doesn't rely on money or hacking skill.
 [Sleeves](sleeves.md) help complete BN6 more quickly since they can share some Bladeburner tasks with the player but aren't a requirement.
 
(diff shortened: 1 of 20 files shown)


src/Documentation/doc/en/advanced/bitnode_recommendation_short_guide.md
```diff
@@ -70,7 +70,7 @@ It provides a variety of useful bonuses.
 
 ## Save these for later
 
-BN7 used to give access to the [Bladeburner API](https://github.com/bitburner-official/bitburner-src/blob/stable/markdown/bitburner.bladeburner.md), but now it doesn't.
+BN7 used to give access to the Bladeburner API, but now it doesn't.
```
d0sboots: I'd re-add this one. 

I think a good rule-of-thumb is having that at least one link per section potentially makes sense. More than that might be going overboard (although it really doesn't hurt much, in the actual document it still flows reasonably.)

src/ui/MD/a.tsx
```diff
@@ -43,8 +23,19 @@ export const A = (props: React.PropsWithChildren<{ href?: string }>): React.Reac
...
+  const onClick = (event: React.MouseEvent<HTMLButtonElement, MouseEvent>) => {
+    navigator.navigate(href, event.ctrlKey);
+  };
+  const sx = href.includes(relativeUrlOfNsApiPage)
+    ? {
+        textDecorationThickness: "3px",
+        textUnderlineOffset: "5px",
+        color: Settings.theme.info,
+      }
+    : {};
```
d0sboots: What's this about? The reasoning is probably worth leaving as a comment.

src/ui/React/Documentation.tsx
```diff
@@ -150,15 +144,8 @@ export function convertNavigatorHref(
...
-    // TODO: Remove this case after converting all these links to relative links.
-    // HTTP URL pointing to NS docs.
-    // Convert https://github.com/bitburner-official/bitburner-src/blob/stable/markdown/bitburner.foo.md to nsDoc/bitburner.foo.md
-    if (href.startsWith(prefixOfHttpUrlOfNsDocs)) {
-      path = asFilePath(`nsDoc/${href.replace(prefixOfHttpUrlOfNsDocs, "")}`);
-    } else {
-      // HTTP URL pointing to other places.
-      return { path: href, forceOpenExternally: true };
-    }
+    // HTTP URL pointing to other places.
```
d0sboots: I would leave a comment that there are rare/historical usages of links to our own docs that still use this format, and we're OK with them being external links.

d0sboots (commented):
Nice cleanup work!

src/ui/MD/a.tsx
catloversg: After checking #1697, I decided to also highlight external links. I added comments explaining the reason.

PR #2523, opened 2026-02-22
DOCUMENTATION: Update BitNode recommendation short guide

catloversg:
While converting links in #2521, I noticed the paragraph of BN7 gives questionable information.

> BN7 used to give access to the Bladeburner API, but now it doesn't.

I think the OP of this guide wanted to say BN7 is not the only BN that gives access to Bladeburner API now. However, the phrasing here actually means BN7 does *not* give access to Bladeburner API, which is wrong.

> Because of that change this [BitNode](bitnodes.md) is relatively low priority and should definitely happen after BN6.

I don't think this is good advice, especially when it's in the "Save these for later" part. We usually see 4 groups of players:
- Do not like/need Bladeburner: Ignore both BN6 and BN7.
- Really like/need Bladeburner: Beat both BN6 and BN7, then use Bladeburner to beat other BNs.
- Somewhat like/need Bladeburner, but do not want to waste too much time to beat both:
  - Beat only BN7 to unlock Bladeburner and get as many Bladeburner-specific rewards as possible.
  - Beat only BN6 to unlock Bladeburner and save their time (BN6 is easier than BN7).

Based on my experience, the last group is the minority. Most people are in the first two groups (beat both BNs or none). If you must choose between 2 BNs, BN7 is actually the better choice, IMO. BN6 is only good if you want to unlock Bladeburner ASAP, but that sounds like something for speedrunning, and Bladeburner is a terrible choice for speedrunning.

I have not seen anyone beating BN6, using Bladeburner in other BNs, and then beating BN7 later.

### DIFF
--- a/src/Documentation/doc/en/advanced/bitnode_recommendation_short_guide.md
+++ b/src/Documentation/doc/en/advanced/bitnode_recommendation_short_guide.md
@@ -41,8 +41,8 @@ BN6 introduces the [Bladeburner](bladeburners.md) mechanic and its corresponding
 Bladeburner is an alternative method to beat BitNodes which doesn't rely on money or hacking skill.
 [Sleeves](sleeves.md) help complete BN6 more quickly since they can share some Bladeburner tasks with the player but aren't a requirement.
 
-BN7 unlocks Bladeburner and gives access to the Bladeburner APIs like BN6, but it's a bit harder than BN6 while giving more Bladeburner-specific rewards.
-Source-File 7 buffs Bladeburner's multipliers, and Source-File 7.3 gives free access to an augmentation that lets you perform Bladeburner and non-Bladeburner actions at the same time.
+BN7 also unlocks Bladeburner like BN6, but it's a bit harder than BN6 while giving more Bladeburner-specific rewards,
+such as an augmentation that lets you perform Bladeburner and non-Bladeburner actions at the same time.
 
 BN10 unlocks two new mechanics: [Sleeves](sleeves.md) and [Grafting](grafting.md).
 [Sleeves](sleeves.md) act as additional players and can perform tasks like studying, training, committing crime and working for factions independently of the player.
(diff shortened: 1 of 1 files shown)


src/Documentation/doc/en/advanced/bitnode_recommendation_short_guide.md
```diff
@@ -41,8 +41,11 @@ BN6 introduces the [Bladeburner](bladeburners.md) mechanic and its corresponding
 Bladeburner is an alternative method to beat BitNodes which doesn't rely on money or hacking skill.
 [Sleeves](sleeves.md) help complete BN6 more quickly since they can share some Bladeburner tasks with the player but aren't a requirement.
 
+BN7 unlocks Bladeburner and gives access to the Bladeburner APIs like BN6, but it's a bit harder than BN6 while giving more Bladeburner-specific rewards.
+Source-File 7 buffs Bladeburner's multipliers, and Source-File 7.3 gives free access to an augmentation that lets you perform Bladeburner and non-Bladeburner actions at the same time.
```
d0sboots:
```suggestion
, such as an augmentation that lets you perform Bladeburner and non-Bladeburner actions at the same time.
```

src/Documentation/doc/en/advanced/bitnode_recommendation_short_guide.md
```diff
@@ -41,8 +41,11 @@ BN6 introduces the [Bladeburner](bladeburners.md) mechanic and its corresponding
 Bladeburner is an alternative method to beat BitNodes which doesn't rely on money or hacking skill.
 [Sleeves](sleeves.md) help complete BN6 more quickly since they can share some Bladeburner tasks with the player but aren't a requirement.
 
+BN7 unlocks Bladeburner and gives access to the Bladeburner APIs like BN6, but it's a bit harder than BN6 while giving more Bladeburner-specific rewards.
```
d0sboots:
```suggestion
BN7 also unlocks Bladeburner like BN6, but it's a bit harder than BN6 while giving more Bladeburner-specific rewards
```
I'm making a few optional suggestions to cut this down to the bare minimum, in keeping with the "short guide" theme.

d0sboots (approved):

PR #2525, opened 2026-02-22
PIPE: Add pipe support for passing data into and out of terminal commands

ficocelliguy:
A number of players have asked for support for pipes in the terminal, or have been confused as to its lack of support:
https://discord.com/channels/415207508303544321/415207839246581781/1411811521935315016
https://discord.com/channels/415207508303544321/415207839246581781/1275557408726323260
https://discord.com/channels/415207508303544321/415207508303544323/932487013922390016
https://discord.com/channels/415207508303544321/415207508303544323/933463698431967283
https://discord.com/channels/415207508303544321/415207508303544323/573291980801703946

So, I've added support for them!

image

Deployed and testable at:
https://ficocelliguy.github.io/bitburner-src-2/

Features (and test cases):

- `echo` command added
- piping to text file support added - will overwrite if `>` is used, otherwise will append ( `>>` )
```
echo test;
echo test data > newFile.txt;
echo overwrite that data > newFile.txt;
```

- support for grepping string literals added
- Support for passing pipe output as additional args for any terminal command added
- support for multiple pipes in a chain added
```
echo Hello World > newFile.txt;
ls | grep .txt;
ls | grep .txt > fileList.txt;
```

- new  ns property `stdin` added: it is a port handle that receives any data passed to the script via pipes
- Support for piping to write the contents of script JS files added
- Support for piping script terminal output to another output location added
```
echo "export function main(ns) { ns.tprint('Input received: ', ns.stdin.read() ); }" > myScript.js;
echo arg1 arg2 | myScript.js;
echo arg3 arg4 | myScript.js > newLogFile.txt;
```

- piping from tail output into additional sources added
```
run myScript arg1; tail myScript arg1 >> tailLogFile.txt;
```


src/Documentation/doc/en/programming/terminal_pipes_and_redirects.md
```diff
@@ -1,28 +1,151 @@
...
+Writes command output to a file. If the file already exists, its contents are replaced.
+
+```
+[home /]> echo "first" > notes.txt
+[home /]> echo "second" > notes.txt
+```
+
+After this, `notes.txt` contains only `second`.
+
+**Note:** Script files (`.js`, `.ts`, `.jsx`, `.tsx`) can only be written to with `>` if they are empty files. Attempts to overwrite a non-empty script file are blocked. This is to prevent accidentally deleting player scripts.
```
d0sboots: I understand the motivation behind this, but I think I am mildly against it. Here's my reasoning:

* It is (yet another) incompatibility with unix logic.
* It adds more edge cases to our code.
* It makes writing utilities that generate code significantly more annoying, which (AFAICT) is the only legitimate use of `> script.js` in the first place.
* (Most importantly) using redirects is already a pretty advanced feature, so I don't think people accidentally overwriting their scripts by using `>` is something we need to be overly worried about.

src/Documentation/doc/en/programming/terminal_pipes_and_redirects.md
```diff
@@ -1,28 +1,151 @@
...
+[home /]> echo "some data" | myScript.js | anotherScript.js > output.txt
+```
+
+### Input redirect: `<`
+
+Reads a file and provides its contents as stdin to a command. This can only be used as the **first** command in a pipe chain.
 
 ```
-[home /]> echo test123 >> newFile.txt
+[home /]> cat < myfile.txt | cat
```
d0sboots: I would use your `grep` example from above here, to show how `<` can be used to avoid needing a pipe.
`grep "error" < myfile.txt`

I get that you want to show the "only first" property, so maybe it can be "in addition to" instead of replacing.

src/Documentation/doc/en/programming/terminal_pipes_and_redirects.md
```diff
@@ -1,28 +1,151 @@
...
 ```
 
-After this, `newFile.txt` will be created (if it didn't exist) and will contain `test123`
+If you try to use `<` anywhere other than the first position, you will get an error.
+
+### Semicolons: `;`
+
+Semicolons let you run multiple independent commands on one line. Each command is separate — they do not share pipes or redirects.
```
d0sboots, line 105: I'd add something about sequencing here. Specifically, in Bitburner (unlike in unix), each command runs in parallel IIRC, and this is noticable if you try to do `hack;hack`.

src/Documentation/doc/en/programming/terminal_pipes_and_redirects.md
```diff
@@ -1,28 +1,151 @@
...
+
+```
+[home /]> echo "foo" > file1.txt; echo "bar" > file2.txt
+```
+
+### The `$!` variable
+
+`$!` expands to the PID (process ID) of the last script that was started with `run`. If the last command was not a `run`, it expands to `-1`.
```
d0sboots: Why do the second part? Why not just leave it as the last command started with run?

I'm thinking especially of scenarios like this:
```
run myScript.js
> Running script with 1 thread, pid 8 and args: [].
echo $!
> 8
echo $!
> -1
```
feesbadman

src/Documentation/doc/en/programming/terminal_pipes_and_redirects.md
```diff
@@ -1,28 +1,151 @@
...
+
+You can also use it with `echo` to capture the PID:
+
+```
+[home /]> run myScript.js; echo $! > pid.txt
+```
+
+### Piping into scripts
+
+When a script is part of a pipe chain, it can read from stdin using `ns.getStdin()`. The script receives whatever the previous command wrote to stdout.
```
d0sboots: Just say "the terminal" instead of stdout, it's clearer in our case.

src/Documentation/doc/en/programming/terminal_pipes_and_redirects.md
```diff
@@ -58,3 +181,27 @@ export async function main(ns) {
...
+
+Bitburner's terminal supports a useful subset of Unix pipes and redirects, but it is not a full Unix shell. This is intentional - players are encouraged to solve problems via script, or write their own terminal utilities as scripts.
+
+The following common Unix features are **not supported**:
+
+- **Here-documents (`<< EOF`)** — There is no way to provide multi-line string literals as stdin inline in the terminal.
+- **Here-strings (`<<<`)** — The `<<<` operator for passing a single string directly to stdin is not supported.
+- **Stderr redirect (`2>`, `2>>`, `&>`)** — There is no way to separately redirect or suppress error output.
+- **Process substitution (`<(cmd)`, `>(cmd)`)** — Commands cannot be used in place of file arguments.
+- **Background jobs (`&`)** — You cannot run commands in the background from the terminal using `&`.
```
d0sboots: This needs to be reworded a bit - it's more the case that commands are *always* run in the background.

src/Documentation/doc/en/programming/terminal_pipes_and_redirects.md
```diff
@@ -58,3 +181,27 @@ export async function main(ns) {
...
+```
+
+---
+
+## Unsupported Features
+
+Bitburner's terminal supports a useful subset of Unix pipes and redirects, but it is not a full Unix shell. This is intentional - players are encouraged to solve problems via script, or write their own terminal utilities as scripts.
+
+The following common Unix features are **not supported**:
+
```
d0sboots, line 209: Should probably add right at the top: Multi-line command processing

src/Terminal/commands/common/editor.ts
```diff
@@ -36,7 +37,7 @@ export async function main(ns) {
 
 export function commonEditor(command: string, { args, server, vim }: EditorParameters, allowZeroFiles = false): void {
   if (args.length < 1 && !allowZeroFiles) {
-    return Terminal.fatal(`Incorrect usage of ${command} command. Usage: ${command} [scriptname]`);
+    return Terminal.error(`Incorrect usage of ${command} command. Usage: ${command} [scriptname]`, getTerminalStdIO());
```
d0sboots: What is your logic for turning these, and things like these, back to error?

src/Terminal/commands/hack.ts
```diff
@@ -24,9 +24,9 @@ export function hack(
   server: BaseServer,
   stdIO: StdIO,
 ): undefined | TerminalAction {
-  if (args.length !== 0) return Terminal.fatal("Incorrect usage of hack command. Usage: hack");
-  if (server.purchasedByPlayer) return Terminal.fatal("Cannot hack your own machines!");
-  if (!server.hasAdminRights) return Terminal.fatal("You do not have admin rights for this machine!");
+  if (args.length !== 0) return Terminal.fatal("Incorrect usage of hack command. Usage: hack", stdIO);
```
d0sboots, line 27: I thought the whole point of fatal was that it *didn't* need this extra argument?

src/Terminal/getTabCompletionPossibilities.ts
```diff
@@ -317,10 +317,11 @@ export async function getTabCompletionPossibilities(fullTerminalText: string, ba
       loadedModule = await compile(script, currServ.scripts);
     } catch (e) {
       const errorData = parseUnknownError(e);
-      Terminal.fatal(
+      Terminal.error(
```
d0sboots: Stuff like this definitely feels like it should not be pipeable.

src/Terminal/Terminal.ts
```diff
@@ -209,6 +207,7 @@ export class Terminal {
           setTimeout(resolve, durationSec * 1000);
           cancel = () => {
             cancelled = true;
+            stdIO.close();
```
d0sboots: This should not be here. It should be part of the higher-level handling of TerminalAction. (I.e. Terminal itself should handle this, not every implementor of TerminalAction.)

src/NetscriptFunctions.ts
```diff
@@ -1517,7 +1517,21 @@ export const ns: InternalAPI<NSFull> = {
   },
   getStdin: (ctx) => () => {
     const stdinHandle = ctx.workerScript.scriptRef.stdin?.handle;
-    return stdinHandle ? new PortHandle(stdinHandle.n) : null;
+    if (!stdinHandle) {
+      return null;
+    }
+    const handle = new PortHandle(stdinHandle.n);
+    // Provide only the methods the player needs to access (or write to) stdio
+    return {
```
d0sboots: This creates a new set of closures for every handle. We definitely want to be returning a class instead. If PortHandle isn't that class for some reason, then we should make a new one - but my impression was that PortHandle *was* that class.

d0sboots (commented):
I'm gonna have to review the whole thing from the top aren't I 😭

src/Terminal/commands/common/editor.ts
ficocelliguy: fatal explicitly closes the provided stdIO. For places where that didn't seem important, I used error. I can use whichever you think makes sense, though, for things like this.

src/Terminal/commands/hack.ts
ficocelliguy: fatal is used to close the StdIO, as opposed to error which just logs red text. There are only a handful of places where there isn't a StdIO already. More importantly, I kept running into places that missed passing the StdIO to fatal after the big merge conflict resolution, so it seemed more future-proof to require it.

src/Terminal/Terminal.ts
ficocelliguy: I've refactored this to include the stdIO in the TerminalAction and close it from the terminal cancel logic instead of here.

src/NetscriptFunctions.ts
ficocelliguy: PortHandle is a class. I did this to avoid exposing the port number (which are not valid for player use cases.) if a player did something like `ns.peek(ns.getStdin().n)` it would throw an error about port numbers needing to be positive.

I could make a new class to do something similar, exposing only the relevant methods? or maybe I can just ignore this odd use case?

src/Documentation/doc/en/programming/terminal_pipes_and_redirects.md
ficocelliguy: originally this was to be more unix-like - everything has a pid, not just running scripts - but I see your point. I've removed the -1 overwrite.

src/Documentation/doc/en/programming/terminal_pipes_and_redirects.md
ficocelliguy: "read from the terminal" is what current `grep -p` does, unlike getStdin. stdin isn't the visible terminal broadly, but only data piped in from an output redirect of some sort. I've re-worded this to try to improve the clarity.

src/Documentation/doc/en/programming/terminal_pipes_and_redirects.md
ficocelliguy: long-running commands are awaited one at a time when separated by semicolons in dev currently, afaik as a result of the TerminalAction changes you recently worked on
image

src/Documentation/doc/en/programming/terminal_pipes_and_redirects.md
ficocelliguy: I deleted a script accidentally messing around with this, which was pretty rough, and one of the few ways players can do this. But I guess rm or similar could too? It's a source control check, I suppose. I've pulled out this restriction.

src/NetscriptFunctions.ts
ficocelliguy: I've changed this to return an IOStream class instance. The other properties (e.g. ns.getStdin().handle.n) are undocumented in the return type, so are unlikely to cause confusion for players

ficocelliguy:
I think this meets all of d0s's requirements (asides from a couple of pending style notes). @catloversg can I ask you to review this, now that it's in a stable state?

catloversg:
I'll review this PR after reviewing #2889 and #2992.

E9cipher:
Is this ready yet or are there any more changes to come?

ficocelliguy:
@E9cipher this is feature-complete and waiting on reviews. (I've just done a pass of minor fixes after asking Claude to look for bugs)

PR #2527, opened 2026-02-23
CODEBASE: Update comment of LoadingScreen of ComplexPage enum

catloversg:
`LoadingScreen` was added in #1474. After checking that PR, the comment (`Has no PageContext, and thus toPage() cannot be used`), and its current callers, I still don't understand why it's in `ComplexPage` instead of `SimplePage` (I don't get what that comment says).

I briefly tested the UI, and I didn't see any problems.

### DIFF
--- a/src/ui/Enums.ts
+++ b/src/ui/Enums.ts
@@ -11,7 +11,6 @@ export enum ToastVariant {
  * transition to. You can use setPage() with these.
  */
 export enum SimplePage {
-  LoadingScreen = "Loading Screen",
   ActiveScripts = "Active Scripts",
   RecentlyKilledScripts = "Recently Killed Scripts",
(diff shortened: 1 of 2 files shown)


d0sboots:
OK, I had to do code archeology to reconstruct my state of mind. You're right, the comment wasn't as clear as it should have been.

"Has no PageContext, and thus toPage() cannot be used" was not a statement about why it had to be placed in `ComplexPage` exactly. Here's an alternate wording for the comment that might make things more clear:
```
// LoadingScreen is a special state that should never be returned to after the initial game load. To enforce this, it is
// constructed as a ComplexPage with no PageContext, and thus toPage() cannot be used (since no overload will fit it).
```

catloversg:
I reverted the changes and updated the comments.

Issue #2530, opened 2026-02-24
darknet servers can scp to darkweb when airgapped


krushia:
In my darknet crawler...
````
    // Get a list of all darknet hostnames directly connected to the current server
    const nearbyServers = ns.dnet.probe();

    // disconnection experiment
    if (!disconnecttestdone && ns.dnet.isDarknetServer() && ns.self().server !== "darkweb" && ns.self().server !== "home" && nearbyServers.length == 0) {
      ns.write("disconnectedtest.txt");
      const dctp = ns.scp("disconnectedtest.txt", "darkweb");
      ns.tprint(`Performed disconnected scp test to darkweb from ${ns.self().server}: ${dctp}`);

      disconnecttestdone = true;
    }
````
Expected to never work, but instead it always works.
Also I've noticed that ports can be used to bypass a lot of the darknet p2p problems.

[bitburnerSave_1771961653_BN1x1.json.gz](https://github.com/user-attachments/files/25529152/bitburnerSave_1771961653_BN1x1.json.gz)

gmcew:
The `scp` function working with remote sessions and not strictly requiring local connections like `exec` requires is intentional. There's a summary in-game at the `Glossary of Terms` in the Darknet Help page on the bottom right of the network screen:

>Session - A script needs a session on a darknet server in order to scp or exec targeting that server. A script gets a session after it calls dnet.authenticate with the correct password. Scripts can also get a session by calling dnet.connectToSession (a sync API with low RAM cost) after any script authenticates successfully on that server at least once. Note that sessions are per-pid (running instance of a script) - each script needs to individually get a session with a specific darknet server in order to exec or scp to that server.
>Connected - Each server on the network has specific other servers it is linked to. These are the links seen in the UI, the servers that appear when using "scan" in the terminal, or that can be seen by calling dnet.probe. This kind of direct connection is required for most dnet API methods. ns.exec requires either a direct connection, or a backdoor, or a stasis link (which also sets a backdoor) to target a darknet server.

Use of ports is also an intended method to handle the limitations on remote access.

ficocelliguy:
Early in development I did prevent scp without a direct connection, but in playtesting it was found to unduly hamper players who used files to communicate (as opposed to ports). Use of files/ports to bypass some of the peer-to-peer puzzles is (one possible) intended solution, yes.

d0sboots:
[...]

PR #2533, opened 2026-02-25
DNET: Rebalance / player feedback

### LINKED ISSUE #2532: ns.dnet.unleashStormSeed doesn't return a promise
`ns.dnet.unleashStormSeed` docs say it returns a promise but it actually doesn't. It does return a `DarknetResult`.

ficocelliguy:
- improve docs of getServerUsedRam to clarify that it includes blocked ram
- fix jsdoc of unleashStormSeed to match its ts type. Closes #2532 
- OpenWebAccessPoint puzzle (the packet capture one) now also frequently includes its password or a connected server's password in its "noise" heartbleed logs
- make packet capture linear with threads
- slow down mutations outside of BN15
- Reduce movement mutation chance slightly to slow down player completion speed
- reduce connection density slightly at higher levels
- slightly reduce auth time and scaling from charisma
- slightly reduce charisma xp gained from auth

### DIFF
--- a/src/DarkNet/effects/effects.ts
+++ b/src/DarkNet/effects/effects.ts
@@ -62,12 +62,14 @@ export const handleFailedAuth = (server: DarknetServer, threads: number) => {
  * @param person - the player's character
  * @param attemptedPassword - the password being attempted
  * @param threads - the number of threads used for the password attempt (which speeds up the process)
+ * @param linear - if true, the time scaling is linear with the number of threads instead of having diminishing returns
  */
 export const calculateAuthenticationTime = (
   darknetServerData: DarknetServerData,
   person: IPerson = Player,
   threads = 1,
   attemptedPassword = "",
+  linear = false,
 ) => {
   const chaRequired = darknetServerData.requiredCharismaSkill;
   const difficulty = darknetServerData.difficulty;
@@ -76,7 +78,7 @@ export const calculateAuthenticationTime = (
   const diffFactor = 5;
   const baseTime = 500;
 
-  const threadsFactor = 1 / (1 + 0.2 * (threads - 1));
+  const threadsFactor = linear ? threads : 1 / (1 + 0.2 * (threads - 1));
   const skillFactor = (diffFactor * chaRequired + baseDiff) / (person.skills.charisma + 100);
   const backdoorFactor = getBackdoorAuthTimeDebuff();
   const applyUnderleveledFactor = person.skills.charisma <= chaRequired && darknetServerData.depth > 1;
--- a/src/DarkNet/effects/effects.ts
+++ b/src/DarkNet/effects/effects.ts
@@ -76,10 +76,10 @@ export const calculateAuthenticationTime = (
 
   const baseDiff = (difficulty + 1) * 100;
   const diffFactor = 5;
-  const baseTime = 600;
+  const baseTime = 850;
 
   const threadsFactor = linear ? 1 / threads : 1 / (1 + 0.2 * (threads - 1));
-  const skillFactor = (diffFactor * chaRequired + baseDiff) / (person.skills.charisma * 0.7 + 100);
+  const skillFactor = (diffFactor * chaRequired + baseDiff) / (person.skills.charisma + 150);
   const backdoorFactor = getBackdoorAuthTimeDebuff();
   const applyUnderleveledFactor = person.skills.charisma <= chaRequired && darknetServerData.depth > 1;
   const underleveledFactor = applyUnderleveledFactor ? 1.5 + (chaRequired + 50) / (person.skills.charisma + 50) : 1;
--- a/src/DarkNet/effects/effects.ts
+++ b/src/DarkNet/effects/effects.ts
@@ -78,7 +78,7 @@ export const calculateAuthenticationTime = (
   const diffFactor = 5;
   const baseTime = 600;
 
-  const threadsFactor = linear ? 1 / threads : 1 / (1 + 0.2 * (threads - 1));
+  const threadsFactor = 1 / (linear ? threads : 1 + 0.2 * (threads - 1));
   const skillFactor = (diffFactor * chaRequired + baseDiff) / (person.skills.charisma * 0.7 + 100);
   const backdoorFactor = getBackdoorAuthTimeDebuff();
   const applyUnderleveledFactor = person.skills.charisma <= chaRequired && darknetServerData.depth > 1;
--- a/src/DarkNet/effects/effects.ts
+++ b/src/DarkNet/effects/effects.ts
@@ -79,8 +79,6 @@ export const calculateAuthenticationTime = (
   const baseTime = 850;
 
   const threadsFactor = 1 / (linear ? threads : 1 + 0.2 * (threads - 1));
-  const skillFactor = (diffFactor * chaRequired + baseDiff) / (person.skills.charisma * 0.7 + 100);
-  const threadsFactor = linear ? 1 / threads : 1 / (1 + 0.2 * (threads - 1));
   const skillFactor = (diffFactor * chaRequired + baseDiff) / (person.skills.charisma + 150);
   const backdoorFactor = getBackdoorAuthTimeDebuff();
   const applyUnderleveledFactor = person.skills.charisma <= chaRequired && darknetServerData.depth > 1;
--- a/src/DarkNet/effects/effects.ts
+++ b/src/DarkNet/effects/effects.ts
@@ -76,10 +76,10 @@ export const calculateAuthenticationTime = (
 
   const baseDiff = (difficulty + 1) * 100;
   const diffFactor = 5;
-  const baseTime = 500;
+  const baseTime = 600;
 
   const threadsFactor = linear ? threads : 1 / (1 + 0.2 * (threads - 1));
-  const skillFactor = (diffFactor * chaRequired + baseDiff) / (person.skills.charisma + 100);
+  const skillFactor = (diffFactor * chaRequired + baseDiff) / (person.skills.charisma * 0.7 + 100);
   const backdoorFactor = getBackdoorAuthTimeDebuff();
   const applyUnderleveledFactor = person.skills.charisma <= chaRequired && darknetServerData.depth > 1;
   const underleveledFactor = applyUnderleveledFactor ? 1.5 + (chaRequired + 50) / (person.skills.charisma + 50) : 1;
@@ -130,10 +130,9 @@ export const getMultiplierFromCharisma = (scalar = 1) => {
   );
 };
 
-// TODO: balance xp gain
 export const calculatePasswordAttemptChaGain = (server: DarknetServerData, threads: number = 1, success = false) => {
   const baseXpGain = 3;
-  const difficultyBase = 1.12;
+  const difficultyBase = 0.8;
   const xpGain = baseXpGain + difficultyBase ** server.difficulty;
   const alreadyHackedMult = server.hasAdminRights ? 0.2 : 1;
   const successMult = success && !server.hasAdminRights ? 10 : 1;
--- a/src/DarkNet/effects/effects.ts
+++ b/src/DarkNet/effects/effects.ts
@@ -78,7 +78,7 @@ export const calculateAuthenticationTime = (
   const diffFactor = 5;
   const baseTime = 600;
 
-  const threadsFactor = linear ? threads : 1 / (1 + 0.2 * (threads - 1));
+  const threadsFactor = linear ? 1 / threads : 1 / (1 + 0.2 * (threads - 1));
   const skillFactor = (diffFactor * chaRequired + baseDiff) / (person.skills.charisma * 0.7 + 100);
   const backdoorFactor = getBackdoorAuthTimeDebuff();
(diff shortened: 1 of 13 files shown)


src/DarkNet/utils/darknetNetworkUtils.ts
```diff
@@ -4,11 +4,13 @@ import { AIR_GAP_DEPTH, MS_PER_MUTATION_PER_ROW, NET_WIDTH } from "../Enums";
 import { GetAllServers } from "../../Server/AllServers";
 import { getNetDepth } from "../effects/labyrinth";
 import { CONSTANTS } from "../../Constants";
+import { Player } from "@player";
 
 export const getDarknetCyclesPerMutation = () => {
   const depth = getNetDepth();
   const cycleRate = MS_PER_MUTATION_PER_ROW / CONSTANTS.MilliPerCycle;
-  return cycleRate / depth;
+  const rateMultiplier = Player.bitNodeN !== 15 ? 0.5 : 1;
```
catloversg: This does not seem right. The number of cycles per mutation is halved, so the mutation speed increases, not decreases.

src/DarkNet/effects/effects.ts
```diff
@@ -62,22 +62,24 @@ export const handleFailedAuth = (server: DarknetServer, threads: number) => {
...
   const difficulty = darknetServerData.difficulty;
 
   const baseDiff = (difficulty + 1) * 100;
   const diffFactor = 5;
-  const baseTime = 500;
+  const baseTime = 600;
 
-  const threadsFactor = 1 / (1 + 0.2 * (threads - 1));
-  const skillFactor = (diffFactor * chaRequired + baseDiff) / (person.skills.charisma + 100);
+  const threadsFactor = linear ? threads : 1 / (1 + 0.2 * (threads - 1));
```
catloversg: This means when `linear` is true, the more threads you use, the longer the auth time is. Is this intentional? You use this when calculating the delay time of packet capturing, so the optimal way to minimize the API delay is to use only 1 thread. That sounds pretty weird.

src/DarkNet/models/packetSniffing.ts
```diff
@@ -200,8 +200,8 @@ const getLogNoise = (server: DarknetServer, logDate: Date): LogEntry => {
     return log(`--${randomServer.password}--`);
   }
 
-  if (server.modelId === ModelIds.packetSniffer && Math.random() < 0.5 / (server.difficulty + 1)) {
-    return log("Authentication successful: " + server.password);
+  if (server.modelId === ModelIds.packetSniffer && Math.random() < 0.7 - server.difficulty * 0.01) {
```
catloversg, line 203: This looks like a big jump in the probability. For example, at difficulty 10, the chance jumps from 0.04545 to 0.6. Is this intentional?
https://www.desmos.com/calculator/khocqyulmv

src/DarkNet/models/packetSniffing.ts
```diff
@@ -212,6 +212,16 @@ const log = (message: string, pid = -1) => ({
   pid,
 });
 
+const addPacketSnifferNoise = (server: DarknetServer) => {
+  const connectedServers = server.serversOnNetwork;
+  if (Math.random() < 0.3 || connectedServers.length === 0) {
```
catloversg, line 219: How can `connectedServers.length === 0` happen? When a server is not connected to any other servers, is there a way to interact with it to generate log noise? Or is log noise generated randomly for all servers periodically?

ficocelliguy: In some scenarios you can interact with the server via the UI to get logs to appear while it is not connected to anything (specifically if another server moves and breaks the connection chain to the current server). It likely will not come up very often, but I wanted to handle it gracefully.

src/DarkNet/models/packetSniffing.ts
ficocelliguy: yes. I want the packet sniffer puzzle to often have viable passwords in the "noise" data in between player heartbleed logs. It was so rare before no player actually discovered it, and I want this to be an alternate approach/solution to it in response to the feedback I've gotten on it.

src/DarkNet/utils/darknetNetworkUtils.ts
ficocelliguy: I need to do less midnight math, it seems XD

src/DarkNet/effects/effects.ts
ficocelliguy: Yeah, it should be the inverse, thank you

src/DarkNet/effects/effects.ts
```diff
@@ -76,10 +76,10 @@ export const calculateAuthenticationTime = (
 
   const baseDiff = (difficulty + 1) * 100;
   const diffFactor = 5;
-  const baseTime = 500;
+  const baseTime = 600;
 
   const threadsFactor = linear ? threads : 1 / (1 + 0.2 * (threads - 1));
-  const skillFactor = (diffFactor * chaRequired + baseDiff) / (person.skills.charisma + 100);
+  const skillFactor = (diffFactor * chaRequired + baseDiff) / (person.skills.charisma * 0.7 + 100);
```
d0sboots: This makes the auth take longer, but in a way that is not easy to intuitively analyze. IMO it would be better to multiply and divide the denominator by 0.7 to get:
`skillFactor = (diffFactor * chaRequired + baseDiff) / ((person.skills.charisma + 143) * 0.7)`
Then you can factor the constant out to baseTime to be left with:
```
const baseTime = 857;
const skillFactor = (diffFactor * chaRequired + baseDiff) / (person.skills.charisma + 143);
```
Tweak the numbers to fit your personal preference for round values - 850 and 150 seem nice.

src/DarkNet/effects/effects.ts
```diff
@@ -62,22 +62,24 @@ export const handleFailedAuth = (server: DarknetServer, threads: number) => {
...
   const difficulty = darknetServerData.difficulty;
 
   const baseDiff = (difficulty + 1) * 100;
   const diffFactor = 5;
-  const baseTime = 500;
+  const baseTime = 600;
 
-  const threadsFactor = 1 / (1 + 0.2 * (threads - 1));
-  const skillFactor = (diffFactor * chaRequired + baseDiff) / (person.skills.charisma + 100);
+  const threadsFactor = linear ? 1 / threads : 1 / (1 + 0.2 * (threads - 1));
```
d0sboots:
```suggestion
  const threadsFactor = 1 / (linear ? threads : 1 + 0.2 * (threads - 1));
```

src/DarkNet/models/packetSniffing.ts
d0sboots, line 203: Optional: The ordering of the clauses puts a ceiling on how often this can fire; Even assuming the best case for the 2nd clause you only have a `.8*1*.95*.9*.9*.95 = 0.58482` chance of reaching this line. If you want this to be common, I would move it to the top.

d0sboots (commented):
I have no opinion on the actual values you've chosen; they seem reasonable enough.

src/DarkNet/models/packetSniffing.ts
ficocelliguy: I want the player to be able to use this to sometimes get new passwords, even after they've authenticated here. if it is TOO common, though, it will just be a nearby password scraping machine and potentially speed things up a lot. I'm interested in player feedback after this goes in, though.

catloversg: The new chance is very high (0.6 at depth 10 and 0.3 at depth 40), so it's fine if it's affected by that 0.58482. Same as fico, I'm also worried that a high chance will make this puzzle spill out the password of nearby servers too easily. Let's see how this works after more testing.

src/DarkNet/models/packetSniffing.ts
catloversg: If you don't mind, please add a comment. For example:
```
// The server may become secluded if the player is interacting with it via UI and all nearby servers move and break
// their connections.
```

test/jest/Netscript/Darknet.test.ts
```diff
@@ -48,6 +48,7 @@ beforeAll(() => {
 });
 beforeEach(() => {
   DarknetState.offlineServers = new Set();
+  DarknetState.allowMutating = false;
```
catloversg: Is this part of the attempt to find out why the test is flaky? If network mutation is related to the issue, disabling it here may hide the real reason.

ficocelliguy: we don't want mutation during these unit tests either way. We can manually call the movement etc methods if we want to test those scenarios, but we don't want relevant servers going offline mid-unit test. If the test is never flaky again as a result, I'm happy with that.

catloversg: I found why the test is flaky. Unexpected network mutation is the actual reason. Dnet can only be mutated in two cases:
- The game engine calls `mutateDarknet` after a number of cycles pass.[1]
- Webstorm.[2]

Setting `DarknetState.allowMutating` only prevents [1], but [1] never happens in Jest tests (we don't call `Engine.update()` in Jest tests). [2] unexpectedly moves/deletes servers between tests. I think [2] not only makes the Jest tests flaky, but it also can create bugs or weird behaviors on prestige. I'm testing a fix for that.

Note that even if my fix works and the tests are not flaky anymore, we still should not manually disable `DarknetState.allowMutating` in the tests. Dnet should not be mutated while a test is running, regardless of `DarknetState.allowMutating`. If it mutates, it means there is a bug. In this case, the bug is that webstorm is not cancelled correctly.

gmcew:
Is it possible to state the requirements we concluded in #2530 in this PR? I think d0s' suggestion to add a detailed summary to `scp` for which of destination and source need a session (and clarify the local connection point) would close that issue too.

test/jest/Netscript/Darknet.test.ts
d0sboots: I agree with catlover here.

test/jest/Netscript/Darknet.test.ts
```diff
@@ -878,9 +879,19 @@ describe("Non-darkweb darknet server", () => {
   test("authenticate from darkweb", async () => {
     const ns = getNsOnDarkWeb();
     const target = getFirstDarknetServerAdjacentToDarkWeb();
-    const result = await ns.dnet.authenticate(target, getDarknetServerOrThrow(target).password);
-    expect(result.success).toStrictEqual(true);
+    const server = getDarknetServerOrThrow(target);
+    const result = await ns.dnet.authenticate(target, server.password);
+    // Logging details for debugging flaky test
+    if (!result.success) {
```
d0sboots, line 884: Even though catlover has identified the flakiness, I am fine with leaving this in in-case anything crops up in the future.

src/DarkNet/models/packetSniffing.ts
d0sboots: I would slightly prefer to control that by directly tuning the probability with it at the top, instead of indirectly tuning it by having this chance multiplied by the chance of reaching this line. (Which you have to calculate, and can be affected by other tuning choices we make.) But I am OK with leaving it as-is.

d0sboots:
> Is it possible to state the requirements we concluded in #2530 in this PR? I think d0s' suggestion to add a detailed summary to `scp` for which of destination and source need a session (and clarify the local connection point) would close that issue too.

We can do that here or another PR, either way.

test/jest/Netscript/Darknet.test.ts
ficocelliguy: There's no way to prevent other unit tests from running during the "mutateDarknet and webstorm > webstorm" unit test that I can find. "cancelling" it won't do much as the sleeps have been disabled and it effectively finishes in a few ms. We also mock the sleep function, making waiting around for a while harder to try and let it all finish. What is a good approach to this?

d0sboots: Per discord discussion, I think we have various options, but it might be best to scope major test changes to another PR, as they have an unknown time budget. :)

ficocelliguy: I'm going to simply move the tests that run webstorm to the bottom for now - we can do a more robust improvement in another PR

src/DarkNet/models/packetSniffing.ts
ficocelliguy: I have intended it to take the place of some of the generic liveliness check logs, rather than reducing the amount of lore & other flavorful messages. If we need to tweak it more and it gets clunky to do so where it is, though, I can move it.

test/jest/Netscript/Darknet.test.ts
catloversg: I'm not really against the idea of keeping the debug code here, but I don't think it will help. Depending on the specific delay in `launchWebstorm` and the delay in other async APIs, the flakiness can happen in different tests. In my test branch, it usually happens on `scp to dnet server and exec on dnet server`, but I also see it happens on other async code. At https://github.com/bitburner-official/bitburner-src/actions/runs/22512528669/job/65224395277, it happens on `Non-darkweb darknet server › connectToSession from darkweb`.

d0sboots: Yes, I also doubt it will help, but IDK almost never against more verbose code in tests, as long as it doesn't disturb the passing path.

PR #2535, opened 2026-02-25
BUGFIX: Fix issues with RFA auto-reconnecting feature

catloversg:
Bug 1:

https://github.com/user-attachments/assets/ef597f80-e900-4e52-b309-61f7ce1f100a

When we intentionally close the current connection before starting a new one, we call `close()` with a custom error code and check this code later in the event handler. This works with `ws` (https://www.npmjs.com/package/ws) and Deno's websocket server but does not work with Bun's websocket server and `websocket` (https://www.npmjs.com/package/websocket). When using the latter packages, the event code is not our custom error code.

Bug 2:

https://github.com/user-attachments/assets/30c04031-74fb-48e0-b753-d78e99d1a415

When clicking on the "Connect" button multiple times and there is no running RFA server, many retry attempts are scheduled, and many connections will be spawned and hit the RFA server at the same time later.

### DIFF
--- a/src/RemoteFileAPI/Remote.ts
+++ b/src/RemoteFileAPI/Remote.ts
@@ -6,11 +6,11 @@ import { Settings } from "../Settings/Settings";
 import { EventEmitter } from "../utils/EventEmitter";
 import type { getRemoteFileApiConnectionStatus } from "./RemoteFileAPI";
 
+const timeOutIds = new Set<number>();
+
 function showErrorMessage(address: string, detail: string) {
   SnackbarEvents.emit(`Error with websocket ${address}, details: ${detail}`, ToastVariant.ERROR, 5000);
 }
-
-const eventCodeWhenIntentionallyStoppingConnection = 3000;
 
 export const RemoteFileApiConnectionEvents = new EventEmitter<[ReturnType<typeof getRemoteFileApiConnectionStatus>]>();
 
@@ -26,7 +26,15 @@ export class Remote {
   }
 
   public stopConnection(): void {
-    this.connection?.close(eventCodeWhenIntentionallyStoppingConnection);
+    // Cancel all pending retries immediately. This function is only called when we intentionally close the current
+    // connection before starting a new one. The new connection will retry on its own if needed.
+    timeOutIds.forEach((id) => window.clearTimeout(id));
+    timeOutIds.clear();
+
+    if (this.connection) {
+      this.connection.intentionallyClosed = true;
+    }
+    this.connection?.close();
     RemoteFileApiConnectionEvents.emit("Offline");
   }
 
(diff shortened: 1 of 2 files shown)


src/RemoteFileAPI/Remote.ts
```diff
@@ -62,14 +70,17 @@ export class Remote {
...
+    const thisConnection = this.connection;
     this.connection.addEventListener("close", (event) => {
       /**
        * On Bitburner side, we may intentionally close the connection. For example, we do that before starting a new
        * connection. In this event handler, we do things that are only necessary when the connection is closed
        * unexpectedly (e.g., show a warning, reconnect after a delay), so we need to check whether the close event is
        * unexpected.
        */
-      if (event.code === eventCodeWhenIntentionallyStoppingConnection) {
+      if (thisConnection.intentionallyClosed) {
```
d0sboots: Optional: You could use `event.currentTarget.intentionallyClosed` instead of needing to capture.

src/RemoteFileAPI/Remote.ts
```diff
@@ -93,6 +105,7 @@ export class Remote {
 
           this.startConnection(attempts);
         }, Settings.RemoteFileApiReconnectionDelay * 1000);
+        timeOutIds.add(timeOutId);
```
d0sboots, line 105: Optional: If you call this directly on the result of `setTimeout`, instead of using a temporary, it will keep the add/delete pair together. It might cause the indent level to get worse, though.

d0sboots (commented):
I feel like there is more to bug 2 than what is being fixed here. Specifically, it shouldn't be possible to spam connection attempts like that in the first place, or if you do, it should be cancelling all the previous ones. At any given time, there should only be one "live" connection.

However, if that's too big of a change from the current model, I'm OK with merging this as-is.

src/RemoteFileAPI/Remote.ts
catloversg: I'm not sure how to do that. For example:
```js
        timeOutIds.add(
          window.setTimeout(() => {
            // timeOutIds.delete(foo);
            // Other code
          }, Settings.RemoteFileApiReconnectionDelay * 1000),
        );
```
Where do I call `timeOutIds.delete`?

catloversg:
> I feel like there is more to bug 2 than what is being fixed here. Specifically, it shouldn't be possible to spam connection attempts like that in the first place, or if you do, it should be cancelling all the previous ones. At any given time, there should only be one "live" connection.

Bug 2 is related to bug 1. We only schedule the retry if the close event is unexpected. When pressing the "Connect" button and there is a running RFA server, `event.code` is 3000 and we don't schedule the retry. However, if there is not a running RFA server, `event.code` is always 1006, and the retry is unexpectedly scheduled.

src/RemoteFileAPI/Remote.ts
d0sboots: Oh right, you very much need the variable for the delete. Durrrr.

PR #2537, opened 2026-02-25
REFACTOR: Replace Terminal callbacks with Promises and restructure code

d0sboots:
This removes the TTimer callback-based interface and replaces it with TerminalAction, a Promise-based alternative. Terminal actions that block the terminal now are represented with promises that are awaited until the action is complete. Cancelling an action (with Ctrl-C) rejects the promise.

`wget` is now a terminal-blocking action, witch Ctrl-C cancellation support. Here is what that looks like (first attempt uses Ctrl-C):
https://github.com/user-attachments/assets/709e421e-e234-4b18-8193-5929518dd01b

It also moves individual implementation functions out of Terminal and into the commands, because the new structure allows them to easily live there. This is done as a separate commit for ease of reviewing; I can split it out to a separate PR if desired.

## Testing
Manually tested:
* hack
* grow
* weaken
* backdoor
* open cache
* analyze
* scan-analyze
* wget
* wget failure
* cancelling all of the above with Ctrl-C
* chaining the above with `;`
* wget failure followed by analyze does not cancel analyze


d0sboots:
The flakiness of this test is getting out of control

d0sboots:
I rebased this, because the merge was a bit too hairy.

ficocelliguy (approved):
This looks good and works as expected in my testing.

d0sboots:
There were some non-trivial changes in the merge (in particular, I made `upload` a TerminalAction, with the largest effect being that its status line now disappears when it finishes). I'm submitting this now, before I forget about it, because it's been too long and I don't want it to accumulate more conflicts.

PR #2538, opened 2026-02-26
DARKNET: Cancel webstorm on prestige

catloversg:
There is a flaky test in dnet. While investigating it, I found out the root cause is the webstorm not being cancelled on prestige, so the network can be mutated unexpectedly. For example, if the player runs webstorm and flume, in the new BN, dnet will be mutated at any point (when the current sleep in `launchWebstorm` finishes).

You can check how to reproduce the flaky tests in https://github.com/catloversg/bitburner-src/tree/test-flaky-test. Sample log: https://github.com/catloversg/bitburner-src/actions/runs/22448333429.


ficocelliguy (approved):

catloversg:
We will either use #2542 or #2543.

PR #2539, opened 2026-02-27
MISC: Allow parsing unknown options with data.flags in autocomplete

### LINKED ISSUE #2536: UI: Don't show an error on tab auto-complete
#2328 introduced a feature which shows the user an error message if their `autocomplete` method throws an error (e.g. due to providing an invalid schema)

Unfortunately, this upsets the "tab autocomplete" feature, which works by silently calling `autocomplete` (to get a list of autocomplete args), but ignores (catches and swallows) any command libarg parsing errors because we are not yet parsing the user's command.

Now when a valid tab-autocomplete action is performed, an error message appears, even though the desired outcome is still achieved (tab-autocompletion takes place):

**Example**
My args schema defines a flag `--no-tail-windows`. If I prepare a command to run my script with just `--no-` and hit `<tab>`, I expect this to auto-complete. It does, but also raises an error:
Image

**Expected Behaviour**
No error message in this situation.

**Proposed Solution**
We cannot just revert to the old behaviour of swallowing all exceptions (shown below alongside the new behaviour), or we'll be losing the new feature that helps users troubleshoot their bad implementations:

Image

Instead, we can perhaps take advantage of the fact that libarg is throwing a well-structured error object, with a distinct error type code in this scenario:

Image

This scenario being: we are parsing arguments against a schema, and an argument is invalid (but the schema is valid). We can revert to the old behaviour of returning an empty object only in this case where `err?.code === 'ARG_UNKNOWN_OPTION'`

In the future, if we discover other false-positive error scenarios which should just be allowed to auto-complete, we can add them to the list of exceptions, but so far I haven't found any.

catloversg:
Let's say the autocomplete function is defined like this:
```
export function autocomplete(data, args) {
  data.flags([
    ["foo", true],
  ]);
  return [];
}
```
If the command is "run test.js --f[tab]", the expected behavior is "--foo" being shown in the autocomplete
result. However, the schema defines "foo", not "f", so the arg library throws the ARG_UNKNOWN_OPTION error.
This is the expected behavior of this library, but when it happens in the autocomplete function, it makes
data.flags become useless. We should suppress the error in this case.
Note that it means we also suppress legit error cases like "run test.js --bar[tab]" ("bar" is not defined
in the schema). Handling all these cases is possible, but it's unnecessarily complicated.

Closes #2536.

Alternative: The `arg` library supports an option that works similarly (https://www.npmjs.com/package/arg#permissive), but it slightly changes the behavior of `data.flags`.
```js
export function autocomplete(data, args) {
  console.log("before", args);
  console.log(data.flags([
    ["foo", true],
  ]));
  console.log("after", args);
  return [];
}
```
Now:
```
before ['--f']
{}
after ['--f']
```
With `permissive`:
```
before ['--f']
{foo:true, _: ['--f']}
after ['--f']
```
The behavior of `permissive` makes sense to me, and I think it's more "correct" if we allow unknown options, but I'm not sure if we should do that.

Edit: I changed my mind. The alternative is better.

### DIFF
--- a/src/Documentation/doc/en/basic/autocomplete.md
+++ b/src/Documentation/doc/en/basic/autocomplete.md
@@ -35,15 +35,15 @@ AutocompleteData is an object with the following properties;
 
 ```javascript
   {
-    command:    // the command being run, as seen on the terminal.
-    enums:      // the ns.enums object with various in-game strings.
-    filename:   // the name of the script file containing the autocomplete function.
-    hostname:   // the name of the host server the script would be running on.
-    processes:  // list of all processes running on the current server.
-    servers:    // list of all servers in the game. Some servers are hidden until you satisfy their requirements. This array does not contain those servers if you do not satisfy their requirements.
-    txts:       // list of all text files on the current server.
-    scripts:    // list of all scripts on the current server.
-    flags:      // the same flags function as passed with ns. Calling this function adds all the flags as autocomplete arguments.
+    command:    // The command being run, as seen on the terminal.
+    enums:      // The ns.enums object with various in-game strings.
+    filename:   // The name of the script file containing the autocomplete function.
+    hostname:   // The name of the host server the script would be running on.
+    processes:  // List of all processes running on the current server.
+    servers:    // List of all servers in the game. Some servers are hidden until you satisfy their requirements. This array does not contain those servers if you do not satisfy their requirements.
+    txts:       // List of all text files on the current server.
+    scripts:    // List of all scripts on the current server.
+    flags:      // A function similar to ns.flags(). Calling this function adds all the flags as autocomplete arguments.
   }
 ```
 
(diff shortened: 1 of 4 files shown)


src/Documentation/doc/en/basic/autocomplete.md
d0sboots: Not quite the same function anymore

src/Documentation/doc/en/basic/autocomplete.md
```diff
@@ -66,6 +66,35 @@ export function autocomplete(data, args) {
 }
 ```
 
+## data.flags
+
+This is a function that works nearly identical to `ns.flags()`. The only difference is that it allows unknown options. For example:
```
d0sboots:
```suggestion
This is a function that works nearly identically to `ns.flags()`. The only difference is that it allows unknown options. For example:
```

src/Documentation/doc/en/basic/autocomplete.md
```diff
@@ -66,6 +66,35 @@ export function autocomplete(data, args) {
 }
 ```
 
+## data.flags
```
d0sboots, line 93: Optional: I think this would order better after "args" but before "notes"

PR #2541, opened 2026-02-27
BLADEBURNER: Adjusted team bonus computation to make one member help

### LINKED ISSUE #2540: One team member in Bladeburner isn't helping
### Description
The team member bonus is computed to be `1` with a team count of `0` or `1`.
Expected: adding one team member should increase the odds of success

### Reproducing
Run the following script with at least 5 team members available. For `i=0,1`, the success rates are the same. After `1`, success rates begin to increase.
```js
/** @param {NS} ns */
  for(let i = 0; i < 5; i++) {
    ns.bladeburner.setTeamSize("Operations", "Investigation", i);
    const success = ns.bladeburner.getActionEstimatedSuccessChance("Operations", "Investigation");
    ns.tprint(`${success[0]}, ${success[1]}`);
  }
}
```

https://github.com/bitburner-official/bitburner-src/blob/d0d776700388a8ed90380bdb806b94a3b0090d94/src/Bladeburner/Actions/Operation.ts#L99
I suggest that line be changed to this:
```js
    const teamMultiplier = Math.pow(this.teamCount + 1, 0.05);
```

Note that this is technically a balance change, but it's minor enough that I don't believe it will affect... well, I'm not sure a single person is going to notice this, but I still believe it's a bug and want to report it.

JoshuaCF:
Fixed bug where there was no benefit to having a team size of 1 for a Bladeburner operation/black op

Testing: I ran the script provided in the issue on my current save file and got the following output
image

fixes #2540

### DIFF
--- a/src/Bladeburner/Actions/Operation.ts
+++ b/src/Bladeburner/Actions/Operation.ts
@@ -96,7 +96,7 @@ export const operationSkillSuccessBonus = (inst: Bladeburner) => {
 export function operationTeamSuccessBonus(this: Operation | BlackOperation, inst: Bladeburner) {
   if (this.teamCount && this.teamCount > 0) {
     this.teamCount = Math.min(this.teamCount, inst.teamSize);
-    const teamMultiplier = Math.pow(this.teamCount, 0.05);
+    const teamMultiplier = Math.pow(this.teamCount + 1, 0.05);
     return teamMultiplier;
   }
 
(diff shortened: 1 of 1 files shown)


src/Bladeburner/Actions/Operation.ts
```diff
@@ -96,7 +96,7 @@ export const operationSkillSuccessBonus = (inst: Bladeburner) => {
 export function operationTeamSuccessBonus(this: Operation | BlackOperation, inst: Bladeburner) {
   if (this.teamCount && this.teamCount > 0) {
```
d0sboots: Now that the formula is adjusted, we can take out this "if" and the redundant return, entirely.

JoshuaCF: Adjusted.

PR #2542, opened 2026-02-28
BUGFIX: Fix webstorm by using a mutationLock

d0sboots:
This is an alternative implementation to #2538

Hats off to CatLover for the hard work of running down the flakiness and getting to root cause. When I looked at that fix, I had lingering concerns about how `allowMutating` was used (in general), which led to creating this alternate design.

This approach is focused on guaranteeing safety by ensuring that only one place (the mutationLock) owns the responsibility of setting mutationLock to null, so that anything that needs to be cancelled will be properly notified (by construction).

I didn't reproduce the manual testing, but I added Darknet tests for webstorm concurrency.

This also fixes some incidental bugs found when forcing Math.random to small values for testing.

### DIFF
--- a/test/jest/Darknet/Darknet.test.ts
+++ b/test/jest/Darknet/Darknet.test.ts
@@ -723,10 +723,10 @@ describe("mutateDarknet and webstorm", () => {
     try {
       jest.useFakeTimers();
       const promise = launchWebstorm();
-      expect(DarknetState.allowMutating).toBe(false);
+      expect(DarknetState.mutationLock).toBeTruthy();
 
       await jest.advanceTimersByTimeAsync(10000);
-      expect(DarknetState.allowMutating).toBe(false);
+      expect(DarknetState.mutationLock).toBeTruthy();
 
       const initialServers = serverSnapshot();
       // Low rolls cause stuff to happen, we want deterministic testing.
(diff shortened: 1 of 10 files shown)


ficocelliguy (approved):

test/jest/Darknet/Darknet.test.ts
```diff
@@ -711,6 +718,61 @@ describe("mutateDarknet and webstorm", () => {
...
+
+      await jest.advanceTimersByTimeAsync(10000);
+      expect(DarknetState.mutationLock).toBeTruthy();
+
+      const initialServers = serverSnapshot();
+      // Low rolls cause stuff to happen, we want deterministic testing.
+      // Jest spies keep state, make our own mock so it doesn't eat memory in
+      // case something goes wrong.
+      let count = 0;
+      Math.random = () => count++ * (Number.EPSILON * 1024);
```
catloversg, line 736: What happens if you don't multiply epsilon with 1024?

catloversg (commented):
I only have one question. The rest looks good to me.

I don't have a strong opinion about choosing between this PR and my rework. It's up to you.

test/jest/Darknet/Darknet.test.ts
d0sboots: It makes the test take longer. Things like IP generation loop until a unique IP is found, using the high bits of the rng. So, this cuts the number of those loops by 1024.

PR #2543, opened 2026-02-28
DARKNET: Integrate webstorm with game engine

catloversg:
Context: #2538, #2542.

This is an alternative fix for those PRs. IMO, this entire issue starts from the way we implement webstorm.

Webstorm is the only mechanic that runs concurrently with the game engine. In the cycle-based game engine, we have a "normal" mutation occurring at a fixed pace. However, every time the player runs the storm seed exe, we create a concurrent task that mutates the network. If there is at least one webstorm running, we have to disable the normal mutation, but multiple webstorms can run at the same time. This is too messy. In this PR, I make webstorm work like any other cycle-based mechanics. It's controlled by the sync code in `processDarknet`, which also controls the normal mutation. No more async code.

d0sboots talked about how `allowMutating` is accessed and mutated by functions spread everywhere. In his PR, the mutation lock is called when we need to cancel the running webstorm, and only the lock function can reset the lock to null. In this PR, I try to "contain" the access of webstorm properties in `webstorm.ts`. This file now works like a "manager" class. If you need to know or do something with webstorm, you need to call its functions instead of directly accessing the properties and potentially creating a bug.

Regarding the webstorm properties, I'm not really fond of the design of `DarknetState`. Having a global god-object tends to cause troubles when it becomes increasingly big. For example, I had to put a comment on `serverState` to let other people know that they must use `getServerState` instead of directly using that property. Ideally, I should pull all webstorm properties to `webstorm.ts` and make them non-exported variables. However, it's a bit weird to do that now when everything else is in `DarknetState`, including `lastStormTime`. Therefore, I did not do that and just leave them in `DarknetState` as normal.

This is a PoC, so I have not tested it carefully or written Jest tests for it. If we choose this route, I'll polish it later.

### DIFF
--- a/src/DarkNet/controllers/NetworkGenerator.ts
+++ b/src/DarkNet/controllers/NetworkGenerator.ts
@@ -48,6 +48,7 @@ import type { ScriptFilePath } from "../../Paths/ScriptFilePath";
 import type { Script } from "../../Script/Script";
 import type { CodingContract } from "../../CodingContract/Contract";
 import { getTorRouter } from "../../Server/ServerHelpers";
+import { resetWebstorm } from "../effects/webstorm";
 
 export function initDarkwebServer(): void {
   const existingServer = GetServer(SpecialServers.DarkWeb);
@@ -130,10 +131,10 @@ export const clearDarknet = () => {
     deleteDarknetServer(labyrinth);
   }
 
+  resetWebstorm();
   DarknetState.zoomIndex = 7;
   DarknetState.netViewLeftScroll = 0;
   DarknetState.netViewTopScroll = 0;
-  DarknetState.allowMutating = true;
   DarknetState.openServer = null;
   DarknetState.stockPromotions = {};
   DarknetState.migrationInductionServers = new Map();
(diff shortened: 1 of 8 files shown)


d0sboots:
I think this approach can work, and be safe. However, I don't like it quite as much for several reasons: 

You identified the primary one with the reliance on DarknetState as a god-object. In a hypothetical world where we add another feature like this, that feature would need another set of variables added to DarknetState. (And they would both need to check each other's conditions, as well as their own, to suppress running. Indeed, *every* callsite that needs the equivalent of "allowMutating" would need to check *all* the conditions.) In comparison, the mutationLock-based approach would not need any new state or checks, only a new implementation of the lock for the new code. (And if it was similar enough, we'd factor it out.)

I doubt we'll actually add another webstorm-like feature. But, when all other things are mostly equal, I find hypotheticals like this help illuminate which design patterns are more robust.

Other things I like slightly less:
* The flow of the code is slightly harder to follow with the case statement vs. async + sleeps. But not terribly much so.
* I believe there are several bugs in the current implementation. To me, this points to this approach being easier to mess up.

Pros: The code being sync instead of async is a definite advantage.

d0sboots:
I thought of another example that illustrates the trade-offs involved.

Corp is similar to this implementation, in that it is modeled as a state machine with a state variable and a switch-case for processing. We could theoretically do Corp as an async loop, instead, but it would be a really bad choice because multiple different parts of the code need to know its current state, and thus the state variable is required no matter what.

I think the same applies here: *If* anything cared about the current stage of a webstorm (NS api, UI, etc.), then I would say absolutely, this is the correct implementation. But without that, the stage is an implementation detail, and one that IMO is best expressed via "current line number."

PR #2545, opened 2026-03-01
DNET: Player feedback

ficocelliguy:
- include darknet money in BN12 scaling mults
- add darknet money scaling to some other bitnodes
- reduce cct reward rates, and add a cooldown timer on caches containing ccts
- Reduce phishing money reward scaling
- Reduce phishing xp on failed attempts
- Added decaying cooldown to phishing rewards - it can't spam money rewards immediately after an install
- Chanced CCTs to only spawn from phishing caches, which are on a cooldown, and have multiple other reward possibilities
-  nerf lab augs' cha mult significantly
- nerf dnet server maximum ram 
- reduced packet capture password inclusion rate to avoid having it be stronger than solving puzzles
- Clarify that session is by PID / script instance in Authenticate and connectToSession


src/DarkNet/effects/cacheFiles.ts
```diff
@@ -63,16 +64,12 @@ export const getRewardFromCache = (server: DarknetServer, cacheName: string, sup
 };
 
 export const getCCTReward = (difficulty: number): string => {
-  if (Math.random() < difficulty * 0.2) {
+  const timeSinceLastCCTReward = new Date().getTime() - DarknetState.lastCctRewardTime.getTime();
+  if (Math.random() < difficulty * 0.2 || timeSinceLastCCTReward > 5 * 60 * 1000) {
```
d0sboots: I think you have the direction on this test backward.

In any case, I don't like adding this state. Adding variables like this makes the drop rate *much* harder to analyze, and since the desired effect is just to slow down spawning, I'd much rather just tweak the rate variables that control that directly.

src/DarkNet/effects/cacheFiles.ts
```diff
@@ -63,16 +64,12 @@ export const getRewardFromCache = (server: DarknetServer, cacheName: string, sup
...
     return getMoneyReward(difficulty);
   }
-
-  const contractCount = Math.floor(Math.min(20, difficulty) * 0.2 - 1.5 + Math.random() * 3);
```
d0sboots, line 70: We worked out this formula for a reason, don't just chuck it XD

This is the place to tune how frequently CCTs appear. If you get rid of it, you've lost all your knobs.

src/DarkNet/effects/effects.ts
```diff
@@ -41,7 +41,7 @@ export const handleSuccessfulAuth = (server: DarknetServer, threads: number, pid
   server.hasAdminRights = true;
   addClue(server);
 
-  const cctChance = Math.min(0.12, 0.02 * (server.difficulty - 1));
+  const cctChance = Math.min(0.12, 0.01 * (server.difficulty - 1));
   if (Math.random() < cctChance) {
```
d0sboots: This seems weird and hard to balance. IIUC:
* When you auth successfully, there's a chance to get a cache on the server. This cache might generate ccts.
* Independently of the above, when you auth successfully you might get a cct on the server itself.
* CCTs will periodically spawn on the network in general (which does *not* include darknet servers)

I did not fully understand this until this PR. It seems like a lot for players to understand, too. (And tricky for us to balance.) Certainly the "auth might create a CCT on the server" is not explained anywhere that I can find. Why spawn CCTs on darknet at all?

src/BitNode/BitNode.tsx
d0sboots, line 615: Other money sources are significantly reduced in 3, it seems appropriate to reduce DNET as well?

src/BitNode/BitNode.tsx
```diff
@@ -830,6 +834,8 @@ export function getBitNodeMultipliers(n: number, lvl: number): BitNodeMultiplier
         StaneksGiftPowerMultiplier: 0.5,
         StaneksGiftExtraSize: 2,
 
+        DarknetMoneyMultiplier: 0.5,
```
d0sboots, line 839: Batching is hit really hard in 9, I might reduce DNET a bit more (0.2?) to go with that.

I would also accept the idea that this is meant to be a way to make 9 (currently a notoriously hard node) easier.

src/BitNode/BitNode.tsx
```diff
@@ -979,6 +987,8 @@ export function getBitNodeMultipliers(n: number, lvl: number): BitNodeMultiplier
         StaneksGiftPowerMultiplier: inc,
         StaneksGiftExtraSize: inc,
 
+        DarknetMoneyMultiplier: dec,
```
d0sboots: Optional: Just in general, unless you think you're going to add new Dnet multipliers, I'd group these with the other `MoneyMultiplier`s.

src/DarkNet/effects/phishing.ts
```diff
@@ -38,17 +38,18 @@ export const handlePhishingAttack = (ctx: NetscriptContext, server: DarknetServe
     const bonusTimeFactor = hasDarknetBonusTime() ? 1.3 : 1;
     const depthFactor = 0.1 + server.depth * 0.05;
     const moneyReward =
-      2000 *
+      1000 *
       Player.mults.crime_money *
       depthFactor *
       threads *
-      ((200 + Player.skills.charisma) / 200) *
+      ((400 + Player.skills.charisma) / 400) *
```
d0sboots, line 43: These are relatively deep cuts - the expected amount would be reducing money by 1/8. Is dnet really *that* overpowered for early phishing? The last numbers I saw didn't seem to indicate so.

src/DarkNet/effects/cacheFiles.ts
ficocelliguy: I re-worked this to use a cct cooldown timer that is shared by all dnet mechanics that can give them (auth on server, cache from removing all blocked ram, cache from phishing)

src/DarkNet/effects/cacheFiles.ts
ficocelliguy: This formula it the spawn COUNT - it can generate multiple CCTs at a time in that code. I'm removing it because it was spawning way too many CCTs, and now will only create one, which is also limited by the cooldown timer.

src/DarkNet/effects/effects.ts
ficocelliguy: Flavor-wise, the successful auth doesn't create it, the contract was already on the server and you discovered it when you finally got to see inside of it. .ccts just appear out of the aether on servers in the main game, too.

Your point about this location being hard to balance is true, though, and why I also included the dnet cct creation cooldown timer here too.

src/BitNode/BitNode.tsx
ficocelliguy: My intention was to take the edge off of it a bit, yes

src/DarkNet/effects/phishing.ts
ficocelliguy: I've reverted this specific change to instead use an initial slowdown on phishing rewards right after an install.

(And yes, the feedback has been generally "phishing gives so much cash up front it snowballs significantly")

src/Augmentation/Augmentations.ts
```diff
@@ -1934,9 +1932,9 @@ export const Augmentations: Record<AugmentationName, Augmentation> = (() => {
         "turning every interaction into a calculated success. The technique is sometimes referred to as Solomonoff's Lightsaber, as it is a " +
         "more powerful version of Occam's razor. " +
         "The final augment awarded to those who discover the secrets of the labyrinth.",
-      stats: "This augmentation raises charisma by 50%, hacking by 16%, and company rep by 10%.",
-      charisma: 1.5,
-      hacking: 1.16,
+      stats: "This augmentation raises charisma, hacking, and company rep by 10%.",
+      charisma: 1.1,
```
d0sboots, line 1942: I feel like these are going the wrong direction. As I understand it, these augs are meant to be semi-mandatory for getting through the labs. If their effect is weak, it's basically acknowledging that you can plow through all the labs on base stats alone. 

I feel like the augs should be left alone (or even buffed), and the *base* Cha rewards treated more harshly.

src/DarkNet/effects/SaveLoad.ts
```diff
@@ -4,12 +4,14 @@ import { assertObject } from "../../utils/TypeAssertion";
...
   storedCycles: number;
   hasUsedHeartbleed: boolean;
+  phishingRewardCount: number;
 };
 
 export function getDarkNetSave(): DarknetSaveFormat {
   return {
     storedCycles: Math.floor(DarknetState.storedCycles),
     hasUsedHeartbleed: DarknetState.hasUsedHeartbleed,
+    phishingRewardCount: DarknetState.phishingRewardCount,
```
d0sboots: This continues to go further in a direction I don't like. I think adding this stuff to dnetState is both fragile and makes it harder to properly balance, as well as harder to explain what's going on.

We should just be adjusting the numbers in the various formulas, instead.

src/DarkNet/models/DarknetServerOptions.ts
```diff
@@ -58,7 +58,7 @@ export type DarknetServerOptions = {
 };
 
 export const DnetServerBuilder = (options: DarknetServerOptions): DarknetServer => {
-  const maxRam = 16 * 2 ** Math.floor(options.difficulty / 4);
+  const maxRam = 16 * 2 ** Math.floor(options.difficulty / 6);
```
d0sboots: Optional: This extends the size of the zone where everything has 16GB. It might be nice to have randomized RAM sizes, similar to the main network. Maybe this (without the floor) could form the minimum, and add a small random percent on top?

Another "fun" thing would be to roundToTwo the result, instead of floor. Servers with fractional RAM isn't actually that hard to deal with, but would be a jumpscare XD

src/DarkNet/effects/cacheFiles.ts
d0sboots: Yes, it returned a count, but the formula determined frequency. When you sample a random count, the result is average frequency. By tuning the mean frequency to be less than 1 (if desired), we could drastically reduce spawns, but you removed those knobs.

src/DarkNet/effects/effects.ts
d0sboots: OK, but you didn't address my main point: Why spawn CCTs on darknet at all? 

Does it make the game more fun/interesting to drag your CCT solver and its associated RAM cost along with the rest of your dnet code to handle this one case, or is it just a chore? To me, it feels like "chore," but I'm curious what others think.

ficocelliguy: Players have asked for ways to increase CCT spawn rate before, and this one requires a lot of coding to do, but DOES offer more access to them. You also don't have to drag your solver around, just report the cct existence / location - you can solve ccts remotely. It fits the theme and is a unique reward, rather than just "money" or "xp", and adds another interesting thing to discover on dnet servers.

I've re-instated the formula from before to add an extra tuning knob, as requested.

src/DarkNet/effects/SaveLoad.ts
ficocelliguy: This has been removed, and instead the lab augs carry a dnet money multiplier.

src/Augmentation/Augmentations.ts
ficocelliguy: As in, reduce the charisma xp you get from dnet activities, and have the lab augs be more key? the reason I was leaning towards reducing them is that these augs aren't very hard or slow to get, they don't require rep or money, so I don't want them to be too much of a boost

src/DarkNet/effects/cacheFiles.ts
ficocelliguy: I've re-instated the formula from before to add an extra tuning knob, as requested.

src/Augmentation/Augmentations.ts
d0sboots: Comment deferred to discord

src/DarkNet/models/DarknetServerOptions.ts
```diff
@@ -58,7 +59,7 @@ export type DarknetServerOptions = {
 };
 
 export const DnetServerBuilder = (options: DarknetServerOptions): DarknetServer => {
-  const maxRam = 16 * 2 ** Math.floor(options.difficulty / 6) * [1, 1, 1.1, 1.4][Math.random() * 4];
+  const maxRam = roundToTwo(16 * 2 ** Math.floor(options.difficulty / 6) * [1, 1, 1.1, 1.4][Math.random() * 4]);
```
d0sboots: If you're going to roundToTwo anyway, why not just do a simple uniform range? I.e. `(1 + Math.random() * 0.4)`

Or do you particularly want the RAM to be in discrete bands?

src/DarkNet/effects/cacheFiles.ts
d0sboots: So, now that the knobs are back, the next step is to use them, but that requires understanding what the goals are.

Obviously, one of the goals is to reduce the average number of CCTs given by caches, but there's the separate question of the *variance*. I.e. always giving 2 is the same average as giving between 1-3, but the latter is more "swingy".

Currently the variance is a width of 3 (randomness changes the spawn from +0 to +3 CCTs), which seems like fun to me, but we could tighten it if we want.

Independently, it seems like the average is felt to be too high, so tuning those constants such that the average comes down would be good.

All of this is in the hopes that we can tune CCT rates directly here, and not need an *additional* mechanism in the form of a rate limit. I *really* do not like the interaction of deterministic rate limits + randomness, they are hell to analyze and a pain to explain.

src/DarkNet/models/DarknetServerOptions.ts
ficocelliguy: I was thinking that some servers are fairly to-be-expected powers of two, and some are kinda wonky - like either someone's been overclocking or they've degraded somehow.

src/DarkNet/effects/effects.ts
gmcew: I liked the challenge of remotely solving ccts, since it's so expensive to do locally. I also agree that this significantly changes how they spawn. Would making them one off that only turn up on initial login be too punishing? It's tied to new server spawn rate (and chance of spawning with a cct) then which is easily controllable.

src/DarkNet/effects/cacheFiles.ts
ficocelliguy: other goals are "darknet is a source of ccts because its cool", but also and very importantly "there are not so many ccts from darknet that it rivals other rep gain methods."

src/DarkNet/effects/effects.ts
d0sboots: If we like the idea of them spawning on dnet and solving them remotely, then why not have caches work the same way? (Spawn 1 or more contracts onto the current server.)

This doesn't change the power of contracts (aside from balancing, since tryToSpawn is not guaranteed like directly spawning), but I think works better flavor-wise and is more understandable?

ficocelliguy: caches do sometimes spawn ccts on the current server currently, yes. You get a cache from clearing out all the blocked ram, or rarely from phishing

d0sboots: It doesn't though?
https://github.com/bitburner-official/bitburner-src/blob/90f6db6d241a898eb048b80866ecc2c7e5cf9fea/src/DarkNet/effects/cacheFiles.ts#L74
The cache reward calls `tryGeneratingRandomContract`, which is the generic entry for the CCT generator that spawns them across the network. You need `generateContract` with the current server specified (like how you do in auth) if you want to spawn on the current server.

d0sboots: Also note that `tryGenerating` only has a 1/4 chance to spawn for each count you pass it, whereas `generateContract` is guaranteed. So the numbers need to be adjusted accordingly.

catloversg:
TLDR:
- Cct rewards are so strong that the only thing keeping it under control is the very low spawn rate (1 cct/40 minutes).
- Nerfing the spawn chance may not work the way we want. It's likely that we need an additional mechanism (e.g., a cooldown) to control it.
- The cooldown should be 1 cct/ 40 minutes, which is the same as the normal cct rate. If you want to keep the cooldown at 1 cct / 10 minutes, I suggest reducing the rewards to one quarter.

### Pre-dnet CCT
Let's recap how cct spawn rate and rewards are currently implemeted.

Every 10 minutes, there is a 25% chance of generating a contract (ignoring the `currentNumberOfContracts`). This means the average spawn rate is 1 cct every 40 minutes.

There are 4 types of cct rewards: faction rep, all-factions rep, company rep, money. The reward type is determined at spawn time. When receiving the reward, if the reward type is invalid, it falls back to a different type. Faction rep and all-factions rep fall back to money. Company rep falls back to faction rep and all factions rep (50% chance for each). In a rep farming strategy, this means the chance to get the rep reward is 75%.

Rewards:
- Rep: 2500 * contractDifficulty * bnMult
- Money: 75e6 * contractDifficulty * bnMult

Contract difficulty is randomized, but the cap depends on SF levels. Let's assume the cap is 10 (the highest difficulty of a contract). The avg difficulty is 4.233333333333333. Test code:
```js
/** @param {NS} ns */
  let sum = 0;
  for (const [_, cct] of Object.entries(CodingContractTypes)) {
    sum += cct.difficulty;
  }
  console.log(sum / Object.entries(CodingContractTypes).length);
}
```
On average:
- Rep: 2500 * 4.233333333333333 * 0.75 = 5291.66666666666625
- Money: 75e6 * 4.233333333333333 * 0.75 = 238125000

Reward/s:
- Rep: 5291.66666666666625 / (40 * 60) = 2.2048611111111107
- Money: 238125000 / (40 * 60) = 99218.75

Recap rep/s from #2222:
- Hack 100 + 25 NFGs + all rep augs: 1.5544/s
- Hack 250 + 25 NFGs + all rep augs: 3.8783/s
- Hack 500 + 50 NFGs + all rep augs: 9.9412/s
- Hack 1500 + 50 NFGs + all rep augs: 29.810/s
- Hack 2500 + 50 NFGs + all rep augs: 49.680/s
- Hack 2500 + 75 NFGs + all rep augs + ns.share (1PB): 95.520/s

These are augs from Tian Di Hui, which is one of the most popular factions in early game due to their low-cost rep augs:

Tian Di Hui

Analysis:
Cct rewards are *huge*. In a speedrun, getting an early good contract is game-changing. 5.3k rep and 236e6 (avg values) let you bypass most farming time of early factions like Tian Di Hui. In early resets, the reputation gain rate is usually <1 rep/s, so getting 5.3k reputation can skip 1–2 hours of reputation farming. Even outside of speedruns, getting 2 or 3 contracts within an hour is still a decent boost.

With these huge rewards, the only thing keeping ccts under control is their very low spawn rate (at 1 cct/40 minutes, the average gain is not that significant). This makes cct a difficult reward to balance if it becomes farmable. At a very low spawn rate, it's a rare item (this is why players ask if there are other ways to spawn more of them). However, if we increase the spawn rate, it becomes OP very quickly.

### CCT in dnet

The first thing I notice is how hard it is to estimate the effective spawn rate of cct in dnet. You get cct when successfully auth a dnet server or from opening a cache, and there are 3 ways to get a cache (auth, phishing, and clearing blocked RAM). We also don't have much data on fast farming strategies (the soft-reset method I mentioned on Discord). I think we should remove the cct spawn on auth like d0sboots suggested. That would make the spawn rate easier to tune.

Another thing I notice is the report on how much rep you can get from ccts. If the testing results we get from playtesters are correct, I suspect the effective cct spawn rate is at least 100x higher than what we want. On average, cct only gives ~5.3k rep. At 1 cct/40 minutes, that's 2.2rep/s, which is tiny. However, PsY said most of their rep comes from cct, and I'm fairly sure they have not even tried specifically farming rep via cct. If this is true, nerfing the chance the way we are doing now may not be very effective.

The problem here is almost the same as the one in the auto-infil rebalance. If we nerf the reward too much, the mechanic becomes useless, but if we don't, it's OP. This entire "cct-in-dnet" started from the requests for more ways to obtain ccts. If the spawn rate is extremely low, players won't see any ccts in dnet, but if it's not low enough, the rep gain (and possibly money gain) becomes too high.

We also need to consider 2 groups of players. The first group plays the game normally, while the second group tries to farm cct via soft-reset. If we tune the rate based on the first group, the second group will gain a massive rep boost. If we tune the rate based on the second group (i.e., an extremely low spawn rate), the first group will rarely see any ccts.

In the auto-infil rebalance, we solved this by using a reward formula that penalizes very fast infiltration speeds once they reach a certain threshold. We can also do this in dnet, but it may be a bit too complicated.

My suggestions:
- Use a cooldown like fico is doing. This is a simple and effective safety net. If we fail to balance the spawn rate, it ensures the rewards do not get out of hand.
- Use the current natural spawn rate of cct as the baseline.
- IMO, cct-in-dnet should have the same effective reward rate as the "natural" cct. This means the cooldown should be 1 cct/40 minutes. If you want dnet to appear to provide many ccts, then the rewards need to be reduced. For example, at 1 cct/10 minutes, the reward should be only one quarter (applied only to ccts from dnet servers). If we look only at the average values, the reward rate of cct is tiny (2 rep/s). However, as I mentioned earlier, the actual impact of cct rewards is much larger than what the averages suggest.

### Other thoughts

Please feel free to skip this entire section.

I respect how much you value player feedback and suggestions, but sometimes I think it's important to say no. The whole cct-in-dnet idea started from suggestions to add more sources of ccts. While getting ccts from dnet is a cool idea, cct is a mechanic that is very hard to balance. From a development perspective, the cost–benefit of tuning the cct spawn rate is quite poor.

Regarding my suggested effective reward rate: when people ask for more ways to get ccts, I think what they really want is simply to see more ccts to solve. The natural spawn rate is very low, so players rarely encounter them and often create dummy ccts to test their solvers. Most players probably don't analyze our implementation deeply or intend to suggest something that would make cct rewards exploitable.

In other words, wanting to see more ccts does not necessarily mean wanting a much higher reward rate. We can satisfy that request by generating more ccts while keeping the effective reward rate relatively low. My proposal results in an effective rate of 2x (normal + cct-in-dnet), compared to the current 5x (normal + 4x normal). I think 2x is already sufficient. I have never seen anyone complain that cct rewards are too weak, and getting "lucky" ccts early in a reset is already considered a strong boost.

src/DarkNet/models/DarknetServerOptions.ts
```diff
@@ -60,7 +60,7 @@ export type DarknetServerOptions = {
 
 export const DnetServerBuilder = (options: DarknetServerOptions): DarknetServer => {
   const maxRam = roundToTwo(
-    16 * 2 ** Math.floor(options.difficulty / 6) * [1, 1, 1.15, 1.4][Math.floor(Math.random() * 4)],
+    16 * 2 ** Math.floor(options.difficulty / 6) * [0.5, 1, 1, 1.15, 1.4][Math.floor(Math.random() * 5)],
```
d0sboots: This means difficulty 0-5 servers (like darkweb) have a chance to have only 8GB of RAM, which is super-harsh. I don't think we want to impose that level of difficulty.

ficocelliguy: Good call, I've reworked this to handle base-size-16 servers separately (and have fixed darkweb's ram to a flat 16)

ficocelliguy:
Currently in this branch, ccts from darknet now work like this:
- They can only be found in caches from phishing, which have a base 3 minute cooldown on dropping.
- They are one of 5 reward types that can be rolled from a cache.
- If the cct reward type is rolled, there is a 4 out of 5 chance it simply rewards money instead.

This comes out to a cct chance of  0.25 * 0.2 every three minutes, or one cct cache an hour on average. 

The cct reward cache, if it doesn't pick money, rewards between -1.5 and 3 caches. (If less than 1 is picked, it rewards money instead). Lower level servers are capped lower (1-2 at max depending on depth) and also biased to lower counts.

(All of these numbers can be tweaked further)

src/DarkNet/effects/cacheFiles.ts
```diff
@@ -19,7 +19,7 @@ export const generateCacheFilename = (prefix?: string, isPhishingCache?: boolean
   return resolveCacheFilePath(`${filenamePrefix}_${Math.random().toString().substring(2, 5)}${suffix}`);
 };
 
-export const addCacheToServer = (server: DarknetServer, prefix?: string, isPhishingCache?: boolean) => {
+export const addCacheToServer = (server: DarknetServer, isPhishingCache?: boolean, prefix?: string) => {
```
d0sboots: Let's make isPhishingCache required, it will make everything a lot more explicit.

src/DarkNet/effects/ramblock.ts
```diff
@@ -46,7 +46,7 @@ export const handleRamBlockRemoved = (ctx: NetscriptContext, server: DarknetServ
  * Handles the rewards for fully clearing a Darknet server's RAM block.
  */
 export const handleRamBlockClearedRewards = (server: DarknetServer) => {
-  if (!isLabyrinthServer(server.hostname)) {
+  if (!isLabyrinthServer(server.hostname) && Math.random() < 0.3) {
```
d0sboots: If these type of caches don't generate CCTs, do we need to add this random factor? (I am trying to corral the sheer number of random factors we are sprouting.)

src/DarkNet/models/DarknetServerOptions.ts
```diff
@@ -195,3 +193,10 @@ const l33tifyName = (name: string): string => {
   }
   return updatedName;
 };
+
+const getMaxRam = (difficulty: number): number => {
+  const baseRam = Math.min(16 * 2 ** Math.floor(difficulty / 6), 256);
```
d0sboots: Whoa, that's a pretty big late-network nerf. This basically makes dnet uninteresting as a ram source.

src/DarkNet/models/DarknetServerOptions.ts
d0sboots: Personally, I would have implemented this as a minimum of 16GB. Having two separate formulas seems overly complex, and even for non-darkweb, being under 16GB seems rough. But, if testers think it's not an issue, I won't complain.

d0sboots:
I think we've probably gone too far the other direction and over-nerfed the ccts. I think the core problem that catlover identified was that ccts are hard to balance because they are too strong. If we included a "strength" as part of their parameters on generation, then we could keep the existing random cct generation slow and strong, but make dnet ccts weaker and more frequent. This would not only be easier to balance, but IMO it would be more "fun" since getting more ccts as rewards feels better than seeing them very rarely. Just one potential thought.

catloversg:
> If we included a "strength" as part of their parameters on generation, then we could keep the existing random cct generation slow and strong, but make dnet ccts weaker and more frequent. This would not only be easier to balance, but IMO it would be more "fun" since getting more ccts as rewards feels better than seeing them very rarely.

Agree.

In the previous message, Fico estimated the rate to be about one cct cache per hour on average. We can tune the cooldown and spawn chance to effectively cap the spawn rate at one cct cache every 10 minutes while reducing the reward to 1/6 of the normal amount. This keeps the overall reward the same while letting players see more ccts.

src/DarkNet/models/DarknetServerOptions.ts
ficocelliguy: I changed this to have a minimum of 16 gig, and just one formula.

src/DarkNet/effects/ramblock.ts
ficocelliguy: Removed, good call

src/DarkNet/models/DarknetServerOptions.ts
ficocelliguy: I was thinking of trying to limit the speed that the later servers get authenticated on, but you are right, this also affects a lot of other things that the ram could be used for as well. Reverted.

src/DarkNet/effects/cacheFiles.ts
```diff
@@ -62,16 +65,14 @@ export const getRewardFromCache = (server: DarknetServer, cacheName: string, sup
...
+  const contractCount = Math.min(Math.floor(Math.min(20, difficulty) * 0.1 - 1.5 + Math.random() * 3), 3);
+  if (contractCount < 1) {
     return getMoneyReward(difficulty);
   }
-
-  const contractCount = Math.floor(Math.min(20, difficulty) * 0.2 - 1.5 + Math.random() * 3);
-  if (contractCount <= 0) {
-    return getMoneyReward(difficulty);
+  for (let i = 0; i < contractCount; i++) {
+    generateContract({ server: server.hostname, rewardScaling: 1 / 5 });
```
ficocelliguy, line 77: This now generates contract(s) with a 1/5 multiplier on their rewards, and I removed the 1/5 chance to generate a CCT instead of money. They should appear around every 12 minutes on average, if the player is constantly phishing.

catloversg:
Can you post the calculation behind the new rate (1 cct every 12 minutes)? Similar to this message: https://github.com/bitburner-official/bitburner-src/pull/2545#issuecomment-4031301698

One thing I noticed is the removal of `cctCooldownReached`. This cooldown is the safety net that ensures the effective spawn rate is always capped at the value we want (1 cct every 12 minutes now), even if the spawn rate formula is flawed or the implementation is bugged. If you remove it, you need to make sure everything else works flawlessly, which I think is fairly risky.

What do you think about keeping that hard cap function?

ficocelliguy:
Math after the latest updates:

Phishing caches can only be acquired at most every 3 minutes (though realistically it will be a bit longer unless the player has a LOT of darknet ram and many individual scripts phishing)

There are five possible rewards from those caches - meaning on average one cct cache every 15 minutes.

The cct cache gives between -1.5 and 3 ccts, capped and weighted lower on lower-difficulty servers. The ccts from these caches are scaled to have 1/5 of the normal rewards of a cct.

There is also a cooldown timer to ensure cct caches cannot spawn more often than every 10 minutes, to ensure some consistency and avoid trying to reset for especially good luck.

catloversg:
> Phishing caches can only be acquired at most every 3 minutes

Where is this part? I only see `getPhishingCacheCooldownDuration`, but it's `hasDarknetBonusTime() ? 12_000 : 24_000`.

By the way, please double-check the usage of `&&` in tsx files. I found a misuse of it in `NetworkDisplayWrapper.tsx`:
https://github.com/bitburner-official/bitburner-src/blob/ade79c0f65bc012f2df44854ec7c147064984916/src/DarkNet/ui/NetworkDisplayWrapper.tsx#L255-L270

If `instability` is 0, React will render the number 0. For more information, please check https://react.dev/learn/conditional-rendering#logical-and-operator-.

Many people like using the `!!` trick, but I recommend explicitly writing the boolean condition that you want (e.g., `instability > 0`).

ficocelliguy:
I've reworked that tsx location as well as several others that were not explicitly using booleans in their conditional rendering in darknet. I also re-instated the three-minute timer for phishing caches - it appears to have been rolled back during some of my local editing in this branch earlier, I'm not sure it ever got committed until now, good catch.

catloversg:
Not using `!!` in those places is still okay. It only becomes an issue when the variable is a number.

Nonetheless, converting them to boolean with `!!` does not cause any problems, so it's fine either way.

catloversg:
We can merge this PR after d0sboots reviews it. I haven’t reviewed it in detail, but it looks good at first glance. Most of the changes are balance-related, so even if there are issues, they’re unlikely to cause major breakages. We can always rebalance further later.

PR #2546, opened 2026-03-02
UI: Add inline script RAM usage text to each active script

vadien:
(First time contributor, please let me know if I have missed something)
It can be hard to tell at a glance which scripts are eating up all your RAM, so I've slightly modified the Active Scripts page to include the total RAM used by a script (next to the script name), as well as the RAM usage of a single thread (shown next to thread count). 
show-ram-inline
Note that RAM used per thread is only displayed if threads > 1, so it's not shown at all for the first script.

✅ `npm run lint`
✅ `npm run format`
✅ `npm run test`

### DIFF
--- a/src/ui/ActiveScripts/WorkerScriptAccordion.tsx
+++ b/src/ui/ActiveScripts/WorkerScriptAccordion.tsx
@@ -4,7 +4,7 @@
  */
 import * as React from "react";
 
-import { formatExp, formatThreads, formatRam } from "../formatNumber";
+import { formatExp, formatThreads } from "../formatNumber";
 
 import Table from "@mui/material/Table";
 import TableCell from "@mui/material/TableCell";
@@ -33,6 +33,7 @@ import { convertTimeMsToTimeElapsedString } from "../../utils/StringHelperFuncti
 import { arrayToString } from "../../utils/helpers/ArrayHelpers";
 import { Money } from "../React/Money";
 import { MoneyRate } from "../React/MoneyRate";
+import { roundToTwo } from "../../utils/helpers/roundToTwo";
 
 const useStyles = makeStyles()({
   noborder: {
@@ -69,7 +70,7 @@ export function WorkerScriptAccordion(props: IProps): React.ReactElement {
         <ListItemText
           primary={
             <Typography sx={{ overflowWrap: "break-word" }}>
-              └ {props.workerScript.name} {formatRam(scriptRef.ramUsage * scriptRef.threads)}{" "}
+              └ {props.workerScript.name} ({roundToTwo(scriptRef.ramUsage * scriptRef.threads)}GB){" "}
               {JSON.stringify(props.workerScript.args)}
             </Typography>
           }
@@ -87,7 +88,7 @@ export function WorkerScriptAccordion(props: IProps): React.ReactElement {
                 <TableCell className={classes.noborder}>
                   <Typography>
                     {formatThreads(scriptRef.threads)}{" "}
-                    {scriptRef.threads > 1 ? `(${formatRam(scriptRef.ramUsage)} each)` : ``}
+                    {scriptRef.threads > 1 ? `(${roundToTwo(scriptRef.ramUsage)}GB each)` : ``}
                   </Typography>
                 </TableCell>
               </TableRow>
--- a/src/ui/ActiveScripts/WorkerScriptAccordion.tsx
+++ b/src/ui/ActiveScripts/WorkerScriptAccordion.tsx
@@ -4,7 +4,7 @@
  */
 import * as React from "react";
 
-import { formatExp, formatThreads } from "../formatNumber";
+import { formatExp, formatThreads, formatRam } from "../formatNumber";
 
 import Table from "@mui/material/Table";
 import TableCell from "@mui/material/TableCell";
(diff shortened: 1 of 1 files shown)


catloversg:
Optional:
- Instead of "a.js 6.3GB []", you can show "a.js (RAM: 6.3GB) []". It's debatable if saying "RAM:" is necessary. It's up to you and d0sboots.
- Make the same change in the "Recently Killed" tab (`src\ui\ActiveScripts\RecentScriptAccordion.tsx`).

src/ui/ActiveScripts/WorkerScriptAccordion.tsx
```diff
@@ -69,7 +69,8 @@ export function WorkerScriptAccordion(props: IProps): React.ReactElement {
         <ListItemText
           primary={
             <Typography sx={{ overflowWrap: "break-word" }}>
-              └ {props.workerScript.name} {JSON.stringify(props.workerScript.args)}
+              └ {props.workerScript.name} {formatRam(scriptRef.ramUsage * scriptRef.threads)}{" "}
```
catloversg: I think you should wrap the calculated value in `roundToTwo` (e.g., `roundToTwo(scriptRef.ramUsage * scriptRef.threads)`). You can wrap it with `formatRam` after that if you want, but it's unnecessary.

Internally, we always use that function when doing arithmetic operations on RAM. It does not matter much here when we only display the value (not store it or do any logical operations based on the value), but it's good to be consistent (some old UI code did not do this, though).

vadien: Done!

vadien:
> Optional:
> 
> * Instead of "a.js 6.3GB []", you can show "a.js (RAM: 6.3GB) []". It's debatable if saying "RAM:" is necessary. It's up to you and d0sboots.

I've changed the RAM figure to have enclosing braces e.g. "a.js (6.3GB) []" but not added the word RAM. I think what it represents should be fairly clear in context, especially compared to the arg array brackets. Happy to be overruled on this.

> * Make the same change in the "Recently Killed" tab (`src\ui\ActiveScripts\RecentScriptAccordion.tsx`).

I tried this change and found the information wasn't really helpful in the context of a non-running script, it only really served for consistency. I can add it back if that's preferred but I think that space on the Recently Killed page is better reserved for the hostname it ran on and maybe the PID (I can raise a separate PR for that).

src/ui/ActiveScripts/WorkerScriptAccordion.tsx
d0sboots: I'm going to disagree with catlover here. Using `formatRam` is IMO important, and proper use of `formatRam` makes `roundToTwo` unneeded.

As catlover says, `roundToTwo` is important when doing arithmetic operations on RAM. However, what we are doing here is display, not logic - i.e., we are not storing the value back in a number. This is what makes `roundToTwo` the wrong method to use.

More importantly, `formatRam` handles cases such as changing to TB when the ram usage is very high, and using the user settings to display GiB/TiB instead of GB/TB. Both of those are things we definitely want here.

src/ui/ActiveScripts/WorkerScriptAccordion.tsx
```diff
@@ -84,7 +86,10 @@ export function WorkerScriptAccordion(props: IProps): React.ReactElement {
                   <Typography>└ Threads:</Typography>
                 </TableCell>
                 <TableCell className={classes.noborder}>
-                  <Typography>{formatThreads(props.workerScript.scriptRef.threads)}</Typography>
+                  <Typography>
+                    {formatThreads(scriptRef.threads)}{" "}
+                    {scriptRef.threads > 1 ? `(${roundToTwo(scriptRef.ramUsage)}GB each)` : ``}
```
d0sboots: IMO I would remove the check vs number of threads. My reasoning:
This displays in an expanded list, so saving space is not as important. Hiding this doesn't actually save space anyway, given the table structure involved. And even when threads are 1, wanting to know "per-thread RAM usage" is still reasonable, even if it will match "overall RAM usage" - it saves the user from having to look elsewhere/make a mental jump.

d0sboots (commented):
Thanks for the thoughtful UI change!

I agree with your reasoning around use of parens and only showing it on the active page.

src/ui/ActiveScripts/WorkerScriptAccordion.tsx
vadien: Great points - didn't even consider values other than GB when I first thought of this. Amended in newest commit.

src/ui/ActiveScripts/WorkerScriptAccordion.tsx
vadien: This makes perfect sense given saving mental workload was my whole idea here!

vadien:
Newest commit:
- Switches `roundToTwo()` back to `formatRam()` and changes imports accordingly, also cleaned up my hardcoded 'GB' strings. I double checked this works as intended with TB and up as well as GiB and up.
- Removes 'number of threads' check so now RAM is shown for single threads as well.

✅ `npm run lint`
✅ `npm run format`
✅ `npm run test`

catloversg (approved):

Issue #2552, opened 2026-03-03
(Typo) Space missing in 'sector-12-crime.lit' from 'foodnstuff'


ZeroKobold:
The end of the first sentence of the document reads "criminal activity inSector-12.", but should likely be "criminal activity in Sector-12."

Same goes for "such as The Syndicate or theSlum Snakes." at the end of the second paragraph, which should be "such as The Syndicate or the Slum Snakes."

catloversg:
If you see `shows a significant rise in criminal activity inSector-12.`, it means you are playing the stable version. It was fixed on the dev version. Note that the fix is `{" "}` that I said in your PR.

PR #2553, opened 2026-03-03
BUGFIX: Fixed typo/whitespace issue in sector-12-crime.lit

ZeroKobold:
Adds missing spaces in the text.

Fixes issue #2552


catloversg:
There is no need to add the whitespace. `{" "}` is that whitespace. In jsx/tsx, if you write:
```jsx
      <Typography>
        A recent study by analytics company Wilson Inc. shows a significant rise in criminal activity in<whitespace><newline>
        {CityName.Sector12}.
```
It will render as `shows a significant rise in criminal activity inSector-12` (notice the missing whitespace). This is why we use `{" "}`.

ZeroKobold:
Thank you for the complementary information!

PR #2555, opened 2026-03-06
UI: Show "undefined" instead of -1 as pid in error popup when catching promise errors

catloversg:
Edit: It will show "undefined".

Jeek asked why the pid was -1. I guess showing "Unknown" pid is less confusing.

Test code:
```js
/** @param {NS} ns */
  ns.asleep(0).then(() => {
    throw new Error("foo");
  });
}
```

Before:

before

After:

after

### DIFF
--- a/src/ErrorHandling/ErrorModal.tsx
+++ b/src/ErrorHandling/ErrorModal.tsx
@@ -73,7 +73,7 @@ export function ErrorModal(): React.ReactElement {
             <p>
               Script: {error.scriptName}
               <br />
-              PID: {error.pid}
+              PID: {error.pid ?? "Unknown"}
             </p>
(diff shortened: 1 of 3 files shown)


src/ErrorHandling/ErrorModal.tsx
```diff
@@ -73,7 +73,7 @@ export function ErrorModal(): React.ReactElement {
             <p>
               Script: {error.scriptName}
               <br />
-              PID: {error.pid}
+              PID: {error.pid ?? "Unknown"}
```
d0sboots: Maybe it's just the programmer in me, but I'd rather we let the raw value of "undefined" show through. "Unknown" implies to me that it might have had a PID, and we just don't know what it was. "undefined" means no PID at all.

PR #2556, opened 2026-03-07
API: Add API to minimize and expand tail windows

### LINKED ISSUE #2409: Feature Request: ns.ui.rollUpTail
Request to add to the Netscript UserInterface to allow a function to roll up a tail, something like ns.ui.rollUpTail(). This would allow scripts to make simple tickers through the ns.ui.setTailTitle function. Option to allow unrolling would be nice as well.

Example use: I use a stock market script that has updates on the value of purchased stocks. I currently send the value to the title, but the window always has to be manually rolled up. If there are issues, it would be nice to have a function unroll it so I could see messages.

catloversg:
Test code:

```js
/** @param {NS} ns */
  ns.ui.openTail();
  const pid = 1;
  while (true) {
    ns.ui.setTailMinimized(true, pid);
    await ns.sleep(200);
    ns.ui.setTailMinimized(false, pid);
    await ns.sleep(200);
  }
}
```
```js
/** @param {NS} ns */
  ns.ui.openTail();
  console.log(ns.pid);
  while (true) {
    console.log(ns.getRunningScript().tailProperties?.minimized);
    await ns.sleep(1000);
  }
}
```

Video:

https://github.com/user-attachments/assets/8454a4b6-6398-4c38-9d26-a8930c638e34

Closes #2409.


src/ui/React/LogBoxManager.tsx
```diff
@@ -178,7 +185,7 @@ function LogWindow({ hidden, script, onClose }: LogWindowProps): React.ReactElem
   const rerender = useRerender(Settings.TailRenderInterval);
   const propsRef = useRef(new LogBoxProperties(rerender, rootRef));
   script.tailProps = propsRef.current;
-  const [minimized, setMinimized] = useState(false);
+  const [minimized, setMinimized] = useState(propsRef.current.minimized);
```
d0sboots: The state is contained in tailProps, there is no reason or need for additional state variables.

src/NetscriptFunctions/UserInterface.ts
```diff
@@ -74,6 +74,30 @@ export function NetscriptUserInterface(): InternalAPI<IUserInterface> {
         LogBoxCloserEvents.emit(pid);
       },
 
+    minimizeTail:
```
d0sboots: Optional: These two functions do the same thing. Something like `setTailMinimized` that takes a boolean might be better.

src/NetscriptFunctions/UserInterface.ts
```diff
@@ -74,6 +74,30 @@ export function NetscriptUserInterface(): InternalAPI<IUserInterface> {
...
+      (ctx) =>
+      (_pid = ctx.workerScript.scriptRef.pid) => {
+        const pid = helpers.number(ctx, "pid", _pid);
+        const runningScriptObj = helpers.getRunningScript(ctx, pid);
+        if (runningScriptObj == null) {
+          helpers.log(ctx, () => helpers.getCannotFindRunningScriptErrorMessage(pid));
+          return;
+        }
+        LogBoxWindowMinimizationEvents.emit(pid, true);
```
d0sboots: There's no need for this event emitter. With the RunningScript you can call `runningScriptObj.tailProps?.setMinimized`

src/ui/React/LogBoxManager.tsx
```diff
@@ -258,7 +281,7 @@ function LogWindow({ hidden, script, onClose }: LogWindowProps): React.ReactElem
   }
 
   function minimize(): void {
-    setMinimized(!minimized);
+    toggleMinimized(!minimized);
```
d0sboots: Without the event emitter, you can have the impl here and no need for useCallback.

PR #2561, opened 2026-03-08
MISC: Improve help text of expr command

catloversg:
A player on Steam asked why `expr 1/2 +2` was `0.045454545454545456`, but `expr 1/2 + 2` was `2.5`.

New help text:
```
Usage: expr [mathematical expression]
 
Evaluate a simple mathematical expression. Supports native JavaScript operators:
 
+, -, /, *, **, %
 
Example:
 
    expr 25*2
    expr "25 * 2 ** 10"
 
Except for the dot ("."), parentheses ("(" and ")"), and the letters "e" and "E" (used for scientific notation), 
most non-digit characters are not allowed and will be removed from the input.
If the input contains whitespace, it must be enclosed in single quotes ('), double quotes ("), or backticks (`).
```


d0sboots:
I read the new help text, but it did not provide the answer to the original question at all. Then I calculated what the answer was (1/22), and I began to get an inkling of what must have occurred. It was only when I looked at the source that I fully understood. 

As far as I'm concerned, this is just a bug in how we handle terminal parsing. The parser should only be dealing with strings, not doing any kind of type conversion. 

For practical and compatibility reasons, I understand not fixing this bug any time soon, but I'm not sure the changed help text actually helps (especially when 98% of the time you *don't* need quotes.)

catloversg:
> If the input contains whitespace, it must be enclosed in single quotes ('), double quotes ("), or backticks (`).

This part was added to help with that case. For example, `expr "1/2 +2"` returns the expected result. In many situations, quotes are not required, but using quotes does not cause problems and helps avoid this kind of issue. However, the current help text does not mention the use of quotes.

src/Terminal/HelpText.ts
```diff
@@ -218,15 +218,18 @@ export const HelpTexts: Record<string, string[]> = {
...
     "Example:",
     " ",
-    "    expr 25 * 2 ** 10",
+    "    expr 25*2",
+    `    expr "25 * 2 ** 10"`,
     " ",
-    "Note that letters (non-digits) are not allowed and will be removed from the input.",
+    `Except for the dot ("."), parentheses ("(" and ")"), and the letters "e" and "E" (used for scientific notation), `,
+    `most non-digit characters are not allowed and will be removed from the input.`,
+    `If the input contains whitespace, it must be enclosed in single quotes ('), double quotes ("), or backticks (\`).`,
```
d0sboots: I think the issue I'm having with this wording is the "must". As I said, in most cases it is perfectly fine to leave out the quotes (and indeed, AFAICT that was the original intent of the command). Also, implying that whitespace is the issue seems wrong, since if you include whitespace between *every* set of tokens, you will always be fine.
```suggestion
    "Some edge cases may need to be enclosed in quotes to evaluate correctly.",
```

PR #2562, opened 2026-03-09
BUGFIX: Intelligence skill is not overridden in some calculations

catloversg:
How to reproduce this bug: Set int level to 10k, override it with 1 when bitfluming, then do hacking work for a faction and observe the rep gain.

Tested: Manual testing.

### DIFF
--- a/src/Bladeburner/Actions/Action.ts
+++ b/src/Bladeburner/Actions/Action.ts
@@ -172,7 +172,7 @@ export abstract class ActionClass {
     for (const stat of getRecordKeys(person.skills)) {
       competence += this.weights[stat] * Math.pow(inst.getEffectiveSkillLevel(person, stat), this.decays[stat]);
     }
-    competence *= calculateIntelligenceBonus(person.skills.intelligence, 0.75);
+    competence *= calculateIntelligenceBonus(person, 0.75);
     competence *= inst.calculateStaminaPenalty();
 
     competence *= this.getTeamSuccessBonus(inst);
--- a/src/Bladeburner/Bladeburner.ts
+++ b/src/Bladeburner/Bladeburner.ts
@@ -57,6 +57,7 @@ import { loadActionIdentifier } from "./utils/loadActionIdentifier";
 import { pluralize } from "../utils/I18nUtils";
 import { calculateActionRankGain, calculateActionReputationGain } from "./Formulas";
 import { processWorkStats } from "../Work/Formulas";
+import { getEffectiveIntelligence } from "../PersonObjects/formulas/intelligence";
 
 export const BladeburnerPromise: PromisePair<number> = { promise: null, resolve: null };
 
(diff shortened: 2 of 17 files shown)


catloversg:
The Jest tests in `TeamCasualties.test.ts` failed with this error:
```
    TypeError: Class extends value undefined is not a constructor or null

      15 | };
      16 |
    > 17 | export abstract class LevelableActionClass extends ActionClass {
         |                                                    ^
      18 |   // Static info, not included in save
      19 |   difficultyFac = 1.01;
      20 |   rewardFac = 1.02;

      at Object.ActionClass (src/Bladeburner/Actions/LevelableAction.ts:17:52)
      at Object.require (src/Bladeburner/Actions/Contract.ts:6:1)
      at Object.require (src/Bladeburner/data/Contracts.ts:2:1)
      at Object.require (src/Bladeburner/Bladeburner.ts:43:1)
      at Object.require (src/NetscriptFunctions/Bladeburner.ts:13:1)
      at Object.require (src/NetscriptFunctions.ts:71:1)
      at Object.require (src/NetscriptWorker.ts:12:1)
      at Object.require (src/NetscriptPort.ts:3:1)
      at Object.require (src/Netscript/NetscriptHelpers.tsx:29:1)
      at Object.require (src/DarkNet/effects/offlineServerHandling.ts:4:1)
      at Object.require (src/DarkNet/effects/ramblock.ts:4:1)
      at Object.require (src/DarkNet/models/DarknetServerOptions.ts:16:1)
      at Object.require (src/DarkNet/controllers/ServerGenerator.ts:1:1)
      at Object.require (src/DarkNet/controllers/NetworkMovement.ts:9:1)
      at Object.require (src/DarkNet/effects/effects.ts:13:1)
      at Object.require (src/DarkNet/effects/labyrinth.ts:3:1)
      at Object.require (src/DarkNet/utils/darknetNetworkUtils.ts:5:1)
      at Object.require (src/DarkNet/models/DarknetState.ts:7:1)
      at Object.require (src/DarkNet/effects/webstorm.ts:1:1)
      at Object.require (src/Programs/Programs.ts:17:1)
```
This happens if anything in `src\Bladeburner\Actions\Action.ts` tries to import `PlayerObject` from `src\PersonObjects\Player\PlayerObject.ts`. I'm not entirely happy with the mitigation in the latest commit, but it's the simplest acceptable fix I can come up with right now. I'm open to other suggestions.

d0sboots:
Is there any possible way we can override `Player.skills.intelligence` directly? Either by setting it, or making it a getter/setter function pair?

The issue I have with this approach is that it's going to be very easy for future changes to mess up and reintroduce the problem. The only effective long-term fix I can think of is to override it on the variable that everyone will use naturally.

catloversg:
That's doable, but I have to rework how int skill is saved, loaded and overridden. I'll open a PoC PR for that so we can compare two approaches.

PR #2565, opened 2026-03-11
UI: Add option to autosave scripts on focus change

catloversg:
This is an experimental UI feature. Despite having some caveats, I think it's good enough and helps players avoid a very common issue.

When players share their scripts but the issues they report do not match the behavior indicated by those scripts, it is usually due to one of the following reasons:
- They modify their scripts but they do not save them before running them.[1]
- They modify the scripts on server X, but run scripts on server Y.[2]
- They modify the scripts on server X, scp them to server Y, but their scp code is flawed, so it does not behave as they expect.[3]

We cannot help them in cases [2] and [3], but we can help them avoid [1]. Before making this PR, I tested a PoC that displayed a warning toast when players ran an unsaved script. It worked, but I think an autosave feature would work better.

Caveats:
- On Windows, right-clicking is treated as a focus change. https://github.com/microsoft/monaco-editor/issues/3525
- Clicking on the "Run" button saves the scripts and make the game save twice. I can work around this issue, but the workaround looks a bit too convoluted.
- If the game lags significantly, pressing the "Re-run script" button on the tail window may run the script before the script has been saved.

Demo:

https://github.com/user-attachments/assets/951160f9-f097-4a95-947e-0f10e4f50aa7

### DIFF
--- a/src/Settings/Settings.ts
+++ b/src/Settings/Settings.ts
@@ -275,5 +275,10 @@ export const Settings = {
 
     // Set up initial state for error modal suppression
     toggleSuppressErrorModals(Settings.SuppressErrorModals, true);
+
+    // Disable this feature for existing save files.
+    if (save.MonacoAutoSaveOnFocusChange === undefined) {
+      Settings.MonacoAutoSaveOnFocusChange = false;
+    }
   },
 };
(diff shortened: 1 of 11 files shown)


src/Settings/Settings.ts
```diff
@@ -188,6 +188,8 @@ export const Settings = {
   MonacoStickyScroll: { enabled: false } as StickyScroll,
   /** Whether to show minimap in the script editor */
   MonacoMinimap: { enabled: true } as Minimap,
+  /** Whether to autosave on focus change */
+  MonacoAutoSaveOnFocusChange: true,
```
d0sboots, line 192: I'm OK with this being default-true for new games. However, I'm not so sure about it defaulting to true for existing games. How much effort would it be to split those cases?

PR #2566, opened 2026-03-11
UI: Change description of some script editor options

catloversg:
Some boolean options are prefixed with "Enable", while others are not. We should be consistent.

Capture


d0sboots:
I don't think I agree with the premise of this PR. The format of all the boolean options is a verb phrase in <https://en.wikipedia.org/wiki/Imperative_mood>. In this case, it's not functioning as a command, but rather as a shorthand query: "(Should I) do X?" So where a construction like "Enable Minimap? y/n" makes sense, "Minimap? y/n" does not make as much sense.

catloversg:
That sounds reasonable.

PR #2567, opened 2026-03-11
API: Improve error messages for invalid sleeve numbers

catloversg:
If players pass an invalid sleeve number that is still within the range [0, sleeveCount) (e.g., 0.5), they will see this error message:
```
TypeError: Cannot read properties of undefined (reading 'hp')
    at eval (Sleeve.ts:184:32)
    at Proxy.wrappedFunction (APIWrapper.ts:68:16)
    at main (a.js:4:25)
    at startNetscript2Script (NetscriptWorker.ts:88:9)
```
The alternative fix is to change `checkSleeveNumber`.


d0sboots:
If we were going to change checkSleeveNumber, IMO it should take `unknown` and do the `helpers` call internally.
This is fine.

PR #888, opened 2023-10-23
MISC: Add "Rumors" interface to Faction discovery

jjclark1982:
### Summary of changes

- UI: Add a "Rumors" section to the Faction overview screen, with hints about factions the player may be able to join soon. Rumors can be encountered in Literatures, Messages, and other game events.
- UI: Add tooltips to the Faction overview screen with checklists of invite requirements for "known" factions. A faction is known if you have ever been invited to it.
- API: Add `ns.singularity.getFactionInviteRequirements()`

### Implementation details
- Player, Literature, and Message have a new field `factionRumors`.
- Faction has a new field `discovery` ("unknown", "rumored", "known"), with controls in the Dev Menu.
- CompanyMetadata `hasFaction` field renamed to `relatedFaction` (the field was unused before and is still unused, but was useful in developing rumors).
- `<Requirement>` component from augmentation prereq tooltip moved to `ui/Components/`.

### PR Description

This adds a **Rumors** section to the Faction overview screen, with hints about factions the player may be able to join soon. Much of this information was only available in [an old readthedocs page](https://bitburner.readthedocs.io/en/latest/basicgameplay/factions.html) and did not have any in-game discovery method.

A rumor appears when:
- a faction is mentioned in a literature (when it is displayed, or when it is copied to home for factions with no other rumor trigger)
- a faction is mentioned in a message
- you fail an infiltration (Shadows of Anarchy)
- you were in a faction before installing augmentations (or invited, but not banned)
- you satisfy partial requirements for invitation

Faction Relationship

Each faction has a persistent **discovery** property added to the save file.

- `unknown`: Faction has never been encountered by this player.
- `rumored`: Player has heard rumors of this faction and may be invited soon. The faction's name appears in corrupted text on the Factions screen, with rumor text hinting at hinting at how to be invited.
- `known`: Player has been invited to join this faction at least once in any bitnode. The faction's invite requirements appear in a tooltip.

Faction Discovery

Each faction has new flavor text written for their rumor, with subtle hints about the requirements to join. For example:

> **Silhouette**: Corporate executives with the right moral flexiblity may be invited to find out who they are truly working for. (received when becoming an executive, before satisfying wealth and karma reqs)

> **Tetrads**: A notorious East Asian criminal organization. (received when in East Asia with enough karma but not enough stats)

> **Bladeburners**: NSA would like to have a word with you once you're ready. (received when entering bitnode 6 or 7)

Screenshot 2023-10-23 at 12 10 20 PM

[...]

PR #2569, opened 2026-03-12
MISC: Rework faction rumor

### LINKED ISSUE #2507: Feature request: Rumored factions with enemies
I appreciate how, at the beginning of a new run, the "Rumors" section lists all (?) known (previously joined) factions, with their entrance requirements. However, this doesn't seem to work with faction enemies: joining a faction will remove its enemies from the rumor list, even at the beginning of the next run.

Could this be revised / enhanced to list all known (previously joined) factions regardless?

Thank you.

catloversg:
The rumor feature was added in #888. We discussed the following behaviors on Discord:
- After joining a faction, the rumors of its enemy factions are removed.
- You get rumors of factions that you were in or invited to in the immediately previous reset.

Suggestions:
- Instead of removing rumors of enemy factions, we mark that faction rumor as an "inaccessible" one.
- Make rumors persistent across BNs.

This PR makes these changes:
- Never reset `Player.factionRumors`.
- Never delete items in `Player.factionRumors`. When showing the rumors in the UI, we filter out those belonging to factions that the player is in or invited.
- Allow receiving rumors of banned factions.
- Show the "Factions" tab if there is at least one rumor.

"Inaccessible" rumors:

Capture

Close #2507.

### DIFF
--- a/src/Faction/ui/FactionsRoot.tsx
+++ b/src/Faction/ui/FactionsRoot.tsx
@@ -85,11 +85,9 @@ const FactionElement = (props: FactionElementProps): React.ReactElement => {
     Router.toPage(Page.FactionAugmentations, { faction });
   }
 
-  function acceptInvitation(event: React.MouseEvent<HTMLButtonElement>, faction: Faction): void {
-    if (!event.isTrusted || !faction.alreadyInvited || faction.isBanned) {
-      return;
-    }
-    joinFaction(faction);
+  function acceptInvitation(event: React.MouseEvent<HTMLButtonElement>, faction: FactionName): void {
+    if (!event.isTrusted) return;
+    joinFaction(Factions[faction]);
     props.rerender();
   }
 
@@ -118,7 +116,7 @@ const FactionElement = (props: FactionElementProps): React.ReactElement => {
             <Button onClick={() => openFactionAugPage(props.faction)}>Augments</Button>
           </Box>
         ) : props.faction.alreadyInvited ? (
-          <Button sx={{ height: "48px", mr: 1 }} onClick={(e) => acceptInvitation(e, props.faction)}>
+          <Button sx={{ height: "48px", mr: 1 }} onClick={(e) => acceptInvitation(e, props.faction.name)}>
             Join!
           </Button>
         ) : null}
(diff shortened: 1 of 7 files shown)


src/Faction/ui/FactionsRoot.tsx
```diff
@@ -130,13 +130,29 @@ const FactionElement = (props: FactionElementProps): React.ReactElement => {
                   </>
                 }
               >
-                <span style={{ overflow: "hidden", whiteSpace: "nowrap", textOverflow: "ellipsis" }}>
+                <span
```
d0sboots: No action needed: This is awkward because, by lines the two branches share more code than they differ, but the structure makes it hard to deduplicate cleanly.

src/PersonObjects/Player/PlayerObjectGeneralMethods.ts
```diff
@@ -141,6 +140,7 @@ export function prestigeAugmentation(this: PlayerObject): void {
 export function prestigeSourceFile(this: PlayerObject): void {
   this.entropy = 0;
   this.prestigeAugmentation();
+  this.factionRumors.clear();
```
d0sboots: I'm not certain about this. This is definitely an improvement on how things were, but does it make sense to clear rumors between SFs? Especially if you are clearing the same BN back-to-back? But even when going to other BNs, the requirements for factions don't really change.

catloversg: I'm neutral on this. The rumor system has always been a bit weird. Each faction has a `discovery` property ("unknown", "rumored", "known") that is not reset on prestige. This implies that once you hear a rumor about a faction or "know" it (by receiving an invite or joining it), you never forget it. However, the faction rumor list is reset on prestige.

I guess never resetting the rumor list somewhat makes sense, according to the design of the `discovery` property. Note that this makes the rumor list very long when you enter a BN. Depending on one's perspective, this could be considered either a nice feature (you always have info on how to join factions) or an annoying bug (BN prestige resets almost everything - except for int and SF).

I did that in the latest commit. Let's see what people think when we release this change.

catloversg:
I reverted the last commit. It requires an additional UI change, so it will be handled in a separate PR.

PR #2575, opened 2026-03-16
MISC: Rework intelligence override

catloversg:
Context: #2562.

Note:
- I increased the version code in this PR to ensure the migration code runs only once. Otherwise, it could run multiple times and override the gained int exp. I specifically tested this scenario in the new migration tests.
- `persistentIntelligenceData.exp` can be `persistentIntelligenceExp` or something else.


src/ui/React/CharacterOverview.tsx
```diff
@@ -127,6 +123,8 @@ export function CharacterOverview({ parentOpen, save, killScripts }: OverviewPro
   const [killOpen, setKillOpen] = useState(false);
   const [hasIntelligence, setHasIntelligence] = useState(Player.skills.intelligence > 0);
   const [showBars, setShowBars] = useState(!Settings.DisableOverviewProgressBars);
+  // const [width, setWidth] = useState(145);
```
d0sboots: Commented code

src/Constants.ts
```diff
@@ -7,7 +7,7 @@ export const CONSTANTS = {
   VersionString: "3.0.0dev",
   isDevBranch: true,
   isInTestEnvironment: globalThis.process?.env?.JEST_WORKER_ID !== undefined,
-  VersionNumber: 46,
+  VersionNumber: 47,
```
d0sboots, line 10: Bumping the version in this PR doesn't feel correct. Usually we do this as part of the release rollup.

src/PersonObjects/Person.ts
```diff
@@ -142,6 +146,21 @@ export abstract class Person implements IPerson {
     );
   }
 
+  overrideIntelligence(): void {
+    const persistentIntelligenceSkill = this.calculateSkill(this.persistentIntelligenceData.exp, 1);
+    // If there is no intelligenceOverride, reset exp/skill to persistent values
+    if (Player.bitNodeOptions.intelligenceOverride === undefined) {
```
d0sboots: I think this should be moved to an "else" of the other if. Here's a bug scenario: 

persistentInt is something large like 1000. IntOverride was 20, but then it is changed to 2000. Since it is not less than persistentInt (but not undefined), int remains unchanged at 20.

src/ui/React/CharacterOverview.tsx
```diff
@@ -71,15 +71,11 @@ export function Val({ name, color }: ValProps): React.ReactElement {
...
 
-  if (
-    name === "Int" &&
-    Player.bitNodeOptions.intelligenceOverride !== undefined &&
-    Player.bitNodeOptions.intelligenceOverride < Player.skills.intelligence
-  ) {
+  if (name === "Int" && Player.bitNodeOptions.intelligenceOverride !== undefined) {
     return (
-      <Tooltip title={`Intelligence: ${formatSkill(Player.skills.intelligence)}`}>
+      <Tooltip title={`Intelligence: ${formatSkill(Player.calculateSkill(Player.persistentIntelligenceData.exp, 1))}`}>
```
d0sboots: Optional: Maybe something like "Saved Intelligence" to indicate that it is not currently present? I don't have a great suggestion that is both short and descriptive.

d0sboots (commented):
Although this has a bit more up-front work, I think it will be more long-term viable. I especially like the tests. :)

src/ui/React/CharacterOverview.tsx
catloversg: I decided to call it "Persistent Intelligence". In the BN description, Intelligence is described as "Intelligence is unique because it is permanent and persistent (it never gets reset back to 1).", so I think using the adjective "persistent" here is appropriate.

src/Constants.ts
catloversg: I did that to prevent the migration code from running more than once. A bug scenerio (tested by migration tests):
- Load v2.8.1 with int skill = 100.
- Migration code runs. Persistent skill is set to 100.
- Farm int to skill = 200. Save (version code is still 46).
- Reload. Migration code runs again. Persistent skill is set to 100 again.

Technically, this only affects people running custom builds like us (normal players will run a build with version code set to 47 once we release), but I think setting the version code to 47 now is safer.

The alternative would be to implement "idempotent" migration code, but I could not find a good way to do that.

Another option is to bump the version code in a separate PR and merge it right after this one.

PR #2576, opened 2026-03-16
DEPS: Update dependencies

catloversg:
The number of warnings reported by `npm audit` is starting to get quite large. I fixed most of them in this PR.

Commits with "manually" in the message were created by specifying the version in `devDependencies` and then running `npm i`. In the end, resolving the overridden versions of `minimatch` and `ajv` manually became too time-consuming, so I used `npm audit fix` for those.

The two remaining problematic deps are `@tootallnate/once` and `dompurify`.

`@tootallnate/once` is from `jest-environment-jsdom`.
```
`-- jest-environment-jsdom@29.7.0
  `-- jsdom@20.0.3
    `-- http-proxy-agent@5.0.0
      `-- @tootallnate/once@2.0.0
```
I'm not sure why many dependencies of `jest-environment-jsdom` are being overridden. It happened in #2246. I'll open a separate PR for this issue later.

Monaco specifies a fixed version of `dompurify`, so we'll need to wait for it to be updated upstream.

Tested manually with UI.


d0sboots:
Since the doc check passes, I guess that means API documenter didn't change in a meaningful way.

PR #2579, opened 2026-03-17
DEPS: Update Jest

catloversg:
This PR updates Jest and its related deps to v30, which is the latest major version. Jest v29 is fairly old, and its deps have to use other outdated deps. For example, jest-environment-jsdom v29 depends on jsdom v20, which is why many of its deps end up being overridden.

While testing this PR, I encountered multiple issues caused by babel-plugin-transform-barrels. This plugin was added a long time ago to improve Jest test performance. However, in #2246, I noticed that after upgrading Jest-related dependencies, some tests failed on GitHub CI while still passing on my machine. After more testing now, I found even more issues.

I tested many combinations of environments (my machines and GitHub CI), babel-plugin-transform-barrels versions, and patches (both in Bitburner and babel-plugin-transform-barrels). Unfortunately, I did not document every combination, as I did not expect the situation to become this complex. Below are some notable cases I recorded:
- Plugin v1.0.17: Work on GitHub CI, but not work on my Windows machine.[1]
- Plugin v1.0.23 with barrel imports removed in Bladeburner code: Work on my Windows machine, but not work on GitHub CI.[2]

Log [1]:
```
 FAIL  test/jest/Netscript/Darknet.test.ts
  ● Test suite failed to run

    TypeError: C:\Games\BitburnerDeps\src\Bladeburner\utils\terminalShorthands.ts: The "path" argument must be of type string. Received null

      15 |   });
      16 |
    > 17 |   return babelTransformer.process(sourceText, sourcePath, options);
         |                           ^
      18 | }
      19 |

      at PackageManager.getMainPackageOfModule (node_modules/babel-plugin-transform-barrels/src/packages.js:176:32)
      at ImportSpecifier.get cjsPath [as cjsPath] (node_modules/babel-plugin-transform-barrels/src/barrel.js:439:39)
      at ImportSpecifier.get path [as path] (node_modules/babel-plugin-transform-barrels/src/barrel.js:461:21)
      at importDeclarationVisitor (node_modules/babel-plugin-transform-barrels/src/main.js:26:26)
      at PluginPass.<anonymous> (node_modules/babel-plugin-transform-barrels/src/main.js:80:7)
      at call (node_modules/@babel/traverse/src/visitors.ts:303:14)
      at NodePath.call [as _call] (node_modules/@babel/traverse/src/path/context.ts:36:20)
      at NodePath.call (node_modules/@babel/traverse/src/path/context.ts:21:18)
[...]

PR #2580, opened 2026-03-17
CODEBASE: Remove barrel imports in Bladeburner code

catloversg:
Context: #2579.

This change is not mandatory, but it should be safe to merge. Bladeburner is the only place that we use barrel imports, and the benefit is questionable.


d0sboots:
I believe it's an older style that is generally considered poor form now. For instance, I know MUI recommends against using their barrel imports, for test-performance reasons.

PR #2582, opened 2026-03-17
API: Reduce RAM cost of inGang and inBladeburner APIs

catloversg:
The original suggestion on Discord was to make the RAM cost of `inGang`, `inBladeburner`, and `hasCorporation` consistent. It doesn’t seem particularly important to me, but I’m fine with making the change.

We can either increase the cost of `hasCorporation` or reduce the cost of the other two. I'm fine either way.


d0sboots:
Seems good to me. You can always check by trying a method you would use anyway and catching the error, making the cost of the "has API" check free, but gross.

PR #2583, opened 2026-03-19
MISC: Support importing Steam Cloud save file manually

catloversg:
Players may want to do this if they have accumulated many Steam Cloud save files after using the "Delete Steam Cloud Data" option multiple times. It is also useful for testing.

Tested manually:
- Options => Import Game
- File => Load From File

### DIFF
--- a/electron/saveDataBinaryFormat.js
+++ b/electron/saveDataBinaryFormat.js
@@ -1,13 +1,68 @@
 // The 2 magic bytes of the gzip header plus the mandatory compression type of DEFLATE
-const magicBytes = [0x1f, 0x8b, 0x08];
+const magicBytesOfDeflateGzip = new Uint8Array([0x1f, 0x8b, 0x08]);
+// Base64-encoded string of magicBytesOfDeflateGzip
+const base64EncodingOfMagicBytes = encodeBytesToBase64String(magicBytesOfDeflateGzip);
+// Convert the base64-encoded string to a byte array
+const byteArrayOfBase64EncodingOfMagicBytes = Uint8Array.from(base64EncodingOfMagicBytes, (c) => c.charCodeAt(0));
 
+/**
+ * @param {Uint8Array} bytes
+ * @returns {string}
+ */
+function encodeBytesToBase64String(bytes) {
+  let binaryString = "";
+  for (let i = 0; i < bytes.length; i++) {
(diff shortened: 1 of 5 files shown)


electron/saveDataBinaryFormat.js
```diff
@@ -1,13 +1,68 @@
...
+// Base64-encoded string of magicBytesOfDeflateGzip
+const base64EncodingOfMagicBytes = encodeBytesToBase64String(magicBytesOfDeflateGzip);
+// Convert the base64-encoded string to a byte array
+const byteArrayOfBase64EncodingOfMagicBytes = Uint8Array.from(base64EncodingOfMagicBytes, (c) => c.charCodeAt(0));
 
+/**
+ * @param {Uint8Array} bytes
+ * @returns {string}
+ */
+function encodeBytesToBase64String(bytes) {
```
d0sboots: This function is unneeded. Use `Uint8Array.toBase64` and rely on webpack polyfilling it through coreJS until our browserlist config says it has wide enough support.

(There is some slight complication since this is used on the electron side as well as the browser. However, looking at `navigator.userAgent` on 3.0 gives `Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) bitburner/3.0.0 Chrome/140.0.7339.240 Electron/38.4.0 Safari/537.36`, which is *just* recent enough to support this natively.)

electron/saveDataBinaryFormat.js
```diff
@@ -1,13 +1,68 @@
...
+  let base64String = "";
+  for (let i = 0; i < bytes.length; i++) {
+    base64String += String.fromCharCode(bytes[i]);
+  }
+  const decodedBinaryString = atob(base64String);
+  const result = new Uint8Array(decodedBinaryString.length);
+  for (let i = 0; i < decodedBinaryString.length; i++) {
+    result[i] = decodedBinaryString.charCodeAt(i);
+  }
+  return result;
```
d0sboots: TextDecoder (with one-byte processing) should be faster than a manual loop. It's definitely shorter.
```suggestion
  return Uint8Array.fromBase64(new TextDecoder("windows-1252").decode(bytes));
```

d0sboots:
Oh yeah, Node itself will need the polyfill, that's fun.

PR #2586, opened 2026-03-20
CODEBASE: Configure Babel to use core-js polyfills

catloversg:
Context: https://github.com/bitburner-official/bitburner-src/pull/2583#discussion_r2963758755

By default, Babel only transpiles syntax. It needs to be configured to provide polyfills.

Webpack bundle analyzer report: 
[report.html](https://github.com/user-attachments/files/26144565/report.html)

Build log:
[build-log.txt](https://github.com/user-attachments/files/26144615/build-log.txt)

Tested: Add `console.log(new Uint8Array([0x1f, 0x8b, 0x08]).toBase64());` to `electron\saveDataBinaryFormat.js`. Tested on Chromium 130 via Electron 33.3.1.

Edit: This PR is not needed now due to https://github.com/bitburner-official/bitburner-src/pull/2586#issuecomment-4103256305 and https://github.com/bitburner-official/bitburner-src/pull/2583/commits/703605c7912b9c957ea2dcf49c1df55b29dd8d7a.

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -1,3 +1,23 @@
 module.exports = {
-  presets: ["@babel/preset-react", "@babel/preset-env", "@babel/preset-typescript"],
+  // The default value of sourceType is "module", so we can omit this config. We can also specify "unambiguous" to let
+  // Babel infer the source type and avoid explicitly setting it for saveDataBinaryFormat.js. However, being explicit is
+  // preferable, since we know all of our modules use ES modules except for those in the Electron folder.
+  sourceType: "module",
+  presets: [
+    "@babel/preset-react",
+    [
+      "@babel/preset-env",
+      {
+        useBuiltIns: "usage",
+        corejs: "3.49",
+      },
+    ],
+    "@babel/preset-typescript",
+  ],
+  overrides: [
+    {
+      test: ["./electron/saveDataBinaryFormat.js"],
+      sourceType: "commonjs",
+    },
+  ],
 };
--- a/package-lock.json
+++ b/package-lock.json
@@ -29,6 +29,7 @@
         "bcryptjs": "^2.4.3",
         "clsx": "^1.2.1",
         "convert-source-map": "^2.0.0",
+        "core-js": "^3.49.0",
         "date-fns": "^2.30.0",
         "fast-dice-coefficient": "1.0.3",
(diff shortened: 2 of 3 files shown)


d0sboots:
Something is not working right with the polyfilling. According to the log, it is polyfilling array.push and array.includes, which are ancient, and the reasons it's giving don't like up with caniuse.

catloversg:
`core-js` polyfills a feature if it determines that a browser does not support it properly, even when the browser claims otherwise. For example: https://github.com/zloirock/core-js/issues/480

With array.push, it polyfills for `{ "opera_mobile":"80", "samsung":"22" }`. I assume these browsers have some bugs in these versions.

The last time I tested on the production build, the polyfill is only applied if the browser needs it. For example, `Uint8Array.toBase64` still calls the native function on Chrome 140. The polyfill code only runs in Chromium 130.

Edit: I tested it again by running the web version on my current Chrome and an old Electron wrapper.
- Chrome 146:
  - array.push: native code
  - Uint8Array.toBase64: native code
- Chromium 130:
  - array.push: native code
  - Uint8Array.toBase64: polyfill

Edit: There are two cases in which array.push is polyfilled: https://github.com/zloirock/core-js/blob/a7d544f246de459fd898d9bc79350132026e4088/packages/core-js/modules/es.array.push.js

d0sboots:
I understand that the polyfills are only used if they are needed. The issue is that they add to size (not much) and load time to both load and determine they are not needed, so it is better to just not have them when the feature is widely available. 

Digging deep (thanks to your link), it turns out array.includes was very recently demoted: https://github.com/zloirock/core-js/commit/0ad3e0035b87ac941ccadc643397bb7825d4e694
This represents a maintenance direction that I rather disagree with, and makes me question if we should actually be polyfilling like this at all.

Looking carefully through the build log, I've audited most of the functions it wants to add polyfills for. There reason in all cases boils down to edge cases that we just didn't care about (like wrong error type thrown when early exiting in custom iterators). Since these are all for iOS, and many for common functions, I worry about performance regressions from falling back to polyfilled versions of a lot of stuff. 

In conclusion, I think I've come around to just sticking with what you originally had, since it (weirdly) has less concern about invisible changes and a lot less overhead dealing with stuff I thought already was dealt with.

PR #2588, opened 2026-03-20
WORKFLOW: Update NodeJS used in CI workflows to v25

catloversg:
The test failed in #2583 because Node24 does not support `Uint8Array.toBase64`. Setting up Jest to polyfill it properly in the test environment is not worth the effort.

To keep the changes minimal, I only update NodeJS in CI workflows. I'll update it in other workflows if requested.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,10 +16,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - name: Use Node.js 24
+      - name: Use Node.js 25
         uses: actions/setup-node@v4
         with:
-          node-version: 24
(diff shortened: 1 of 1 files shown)


d0sboots:
Wouldn't this mean we'd need a version requirement of v25 for tests to work from the commandline? Given that maintenance of v25 is only planned until Jun, that doesn't seem great. https://nodejs.org/en/about/previous-releases

catloversg:
This PR is not needed now due to https://github.com/bitburner-official/bitburner-src/pull/2586#issuecomment-4103256305 and https://github.com/bitburner-official/bitburner-src/pull/2583/commits/703605c7912b9c957ea2dcf49c1df55b29dd8d7a.

PR #2590, opened 2026-03-22
ELECTRON: Fix issues in edge cases of using --export-save

catloversg:
How to reproduce this issue:
- Load Electron version. Open Dev tool and delete the database. This simulates a clean install of the Steam version.
- Run with --export-save. Nothing is exported. Check console log: `Uncaught NotFoundError: Failed to execute 'transaction' on 'IDBDatabase': One of the specified object stores was not found.`
- Run without --export-save. The game gets stuck on the loading screen.

When calling `indexedDB.open`, a new database is always created if it does not exist. The current export code does not create the "savestring" object store, so if it runs before the code in `src\db.ts`, the database will be missing the expected object store.

`createObjectStore` can only be called in `onupgradeneeded`, so if we detect a missing object store, we need to either increase the database version (to trigger `onupgradeneeded`) or delete and recreate the database. Because most IndexedDB APIs are callback-based, increasing the version number is the simpler approach.

This type of migration is usually problematic because the dev version upgrades the version number to 2, while the stable version still expects version number 1. However, since we are currently in the transition period of v3, and it is already not possible to switch from the dev version back to the stable version, this should not be an issue.

There are some other issues in `src\db.ts`, but they are not related to this PR. I'll fix them in separate PR.

Tested manually.


catloversg:
@d0sboots This is the last blocking issue before we can release the next beta version.

src/db.ts
```diff
@@ -9,18 +16,24 @@ function getDB(): Promise<IDBObjectStore> {
      * DB is called bitburnerSave
      * Object store is called savestring
      * key for the Object store is called save
-     * Version `1` is important
+     * Version `2` is important
```
d0sboots: Please add to this comment (briefly) with a version history/why we are on version 2.

electron/export.html
```diff
@@ -25,5 +25,62 @@
...
+          const objectStore = transaction.objectStore("savestring");
+          const request = objectStore.get("save");
+          request.onsuccess = () => {
+            if (request.result == null) {
+              alert("There is no save data");
+              return;
+            }
+            let isBinaryFormat;
+            if (request.result instanceof Uint8Array) {
+              // All modules in the Electron folder are CommonJS, so importing them here would be really difficult. The
```
d0sboots, line 69: I don't understand. Why can't this script tag reference an external .js that goes through our regular infrastructure?

This is extra-much because there's generally a lot of small bits of duplicated code in here, that I worry will rot over time as the main part gets changed/updated without updating this.

d0sboots (commented):
Sorry, this one slipped by somehow.

electron/export.html
catloversg: Browsers don't support cjs (no `require`, no `module.exports`), so when we load `saveDataBinaryFormat.js` in this html file via `<script>`, it will throw an error at `module.exports`. Technically, loading that file is okay, all functions would be put into the global scope before the `module.exports` is executed and crashes. Then the current code can call the `isBinaryFormat` function.

However, this only works in this specific case because we only need one function from a very small module `saveDataBinaryFormat.js`. In general, using cjs code in browsers requires a bundler.

If you really want to do that, I guess it's (barely) acceptable, but I don't recommend doing that. I agree code-rotting is a problem, but it's hard to avoid it here without the ugly hack I mentioned (loading `saveDataBinaryFormat.js` via `<script>` and pretending nothing is wrong).

Edit: I assume you mean:
```
    <div>
      <h1>Close me when operation is completed.</h1>
    </div>
    <script src="./saveDataBinaryFormat.js"></script>
    <script>
      const dbRequest = window.indexedDB.open("bitburnerSave", 1);
```
If you mean importing `isBinaryFormat` via `require` (cjs) or `import` (esm) in the same script tag, it's impossible (`saveDataBinaryFormat.js` is cjs, and `require` cannot be used in browsers).

Technically, we can convert `saveDataBinaryFormat.js` to esm and use dynamic import in cjs modules, but that forces the callers of all exported functions of `saveDataBinaryFormat.js` to be async.

d0sboots: Yes, using commonJs in browsers requires a bundler, such as webpack, which is what we use for all our other js. This is what I meant by "an external .js that goes through our regular infrastructure" - why isn't the election code being bundled?

catloversg: Our current infrastructure does not bundle Electron code (it only bundles the web app). In the Electron part of our code, it works as if we are running a NodeJs project in a non-browser environment. Using webpack to bundle Electron code (generating a `main.bundle.js` containing everything in the `electron` folder) is possible, but it complicates the current build process. It also makes prototyping changes slower (I can edit Electron js code and restart app).

To answer the question of why the election code isn't bundled, I think it's because we haven't needed to import and run Electron cjs code directly in html files in the electron folder until now. We do need it in the main TS web app, but that part is already handled by the current webpack setup.

catloversg: The best approach is to convert our Electron code to esm and eventually TS. The latest Electron supports loading esm modules, and the latest NodeJS supports TS type stripping, so this is feasible. It's on my to-do list, but it is a low priority.

d0sboots: OK, I understand why the Electron code is completely separate from our standard build process, but that still leaves us in a rough spot with this change. 

I could live with the duplicated code normally I guess, because as a rarely used function if this bitrots it's not the end of the world. However, I *really* do not like the duplication of the IDB initialization code. If we mess up there (or it goes stale), using this tool could cause permanent save corruption. 

Can you leave that part out entirely? This is supposed to be an export; if the database does not exist (or equivalently is at the wrong version), there is nothing to do. If this utility only reads and never modifies, it's much less scary.

PR #2593, opened 2026-03-22
DNET: Tweaks from player feedback

ficocelliguy:
- When zooming in and out with the mouse wheel in the darknet UI, scroll towards the cursor, rather than the center of the screen

- On higher-difficulty labs, slightly adjust the starting and ending location so they are not always touching the outside wall. This will, in some cases, require players to deal with the loops in lab paths instead of just wall-following

- still show the local surroundings of the selected script on labs that cannot be completed manually

### DIFF
--- a/src/DarkNet/effects/labyrinth.ts
+++ b/src/DarkNet/effects/labyrinth.ts
@@ -334,7 +334,7 @@ export const handleLabyrinthPassword = (
 export const getPositionInLab = (pid: number): [number, number] => {
   const offsetStartAndEnd = getLabyrinthDetails().offsetStartAndEnd;
   if (!DarknetState.labLocations[pid]) {
-    const offsetX = offsetStartAndEnd ? [0, 2, 4][Math.floor(Math.random() * 3)] : 0;
+    const offsetX = offsetStartAndEnd ? Math.floor(Math.random() * 3) * 2 : 0;
     const offsetY = offsetStartAndEnd ? [0, 2, 4][Math.floor(Math.random() * 3)] : 0;
     DarknetState.labLocations[pid] = [1 + offsetX, 1 + offsetY];
   }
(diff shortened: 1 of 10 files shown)


src/DarkNet/effects/labyrinth.ts
```diff
@@ -324,6 +331,17 @@ export const handleLabyrinthPassword = (
   };
 };
 
+export const getPositionInLab = (pid: number): [number, number] => {
+  const offsetStartAndEnd = getLabyrinthDetails().offsetStartAndEnd;
+  if (!DarknetState.labLocations[pid]) {
+    const offsetX = offsetStartAndEnd ? [0, 2, 4][Math.floor(Math.random() * 3)] : 0;
```
d0sboots:
```suggestion
    const offsetX = offsetStartAndEnd ? Math.floor(Math.random() * 3) * 2 : 0;
```
And then there's no array to be changed if we add more offsets.

src/DarkNet/effects/labyrinth.ts
```diff
@@ -350,8 +368,18 @@ const getOrdinalInput = (input: string): number[] | null => {
 
 export const getLabMaze = (): string[] => {
   if (!DarknetState.labyrinth) {
-    const { mazeWidth, mazeHeight } = getLabyrinthDetails();
+    const { mazeWidth, mazeHeight, offsetStartAndEnd } = getLabyrinthDetails();
     DarknetState.labyrinth = generateMaze(mazeWidth, mazeHeight);
+    if (offsetStartAndEnd) {
```
d0sboots: This is all duplicated. Pull out the function that generates the a random starting position; this just needs to subtract that from labyrinth.length - 1.

ficocelliguy: Done!

PR #2594, opened 2026-03-22
DNET: Remove packet capture

ficocelliguy:
- Based on player feedback, the general ns.dnet.packetCapture tool has been removed from the API and docs

- the OpenWebAccessPoint puzzle now always includes a password in the heartbleed logs generated from an auth attempt - usually the current server, but sometimes a neighboring server

- At higher difficulties, that puzzle simply injects the password into a stream of random characters without nice formatting and spaces to make the puzzle trickier


src/DarkNet/models/packetSniffing.ts
```diff
@@ -14,29 +14,24 @@ import { exceptionAlert } from "../../utils/helpers/exceptionAlert";
...
+    // For higher difficulty servers, if the current server's password was chosen rather than a nearby server,
+    // the data simply contains the password among alphanumeric noise without spaces or nice labels to make it trickier to parse
+    const data = getRandomAlphaneumericNoiseData(124 + Math.random() * 20);
+    const insertIndex = Math.floor(Math.random() * (data.length - captureServer.password.length));
+    return data.slice(0, insertIndex) + captureServer.password + data.slice(insertIndex);
+  }
+  const passwordData = ` ${captureServer.hostname}:${captureServer.password} `;
+  const randomData = getRandomData(server, 124 + Math.random() * 20);
+  const insertIndex = Math.floor(Math.random() * (randomData.length - passwordData.length));
+  return randomData.slice(0, insertIndex) + passwordData + randomData.slice(insertIndex);
```
d0sboots, line 23: I thought the easier difficulties were just going to return the password? 

Since there is the colon to help parse, I can see doing this all the time, but in that case don't duplicate the impl.

src/DarkNet/models/packetSniffing.ts
```diff
@@ -73,6 +68,15 @@ const getRandomData = (server: DarknetServer, length: number) => {
   return result;
 };
 
+const getRandomAlphaneumericNoiseData = (length: number) => {
```
d0sboots: Misspelled. Also, don't you have a version of this for generating passwords?

src/DarkNet/models/packetSniffing.ts
```diff
@@ -14,29 +14,24 @@ import { exceptionAlert } from "../../utils/helpers/exceptionAlert";
...
+    const insertIndex = Math.floor(Math.random() * (data.length - captureServer.password.length));
+    return data.slice(0, insertIndex) + captureServer.password + data.slice(insertIndex);
+  }
+  const passwordData = ` ${captureServer.hostname}:${captureServer.password} `;
+  const randomData = getRandomData(server, 124 + Math.random() * 20);
+  const insertIndex = Math.floor(Math.random() * (randomData.length - passwordData.length));
+  return randomData.slice(0, insertIndex) + passwordData + randomData.slice(insertIndex);
 };
 
 const getRandomData = (server: DarknetServer, length: number) => {
```
d0sboots, line 26: Is this still used?

src/DarkNet/models/packetSniffing.ts
```diff
@@ -73,6 +68,15 @@ const getRandomData = (server: DarknetServer, length: number) => {
...
+const getRandomAlphaneumericNoiseData = (length: number) => {
+  const chars = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789";
+  let result = "";
+  for (let i = 0; i < length; i++) {
+    result += chars.charAt(Math.floor(Math.random() * chars.length));
+  }
+  return result;
+};
+
 const getRandomCharsInPassword = (password: string) => {
```
d0sboots, line 60: No action: The fact that most of this file remains indicates that this file was probably misnamed XD

src/DarkNet/models/packetSniffing.ts
catloversg: We implemented this function in at least 3 places (search for `ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789`). I think you should reuse the function in `src\utils\helpers\createRandomString.ts` or `src\utils\StringHelperFunctions.ts`. Either is fine.

I'll refactor and remove the duplicate functions later.

src/DarkNet/models/packetSniffing.ts
ficocelliguy: The server model is still called "packet capture" internally, so there's the shared label to tie them together. I can rename them if needed, though.

src/DarkNet/models/packetSniffing.ts
ficocelliguy: Yes, lower difficulty servers still use random lore data and whatnot rather than alphanumeric noise.

src/DarkNet/models/packetSniffing.ts
ficocelliguy: Yeah I do - removed

src/DarkNet/models/packetSniffing.ts
ficocelliguy: I was thinking about returning neighboring server's passwords sometimes, but it doesn't really make sense. I've simplified this accordingly.

PR #2597, opened 2026-03-23
ELECTRON: Mitigate issue of forcefullyCrashRenderer

catloversg:
When calling `loadFile` after `forcefullyCrashRenderer`, the renderer process may not spawn as expected. This issue is tracked at https://github.com/electron/electron/issues/48661.

I implemented this (ugly) workaround a while ago, but wanted to wait for an upstream fix. However, in the latest Electron version, this issue affects both Windows and Linux, so I think we should mitigate it now.

Tested with Electron v41.0.3 on Windows 10 and Debian 13.

### DIFF
--- a/electron/utils.js
+++ b/electron/utils.js
@@ -1,5 +1,5 @@
 /* eslint-disable @typescript-eslint/no-var-requires */
-const { dialog } = require("electron");
+const { app, dialog } = require("electron");
 const log = require("electron-log");
 
 const Store = require("electron-store");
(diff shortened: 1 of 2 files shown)


electron/utils.js
```diff
@@ -9,10 +9,26 @@ function reloadAndKill(window, killScripts) {
...
   window.webContents.forcefullyCrashRenderer();
-  window.loadFile("index.html", killScripts ? { query: { noScripts: true } } : {});
-  window.once("ready-to-show", () => {
-    setZoomFactor(window, zoomFactor);
-  });
+  // Delay the loading to mitigate the issue of forcefullyCrashRenderer.
+  // Check https://github.com/electron/electron/issues/48661 for more information.
+  setTimeout(() => {
+    window.loadFile("index.html", killScripts ? { query: { noScripts: true } } : {});
+    // The first delay may not work, so we try to load again after another delay.
```
d0sboots: If one could be flaky, theoretically two could be flaky as well. Why not structure this as a loop instead? That would also let you lower the timeout (maybe to 0).

PR #2599, opened 2026-03-25
ELECTRON: Import correct cloud file when multiple exist

catloversg:
When retrieving Steam cloud data, we always use the first file returned from `steamworksClient.cloud.listFiles`. However, this API can return multiple files, and their order is not guaranteed.

MRE:
- Load save id "foo" on machine 1 and use "Delete Steam Cloud Data" option to ensure this save file is synced to Steam Cloud.
- Login Steam on machine 2 and let it sync "foo" to the remote folder.
- Use "Delete Steam Cloud Data" option again on machine 1 to sync another save id "bar" to Steam Cloud.
- After a while, Steam will sync "bar" to machine 2. Calling `steamworksClient.cloud.listFiles` on machine 2 now returns both "foo" and "bar" with no guaranteed order.

The fix is to use the most recently persisted cloud file instead of the first file in that array.

Note:
- In the case of having multiple cloud files, `steamworksClient.cloud.writeFile` always fails. This issue is tracked in #2105. I'm holding off on that PR and wait for v3 release.
- There are many todo comments added in this PR. I can remove them if you want.

This PR uses a new version of my fork of steamworks.js.
- https://github.com/ceifa/steamworks.js/pull/207
- Transparency log entry: https://search.sigstore.dev/?logIndex=1133352212

### DIFF
--- a/electron/global.d.ts
+++ b/electron/global.d.ts
@@ -1,0 +1,23 @@
+import type { BrowserWindow } from "electron";
+
+declare global {
+  var steamworksError: Error | undefined;
+  var app_handlers: {
+    stopProcess: (window: BrowserWindow) => void;
+  };
+  namespace Electron {
+    interface BrowserWindow {
+      gameInfo?: {
+        player?: {
+          identifier: string;
+          playtime: number;
+          lastSave: number;
+        };
+        game?: {
+          version: string;
+          hash: string;
+        };
+      };
+    }
+  }
+}
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -55,5 +55,15 @@ module.exports = {
         "@typescript-eslint/prefer-literal-enum-member": ["off"],
       },
     },
+    /**
+     * TypeScript requires the "var" keyword within "declare global" to correctly merge variables into the global
+     * namespace.
+     */
+    {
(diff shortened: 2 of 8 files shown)


electron/global.d.ts
```diff
@@ -0,0 +1,23 @@
+import type { BrowserWindow } from "electron";
+
+declare global {
+  var steamworksError: Error | undefined;
+  var app_handlers: {
+    stopProcess: (window: BrowserWindow) => void;
+  };
+  namespace Electron {
+    interface BrowserWindow {
```
d0sboots, line 9: If you are importing the type, why do you have to redefine it here?

d0sboots (commented):
I'm fine with the TODOs, I don't want to expand the scope of this beyond what it already is.

electron/global.d.ts
catloversg: The import is used to define what `BrowserWindow` is.

This part is for declaration merging. If the engine successfully loads the save data, we send an event containing `player` and `game` info via "push-game-ready" channel. This info is then "attached" to the browser window (`window.gameInfo`). Declaration merging allows TypeScript to understand what `gameInfo` in `BrowserWindow` is.

PR #2601, opened 2026-03-26
REFACTOR: Remove duplicate random alphanumeric string functions

catloversg:
It's just as the title said.

Tested together with #2600.


src/utils/StringHelperFunctions.ts
```diff
@@ -78,15 +78,15 @@ export function containsAllStrings(arr: string[]): boolean {
 }
 
 // Generates a random alphanumeric string with N characters
-export function generateRandomString(n: number): string {
-  let str = "";
+export function getRandomAlphanumericString(length: number): string {
   const chars = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789";
+  const result = [];
```
d0sboots: Don't change the impl. It is known that JS engines optimize += well, typically better than join.

PR #2602, opened 2026-03-27
DOCUMENTATION: Fix newline issues in IPvGO docs and add missing RAM cost

catloversg:
In IPvGO docs, we usually try to display the example boards while preserving whitespace and newline as-is. However, this is not simple. We have already tried many approaches (triple backticks, `<pre>`, `<br />` + 2 whitespaces, using many newlines). Using triple backticks should be the best, but `api-documenter` breaks it, so we need to try different approaches, each with its own issues. Please check #1171 and #2350 for more information.

Nowadays, `api-documenter` emits HTML tag for tables, so we can use triple backticks. It works correctly both in GitHub and the in-game docs viewer.

@ficocelliguy In most APIs, we provide a short description of the API at the top and put the long, detailed information after `@remarks`. The upside is that this makes the table of API list look clean (e.g., `markdown\bitburner.singularity.md`). The downside is players need to click on each API to see its detail.

With IPvGO and dnet, the documentation style is opposite. Most of the details are placed before `@remarks`, making the table large. This is not necessarily a bad thing, but I want to make sure this is intentional. You don't have to change them if this is what you prefer.

### DIFF
--- a/markdown/bitburner.bladeburner.getstamina.md
+++ b/markdown/bitburner.bladeburner.getstamina.md
@@ -21,7 +21,7 @@ Array containing current stamina and max stamina.
 
 RAM cost: 4 GB
 
-Returns an array with two elements: \* \[Current stamina, Max stamina\]
+Returns an array with two elements: \[Current stamina, Max stamina\]
 
 ## Example
 
--- a/markdown/bitburner.cloud.deleteserver.md
+++ b/markdown/bitburner.cloud.deleteserver.md
@@ -56,7 +56,7 @@ True if successful, and false otherwise.
 
 ## Remarks
 
-2.25 GB
+RAM cost: 2.25 GB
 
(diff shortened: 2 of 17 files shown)


ficocelliguy:
I don't have a strong preference either way. I'm happy to change the docs to have the details under remarks if that would make things cleaner.

catloversg:
The current style of most APIs looks clean, so I like it a bit more. However, I don't have a strong opinion about it, and I know that double-checking all IPvGO and dnet APIs is a tedious (and boring) task.

It's up to you and d0sboots. If it's not worth the effort, just keep it as is.

d0sboots:
I also have a slight preference toward details under remarks, but most important is to be consistent (within a namespace). For instance, here's how IPvGO looks like on my phone rn: 
Screenshot_20260327-102914

Everything ends up very squashed because that column isn't meant for lots of text.

PR #2603, opened 2026-03-27
CCT: Generate more frequent and lower-reward coding contracts

ficocelliguy:
This PR changes CCT generation to have 1/3 the reward as before, but generate 3x more often on average, using the new reward scalar feature of CCTs.

- It adds "more CCTs", which players have asked for, without increasing the rewards from CCTs
- It keeps the existing generation cadence, but makes three rolls at generating a CCT instead of one. This means CCTs will not spawn any sooner than it does currently after starting the game.
- It lowers the rewards from each CCT, making their impact less "spiky." You are more likely to get SOME rewards of the type you want, but much less likely to get thousands and thousands of rep for a target faction all at once.
- Offline-generated CCTs are currently unchanged, but they could also be given a similar treatment if desired.

The exact numbers of this PR are, of course, open for adjustment.

### DIFF
--- a/src/CodingContract/ContractGenerator.ts
+++ b/src/CodingContract/ContractGenerator.ts
@@ -13,7 +13,7 @@ import { ContractFilePath, resolveContractFilePath } from "../Paths/ContractFile
 import { clampNumber } from "../utils/helpers/clampNumber";
 import { getRandomAlphanumericString } from "../utils/StringHelperFunctions";
 
-export function tryGeneratingRandomContract(numberOfTries: number): void {
+export function tryGeneratingRandomContract(numberOfTries: number, rewardScaling: number = 1): void {
   /**
    * We try to generate a contract every 10 minutes. 525600 is the number of tries in 10 years. There is no reason to
    * support anything above that. We tested this number (525600) on a very old machine. It took only 300-350ms to
@@ -64,12 +64,12 @@ export function tryGeneratingRandomContract(numberOfTries: number): void {
     if (random > 100 / (399 + Math.exp(0.0012 * currentNumberOfContracts))) {
       continue;
     }
-    generateRandomContract();
+    generateRandomContract(rewardScaling);
     ++currentNumberOfContracts;
   }
 }
 
-export function generateRandomContract(): void {
+export function generateRandomContract(rewardScaling: number = 1): void {
   // Choose random server
   const randServer = getRandomServer();
   if (randServer === null) {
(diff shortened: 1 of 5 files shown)


gmcew:
Making ccts a flatter curve of gains seems good! 

My only observation is that this will increase the backlog of unsolved ccts for new players without a full suite of solvers since more spawn. Since each is worth less it may reduce the incentive to work on that solver, but I hope in that instance seeing more consistent gains is enough to encouraging solving them instead. 

It may be worth looking at tweaking the SF scaling on what pool of ccts is available at the same time to ensure we don't drown new players with a large pile of ccts they don't have an answer for yet?

ficocelliguy:
It looks like the CCTs are limited to difficulty 1 if you don't have any source-files yet, which means there are only 5 distinct puzzles that can spawn. Once you know The Truth it is probably fine to have a much larger pool (~18 it looks like, with 1 source-file)? CCTs don't really get in your face - they're fairly invisible unless you're actively looking for and interacting with them. But if you have a suggestion for how that curve should change (or if we should scale down spawn chance harder if there is a bunch of CCTs sitting around already) I'd be interested.

catloversg:
This looks good to me. I don't have any suggestions for what gmcew said right now.

The comments in `tryGeneratingRandomContract` need to be updated.

Simulator:
```js
/** @param {NS} ns */
  console.clear();
  const maxAttempts = 30 * 24 * 6;
  let count = 0;
  for (let i = 0; i < 10000000; ++i) {
    if (i > maxAttempts) {
      break;
    }
    if (Math.random() < getChance(count)) {
      ++count;
    }
  }
  console.log(maxAttempts, count)
}

function getChance(currentNumberOfContracts) {
  return 100 / (399 + Math.exp(0.0012 * currentNumberOfContracts));
}
```
~`After 4320 attempts, it should generate ~3200-3300 ccts. However, according to the desmos link at https://github.com/bitburner-official/bitburner-src/pull/1657#discussion_r1769479596, it should generate only ~1000 ccts. Why is the difference so big?`~

Edit: I forgot to invert the check when moving `++count`. It should be correct now. This simulator can be used for calculating the new numbers (change `++count` to `count += 3`).

ficocelliguy:
@catloversg I think it should be `Math.random() < getChance(count)` not `Math.random() > getChance(count)` in your simulator - in the code it skips generation if the random roll is above the calculated chance

gmcew:
I'm not sure the pool _needs_ changing, I just wanted to highlight that's a potential area of feedback we've got another lever for.

CCTs often go unnoticed for a while, an intent for the SF cap was to make them not too overwhelming when they are first noticed - I think that still holds here since there's still incentive to solve the easiest pool first.

d0sboots:
I agree with this, but I think we should go farther in two ways: 
* Offline should work the same as online. 
* If the new "standard" is 1/3, we should refactor that into the base so that that the base is 1/3 as strong and the scaling value is 1.

Issue #2604, opened 2026-03-29
UI: Improve discoverability of endgame mechanics


catloversg:
For a long time, I have observed that players sometimes overlook elements in the UI. One of the most common cases is the button to purchase and upgrade Sleeves in The Covenant faction UI. Because faction UIs are usually consistent, when a button is hidden in pre-endgame and only appears in endgame, players may not notice the change and assume there is nothing new.

For some mechanics (corp, CotMG), we show clear hints in their associated pre-endgame locations (city hall, church) to indicate the presence of these mechanics. We should apply the same approach more broadly. Instead of hiding relevant UI elements, we should provide hints. This teaches players that certain locations or factions have mechanics that become available later and encourages them to revisit. From a lore perspective, this is also more engaging. 

I'll open a PR to demonstrate this UI change for Sleeves. If it's good, I'll expand it to other mechanics.

ficocelliguy:
I like this idea a lot. This could also be done with a handbook-style lit file that appears upon starting specific BNs with some basics or key hints.

PR #2605, opened 2026-03-29
UI: Show hints of Sleeves mechanic in pre-endgame

catloversg:
Context: #2604.

This PR is the foundation for UI changes that I mentioned in that issue.
All factions now have this new `Special Campaign` section:

Tooltip:

In factions that offer endgame mechanics, this section shows the relevant hints:

When clicking on the "Research" button:

Some other mechanics:

Bladeburner:
bladeburner

CotMG:
CotMG

This encourages players to check this section across all factions, as new mechanics may be unlocked later.

The text for The Covenant's campaign is bland. I have a PR that improves the lore; "The Beginning of True Immortality" is part of that PR.

### DIFF
--- a/src/Faction/ui/CovenantCampaign.tsx
+++ b/src/Faction/ui/CovenantCampaign.tsx
@@ -4,6 +4,7 @@ import React, { useState } from "react";
 import { CovenantPurchasesRoot } from "../../PersonObjects/Sleeve/ui/CovenantPurchasesRoot";
 import { Modal } from "../../ui/React/Modal";
 import { Option } from "./Option";
+import { knowAboutBitverse } from "../../BitNode/BitNodeUtils";
 
 function CovenantIncompleteCampaign() {
   const [open, setOpen] = useState(false);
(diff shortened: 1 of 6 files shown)


src/Faction/ui/CovenantCampaign.tsx
```diff
@@ -0,0 +1,44 @@
...
+  return (
+    <>
+      <Option
+        buttonText={"Research"}
+        infoText={"The Beginning of True Immortality"}
+        onClick={() => setOpen(true)}
+      ></Option>
+      <Modal open={open} onClose={() => setOpen(false)}>
+        <Typography component="div">
+          The research team does not want to reveal their goal at this moment. It's best to leave now and come back
```
d0sboots: I think this is a good start, but it's too open-ended to give a good hint as to when the player should check back. IMO this check/text should be split into two parts: One hinting towards jump3r/the truth if you're not endgame, the other hinting more directly towards a specific bitnode.

catloversg: I made minimal changes based on your feedback. Since this entire campaign will be changed in my lore PR, I only made minimal changes here.

PR #2607, opened 2026-04-02
COMMAND: Rename download to export

hexagonrecursion:
The game is so immersive that the name of the `download` command confuses me: to me "download" sounds like it would download something to my in-game "home computer".

Proposed solution: rename the command to `export`


d0sboots:
General consensus on discord is that `download` makes more sense, because people think `export` would be for exporting your savegame. Sorry.

PR #2608, opened 2026-04-02
MISC: Make implicit string conversion consistent across all coding contracts

catloversg:
When players pass a string answer to a contract that expects a non-string answer, we convert that string to the expected format. However, this conversion is inconsistent and undocumented. For example, depending on the specific implementation of `convertAnswer`, the empty string can be converted to null, an empty array, an array containing an empty string, or an array containing NaN.

This PR makes the behavior consistent, predictable, and well-documented.

Test save file (I modified the solvers from Insight's scripts):
[cct-string-conversion.gz](https://github.com/user-attachments/files/26440565/cct.gz)

Log:
```
If you pass a string to ns.codingcontract.attempt() for contracts that require a non-string answer, the
game will convert that string to the expected format. This string conversion was inconsistent and had many
undocumented behaviors. Now the rules are consistent and well-documented. Please check the "Coding Contracts"
page for more information.
There are 7 contracts that are affected by this change:
- Find All Valid Math Expressions
- Generate IP Addresses
- Largest Rectangle in a Matrix
- Merge Overlapping Intervals
- Proper 2-Coloring of a Graph
- Sanitize Parentheses in Expression
- Spiralize Matrix
Note that this change only affects the string conversion. The solution format of these contracts is an array,
so if you pass the solution, which is an array, as is, you will not have any problems.
If your code converts the array to a string, you should check these contracts, especially the "Sanitize
Parentheses in Expression" contract and others that require a string array.
- Sanitize Parentheses in Expression: Previously, if you passed an empty string to this contract, it was
converted to an array containing an empty string. Now, it's converted to an empty array.
- Read the "General rules", "String conversion", and "Tips" sections on the "Coding Contracts" page carefully.

ns.codingcontract.attempt
FindAllValidMathExpressions
GenerateIPAddresses
LargestRectangleInAMatrix
MergeOverlappingIntervals
Proper2ColoringOfAGraph
SanitizeParenthesesInExpression
SpiralizeMatrix
Find All Valid Math Expressions
Generate IP Addresses
Largest Rectangle in a Matrix
Merge Overlapping Intervals
Proper 2-Coloring of a Graph
Sanitize Parentheses in Expression
Spiralize Matrix

a.js: (Line number: 26)
lib.js: (Line numbers: 58, 147, 172, 415, 458, 576)
```

### DIFF
--- a/src/Documentation/doc/en/basic/codingcontracts.md
+++ b/src/Documentation/doc/en/basic/codingcontracts.md
@@ -40,24 +40,91 @@ The [`getContractTypes`](../../../../../markdown/bitburner.codingcontract.getcon
 
 ## Submitting Solutions
 
+### General rules
+
 Different contract problem types will require different types of solutions.
 Some may be numbers, others may be strings or arrays.
+
 If a contract asks for a specific solution format, then use that.
 Otherwise, follow these rules when submitting solutions:
 
-- String-type solutions should **not** have quotation marks surrounding the string (unless specifically asked for).
+- String-type solutions (e.g., Shortest Path in a Grid) should **not** have quotation marks surrounding the string (unless specifically asked for). For example, if your answer is `foo` (3 characters: f, o, o), just submit those 3 characters. Don't submit `"foo"` (5 characters).
   Only quotation marks that are part of the actual string solution should be included.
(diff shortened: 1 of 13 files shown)


src/Documentation/doc/en/basic/codingcontracts.md
```diff
@@ -40,24 +40,91 @@ The [`getContractTypes`](../../../../../markdown/bitburner.codingcontract.getcon
...
+ns.codingcontract.attempt(answer, "filename.cct");
+```
+
+There is no conversion!
+
+In the "General rules" section above, with array-of-strings solutions, we say `Don't use single quotes or backticks`.
+However, this code works:
+
+```js
+const firstString = "foo"; // Single quotes
```
d0sboots: You used double-quotes anyway

PR #2609, opened 2026-04-02
COMMAND: add upload (inverse of download)

### LINKED ISSUE #1522: Add "upload" terminal command
Summary: Add a terminal command `upload` which uploads file(s) from your IRL computer to the game.  This is the inverse of the `download` command.

Currently there exists a command `download` in the terminal:
```
download [script/text file]      Downloads scripts or text files to your computer
```
which downloads files from the game to your IRL computer.  If you download multiple files, they will be zipped.

However, no corresponding command exists to upload files from your IRL computer to the game.  This means that the `download` command isn't as useful as it could be, because it cannot effectively be used for:
- Transferring scripts between games
- backing up your scripts (backup without restore isn't much)
- editing your scripts externally without having to set up the API server

This issue is a suggestion to create a command `upload` which allows the player to select file(s)/folders to upload from their IRL computer to the game.  This would be the inverse of the `download` command.

Effects:
- Enables easy backup (and restore) of all your scripts and files
- Creates a mid-point between "edit in your browser" and "install third-party software" for editing scripts
- Enables easy transfer of scripts between games

Save file: any
Scripts: at least one
Steps: none
Version: v2.6.2 (633da3830)

EDIT: Apologies, I can't seem to figure out how to add a label to this - this should be an "enhancement" I think.

hexagonrecursion:
I want to upload the zip archive I got by running the download command, but:
1. I do not want to have to copy and paste files into the game one file at a time
2. I do not want to have to run an API server

closes https://github.com/bitburner-official/bitburner-src/issues/1522

Add a command that takes a destination directory as an argument, asks the user to open a zip archive and then unpacks the zip archive into the directory.

* ~I named the new command "import" instead of "upload" due to https://github.com/bitburner-official/bitburner-src/pull/2607~
* ~I will squash the commits once I am done writing the code~
* [x] manually test in Firefox
* [x] manually test in Chrome
* I did not test this PR in electron

## _File content_ encoding is assumed to be utf-8
* A zip archive contains files. The files are stored in the archive as contiguous sequences of bytes. To read a text file from the archive we have to know its encoding
* As far as I can tell, the `download` command always produces archives, where file content is encoded as utf-8, but other software may produce archives with different file encodings
* In my opinion, support for other file encodings should be out of scope for this pull request. If necessary, it can be added later

## _File name_ encoding is assumed to be utf-8
* [APPNOTE.txt](https://pkwaredownloads.blob.core.windows.net/pem/APPNOTE.txt) gives the programs that create zip archives several options for storing file names in the archive:
    1. They can encode the file name in utf-8 and set the general purpose bit 11
    2. They can use a Info-ZIP Unicode Path Extra Field to store a utf-8 encoded file name
    3. They can do neither of the above, which means that the program reading the archive will have to somehow guess the encoding. APPNOTE.txt says it is supposed to be IBM Code Page 437, but in practice it could be anything
* I would expect any program that creates zip archives containing non-ASCII filenames and wants the archives to unpack correctly on any computer regardless of language settings to use option 1 and/or option 2 because APPNOTE.txt does not define an interoperable way to communicate file name encoding unless the encoding is utf-8
* I am aware of one program that does not obey this expectation: when I used the built-in file manager on Windows 10 to create a zip archive with cyrillic file names, it created an archive with file names encoded in cp866 and it did not add a Unicode Path Extra Field. As far as I can tell by examining it in a hex editor, the archive it created does not contain any metadata to determine the encoding (not even non-standard windows-specific metadata). I would be very surprised if the same file manager can correctly unpack the archive on a machine with different language settings
* I have seen multiple programs that create zip archives store ASCII-only filenames (file names consisting of unicode codepoints U+0000 through U+007F) using option 3
* It looks like the `download` command stores ASCII-only filenames using option 3 and uses option 1 for filenames containing non-ASCII characters
* According to the documentation of JSZip, it should be able to correctly unpack archives that use option 1 or option 2. Archives that use option 3 with ASCII-only file names should unpack correctly too because ASCII can be decoded as utf-8
* In my opinion, support for file name encodings other than utf-8 should be out of scope for this pull request. If needed, it can be added later

## Directory separator is assumed to be /
* APPNOTE.txt says "All slashes MUST be forward slashes '/' as opposed to backwards slashes '\' for compatibility with Amiga and UNIX file systems etc"
* It looks like the archives created by the `download` command always use /


d0sboots:
I think `upload` is probably the better name here, especially since the other one is going to remain `download`. Other points:

* Don't worry about testing in Electron. If it works in the browser, it'll work.
* Don't squash your commits. It makes it harder to review, and we do that automatically as part of merging.

hexagonrecursion:
@d0sboots Thanks. I renamed it to "upload"

hexagonrecursion:
I consider this pull request ready for review

Sphyxis:
Why only .zip?  Why not a plain old directory with scripts in it as well?

d0sboots:
Yes, I agree that basic files should be supported as well as zip files. I don't think HTML will let you do a whole directory at a time (you have to select all the files in the directory instead), but that's fine. 

Given that this overwrites things, it would be very nice to have a confirmation UI of the file paths that will be created/overwritten with a final ok/cancel before actual action is taken.

hexagonrecursion:
Actually. Uploading a whole directory is possible. I marked this pull request as a draft because I am working on an alternative that does exactly that.

hexagonrecursion:
* Closing in favor of #2659

PR #2612, opened 2026-04-02
BUGFIX: Fix tab completion for multi-word quoted autocomplete options

### LINKED ISSUE #1014: autocomplete does not work well with multi-word data
Multi-word autocomplete entries don't work well. If you have fairly unique entries, it'll mostly work. You still have to be aware that you need to add `"` around the options you return from the `autocomplete` function, otherwise it won't autocomplete with quotes - which makes them separate arguments.

If you have options that differ only *after* the first word, then the whole thing just fails as-is - even with quotes.

For example, use this as the autocomplete, and attempt to autocomplete "nonunique second word": 
```js
export function autocomplete() {
  return ['"unique first"', '"nonunique second word"', '"nonunique sea word"'];
}
```
You'll get to `"nonunique se` and it will no-longer register any autocomplete options!

If you then add `'seriously'` to the list of options (`['"unique first"', '"nonunique second word"', '"nonunique sea word"', 'seriously']`), then it would autocomplete to `"nonunique seriously` from `"nonunique se`

The only way around this is to do a ton of parsing in the autocomplete function to try and figure out what already exists, and then only return the next parts of matching options, and as was mentioned in #522, it's rather complex to try and figure out what is already set and what isn't.

Version: v2.5.2 (c6141f2ad)

I've already written myself a wrapper to deal with this issue (and a few more), but I just wanted to bring it to light, since I couldn't find a bug that dealt with this particular part of the autocomplete.

For reference: this came up as I was trying to add autocomplete for contract types.

lstutzman:
## Summary

Fixes #1014

Tab completion breaks when a script's `autocomplete()` returns multi-word quoted options (e.g., `"nonunique second word"`). Typing `"nonunique se` + Tab produces no completions because the system only matches the last space-separated word (`se`) against the full quoted options.

## Root Cause

Two places use the regex `/[^ ]*$/` to extract the "current text" being completed:

1. `getTabCompletionPossibilities.ts:27` — extracts `currentText` for matching against options
2. `TerminalInput.tsx:259,264` — replaces the matched portion with the selected completion

This regex splits on spaces blindly and doesn't understand quotes. When mid-quote (e.g., `"nonunique se`), it extracts only `se` instead of the full quoted fragment `"nonunique se`, so the `startsWith` match against `"nonunique second word"` fails.

## Fix

- Added `extractCurrentText()` helper that detects unclosed double quotes (odd quote count) and extracts everything from the opening `"` as a single token. Falls back to the original space-split behavior when all quotes are balanced.
- Updated `TerminalInput.tsx` to use the same quote-aware logic for the replacement regex — when mid-quote, replaces from the opening `"` instead of just the last word.

## Test script

```js
// Save as autocomplete-test.js and run: run autocomplete-test.js
export function autocomplete() {
  return ['"unique first"', '"nonunique second word"', '"nonunique sea word"'];
}
  ns.tprint(ns.args);
}
```

Type `run autocomplete-test.js "nonunique se` + Tab → should show both matching options.

## Test

6 new unit tests for `extractCurrentText` covering: unquoted input, trailing space, unclosed quote, closed quotes, empty input, and single-word quote.

## Checklist

- [x] Branched from `dev`
- [x] `npm run lint` passes
- [x] `npm run format` passes
- [x] `npm run test` passes (51 suites, 4397 tests)
- [x] Test added that reproduces the bug

### DIFF
--- a/src/Terminal/getTabCompletionPossibilities.ts
+++ b/src/Terminal/getTabCompletionPossibilities.ts
@@ -17,14 +17,21 @@ import { parseUnknownError } from "../utils/ErrorHelper";
 import { DarknetServer } from "../Server/DarknetServer";
 import { CompletedProgramName } from "@enums";
 
+/** Extract the text being autocompleted, handling unclosed double quotes as a single token */
+export function extractCurrentText(terminalText: string): string {
+  const quoteCount = (terminalText.match(/"/g) || []).length;
+  if (quoteCount % 2 === 1) return terminalText.substring(terminalText.lastIndexOf('"'));
+  return /[^ ]*$/.exec(terminalText)?.[0] ?? "";
+}
+
 /** Suggest all completion possibilities for the last argument in the last command being typed
  * @param terminalText The current full text entered in the terminal
  * @param baseDir The current working directory.
  * @returns Array of possible string replacements for the current text being autocompleted.
  */
 export async function getTabCompletionPossibilities(terminalText: string, baseDir = root): Promise<string[]> {
-  // Get the current command text
-  const currentText = /[^ ]*$/.exec(terminalText)?.[0] ?? "";
+  // Get the current command text, treating unclosed quotes as a single token
+  const currentText = extractCurrentText(terminalText);
   // Remove the current text from the commands string
   const valueWithoutCurrent = terminalText.substring(0, terminalText.length - currentText.length);
   // Parse the commands string, this handles alias replacement as well.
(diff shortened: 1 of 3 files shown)


lstutzman:
Rebased onto latest dev, all tests pass. Quick note on scope:

This handles the main reported issue — multi-word quoted options from script autocomplete() breaking after the first space within the quote. The approach is a simple quote-counting heuristic: odd number of double quotes means we're mid-quote, so treat everything from the last opening `"` as the current token.

Known limitations (not regressions — upstream doesn't handle these either):
- Escaped quotes (`\"` inside strings) would throw off the count
- Single-quoted multi-word args aren't handled

Both of these are edge cases of edge cases — the core issue is the double-quote path that the reporter described, and that works.

PR #2613, opened 2026-04-02
UI: Navigate to gym/university instead of city when stopping focusing on gym/class work

### LINKED ISSUE #710: Inconsistency in "do something else simultaneously"
When working for a faction or corp, if you click "do something else simultaneously", it returns to the faction details or corp location window

When working out at a gym, or studying at a university, if you click "do something else simultaneously", it returns to the City screen for the city the player is in.

In most cases, at least for myself, when I am working out at the gym, I am trying to target specific Str/Def/Dex/Agi scores, e.g. for faction invites or to get infiltration difficulty down, so when I am working on one, I wait for it to hit the target number then immediately switch to the next training type. For this reason, it would be preferable if "do something else simultaneously" returned me to the Gym interface, rather than the City - it is fewer clicks to get to the next training type that I wish to do. 

I don't think this would be quite as important for the University studying, because I don't find myself switching between different courses, but for consistency's sake, it probably should also return the player to the University screen rather than the City screen

editing to clarify a little about the "inconsistency" claim: 
When working for a Faction or Corp, the "Do something else simultaneously" button essentially brings you back to the screen that you had to be on to get to that active task, but when studying or working out, it brings you back to the city screen for the city you were in - I am proposing that for studying/working out, it should also bring you back to the screen you had to be on to get to that task - either that gym's interface, or that university's interface

lstutzman:
## Summary

Fixes #710

"Do something else simultaneously" during gym/university work navigated to the City page, while the same button during faction/company work correctly returned to the originating page. Fixed to navigate to the specific gym or university location.

## Root Cause

In `WorkInProgressRoot.tsx`, the `isClassWork` handler for both `cancel` and `unfocus` used `Router.toPage(Page.City)`, while faction work used `Router.toPage(Page.Faction, { faction })` and company work used `Router.toPage(Page.Job)`. The ClassWork object already has a `.location` property — it just wasn't being used.

## Fix

Changed both `Router.toPage(Page.City)` calls in the `isClassWork` block to `Router.toPage(Page.Location, { location: Locations[classWork.location] })`.

2-line change.

## Demo

After the fix, clicking "Do something else simultaneously" while training at Crush Fitness Gym navigates to the Crush Fitness Gym page (showing Train Strength/Defense/Dexterity/Agility options), not the City map.

gym-fix-demo

## Checklist

- [x] Branched from `dev`
- [x] `npm run lint` passes
- [x] `npm run format` passes
- [x] `npm run test` passes (52 suites, 4395 tests)
- [x] UI change includes GIF

### DIFF
--- a/src/ui/WorkInProgressRoot.tsx
+++ b/src/ui/WorkInProgressRoot.tsx
@@ -262,10 +262,10 @@ export function WorkInProgressRoot(): React.ReactElement {
       buttons: {
         cancel: () => {
           Player.finishWork(true);
-          Router.toPage(Page.City);
+          Router.toPage(Page.Location, { location: Locations[classWork.location] });
         },
         unfocus: () => {
-          Router.toPage(Page.City);
+          Router.toPage(Page.Location, { location: Locations[classWork.location] });
           Player.stopFocusing();
         },
       },


test/jest/ui/WorkInProgressRoot.test.ts
```diff
@@ -0,0 +1,54 @@
+import { ClassWork } from "../../../src/Work/ClassWork";
```
d0sboots: These tests aren't testing anything relevant to the bug, or indeed anything particularly useful IMO, and should just be removed.

lstutzman: Fair point. I am struggling to find a way to accurately test this automated without adding additional testing deps. Manual testing proves the fix but anyway... removed the test file.

catloversg (approved):

PR #2614, opened 2026-04-02
BUGFIX: Fix tab auto-complete not correcting folder casing

### LINKED ISSUE #1708: Terminal: tab auto-complete doesn't correct folder casing

lstutzman:
## Summary

Fixes #1708

Tab completion corrects filename casing but not folder casing. Typing `run tasks/` when the real folder is `Tasks/` produces `tasks/foo.js` instead of `Tasks/foo.js`, causing files to be created in wrong-cased phantom directories.

## Root Cause

`resolveDirectory()` is purely syntactic — it resolves `../` and `./` navigation but preserves the user's casing. When the user types `tasks/`, `resolveDirectory` returns `"tasks/"` (wrong case) rather than `"Tasks/"`. The output path is then constructed as `relativeDir + filename`, inheriting the wrong-cased `relativeDir` from user input.

## Fix

After `resolveDirectory`, look up the canonical directory from `getAllDirectories()` via case-insensitive match. Correct both `baseDir` (used for matching) and `relativeDir` (used for output) with the canonical casing. The `relativeDir` correction preserves navigation prefixes (`../`, `./`) while replacing only the directory name portion.

## Test

3 new test cases in `tabCompletion.test.ts`:
- Wrong-cased folder name (`Folder1/` → `folder1/`)
- Fully uppercased folder name (`ANOTHERFOLDER/` → `anotherFolder/`)
- Wrong-cased folder with `./` prefix (`./FOLDER1/` → `./folder1/`)

## Note

PR #1929 addressed the same issue but was abandoned with failing tests and broken logic. This PR takes a different approach — correcting the casing at the source (after `resolveDirectory`) rather than modifying the matching logic in `addGeneric`.

## Checklist

- [x] Branched from `dev`
- [x] `npm run lint` passes
- [x] `npm run format` passes
- [x] `npm run test` passes (51 suites, 4392 tests)
- [x] Test added that reproduces the bug

### DIFF
--- a/src/Terminal/getTabCompletionPossibilities.ts
+++ b/src/Terminal/getTabCompletionPossibilities.ts
@@ -61,19 +61,33 @@ export async function getTabCompletionPossibilities(terminalText: string, baseDi
     if (path === null) return [];
     baseDir = path;
     // Correct folder casing: resolveDirectory is purely syntactic and preserves user's casing.
-    // Look up the canonical directory from the server's actual files.
-    // Prefer an exact case match first — only fall back to case-insensitive when no exact match
-    // exists. This handles the edge case where two directories differ only in case (e.g., both
-    // Tasks/ and tasks/ exist) — we respect the user's typed casing if it matches a real directory.
+    // Look up the canonical directory from the server's actual files. Prefer an exact case match
+    // first, falling back to case-insensitive, so existing correctly-cased paths are respected.
     if (path !== root) {
       const allDirs = getAllDirectories(Player.getCurrentServer());
       const canonical = allDirs.has(path) ? path : [...allDirs].find((d) => d.toLowerCase() === path.toLowerCase());
       if (canonical && canonical !== path) {
         baseDir = canonical;
-        // Fix relativeDir: replace the resolved directory suffix with canonical casing,
-        // preserving any navigation prefix (e.g., "../" or "./")
-        if (relativeDir.toLowerCase().endsWith(path.toLowerCase())) {
-          relativeDir = relativeDir.substring(0, relativeDir.length - path.length) + canonical;
+        // Rebuild relativeDir with canonical casing while preserving the user's navigation style.
+        // path is absolute-from-server-root; relativeDir is user-relative, so they cannot be
+        // compared by string suffix — peel the navigation prefix (leading "/" and any run of
+        // "./"/"../") off relativeDir, count the typed directory segments that follow, and
+        // replace them with the trailing N segments of the canonical path.
+        let navPrefix = "";
+        let rest = relativeDir;
+        if (rest.startsWith("/")) {
+          navPrefix = "/";
+          rest = rest.substring(1);
+        }
+        while (rest.startsWith("./") || rest.startsWith("../")) {
+          const len = rest.startsWith("../") ? 3 : 2;
+          navPrefix += rest.substring(0, len);
+          rest = rest.substring(len);
+        }
+        const typedSegmentCount = rest.match(/[^/]+\//g)?.length ?? 0;
+        if (typedSegmentCount > 0) {
+          const canonicalSegments = canonical.match(/[^/]+\//g) ?? [];
+          relativeDir = navPrefix + canonicalSegments.slice(-typedSegmentCount).join("");
         }
       }
     }
--- a/src/Terminal/getTabCompletionPossibilities.ts
+++ b/src/Terminal/getTabCompletionPossibilities.ts
@@ -60,7 +60,24 @@ export async function getTabCompletionPossibilities(terminalText: string, baseDi
     // No valid terminal inputs contain a / that does not indicate a path
     if (path === null) return [];
     baseDir = path;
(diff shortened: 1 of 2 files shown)


d0sboots:
The real tricky part here is when there are multiple directories that differ only in case. From your description, I'm not sure how robust your approach is vs that.

lstutzman:
Good catch - you're right. Bitburner's file storage is fully case-sensitive so a server can absolutely have both `Tasks/foo.js` and `tasks/bar.js`.

The original `.find()` with case-insensitive match would non-deterministically pick one based on Set insertion order, potentially "correcting" the user's already-correct casing to the wrong directory.

Updated the fix to prefer an exact case match first:

```typescript
const canonical = allDirs.has(path)
  ? path
  : [...allDirs].find((d) => d.toLowerCase() === path.toLowerCase());
```

If the user typed a casing that actually exists as a real directory, we leave it alone. We only "correct" when the typed casing doesn't match any real directory — meaning there's at most one case-insensitive match that makes sense.

lstutzman:
Rebased onto latest dev, all tests pass. Recap of what changed since d0sboots' feedback:

The fix now checks for an exact case match first via `allDirs.has(path)`, and only falls back to case-insensitive `.find()` when no exact match exists. So if a player actually has both `Tasks/` and `tasks/` on the same server, typing the correct casing leaves it alone — we only "correct" when the typed casing doesn't correspond to any real directory.

The test covers three scenarios: wrong-cased folder name, fully uppercased, and wrong-cased with `./` prefix. All produce correctly-cased output while preserving any navigation prefix.

src/Terminal/getTabCompletionPossibilities.ts
```diff
@@ -53,7 +53,24 @@ export async function getTabCompletionPossibilities(terminalText: string, baseDi
...
+    // exists. This handles the edge case where two directories differ only in case (e.g., both
+    // Tasks/ and tasks/ exist) — we respect the user's typed casing if it matches a real directory.
+    if (path !== root) {
+      const allDirs = getAllDirectories(Player.getCurrentServer());
+      const canonical = allDirs.has(path) ? path : [...allDirs].find((d) => d.toLowerCase() === path.toLowerCase());
+      if (canonical && canonical !== path) {
+        baseDir = canonical;
+        // Fix relativeDir: replace the resolved directory suffix with canonical casing,
+        // preserving any navigation prefix (e.g., "../" or "./")
+        if (relativeDir.toLowerCase().endsWith(path.toLowerCase())) {
```
d0sboots: I don't think this code does what the comment says. For instance, if basedir was "foo/baz", currentText is "../bar/ta", actual file is "foo/Bar/Task.js", relativeDir will be "../bar/”, path will be "foo/bar/”, canonical will be "foo/Bar/" which changes baseDir, relativeDir does *not* end with path so it won't be fixed.

lstutzman:
Yeah, you're right about this, sorry for the runaround. Traced through your example properly and the old code really does spit out `../bar/Task.js`.

What I got wrong: I was treating `relativeDir` and `path` like they belonged in the same frame, when they don't. `path` is absolute from the server root (that's what resolveDirectory hands back) but `relativeDir` is just whatever the user typed up to the last slash. They only line up as strings in the one shape all my tests happened to use: baseDir=root, no `../`. Any time the resolved path picks up a segment the user never typed, like your `foo/` surviving the `../bar/` pop, `relativeDir.endsWith(path)` is false and the whole correction silently no-ops. Same story for a plain `bar/` typed from cwd `foo/` when the real dir is `foo/Bar/`. Nothing exotic about the input, my test coverage just missed the entire non-root-cwd dimension.

Rewrote the correction to work off the actual structure. Peel the leading `/` and any run of `./`/`../` off the front of `relativeDir`, count the typed directory segments that follow, and replace them with the last N segments of the canonical path. Keeps whatever navigation style the user was typing while fixing the casing.

Added the tests I should've had the first time: forward path from a non-root cwd, `../` navigation to a wrong-case destination (your example), and a multi-segment wrong-case path. Full suite green. Also rebased on dev while I was in there since the branch had fallen behind.

PR #2617, opened 2026-04-02
UI: Fix scrollTo firing on every render instead of page change

### LINKED ISSUE #2616: UI: GameRoot useEffect triggers window.scrollTo on every render
## Description

The `useEffect` in `GameRoot.tsx` (line 286-288) that scrolls to the top of the page has no dependency array:

```tsx
useEffect(() => {
  if (pageWithContext.page !== Page.Terminal) window.scrollTo(0, 0);
});
```

Without a dependency array, React runs this effect after **every render**. On pages that re-render frequently via `useCycleRerender()` (Corporation, Gang, Stock Market — 5x/sec), this calls `window.scrollTo(0, 0)` 5 times per second even though the page hasn't changed.

`window.scrollTo` is a layout-forcing DOM operation that triggers the browser's layout recalculation pipeline, adding unnecessary overhead to every render cycle.

## Expected Behavior

`scrollTo(0, 0)` should only fire when the page actually changes, which is the clear intent of the code.

## Fix

Add `[pageWithContext.page]` as the dependency array:

```tsx
useEffect(() => {
  if (pageWithContext.page !== Page.Terminal) window.scrollTo(0, 0);
}, [pageWithContext.page]);
```

One-line change. The effect only depends on `pageWithContext.page`, so it should only re-run when the page changes.

lstutzman:
## Summary

Fixes #2616

The `useEffect` in `GameRoot.tsx` that scrolls to the top of the page had no dependency array, causing `window.scrollTo(0, 0)` to fire after every React render. On pages using `useCycleRerender()` (Corporation, Gang, Stock Market), this triggered ~5 unnecessary `scrollTo` calls per second and forced browser layout recalculations.

## Root Cause

```tsx
useEffect(() => {
  if (pageWithContext.page !== Page.Terminal) window.scrollTo(0, 0);
}); // ← no dependency array = runs every render
```

Without a dependency array, React runs the effect after every render. The intent is clearly to scroll on page change only.

## Fix

Added `[pageWithContext.page]` as the dependency array. The effect only depends on the current page, so it should only re-run when the page changes.

One-line change.

## Test

New test file `test/jest/ui/GameRootScrollTo.test.ts` with 3 tests:
- Demonstrates the bug: without deps, `scrollTo` fires on every render (5 calls for 5 renders)
- With the fix: `scrollTo` fires only on page change (2 calls for 5 renders across 2 pages)
- Terminal page is correctly excluded from scrolling

## Checklist

- [x] Branched from `dev`
- [x] `npm run lint` passes
- [x] `npm run format` passes
- [x] `npm run test` passes (52 suites, 4394 tests)
- [x] Test added

### DIFF
--- a/src/ui/GameRoot.tsx
+++ b/src/ui/GameRoot.tsx
@@ -285,7 +285,7 @@ export function GameRoot(): React.ReactElement {
 
   useEffect(() => {
     if (pageWithContext.page !== Page.Terminal) window.scrollTo(0, 0);
-  });
+  }, [pageWithContext.page]);
 
   function softReset(): void {
     dialogBoxCreate("Soft Reset!");


catloversg:
The "fix" is correct, but the test is unnecessary and misleading as I said in #2620. I use double quotes here because while this fix is technically correct, the issue mentioned in the description does not actually exist (i.e., `GameRoot` is not rerendered as you said).

The wrong assumption in this PR and the linked issue is that `scrollTo` is called every time a mechanic is rerendered due to `useCycleRerender`. You can test this easily like this:
- Add `console.log("render root");` in `src\ui\GameRoot.tsx` right before the `return` block.
- Add `console.log("render corp");` in `src\Corporation\ui\CorporationRoot.tsx` right before the `return` block.
- Open the corp tab and check the console.

The corp tab is rerendered 5 times per second, but the parent component (`GameRoot`) is not.

The test code should be removed. If you want to test UI code written in React, you need to use a specialized testing library designed for that job. Your test code is testing itself, not the wrong code. And it's testing a nonexistent issue.

lstutzman:
You're right — I tested it and GameRoot doesn't re-render from child useCycleRerender calls. My assumption about the rendering cascade was wrong. Thanks for walking me through the reasoning.

Removed the test and updated the commit message. The fix itself is still worth having IMO — it makes the dependency contract explicit so the intent is clear. But it's a code quality thing, not a perf thing. Happy to close if you'd rather not bother with it.

catloversg (approved):

lstutzman:
Rebased onto latest dev, tests pass. This is just the one-line dep array addition — test was removed per catloversg's feedback.

To be clear on what this actually does: it makes the useEffect's dependency explicit rather than implicit. The scroll-to-top only needs to fire on page change, and adding `[pageWithContext.page]` makes that contract visible in the code rather than relying on the fact that GameRoot happens to not re-render often. Defensive, not urgent.

PR #2620, opened 2026-04-02
CODEBASE: Add dependency array to TerminalInput keydown useEffect

### LINKED ISSUE #2618: UI: TerminalInput keydown listener re-registers on every render
## Description

The `useEffect` in `TerminalInput.tsx` (lines 210-225) that registers a document-level `keydown` listener has no dependency array:

```tsx
useEffect(() => {
  function keyDown(this: Document, event: KeyboardEvent): void {
    // ...handler logic...
  }
  document.addEventListener("keydown", keyDown);
  return () => document.removeEventListener("keydown", keyDown);
}); // ← no dependency array
```

Without a dependency array, React runs the cleanup and re-setup on every render. Since `TerminalInput` is a controlled input, it re-renders on every keystroke — causing an unnecessary `removeEventListener` + `addEventListener` cycle per character typed.

The handler only captures `terminalInput` (a ref, stable across renders) and `Terminal` (a global singleton). No actual dependencies change between renders.

## Fix

Add `[]` as the dependency array. One-line change.

lstutzman:
## Summary

Fixes #2618

The `useEffect` that registers a document-level `keydown` listener in `TerminalInput.tsx` had no dependency array, causing React to tear down and re-create the listener on every render. Since the terminal re-renders on every keystroke, this caused an unnecessary `removeEventListener` + `addEventListener` cycle per character typed.

## Root Cause

```tsx
useEffect(() => {
  function keyDown(...) { ... }
  document.addEventListener("keydown", keyDown);
  return () => document.removeEventListener("keydown", keyDown);
}); // ← no dependency array = runs every render
```

The handler only captures `terminalInput` (a ref, stable) and `Terminal` (a singleton). No dependencies change between renders.

## Fix

Added `[]` as the dependency array. One-line change.

## Test

New test `test/jest/Terminal/terminalKeydownRegistration.test.ts`:
- Verifies correct behavior: listener registered once, removed once
- Demonstrates the bug: without deps, 5 renders = 5 add/remove cycles

## Checklist

- [x] Branched from `dev`
- [x] `npm run lint` passes
- [x] `npm run format` passes
- [x] `npm run test` passes (52 suites, 4393 tests)
- [x] Test added

### DIFF
--- a/src/Terminal/ui/TerminalInput.tsx
+++ b/src/Terminal/ui/TerminalInput.tsx
@@ -222,7 +222,7 @@ export function TerminalInput(): React.ReactElement {
     }
     document.addEventListener("keydown", keyDown);
     return () => document.removeEventListener("keydown", keyDown);
-  });
+  }, []);
 
   async function onKeyDown(event: React.KeyboardEvent<HTMLInputElement | HTMLTextAreaElement>): Promise<void> {
     const ref = terminalInput.current;


catloversg:
The fix is correct, but the Jest test is unnecessary and misleading.

I notice that you write this kind of Jest test in many UI-related PRs, but instead of testing the wrong code and the fix, these tests *simulate* the wrong behavior and the fix. This means your tests are actually testing the test code (the simulator).

This kind of testing gives the "illusion" that new tests cover this issue and ensure that there is no regression in the future. However, if we mistakenly revert the fix, those tests won't catch it. In some cases, these tests misleadingly make us think an issue exists when it actually does not, or the issue is a completely different one. For example, in #2617, you are testing the wrong thing based on a wrong assumption. I'll explain what is wrong there.

d0sboots:
I'll agree with CatLover that although in general we welcome better test coverage, for many of these I don't think the tests are actually helping much. (But for some, especially the ones fixing logic bugs, they're very welcome!)

lstutzman:
Yeah, that's a fair criticism across the board. The simulated tests were testing themselves rather than the actual component — lesson learned. Removed the test, kept just the one-line fix.

For the logic bug PRs (alias recursion, skillMaxUpgradeCount, autocomplete) the tests exercise real exported functions so those should be fine. For the UI/React ones I'll stick to the code change and let review carry the weight.

catloversg (approved):

Issue #924, opened 2023-11-09
Prompt of type text loses input focus randomly


Lloople:
When opening a prompt with `ns.prompt('Some text, {type:'text'})`. Sometimes the caret is set on the terminal at background instead of the input on the prompt.

Code to reproduce:

```js

/** @param {NS} ns */
export async function main(ns) {

    const result = await ns.prompt('Some dummy text', {type: 'text'})
    
}
```

Also tried this with @Caldwell-74 's help. The input acquires the focus but it loses it a moment after.

```js
/** @param {NS} ns */
export async function main(ns) {

    const action = await askForAction(ns)
    
}

async function askForAction(ns) {

    const userPrompt = ns.prompt('Test', {type: 'text'})
    
    await ns.asleep(500)

    eval('document').querySelector('.MuiInputBase-input').focus()

    return await userPrompt
}
```

- Version of the game: Steam 2.5.0
- OS: MacOS Ventura

catloversg:
Please check #2622 to see why I reopen this issue.

catloversg:
In #2622, I mentioned some inconsistent behaviors related to the prompt losing focus. This issue will be fixed in #2631. Please check that PR for more information.

PR #2622, opened 2026-04-03
UI: Follow-up to #2615

catloversg:
This is the follow-up to #2615.

The fix in that PR is partially correct (it needs to handle another case), but it fixes a different issue, not #924. The issue described in that PR does not exist at all.

Test code:
```jsx
// a.js
/** @param {NS} ns */
  ns.ui.openTail();
  await ns.prompt("Some dummy text", { type: "text" });
}
```

Comment out the fix, run `test.jsx`. ~The text box of the prompt dialog automatically gets focus and if you don't click anything else, it won't lose focus~ (Edit: It actually may lose focus; more about this later). Now type text in that text box. The focus does not change, and it works as expected. That PR claims that it's impossible to type in the prompt dialog, but it's actually possible.

Let's take a closer look at #924 and its second example. In that test code, we sleep for a brief moment and then intentionally focus on the terminal. Obviously, if you focus on the terminal, the prompt dialog will lose its focus.

~I think #924 is describing this behavior: Run my test code, but when the prompt dialog is shown, instead of typing in its text box, you click the "Some dummy text" part. When you click on that text, the text box of the prompt dialog loses focus and the text you type will appear in the terminal, then the terminal gets focus. IMO, this is not an issue at all. Note that if you remove the `ns.ui.openTail();`, you can reproduce the loss of focus. However, you can type text as normal.~
Edit: There is a specific way to make the prompt modal lose focus after being rendered (you can still type in its text box, though). The examples in #924 do not show it. I'll add more information to #924 later.

Now let's take a look at this analysis:
```
The prompt's onKeyDown handler does call event.stopPropagation(), but the terminal's listener is a native
document.addEventListener("keydown") which can still receive the event depending on the event dispatch order.
```
This is wrong, at least in this context. `event.stopPropagation()` does its job properly. You can put a debugger in the handlers to verify it. If you comment out `event.stopPropagation()`, the prompt dialog will lose focus after it's shown, and even if you focus on it again, it always loses focus when you type. In short, `event.stopPropagation()` already prevents the issue that this PR describes.

The fix in this PR addresses another issue. Test code:
```jsx
// test.jsx
/** @param {NS} ns */
  ns.ui.openTail();
  ns.clearLog();
  ns.printRaw(<input autoFocus />);
  ns.printRaw(<div contentEditable style={{backgroundColor: "white"}} />);
}
```
If you focus on these elements and type in them, they will lose focus and the text will appear in the terminal. It happens because they don't prevent event propagation like the prompt dialog does.

Now let's check the coding contract modal. It does not call `event.stopPropagation()`, but you can type the answer normally. Why is that? The trick is here:
https://github.com/bitburner-official/bitburner-src/blob/63aa4d2a45f5d1a827e0a43f10c853ed702e6e4c/src/Terminal/ui/TerminalInput.tsx#L212
We have a special check for the coding contract modal.

This highlights an issue: If there is a text box rendered in the terminal tab, by default (without using `event.stopPropagation()` or having a special check in Bitburner's side), you cannot type in that text box. The fix in #2615 resolves this issue. Note that we need to check the `contentEditable` attribute  with `target.isContentEditable`, as I do in this PR.

A final note about #924: As I said above, there are ways to reproduce the loss of focus, but it's unrelated to the ability to type in the text box. It's a different issue. For example:
```jsx
// a.js
/** @param {NS} ns */
  // ns.ui.openTail();
  await ns.prompt("Some dummy text", { type: "text" });
}
```
The prompt dialog gets focus with `ns.ui.openTail()`, but it does not if we comment out the `openTail` call. This may be related to the tail window. I'll investigate it later.
Edit 1: Even without the `openTail` call, the prompt dialog still gets focus once after loading the game. However, it does not get focus the next time we run `a.js`. With the `openTail` call, depending on how we run `a.js`, the prompt dialog may or may not get focus.
Edit 2: I found the issue. I'll explain it in #924 later.

Regarding the removal of the Jest test, I believe this is the right move. As I said in other PRs, this kind of Jest test is misleading. It checks the logic of the change (the new if condition), so it's not exactly useless. However, it's testing itself (the test code), not the wrong code in `src\Terminal\ui\TerminalInput.tsx`. If we mistakenly revert the fix, this test will still pass as normal. IMO, this test is only useful for verifying the logic while reviewing the PR. If we add it to our test suite, it will make us think that this issue is covered by a test, but it's actually not.


d0sboots:
Ah yes, I missed that the test was creating its own handler instead of actually testing the terminal.

PR #2626, opened 2026-04-04
API: Add weakenEffect to formulas.hacking namespace

### LINKED ISSUE #2081: Formulas API should provide formulas to calculate weaken effect
The API has four functions each for `hack` and `grow`, including formulas that calculate the effect of using a given number of threads and the number of threads required to achieve a desired effect. However, there are no such formulas for `weaken`. They would be unnecessary if `weaken` had a consistent effect of -0.05 security per thread, but extra cores on a server boost the effect of `weaken`. The only in-game way of calculating what effect this is is to use the hefty 1 GB `weakenAnalyze` function, and reverse engineering the effect of cores through said function may prove difficult as there seems to be a secret extra modifier to `weaken`'s effect. Access to the formulas API seems intended to let the player save RAM by avoiding the expensive `Analyze` functions, but it is incomplete without a counterpart to `weakenAnalyze`.

lstutzman:
## Summary

Fixes #2081

The Formulas API has functions for hack and grow effects but no equivalent for weaken, forcing players to use the 1 GB `weakenAnalyze` for what is fundamentally `0.05 * threads * coreBonus`. More importantly, the core bonus formula (`1 + (cores - 1) / 16`) is not documented or discoverable in-game without reading the source code. The Formulas API is meant to expose these internals — this completes the set.

## Fix

Added `formulas.hacking.weakenEffect(threads, cores?)` that delegates to the existing `getWeakenEffect()` in ServerHelpers. Zero RAM cost, gated by Formulas.exe ownership — same pattern as all other formula functions.

Unlike other hacking formulas, weaken effect doesn't depend on server or player properties, so the signature takes `(threads, cores?)` rather than `(server, player)` — matching `weakenAnalyze`.

## Files Changed

- `src/NetscriptFunctions/Formulas.ts` — add weakenEffect function
- `src/Netscript/RamCostGenerator.ts` — add RAM cost entry (0)
- `src/ScriptEditor/NetscriptDefinitions.d.ts` — add type signature + JSDoc

## Test

5 tests in `test/jest/Netscript/WeakenEffect.test.ts` covering the underlying formula:
- Single core (0.05 per thread)
- Multi-core with core bonus (8 cores = 1.4375x)
- Zero threads (returns 0)
- Edge cases

## Checklist

- [x] Branched from `dev`
- [x] `npm run lint` passes
- [x] `npm run format` passes
- [x] `npm run test` passes (54 suites, 4417 tests)
- [x] Test added


d0sboots:
You need to run npm doc. That should conveniently rerun the flaky test.

lstutzman:
Done — ran npm doc and pushed the generated API docs for weakenEffect.

PR #2627, opened 2026-04-04
DNET: Remove bonus time effect on authentication and heartbleed speed; fix ram rounding

ficocelliguy:
To prevent bonus time from messing with the timing of authentication, this PR removes bonus time as a factor from calculateAuthenticationTime. It remains only as a factor for xp gained from auth and money gained from phishing.

In addition, this PR ensures that the ram used on a server is always rounded to four digits, to prevent tiny floating point imprecisions from messing with available ram.

### DIFF
--- a/src/DarkNet/effects/ramblock.ts
+++ b/src/DarkNet/effects/ramblock.ts
@@ -12,6 +12,7 @@ import type { DarknetServerData, Person as IPerson } from "@nsdefs";
 import { clampNumber } from "../../utils/helpers/clampNumber";
 import { ResponseCodeEnum } from "../Enums";
 import { isLabyrinthServer } from "./labyrinth";
+import { roundToN } from "../../utils/helpers/roundToTwo";
 
 /*
  * Handles the effects of removing some blocked RAM from a Darknet server.
@@ -21,7 +22,7 @@ export const handleRamBlockRemoved = (ctx: NetscriptContext, server: DarknetServ
   const difficulty = server.difficulty + 1;
 
   const ramBlockRemoved = getRamBlockRemoved(server, threads);
-  server.blockedRam -= ramBlockRemoved;
+  server.blockedRam = roundToN(server.blockedRam - ramBlockRemoved, 4);
   server.updateRamUsed(server.ramUsed - ramBlockRemoved);
 
   if (server.blockedRam <= 0) {
--- a/src/DarkNet/effects/ramblock.ts
+++ b/src/DarkNet/effects/ramblock.ts
@@ -101,5 +101,5 @@ export const getRamBlock = (maxRam: number): number => {
     return [16, 32, maxRam - 8][Math.floor(Math.random() * 3)];
   }
 
-  return [maxRam, maxRam - 8, maxRam - 64, maxRam / 2][Math.floor(Math.random() * 4)];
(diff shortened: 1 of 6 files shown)


src/utils/helpers/roundToTwo.ts
```diff
@@ -3,5 +3,15 @@
...
-  return Math.round(decimal * 100) / 100;
+  return roundToN(decimal, 2);
+}
+
+/**
+ * Rounds a number to n decimal places.
+ * @param decimal A decimal value to trim to n places.
+ * @param digits the number of digits to trim to.
+ */
+export function roundToN(decimal: number, digits: number): number {
```
d0sboots: No, it *must* be roundToTwo. There are many places throughout the code that have this as a fundamental assumption, and finer-grained amounts will *not* be supported correctly - in fact, they will continue to lead to this bug. For instance, if ramBlockRemoved is .004, the use of roundToN will cause ramBlock to be updated, but use of roundToTwo in script ram calculation will round that back to 0, causing ramUsed to not be updated.

It is *vital* that all ram calculations be performed the same way.

Edit: The way you can think of it is that we really wanted our unit to be "10MB" and use an integer for it, and we are emulating it via roundToTwo. We don't actually want any decimal math here.

src/Server/BaseServer.ts
```diff
@@ -233,7 +234,7 @@ export abstract class BaseServer implements IServer {
   }
 
   updateRamUsed(ram: number): void {
-    this.ramUsed = clampNumber(ram, 0, this.maxRam);
+    this.ramUsed = clampNumber(roundToN(ram, 4), 0, this.maxRam);
```
d0sboots: Since the clamp exists to protect against infinity, NaN etc., if we apply roundToTwo here (which I agree is a good idea), it should go outside of the clamp.

src/utils/helpers/roundToTwo.ts
ficocelliguy: updated to use roundToTwo everywhere.

src/DarkNet/effects/ramblock.ts
```diff
@@ -21,7 +22,7 @@ export const handleRamBlockRemoved = (ctx: NetscriptContext, server: DarknetServ
   const difficulty = server.difficulty + 1;
 
   const ramBlockRemoved = getRamBlockRemoved(server, threads);
```
d0sboots, line 24: There's a subtlety here. `getRamBlockRemoved` *itself* should also use roundToTwo (internally). This matters so that formulas gets the same effective numbers that are being used here.
Actually, the same is also true of `getRamBlock` - it *almost* doesn't need it, but there's a / 2 that can interact with the new fractional getMaxRam in bad ways, and it's best for safety anyway.

(Note that it's still needed here as well, to protect against incorrect rounding after the subtraction.)

src/DarkNet/effects/ramblock.ts
d0sboots:
```suggestion
  return roundToTwo(clampNumber(baseAmount * difficultyFactor * threads * charismaFactor, 0, remainingRamBlock));
```

d0sboots:
```suggestion
  return roundToTwo([maxRam, maxRam - 8, maxRam - 64, maxRam / 2][Math.floor(Math.random() * 4)]);
```

d0sboots: This should be formatRam (to respect the user's choice of units)

PR #2634, opened 2026-04-05
DOCUMENTATION: Clarify scp and exec darknet permissions in API docs

### LINKED ISSUE #2530: darknet servers can scp to darkweb when airgapped

lstutzman:
The scp JSDoc mentions needing a session for darknet servers, but doesn't explain that unlike exec, scp doesn't need a direct connection — it works at any distance. This is the main thing that confused the reporter in #2530.

Updated the scp note to spell out the full picture: destination needs a session (but no connection), source needs nothing at all. Added cross-references to `dnet.authenticate` and `dnet.connectToSession` so players know how to actually get a session.

Also added a darknet note to exec, which had zero mention of darknet/session/connection requirements.

Fixes #2530

### DIFF
--- a/src/ScriptEditor/NetscriptDefinitions.d.ts
+++ b/src/ScriptEditor/NetscriptDefinitions.d.ts
@@ -7905,6 +7905,11 @@ export interface NS {
    * // arguments to the script.
    * ns.exec("foo.js", "foodnstuff", 5, 1, "test");
    * ```
+   *
+   * For darknet servers: A session must be established with the target server, and the script must be
+   * running on a server that is directly connected to the target, or the target must have a backdoor or
+   * stasis link installed.
+   *
    * @param script - Filename of script to execute. This file must already exist on the target server.
    * @param host - Hostname/IP of the target server on which to execute the script.
    * @param threadOrOptions - Either an integer number of threads for new script, or a {@link RunOptions} object. Threads defaults to 1.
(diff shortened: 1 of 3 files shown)


src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -8042,7 +8047,11 @@ export interface NS {
    * ns.scp(files, server, "home");
    * ```
    *
-   * For password-protected servers (such as darknet servers), a session must be established with the destination server before using this function. (The source server does not require a session.)
+   * For darknet servers: The destination requires a session, but unlike {@link NS.exec | exec}, does not
+   * require a direct connection — scp works at any distance. The source server has no darknet requirements
+   * (no session, no connection). Use {@link Darknet.authenticate | dnet.authenticate} (requires direct
+   * connection) or {@link Darknet.connectToSession | dnet.connectToSession} (works at any distance) to
+   * establish a session.
```
ficocelliguy, line 8054: I think this can be clarified a bit more. `(no session, no connection)` isn't super clear - can you make that a full phrase, like `(no session or connection needed)` ?

Also, `(works at any distance)` can probably be simplified to something like `(at any distance)` so it flows better in the sentence

lstutzman:
Thanks for the review - reworded both phrases as suggested. Also regenerated the markdown doc.

ficocelliguy (approved):

PR #2635, opened 2026-04-06
API: Rename ns.gang.getOtherGangInformation to getAllGangInformation

### LINKED ISSUE #2547: ns.gang.getOtherGangInformation actually returns all gangs information (including own gang)
A bit of a discrepancy in naming. The function comment detail actually explains that it shows all gang's info, but the name of the function itself is quite misleading.

lstutzman:
Fixes #2547

`getOtherGangInformation` returns info about all gangs including the player's own, so "Other" was misleading. d0sboots and catloversg both agreed `getAllGangInformation` is a better name.

### What changed

- Renamed the function in the implementation, type definitions, and RAM cost table
- Added a `setRemovedFunctions` deprecation stub so scripts using the old name get a clear error pointing to the new one
- Added a migration entry in `APIBreaks/3.0.0.ts` so existing player scripts get auto-migrated
- Added changelog entry to the 3.0.0 breaking changes section
- Regenerated API docs via `npm run doc`

### Test

Added `test/jest/Netscript/Gang.test.ts` — verifies `getAllGangInformation` returns all 7 gangs with correct shape and returns copies (not references).

```
npm run lint   # ✓
npm run format # ✓
npm run test   # 54 suites, 4416 tests pass
```


src/Constants.ts
```diff
@@ -125,6 +125,7 @@ export const CONSTANTS = {
 - Rename setAutoJobAssignment API to setJobAssignment (#2146) (@catloversg)
 - Make nuke and port cracking APIs return false instead of throwing error (#1023, #2153) (@Hydrogeniouss, @catloversg)
 - Standardize names of Stock APIs (#2173) (@catloversg)
+- Rename ns.gang.getOtherGangInformation to ns.gang.getAllGangInformation (#2547)
```
catloversg: Don't update the changelog in normal PRs. The changelog will be updated right before we release a new version.

catloversg (commented):
The poll on Discord does not give a clear result whether we should do this, at least right now.

lstutzman:
Thanks, removed the changelog entry. Will remember for future PRs.

PR #2502, opened 2026-02-14
API: Remove RAM cost of hacknet namespace and set RAM cost of each hacknet API

catloversg:
For more information, please check #2498.

Test code:
```js
/** @param {NS} ns */
export async function main(ns) {
  ns.ui.openTail();
  ns.hacknet.numNodes();
  ns.hacknet.purchaseNode();
  ns.hacknet.getPurchaseNodeCost();
  ns.hacknet.getNodeStats(0);
  ns.hacknet.upgradeLevel(0);
  ns.hacknet.upgradeRam(0);
  ns.hacknet.upgradeCore(0);
  ns.hacknet.upgradeCache(0);
  ns.hacknet.getLevelUpgradeCost(0);
  ns.hacknet.getRamUpgradeCost(0);
  ns.hacknet.getCoreUpgradeCost(0);
  ns.hacknet.getCacheUpgradeCost(0);
  ns.hacknet.numHashes();
  ns.hacknet.hashCost("Company Favor");
  ns.hacknet.spendHashes("Company Favor");
  ns.hacknet.maxNumNodes();
  ns.hacknet.hashCapacity();
  ns.hacknet.getHashUpgrades();
  ns.hacknet.getHashUpgradeLevel("Company Favor");
  ns.hacknet.getStudyMult();
  ns.hacknet.getTrainingMult();
}
```

```
Accessing the hacknet namespace incurred a one-time cost of 4 GB of RAM, and each hacknet API did not incur 
RAM cost. Now the hacknet namespace does not incur RAM cost, but each hacknet API incurs small RAM cost.

Usage of the following functions may have been affected:
ns.hacknet.numNodes
ns.hacknet.purchaseNode
ns.hacknet.getPurchaseNodeCost
ns.hacknet.getNodeStats
ns.hacknet.upgradeLevel
ns.hacknet.upgradeRam
ns.hacknet.upgradeCore
ns.hacknet.upgradeCache
ns.hacknet.getLevelUpgradeCost
ns.hacknet.getRamUpgradeCost
ns.hacknet.getCoreUpgradeCost
ns.hacknet.getCacheUpgradeCost
ns.hacknet.numHashes
ns.hacknet.hashCost
ns.hacknet.spendHashes
ns.hacknet.maxNumNodes
ns.hacknet.hashCapacity
ns.hacknet.getHashUpgrades
ns.hacknet.getHashUpgradeLevel
ns.hacknet.getStudyMult
ns.hacknet.getTrainingMult

[...]

PR #2639, opened 2026-04-07
DOCUMENTATION: Update RAM cost of hacknet APIs and remove unnecessary RAM cost docs

catloversg:
We recently changed the RAM cost of hacknet APIs in #2502.


src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -6999,7 +6999,7 @@ interface UserInterface {
 export interface NS {
   /**
    * Namespace for {@link Hacknet | hacknet} functions. Some of this API contains spoilers.
-   * @remarks RAM cost: 4 GB.
+   * @remarks RAM cost: 0GB.
```
d0sboots: Just remove this, like other namespaces it doesn't have a cost now.

PR #2645, opened 2026-04-09
UI: Consistently calculate BitNode "level"

catloversg:
Strictly speaking, there is no BitNode "level". However, saying "Enter BN1.2" is shorter than saying "Enter BN1 with SF1.1". This is how we display it in the BitVerse UI and other places. However, this "level" is not consistently calculated in all places. For example, with SF1.3, the import comparison popup shows "BitNode 1-4" instead of "BitNode 1-3".

This PR adds a utility function to consistently calculate this value and uses it in the relevant places.

Tested manually.

### DIFF
--- a/src/BitNode/BitNodeUtils.ts
+++ b/src/BitNode/BitNodeUtils.ts
@@ -95,9 +95,9 @@ export function finishBitNode() {
 }
 
 /**
- * Strictly speaking, there is no BitNode "level". However, saying "Enter BN1.2" is shorter than saying "Enter BN1 with
- * SF1.1". This is how we display it in the BitVerse UI and other places. This function is used to consistently
- * calculate this "level".
+ * BitNode level is not something that is stored, but rather calculated from the current BN and SF level. The concept
+ * appeared because saying "Enter BN1.2" is shorter than saying "Enter BN1 with SF1.1". This is how we display it in the
+ * BitVerse UI and other places. This function is used to consistently calculate this "level".
  */
 export function getBitNodeLevel(bn = Player.bitNodeN, sfLevel = Player.activeSourceFileLvl(bn)): number {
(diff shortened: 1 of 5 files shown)


src/BitNode/BitNodeUtils.ts
```diff
@@ -93,3 +93,12 @@ export function finishBitNode() {
   }
   wd.backdoorInstalled = true;
 }
+
+/**
+ * Strictly speaking, there is no BitNode "level". However, saying "Enter BN1.2" is shorter than saying "Enter BN1 with
```
d0sboots:
```suggestion
 * BitNode level is not something that is stored, rather it is calculated from the current BN and SF level.  The concept appeared because saying "Enter BN1.2" is shorter than saying "Enter BN1 with
```
Reworded slightly. What you had made it sound like BN level isn't "real," but it very much is a real thing (as evidenced by how many places use this function), it's just not explicitly stored.

src/BitNode/ui/BitnodeMultipliersDescription.tsx
```diff
@@ -56,8 +56,7 @@ export const BitNodeMultipliersDisplay = ({ n, level, hideMultsIfCannotAccessFea
   // If not, then we have to assume that we want the next level up from the
   // current node's source file, so we get the min of that, the SF's max level,
   // or if it's BN12, ∞
-  const maxSfLevel = n === 12 ? Number.MAX_VALUE : 3;
-  const mults = getBitNodeMultipliers(n, level ?? Math.min(Player.activeSourceFileLvl(n) + 1, maxSfLevel));
+  const mults = getBitNodeMultipliers(n, level ?? getBitNodeLevel(n, Player.activeSourceFileLvl(n)));
```
d0sboots: Optional: should be able to use the default for the 2nd arg

PR #2646, opened 2026-04-09
CLI: Add "hidden" mkdir command

### LINKED ISSUE #2637: Add mkdir support
It has confused many players that there is no mkdir command in the game. Adding support would give a nice bit of realism.

Currently the filesystem is the same as azure cloud's directory solution - directories are just part of file names and cannot exist alone/empty. I don't think it is worth overhauling that for some simple flavor. However, we could add a file type that is invisible to the player and just exists to hold directory paths, and makedir would create such a file.

catloversg:
For more information, please check #2637.

Screenshot:

Capture

Closes #2637.

### DIFF
--- a/src/Terminal/Terminal.ts
+++ b/src/Terminal/Terminal.ts
@@ -78,6 +78,7 @@ import { commitHash } from "../utils/helpers/commitHash";
 import { apr1 } from "./commands/apr1";
 import { changelog } from "./commands/changelog";
 import { clear } from "./commands/clear";
+import { mkdir } from "./commands/mkdir";
 import { currentNodeMults } from "../BitNode/BitNodeMultipliers";
 import { Engine } from "../engine";
 import { Directory, resolveDirectory, root } from "../Paths/Directory";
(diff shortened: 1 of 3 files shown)


src/Terminal/Terminal.ts
```diff
@@ -877,7 +882,7 @@ export class Terminal {
 }
 
 function findSimilarCommands(command: string): string[] {
-  const commands = Object.keys(TerminalCommands);
+  const commands = supportedCommands.slice();
```
d0sboots: Optional: You can just replace "commands" with "supportedCommands" here, it isn't mutated.

src/Terminal/getTabCompletionPossibilities.ts
```diff
@@ -84,7 +84,7 @@ export async function getTabCompletionPossibilities(terminalText: string, baseDi
 
   const addAliases = () => addGeneric({ iterable: Aliases.keys() });
   const addGlobalAliases = () => addGeneric({ iterable: GlobalAliases.keys() });
-  const addCommands = () => addGeneric({ iterable: Object.keys(TerminalCommands) });
+  const addCommands = () => addGeneric({ iterable: supportedCommands.slice() });
```
d0sboots: addGeneric *shouldn't* be mutating its argument, so a defensive slice isn't needed

PR #2650, opened 2026-04-10
MISC: Reduce achievements check interval

catloversg:
Players often ask why they haven't received an achievement despite meeting the requirements. Usually, they just need to wait for the next periodic check, but sometimes they don't actually satisfy the requirements as they thought.

~Adding this button allows players to trigger a manual check and get immediate feedback.~
Edit: This PR nows increases the frequency of checking achievements instead of adding a new button.

I also removed redundant styles in `src/Achievements/AchievementsRoot.tsx`:
- `width: 50` is overridden by `width: "100%"`
- `userSelect: "none"`: I don't see why we need to prevent players from selecting or highlighting text and other elements.

Screenshot:

Capture


src/Achievements/AchievementsRoot.tsx
```diff
@@ -1,28 +1,22 @@
...
+    <div style={{ width: "100%" }}>
       <Typography variant="h4">Achievements</Typography>
       <Box mx={2}>
         <Typography>
           Achievements are persistent rewards for various actions and challenges. A limited number of Bitburner
           achievements have corresponding achievements in Steam.
+          <br />
+          Achievements are automatically checked periodically. You can also manually check with the button below.
         </Typography>
+        <Button onClick={() => calculateAchievements()}>Check achievements</Button>
```
d0sboots: This seems a bit janky. Is it an actual performance problem to up the frequency on achievement checking to the point where the lag isn't that noticable?

catloversg: I set the delay to 5 seconds. `calculateAchievements()` is very fast, so there should not be any performance problems.

src/engine.tsx
```diff
@@ -215,7 +215,7 @@ const Engine = {
 
     if (Engine.Counters.achievementsCounter <= 0) {
       calculateAchievements();
-      Engine.Counters.achievementsCounter = 300;
+      Engine.Counters.achievementsCounter = 25;
```
d0sboots: How fast is "very fast?" As long as it's reasonably "faster than saving your game," which is an operation that can impact batchers but we still do pretty frequently, I'd say push this all the way to once a second - that's the speed that no one will notice the lag, whereas five seconds is just long enough to think, "did it trigger?"

catloversg: `calculateAchievements` takes less than 1ms. `saveObject.saveGame` takes at least ~20ms.

I changed the delay to 1 second.

PR #2659, opened 2026-04-13
CLI: Add new command to upload a directory

### LINKED ISSUE #1522: Add "upload" terminal command

hexagonrecursion:
closes https://github.com/bitburner-official/bitburner-src/issues/1522

* This pull request is an alternative to https://github.com/bitburner-official/bitburner-src/pull/2609

Unlike #2609, this pull request shows the user a directory-picker dialog and uploads a whole directory.

1. I like this version better because if a user on Windows wants to upload a directory and the file names contain non-ASCII characters and if they create a .zip archive from the directory by using the built-in windows file manager, the code in #2609 will not decode the file names correctly
2. I also like that this version lets me skip the step of having to create a .zip archive to upload files from my local git repository

- [x] manually tested in Firefox 149.0.2 on Linux
- [x] manually tested in Chromium 146.0.7680.177 on Linux
- [x] manually tested in Firefox 149.0 on Windows
- [x] manually tested in Chrome 147.0.7727.56 on Windows
- I did not test electron

## Added confirmation

Thanks, @d0sboots , for the suggestion

Screenshot 2026-04-13 at 17-38-27 dev Bitburner v3 0 0dev (943cc4e3e)

## _File content_ encoding is assumed to be utf-8

* The `download` command always produces files encoded as utf-8
* In my opinion, support for other file encodings should be out of scope for this pull request. If necessary, it can be added later

## _File name_ encoding
File name encoding is irrelevant because the browser performs the decoding and gives us a `string`

## Directory separator is always /
* [mdn article File: webkitRelativePath property](https://developer.mozilla.org/en-US/docs/Web/API/File/webkitRelativePath) links to this specification: https://wicg.github.io/entries-api/#dom-file-webkitrelativepath
* The specification says "The webkitRelativePath getter steps are to return [this](https://webidl.spec.whatwg.org/#this)’s [relative path](https://wicg.github.io/entries-api/#relative-path), or the empty string if not specified"
* "A relative path is a string consisting of one or more [path segments](https://wicg.github.io/entries-api/#path-segment) joined by '/' (U+002F SOLIDUS) that does not start with '/' (U+002F SOLIDUS)."
* I tested my code in both Chrome and Firefox on Windows. It worked

## Browser support
* According to https://developer.mozilla.org/en-US/docs/Web/API/File/webkitRelativePath and https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/webkitdirectory this requires firefox >= 50 (released 2016-11-15)
* According to https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/webkitdirectory this requires Chrome >= 72 ([due to a bug](https://crbug.com/124187)) (released 2019-01-29)

## Uploading files instead of whole directories
Unfortunately it appears that there are only a directory picker dialog and a file picker dialog, but there is no single dialog that would allow the user to select both files and directories. If we want to have a command that shows a file picker, we will need two commands


src/Terminal/commands/upload.ts
```diff
@@ -0,0 +1,140 @@
+import { Terminal } from "../../Terminal";
+import { BaseServer } from "../../Server/BaseServer";
```
catloversg:
```suggestion
import type { BaseServer } from "../../Server/BaseServer";
```

src/Terminal/commands/upload.ts
```diff
@@ -0,0 +1,140 @@
...
+import { combinePath, isFilePath } from "../../Paths/FilePath";
+import { hasTextExtension, validTextExtensions } from "../../Paths/TextFilePath";
+import { hasScriptExtension, validScriptExtensions } from "../../Paths/ScriptFilePath";
+import { PromptEvent } from "../../ui/React/PromptManager";
+import { ContentFilePath } from "src/Paths/ContentFile";
```
catloversg:
```suggestion
import type { ContentFilePath } from "src/Paths/ContentFile";
```

src/Terminal/commands/upload.ts
```diff
@@ -0,0 +1,140 @@
...
+          reject(new Error("PromptEvent got a string, expected boolean"));
+        } else {
+          resolve(value);
+        }
+      },
+    });
+  });
+}
+
+function isFileExists(server: BaseServer, file: ContentFilePath): boolean {
```
catloversg: Nitpick: Both "is" and "exist" are verbs. This kind of function is usually named "fileExists".

This function is only used in one place, so it could also be inlined.

src/Terminal/commands/upload.ts
```diff
@@ -0,0 +1,140 @@
...
+
+async function uploadAsync(args: (string | number | boolean)[], server: BaseServer) {
+  if (args.length !== 1) {
+    return Terminal.error("Incorrect usage of upload command. Usage: upload [dir]");
+  }
+  const destinationInput = String(args[0]);
+  const destination = Terminal.getDirectory(destinationInput);
+  if (destination === null) {
+    return Terminal.error(`Could not resolve ${destinationInput} as a Directory`);
+  }
```
catloversg, line 49: I think `const destForPrint = destination === "" ? "/" : destination;` should be moved to below this check and `destForPrint` should be used instead of `destination` in `Upload files to ${destination}?`. Otherwise, the popup will look like this:

Image

src/Terminal/commands/upload.ts
```diff
@@ -0,0 +1,140 @@
...
+  if (files === null || files.length === 0) {
+    return;
+  }
+  const withPath: (
+    | { badPath: string }
+    | { overwrite: ContentFilePath; file: File }
+    | { create: ContentFilePath; file: File }
+  )[] = [...files].map((f) => {
+    const { webkitRelativePath } = f;
+    const path = webkitRelativePath.substring(1 + webkitRelativePath.indexOf("/"));
```
catloversg, line 74: Optional: Add a comment with example(s) of `webkitRelativePath` and the parsed `path`. The logic here is simple, so this is not required, but examples are always helpful, especially for string processing code.

src/Terminal/commands/upload.ts
```diff
@@ -0,0 +1,140 @@
...
+  const overwrite = withPath.filter((item) => "overwrite" in item);
+  const skipped = withPath.filter((item) => "badPath" in item);
+  const create = withPath.filter((item) => "create" in item);
+  let lines = [`Upload files to ${destination}?`];
+  if (overwrite.length !== 0) {
+    lines = [
+      ...lines,
+      "",
+      `${overwrite.length} files will be overwritten:`,
+      ...overwrite.map(({ overwrite }) => `O ${overwrite}`),
```
catloversg: I think the prefix "O", "S", and "C" are more confusing than helpful. For example:

Image

I thought "O" (the uppercase "o") was zero and wondered why there was a zero here.

src/Terminal/commands/upload.ts
```diff
@@ -0,0 +1,140 @@
...
+    ];
+  }
+  if (skipped.length !== 0) {
+    const extensions = [...validScriptExtensions, ...validTextExtensions];
+    const last = extensions.pop() as string;
+    const allValid = extensions.join(", ") + " and " + last;
```
catloversg: This is nice, but using `extensions.join(",")` directly later is simpler.

Image

This looks okay to me. There is no need for fancy spliting and joining just for the word "and".

src/Terminal/commands/upload.ts
```diff
@@ -0,0 +1,140 @@
...
+    lines = [
+      ...lines,
+      "",
+      "Characters * ? [ ] ! \\ ~ | # \" ' and whitespace are not allowed in file paths.",
+      `Only file extensions ${allValid} are allowed.`,
+      "A file name must have at least one character before the extension.",
+      "",
+      `${skipped.length} files will be skipped due to prohibited file paths:`,
+      ...skipped.map(({ badPath }) => `S ${badPath}`),
```
catloversg: Same comment about "S".

src/Terminal/commands/upload.ts
```diff
@@ -0,0 +1,140 @@
...
+    ];
+  }
+  if (create.length !== 0) {
+    lines = [...lines, "", `${create.length} new files will be created:`, ...create.map(({ create }) => `C ${create}`)];
```
catloversg: Same comment about "C".

catloversg (commented):
This is my early feedback. I will review it again later after more testing (I am currently focusing on fixing bugs in other issues).

The goal/scope of this PR looks good to me. I actually like that it does not support .zip files. It may seem like a missing feature at first, but supporting directory uploads is sufficient.

For a niche feature like this, I prefer starting with a minimal set of features and adding more later if there is clear demand.

d0sboots:
Oh cool, I didn't know directory-picking was a thing! That's definitely the way to go. It looks like it's been around on desktop for a long time now, but I see this gem on mobile:
image
XD (still won't be a problem for us)

If need be, we can augment this later with a second button to allow adding individual files. I agree with CatLover that doing the basic thing first is correct, and we'll add more features if/when they seem important.

hexagonrecursion:
@catloversg Thanks. Done

catloversg (approved):

catloversg:
I committed some changes.

PR #2660, opened 2026-04-14
BUGFIX: Intelligence data is incorrectly migrated when Intelligence is not unlocked

catloversg:
MRE: Load the attached save file. It does not have access to Intelligence, but the overview window shows intelligence skill level.

For normal stats, 0 exp means skill level 1. For Intelligence, 0 exp can be either 0 or 1 depending on conditions. Intelligence starts at skill level 0. When entering BN5 or being granted SF5, if the skill level is 0, it gets set to 1. The overview window shows Intelligence when the skill level is not 0.

During migration, without the fix in this PR, the skill level is always set to 1 when exp is 0. This causes the UI to accidentally expose intelligence data.

Another issue is that `gainIntelligenceExp` allows gaining exp when the skill level is greater than 0. This assumes that skill level > 0 means Intelligence is unlocked, which is not always true (e.g., a bug like this one can incorrectly set the skill level).

This PR makes 3 changes:
- Don't set intelligence data if the player has not unlocked it.
- Reset the wrong data created by the bug above.
- Remove the wrong condition in `gainIntelligenceExp`.

This PR is a hotfix, so I kept it simple. There are other minor inconsistencies in how we check whether to show intelligence data, but they are not critical as long as the data is correct. I will address them in follow-up PRs.

Save file:
[v2.8.1_SF1.1_SF10.3.gz](https://github.com/user-attachments/files/26716504/v2.8.1_SF1.1_SF10.3.gz)


src/PersonObjects/Person.ts
```diff
@@ -147,6 +147,10 @@ export abstract class Person implements IPerson {
   }
 
   overrideIntelligence(): void {
+    // Do not set anything if the player has not unlocked Intelligence.
+    if (Player.sourceFileLvl(5) === 0 && Player.bitNodeN !== 5) {
+      return;
```
d0sboots, line 152: Does it make sense to do the same thing as the migration (zeroing out int) in this block, as a safety check?

Without a bug, it is impossible for such code to matter. But if another bug happens, it would reset int on prestige. The downside is that it would also mask the bug.

I'm merging this anyway, while we consider/discuss.

catloversg: We already reset int on BN prestige if the player has not unlocked int.

https://github.com/bitburner-official/bitburner-src/blob/c21d1f44b2cf460fb92a88f17168f65f3473faa5/src/RedPill.tsx#L64-L73

The logic here is questionable, though. I'm writing test cases to check it, especially that else-if; it is likely incorrect. Let's discuss it further in the follow-up PR.

PR #2661, opened 2026-04-14
DEVMENU: Add tool to expose internal objects

catloversg:
To access internal objects, we usually use the "module raid" trick. However, with the current version of Monaco, using this trick in combination with the in-game editor floods the console with errors. Because of this, when testing the latest production builds, I sometimes use the React exploit to access the dev menu instead of module raid. I'm not sure whether that exploit can expose internal objects, and I haven't tested it.

This PR allows access to commonly used internal objects via the dev menu without relying on module raid. It also exposes `Go` and `DarknetState`.

This is a niche use case, and I may be the only one using it. If you prefer not to merge this, I can remove the dev menu changes and only expose those two objects.

### DIFF
--- a/src/DevMenu/Tools.ts
+++ b/src/DevMenu/Tools.ts
@@ -1,0 +1,40 @@
+import { Companies } from "../Company/Companies";
+import { DarknetState } from "../DarkNet/models/DarknetState";
+import { Factions } from "../Faction/Factions";
+import { Go } from "../Go/Go";
+import { Player } from "../Player";
+import { saveObject } from "../SaveObject";
+import { GetAllServers } from "../Server/AllServers";
+
+declare global {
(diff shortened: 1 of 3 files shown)


d0sboots:
I am 90% sure the functionality of this can be accomplished through other means, without spamming console errors. I think the correct place for such code is a pinned post in #expliots-and-other-shenanigains, rather than inside the game.

catloversg:
I will close this PR. I initially planned to remove the dev menu changes and only expose `Go` and `DarknetState`, but on further thought, this does not seem very useful.

Since #1611, I have almost never used `globalThis.Bitburner` as intended. In the dev environment, we typically run custom builds that expose more than what is available in `globalThis.Bitburner`. In the production environment, `globalThis.Bitburner` is not exposed at all. As a result, #1611 is less useful than I initially expected.

PR #2662, opened 2026-04-14
BLADEBURNER: Restore check against available team members

gmcew:
Context and steps to reproduce at [discord](https://discord.com/channels/415207508303544321/415213413745164318/1493469467559329843)
Partial revert of #2541

Post restoration of this `min`, the UI boxes correctly cap at however many teammates have survived/are assigned. From my experimentation, the >0 check is still not necessary - if I missed an edge case I'll restore it too.

`lint`,`format`,`test` and `doc` all completed

### DIFF
--- a/src/Bladeburner/Actions/Operation.ts
+++ b/src/Bladeburner/Actions/Operation.ts
@@ -93,6 +93,8 @@ export const operationSkillSuccessBonus = (inst: Bladeburner) => {
   return inst.getSkillMult(BladeburnerMultName.SuccessChanceOperation);
 };
 
-export function operationTeamSuccessBonus(this: Operation | BlackOperation) {
-  return Math.pow(this.teamCount + 1, 0.05);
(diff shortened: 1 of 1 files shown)


catloversg:
This mitigates the regression caused by #2541, but this regression exposes a larger bug in `teamCount` of Operation and at least 2 bugs in Black Operations.

We can merge this PR if d0sboots wants to mitigate the regression ASAP, but it may be better to wait. I'm testing another PR for fixing the bug in `teamCount` of Operation, which will supersede this PR, along with additional fixes for Black Operations.

gmcew:
It doesn't surprise me there's other problems here; it's not the best explored sub mechanic. I'd rather a larger more effective fix if you've got one in progress. 

I'm keeping it as draft just in case, I'll close it when yours overtakes this.

d0sboots:
I'm happy to wait. BB is a viper's nest.

PR #2665, opened 2026-04-16
TOOL: Add Nix for reproducible local dev environment

66Ton99:
## Summary
- add a `flake.nix` dev shell for local Bitburner source builds
- pin dependencies in `flake.lock` so contributors get a reproducible environment
- provide the core tools needed to work on the repo: `bash`, `git`, and `nodejs_24`

## Why
Setting up a local environment can drift across machines depending on the host package manager, shell setup, and installed Node version. This adds a Nix-based entrypoint that makes the baseline development toolchain reproducible and easier to bootstrap for contributors who already use Nix or Mac, even WSL.

The shell intentionally stays small and focuses on the essentials needed to enter the repo and run the existing project workflows, while keeping the Node version aligned with the repo's current expectations.

## Testing
- ran `nix develop --command bash -lc 'node --version && git --version && bash --version | head -n 1'`
- verified the shell provides `node v24.14.0`, `git 2.53.0`, and `bash`

## Notes
- no runtime game code was changed
- this PR only adds Nix development environment files

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,61 @@
+{
+  "nodes": {
+    "flake-utils": {
+      "inputs": {
+        "systems": "systems"
+      },
+      "locked": {
+        "lastModified": 1731533236,
+        "narHash": "sha256-l0KFg5HjrsfsO/JpG+r7fRrqm12kzFHyUHqHCVpMMbI=",
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "rev": "11707dc2f618dd54ca8739b309ec4fc024de578b",
+        "type": "github"
+      },
+      "original": {
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "type": "github"
+      }
+    },
+    "nixpkgs": {
+      "locked": {
+        "lastModified": 1776169885,
+        "narHash": "sha256-l/iNYDZ4bGOAFQY2q8y5OAfBBtrDAaPuRQqWaFHVRXM=",
+        "owner": "NixOS",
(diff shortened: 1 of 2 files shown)


d0sboots:
I had only vaguely heard of nixos before this point. Afaik, no one else who contributes uses it either. Although the motivation of reproducible builds makes sense, we accomplish that in other ways (specifically, using github workflows to take personal machines out of the equation entirely, since they could be compromised).

The main issue I have with this is it's new infrastructure that I don't think any of the core maintainers will maintain.

66Ton99:
That makes sense. I’m not proposing Nix as required project infrastructure or as a replacement for your current CI/workflow setup. This PR only adds an optional local development shell (flake.nix + flake.lock), and it does not touch runtime game code or change the existing contributor workflow. The shell is intentionally minimal and only provides the basic tools needed to work on the repo. For contributors who already use Nix, it makes local setup faster and more reproducible; for everyone else, it can be ignored completely. 
It’s also not limited to one Linux niche: official Nix installation/support covers Linux, macOS, and Windows via WSL2, so this can help contributors across platforms, not just NixOS users. If the main concern is ownership, I’m happy to handle follow-up updates when the pinned toolchain needs to change. And because this is isolated to dev-environment files, if it ever stops being useful it can be removed cleanly without affecting the game itself.

d0sboots:
I think as long as this is scoped to a thing you primarily would use, it's best to just keep it as files local to your workspace. If I hear demand/interest for this from other parties, then it makes sense to think about keeping them in a shared location.

PR #2666, opened 2026-04-16
BUGFIX: Follow-up to #2660

catloversg:
Context: https://github.com/bitburner-official/bitburner-src/pull/2660#discussion_r3082824399

https://github.com/bitburner-official/bitburner-src/blob/45bce6e45e0fd618194cd535559e774a5ae69f08/src/RedPill.tsx#L64-L70

This else-if has some problems:
- It does not reset (incorrect) int data when destroying WD (it should be an if instead of an if-else).
- It checks `newBitNode` instead of `destroyedBitNode`, which appears to be a bug. Even if the first issue is fixed, this still fails in cases like: starting in BN1 with incorrect int data, then destroying WD and entering BN5 (`newBitNode = 5`). In this case, the check does not reset the data.

Another questionable behavior: exp is not reset when bitfluming from BN5 to BN5 again (without SF5). I'm not sure if we should "fix" this.

I added some Jest tests and highlighted these issues with `// WIP` comments. There is also some commented-out code. I'll remove them after we agree on the expected behavior.

### DIFF
--- a/src/PersonObjects/Person.ts
+++ b/src/PersonObjects/Person.ts
@@ -149,6 +149,14 @@ export abstract class Person implements IPerson {
   overrideIntelligence(): void {
     // Do not set anything if the player has not unlocked Intelligence.
     if (Player.sourceFileLvl(5) === 0 && Player.bitNodeN !== 5) {
+      // WIP: This is not enough. Try disabling the code in RedPill.tsx, enabling this code, then running tests.
+      // At this point, "bitNodeN" is "newBitNode", but we need both "newBitNode" and "destroyedBitNode".
+      // this.skills.intelligence = 0;
+      // this.exp.intelligence = 0;
+      // this.persistentIntelligenceData.exp = 0;
+      // if (Player.bitNodeN === 5 && Player.skills.intelligence === 0) {
+      //   Player.skills.intelligence = 1;
+      // }
       return;
     }
     const persistentIntelligenceSkill = this.calculateSkill(this.persistentIntelligenceData.exp, 1);
--- a/src/PersonObjects/Person.ts
+++ b/src/PersonObjects/Person.ts
@@ -147,16 +147,14 @@ export abstract class Person implements IPerson {
   }
 
   overrideIntelligence(): void {
-    // Do not set anything if the player has not unlocked Intelligence.
+    // Reset intelligence data if the player has not unlocked Intelligence.
+    // Note that this check cannot reset intelligence data in some edge cases (e.g., bitflume from non-BN5 to BN5). This
+    // is an accepted limitation.
+    // For more information, please check https://github.com/bitburner-official/bitburner-src/pull/2666
     if (Player.sourceFileLvl(5) === 0 && Player.bitNodeN !== 5) {
-      // WIP: This is not enough. Try disabling the code in RedPill.tsx, enabling this code, then running tests.
-      // At this point, "bitNodeN" is "newBitNode", but we need both "newBitNode" and "destroyedBitNode".
-      // this.skills.intelligence = 0;
-      // this.exp.intelligence = 0;
-      // this.persistentIntelligenceData.exp = 0;
-      // if (Player.bitNodeN === 5 && Player.skills.intelligence === 0) {
-      //   Player.skills.intelligence = 1;
-      // }
+      this.skills.intelligence = 0;
+      this.exp.intelligence = 0;
+      this.persistentIntelligenceData.exp = 0;
       return;
     }
     const persistentIntelligenceSkill = this.calculateSkill(this.persistentIntelligenceData.exp, 1);
--- a/src/RedPill.tsx
+++ b/src/RedPill.tsx
@@ -63,7 +63,10 @@ export function enterBitNode(
 
   if (!isFlume) {
     giveSourceFile(destroyedBitNode);
-  } else if (Player.sourceFileLvl(5) === 0 && newBitNode !== 5) {
+  }
+  // WIP
+  if (Player.sourceFileLvl(5) === 0 && destroyedBitNode !== 5) {
+    // if (Player.sourceFileLvl(5) === 0 && newBitNode !== 5) {
     Player.skills.intelligence = 0;
     Player.exp.intelligence = 0;
     Player.persistentIntelligenceData.exp = 0;
--- a/src/RedPill.tsx
+++ b/src/RedPill.tsx
@@ -35,9 +35,6 @@ function giveSourceFile(bitNodeNumber: number): void {
     }
   } else {
     Player.sourceFiles.set(bitNodeNumber, 1);
-    if (bitNodeNumber === 5 && Player.skills.intelligence === 0) {
-      Player.skills.intelligence = 1;
-    }
     dialogBoxCreate(
       <>
         You received a Source-File for destroying a BitNode!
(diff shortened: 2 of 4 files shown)


src/RedPill.tsx
```diff
@@ -63,7 +63,10 @@ export function enterBitNode(
 
   if (!isFlume) {
     giveSourceFile(destroyedBitNode);
-  } else if (Player.sourceFileLvl(5) === 0 && newBitNode !== 5) {
+  }
+  // WIP
+  if (Player.sourceFileLvl(5) === 0 && destroyedBitNode !== 5) {
```
d0sboots: This check is definitely wrong now. It no longer removes int when you bitflume away from BN5 (`sourceFileLvl(5) === 0 && destroyedBitNode === 5`), which was the primary purpose of this code.

Looking at this "if", I think all of this simply should not be done here. We have the duplicate check in `overrideIntelligence` that I commented on earlier. That is called (transitively) via `prestigeSourceFile`, which happens a bit lower in this function, as well as elsewhere.

Instead, IMO we should just set int to 0 in `overrideIntelligence` before returning. Similarly, we should do the "set skill to 1" check there as well - it logically makes sense there, being called "overrideIntelligence." (That code is actually pointlessly duplicated within RedPill.tsx right now.) The one questionable part is that `overrideIntelligence` takes effect for sleeves too, which seems correct but maybe their skill isn't supposed to start at 1?

d0sboots: Oh, I didn't quite answer your question - IMO it is correct for "bitflume from BN5 to BN5" to preserve int. If someone really wants to reset their int, they can bitflume to a different node first.

catloversg: I updated the check in RedPill.tsx. I tried your suggestion to reset in `overrideIntelligence`, but it does not seem to have enough information for our needs. Please check my new WIP comments and tests.

test/jest/Netscript/Singularity.test.ts
```diff
@@ -646,3 +670,158 @@ describe("purchaseProgram", () => {
...
+
+      // Bitflume from non-BN5 to non-BN5.
+      expect(Player.bitNodeN).toStrictEqual(1);
+      manuallySetIntelligenceExp(50);
+      expectIntelligenceExp(50);
+      ns.singularity.b1tflum3(1);
+      // Reset intelligence data
+      expectInitialIntelligenceData();
+
+      // Bitflume from non-BN5 to BN5.
```
d0sboots, line 758: I don't think this case is critical. Yes, it's nice if it works this way for safety, but this is strictly cleaning up erroneous behavior *in the edge case of fluming to BN5* - it's not worth contorting the code to get it to work.

What we *do* want to test is that bitfluming to BN5 in a normal circumstance (initialIntelligenceData) results in 0 int exp. (1 skill)

test/jest/Netscript/Singularity.test.ts
```diff
@@ -646,3 +670,158 @@ describe("purchaseProgram", () => {
...
+      // Destroy WD of non-BN5 and jump to non-BN5.
+      expect(Player.bitNodeN).toStrictEqual(1);
+      manuallySetIntelligenceExp(50);
+      expectIntelligenceExp(50);
+      setUpBeforeDestroyingWD();
+      ns.singularity.destroyW0r1dD43m0n(1);
+      // Reset intelligence data
+      expectInitialIntelligenceData();
+
+      // Destroy WD of non-BN5 and jump to BN5.
```
d0sboots: Same notes as above for this case.

src/PersonObjects/Person.ts
```diff
@@ -149,6 +149,14 @@ export abstract class Person implements IPerson {
   overrideIntelligence(): void {
     // Do not set anything if the player has not unlocked Intelligence.
     if (Player.sourceFileLvl(5) === 0 && Player.bitNodeN !== 5) {
+      // WIP: This is not enough. Try disabling the code in RedPill.tsx, enabling this code, then running tests.
+      // At this point, "bitNodeN" is "newBitNode", but we need both "newBitNode" and "destroyedBitNode".
```
d0sboots: I don't see why we need both? The old code used newBitnode, and I consider it to be essentially correct.

src/PersonObjects/Person.ts
```diff
@@ -149,6 +149,14 @@ export abstract class Person implements IPerson {
...
+      // this.skills.intelligence = 0;
+      // this.exp.intelligence = 0;
+      // this.persistentIntelligenceData.exp = 0;
+      // if (Player.bitNodeN === 5 && Player.skills.intelligence === 0) {
```
d0sboots: This code needs to be after the if block, it can never trigger if it's inside.

...OK, how did this ever work, actually. The lines that set intelligence to 1 are in RedPill.tsx, *before* it calls prestigeSourceFile, which calls overrideIntelligence. All paths here (that are don't early-return, but you can verify that the early-return is logically exclusive with setting the skill to 1) unilaterally reset `this.skills.intelligence`. So... how was that line ever doing anything?

Looking into it more, the answer is "it wasn't." The skill is being set to 1 by the clamp inside `calculateSkill` (this was the cause of the bug you fixed where an early return was needed), and the other code is wholly useless now.

test/jest/Netscript/Singularity.test.ts
d0sboots: What we *also* want to see is that bitfluming (or installing augs) while you are in non-BN5 will clean up erroneous int.

Edit: The "installing augs" case is why I wanted it to be in `overrideIntelligence`.

catloversg: I'm confused by these two comments.

Let's ignore the `overrideIntelligence` and other comments at the moment and focus on what we should test and the expected behavior.

In the first comment, you said it's nice to have "Bitflume from non-BN5 to BN5" test, but it's not worth to change the code to make it work. In the second comment, you want to clean up erroneous int when bitfluming or installing augs while you are in non-BN5. "bitfluming while you are in non-BN5" is exactly the "Bitflume from non-BN5 to BN5" test.

```ts
      // Bitflume from non-BN5 to BN5.
      expect(Player.bitNodeN).toStrictEqual(1);
      manuallySetIntelligenceExp(50);
      expectIntelligenceExp(50);
      ns.singularity.b1tflum3(5);
      // Reset intelligence data and skill = 1
      expectIntelligenceExp(0);
```
In short, is this test required or not?

If the int data is corrupted in non-BN5 and you bitflume to BN5, the check in the reset code must have access to "destroyedBitNode".

In `overrideIntelligence`, we don't have access to "destroyedBitNode", so we must explicitly accept the fact that we cannot check the case of "Bitflume from non-BN5 to BN5". In other words, if we reset in `overrideIntelligence`, there is a specific case that we cannot reset corrupted int data. I want to make sure this is the behavior that you want, but I don't see why we should accept that limitation.

d0sboots: Yes, you are correct about what I want. I am reasoning possibly backwards from the constraints of the code, rather than creating a strict spec of expected behavior and going from there.

In broad strokes, the properties I want are: 
* No duplicated code, or it is duplicated as little as possible
* In line with the first goal, dealing with int should be in a single place, not scattered in multiple places (as much as possible) 
* The "correct" scenarios should work (obviously) 
* BN5 flume -> BN5 shouldn't reset int, although that's nice to have as opposed to required
* Incorrect int is usually caught and fixed before very long. Note that this is the *lowest* priority, since if this happens it indicates a different bug in our code (or savegame editing etc). So it's fine if it misses some edge cases (such as flume from BN1 to BN5).

d0sboots: To hopefully clear up the confusion: My second comment about having a "not in BN5" test was referring to installing augs not in BN5 (newBitnode = BN1), or doing bitflume from BN1 to BN2. My point was that having newBitnode = BN5 *specifically* is an edge case that IMO doesn't need to clean up invalid int. (It would be nice, but not at the cost of sacrificing the other properties.)

catloversg: In order to reset when installing augs, I have to move the call of `overrideIntelligence` from `prestigeSourceFile` to `prestigeAugmentation` in the latest commit (there is a new test for this case). This looks a bit weird to me, to be honest.

PR #2667, opened 2026-04-16
BUGFIX: calculateExp throws errors in edge cases

catloversg:
MRE: Load a clean new save file, set SF12 to 35842 or higher (`Player.sourceFiles.set(12, 35842)`), then bitflume to BN12.

```
clampNumber.ts:16 Uncaught Error: NaN passed into clampNumber()
    at clampNumber (clampNumber.ts:16:78)
    at calculateSkill (skill.ts:16:81)
    at calculateExp (skill.ts:24:21)
    at setInitialExpForPlayer (Prestige.ts:78:137)
    at prestigeSourceFile (Prestige.ts:370:3)
    at enterBitNode (RedPill.tsx:75:64)
    at onClick (PortalModal.tsx:120:61)
    at HTMLUnknownElement.callCallback (react-dom.development.js:3945:14)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:3994:16)
    at invokeGuardedCallback (react-dom.development.js:4056:31)
```

In BN12, with a high SF12 level, the `mult` passed to `calculateExp` can be 0, so `value` can be Infinity. Before #2274, `value` would be clamped to `Number.MAX_VALUE` in that case. However, after the change in #2274, `calculateSkill` will be called in the new if block. With exp = Infinity and mult = 0, `Math.floor(mult * (32 * Math.log(exp + 534.6) - 200))` evaluates to NaN.

### DIFF
--- a/src/PersonObjects/formulas/skill.ts
+++ b/src/PersonObjects/formulas/skill.ts
@@ -5,6 +5,11 @@ import { clampNumber } from "../../utils/helpers/clampNumber";
  * stat level. Stat-agnostic (same formula for every stat)
  */
 export function calculateSkill(exp: number, mult = 1): number {
+  // Mult can be 0 in BN12 when the player has a very high SF12 level. In this case, the skill level will never change
+  // from its initial value (1 for most stats, except intelligence).
+  if (mult === 0) {
+    return 1;
+  }
   const value = Math.floor(mult * (32 * Math.log(exp + 534.6) - 200));
   return clampNumber(value, 1);
 }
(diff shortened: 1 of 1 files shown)


d0sboots:
As the stacktrace shows, this is not the only place to apply the fix. `calculateSkill` is the function that is throwing; it should always return a value when passed a mult of 0. (Probably the correct value to return is 1, since with a mult of 0 skill can never change from the initial value.)

However, the line you changed still needed to be changed.

PR #2670, opened 2026-04-16
MISC: Clear recent scripts when installing augmentations

### LINKED ISSUE #2669: sometimes the list of recently finished scripts is missing entries
Under specific circumstances, scripts do not show up in the list of recently finished scripts, making it difficult (impossible?) to access their logs.

#### steps to reproduce

1. load or reload a save file
2. choose a small non-negative integer `n`, less than the "Recently killed script size" (see game Options)
3. run any script (due to step 1, this should have PID 1)
4. repeat the previous step `n` times to generate some additional PIDs
5. go to the Options and "Soft Reset" the game (alternatively you can install augmentations if you have some purchased)
6. (optional) go to "Active Scripts" → "Recently Killed" and observe that the scripts you ran earlier are still listed there
7. run any script (it has PID 1 again) and wait for it to terminate
8. check the recently finished scripts: The script from step 7 is not listed.

Step 7 can be repeated `n` times without having anything show up in the list of recently finished scripts. After that the scripts have new PIDs and show up again.

#### What is going on?
Installing augmentations (or soft resetting) resets the PID counter to 1, but does not clear the list of recently finished scripts. The [list of recently finished scripts identifies scripts by PID for de-duplication](https://github.com/bitburner-official/bitburner-src/blob/stable/src/Netscript/RecentScripts.ts), thereby preventing any scripts with a PID which already exists in the list from being listed.

#### scripts
any working script will do, you can use this one:
```javascript
export async function main() {}
```
---
operating system: Linux
browser: Firefox 149.0.2 (64-Bit)

[example_savefile.zip](https://github.com/user-attachments/files/26802401/example_savefile.zip)

Mathekatze:
# MISC: clear recent scripts when installing augmentations

Also refactors the PID generation code.

fixes #2669 

# bug fix

## before fix
before_pre_reset

soft reset
before_post_reset
The list of recently finished scripts only shows the two instances from before the soft reset and the last one that was ran:
before_list

## after fix
after_pre_reset

soft reset
after_post_reset
Now all script instances have different PIDs and the list of recently finished scripts shows all five of them:
after_list

## test script
```javascript
/** @param {NS} Bitburner */
export async function main(Bitburner) {
  const rggrHistory = Bitburner.getRecentScripts();
  const rghPIDs = rggrHistory.map(grInstance => grInstance.pid);
  const hMinPID = Math.min(...rghPIDs);
  const hMaxPID = Math.max(...rghPIDs);
  const hOwnPID = Bitburner.pid;
  if(hMaxPID >= hOwnPID) {
    Bitburner.tprint(`WARNING: The PIDs have been reset!`);
    const nMargin = hMinPID - hOwnPID - 1;
    if(nMargin > 0) {
      Bitburner.tprint(`INFO: You should be able to run at least ${nMargin} scripts without PID collisions.`);
    } else {
      Bitburner.tprintf(`WARNING: There might be some PID collisions!`);
    }
  } else {
    Bitburner.tprint(`INFO: The PIDs seem to be fine.`);
  }
  if(hOwnPID in rghPIDs) {
    Bitburner.tprint(`SUCCESS: PID collision: There already is a script with PID ${hOwnPID} in the history!`);
    Bitburner.print(`WARNING: The log containing this message may get lost.`);
  }
}
```

### DIFF
--- a/src/Netscript/Pid.ts
+++ b/src/Netscript/Pid.ts
@@ -9,29 +9,28 @@ export function generateNextPid(): number {
 
   // Cap the number of search iterations at some arbitrary value to avoid
   // infinite loops. We'll assume that players won't have a million running scripts.
-  for (let attemptCounter = 0; attemptCounter < 1e6; ++attemptCounter) {
+  for (let attemptCounter = 0; attemptCounter < 1e6; ++attemptCounter, ++pidCandidate) {
+    // ensure the candidate PID is a safe integer
+    if (pidCandidate >= Number.MAX_SAFE_INTEGER) {
+      pidCandidate = 1;
+    }
+    // ensure the PID is not in use
+    if (workerScripts.has(pidCandidate)) {
+      continue;
+    }
     /* Also check the recentScripts: The function `AddRecentScript` de-duplicates scripts by PID.
        * Therefore scripts which re-use a PID from that list do not show up the list of recent scripts
        * upon termination (if the previous script with that PID is still in the list at that point).
        * Such a PID re-use could happen after installing augmentations, which resets the PID counter
        * but does not clear the list recent scripts. */
-    if (!workerScripts.has(pidCandidate)
-         && recentScripts.some((r) => (r.runningScript.pid === pidCandidate))) {
-      // found a PID that's not in use
-      // set the counter to the successor of the found PID, wrapping back to 1 when it gets too high
-      pidCounter = pidCandidate + 1;
-      if (pidCounter >= Number.MAX_SAFE_INTEGER) {
-        pidCounter = 1;
-      }
-      return pidCandidate;
+    if(recentScripts.some((r) => (r.runningScript.pid === pidCandidate))) {
+      continue;
     }
-
-    ++pidCandidate;
-    if (pidCandidate >= Number.MAX_SAFE_INTEGER) {
-      pidCandidate = 1;
-    }
+    // found a PID that's not in use
+    pidCounter = pidCandidate + 1;
+    return pidCandidate;
   }
-  // ran out of attempts without finding a valid unused pid :-(
+  // ran out of attempts without finding an unused PID :-(
   return -1;
 }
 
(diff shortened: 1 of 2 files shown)


d0sboots:
This is the wrong fix. The recently killed scripts should be flushed on actions that reset the pid counter, instead.

catloversg:
I think it's fine to have access to the list of recent scripts after installing augs, but I agree that we should clear that list instead of adding a new check to `generateNextPid`.

By the way, the refactor of `generateNextPid` looks good to me. It's cleaner than current code (e.g., with the `i < 1e6` condition, the `i === Number.MAX_SAFE_INTEGER - 1` check is dead code), so we can keep most of it after removing the new check of `recentScripts`.

Mathekatze:
Thanks for your replies! I have removed the additonal check now. Instead, `resetPidCounter` clears the list of recently finished scripts now.

src/Netscript/Pid.ts
```diff
@@ -1,40 +1,37 @@
...
+  return -1;
 }
 
 export function resetPidCounter(): void {
+  /* The function `AddRecentScript` de-duplicates scripts by PID.
+   * Therefore the list has to be cleared when resetting the PID counter.
+   * Otherwise scripts which re-use a PID from the list of recent scripts
+   * do not show up there when they finish (if the previous script with
+   * that PID is still in the list at that point). */
+  recentScripts.splice(0);
```
ficocelliguy: This may make more sense to do where we kill all scripts during "prestige" rather than here - the recently killed scripts page isn't really related to the pid counter for scripts

ficocelliguy (changes requested):

src/Netscript/Pid.ts
catloversg: Just in case you wonder what fico means, check the callers of `resetPidCounter` in `src/Prestige.ts`. `prestigeSourceFile` correctly clears `recentScripts`. You need to do that for `prestigeAugmentation`.

PR #2673, opened 2026-04-18
UI: Ensure intelligence override is a positive integer

catloversg:
This constraint is enforced by `validateBitNodeOptions` when using NS APIs, but the documentation incorrectly states that it must be a non-negative integer. The UI also does not enforce this rule.

Capture

Allowing a value of 0 is useful if the player wants to "pretend" that they have not unlocked Intelligence, but this use case is too niche to support.


d0sboots:
I don't think it's actually super niche, but I'm fine not supporting it if it causes annoying edge cases.

PR #2680, opened 2026-04-21
API: Remove completion from GraftingWork.

jeek:
API cost on ns.singularity.getCurrentWork() is .5GB. Currently, in the event that the work is Grafting, an additional "completion" promise is added to the result. All other promise-generating functions are both standalone and at least 1GB.

This patch removes completion from grafting work. It may be suitable for later addition as a standalone singularity call, albeit, one that would duplicate ns.grafting.waitForOnGoingGrafting() that is already in the game at 1GB.

### DIFF
--- a/src/Work/GraftingWork.tsx
+++ b/src/Work/GraftingWork.tsx
@@ -24,14 +24,6 @@ export class GraftingWork extends Work {
   augmentation: AugmentationName;
   unitCompleted: number;
   unitRate: number;
-  completionPromisePair: PromisePair<void> = { promise: null, resolve: null };
-
-  get completion(): Promise<void> {
-    if (!this.completionPromisePair.promise) {
-      this.completionPromisePair.promise = new Promise((r) => (this.completionPromisePair.resolve = r));
-    }
-    return this.completionPromisePair.promise;
-  }
 
   constructor(params?: GraftingWorkParams) {
     super(WorkType.GRAFTING, params?.singularity ?? true);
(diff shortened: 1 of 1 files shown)


catloversg:
First, I'll add context to this PR. You did not state it in the description, but the main goal of this PR is to make the object returned from `getCurrentWork` clonable by `structuredClone`, which in turn enables RAM dodging this API via ports.

Second, this change introduces a new "rule" in the API design: we will *never* add promises to getTask-like APIs (`Singularity.getCurrentWork` and `Sleeve.getTask`), since promises are unclonable by `structuredClone`. This rule and its rationale should be documented in the PR description.

Third, you don't want to change your RAM-dodging code to support this `completion` promise, but what about `nextCompletion` in sleeve tasks? Should we remove `nextCompletion` in all sleeve tasks? What about the suggestion of adding `nextCompletion` to sleeve crime task on Discord?

In the current state, I'll veto this change. I'll accept it under these conditions:
- Implement the suggested API to get the `completion`/`nextCompletion` promise (e.g., `getNextCompletionOfTask` or `waitForNextCompletionOfTask`). This API covers the current `completion` and future `nextCompletion` of other tasks (as I said, there were suggestions of adding more promises, not fewer).
- It is not coherent to remove `completion` in the normal task (the grafting task) while retaining `nextCompletion` in sleeve tasks. In your specific case, your RAM-dodging code will still crash if you call `Sleeve.getTask` and your sleeves are doing Bladeburner tasks, which have `nextCompletion`. Therefore, the sleeve API should be updated as follows:
  - Remove `nextCompletion` in sleeve tasks.
  - Add an API to get the `nextCompletion` promise of sleeve tasks in the Sleeve namespace.

You don't have to do all these changes in this PR. They are preconditions for adding that new "rule".

d0sboots:
We are planning on removing "completion", but only to rename it to "nextCompletion". See #2684 for the full plan here. Although people will continue to trip over this (possibly even more than before), hopefully it will become more obvious why.

PR #2690, opened 2026-04-25
BUGFIX: DarkscapeNavigator program is not granted on BN prestige when having SF15

catloversg:
MRE:
- Load a clean save file.
- Set SF15.1
- Bitflume and check the home server.

We do this in `prestigeAugmentation`, but not in `prestigeSourceFile`.


d0sboots:
I've long been thinking that we should make a shared helper *specifically in this file* that covers the common bits of both, so that what remains can just be the unique parts...

PR #2696, opened 2026-04-26
HACKNET: fix application of limits in cost functions

lucebac:
Hacknet Node/Server cost calculation function fail to correctly calculate the upgrade cost under certain conditions: if the node/server is not at max level/ram/core/cache and the amount of upgrades surpasses the maximum (i.e. current cores + new cores > max_cores), the cost will incorrectly be calculated event for stats which are not allowed. See the following example:
```js
  ns.print(`node id: 0, node level: ${ns.hacknet.getNodeStats(0).level}, cost for 5_000 levels: ${ns.hacknet.getLevelUpgradeCost(0, 5_000)}`);
  ns.print(`node id: 1, node level: ${ns.hacknet.getNodeStats(1).level}, cost for 5_000 levels: ${ns.hacknet.getLevelUpgradeCost(1, 5_000)}`);
}
```
Which will output this:
```
node id: 0, node level: 200, cost for 5_000 levels: Infinity
node id: 1, node level: 1, cost for 5_000 levels: 1.5738957360987797e+89
```

I've fixed this edge case for all hacknet cost calculation functions and also added a few tests which should hopefully catch the edge cases in the future.

### DIFF
--- a/src/Hacknet/formulas/HacknetNodes.ts
+++ b/src/Hacknet/formulas/HacknetNodes.ts
@@ -16,7 +16,7 @@ export function calculateLevelUpgradeCost(startingLevel: number, extraLevels = 1
     return 0;
   }
 
-  if (startingLevel >= HacknetNodeConstants.MaxLevel || startingLevel + extraLevels > HacknetNodeConstants.MaxLevel) {
+  if (startingLevel + sanitizedLevels > HacknetNodeConstants.MaxLevel) {
     return Infinity;
   }
 
@@ -39,7 +39,7 @@ export function calculateRamUpgradeCost(startingRam: number, extraLevels = 1, co
 
   if (
     startingRam >= HacknetNodeConstants.MaxRam ||
-    startingRam * Math.pow(2, extraLevels) > HacknetNodeConstants.MaxRam
+    startingRam * Math.pow(2, sanitizedLevels) > HacknetNodeConstants.MaxRam
   ) {
     return Infinity;
   }
@@ -69,7 +69,7 @@ export function calculateCoreUpgradeCost(startingCores: number, extraLevels = 1,
     return 0;
   }
 
-  if (startingCores >= HacknetNodeConstants.MaxCores || startingCores + extraLevels > HacknetNodeConstants.MaxCores) {
+  if (startingCores >= HacknetNodeConstants.MaxCores || startingCores + sanitizedCores > HacknetNodeConstants.MaxCores) {
     return Infinity;
   }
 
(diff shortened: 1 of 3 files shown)


src/Hacknet/formulas/HacknetNodes.ts
```diff
@@ -16,7 +16,7 @@ export function calculateLevelUpgradeCost(startingLevel: number, extraLevels = 1
     return 0;
   }
 
-  if (startingLevel >= HacknetNodeConstants.MaxLevel) {
+  if (startingLevel >= HacknetNodeConstants.MaxLevel || startingLevel + extraLevels > HacknetNodeConstants.MaxLevel) {
```
catloversg: Do we really need the first check? It is redundant if `extraLevels` is non-negative. Technically, callers can pass a negative value, but given how the rest of the function behaves, a negative `extraLevels` always results in 0.

I think other functions behave the same way.

test/jest/Hacknet.test.ts
```diff
@@ -0,0 +1,135 @@
...
+    it("bails out for NaN", () => {
+      expect(calculateLevelUpgradeCostNode(1, NaN, 1)).toBe(0);
+    });
+    it("fails for extraLevels < 1", () => {
+      expect(calculateLevelUpgradeCostNode(1, -1, 1)).toBe(0);
+      expect(calculateLevelUpgradeCostNode(1, 0, 1)).toBe(0);
+      expect(calculateLevelUpgradeCostNode(1, 0.4, 1)).toBe(0);
+      expect(calculateLevelUpgradeCostNode(1, 0.5, 1)).toBeGreaterThan(0);
+      expect(calculateLevelUpgradeCostNode(1, 0.9999999999999999, 1)).toBeGreaterThan(0);
+      expect(calculateLevelUpgradeCostNode(1, 1, 1)).toBeGreaterThan(0);
```
catloversg:
```js
      expect(calculateLevelUpgradeCostNode(1, 0.5, 1)).toBeGreaterThan(0);
      expect(calculateLevelUpgradeCostNode(1, 0.9999999999999999, 1)).toBeGreaterThan(0);
      expect(calculateLevelUpgradeCostNode(1, 1, 1)).toBeGreaterThan(0);
```
These tests are correct, but why are they under "fails for extraLevels < 1"?

Values like 0.5 and 0.9999999999999999 are rounded up to 1, so the cost is greater than 0. They do not seem like failures. The last test is clearly not a failure.

src/Hacknet/formulas/HacknetNodes.ts
lucebac: I've tested the behavior of the functions with the conditions simplified to only the second condition and the test results did indeed not change. Adjusted all the functions to the simpler condition.

test/jest/Hacknet.test.ts
lucebac: No specific reason for that. I believe I rephrased the test title and just didn't check whether all steps still fit. I've split the test up and named the new sections appropriately.

PR #2705, opened 2026-04-29
CODEBASE: Add comments explaining redundant check in product calculation

catloversg:
This check is only relevant when processing materials, where the total size of output materials may differ from the total size of input materials. For products, the output size always equals the total size of the input materials, so the check is unnecessary.

https://github.com/bitburner-official/bitburner-src/blob/c72a2928eb6fa50678fa4793ea8cef2cef1a2c5a/src/Corporation/Product.ts#L208-L212

https://github.com/bitburner-official/bitburner-src/blob/c72a2928eb6fa50678fa4793ea8cef2cef1a2c5a/src/Corporation/Division.ts#L876-L886

`netStorageSize` is always 0 or a very small value (due to floating-point imprecision).
- Pharmaceutical: -6.938893903907228e-18
- Robotics: 5.551115123125783e-17
- Real Estate: 2.7755575615628914e-17

### DIFF
--- a/src/Corporation/Division.ts
+++ b/src/Corporation/Division.ts
@@ -873,6 +873,18 @@ export class Division {
           }
           prod *= corpConstants.secondsPerMarketCycle * marketCycles;
 
+          //Calculate net change in warehouse storage making the Products will cost
+          let netStorageSize = product.size;
+          for (const [reqMatName, reqQty] of getRecordEntries(product.requiredMaterials)) {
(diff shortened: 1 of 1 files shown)


d0sboots:
Although this is true, I don't think it's a *requirement* of products to act this way. So, I'm not sure that we should remove the check.

(I.e., they could produce more than the size of the input, potentially.)

PR #2706, opened 2026-04-29
UI: Add backdoor status to "scan-analyze"

abbyintheattic:
In my opinion, backdoor status should appear on `scan-analyze`. I copied the code that displayed backdoor status on `analyze` into the code for `scan-analyze` to achieve this goal.

As instructed by CONTRIBUTING.md, I ran `npm run format` and it made some changes that didn't seem to affect the functionality in testing.

### DIFF
--- a/src/Terminal/Terminal.ts
+++ b/src/Terminal/Terminal.ts
@@ -649,7 +649,12 @@ export class Terminal {
       } else {
         this.print(`${infoPrefix}Root Access: ${hasRoot}` + "\n");
       }
-      this.print(`${infoPrefix}RAM: ${formatRam(server.maxRam)}` + "\n");
+      this.print(
+        `${infoPrefix}RAM: ${formatRam(server.maxRam)}` +
+          ", Backdoor: " +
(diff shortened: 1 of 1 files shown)


catloversg:
For context, we intentionally keep the built-in CLI commands minimal. The goal is for them to be just usable enough for simple cases, but limited enough to encourage players to build their own tools.

This change is not too bad, so I won't veto it. d0sboots will make the call for it.

My (minor) concern is feature creep. Once we allow improvements to a tool, it becomes harder to reject similar improvements later. For example, permitting the addition of backdoor status while rejecting other properties creates inconsistency.

d0sboots:
I think I'm going to veto this, both for the reason catlover mentioned and because scan-analyze prints a recursive tree, so every line that's added also makes it harder to see the tree structure. I'm not sure that's backdoor status, which is only semi-frequently useful, is enough to overcome this problem. This could be solved with command-line switches, but that's *definitely* getting into the territory of "the base tools are awkward to encourage building your own."

PR #2711, opened 2026-05-01
RELEASE: 3.0.0

Snarling:
Changes for stable release version
* Added 3.0.0 changelog to changelog.md doc file
* Updated info in constants for 3.0.0 non-dev release


Snarling:
@catloversg @d0sboots Test failure is due to differing behavior in the `clampNumber` function in utils/helpers/clampNumber.ts based on the value of CONSTANTS.isDevBranch. When true it throws an error, when false it will return the minimum value.

The test expects an error to be thrown for calculateExp(NaN) regardless of whether we're on dev branch or not, and that function ends up calling `clampNumber(NaN, 0)`. On dev branch this throws an error, on stable release it would return the value 0. I'm not sure why we would want different behavior in this function based on whether we're on a dev release but this seems incorrect to me.

Only other thing that gave me pause in preparing the release was 24 vulnerabilities shown during `npm ci`, many of them are for dev dependencies but not all of them. 11 of them have auto-fixes with `npm audit fix` but I didn't apply that in case there was some breaking change (even though npm audit fix is supposed to avoid breaking changes). I'll leave it up to y'all as far as which of these (if any) we need to address before release.

PR #2715, opened 2026-05-02
Update Stable Branch to 3.0.0 Release Build

Snarling:
Updating stable using a PR instead of just deleting the stable branch and recreating it based on the release version, which is what I had been doing in the past for stable branch updates.


Snarling:
Squash and merge is usually how we handle merging PRs to dev, but to merge the dev changes onto stable I'm doing a merge commit instead. I think squashing all the commits from dev might cause complications the next time when updating stable.

PR #2721, opened 2026-05-02
DNET: Remove TS type annotation from doc example script

ficocelliguy:
This type annotation causes an error if the player copies this as-is into a JS script.


hexagonrecursion:
* Is the example even necessary? In 3.0 --tail is automatically added to autocomplete options. See https://github.com/bitburner-official/bitburner-src/pull/2103

ficocelliguy:
it's not strictly needed, no, but I partially added it to make autocomplete more discoverable to players

PR #2722, opened 2026-05-03
DOC: Update list of RFA community tools

CTNOriginals:
The remote-api page did not yet include BitburnerGoFilesync which is the first change I committed.

When I was editing the description of each tool underneath the links, I thought of a cleaner way to describe each tool which will also be a bit more informative to the users that want to chose a tool.
Each description for each tool is taken from their README which was the first line in all of them.

If the original way is preferred over the new one, i can revert it as i did also commit the change i would have made [without reformatting the links](https://github.com/bitburner-official/bitburner-src/commit/853c60714353b6b21a5724ba162bfdc58262e55e).

Here is a preview of what it looks like in game (not entirely up to date with actual content):
image


src/Documentation/doc/en/programming/remote_api.md
```diff
@@ -13,11 +13,10 @@ All these tools support synchronizing scripts to Bitburner and transpiling TypeS
 
 Links:
 
-- https://github.com/bitburner-official/typescript-template
-- https://github.com/Tanimodori/viteburner
-- https://github.com/shyguy1412/bb-external-editor
-
-`typescript-template` has a small set of options and features. Its simplicity is by design. `viteburner` and `bb-external-editor` have more fancy features.
+- [typescript-template](https://github.com/bitburner-official/typescript-template): The official template for synchronizing Typescript/Javascript from your computer to the game.
```
catloversg: Optional: Remove the "official" word. It's not wrong, but it may make players think this tool is "better" than others (official vs non-official).

src/Documentation/doc/en/programming/remote_api.md
```diff
@@ -13,11 +13,10 @@ All these tools support synchronizing scripts to Bitburner and transpiling TypeS
 
 Links:
 
-- https://github.com/bitburner-official/typescript-template
-- https://github.com/Tanimodori/viteburner
-- https://github.com/shyguy1412/bb-external-editor
-
-`typescript-template` has a small set of options and features. Its simplicity is by design. `viteburner` and `bb-external-editor` have more fancy features.
```
catloversg, line 20: This line should be kept and updated. The new description does not exactly replace this "TLDR".

catloversg (commented):
One thing to note: I saw the order suggestion on Discord, and I'm fairly against it. I know the intention of this new order is "simple, simple, complex, complex", not "tool foo is better than tool bar". However, if we use a "meaningful" order, we also imply that the list has a specific order. If another contributor creates tool baz and they want to change the order with this argument: "My tool is absolutely the best, and the list order must reflect it", what are we going to do?

My point is that the list here has no particular order, and that's intentional. I would even argue that we should say "Links (in no particular order):". Technically, back when I added this list, I ordered it roughly based on creation time. It may not be the best approach, but it clearly shows we don't specifically recommend anything.

CTNOriginals:
On the note of the order, I can indeed also see the issue you explained happening down the line, and if not the proposed order, i think it should indeed be the order of creation of the tool.

CTNOriginals:
im having a stroke trying to make these changes on my windows machine... so i will do them instead tomorrow.

note to self, changes:
```md
## Community tools

All these tools support synchronizing scripts to Bitburner and transpiling TypeScript/JSX to JavaScript. Note that Bitburner has native support for TypeScript/JSX.

Links:

- [typescript-template](https://github.com/bitburner-official/typescript-template): A template for synchronizing Typescript/Javascript from your computer to the game.
- [viteburner](https://github.com/Tanimodori/viteburner): Daemon tools of bitburner using vite for script transform, file syncing, RAM monitoring and more!
- [bb-external-editor](https://github.com/shyguy1412/bb-external-editor): This tool uses esbuild to transpile and bundle your scripts. It supports JS, TS and React as well as importing from any browser-compatible npm library out of the box.
- [BitburnerGoFilesync](https://github.com/CTNOriginals/BitburnerGoFilesync): A standalone binary cli tool that doesn't require any setup or third party libraries. It is designed to be very minimal and easy to use out of the box.

`typescript-template` and `BitburnerGoFilesync` both have a small set of options and features, their simplicity is by design.  
`viteburner` and `bb-external-editor` have more fancy features and may offer more control for specific use cases.
```

package-lock.json
```diff
@@ -6,7 +6,7 @@
   "packages": {
     "": {
       "name": "bitburner",
-      "version": "3.0.0",
+      "version": "3.0.1",
```
d0sboots: Revert this file (it won't affect anything you're doing). We'll handle it in a PR of it's own.

PR #2723, opened 2026-05-03
UI: Show hints of Gang mechanic in pre-endgame

catloversg:
Context: #2604.

This PR is the same as #2605, but for gang.

Pre-endgame:

pre-endgame-1

pre-endgame-2

Errors:

error-1

error-2

error-3

error-4


d0sboots:
This looks good. However, I'm sitting on bigger and/or risky stuff for about a week in case we need a 3.0.1. Reply again if this one gets lost.

d0sboots:
I think we will have 3.0.1 before too long, but at the same time I don't think there's anything so urgent popping up that I'm fine with continuing regular submits.

PR #2724, opened 2026-05-03
BUGFIX: Player can switch tabs without losing focus on current work

catloversg:
MRE:
- Load a clean save file.
- Training at a gym.
- Press the RFA status icon in the overview window.
- The current tab switches to the setting page, but `Player.focus` is true.

I initially planned to fix the bug in `RemoteFileApiConnectionStatus.tsx`, but decided to centralize the check instead, as noted in the comments.

### DIFF
--- a/src/ui/GameRoot.tsx
+++ b/src/ui/GameRoot.tsx
@@ -266,6 +266,13 @@ export function GameRoot(): React.ReactElement {
           }
           break;
       }
+      // If the current page is Page.Work, the player is focusing on their current work. Switching to another page ends
+      // that focus, so we must call Player.stopFocusing() immediately after Router.toPage() to keep Player.focus in
+      // sync. Instead of repeating this logic wherever Router.toPage() is called, we should centralize the check and
+      // the Player.stopFocusing() call here.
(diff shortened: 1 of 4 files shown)


src/ui/GameRoot.tsx
```diff
@@ -266,6 +266,13 @@ export function GameRoot(): React.ReactElement {
           }
           break;
       }
+      // If the current page is Page.Work, the player is focusing on their current work. Switching to another page ends
+      // that focus, so we must call Player.stopFocusing() immediately after Router.toPage() to keep Player.focus in
+      // sync. Instead of repeating this logic wherever Router.toPage() is called, we should centralize the check and
+      // the Player.stopFocusing() call here.
+      if (pageWithContext.page === Page.Work && page !== Page.Work && Player.currentWork && Player.focus) {
+        Player.stopFocusing();
```
d0sboots, line 274: We should also clean up all the places this is currently scattered about.

catloversg:
We can also centralize the `Player.startFocusing()` call in the same way, but I'll do that in a separate PR.

d0sboots:
Yeah, what all those sites really show is that the if focusing/Router.ToPage stuff should get moved centrally into Player.startWork (if it has a dep on router) or a new helper (if needed to avoid the circular dep)

PR #2736, opened 2026-05-06
DNET: Reset IP when servers restart

ficocelliguy:
At player's request, this PR assigns a new, random, IP to a darknet server when it is restarted, as if it was re-connected to the web and assigned a new IP. 

Context:
https://discord.com/channels/415207508303544321/1469814021883560046/1501258707827167242
https://discord.com/channels/415207508303544321/1469814021883560046/1501266896551543025

### DIFF
--- a/src/DarkNet/controllers/NetworkMovement.ts
+++ b/src/DarkNet/controllers/NetworkMovement.ts
@@ -1,10 +1,4 @@
-import {
-  changeIp,
-  connectServers,
-  DeleteServer,
-  disconnectServers,
-  GetServer,
-} from "../../Server/AllServers";
+import { changeIp, connectServers, DeleteServer, disconnectServers, GetServer } from "../../Server/AllServers";
 import {
   DarknetEvents,
   DarknetState,
@@ -66,7 +60,6 @@ export const mutateDarknet = (): void => {
   if (Math.random() > depthSpeedFactor) {
     return;
   }
-
 
   restartRandomServer();
   restartRandomServer();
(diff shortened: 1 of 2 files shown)


src/DarkNet/controllers/NetworkMovement.ts
```diff
@@ -61,6 +61,10 @@ export const mutateDarknet = (): void => {
     return;
   }
 
+  restartRandomServer();
```
d0sboots: Is this testing code that got left in? It's a silent balance change, and a *huge* one.

d0sboots (commented):
I don't think I agree with this change, as it represents a significant nerf/destruction of a strategy for no reason given. 

Specifically, right now you can use ips as a key for your dnet data instead of hostnames, to get around some of the issues that hostnames have. This trades the memorability (and this easier debugability, somewhat) of hostnames for the more straightforward nature of ips, which have no surprises. It also means ips have a use, because they were otherwise totally vestigal.

But if ips can change without the underlying server changing (the passwords are still valid), then hostnames are clearly always the only valid key to use.

(This would also be a breaking change for various people.)

src/DarkNet/controllers/NetworkMovement.ts
ficocelliguy: yes it is; removed.

ficocelliguy:
I'm fine with this not going though if you are opposed to it. It was an off-the-cuff suggestion from some players.

d0sboots:
Yeah, if this is just being done for flavor I don't think that outweighs the actual balance and breaking-change aspects here.

PR #2746, opened 2026-05-08
DNET: simplify getServer return type; add all missing dnet properties to getServerAuthDetails

ficocelliguy:
Multiple players have run into issues or complained about the odd union type of getServer in the current 3.0 branch. To avoid having a disparate return type of DarknetServer | Server , the return type is simply Server now. However, getServer can still fetch the information for darknet servers if needed, to avoid breaking player scripts. This behavior is deprecated.

Context:
https://discord.com/channels/415207508303544321/415247422638522395/1502386651434062068


markdown/bitburner.ns.getserver.md
```diff
@@ -6,12 +6,10 @@
 
 Returns data of a server.
 
-If the server is a darknet server and has recently gone offline, it will return a dummy server object with `isOnline: false`<!-- -->.
```
d0sboots, line 9: I believe this needs to be kept. Because getServer will remain the only way to get basic server info about dnet servers, it needs to function fully in that case. To my mind, that means the `isOnline` property needs to stay supported instead of deprecated.

However, the return type can become `Server & { isOnline?: boolean }`, which shouldn't cause ts type problems.

src/NetscriptFunctions.ts
```diff
@@ -934,7 +942,7 @@ export const ns: InternalAPI<NSFull> = {
         ...exampleDarknetServerData,
         hostname: isIp ? "" : host,
         ip: isIp ? host : "",
-      } satisfies DarknetServerData & { isOnline: boolean };
+      } satisfies DarknetServerData & { isOnline: boolean } as unknown as Server;
```
d0sboots: This doesn't need to satisfy DnetServerData anymore, but it will need to satisfy Server. I suppose we should keep the fake dnetserverdata fields here for back-compat.

src/NetscriptFunctions.ts
```diff
@@ -958,7 +966,10 @@ export const ns: InternalAPI<NSFull> = {
         logTrafficInterval: server.logTrafficInterval,
         isStationary: server.isStationary,
         purchasedByPlayer: false,
-      } satisfies DarknetServerData & { isOnline: boolean };
+        // FUTURE: remove this deprecated support for darknet servers.
+        // This uses the type casting to have a clean return type,
+        // but also to avoid breaking scripts that expect it to work for darknet servers.
+      } satisfies DarknetServerData & { isOnline: boolean } as unknown as Server;
```
```suggestion
      } satisfies Server & { isOnline: boolean };
```

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -8209,16 +8217,13 @@ export interface NS {
   /**
    * Returns data of a server.
    *
-   * If the server is a darknet server and has recently gone offline, it will return a dummy server object with
```
d0sboots, line 8212: "If the server is a darknet server, it will also contain the "isOnline" field."

d0sboots (commented):
@catloversg Definitely want your thoughts on this one. This was the option I could think of that seemed both reasonably clean and doesn't have api-break issues, but it's still adding on some more tech debt in the form of deprecated properties.

src/NetscriptFunctions.ts
ficocelliguy: `Object literal may only specify known properties, and 'depth' does not exist in type 'Server & { isOnline: boolean; }'.`

What should I do about this? should I exclude the extra darknet server properties, and add fake/placeholder ones for the Server properties?

src/NetscriptFunctions/Darknet.ts
```diff
@@ -704,3 +712,61 @@ export function NetscriptDarknet(): InternalAPI<DarknetAPI> {
     },
   };
 }
+
+export const getDarknetPropertiesForDeprecationSupport = (dnetServer?: DarknetServer | undefined) => ({
+  isOnline: {
```
d0sboots: I mentioned this before, but I think isOnline will need to remain permanently, although it is a debatable point. 

Let's consider the hypothetical end state where we have removed these deprecated properties. Since getServer is still the expected way to get server properties for darknet servers, it still needs to work with offline servers and presumably still needs a way to signal that they're offline. You *could* say "getServerAuthDetails must be used for that," but I can think of cases where it would be useful to call getServer only and save the ram cost. 

Also, it is currently the most reliable (and documented) flag we have for distinguishing between darknet and non-darknet servers in the return. 

This means that the return type of getServer will never be a simple `Server`. This is what I meant when I said all the solutions here are compromises with upsides and downsides.

src/NetscriptFunctions.ts
```diff
@@ -74,7 +74,7 @@ import { NetscriptCorporation } from "./NetscriptFunctions/Corporation";
 import { NetscriptFormulas } from "./NetscriptFunctions/Formulas";
 import { NetscriptStockMarket } from "./NetscriptFunctions/StockMarket";
 import { NetscriptGrafting } from "./NetscriptFunctions/Grafting";
-import type { NS, RecentScript, ProcessInfo, NSEnums, Server as NSInterfaceServer, DarknetServerData } from "@nsdefs";
+import type { NS, RecentScript, ProcessInfo, NSEnums, Server as NSInterfaceServer, Server } from "@nsdefs";
```
d0sboots: Remove the second alias here, should just be using NSInterfaceServer. I've actually surprised this compiles.

d0sboots (commented):
I think DarknetServerData itself can/should be removed from nsdefs and either be brought internal, or removed entirely. I need to look at this from my computer to be sure of that, though.

ficocelliguy:
DarknetServerData was being used in the darknet formulas methods. However, only a handful of properties actually matter for those, and both DarknetServerData and ServerAuthDetails have them. I've migrated the public API to expect a ServerAuthDetails , but they work fine with either for backwards compatability.

src/NetscriptFunctions/Darknet.ts
ficocelliguy: I'm fine with that, an additional property won't cause the kinds of issues that players are running into currently with the disparate union type

src/DarkNet/utils/darknetServerUtils.ts
```diff
@@ -19,3 +19,51 @@ export function getDarknetServerOrThrow(host: string): DarknetServer {
   }
   return server;
 }
+
+/**
+ * Darknet server data.
+ * @public
```
d0sboots: Not public anymore. :)

src/DarkNet/Enums.ts
```diff
@@ -1,4 +1,5 @@
-import type { _ValueOf, DarknetServerData } from "@nsdefs";
+import { _ValueOf, ServerAuthDetails } from "@nsdefs";
```
d0sboots: import type still applies I think?

src/DarkNet/Enums.ts
```diff
@@ -88,3 +89,20 @@ export const exampleDarknetServerData: DarknetServerData = {
...
+  data: "",
+  logTrafficInterval: -1,
+  passwordLength: -1,
+  passwordFormat: "numeric",
+  blockedRam: -1,
+  difficulty: -1,
+  depth: -1,
+  requiredCharismaSkill: -1,
+  isStationary: false,
+};
```
d0sboots: as const

src/Netscript/TypeAssertion.ts
```diff
@@ -1,7 +1,6 @@
-import { exampleDarknetServerData } from "../DarkNet/Enums";
-import type { DarknetServerData } from "@nsdefs";
 import type { NetscriptContext } from "./APIWrapper";
 import { errorMessage } from "./ErrorMessages";
+import { ServerAuthDetails } from "@nsdefs";
```
d0sboots: import type

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -4472,11 +4424,16 @@ export type CacheResult = {
  * @public
  */
 interface ServerAuthDetails {
+  /** The hostname of the server */
+  hostname: string;
```
d0sboots: This is a little weird. It's extra weird because ip is not present, and the two usually go together (i.e. the NS api tries very hard to not prefer hostname or ip over the other).

I looked into the implementation, and I think the only reason you're adding this is actually due to a deficiency in the current implementation. Right now, you use the hostname to look up the server password, to calculate shared characters for timing attack. This means that the formulas functions have no way to calculate the additional time added by timing attack, and have to duplicate this calculation on their own. The number of correct characters should actually be an optional parameter to the function, and then players can make use of it *and* you can remove the need to pass hostname to this function.

I'm not sure if this PR is necessarily the place for that; I guess it depends on how much work it is. I do know that based on this, we shouldn't be adding hostname here.

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -6536,26 +6495,26 @@ interface BladeburnerFormulas {
 interface DarknetFormulas {
   /**
    * Gets the time it will take to authenticate a server.
-   * @param darknetServerData - The server to check authentication time on.
+   * @param serverDetails - The server to check authentication time on.
    * @param threads - The number of threads to use for the authentication. Optional, defaults to 1
    * @param player - The player object. Optional, defaults to the current player status
    */
-  getAuthenticateTime(darknetServerData: DarknetServerData, threads?: number, player?: Person): number;
+  getAuthenticateTime(serverDetails: ServerAuthDetails, threads?: number, player?: Person): number;
```
d0sboots: If you do add correctCharacters here, it probably makes sense to add an optional `options` param to the end instead. This *might* be the last thing we need to add, but it easily might not.

src/ScriptEditor/NetscriptDefinitions.d.ts
ficocelliguy: I've removed hostname from here and reworked the auth time methods to take a correct char count, instead of a password and then looking up the diff.

src/NetscriptFunctions/Formulas.ts
```diff
@@ -482,11 +482,12 @@ export function NetscriptFormulas(): InternalAPI<IFormulas> {
     dnet: {
       getAuthenticateTime:
         (ctx) =>
-        (_darknetServerData, _threads, _player): number => {
+        (_darknetServerData, _threads, _player, correctCharactersInPassword): number => {
```
d0sboots: Although not strictly required, the `_` at the front of the name signals that these are private/non-typechecked values.

src/utils/APIBreaks/3.0.1.ts
```diff
@@ -0,0 +1,19 @@
+import type { VersionBreakingChange } from "./APIBreak";
+
+export const breakingChanges301: VersionBreakingChange = {
+  apiBreakingChanges: [
+    {
+      brokenAPIs: [
+        {
+          name: "ns.dnet.getServerAuthDetails",
+          migration: {
+            searchValue: "ns.dnet.getServerAuthDetails",
```
d0sboots: Since this is a unique word (I believe, double-check me), we should search/replace on just `getServerAuthDetails` to catch code where people have captured the namespace to a variable or are using ram dodging or other things.

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -6528,26 +6493,32 @@ interface BladeburnerFormulas {
...
    * @param threads - The number of threads to use for the authentication. Optional, defaults to 1
    * @param player - The player object. Optional, defaults to the current player status
+   * @param correctCharactersInPassword - only used for 2G_cellular model servers. The number of correct characters in the attempted password. Optional, defaults to 0
    */
-  getAuthenticateTime(darknetServerData: DarknetServerData, threads?: number, player?: Person): number;
+  getAuthenticateTime(
+    serverDetails: DarknetServerDetails,
+    threads?: number,
+    player?: Person,
+    correctCharactersInPassword: number,
```
```suggestion
    correctCharactersInPassword?: number,
```

src/utils/SaveDataMigrationUtils.ts
```diff
@@ -645,7 +645,7 @@ Error: ${e}`,
     }
     showAPIBreaks("3.0.0", breakingChanges300);
   }
-  if (ver < 50) {
+  if (ver < 51) {
```
d0sboots, line 648: I think this is supposed to be 50, although it can be 51 rn for testing

ficocelliguy: 50 is 3.0.0 as far as I can tell
https://github.com/bitburner-official/bitburner-src/blame/dev/src/Constants.ts

catloversg:
I committed some non-trivial changes.

There were some subtle issues with `getServer`, so I rewrote this function.

When the server is offline:
```ts
      const result = {
        isOnline: false,
        ...exampleDarknetServerData,
        hostname: isIp ? "" : host,
        ip: isIp ? host : "",
        ...serverProps,
      };
      setDeprecatedProperties(result, getDarknetPropertiesForDeprecationSupport());
```
You don't pass anything to `getDarknetPropertiesForDeprecationSupport`, so that function will use its default values, but its default values are not the "dummy data" in `exampleDarknetServerData`. For example, `getDarknetPropertiesForDeprecationSupport` spefifies `dnetServer?.blockedRam ?? -1`, but dummy data's `blockedRam` should be 0.

The issue of inconsistent dummy data is the same with `getServerDetails`, so I created a new `exampleDarknetServerDetails`. Dummy data is not real data, but we should still be consistent.

When the server is an online dnet server:
```ts
    if (server instanceof DarknetServer) {
      const result = {
        isOnline: true,
        // truncated
        ...serverProps,
      };
      setDeprecatedProperties(result, getDarknetPropertiesForDeprecationSupport(server));
      return result satisfies Server & { isOnline: boolean };
    }
```
This assumes that port-related props and `organizationName` of a dnet server are *always* whatever values of `serverProps`, which is not true. For example, what if we specify custom `organizationName` for some special dnet servers?

Another note: If you use a "default" object to provide default values, please put it on top when building a new object. Instead of:
```js
const foo = {
  baz: 0,
  ...bar,
}
```
Please use:
```js
const foo = {
  ...bar,
  baz: 0,
}
```
Otherwise, the default values in `bar` may mistakenly override the `baz` value.

`getDarknetPropertiesForDeprecationSupport` also marks `isOnline` as deprecated, but IIUC, it's a normal field.

`assertDarknetServerDetails` (renamed from `assertDarknetServerData`) asserts the unknown data and guarantees it's `DarknetServerDetails`, but it actually only checks 4 specific properties. This is fine now, but it will create bugs later when dnet formulas APIs start using more properties. The current implementation basically lies to the TS compiler.

Note for myself and d0sboots: Refactor `DarknetServerData`. This interface was designed in a different context, with different goals and constraints. Moving it as-is like this still works for now, but it will confuse future contributors. Its design made sense when it was originally introduced, but it no longer fits the current context.

This interface is internal now, so we can refactor without worrying about breaking NS APIs. I'll explain the rationale and handle the refactor in a separate PR.

src/ScriptEditor/NetscriptDefinitions.d.ts
catloversg: This suggestion looks good to me.

src/utils/SaveDataMigrationUtils.ts
d0sboots: Ah yes that is correct.

src/ScriptEditor/NetscriptDefinitions.d.ts
d0sboots: Discussed on discord, fico is pretty certain nothing more will be added.

Issue #2735, opened 2026-05-06
Harden saving to avoid savegame corruption


d0sboots:
In issues #2729 and #2730 permanent savegame corruption is observed, and for a while at least it was suspected that it was due to bad interaction with `__proto__` due to player's own scripts. This perhaps wasn't the case, but what *is* the case is that player actions can *certainly* inadvertently mess up the save process, permanently bricking their save without it being obvious until load time.

[corrupting_save_BN1x1.json.gz](https://github.com/user-attachments/files/27429653/corrupting_save_BN1x1.json.gz)

The attached save has a very simple script, `destroySave.js`.
```js
/** @param {NS} ns */
export async function main(ns) {
  JSON.stringify = globalThis.console.log;
}
```

Running it and then saving the game will destroy the save, without any indication (errors or popups) until the damage has already been done. I propose that we add some more defensive checks into the saving part of the pipeline, similar to what we have on the loading side, so that these type of errors bomb out to the recovery screen (probably) or show a toast (maybe) before permanent corruption occurs.

catloversg:
We sometimes come up with the same plan!

#2729 is caused by mishandling of emoji (the name generator incorrectly reverses the string). I'm not sure what causes #2730, but I'm still hardening `db.ts` and other code.

I'm also very tempted to temporarily remove `__proto__` and `constructor`. I have not found a way to corrupt the save data with them, but I can force a crash and many UI bugs with `Object.assign(db[host], foo)` (a bug report on Discord actually contains this code). Although I like these new challenges, I think we should add them later in a safer way. What do you think about that?

d0sboots:
Aha, that last code snippet was the missing piece I needed to understand why `__proto__` is crashing things. I think we can leave constructor in - it's not nearly as dangerous in this respect - but yeah, there's no good way to prevent too much accidental chaos that's just too hard to debug for newbies with `__proto__` rn.

PR #2755, opened 2026-05-11
MISC: Harden saving to avoid save data corruption

catloversg:
Context: #2735.

If save data is invalid, we will show this error popup:

Capture

### DIFF
--- a/src/utils/Protections.ts
+++ b/src/utils/Protections.ts
@@ -20,6 +20,5 @@ if (window.indexedDB) {
   Object.defineProperty(window, "indexedDB", {
     value: window.indexedDB,
     writable: false,
-    configurable: false,
   });
 }
--- a/src/utils/Protections.ts
+++ b/src/utils/Protections.ts
@@ -1,5 +1,5 @@
 // This file is imported for side effects only.
-/* Prevent inadvertantly redefining certain window properties,
+/* Prevent inadvertently redefining certain window properties,
    which are known to cause unrecoverable game errors when redefined.
    The player is able to redefine these properties as writable if desired. */
 Object.defineProperties(window, {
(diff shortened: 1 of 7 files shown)


src/utils/Protections.ts
```diff
@@ -1,9 +1,25 @@
...
+Object.freeze(IDBTransaction.prototype);
+Object.freeze(IDBObjectStore.prototype);
+Object.freeze(IDBRequest.prototype);
+Object.freeze(IDBOpenDBRequest.prototype);
+if (window.indexedDB) {
+  Object.freeze(window.indexedDB);
+  Object.defineProperty(window, "indexedDB", {
+    value: window.indexedDB,
+    writable: false,
+    configurable: false,
```
d0sboots: Per the above comment, I think we should leave this configurable. This isn't meant to prevent deliberate manipulation, just accidental manipulation.

src/utils/SaveDataUtils.ts
```diff
@@ -35,6 +35,17 @@ async function decompress(binaryData: Uint8Array<ArrayBuffer>): Promise<string>
 }
 
 export async function encodeJsonSaveString(jsonSaveString: string): Promise<SaveData> {
+  if (jsonSaveString == null) {
+    throw new InvalidSaveData(`jsonSaveString is ${jsonSaveString}`);
+  }
+  if (typeof jsonSaveString !== "string") {
+    console.error(jsonSaveString);
+    throw new InvalidSaveData(`jsonSaveString is ${typeof jsonSaveString}`);
```
d0sboots:
```suggestion
    throw new InvalidSaveData(`type of jsonSaveString is ${typeof jsonSaveString}`);
```

src/utils/SaveDataUtils.ts
```diff
@@ -35,6 +35,17 @@ async function decompress(binaryData: Uint8Array<ArrayBuffer>): Promise<string>
...
+  }
+  if (!jsonSaveString.startsWith(`{"ctor":"BitburnerSaveObject"`)) {
+    console.error(jsonSaveString);
+    throw new InvalidSaveData(`Invalid jsonSaveString: ${jsonSaveString.slice(0, 100)}`);
```
d0sboots:
```suggestion
    throw new InvalidSaveData(`Invalid jsonSaveString (doesn't seem to contain a BitburnerSaveObject): ${jsonSaveString.slice(0, 100)}`);
```

d0sboots (commented):
What is the exact effect of throwing an error? Does it pop up a box each time autosave tries to kick in?

Basically, I'm trying to figure out how playable the game remains in this circumstance. It's not necessarily good if the game remains playable, because without expert intervention you won't be able to save your progress so any time spent in this state is wasted.

catloversg:
> Does it pop up a box each time autosave tries to kick in?

Yes.

This is annoying enough for players to realize something went wrong and consider stopping playing to avoid wasting their progress.

PR #2756, opened 2026-05-12
BUGFIX: getAugmentationBasePrice ignores bitnode mult for SoA augs

### LINKED ISSUE #2714: getAugmentationBasePrice() returns wrong value for SoA augmentations
Like the title says, the singularity function `getAugmentationBasePrice()` returns an incorrect value for augmentations from the Shadows of Anarchy faction. Or rather, it can, in bitnodes where the `AugmentationMoneyCost` multiplier isn't 1.

The function automatically multiplies an augmentation's base cost with the bitnode price multiplier before returning it.

https://github.com/bitburner-official/bitburner-src/blob/f4c4b177974b87386f6288dd2fde4504e01a7e28/src/NetscriptFunctions/Singularity.ts#L140-L145

However, SoA costs are calculated differently from other augmentations and only scale with themselves, no other multipliers:

https://github.com/bitburner-official/bitburner-src/blob/f4c4b177974b87386f6288dd2fde4504e01a7e28/src/Augmentation/AugmentationHelpers.ts#L150-L152

You can check this easily in-game by entering BN 11 (has a 200% augmentation money cost multiplier) and running, for example:

```
const aug = "SoA - Might of Ares"
const base = ns.singularity.getAugmentationBasePrice(aug);
const price = ns.singularity.getAugmentationPrice(aug);
ns.tprintf("Base price: %d\nActual price: %d",base,price)
```

SAY-5:
Fixes #2714.

SoA augmentations don't scale with the bitnode `AugmentationMoneyCost` multiplier; their cost is computed as `baseCost * SoACostMult^soaAugCount` in `getAugCost`. But `ns.singularity.getAugmentationBasePrice` was always multiplying by `currentNodeMults.AugmentationMoneyCost`, so it returned the wrong base price in any bitnode where that mult isn't 1 (e.g. BN11 = 2.0).

Exposed `soaAugmentationNames` from `AugmentationHelpers` and short-circuited the SoA case in `getAugmentationBasePrice` to return `aug.baseCost` directly, matching how `getAugCost` computes the SoA money cost.

### DIFF
--- a/markdown/bitburner.singularity.getaugmentationbaseprice.md
+++ b/markdown/bitburner.singularity.getaugmentationbaseprice.md
@@ -58,5 +58,5 @@ Base price of the augmentation, before the player's price multiplier.
 
 RAM cost: 2.5 GB \* 16/4/1
 
-This excludes the player's price multiplier, but does include the relevant BitNode multipliers (e.g. the Shadows of Anarchy augmentation cost mult).
+This excludes the player's price multiplier, but does include the relevant BitNode multiplier (for all augs that aren't part of Shadows of Anarchy, which doesn't use BitNode multipliers).
 
--- a/markdown/bitburner.singularity.getaugmentationbaseprice.md
+++ b/markdown/bitburner.singularity.getaugmentationbaseprice.md
@@ -52,9 +52,11 @@ Name of Augmentation.
 
 number
 
-Base price of the augmentation, before price multiplier.
+Base price of the augmentation, before the player's price multiplier.
 
 ## Remarks
 
 RAM cost: 2.5 GB \* 16/4/1
+
+This excludes the player's price multiplier, but does include the relevant BitNode multipliers (e.g. the Shadows of Anarchy augmentation cost mult).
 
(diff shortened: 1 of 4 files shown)


d0sboots:
I'll suggest that you also update the docs for this function (in ScriptEditor/NetscriptDefinitions.d.ts) to make it clear that although most multipliers aren't included, the bitnode multipliers *are* (when appropriate).

SAY-5:
Updated the getAugmentationBasePrice doc to note that BitNode multipliers (like the SoA cost mult) are included even though the player's price multiplier isn't, and regenerated the markdown.

markdown/bitburner.singularity.getaugmentationbaseprice.md
```diff
@@ -52,9 +52,11 @@ Name of Augmentation.
...
 number
 
-Base price of the augmentation, before price multiplier.
+Base price of the augmentation, before the player's price multiplier.
 
 ## Remarks
 
 RAM cost: 2.5 GB \* 16/4/1
 
+This excludes the player's price multiplier, but does include the relevant BitNode multipliers (e.g. the Shadows of Anarchy augmentation cost mult).
```
d0sboots: The wording you had confused me, because it seemed to say the opposite of how it works.
```suggestion
This excludes the player's price multiplier, but does include the relevant BitNode multiplier (for all augs that aren't part of Shadows of Anarchy, which doesn't use BitNode multipliers).
```

SAY-5:
Applied your wording suggestion to both NetscriptDefinitions.d.ts and the generated markdown. Thanks for the review.

PR #2757, opened 2026-05-12
MISC: Temporarily remove servers with unusual hostnames

catloversg:
Context: https://github.com/bitburner-official/bitburner-src/issues/2735#issuecomment-4387196963.


src/Server/AllServers.ts
```diff
@@ -148,6 +148,10 @@ export function loadAllServers(saveString: string): void {
     if (!(server instanceof Server) && !(server instanceof HacknetServer) && !(server instanceof DarknetServer)) {
       throw new Error(`Server ${serverName} is not an instance of Server or HacknetServer or DarknetServer.`);
     }
+    // Temporarily skip loading servers with these unusual hostnames.
```
d0sboots: Honestly, I don't think we even need this. Darknet servers are very ephemeral; as soon as we stop creating these servers they will basically disappear. Adding loading code like this will make them disappear slightly faster, but also adds the possibility of bugs and permanent code that becomes more-or-less instantly unneeded. 

I'd probably feel differently if this PR were a bugfix, but since it's just removing a feature IMO the balance is different.

test/jest/__snapshots__/Save.test.ts.snap
```diff
@@ -45,9 +45,7 @@ exports[`load/saveAllServers 1`] = `
           ]
         ]
       },
-      "serversOnNetwork": [
-        "__proto__"
-      ],
+      "serversOnNetwork": [],
```
d0sboots: Leaving the loading code alone means we can leave this test as-is, and IMO it's good to keep this in as a stress test if our internal name handling.

Issue #1997, opened 2025-03-05
BUG: "hacknet" property of MoneySource incorrectly requires RAM


iclasen:
Steps to reproduce:

Attempt to use the [hacknet](https://github.com/bitburner-official/bitburner-src/blob/dev/markdown/bitburner.moneysource.hacknet.md) property of [MoneySource](https://github.com/bitburner-official/bitburner-src/blob/dev/markdown/bitburner.moneysource.md)

```
/** @param {NS} ns */
export async function main(ns) {
  ns.getMoneySources().sinceInstall.hacknet
}

```

![Image](https://github.com/user-attachments/assets/de4bf65c-8b79-46d3-93ba-1c164b2d59ef)

Proposed Solution:

Unless I am wildly misunderstanding something. I do not believe any property should have a RAM cost. I assume this is due to a naming collision with the [Hacknet API](https://github.com/bitburner-official/bitburner-src/blob/dev/markdown/bitburner.hacknet.md)

Issue #1894, opened 2025-01-07
Incorrect RAM calculation if a user define class method has the same name as a NS API


langerest:
Version 2.7.0 A minimal sample script to reproduce the issue
test.js
```
class TestClass {
    constructor() {
    }
    run() {
    }
}
function main(ns) {
    let test = new TestClass();
    test.run();
}
```

mem test.js shows
[home /]> mem test.js 
This script requires 2.60GB of RAM to run for 1 thread(s)
  1.60GB | baseCost (misc)
  1.00GB | run (fn)

Even though the test.js did not include ns.run()

catloversg:
Duplicate of #875.

Issue #1674, opened 2024-10-01
`GangMemberInfo.hack` property access adds RAM cost


aife-finklea:
Accessing the `hack` property of a `GangMemberInfo` object in a script adds the RAM cost for `ns.hack()` to the script RAM. 

Problematic script (this is the entire file):
```js
/** @param {NS} ns **/
export async function main(ns) {
  ns.alert(ns.gang.getMemberInformation("CoolGuy").hack);  // Script total RAM cost: 3.70 GiB
}
```

No RAM penalty added for accessing a different property of the same object:
```js
/** @param {NS} ns **/
export async function main(ns) {
  ns.alert(ns.gang.getMemberInformation("CoolGuy").str); // Script total RAM cost: 3.60 GiB
}
```

aife-finklea:
I'm an idiot and didn't check the closed issues. Duplicate of #298

aife-finklea:
I'm an idiot and didn't check the closed issues. Duplicate of #298 and #875

Issue #298, opened 2023-01-05
Player is charged ram cost for using variable/property/function names of ns functions


RiversJohn:
Bitburner 2.2.1 : c46cedd5

Below test script will result in an additional 150mb RAM cost for the script due to line 25 call to readback.weaken
Being a C# programmer not a JS guy i can't offer too much insight as to what's going on but i'd imagine this is due
to dynamic typing and inability to type resolve calls so the mem function simply looks for .weaken maybe?

```
/** @param {NS} ns */
export const weakenSecBoost = 0.05;
export const growSecCost = 0.004;
export const hackSecCost = 0.002;
export const globalPort = 1;

export async function main(ns) {
    var globals =
    {
            "weaken": weakenSecBoost,
            "grow": growSecCost,
            "hack": hackSecCost
    };

    var globalsJSON = JSON.stringify(globals);

    var globalsPort = ns.getPortHandle(globalPort);

    await globalsPort.write(globalsJSON);

    var readback;

    readback = JSON.parse(await globalsPort.read());

    ns.tprint(readback.weaken);
}

```

Snarling:
Any ns function names are essentially reserved keywords if you don't want to pay their ram cost. For the time being there's no way around this, and would require a significant ramcost system rework to make it so it does not work like this.

As a workaround you can write that as readback["weaken"] to avoid the ramcost.

Snarling:
This was partially fixed by the addition of the [ramOverride](https://github.com/bitburner-official/bitburner-src/blob/dev/markdown/bitburner.runoptions.ramoverride.md) property on [runOptions](https://github.com/bitburner-official/bitburner-src/blob/dev/markdown/bitburner.runoptions.md), which can take the place of the threads argument for [ns.run](https://github.com/bitburner-official/bitburner-src/blob/dev/markdown/bitburner.ns.run.md), [ns.exec](https://github.com/bitburner-official/bitburner-src/blob/dev/markdown/bitburner.ns.exec.md), etc

This is the extent to which this is really able to be addressed, so I am closing the issue. Sorry this is not a more complete fix, but there are limitations we have to work within.

PR #2758, opened 2026-05-12
BUGFIX: Fix static RAM parsing false positives from identifier-name collisions

### LINKED ISSUE #875: Property name in GangMemberInfo and GangMemberAscension causes unnecessary RAM cost
### Steps to reproduce
1. Write ns2 codes below:
```
ns.gang.getMemberInformation('').hack
ns.gang.getAscensionResult('').hack
```
2. Check RAM usage

### Result
RAM cost of function hack() is calculated:
image
which is irrelevant in gang managing scenairos.

### Game version
Bitburner v2.5.0 (b87b8b4be)

bobbybaxter:
## Summary

Static RAM calculation over-charged scripts whenever a user-defined identifier shared a name with an NS API function. Several long-standing user reports trace back to the same root cause inside `parseOnlyCalculateDeps` / `parseOnlyRamCalculate` in `src/Script/RamCalculations.ts`:

1. The walker descended into both sides of every `MemberExpression`, so for any `expr.api` the property name `api` was added as a bare identifier reference.
2. The resolver looked up bare names in `RamCosts` with a recursive search that walked into nested namespaces, so a local variable named `attempt` matched `RamCosts.codingcontract.attempt`, etc.
3. There was no scope analysis at all, so a parameter named `readback` anywhere in the file made `readback.weaken` charge for `weaken()` everywhere else.

This PR rewrites the member-chain handling around explicit, scope-aware path collection.

## What changes (high level)

- **Static path collection.** A new `collectStaticMemberExpressionPath` walks a non-computed `MemberExpression` chain and returns the dotted access path when the root is an `Identifier`, `ThisExpression`, `Super`, or `PrivateIdentifier`. Dynamic roots (e.g. a `CallExpression` like `ns.gang.getMemberInformation('').hack`) return `null` (no path is collected), so no leaf ref enters the dep map. This alone fixes the dynamic-root family of bugs (`#875`, `#1674`, `#1997`).
- **Strict `RamCosts` lookup.** `findFunc`'s recursive search is replaced with `lookupRamCost`. Dotted refs traverse `RamCosts` exactly; bare refs only match top-level entries. A local named `attempt` can no longer be mis-resolved to `codingcontract.attempt`.
- **Property-walk restricted to computed access.** `MemberExpression` now only descends into `node.property` when it's computed (`obj[expr]`). Static property names are captured exclusively through path collection, so leaf names like `hack` in `someCall().hack` are never treated as identifier references.
- **Per-function lexical scope analysis.** A new `ScopeFrame { params, locals }` is pushed/popped on entry to every `FunctionDeclaration`, `FunctionExpression`, and `ArrowFunctionExpression`. A `rootIsVisibleParam(name, scopes)` lookup walks the chain innermost-to-outermost, with `locals` shadowing outer `params`. The bare-leaf RamCosts fallback (used for renamed-`ns` patterns like `main(X){ X.hack() }`) only fires when the chain root is `this`, `super`, or a visible parameter. `var readback` in `main` correctly shadows a same-named parameter in an unrelated helper, so `readback.weaken` does not charge.

## Related issues

- closes #875 — `ns.gang.getMemberInformation('').hack` and similar dynamic-root accesses
- addresses #1674 — same shape, on `GangMemberInfo`
- addresses #1997 — `ns.getMoneySources().sinceInstall.hacknet`
- addresses #298 — `readback.weaken` (object property name collides with `weaken()`)
- addresses #1894 — user class method `test.run()` collides with `ns.run()`

## Files

- `src/Script/RamCalculations.ts`
  - New helpers: `unwrapExpressionForMemberPath`, `collectStaticMemberExpressionPath`, `ramCostRefsFromStaticPath`, `lookupRamCost`, `ScopeFrame`, `collectPatternNames`, `visitForLocals`, `buildScopeFrame`, `rootIsVisibleParam`, and a `withFunctionScope` helper inside `parseOnlyCalculateDeps`.
  - `State` now carries `scopes: ScopeFrame[]`.
  - The top-level walk seeds `scopes: [moduleFrame]`; the top-level `FunctionDeclaration` handler additionally pushes the function's own frame before walking its body.
  - `commonVisitors` adds `FunctionDeclaration`, `FunctionExpression`, and `ArrowFunctionExpression` visitors that push/pop a `ScopeFrame` around their body walk. The `Object.assign` order is `(commonVisitors, topLevelOverrides)` so the outer-most walk still routes through the top-level `FunctionDeclaration` handler (which keeps `checkRamOverride` and per-function dep-map keys working).
  - `MemberExpression` collects a static path, computes `allowLeafFallback = root === "this" || root === "super" || rootIsVisibleParam(root, st.scopes)`, and only descends into `node.property` for computed access.
  - The old recursive `findFunc` is removed.

- `test/jest/Netscript/StaticRamParsingCalculation.test.ts`
  - Un-skipped the existing `Function 'getTask' that can be confused with Sleeve.getTask` test (was `it.skip` with a TODO that pointed at this fix).
  - New: parameter named `attempt` does not pick up `codingcontract.attempt` RAM.
  - New: `ns.gang.getMemberInformation('').hack` (issue #875).
  - New: `readback.weaken` (issue #298, simple form).
  - New: `test.run()` user class method (issue #1894, simple form).
  - New: `var readback` shadows unrelated `decode(readback)` param (issue #298 contrived).
  - New: `let test` shadows unrelated `helper(test)` param (issue #1894 contrived).
  - New: renamed-`ns` parameter wins over a module-level `var X` of the same name.

### Automated

- `npx jest test/jest/Netscript/StaticRamParsingCalculation.test.ts test/jest/Netscript/RamCalculation.test.ts` passes (146 passed, 1 unrelated `it.skip`).
- `npm run lint` is clean.
- `npm run format` is clean.

### Manual

Tested in-game with the following minimal scripts; the right-hand column is what `mem <script>.js` reports after this PR.

| Pattern | Before | After |
|---|---|---|
| `await ns.hack("n00dles")` | 1.70 GB | 1.70 GB |
| `ns.hacknet.purchaseNode(0)` | 2.10 GB | 2.10 GB |
| `ns.gang.getMemberInformation('').hack` (#1674) | 3.70 GB | **3.60 GB** |
| `ns.getMoneySources().sinceInstall.hacknet` (#1997) | 6.60 GB | **2.60 GB** |
| `var readback; readback.weaken` (#298) | 1.75 GB | **1.60 GB** |
| `class T { run(){} } new T().run()` (#1894) | 2.60 GB | **1.60 GB** |
| `main(X){ X.hack(...) }` (renamed `ns`) | 1.70 GB | 1.70 GB |
| `this.ns.hack(...)` in a method | 1.70 GB | 1.70 GB |
| `this.#ns.grow(...)` private field | 1.75 GB | 1.75 GB |

## Notes for reviewers

- The PR introduces no new public API and no changes to `calculateRamUsage`'s signature. The entire fix is scoped to the per-module dependency walk.
- Scope analysis is intentionally lightweight: it tracks parameters and the names declared in the same function-scope as those parameters. `var`/`let`/`const` are treated uniformly as function-scoped — a mild over-approximation of block-scoped `let`/`const` that biases toward the pessimistic side of the RAM pass.
- One pre-existing `it.skip`-ed test ("Importing a function from a library that contains a class") remains skipped. Its TODO is about `import * as foo` wildcard expansion in dependency resolution, which is unrelated to the member-expression / scope handling fixed here.

### DIFF
--- a/src/Script/RamCalculations.ts
+++ b/src/Script/RamCalculations.ts
@@ -72,15 +72,16 @@ function getNumericCost(cost: number | (() => number)): number {
   return typeof cost === "function" ? cost() : cost;
 }
 
-function unwrapExpressionForMemberPath(node: acorn.Expression | acorn.Super): acorn.Expression | acorn.Super {
+/** Skip over `(...)` and `?.` wrappers to reach the actual expression. */
+function unwrapExpression(node: acorn.Expression | acorn.Super): acorn.Expression | acorn.Super {
   let n: acorn.Node = node as acorn.Node;
   for (;;) {
     if (n.type === "ParenthesizedExpression") {
-      n = (n as acorn.ParenthesizedExpression).expression as acorn.Node;
+      n = (n as acorn.ParenthesizedExpression).expression;
       continue;
     }
     if (n.type === "ChainExpression") {
-      n = (n as acorn.ChainExpression).expression as acorn.Node;
+      n = (n as acorn.ChainExpression).expression;
       continue;
     }
     return n as acorn.Expression | acorn.Super;
@@ -88,27 +89,20 @@ function unwrapExpressionForMemberPath(node: acorn.Expression | acorn.Super): ac
 }
 
 /**
- * Collects a.b.c from a non-computed member chain whose root is a simple Identifier,
- * ThisExpression, or Super. Returns null for dynamic roots (e.g. call results) or
- * computed access we can't statically resolve.
+ * Collects `a.b.c` from a non-computed member chain whose root is an Identifier, ThisExpression,
+ * or Super. Returns null for dynamic roots (e.g. call results) or computed access we can't
+ * statically resolve.
  */
-function collectStaticMemberExpressionPath(node: acorn.MemberExpression): string[] | null {
+function collectStaticMemberPath(node: acorn.MemberExpression): string[] | null {
   const parts: string[] = [];
   let current: acorn.MemberExpression = node;
   for (;;) {
-    if (current.computed) {
-      return null;
-    }
+    if (current.computed) return null;
     const prop = current.property;
-    // PrivateIdentifier (e.g. `obj.#name`) is treated as a static name. Identifier is the common case.
-    if (prop.type === "Identifier") {
-      parts.unshift(prop.name);
-    } else if (prop.type === "PrivateIdentifier") {
-      parts.unshift("#" + prop.name);
-    } else {
-      return null;
-    }
-    const obj = unwrapExpressionForMemberPath(current.object as acorn.Expression);
+    if (prop.type === "Identifier") parts.unshift(prop.name);
+    else if (prop.type === "PrivateIdentifier") parts.unshift("#" + prop.name);
+    else return null;
+    const obj = unwrapExpression(current.object as acorn.Expression);
     if (obj.type === "Identifier") {
       parts.unshift(obj.name);
       return parts;
@@ -130,45 +124,9 @@ function collectStaticMemberExpressionPath(node: acorn.MemberExpression): string
 }
 
 /**
- * Maps a static access path to candidate RamCosts ref strings. RamCosts has no top-level
- * `ns` wrapper, so we strip a leading `ns` (and any receiver marker) before joining.
- *
- * The leaf is returned as a separate candidate only when `allowLeafFallback` is true. The
- * caller decides eligibility based on scope: typically true when the chain root is
- * `this`/`super` or a parameter visible in the current lexical scope (and not shadowed by
- * a local declaration). See `rootIsVisibleParam` and the MemberExpression visitor.
- *
- * Without the leaf fallback, identifiers like `readback.weaken` (issue #298) or
- * `test.run()` (issue #1894) are NOT attributed NS RAM. Dynamic roots like
- * `someCall().hack` never produce a path here at all, which fixes issue #875.
- *
- * The strict bare-lookup in `lookupRamCost` (top-level only) further ensures the leaf
- * candidate cannot pick up nested API names (e.g. `obj.codingcontract.attempt` won't
- * resolve to `codingcontract.attempt` via the bare leaf "attempt").
- */
-function ramCostRefsFromStaticPath(path: string[], allowLeafFallback: boolean): string[] {
-  let segments = path;
-  if (segments[0] === "this" || segments[0] === "super") {
-    segments = segments.slice(1);
-  }
-  if (segments[0] === "ns") {
-    segments = segments.slice(1);
-  }
-  if (segments.length === 0) {
-    return [];
-  }
-  const refs: string[] = [segments.join(".")];
-  if (allowLeafFallback && segments.length > 1) {
-    refs.push(segments[segments.length - 1]);
-  }
-  return refs;
-}
-
-/**
- * A lexical scope frame used to decide whether an identifier root in a static member
- * chain (e.g. the `X` in `X.hack`) is actually a function parameter visible at that
- * point — and thus a plausible alias for `ns`, eligible for a bare-leaf RamCosts
- * fallback — or is shadowed by a local declaration in the same scope.
+ * A lexical scope frame used to decide whether an identifier root in a static member chain
+ * (e.g. the `X` in `X.hack`) is a function parameter visible at that point — and thus a plausible
+ * alias for `ns` — or is shadowed by a local declaration in the same scope.
  */
 interface ScopeFrame {
   /** Identifier names bound as parameters when entering this function. */
@@ -192,11 +150,8 @@ function collectPatternNames(p: acorn.Pattern | null | undefined, out: Set<strin
       return;
     case "ObjectPattern":
       for (const prop of p.properties) {
-        if (prop.type === "RestElement") {
-          collectPatternNames(prop.argument, out);
-        } else {
-          collectPatternNames(prop.value, out);
-        }
+        if (prop.type === "RestElement") collectPatternNames(prop.argument, out);
+        else collectPatternNames(prop.value, out);
       }
       return;
     case "ArrayPattern":
@@ -208,15 +163,11 @@ function collectPatternNames(p: acorn.Pattern | null | undefined, out: Set<strin
 }
 
 /**
- * Scans `node` (a function body or program) for declarations that introduce names into the
- * same scope as the enclosing function's params. Does NOT descend into nested function
- * bodies — those open new scopes — but does descend through blocks, conditionals, loops,
- * try/catch, switch, labels, and export declarations.
- *
- * `var`/`let`/`const` are treated uniformly as function-scoped. This is a mild
- * over-approximation for block-scoped `let`/`const`, but the only consequence is that a
- * block-scoped local can still shadow an outer parameter — which matches the pessimistic
- * intent of the static RAM pass.
+ * Scans `node` (a function body or program) for declarations introduced into the same scope as
+ * the enclosing function's params. Does NOT descend into nested function bodies — those open new
+ * scopes — but does descend through blocks, conditionals, loops, try/catch, switch, and labels.
+ * `var`/`let`/`const` are treated uniformly; the only consequence is that block-scoped names can
+ * still shadow outer params, which matches the pessimistic intent of the pass.
  */
 function visitForLocals(node: acorn.Node | null | undefined, out: Set<string>): void {
   if (!node) return;
@@ -315,9 +266,9 @@ function buildScopeFrame(params: acorn.Pattern[], body: acorn.Node | null | unde
 }
 
 /**
- * Looks up `name` against the active scope chain. Walks innermost to outermost. The first
- * frame that declares `name` as a local shadows any outer params. Otherwise, the first
- * frame that lists `name` as a parameter signals an `ns`-alias-eligible identifier.
+ * Walks the scope chain innermost-to-outermost. The first frame that declares `name` as a local
+ * shadows any outer params; otherwise the first frame that lists `name` as a parameter signals an
+ * `ns`-alias-eligible identifier.
  */
 function rootIsVisibleParam(name: string, scopes: ScopeFrame[]): boolean {
   for (let i = scopes.length - 1; i >= 0; i--) {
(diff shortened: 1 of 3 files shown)


d0sboots:
The intent of this is great, and I can tell you've put in work to try and fix most of the edge cases here. However, there are some you may not know about, and some are outright unavoidable.

It might take a bit for me to review this all, so I'll leave you with these cases to consider for now: 
* Some scripts explicitly use the construct `ns.hack` (no function call) in order to incur a static RAM cost, because they use RAM-dodging techniques that avoid the static cost check in other ways already. This would break that pattern. 
* You claim to be adding lexical analysis, but you're light on the details of how that works. If a variable is assigned inside an `if`, what is the type of the variable for the purposes of your new RAM-checking? Obviously, this cannot be solved perfectly, so it's important that it have understandable rules.

hexagonrecursion:
@bobbybaxter  It would be great if you pushed your build of bitburner to github pages to allow people to test whether this breaks their scripts before the pull request is merged. There are many clever patterns that may break unless we are very careful.

catloversg:
First, I appreciate your effort in fixing this long-standing issue, but before discussing the specific changes in this PR, we should first discuss the goal of the static RAM calculator. This is very important.

In Bitburner, calling an NS API requires paying that API's RAM cost. The combined RAM usage of all running scripts on a server cannot exceed the server’s maximum RAM. If a script calls an API when insufficient free RAM is available, the script is terminated. In other words, any running script can be killed at any time due to insufficient free RAM, which is especially unfriendly to new players.

To mitigate this, we allocate a fixed amount of "static RAM" when a script starts. This acts as the script's prepaid RAM allocation. As long as the script does not exceed this amount at runtime, it will not be killed. For example, suppose `foo.js` has a static RAM allocation of 4 GB and the server currently has 16 GB of free RAM. When `foo.js` starts, the server's free RAM decreases to 12 GB. As long as the runtime RAM usage of `foo.js` remains at or below 4 GB, the script will not be terminated.

This leads to the next question: how do we calculate static RAM? The simplest solution is not to calculate it at all and instead require players to specify it explicitly. However, that would be unacceptable from a gameplay perspective. This is where the static RAM calculator comes in. The calculator is a "pessimistic" QoL feature that estimates the script RAM cost for us. The key word here is "pessimistic".

A non-obvious characteristic of this calculator is that false positives are annoying but still acceptable, whereas false negatives are unacceptable. With false positives, you waste some RAM, but your scripts still work. With false negatives, however, your scripts may be killed at any time when the dynamic RAM calculator catches the discrepancy.

Now let's check some examples:

Example 1:
```js
/** @param {NS} ns */
  const _ns = ns;
  _ns.hack();
}
```
This is the most basic false negative case.

Example 2:
```js
/** @param {NS} ns */
  const g = ns.gang;
  g.getMemberInformation("");
}
```
This is a very popular pattern that many players use.

Example 3:
```js
let _ns;
/** @param {NS} ns */
  _ns = ns;
  _ns.hack();
}
```
This is similar to example 1, but `_ns` is a module-scope variable.

Example 4:
```js
/** @param {NS} ns */
  globalThis.ns = ns;
  foo();
}
function foo() {
  globalThis.ns.hack();
}
```
This is another popular pattern which involves `globalThis`.

Example 5:
```js
import {foo} from "/lib";
/** @param {NS} ns */
  foo().hack();
}
```
`foo()` lets you access an ns instance of another script. `foo` is not too hard to statically analyze.

Example 6:
```js
import {extremelyComplicatedObfuscatedFunction} from "/lib";
/** @param {NS} ns */
  extremelyComplicatedObfuscatedFunction().hack();
}
```
This is the same as example 5, but `foo` is now a ten-million-line obfuscated function which is practically impossible to statically analyze.

You can argue that most of these examples are impractical (example 2 is practical, though). However, practicality is not the point here. The point is that the static RAM calculator is a QoL feature, not a security tool or an optimization tool. Its primary goal is to find a "safe" RAM cost that prevents scripts from being killed unexpectedly by the dynamic RAM calculator.

In short, any changes to the static RAM calculator must ensure that they do not introduce false negatives. Having a script with a non-optimal RAM cost that still works is far better than having a near-optimal one that gets killed randomly.

Another thing to keep in mind: if you want to accurately calculate the total RAM cost of a script at runtime without running it, you need to solve the Halting Problem, which is impossible.

Now back to the PR. You are trying to improve the static analysis code to remove the false positives. However, while doing that, you are also creating false negatives. For example, this is one of your tests:
```js
  ns.gang.getMemberInformation('').hack;
  ns.gang.getAscensionResult('').hack;
}
```
It proves #875 is fixed, but how about:
```js
  const g = ns.gang;
  g.getMemberInformation('').hack;
  g.getAscensionResult('').hack;
}
```
This is the same as example 2 above. When running this script, instead of merely wasting a bit more RAM, it just crashes.

In general, I'm very hesitant on making non-trivial changes to the static RAM calculator. The first steps should be adding more tests that cover the examples I posted here and ensuring that your changes fix all of these false negatives.

hexagonrecursion:
I have tried examples provided by catloversg. Here are the results:

### Example 1
correct ram cost: 1.7 GB
ram cost in commit a423a0772cc4c7860aced19dd334f453e4423487: 1.6 GB

### Example 2
correct ram cost: 3.6 GB
ram cost in commit a423a0772cc4c7860aced19dd334f453e4423487: 1.6 GB

### Example 3
correct ram cost: 1.7 GB
ram cost in commit a423a0772cc4c7860aced19dd334f453e4423487: 1.6 GB

### Example 4
correct ram cost: 1.7 GB
ram cost in commit a423a0772cc4c7860aced19dd334f453e4423487: 1.6 GB

### Examples 5 and 6
I do not understand the examples well enough to fill in the missing bits and test them.

## Notes
I agree with catloversg that 2 is a practical example. I would argue that 3 and 4 are also useful patterns that some players probably rely on. The pattern in example 3 works fine in any script that you do not need to be able to run multiple instances of in parallel. The pattern in example 4 is more limited, but is still fine as long as only one script does this at a time (or as long as different scripts use different global variables).

## My patterns

I would like to add several patterns that I use in my scripts

### Example 7

```js
/** @param {NS} ns */
  const {
    createCorporation,
    expandIndustry,
    buyTea,
    buyMaterial,
  } = ns.corporation;
  createCorporation('corp', false);  // requires bitnode 3
  expandIndustry('Agriculture', 'ag');
  buyTea('ag', 'Sector-12');
  buyMaterial('ag', 'Sector-12', 'Water', 1);
}
```
correct ram cost: 81.6 GB
ram cost in commit a423a0772cc4c7860aced19dd334f453e4423487: 1.6 GB

### Example 8
```js
/** @param {NS} ns */
  const foo = new Foo(ns);
  await foo.bar();
}

class Foo {
  constructor(ns) {
    this.ns = ns;
  }
  async bar() {
    const ns = this.ns;
    await ns.hack('n00dles');
  }
}
```

commit a423a0772cc4c7860aced19dd334f453e4423487 correctly determines that the cost of this example is 1.7 GB, but we should add a test that covers this pattern to guard against breakage in the future.

I like this pattern better than `this.ns.hack()` because it makes it easier to move code into and out of class methods when refactoring.

### Example 9
Sometimes I bundle several function parameters together into an object. Sometimes that object contains `ns`:
```js
/** @param {NS} ns */
  await foo({ns, notImportant: 'something'}, 42);
}

async function foo(ctx, alsoNotImportant) {
  await ctx.ns.hack('n00dles');
}
```

Just like example 8, commit a423a0772cc4c7860aced19dd334f453e4423487 correctly determines that the cost of this example is 1.7 GB, but we should probably add a test for it.

bobbybaxter:
I came into this expecting scrutiny since I knew this was not a trivial change. You all have much deeper experience here than I do, so I appreciate your review and insights on this. My hope is for this conversation to lead to a positive improvement, but I understand there are tradeoffs and the current implementation might just be the best solution for the time being.

I completely understand and agree with the purpose for making this feature pessimistic. I would argue that there are situations where the false positives are more than just annoying. Naming a variable `hack` and incurring a 0.10 GB charge for an unrelated `ns.hack()` is indeed annoying; however, I would consider naming a variable `attempt` and incurring a 10 GB charge for an unrelated `ns.codingcontract.attempt()` feels closer to the criteria of "unacceptable".

Since @d0sboots asked for more clarity, I'll start there. The checking logic was originally only using lexical analysis, which is literally looking at what names exist in the program text and where they are introduced (function params, local variables, function/class declarations, etc.). After all of the additional use cases, I've expanded it to also use static patterns of expressions to handle aliases and destructuring, and rules for call/member shapes (like #875 ). The type of the variable or what value it holds is not inferred.

Here are the big patterns I'm currently leaning on:
- For paths, it only trusts full dotted chains (like `ns.gang.getMemberInformation) when it can resolve them cleanly against `RamCosts`. For bare names, we only fall back to charging a lone leaf (e.g. just `hack`) when the chain clearly starts from something we treat as the script's Netscript handle (e.g. the `main(ns)` parameter or an `ns`-like alias), so we do not treat every random `.hack` as `ns.hack`
- Follow `ns` where players usually put it (e.g. renamed params, `const x = ns`, `const g = ns.gang`, etc.) so we still recognize real Netscript usage and unrelated calls don't get charged as if they were the full API object.

With the help of your additional examples, I made some modifications. Here is how these examples are handled:

## Example 1:
```js
/** @param {NS} ns */
  const _ns = ns;
  _ns.hack();
}
```
- The walker builds a ScopeFrame and adds `ns` to its `params` (which will later make `ns` count as "Netscript-shaped" for member paths).
- `_ns` is recorded as a full-`ns` alias after going through `commonVisitors()` -> `VariableDeclaration` -> `rhsIsNsLike() === true`.
- `_ns.hack()` is a `MemberExpression`. The walker collects a static chain `["_ns", "hack"]` via `collectStaticMemberPath`, then strips the receiver when it is an `ns` alias via `normalizeMemberPathForRam`. So, `["_ns", "hack"]` becomes `["hack"]`. That string is looked up against `RamCosts`, which includes a top-level `hack` entry, so you get the same RAM charge as `ns.hack()`

The current and new cost are the same:
```
This script requires 1.70GB of RAM to run for 1 thread(s)
  1.60GB | baseCost (misc)
  0.10GB | hack (fn)
```

## Example 2:
```js
/** @param {NS} ns */
  const g = ns.gang;
  g.getMemberInformation("");
}
```
- `const g = ns.gang` is handled in `VariableDeclaration` -> `extractNsNamespaceAlias` which sees a non-computed member on the visible `ns` param and records `namespaceAliases.set("g", "gang")`.
- For `g.getMemberInformation("")`, the static chain is `["g", "getMemberInformation"]`. `normalizeMemberPathForRam` will use that `nameSpaceAliases` to substitute the "g" so it'll come out looking like `["gang", "getMemberInformation"]`, which will then get looked up against `RamCosts`.

The current and new cost are the same:
```
This script requires 3.60GB of RAM to run for 1 thread(s)
  2.00GB | gang.getMemberInformation (fn)
  1.60GB | baseCost (misc)
```

## Example 3:
```js
let _ns;
/** @param {NS} ns */
  _ns = ns;
  _ns.hack();
}
```
- The walker builds a ScopeFrame and adds `_ns` in `locals` and `ns` in `params`.
- `_ns = ns` hits `AssignmentExpression` with `operator === "="`. `rhsIsNsLike(ns)` will evaluate to true, so `_ns` is added to `nsParamAliases` even though the binding was declared outside of the function.
- `_ns.hack()` gets normalized the same as Example 1: `["_ns", "hack"]` -> `["hack"]`, then looked up in `RamCosts` as top-level `hack`

The current and new cost are the same:
```
This script requires 1.70GB of RAM to run for 1 thread(s)
  1.60GB | baseCost (misc)
  0.10GB | hack (fn)
```

## Example 4:
```js
/** @param {NS} ns */
  globalThis.ns = ns;
  foo();
}
function foo() {
  globalThis.ns.hack();
}
```
- The module is prescanned by `moduleGlobalThisNsSlotPrefixMap` and looks for plain assignments `globalThis.<slot> = ...` where the RHS is either the identifier `ns` or a non-computed `ns.<segment>`
- For `globalThis.ns = ns`, the map records `"ns" -> []` (empty prefix = the whole Netscript handle was stored on that slot). The scan is module-wide and order-insensitive, which is the same pessimism as before. Any later read of `globalThis.ns.*` may be hitting real APIs because `globalThis` is shared at runtime.
- In `foo`, `globalThis.ns.hack` has the static chain `["globalThis", "ns", "hack"]`. `normalizeMemberPathForRam` sees `globalThis` + slot `ns`, finds the stored prefix `[]`, and rewrites to `["hack"]`, which resolves like `ns.hack()`.
- NOTE: the prescan's RHS is only `ns` or `ns.<id>` (literal `ns` identifier on the RHS), so patterns like `globalThis.x = _ns` are not modeled. I believe I would need to think this one through more to extend this functionality.

The current and new cost are the same:
```
This script requires 1.70GB of RAM to run for 1 thread(s)
  1.60GB | baseCost (misc)
  0.10GB | hack (fn)
```

## Example 7:
```js
/** @param {NS} ns */
  const {
    createCorporation,
    expandIndustry,
    buyTea,
    buyMaterial,
  } = ns.corporation;
  createCorporation('corp', false);  // requires bitnode 3
  expandIndustry('Agriculture', 'ag');
  buyTea('ag', 'Sector-12');
  buyMaterial('ag', 'Sector-12', 'Water', 1);
}
```
- `const {...} = ns.corporation` goes through `recordDestructuredNamespaceAliases`. the init is `ns.corporation` so the namespace segment is `corporation` and each simple binding in the pattern gets `apiLeafAliases.set(localName, "corporation.<apiName>")`.
- Each bare identifier callee (i.e. `createCorporation`, etc) is seen by the `Identifier` visitor, which checks `apiLeafAliases` first and records the dotted ref instead of just the bare name.
- Static allocation includes all four `corporation.*` APIs that are invoked.

The current and new cost are the same:
```
This script requires 81.60GB of RAM to run for 1 thread(s)
 20.00GB | corporation.createCorporation (fn)
 20.00GB | corporation.expandIndustry (fn)
 20.00GB | corporation.buyTea (fn)
 20.00GB | corporation.buyMaterial (fn)
  1.60GB | baseCost (misc)
```

## Example 8:
```js
/** @param {NS} ns */
  const foo = new Foo(ns);
  await foo.bar();
}

class Foo {
  constructor(ns) {
    this.ns = ns;
  }
  async bar() {
    const ns = this.ns;
    await ns.hack('n00dles');
  }
}
```
- `constructor(ns)` is only the class's parameter name, so `rhsIsNsLike` still recognizes `this.ns` / `super.ns` as Netscript-shaped for the initializer `const ns = this.ns` inside `bar()`, so the inner `ns` binding is added to `nsParamAliases` (the same as the previous `const _ns = ns` example)
- the callee `ns.hack` has a static chain of `["ns", "hack"]`, which normalizes to `["hack"]` using that alias set and resolves `hack` like a direct `main(ns)` call

The current and new cost are the same:
```
This script requires 1.70GB of RAM to run for 1 thread(s)
  1.60GB | baseCost (misc)
  0.10GB | hack (fn)
```

## Example 9:
```js
/** @param {NS} ns */
  await foo({ns, notImportant: 'something'}, 42);
}

async function foo(ctx, alsoNotImportant) {
  await ctx.ns.hack('n00dles');
}
```
- On `ctx.ns.hack`, the walker builds the static member path `["ctx", "ns", "hack"]` via `collecStaticMemberPath` on the `MemberExpression callee).
- `normalizeMemberPathForRam` leaves it unchanged: `ctx` is not an `ns` alias, not in `namespaceAliases`, and not a `globalThis` slot from the module prescan.
- `allowLeafFallback` is true because `ctx` is a parameter of `foo` (`rootIsVisibleParam`), so `ramCostRefsFromStaticPath` adds both the full dotted string and the last segment `hack` as lookup candidates.
- `lookupRamCost` resolves the bare `hack` to the top-level `RamCosts.hack` entry

The current and new cost are the same:
```
This script requires 1.70GB of RAM to run for 1 thread(s)
  1.60GB | baseCost (misc)
  0.10GB | hack (fn)
```

## Example 10:
```js
  const g = ns.gang;
  g.getMemberInformation('').hack;
  g.getAscensionResult('').hack;
}
```
- `g` is still a `namespaceAliases` -> `gang` head, so `g.getMemberInformation` and `g.getAscensionResult` normalize to `gang.getMemberInformation` and `gang.getAscensionResult`, and they are charged as the known ns APIs.
- The leaf `.hack` is a property on the result of each call. The `MemberExpression` visitor's call-result branch avoids charging a top-level `hack` RAM line from the property name (the same idea as `ns.gang.getMemberInformation('').hack)

The current cost is:
```
This script requires 5.70GB of RAM to run for 1 thread(s)
  2.00GB | gang.getMemberInformation (fn)
  2.00GB | gang.getAscensionResult (fn)
  1.60GB | baseCost (misc)
  0.10GB | hack (fn)
```

The proposed cost is:
```
This script requires 5.60GB of RAM to run for 1 thread(s)
  2.00GB | gang.getMemberInformation (fn)
  2.00GB | gang.getAscensionResult (fn)
  1.60GB | baseCost (misc)
```

## Notes
- I added tests to capture the same values here
- I also added a new section in `RamCalculations.test.ts` to test against many of these scenarios to ensure the dynamic RAM calculation doesn't exceed the static RAM.
- @hexagonrecursion, good idea about pushing my build. I'll try to get to that when I find the time.
- **EDIT**: I'm currently working on solutions to Examples 5 and 6 because they are not handled correctly in the current state.

hexagonrecursion:
> ## Example 9:
> 
> ```js
> export async function main(ns) {
>   const g = ns.gang;
>   g.getMemberInformation('').hack;
>   g.getAscensionResult('').hack;
> }
> ```

* This is not the example number 9. catloversg did not assign a number to this example. I propose we call this example number 10
* You have skipped the example number 9 that I wrote

bobbybaxter:
> > ## Example 9:
> > ```js
> > export async function main(ns) {
> >   const g = ns.gang;
> >   g.getMemberInformation('').hack;
> >   g.getAscensionResult('').hack;
> > }
> > ```
> 
> * This is not the example number 9. catloversg did not assign a number to this example. I propose we call this example number 10
> * You have skipped the example number 9 that I wrote

Thanks for catching that. I've edited the comment to fix this issue.

catloversg:
I appreciate the quick response. This is a complicated and risky PR, so we will need some time to review it. Don't worry if you don't receive much feedback right away.

d0sboots:
There is a correlary to example 10 that you didn't list.  (Let's call it example 11.) This is `ns.hack` as a property access, which *should* charge cost, as this is currently one of the most reliable ways to add static RAM cost for a specific function without any run-time impact. 

Separating that from example 10 will be tricky, but I suspect you've added enough machinery that it might be possible.

bobbybaxter:
With this new commit, instead of going into details like the above examples, it would make more sense to review the new [StaticRamParsingCalculation.test.ts](https://github.com/bobbybaxter/bitburner-src/blob/26b41d23c7cd3b6732f0627e275ec1824ea5a28c/test/jest/Netscript/StaticRamParsingCalculation.test.ts) file, or pull down the branch and run that test `npm run test StaticRamParsingCalculation.test.ts` yourself.

I've included almost* all of the examples from above, plus some extras that came up as I was thinking through these. I also refactored the test file so it better explains what examples we're setting up and what outcomes we are asserting.

I agree that this is a high-stakes change, hence the hardening of the tests. And I'm sure there's room for improvement, so let me know if you see any mistakes, if anything needs tightening, or if there are any situations I haven't covered.

\*NOTE: I didn't include Example 6, because Example 5 already covers the base scenario and my change doesn't substantially affect the currently way this is handled. If I'm reading it correctly, it currently loads every imported module in full, and each dependency is parsed with `parseAST` on the whole file text, then `parseOnlyCalculateDeps` walks the entire AST (including every top-level function body). So, a hypothetical ten-million-line obfuscated function would stress the same parse-and-walk pipeline today.

Another note: I feel like these changes should make possible the ability to handle the bracket notation function calls, so there isn't a need to call something like `void ns.hack` just to incur the charge of that bracket call; however, I honestly just didn't want to handle that yet.

catloversg:
You don't need to worry about Example 6. It is only there as an extreme case. I included it to illustrate that the analyzer should not attempt overly fancy static analysis because it's impossible to perfectly determine whether a function can return an ns instance. The tests only need to cover Example 5.

d0sboots:
> Another note: I feel like these changes should make possible the ability to handle the bracket notation function calls, so there isn't a need to call something like `void ns.hack` just to incur the charge of that bracket call; however, I honestly just didn't want to handle that yet.

I don't understand what you're saying here; can you clarify?

bobbybaxter:
> > Another note: I feel like these changes should make possible the ability to handle the bracket notation function calls, so there isn't a need to call something like `void ns.hack` just to incur the charge of that bracket call; however, I honestly just didn't want to handle that yet.
> 
> I don't understand what you're saying here; can you clarify?

I believe this would fall under the first scenario you mentioned:
> Some scripts explicitly use the construct ns.hack (no function call) in order to incur a static RAM cost, because they use RAM-dodging techniques that avoid the static cost check in other ways already. This would break that pattern.

Like the following example:
```js
  void ns.hack; // added to charge for the function ns.hack()
  const host = "n00dles";
  await ns["hack"](host); // because calling it this way doesn't charge for ns.hack()
}
```

I have a feeling I could update the parser to handle when `ns` functions are called using bracket notation, but I feel like since it's currently working this way, it's not a regression to leave this functionality as it is right now.

hexagonrecursion:
Today it occurred to me that the simple heuristic that the game currently uses has an advantage: if the player asks for help on discord because their script is affected by a false negative or a false positive, it is reasonably easy to explain how the heuristic works and provide tips on how to work around its flaws.

A more complex heuristic will still have some false negatives and some false positives, but it might be much harder to explain how it works

bobbybaxter:
> Today it occurred to me that the simple heuristic that the game currently uses has an advantage: if the player asks for help on discord because their script is affected by a false negative or a false positive, it is reasonably easy to explain how the heuristic works and provide tips on how to work around its flaws.
> 
> A more complex heuristic will still have some false negatives and some false positives, but it might be much harder to explain how it works

Sure, the parser code grew but I don't think the mental model should be too difficult to explain. Maybe something like:
- The current rule of thumb is "The game charges RAM for any name in your script that matches any Netscript function name".
- The new heuristic would be "The game charges RAM for uses of `ns.` (or aliases of `ns`)".

d0sboots:
> Like the following example:

```js
  void ns.hack; // added to charge for the function ns.hack()
  const host = "n00dles";
  await ns["hack"](host); // because calling it this way doesn't charge for ns.hack()
}
```

Ah, but that is because `ns["hack"](host)` syntax is *explicitly* used to avoid the static ram checker. If that were to charge ram, it would be a major compatibility issue, to the point that I don't think it can (or should) ever be changed. People use that for "ram dodging," where ns functions are called dynamically without incurring a static ram cost, and the static cost must be accounted for another way (usually through `ns.ramOverride`). Using `ns.hack` as a property is another way, which allows ram-dodged library functions to have non-dodged wrappers that account for their cost in the normal fashion.

Edit: This is a glimpse into the complexities of how there isn't just "right" and "wrong" here, but a complex ecosystem of behaviors and people's scripts that have grown up around the current way the static checker functions. Some parts of it are straightforwardly "bugs," but others are almost certainly being relied on by various people.

> Today it occurred to me that the simple heuristic that the game currently uses has an advantage: if the player asks for help on discord because their script is affected by a false negative or a false positive, it is reasonably easy to explain how the heuristic works and provide tips on how to work around its flaws.

Yes, that is a significant strength of the current system. If the new system works "well enough" then the benefits may exceed the disadvantages, but there will always be weird edge cases, and a more complicated system will always have more complicated edge cases.

bobbybaxter:
> > Like the following example:
> 
> ```js
> export async function main(ns) {
>   void ns.hack; // added to charge for the function ns.hack()
>   const host = "n00dles";
>   await ns["hack"](host); // because calling it this way doesn't charge for ns.hack()
> }
> ```
> 
> Ah, but that is because `ns["hack"](host)` syntax is _explicitly_ used to avoid the static ram checker. If that were to charge ram, it would be a major compatibility issue, to the point that I don't think it can (or should) ever be changed. People use that for "ram dodging," where ns functions are called dynamically without incurring a static ram cost, and the static cost must be accounted for another way (usually through `ns.ramOverride`). Using `ns.hack` as a property is another way, which allows ram-dodged library functions to have non-dodged wrappers that account for their cost in the normal fashion.
> 
> Edit: This is a glimpse into the complexities of how there isn't just "right" and "wrong" here, but a complex ecosystem of behaviors and people's scripts that have grown up around the current way the static checker functions. Some parts of it are straightforwardly "bugs," but others are almost certainly being relied on by various people.

Yeah, I get where you're coming from here. I'd be interested in seeing a real use case for this pattern in particular because, from this small example, the benefits aren't apparent to me.

catloversg:
Parallel to this PR, I'll build infrastructure to test real scripts from our players. We have tons of save files from players through GitHub issues and Discord. This will help us determine which edge cases we need to handle and which can reasonably be ignored.

NagaOuroboros:
> Yeah, I get where you're coming from here. I'd be interested in seeing a real use case for this pattern in particular because, from this small example, the benefits aren't apparent to me.

Here's an actual example from my "universal worker":
```ts
// . . .
ns.ramOverride(BASE_COST + ns.getFunctionRamCost(method));
await ns[method](target, opts);
// . . .
```
In my code, I ensure `method` can only be one of the strings `"hack"`, `"grow"`, or `"weaken"`.  All three NS methods share the same function signature (`(host?: string | undefined, opts?: BasicHGWOptions | undefined): Promise<number>`), but have different RAM costs.  Rather than 'dodging' the cost, I dynamically adjust the RAM cost of the script before invoking the method, since the Static RAM Checker can't predict what value `method` will have.  It also ensures I only pay the RAM needed for method that gets called, rather than the sum of all three, i.e. using `if..else` or `switch..case`.

d0sboots:
@NagaOuroboros That isn't quite the case they were talking about, because it doesn't have any static strings/function calls for the ram-checker to match - it's all dynamic.

A good example of the techniques I'm taking about in practice would be https://github.com/d0sboots/bitburner/blob/main/lib/formulas.ts

Berdes:
I was also considering improving the static RAM cost analysis to reduce the false-positives. I'm super happy to see that someone is already working on it!

I found a couple of cases where the new implementation is missing some costs. I think most of those patterns are reasonable enough that I wouldn't be surprised if some player are using them.

## Example 11
Namespace alias and destructuring don't work well together.
```js
  const gang = ns.gang;
  const { createGang } = gang;
  createGang("");
}
```

## Example 12
Namespace alias don't seem to be handled well if the alias is not the same name as the namespace.
```js
  create(ns.gang);
}

function create(g) {
  g.createGang();
}

// One-liner version that also breaks.
  (g => g.createGang())(ns.gang);
}
```

Using `gang` as parameter works.

## Example 13
Local namespace alias is missed if both `ns` and the namespace are renamed.
```js
export async function main(n) {
  const g = n.gang;
  g.createGang();
}
```

## Example 14
Similar to examples 5 and 6, but getting ns out of an array instead.
```js
  [ns][0].hack();
}
```

This one is a bit of a stretch, but I'm using as yet another example of of the pattern that comes out of the previous 3 examples.

## Example 15
The shadowing logic is too optimistic: the scoped ns is incorrectly shadowing the function argument.
```js
  ns.gang.createGang();
  {
    const ns = 0;
  }
}
```

If you want to stick with the simplified scope analysis you're doing, you could instead assume that an identifier that matches both a parameter and a local variable can be either. If you have nested scopes, you can search for the identifier in parent scopes:
 * if you see a local variable, it could be a match, but you also need to check parent scopes.
 * if you see a function parameter, it could be a match. You also don't need to check parent scopes since you're guaranteed that it has shadowed anything declared with the same name in ancestor scopes.

## Alternative approach

I think we can achieve the stated goal of avoiding to charge the ram cost of `codingcontract.attempt` if the user has a local `const attempt` somewhere in a much simpler way that shouldn't cause any false-positives.

As far as I can tell, ignoring things like `eval`, `operator[]` and similar (which are already excluded from static ram calculations), the only way to access something from `ns` is by using the property access operator `.` or the chaining operator `?.` at some point (EDIT: and also destructuring). The key here is that we should only scan the right hand side of those operators to find "NS names" (things like `hack`, `createGang`, etc).

## More advanced analysis

It might also be possible to exclude some property access operators and chaining operators from the ram cost lookup if static analysis proves those cannot ever be derived from NS. The way this can be done is by flagging anything that *could* be from NS as derived from NS (any function parameter could) and then propagate this flag to local variables if at any point they could get "poisoned" by NS. However, there are *a lot* of ways that a variable can get poisoned:
 * if it's on the lhs of an assignment where the rhs is an expression involving something poisoned: `let obj = ns`, obj is poisoned.
 * if it's used as function argument (the function could modify the object to poison it)
 * if it's used in a for...of loop where the loop variable gets poisoned: `for (let a of obj) { a.foo = ns; }`, obj is poisoned.
 * if the result of an expression involving the variable gets poisoned: `let a = obj.mem; a = ns`, obj is poisoned.
 * an expression involving something poisoned can actually be anything because of prototypes: `Number.__prototype__.ns = ns`, now all numbers are poisoned.

So basically, there is no way to guarantee that an object is never going to be poisoned.

## About the tests

Even in the event where it's decided that we shouldn't change the static ram cost calculation (I'd be sad), I think all the tests should still be added as that would still be a significant progress towards any future effort to improve this part of the codebase.

Berdes:
So, I implemented the simplified alternative approach that I mentioned in my previous message and it looks like it's working pretty well: https://github.com/bitburner-official/bitburner-src/compare/dev...Berdes:bitburner-src:static-ram

All the original tests are passing (including one of the skipped ones) and the additional tests that should cover the examples provided here are also passing. Those cover both examples of ram costs that should be identified and instances of local variables/functions that shouldn't trigger ram costs.

Should I open a separate PR for my changes, or would you prefer for them to be included in the current PR (if that's possible, I haven't used Github all that much)? I don't have a strong preference either way as long as I get proper attribution for the part of the change I authored.

d0sboots:
I'd say open a separate PR and that will make it easier to compare/contrast.

Berdes:
Sounds good. I opened #2801 .

PR #2766, opened 2026-05-15
UI: Show data file names when opening darknet cache files

catloversg:
This is a suggestion on Discord. A player asked us what "data file cache" means. It seems they did not realize that opening a cache may generate new files (.lit or .txt) on the corresponding dnet server. Showing the file names would make this clearer.

Capture

### DIFF
--- a/src/DarkNet/effects/cacheFiles.ts
+++ b/src/DarkNet/effects/cacheFiles.ts
@@ -118,9 +118,7 @@ export const getStockReward = (difficulty: number): string => {
 };
 
 export const getDataFileReward = (difficulty: number, server: DarknetServer): string => {
-  const dataFiles = [];
-  dataFiles.push(...addClue(server));
-  dataFiles.push(...addClue(server));
+  const dataFiles = [...addClue(server), ...addClue(server)];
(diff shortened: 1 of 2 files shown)


ficocelliguy (approved):

src/DarkNet/effects/cacheFiles.ts
```diff
@@ -117,14 +118,13 @@ export const getStockReward = (difficulty: number): string => {
 };
 
 export const getDataFileReward = (difficulty: number, server: DarknetServer): string => {
-  const currentDataFiles = server.textFiles.size;
-  addClue(server);
-  addClue(server);
-  const dataFilesGained = server.textFiles.size - currentDataFiles;
-  if (dataFilesGained === 0) {
+  const dataFiles = [];
```
d0sboots:
```suggestion
  const dataFiles = [...addClue(server), ...addClue(server)];
```

d0sboots (commented):
Not for this PR, but this continues adding information to the "message" field that then needs to be parsed by scripts. We should probably have a more descriptive schema for cache rewards.

PR #2772, opened 2026-05-16
UI: combine favor-related information and formulas into a single tooltip

Mathekatze:
This removes the tooltip for reputation and instead moves the information that was in it to the favor tooltip. This combined tooltip now allows to see all of the information about the relations between favor and reputation at once.

Additionally, while I was at it, I wrapped all the variable names in the MathNotations as `\text` to make them look less janky (compare how "favor" is written in the formulas in the before and after screenshots).

## before
before_reputation
before_favor

## after
after_favor2


d0sboots:
I'm not sure that I agree with this change. The numbers had an icon on them, so it was not hidden that both of them had tooltips. And having two tooltips allows for the information to be split up, and the formulas to be more prominent. Squishing everything into one makes everything smaller, and IMO harder to read.

I do agree with the TeX formatting fixes, though.

PR #2781, opened 2026-05-17
UI: Improve visual effect of "SoA - Flood of Poseidon" augmentation

catloversg:
A player asked why the "SoA - Flood of Poseidon" augmentation did not provide any hints. I did not understand what they meant until they mentioned that they are color-blind. The UI hint provided by this augmentation relies on the ability to distinguish colors, which is problematic for color-blind players.

This PR adds a border around the target symbol so that color-blind players can still identify it. The goal is similar to #1380, but applied to the "Match the symbols" game.

Before:

before

After:

after

### DIFF
--- a/src/Infiltration/ui/Cyberpunk2077Game.tsx
+++ b/src/Infiltration/ui/Cyberpunk2077Game.tsx
@@ -43,7 +43,7 @@ export function Cyberpunk2077Game({ stage }: IProps): React.ReactElement {
         <Typography variant="h5" color={Settings.theme.primary}>
           Targets:{" "}
           {stage.answers.map((a, i) => {
-            if (i == stage.currentAnswerIndex)
+            if (i === stage.currentAnswerIndex)
               return (
                 <span key={`${i}`} style={{ fontSize: "1em", color: Settings.theme.infolight }}>
                   {a}&nbsp;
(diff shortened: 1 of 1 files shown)


src/Infiltration/ui/Cyberpunk2077Game.tsx
```diff
@@ -70,7 +70,13 @@ export function Cyberpunk2077Game({ stage }: IProps): React.ReactElement {
               sx={{
                 fontSize: fontSize,
                 color: item.color,
-                border: item.selected ? `2px solid ${Settings.theme.infolight}` : "unset",
+                // Always use a 2px border to avoid small symbol position shifts caused by size changes (element vs
```
d0sboots: If the border should always be 2px, then IMO that should be pulled outside the ternary chain.

src/Infiltration/ui/Cyberpunk2077Game.tsx
```diff
@@ -70,7 +70,13 @@ export function Cyberpunk2077Game({ stage }: IProps): React.ReactElement {
...
+                // element + 2px border).
+                border: item.selected
+                  ? `2px solid ${Settings.theme.infolight}`
+                  : hasAugment && item.content === stage.answers[stage.currentAnswerIndex]
+                  ? `2px solid ${Settings.theme.warning}`
```
d0sboots: This only differs from the current selection by color, which is a similar problem. What about using a different style liked dashed or dotted?

PR #2784, opened 2026-05-18
MISC: Allow exporting save data even if game fails to load

catloversg:
In the Steam version, players can use `--export-save` flag to load `export.html` and export their save data. This is useful if the game fails to load for an unknown reason. This PR adds the same feature to the web version.
- Stable: https://bitburner-official.github.io/export.html
- Dev: https://bitburner-official.github.io/bitburner-src/export.html

After merging this PR, I'll ask Snarling to upload `export.html` to https://github.com/bitburner-official/bitburner-official.github.io.

### DIFF
--- a/export.html
+++ b/export.html
@@ -27,6 +27,12 @@
     </div>
     <!-- Use esm for top-level await -->
     <script type="module">
+      if (!window.indexedDB || typeof window.indexedDB.databases !== "function") {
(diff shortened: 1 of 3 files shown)


d0sboots:
Will this not be handled by the typical deploy scripts? Or are the top-level files entirely manual?

catloversg:
Let's discuss in #2788.

PR #2786, opened 2026-05-19
FIX: Autocompletion not working on scripts when autocompleting a filepath

Olacken:
### The issue
Fix the autocompletion not calling the autocomplete function  of the script when trying to  run a script using relative path while typing a file path containing /

For example :
You have a server with the files
```
temp.js hack/script.js hack/utils/listServers.js
``` 
The autocomplete will run fine until you type the / and then it will stop working
```shell
/>run temp.js hack/
/>run temp.js hack/scr
/>run temp.js hack/util
``` 
All of those case will fail to fetch the script  because it is trying to find it at /hack/temp.js

If you use an absolute path like
```shell
/>run /temp.js hack/
/>run /hack/script.js ./util
``` 
It will works as it loads the script correctly

Also the next ones will also work since the autocomplete directory and the script directory are the same(until you change autocomplete directory by reaching the next /)
```shell
hack/>run script.js hack/sc
hack/>run utils/listServers.js hack/uti
/>run temp.js ./hac
``` 

### The root cause of the issue
The reason for the bug is that when the getTabCompletionPossibilities function is run, the working directory(baseDir) is passed as an argument.
But when completing a path(aka an input with a /) it is changed to be the absolute path until the last / of the current autocompletion item.
But that same working directory is later used to find the location of the script (to get it's autocomplete function)

### The fix
The working directory was decoupled from the path of the current autocomplete in order to avoid interference between the two

PS: I ran the lint and format but couldn't find how to run the game on my distribution so I tested it by modifying directly the main bundle on my game since it's a very small patch

### DIFF
--- a/src/Terminal/getTabCompletionPossibilities.ts
+++ b/src/Terminal/getTabCompletionPossibilities.ts
@@ -52,14 +52,14 @@ export async function getTabCompletionPossibilities(terminalText: string, baseDi
 
   /** The directory portion of the current input */
   let relativeDir = "";
+  let path = baseDir;
   const slashIndex = currentText.lastIndexOf("/");
 
   if (slashIndex !== -1) {
     relativeDir = currentText.substring(0, slashIndex + 1);
-    const path = resolveDirectory(relativeDir, baseDir);
+    path = resolveDirectory(relativeDir, baseDir);
     // No valid terminal inputs contain a / that does not indicate a path
     if (path === null) return [];
-    baseDir = path;
     pathingRequiredMatch = currentText.replace(/^.*\//, path).toLowerCase();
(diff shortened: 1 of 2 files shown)


catloversg:
You need to specify the exact issue (with concrete examples) that this PR resolves. Autocomplete has many known issues, so we want to know which issue you are trying to fix.

catloversg:
I added some Jest tests.

This PR looks good to me, but some behaviors seem fairly strange, especially in the last two tests. They are not caused by the new changes, but rather by the current implementation (`addGeneric`). I'm not sure whether they are intentional.

@d0sboots Please double-check this PR and my Jest tests.

PR #2790, opened 2026-05-19
Fix tutorial blocking on n00dles.js filename casing

### LINKED ISSUE #2789: Tutorial Commands Not Working
Image

Very early on I am trying to enter nano, which the tutorial is requesting I do, and I get blocked. I did the `n00dles` connection and ran `weaken` `hack` commands as requested, perhaps 1 extra time to see what happens, then I'm into a tutorial block.

rpmullig:
## Summary

- Tutorial terminal gate accepts case-insensitive nano/run/tail with n00dles.js and normalizes the filename before the command runs.
- Create-script tutorial text shows a single command with (check case) instead of a misleading nano-only step.
- Step-specific error messages for those tutorial steps.

Fixes #2789

## Test plan

- [ ] Start a new game, reach the create-script tutorial step after home.
- [ ] nano n00dles.js advances and opens the editor.
- [ ] nano N00dles.js also advances and opens n00dles.js.
- [ ] nano alone shows: Enter: nano n00dles.js (check case).
- [ ] Later tutorial run/tail steps accept mixed-case n00dles.js.

### DIFF
--- a/src/Terminal/Terminal.ts
+++ b/src/Terminal/Terminal.ts
@@ -718,6 +718,7 @@ export class Terminal {
       }
       const errorMessageForBadCommand =
         "Bad command. Please follow the tutorial or click 'Exit Tutorial' if you'd like to skip it.";
+      const tutorialArg = (arg: string | number | boolean) => String(arg).toLowerCase();
       switch (ITutorial.currStep) {
         case iTutorialSteps.TerminalHelp:
           if (commandArray.length === 1 && commandArray[0] === "help") {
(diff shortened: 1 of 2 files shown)


catloversg:
Filenames in Bitburner are case-sensitive. With this change, the tutorial would teach players that filenames are case-insensitive (e.g., `nano foo.js` and `nano FOO.js` are the same), which is wrong.

rpmullig:
Sorry! You're correct. I put `N00dles` and it should've been `n00dles`. Enjoying it so far!

rpmullig:
Closing due to typo error in command line.

PR #2792, opened 2026-05-20
DNET: tweak cha augments & xp; fix script counts

ficocelliguy:
- Fix for the server count and tooltip not correctly handling duplicates of the same script
- Added darknet money to NFG
- Reduced charisma mult from early-game augments, and reduced xp from darknet auth. Players have still been commenting on how quickly charisma levels are gained with little rffort.


gmcew:
May not the PR for it, but I reasoned here's a good place to highlight it anyway: BN13 probably deservers a cha nerf in the BN mults same as the other [stats](https://github.com/bitburner-official/bitburner-src/blob/dev/src/BitNode/BitNode.tsx#L1003).

ficocelliguy:
@gmcew [it's currently at 0.1](https://github.com/bitburner-official/bitburner-src/blob/6095f2e1ace36b7e51897406f56ed7cc73cf306b/src/BitNode/BitNode.tsx#L1039) , do you think that isn't a good value for it?

gmcew:
Not the ` DarknetMoneyMultiplier: 0.1,`, I mean for the charisma stat itself: for BN13 I see the following in the link in my previous comment. I think I'm on dev, I'm happy to be told I missed it though!
```ts
return new BitNodeMultipliers({
        HackingLevelMultiplier: 0.25,
        StrengthLevelMultiplier: 0.7,
        DefenseLevelMultiplier: 0.7,
        DexterityLevelMultiplier: 0.7,
        AgilityLevelMultiplier: 0.7,

        CloudServerSoftcap: 1.6
```

src/DarkNet/ui/ServerSummary.tsx
```diff
@@ -41,7 +43,8 @@ export function ServerSummary({
         }`
       : "No data files on server";
   const contractCount = server.contracts.length;
-  const runningScriptNames = Array.from(server.runningScriptMap.keys()).map((script) => script.replace("*[]", ""));
+  const runningScriptsCount = getScriptCount(server.runningScriptMap);
```
d0sboots: This doesn't feel like it belongs in this PR at all, it has nothing to do with balancing dnet.

However, if you're gonna do it here: this function is completely redundant. You have to get all the names anyway, you can just look at the size of that list. (And we are in fact doing that elsewhere.)

src/DarkNet/ui/ServerSummary.tsx
```diff
@@ -152,3 +155,17 @@ export function ServerSummary({
     </div>
   );
 }
+
+const getScriptNameStrings = (runningScriptMap: Map<ScriptKey, Map<number, RunningScript>>) => {
```
d0sboots: Optional: If you passed the server instead, the signature would be a lot less complicated.

src/DarkNet/ui/ServerSummary.tsx
```diff
@@ -152,3 +155,17 @@ export function ServerSummary({
...
+  return Array.from(runningScriptMap.entries()).map(([script, dataMap]) => cleanScriptName(script, dataMap.size));
```
d0sboots: This is dealing with implementation details surrounding the key, and it's totally unnecessary. All you need is
```suggestion
  const filenames = [];
  for (const map of server.runningScriptMap.values()) {
    for (const rs of map.values()) {
      filenames.push(rs.filename);
    }
  }
  return filenames;
```
You can do it with map and reduce too, but it gets harder to read and involves more temporary array creation.

Since you only do this once, you can save the `return` by just avoiding the helper function and creating `runningScriptNames` as the initial array.

src/DarkNet/ui/ServerSummary.tsx
ficocelliguy: If you have ten scripts running, five are foo.js, five are bar.js, the name list has two items (intentionally, with a counter on each as part of the strings). That is the original bug I am solving here - the script count is not the same as the running script unique filename count.

I want the name tooltip to collapse duplicate names for simplicity, and the script count to be the true distinct PID count running on the server. It doesn't tell you the arguments of the running script, but that's fine; it's supposed to be a small tooltip and not replace the running scripts page.

I'm fine pulling this out to its own PR if you would like.

src/DarkNet/ui/ServerSummary.tsx
```diff
@@ -7,6 +7,7 @@ import { formatToMaxDigits } from "./uiUtilities";
 
 import type { DarknetServer } from "../../Server/DarknetServer";
 import { DarknetConstants } from "../Constants";
+import { BaseServer } from "../../Server/BaseServer";
```
catloversg: I did not test, but I think type-only import should work here.

src/DarkNet/ui/ServerSummary.tsx
```diff
@@ -152,3 +154,22 @@ export function ServerSummary({
     </div>
   );
 }
+
+const getScriptNameStrings = (server: BaseServer) => {
+  const filenames = [];
+  for (const map of server.runningScriptMap.values()) {
```
catloversg: Using `runningScriptMap.keys` should work. For example:
```js
[...server.runningScriptMap.keys().map(v => cleanScriptName(v))]
```
Using an explicit loop is also a good idea. I was just too lazy to write the example that way.

catloversg (commented):
Optional: New utility functions can be inlined (they are used in only one place).

src/DarkNet/ui/ServerSummary.tsx
ficocelliguy: I need to know the count for each script name to show like a `x3` or whatever next to the ones with multiple instances running, so I need more than just the keys. (I was originally doing basically what you suggest - pulling an array from the .entries() value of the map then iterating over it - and d0s proposed pulling it out to a for loop)

catloversg:
```ts
const getScriptNameStrings = (server: BaseServer) => {
  const filenameAndCounts = [];
  for (const [scriptName, map] of server.runningScriptMap) {
    const cleanedName = scriptName.slice(0, scriptName.indexOf("*"));
    const countString = map.size > 1 ? ` x${map.size}` : "";
    filenameAndCounts.push(cleanedName + countString);
  }
  return filenameAndCounts;
};
```
Using `scriptName` directly from the entry lets you avoid using a bunch of `?` and `??` in `map.values().next()?.value?.filename ?? ""`. It also avoids unnecessary access to values in the map.

Edit: I took another look at how `runningScriptMap` is implemented. The key is filename+args, so using `scriptName` may not do what you want. I guess using the updated suggestion from d0sboots is the way to go.

src/DarkNet/ui/ServerSummary.tsx
```diff
@@ -152,3 +154,22 @@ export function ServerSummary({
...
+    const scriptName = map.values().next()?.value?.filename ?? "";
+    filenames.push(cleanScriptName(scriptName, map.size));
+  }
+  return filenames;
+};
+
+const cleanScriptName = (scriptName: string, count: number) => {
+  const cleanedName = scriptName.slice(0, scriptName.indexOf("*"));
```
d0sboots: This part is unneeded when you're dealing directly with clean names in the first place

src/DarkNet/ui/ServerSummary.tsx
```diff
@@ -152,3 +154,22 @@ export function ServerSummary({
...
   );
 }
+
+const getScriptNameStrings = (server: BaseServer) => {
+  const filenames = [];
+  for (const map of server.runningScriptMap.values()) {
+    const scriptName = map.values().next()?.value?.filename ?? "";
+    filenames.push(cleanScriptName(scriptName, map.size));
+  }
+  return filenames;
```
d0sboots: I was unclear on what you were trying to do at first. Unfortunately, I think to do what you want you're going to need to make your own map of script name -> count, and build that by traversing the RunningScript structure. Otherwise, there's no guarantee on correctness - for instance, what you have right now will fail if you run the same script with different arguments.

If you take catlover's suggestion to do everything inline, you can also do the total count as part of the same loop very simply. Here's my updated crack at it:
```suggestion
  let runningScriptCount = 0;
  const scripts = new Map<string, number>();
  for (const map of server.runningScriptMap.values()) {
    const rs = map.values().next().value;
    if (!rs) continue;
    runningScriptCount += map.size;
    let count = (scripts.get(rs.filename) ?? 0) + map.size;
    scripts.set(rs.filename, count);
  }
  const runningScriptNames = scripts.entries().map((name, count) =>name + (count === 1 ? "" : "x" + count));
```

ficocelliguy: I've applied this, tweaked it slightly, and moved it to be inline

PR #2804, opened 2026-05-24
DNET: Fix issues in MathML puzzle

ficocelliguy:
- Show correct stat in heartbleed error message about cha requirement
- Set a minimum in the MathML arithmatic expression puzzle to prevent a single number from being the hint and password
- Fix the "close enough" detection for floating point passwords to correctly handle if the password is negative

### DIFF
--- a/src/DarkNet/controllers/ServerGenerator.ts
+++ b/src/DarkNet/controllers/ServerGenerator.ts
@@ -514,7 +514,7 @@ export const parseSimpleArithmeticExpression = (expression: string): number => {
 
 export const generateSimpleArithmeticExpression = (difficulty: number): string => {
   const operators = ["+", "-", "*", "/"];
-  const operatorCount = Math.min(Math.floor(difficulty / 4), 1);
+  const operatorCount = Math.max(Math.floor(difficulty / 4), 1);
   const expression = [];
   for (let i = 0; i < operatorCount; i++) {
     expression.push(Math.ceil(Math.random() * 98));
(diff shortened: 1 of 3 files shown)


src/DarkNet/controllers/ServerGenerator.ts
```diff
@@ -514,7 +514,7 @@ export const parseSimpleArithmeticExpression = (expression: string): number => {
 
 export const generateSimpleArithmeticExpression = (difficulty: number): string => {
   const operators = ["+", "-", "*", "/"];
-  const operatorCount = Math.floor(difficulty / 4);
+  const operatorCount = Math.min(Math.floor(difficulty / 4), 1);
```
d0sboots: Doesn't this need to be max? This is why we prefer using clamp.

But, how is difficulty becoming negative in the first place?

src/DarkNet/effects/authentication.ts
```diff
@@ -155,7 +155,8 @@ export const isCloseToCorrectPassword = (
   logIfNotClose = false,
 ): boolean => {
   const difference = Math.abs(attemptedPassword - Number(correctPassword));
-  const result = difference < 0.01 || difference / Number(correctPassword) < 0.005;
+  const fractionalDiff = Math.abs(difference / Number(correctPassword));
```
d0sboots, line 158: I said this in the other PR - IMO the password should never be negative. So this doesn't feel like the right fix.

ficocelliguy: That would require re-rolling the arithmetic expression generator each time it happens to result in something below zero, which feels clunky to me. When you're subtracting and dividing and such, sometimes you get negative numbers.

src/DarkNet/controllers/ServerGenerator.ts
ficocelliguy: This is just to handle difficulty 0, in which case there is no expression, just a number, without setting it to a minimum of 1 operator

PR #2805, opened 2026-05-24
BUGFIX: Fixed MathML darknet nodes with negative valued passwords would incorrectly allow wrong passwords

TrixTrl:
Fixed bug where MathML darknet nodes with negative valued math expressions would incorrectly allow wrong passwords

### DIFF
--- a/src/DarkNet/effects/authentication.ts
+++ b/src/DarkNet/effects/authentication.ts
@@ -155,7 +155,7 @@ export const isCloseToCorrectPassword = (
   logIfNotClose = false,
 ): boolean => {
   const difference = Math.abs(attemptedPassword - Number(correctPassword));
-  const result = difference < 0.01 || difference / Number(correctPassword) < 0.005;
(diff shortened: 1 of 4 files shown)


src/DarkNet/effects/authentication.ts
```diff
@@ -155,7 +155,7 @@ export const isCloseToCorrectPassword = (
   logIfNotClose = false,
 ): boolean => {
   const difference = Math.abs(attemptedPassword - Number(correctPassword));
-  const result = difference < 0.01 || difference / Number(correctPassword) < 0.005;
+  const result = difference < 0.01 || Math.abs(difference / Number(correctPassword)) < 0.005;
```
d0sboots, line 158: I didn't think this is the right fix. Although this patches an issue with negative passwords, I don't think passwords should be negative in the first place.

ficocelliguy:
This was resolved by PR #2804

PR #2806, opened 2026-05-25
NETSCRIPT: Adds functions to set aliases programmatically.

vivian-wei-008:
Adds `ns.alias()`, `ns.unalias()`, and `ns.clearAliases()` to set aliases programmatically in the terminal. 

image
image

```js
// testAlias.js
/** @param {NS} ns */
  ns.alias("target", "run aliastarget.js")
  ns.alias("globalTarget", "aliastarget.js", true)
}
```

```js
// aliastarget.js
/** @param {NS} ns */
  ns.tprint("Test success!")
}
```

```js
// clearAliases.js
/** @param {NS} ns */
  ns.clearAliases()
}
```


src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -9107,6 +9107,43 @@ export interface NS {
    */
   getSharePower(): number;
 
+  /**
+   * Programmatically sets an alias.
+   * @param alias - The keyword to set.
```
catloversg: We usually use this order:
```
Short description

@remarks
RAM cost

Full description

@example

@param
@returns
```

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -9107,6 +9107,43 @@ export interface NS {
...
+   * ```
+   *
+   * @remarks
+   * RAM cost: 0.05 GB
+   */
+  alias(alias: string, substitution: string, global?: boolean): void;
+
+  /**
+   * Clears an existing alias.
+   * @param alias - The alias to clear.
```
catloversg: Same comment about the order.

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -9107,6 +9107,43 @@ export interface NS {
...
+   * @returns - True if there was a previous alias set.
+   * @remarks
+   * RAM cost: 0.05 GB
+   */
+  unalias(alias: string): boolean;
+
+  /**
+   * Clears every alias.
```
catloversg: Optional: "Clears all aliases."

src/NetscriptFunctions.ts
```diff
@@ -1411,6 +1412,41 @@ export const ns: InternalAPI<NSFull> = {
   getFavorToDonate: () => () => {
     return Math.floor(CONSTANTS.BaseFavorToDonate * currentNodeMults.FavorToDonateToFaction);
   },
+  alias: (ctx) => (_alias, _substitution, _global) => {
+    const alias = helpers.string(ctx, "alias", _alias);
+    const substitution = helpers.string(ctx, "substitution", _substitution);
+    const global = helpers.boolean(ctx, "global", _global ?? false);
+
+    if (global) {
```
catloversg: You are calling `addAlias`/`addGlobalAlias` directly instead of `parseAliasDeclaration`.

`parseAliasDeclaration` handles quote parsing and edge cases such as `alias =foo` (invalid alias usage). Quote parsing is not particularly important when scripting, but removing it may introduce subtle bugs and behavioral discrepancies between CLI commands and NS APIs. 

The bigger issue is the edge cases such as `alias =foo`. `ns.alias` currently allows `ns.alias("", "foo")`.

src/NetscriptFunctions.ts
```diff
@@ -1411,6 +1412,41 @@ export const ns: InternalAPI<NSFull> = {
...
+      addGlobalAlias(alias, substitution);
+      helpers.log(ctx, () => `Added global alias ${alias}: ${substitution}`);
+    } else {
+      addAlias(alias, substitution);
+      helpers.log(ctx, () => `Added alias ${alias}: ${substitution}`);
+    }
+  },
+  unalias: (ctx) => (_alias) => {
+    const alias = helpers.string(ctx, "alias", _alias);
+    if (alias) {
```
catloversg: You are printing an error message saying "no alias with that name found", but you are checking if the input alias is an empty string.

I guess you mean something like this:
```js
  unalias: (ctx) => (_alias) => {
    const alias = helpers.string(ctx, "alias", _alias);
    const removed = removeAlias(alias);
    if (removed) {
      helpers.log(ctx, () => `Successfully removed the "${alias}" alias.`);
    } else {
      helpers.log(ctx, () => `Failed to remove the "${alias}" alias: no alias with that name found.`);
    }

    return removed;
  },
```

catloversg (commented):
Considerations:
- Move these APIs to `ns.ui` or somewhere else.
- Optional: Add some more APIs as d0sboots mentioned on Discord.

vivian-wei-008:
Made a couple changes
* Moved alias functions to the UI namespace, as was suggested in discord. 
* Added getAlias() and getAllAliases() to get information about existing aliases. 
* Altered Alias input validation to use the same regex as the original alias term. This closes gaps in behavior. 

In-game test scripts: 
```js
// clearAliases.js
/** @param {NS} ns */
  ns.tprint(...ns.ui.getAllAliases())
  ns.ui.clearAliases()
  const printTarget = ["test", ...ns.ui.getAllAliases()]
  ns.tprint(printTarget)
}
```

```js
// testAlias.js
/** @param {NS} ns */
  ns.ui.alias("target", "run aliastarget.js")
  ns.tprint(ns.ui.getAlias("target")) // {"substitution":"run aliastarget.js","isGlobal":false}
  ns.tprint(ns.ui.getAlias("unknownAlias")) // undefined
  ns.tprint(ns.ui.unalias("target")) // true
  ns.tprint(ns.ui.unalias("target")) // false
  ns.tprint(ns.ui.getAlias("target")) // undefined
  ns.ui.alias("globalTarget", "aliastarget.js", true)
  ns.tprint(...ns.ui.getAllAliases()) // ["globalTarget",{"substitution":"aliastarget.js","isGlobal":true}]
}
```

```js
// testAssert.js
/** @param {NS} ns */
  ns.ui.alias("test", "aliastarget.js ", true) // Should succeed
  ns.ui.alias(" 4752", "run aliastarget.js") // Should fail
  // ns.ui.alias(" 4752 ❌❌❌❌", "run aliastarget.js") // Should fail
}
```

image

image

d0sboots:
I will check this over more thoroughly in a bit, but one thing I was going to suggest previously and didn't mention: with the ability to see all the aliases, I don't think we need a clear function built-in. The player can loop and unset if they want.

src/NetscriptFunctions/UserInterface.ts
```diff
@@ -5,11 +5,24 @@ import { defaultTheme } from "../Themes/Themes";
...
+
+/** Converts the provided value to a string and ensures it satisfies the alias condition, throwing if it is not  */
+export function parseAsAlias(ctx: NetscriptContext, argName: string, v: unknown): string {
+  if (typeof v === "number") v = v + ""; // cast to string;
+  assertStringWithNSContext(ctx, argName, v);
+  const matches = v.match(aliasRegex);
+  if (matches == null || matches.length !== 1 || matches[0] !== v) {
+    throw helpers.errorMessage(ctx, `'${argName}' must only contain letters, numbers, or any of these symbols: |!%,@-`);
+  }
+  return v;
```
vivian-wei-008, line 28: Forgot to write this earlier. This is essentially a helper ctx, but it's too specific to live in the helpers and seems like it'd import too many unusual things into alias to live in alias. I have no strong opinions about where this should live, I'd put it in helpers, but I'm not sure about the existing code standards.

vivian-wei-008:
Bump

d0sboots:
I haven't forgotten it, but I am being a bit slow

src/Netscript/RamCostGenerator.ts
```diff
@@ -452,6 +452,11 @@ const ui = {
   getGameInfo: 0,
   clearTerminal: 0,
   windowSize: 0,
+  alias: 0.05,
```
d0sboots: Like the rest of the things in ui, I think we can make these free. They're not exactly the same, but it's fine.

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -7002,6 +7002,70 @@ interface UserInterface {
...
   clearTerminal(): void;
+
+  /**
+   * Programmatically sets an alias.
+   * @remarks
+   * RAM cost: 0.05 GB
+   *
+   * Programmatically sets an alias. This is functionally equivalent to if you typed ``alias ${alias}=${substitution}`` in the terminal.
+   * This function throws an error if alias/substitution are empty strings after leading and trailing whitespace are removed. It also throws if
+   * alias has any white space in the string in the middle of the string.
```
```suggestion
   * alias has any invalid characters (not alphanum or `|!%,@-`).
```

src/NetscriptFunctions/UserInterface.ts
```diff
@@ -5,11 +5,24 @@ import { defaultTheme } from "../Themes/Themes";
...
 import { Terminal } from "../../src/Terminal";
 import { helpers, wrapUserNode } from "../Netscript/NetscriptHelpers";
 import { assertAndSanitizeMainTheme, assertAndSanitizeStyles } from "../JsonSchema/JSONSchemaAssertion";
 import { LogBoxCloserEvents, LogBoxEvents } from "../ui/React/LogBoxManager";
+import { addGlobalAlias, addAlias, removeAlias, Aliases, GlobalAliases, aliasRegex } from "../Alias";
+import { assertStringWithNSContext } from "../Netscript/TypeAssertion";
+
+/** Converts the provided value to a string and ensures it satisfies the alias condition, throwing if it is not  */
+export function parseAsAlias(ctx: NetscriptContext, argName: string, v: unknown): string {
+  if (typeof v === "number") v = v + ""; // cast to string;
```
d0sboots: Use helpers.string here.

src/NetscriptFunctions/UserInterface.ts
d0sboots: This is fine.

src/NetscriptFunctions/UserInterface.ts
```diff
@@ -189,5 +202,89 @@ export function NetscriptUserInterface(): InternalAPI<IUserInterface> {
       helpers.log(ctx, () => `Clearing terminal`);
       Terminal.clear();
     },
+
+    alias: (ctx) => (_alias, _substitution, _global) => {
+      const alias = parseAsAlias(ctx, "alias", _alias);
+      const substitution = helpers.string(ctx, "substitution", _substitution).trim();
```
d0sboots: Don't trim here. Setting it programmatically should set it as-is, even down to quotes and whitespace. (Removing those is part of shell parsing.)

src/ScriptEditor/NetscriptDefinitions.d.ts
d0sboots: I'd add a note that there can only be one alias for a given keyword (either local or global), and a prior one will be silently overwritten.

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -7002,6 +7002,70 @@ interface UserInterface {
...
+   * @remarks
+   * RAM cost: 0.05 GB
+   *
+   * @param alias - The alias to clear.
+   * @returns - True if there was a previous alias set.
+   */
+  unalias(alias: string): boolean;
+
+  /**
+   * Clears all aliases.
```
d0sboots: I don't think this is needed. It can be constructed simply from getAllAliases and unalias.

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -7002,6 +7002,70 @@ interface UserInterface {
...
+   * @remarks
+   * RAM cost: 0.05 GB
+   */
+  clearAliases(): void;
+
+  /**
+   * Gets the substitution for a specific alias name, or "undefined" if it hasn't been set.
```
d0sboots: Similarly, I didn't think we need this - you can getAllAliases and then do a lookup.

d0sboots (commented):
Although I left a bunch of small nit comments, this is basically good to go. Good job.

d0sboots:
A side comment for the future: Please don't use `rebase` + force-push after review has started on a PR. (Use `merge` to resolve conflicts with upstream instead.) Because it rewrites history, it loses progress for the reviewer(s) and they have to go over the whole thing again.

PR #2815, opened 2026-05-27
DNET: Add scaling depth each time you complete the bonus lab

ficocelliguy:
Players have been asking for darknet to take a bit longer at higher levels. Currently, as players get many augs, they can quickly reach the bottom of the 'net and repeatedly complete the bonus lab. I've had multiple people mention that it would be nice if the net got harder or deeper the more you completed it:
https://discord.com/channels/415207508303544321/415213435974975508/1496580682372087881

This pr increases the depth of the darknet each time you complete the bonus lab (up to a cap).
It also slightly decreases the overall server density and eventually caps the number of servers entirely, making later completions harder - both because there is more distance and because there are fewer servers/connections.
Finally, it increases the size of the lab maze (again, up to a cap)

In order to support this, I have also trimmed down some of the components and styling in the darknet, and made ore use of memoization, to ensure the UI is responsive enough to handle the largest nets.

This is deployed for testing at https://ficocelliguy.github.io/bitburner-src/

### DIFF
--- a/src/DarkNet/Constants.ts
+++ b/src/DarkNet/Constants.ts
@@ -9,3 +9,13 @@ export const DarknetConstants = {
 } as const;
 
 export const MAX_PASSWORD_LENGTH = 50;
+export const VERTICAL_CONNECTION_CHANCE = 0.3;
+export const AIR_GAP_DEPTH = 8;
+export const NET_WIDTH = 8;
+export const MAX_NET_DEPTH = 100;
+export const SERVER_DENSITY = 0.6;
+export const LOW_LEVEL_SERVER_DENSITY = 0.7;
+export const MS_PER_MUTATION_PER_ROW = 30_000; // 30 seconds
+export const MAXIMUM_DNET_SERVER_COUNT = 120;
+export const MAXIMUM_DIFFICULTY = 36;
+export const MAX_MAZE_BONUS_SIZE = 50;
--- a/src/DarkNet/Enums.ts
+++ b/src/DarkNet/Enums.ts
@@ -5,10 +5,12 @@ export const HORIZONTAL_CONNECTION_CHANCE = 0.5;
 export const VERTICAL_CONNECTION_CHANCE = 0.3;
 export const AIR_GAP_DEPTH = 8;
 export const NET_WIDTH = 8;
-export const MAX_NET_DEPTH = 40;
+export const MAX_NET_DEPTH = 200;
 export const SERVER_DENSITY = 0.6;
 export const LOW_LEVEL_SERVER_DENSITY = 0.7;
 export const MS_PER_MUTATION_PER_ROW = 30_000; // 30 seconds
+export const MAXIMUM_DNET_SERVER_COUNT = 120;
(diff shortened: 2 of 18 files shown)


src/DarkNet/controllers/NetworkMovement.ts
```diff
@@ -214,24 +227,32 @@ export const addLowLevelServersIfNeeded = (): void => {
...
+
 const isImmutable = (server: DarknetServer): boolean =>
   server === DarknetState.openServer || server.isConnectedTo || server.hasStasisLink;
 
 export const moveDarknetServer = (
   server: DarknetServer,
   maxDepthDecrease = 3,
   maxDepthIncrease = 3,
-  startingDepth = server.difficulty,
+  startingDepth = server.depth,
```
d0sboots, line 255: This fundamentally changes how dnet works. Before, servers spawned and were kept in bands near where they spawned - that's why this variable is called startingDepth. Now, difficult servers are free to wander in from high up in the net over time. 

I worry that this will make the early parts of dnet much trickier to solve, especially if you let it sit for a while.

src/DarkNet/ui/ServerStatusBox.tsx
```diff
@@ -31,41 +31,61 @@ export function ServerStatusBox({ server, enableAuth, classes }: DWServerProps):
...
+    server.depth,
+    server.leftOffset,
+    server.blockedRam,
+    server.requiredCharismaSkill,
+    server.caches.length,
+    server.contracts.length,
+    server.messages.length,
+    server.programs.length,
+    server.textFiles.size,
+    server.runningScriptMap.size,
```
d0sboots, line 89: Not only is this gross, it is obviously bug-prone, as evidenced by the bug you've introduced here. (The text can change due to the scripts changing without the size changing.) I don't think you can correctly implement this without still calling the render function for serverstatusbox with the proper parameters. 

Back this part out to it's own change - it's unrelated to the rest, and it's going to require a lot more design than the rest.

src/DarkNet/ui/ServerSummary.tsx
```diff
@@ -69,90 +77,87 @@ export function ServerSummary({
...
+      >
+        <Description className={classes.serverStatusIcon} />
+        {fileCount}
+      </span>,
     );
   }
   if (server.blockedRam) {
     components.push(
-      <Tooltip
+      <span
```
d0sboots, line 152: How does this look now that all the tooltips are gone? How does that even work?

src/DarkNet/Enums.ts
```diff
@@ -2,13 +2,6 @@ import type { _ValueOf, DarknetServerDetails } from "@nsdefs";
 import type { DarknetServerData } from "./utils/darknetServerUtils";
 
 export const HORIZONTAL_CONNECTION_CHANCE = 0.5;
```
d0sboots, line 4: Shouldn't this move to constants, too?

src/DarkNet/Constants.ts
d0sboots, line 1: It doesn't really make sense for these to be in their own object, but you didn't need to clean that up now.

PR #2817, opened 2026-05-28
BUGFIX: Fix an edge case with ram rounding

d0sboots:
Because `maxRam` is not necessarily an integer, `maxRam- 8` will not necessarily be correctly rounded, *even if* `maxRam` is. For instance, `32 * 1.15 = 36.8` (rounds correctly naturally), but `36.8 - 8 = 28.799999999999997` (rounds "wrong.") This happens due to crossing precision boundaries at powers of two.

The correct answer is to *always* use `roundToTwo` whenever there's any chance of a fractional ram amount.

### DIFF
--- a/test/jest/Darknet/Darknet.test.ts
+++ b/test/jest/Darknet/Darknet.test.ts
@@ -67,6 +67,8 @@ import { prestigeAugmentation } from "../../../src/Prestige";
 import { initStockMarket, StockMarket, SymbolToStockMap } from "../../../src/StockMarket/StockMarket";
 import { StockSymbol } from "@enums";
 import { GetAllServers } from "../../../src/Server/AllServers";
+import { roundToTwo } from "../../../src/utils/helpers/roundToTwo";
+import { getRamBlock } from "../../../src/DarkNet/effects/ramblock";
 
 beforeAll(() => {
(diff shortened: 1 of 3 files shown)


test/jest/Darknet/Darknet.test.ts
```diff
@@ -965,3 +967,23 @@ describe("CacheReward", () => {
     }
   });
 });
+
+describe("ramblock", () => {
+  test.each([16, 16.01, 32.01, 64.01])("getRamBlock rounds %d correctly", (maxRam: number) => {
+    // This *must* be done within the function, Jest internally relies on
+    // Math.random so the mock must be restored immediately after.
```
catloversg, line 974: > Jest internally relies on Math.random

Which part of Jest internally relies on Math.random?

Mocking `Math.random` and restoring it immediately after are correct, but I want to know the source of this specific claim.

catloversg (commented):
It looks good to me. I only have a question.

test/jest/Darknet/Darknet.test.ts
d0sboots: I don't know. All I know is that when I tried to do the same thing with beforeEach/afterEach, I got really weird failures.

PR #2818, opened 2026-05-28
DNET: "freeze" a server for testing

ficocelliguy:
Based on recent discussion, players are frustrated by how difficult it can be to develop or debug their darknet scripts when the specific model they are testing against disappears and isn't on the network anymore.

To combat this but still keep the core experience, a new method `ns.dnet.freezeServer()` has been added to the api. When called:
- The target becomes immobile and steady (until destroyed by the next webstorm or aug install)
- The target loses all ram
- No xp or caches are rewarded from the server
- The presence of these frozen servers stuck on the network increases instability, similar to having too many backdoored servers


markdown/bitburner.darknet.md
```diff
@@ -58,6 +58,19 @@ If unsuccessful, more detail may be able to be gathered by using heartbleed() to
...
 
+</td></tr>
+<tr><td>
+
+[freezeServer(host)](./bitburner.darknet.freezeserver.md)
+
+
+</td><td>
+
+Overloads a darknet server with feedback to lock it down. It will no longer move or go offline. However, it also loses all of its connections (excluding the one to the script's current server), and all of its max ram, and no longer gives experience.
```
d0sboots: I don't think connections should be changed. Connections can still be added/removed by servers around it moving, so thematically it doesn't really make sense. And (mostly) isolating the server runs counter to the purpose of having this become an auth testing platform.

d0sboots:
```suggestion
Overloads a darknet server with feedback to lock it down. Similar to status link, it will no longer move or go offline, although servers connected to it may still move. However, it also loses all of its max ram, and no longer gives experience.
```

src/DarkNet/controllers/NetworkGenerator.ts
```diff
@@ -176,7 +177,7 @@ export const loadDarknet = () => {
 };
 
 export const addRandomConnections = (server: DarknetServer) => {
-  if (isLabyrinthServer(server.hostname)) {
+  if (isLabyrinthServer(server.hostname) || isFrozen(server)) {
```
d0sboots: Oh, so *this* is why connections to the lab are so rare. They can only be formed unidirectionally from the outside, everything else has a bidirectional chance.

d0sboots: I think these should act the same as stasis linked servers, i.e. no changes to how they connect.

src/DarkNet/controllers/NetworkMovement.ts
```diff
@@ -396,3 +397,15 @@ export const validateDarknetNetwork = (): void => {
     }
   }
 };
+
+export const isFrozen = (server: DarknetServer | DarknetServerData) => {
+  return !server.maxRam && server.isStationary;
```
d0sboots: I think !server.maxRam is enough here. And perhaps that means this helper isn't needed. In the hypothetical where you can unfreeze servers, you've still want 0 ram to gate things like no cache/no xp etc.

src/NetscriptFunctions/Darknet.ts
```diff
@@ -220,6 +221,44 @@ export function NetscriptDarknet(): InternalAPI<DarknetAPI> {
...
+        if (!serverCheck.success) {
+          return {
+            success: false,
+            code: serverCheck.code,
+            message: serverCheck.message,
+            logs: [],
+          };
+        }
+
+        freezeServer(serverCheck.server, getDarknetServerOrThrow(ctx.workerScript.hostname));
```
d0sboots: I have a somewhat controversial suggestion: freezing the server immediately. This means that the server won't be able to disappear while the call is running, preventing a source of annoyance. It also means crafty players could immediately kill the script after running the call, to save on ram-time. Since you don't *want* lots of frozen servers littering the net, this really isn't much of an advantage.

src/DarkNet/controllers/NetworkGenerator.ts
ficocelliguy: Every server on the bottom row of the network is automatically connected to the lab, it's part of the "addGuaranteedConnection" logic whenever a server is moved. Same with the top row of the network connecting to darknet.

E9cipher:
I think this is a great idea. As ficocellyguy said, many players (myself included) find it difficult to debug the scripts on the darknet, and setting a "debug: freeze the server" API would come handy. What do you think about it?

Btw: if it's not ready yet it's not a hurry, just commenting to reopen this debate.

src/DarkNet/controllers/NetworkGenerator.ts
d0sboots: That doesn't match my experience. I recall watching scripts for a long time, waiting for a node to have a connection to the lab.

src/DarkNet/effects/offlineServerHandling.ts
```diff
@@ -148,7 +148,10 @@ export function expectRunningOnDarknetServer(ctx: NetscriptContext): DarknetServ
...
 }
 
+// Having frozen servers, or more than two backdoored servers, increases darknet instability.
 export function getTimeoutChance() {
-  const backdooredDarknetServerCount = getBackdooredDarknetServers().length - 2;
-  return Math.max(Math.min(backdooredDarknetServerCount * 0.03, 0.5), 0);
+  const backdooredDarknetServerCount = getBackdooredDarknetServers().length;
+  const frozenServers = getAllDarknetServers().filter((s) => !s.maxRam).length;
+  const serversAddingToInstability = frozenServers + Math.max(backdooredDarknetServerCount - 2, 0);
+  return Math.max(Math.min(serversAddingToInstability * 0.03, 0.5), 0);
```
d0sboots: Use clamp from our helpers for this, it's much easier to read.

src/DarkNet/utils/darknetNetworkUtils.ts
```diff
@@ -81,7 +81,7 @@ export const getAllAdjacentNeighbors = (x: number, y: number): DarknetServer[] =
   const rowAbove = getServersOnRowAbove(x, true);
   const rowBelow = getServersOnRowBelow(x, true);
   const neighborsOnRow = getNeighborsOnRow(x, y);
-  return [...rowAbove, ...rowBelow, ...neighborsOnRow];
+  return [...rowAbove, ...rowBelow, ...neighborsOnRow].filter((n) => n.maxRam);
```
d0sboots: Doesn't this remove everything? Since n will be an array? Why isn't TS mad about this? 

You already guarded the individual rows.

d0sboots:
> I think this is a great idea. As ficocellyguy said, many players (myself included) find it difficult to debug the scripts on the darknet, and setting a "debug: freeze the server" API would come handy. What do you think about it?
> 
> Btw: if it's not ready yet it's not a hurry, just commenting to reopen this debate.

Yeah, it just fell off my radar for some reason.

src/DarkNet/controllers/NetworkGenerator.ts
ficocelliguy: The connection is added when any server is added the bottom row of the network (which includes being moved) https://github.com/bitburner-official/bitburner-src/blob/df506f56fa16a591d30a0f9b7087b89d3bc8226a/src/DarkNet/controllers/NetworkGenerator.ts#L226

However, looking into this, I discovered that the connection can be removed later - which I have fixed.

src/DarkNet/utils/darknetNetworkUtils.ts
ficocelliguy: because the other DarknetServer[] arrays are spread, n is a single darknet server.
Yeah, I did filter on the individual methods above, so I've removed this for simplicity.

catloversg:
I self-requested a review because I just found bugs in this PR ...

I guess I'll have to open a follow-up PR later. It's not a critical bug, so it's not urgent.

d0sboots:
Sorry, I don't think I get email notifications about that so it's not something I see most of the time. :(

If you have concerns about a PR but don't have time to explain then fully, I'd leave a brief comment like "Hold up on this until I can investigate more"

E9cipher:
I have one suggestion for this @ficocelliguy. I would make the frozen server "backdoored" so it can be accesed at least from home. An alternative is moving the server adjacent to darkweb, but I find the first more elegant.

I didn't see this in the docs. Could we document this and add an example?

Thanks for your time.

catloversg:
Frozen servers are dummy servers for testing scripts launched from other normal servers. I don't see why you need a convenient way to connect to them from home. If you really want to do that for some reason, backdoor the server that is attacking the frozen server, then connect from there.

E9cipher:
Yeah but that can be tricky especially on the darknet. I think a connection to home would really simplify things...

catloversg:
Simplify what? Why do you need a convenient way to connect to the dummy target?

E9cipher:
To deploy the attacker scripts. This method was added so it was simpler to debug and test vulnerabilities on a specified model. If this model is "difficult" enough (e.g freshInstall vs 2g cellular) it will be deep enough to have to modify your darknet script, and imo it's not a very nice approach. You have to edit the working code, add a 2gb ram maybe more overhead, also make the function where it stays next to it, cancels the replicate part.. or you could make a separate script but this doesn't solve the depth problem either...

ficocelliguy:
you can't deploy attacker scripts there - the computer is frozen (0 ram)! You sacrifice it in exchange for stability. you CAN auth on it from neighboring servers, though - that's the idea. 

You can run the singularity backdoor function, or stasis link, on the server you have scripts on that neighbors the newly-frozen, if you want, I guess?

E9cipher:
You've got a point there, but I think it's easier if you have the server next to home/darkweb, so the auth script (incorrectly, I referred to it as deployer) is minimal. I mean, if you have a standard darknet bot, why overcharge it with all the authenticating logic that you may require to analyze a server properly? It's at least an extra 2gb ram which you could use for another thing (like the hacking script).
I think it's easier if the server is backdoored, so you don't depend on the main not. Also, the neighbor server may move, which also has a negative effect on what you're trying to do.

Was this intended when the method was added?

catloversg:
I think you still don't understand what this feature does.

Let's say you already have access to server `foo`. On `foo`, you run an auth script that tries to find the password for server `bar`. The model of server `bar` is a new one, and your auth script is not very good, so server `bar` may move away before you can write a better script. From `foo`, you call `ns.dnet.freezeServer("bar")` to freeze server `bar`. Now the frozen server `bar` won't grant you any rewards when you authenticate it, but it also won't move, so you can take your time improving your script on server `foo`.

Your suggestion is to make server `bar` backdoor-ed when it becomes frozen. I'll be frank with you, this is nonsense. Your auth script is running on server `foo`. You are updating the auth script on `foo`. You can backdoor `foo` if you want an easy way to access `foo` from home. Again (I have asked this question twice), *why do you want to access server `bar` from home?* You still have not found server `bar`'s password, so it's obviously impossible to run a script on server `bar` . It makes no sense to access server `bar`. What are you going to do after accessing server `bar` from home? Creating some files manually on it? For what purposes?

If you wonder what happens if server `foo` goes offline and you can no longer update your auth script, then the answer is `ns.dnet.setStasisLink`. Now you have a safe and stable server for updating scripts (`foo`) and a stable target (`bar`).

Why do we need to add this feature when we can simply add a dummy target next to darkweb so that you can sit on darkweb, slowly write your auth script, easily debug your script without wasting a stasis link? The answer is that, it's not how dnet is designed.

Dnet has two main challenges. First, dnet contains a big number of cct-like puzzles. You can think of dnet as a new big playground of cct-like puzzles. Second, dnet is volatile. The main network that contains normal servers such as n00dles, joeguns is stable, i.e., their servers are always there. They never move (until you prestige), and they never go offline. Your running scripts run forever until they throw an error or you kill them. However, in dnet, your running scripts can be killed at any time (e.g., when the server restarts). This means you must build a resilient mechanism to sync your scripts and restart them from home. In other words, "sit on darkweb, slowly write your auth script, easily debug your script without wasting a stasis link" is intentionally prohibited.

E9cipher:
Thanks for the reply, I now fully understand the purpose of this feature.
To answer your question and close this:
>  why do you want to access server bar from home?

if `bar` if adjacent to home, it acts like the `stasisLink` feature you described, but without wasting the `stasisLink`. If you can deploy scripts from home to server `bar` it'd be a lot easier. But this is intentionally not allowed, so your approach is the right one.
That said, would you recommend documenting this (both the behavior and your example)?

ficocelliguy:
The current documentation is this:

```
Overloads a darknet server with feedback to lock it down. Similar to status link, it will no longer move or go offline, although servers connected to it may still move. However, it also loses all of its max ram, and no longer gives experience.

This technique is sometimes used to sacrifice a new device that appears on the network to make it easier to probe it for weaknesses and develop scripts against it.
```

It doesn't describe the strategy of how best to use it (intentionally) , but it is intended to describe what the tool does. Is it missing something?

catloversg:
> if bar if adjacent to home, it acts like the stasisLink feature you described, but without wasting the stasisLink. If you can deploy scripts from home to server bar it'd be a lot easier.

`bar` (the target) and all non-darkweb dnet servers cannot be adjacent to home. They can only be adjacent to darkweb. In the hypothetical case where `bar` is adjacent to darkweb, darkweb is `foo` in my example. Note that you never need to deploy scripts to server `bar`. `bar` is the target, so there is no need to deploy anything to it. You need to deploy your scripts to server `foo`, not `bar`.

Regarding the documentation, I think the current documentation is already good.

PR #2823, opened 2026-05-29
CITY: Add BN hint to special locations

### LINKED ISSUE #2819: CITY: add hints to BN numbers on special city locations
I was thinking about adding hints to special locations in the city, e.g Sector-12's City Hall. Maybe add something like "Check out BN-3 to uncover the truth about this location". Same for Chongqing's Church of the Machine God, "Check out BN-13 to uncover the truth about this location",.and others. Of course, players on BN 1.1 wouldn't have the hint.

I have created an issue because I don't know if this will be accepted. If it does, I can take the PR for it.

Writing this on a phone, so apologies if there are any typos.

E9cipher:
closes #2819 

Add BN hints to special locations. For example, Sector-12's City Hall now has the following text when the player discovers the Bitverse:
Screenshot from 2026-05-29 21-37-05

Places that changed and now contain the hint:
**Sector-12:** City Hall (Corp)
**Sector-12:** NSA (Bladeburner)
**Chongqing:** Church (CotMG)
**New Tokyo:** Vitalife (Sleeves)

@catloversg I was thinking maybe we can use `CorruptibleText` for those texts. What do you think?

### DIFF
--- a/src/Locations/ui/SpecialLocation.tsx
+++ b/src/Locations/ui/SpecialLocation.tsx
@@ -158,13 +158,28 @@ export function SpecialLocation(props: SpecialLocationProps): React.ReactElement
     );
   }
 
+  function specialLocationNextBNHint(bn_number: number): React.ReactElement {
+    if (knowAboutBitverse()) {
+      return (
+        <>
+          <br />
+          <br />
+          <Typography>You should check out BN-{bn_number} to uncover more details about this place.</Typography>
+        </>
+      )
+    }
+    return <></>;
+  }
+
   function CreateCorporation(): React.ReactElement {
     const [open, setOpen] = useState(false);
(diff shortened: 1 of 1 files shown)


src/Locations/ui/SpecialLocation.tsx
```diff
@@ -165,6 +181,7 @@ export function SpecialLocation(props: SpecialLocationProps): React.ReactElement
         <>
           <Typography>
             <i>A businessman is yelling at a clerk. You should come back later.</i>
+            {specialLocationNextBNHint(3)}
```
d0sboots: Should be outside the typography block.

d0sboots:
I think it is fine to have this be straightforward (no corruptible text) in the endgame. Reexploration is relatively rare, and the secrets are meant to have "opened up" then.

PR #2826, opened 2026-05-30
BUGFIX: Stop using eval when parsing price and amount strings to prevent memory leaks

Sphyxis:
This removes all the eval's in corp code.  These evals, when combined with an open dev console, would cause a memory leak and crash the console and the game.

In short, the eval usage in corps created something in memory that, when combined with the debug console being open, resulted in a memory leak.  This leak would eventually crash the debug console renderer process and later the game itself.

I tested the changes against my corp script and was able to reach round 5 in the same amount of time as usual.

*** I did not however test all alternate export string combinations ***

To reproduce the issue in the current Dev build:

Start a corps.  Get it up to where it starts to export - round 3 should be good.
Hit F12 to open the debugger console.
Open task manager and verify that there are 2 renderer process for Bitburner.
Run the code below to reproduce the issue when in the current Dev build.
Memory usage should balloon for the consoles render process within 20k count.

Redo these steps with the changes in this PR.
You should be able to get past 200k fairly quickly.

MRE Code:
```js
/** @param {NS} ns */
  ns.disableLog("ALL")
  globalThis.webpackRequire ?? webpackChunkbitburner.push([[-1], {}, w => globalThis.webpackRequire = w]);
  const skippedModuleIds = new Set(Object.keys(webpackChunkbitburner[0][1]));
  let p; //Will be the player
  Object.keys(webpackRequire.m).filter(id => !skippedModuleIds.has(id)).forEach(k => Object.values(webpackRequire(k)).find(f => { if (typeof f?.giveExploit === "function") p = f }))
  let count = 0
  let tm = performance.now()
  ns.ui.openTail()
  ns.printf("Count: " + count)
  while (true) {
    p.corporation.process()
    count++
    if (performance.now() - tm > 200) {
      ns.clearLog()
      ns.printf("Count: " + count)
      tm = performance.now()
      await ns.sleep(0)
    }
  }
}
```


src/Corporation/Division.ts
d0sboots: All of this is unneeded now, isn't it?

src/Corporation/FormulaEvaluator.ts
```diff
@@ -0,0 +1,229 @@
+type FormulaNode =
```
d0sboots: This all seems well-constructed, but I'd rather not maintain a whole parser just for this one corner-case of the game. Why not use something like expr-eval instead?

src/Corporation/Division.ts
Sphyxis: Possibly.  I was being overly careful not to take anything out that didn't need to go.  I can take another look.

src/Corporation/FormulaEvaluator.ts
Sphyxis: I've never actually heard of it before.  Looking at it closer it seems the .evaluate() or .parse().evaluate() should work just fine without creating the same memory issue.  It's .toJSFunction() that has to be avoided there.  I'll redo this using that.

Sphyxis:
I've tried to keep the main bulk of the changes inside the helper.ts file.  Let me know if anything else needs modifying.

src/Corporation/helpers.ts
```diff
@@ -8,6 +8,43 @@ import * as corpConstants from "./data/Constants";
...
+corpFormulaParser.functions = {};
+corpFormulaParser.consts = {};
+
+function getCorpFormulaExpression(formula: string): ReturnType<typeof corpFormulaParser.parse> {
+  const parsed = corpFormulaParser.parse(formula);
+  return parsed;
+}
+
+export function evaluateCorpFormula(formula: string, variables: Readonly<Record<string, number>>): number {
+  const result = getCorpFormulaExpression(formula).evaluate(variables) as unknown;
```
d0sboots: I think you can use the 2-arg form of evaluate and get rid of the helper function entirely.

Otherwise, looks good - even simpler than I hoped!

Sphyxis: You are correct.  I've switched to using the 2 arg version and removed the helper.  I also added a few lines of comments as some things are not immediately obvious as to what they are for.

Sphyxis:
I've switched it to the 2 arg version of the .evaluate().  Tested again with no issues after 200k cycles.  No issues with my corp script afterwards.

catloversg (approved):

PR #2827, opened 2026-05-30
DNET: grow lab with bonus lab completions; replace bonus lab reward with new aug

ficocelliguy:
This PR adds a new augment (the "damp socks" discussed in Discord). It is the "Thread of Ariadne", a reference to the woman who found a way to map the labyrinth and return in mythology by marking your path as you go.

This aug can be acquired multiple times, which upgrades its effect (1% to all stats) each time. However, unlike NFG, the effect is additive with each upgrade, not multiplicative. In addition to that source of diminishing returns, the bonus lab also rewards fewer of them over time, down to just one per completion.

This PR also increases the size of the maze in the bonus lab based on the number of Threads the player has (and also scales up the offsets possible for the starting and ending points), up to a cap. This adds some extra challenge to the maze for players pushing for many lab completions.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4775,9 +4775,6 @@
         "arm64"
       ],
       "dev": true,
-      "libc": [
-        "glibc"
-      ],
       "license": "Apache-2.0 AND MIT",
       "optional": true,
       "os": [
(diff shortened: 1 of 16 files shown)


d0sboots:
I have Opinions about how the augs are being handled, but it's too much for comments. I'm gonna do a refactoring PR.

ficocelliguy:
@d0sboots if there's anything I can help with on that, let me know. I'm happy to restructure things.

PR #2830, opened 2026-05-30
API: Add `ns.ui.openCodeEditor()` to open files in the script editor

ficocelliguy:
Players have asked for ways to programmatically open the nano editor, or make custom clickable file links like the terminal `ls` command does, for the purpose of more rich script output or full custom UI. Currently the only way to do this is terminal injection or webpack exploits.

This PR adds a method to the `ui` namespace that allows for opening text or script files in the editor. The player can specify if they want to open the file(s) in vim or nano mode.

In the future, d0s has proposed also adding a full clickable file link getter for use with tPrintRaw, but I wanted to keep this first PR simple.


src/NetscriptFunctions/UserInterface.ts
```diff
@@ -189,5 +193,18 @@ export function NetscriptUserInterface(): InternalAPI<IUserInterface> {
       helpers.log(ctx, () => `Clearing terminal`);
       Terminal.clear();
     },
+
+    nano: (ctx: NetscriptContext) => (_files: any, _vim: any) => {
+      const files = !_files ? [] : Array.isArray(_files) ? _files : [_files];
+      const fileNames = files.map((f) => {
+        const path = helpers.filePath(ctx, "fileName", f);
+        if (!hasScriptExtension(path) && !hasTextExtension(path)) {
+          throw errorMessage(ctx, `Only scripts or text files can be edited. Invalid file path: ${path}`);
```
catloversg:
```suggestion
          throw errorMessage(ctx, `Only scripts and text files can be edited. Invalid file path: ${path}`);
```

src/NetscriptFunctions/UserInterface.ts
```diff
@@ -189,5 +193,18 @@ export function NetscriptUserInterface(): InternalAPI<IUserInterface> {
...
+        }
+        return path;
+      });
+      const vim = helpers.boolean(ctx, "vim", _vim ?? false);
+      commonEditor(vim ? "vim" : "nano", { args: fileNames, server: ctx.workerScript.getServer(), vim }, true);
```
catloversg: With the nano and vim CLI commands, the specified file(s) are opened on the server that the terminal is currently connected to. With this API, files are opened on the server that the current script is running on. This is a behavioral discrepancy.

I'm fine with it, but the TSDoc should explicitly document this difference.

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -7017,6 +7017,17 @@ interface UserInterface {
...
   clearTerminal(): void;
+
+  /**
+   * Opens the specified file(s) in the nano or vim code editor
+   *
+   * @param files - Optional. The file(s) to open in the editor. If not provided, opens the editor to the last edited file, if any.
+   * @param vim - Optional. If true, opens the editor in `vim` mode rather than `nano`
+   *
+   * @remarks
+   * RAM cost: 0.2 GB
```
catloversg: Optional: Document the rationale for the 0.2 GB RAM cost in the PR description.

Players occasionally ask why an API costs foo GB instead of bar GB, and we often cannot answer that question. We can sometimes guess the original reasoning, but we cannot be certain because the original developer usually has left long ago and the rationale was never documented. The explanation can be as simple as "I copied it from another API", but it should still be recorded somewhere.

Note that if you copy the RAM cost from the `clearTerminal` API, it should be 0 GB. That API originally cost 0.2 GB, but it was changed years ago and the current TSDoc is outdated. I'll update it later.

A 0 GB RAM cost is fine with me.

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -7017,6 +7017,17 @@ interface UserInterface {
    * RAM cost: 0.2 GB
    */
   clearTerminal(): void;
+
+  /**
+   * Opens the specified file(s) in the nano or vim code editor
```
catloversg: Optional: `Opens the specified file(s) in the code editor`.

Technically, we only have one code editor. The nano command opens that editor. When the vim command is used, the same editor is opened in the vim mode.

I know this is a fairly questionable nitpick, but I want to be a bit more precise with our wording. The reason is that I have seen players ask why the nano editor is not actually nano. They expect the nano command to open an editor that behaves like the real nano tool, but Bitburner has never had a nano editor (and a vim editor) in that sense.

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -7017,6 +7017,17 @@ interface UserInterface {
...
+   *
+   * @param files - Optional. The file(s) to open in the editor. If not provided, opens the editor to the last edited file, if any.
+   * @param vim - Optional. If true, opens the editor in `vim` mode rather than `nano`
```
catloversg: Optional: "If true, opens the editor in `vim` mode".

catloversg (commented):
In the current state, this API does not work if the current page is the text editor.

Test code (`foo.js`):
```js
/** @param {NS} ns */
  ns.ui.openTail();
  ns.ui.nano("bar.js");
}
```

MRE:
- Run `foo.js` to open the tail log window and `bar.js` in the code editor.
- Close the tab of `bar.js`. Make sure to NOT switch from the code editor to another page.
- Run `foo.js` again via the tail log window. The code editor, which is the current page, does not open `bar.js`.

You can try to solve this issue, but I'm fine if you don't do it. I'll open a separate PR for that.

ficocelliguy:
I've put up a simple fix for the issue where the files are not opened if you are already on the editor page.

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -7014,9 +7014,22 @@ interface UserInterface {
   /**
    * Clear the Terminal window, as if the player ran `clear` in the terminal
    * @remarks
-   * RAM cost: 0.2 GB
+   * RAM cost: 0.0 GB
```
catloversg: Don't change it here. This API is unrelated to this PR. You should do that in #2831.

By the way, "0.0 GB" should be "0 GB". We use "0 GB" everywhere else.

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -7014,9 +7014,22 @@ interface UserInterface {
...
+  /**
+   * Opens the specified file(s) in the code editor.
+   *
+   * This opens files from the server the script is running on, which may be different than the server the terminal is connected to.
+   *
+   * @param files - Optional. The file(s) to open in the editor. If not provided, opens the editor to the last edited file, if any.
+   * @param vim - Optional. If true, opens the editor in `vim` mode
+   *
+   * @remarks
+   * RAM cost: 0.0 GB
```
```suggestion
   * RAM cost: 0 GB
```

src/NetscriptFunctions/UserInterface.ts
```diff
@@ -189,5 +195,23 @@ export function NetscriptUserInterface(): InternalAPI<IUserInterface> {
...
+      const fileNames = files.map((f) => {
+        const path = helpers.filePath(ctx, "fileName", f);
+        if (!hasScriptExtension(path) && !hasTextExtension(path)) {
+          throw errorMessage(ctx, `Only scripts and text files can be edited. Invalid file path: ${path}`);
+        }
+        return path;
+      });
+      helpers.log(ctx, () => `Opening files: ${files.join(", ")}`);
+      const vim = helpers.boolean(ctx, "vim", _vim ?? false);
+      if (Router.page() === ComplexPage.ScriptEditor) {
```
catloversg: This only mitigates the issue, not fix it. You can see the React warning in the dev console. In fact, this workaround hides the real issue in `ScriptEditorRoot.tsx`.

Fixing this issue requires changes in `ScriptEditorRoot.tsx` and other related files. It's not easy, given the awkward design of code in `src/ScriptEditor/ui`. Trying to fix it in this PR is doable, but I don't recommend it. I'll handle it in a separate PR unless you specifically want to address it here.

src/NetscriptFunctions/UserInterface.ts
```diff
@@ -189,5 +195,23 @@ export function NetscriptUserInterface(): InternalAPI<IUserInterface> {
       helpers.log(ctx, () => `Clearing terminal`);
       Terminal.clear();
     },
+
+    nano: (ctx: NetscriptContext) => (_files: any, _vim: any) => {
```
catloversg:
```suggestion
    nano: (ctx: NetscriptContext) => (_files, _vim) => {
```

src/NetscriptFunctions/UserInterface.ts
ficocelliguy: I've rolled back my fix to keep this PR clean.

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -7017,6 +7017,19 @@ interface UserInterface {
...
+
+  /**
+   * Opens the specified file(s) in the code editor.
+   *
+   * This opens files from the server the script is running on, which may be different than the server the terminal is connected to.
+   *
+   * @param files - Optional. The file(s) to open in the editor. If not provided, opens the editor to the last edited file, if any.
+   * @param vim - Optional. If true, opens the editor in `vim` mode
+   *
+   * @remarks
```
catloversg, line 7027: `@remarks` should be above `This opens files from the server ...`.

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -7017,6 +7017,19 @@ interface UserInterface {
...
+   * RAM cost: 0 GB
+   */
+  nano(files?: string | string[], vim?: boolean): void;
```
d0sboots: Please make an options struct for the 2nd argument. It helps fix ambiguity about what happens when unspecified (we'd like unspecified to mean "open in default editor", but that becomes awkward if it's a tri-state value between undefined/false/true), and also my spider-sense is telling me that we'll be adding more options to this one.

src/NetscriptFunctions/UserInterface.ts
```diff
@@ -189,5 +193,19 @@ export function NetscriptUserInterface(): InternalAPI<IUserInterface> {
...
+      const files = !_files ? [] : Array.isArray(_files) ? _files : [_files];
+      const fileNames = files.map((f) => {
+        const path = helpers.filePath(ctx, "fileName", f);
+        if (!hasScriptExtension(path) && !hasTextExtension(path)) {
+          throw errorMessage(ctx, `Only scripts and text files can be edited. Invalid file path: ${path}`);
+        }
+        return path;
+      });
+      helpers.log(ctx, () => `Opening files: ${files.join(", ")}`);
+      const vim = helpers.boolean(ctx, "vim", _vim ?? false);
```
d0sboots: Per other comment: When unspecified, this should use the default editor, instead of always being nano. IMO that's the principle of least surprise (and there's still a way to set it explicitly).

ficocelliguy: `nano` is the "default" in that it is the one most people use and the one shown in the tutorial, right? or is something else the "default" - and if so, what's the terminal command for that? 

the vim flag is either true or false when opening the editor, just based on if you used the terminal command nano or vim - or did I miss something?

src/ScriptEditor/NetscriptDefinitions.d.ts
ficocelliguy: should it be an enum, in that case, rather than a single-property object? or do you expect some other options asides from editor type to be relevant in the future?

src/NetscriptFunctions/UserInterface.ts
catloversg: This may not be related to what d0sboots said, but we have a setting called `Enable Vim as default editor` (`MonacoDefaultToVim`, default: false). It currently only affects `ls`.

Most people use nano instead of vim, so making the default mode non-vim (vim = false) seems reasonable to me. That's also why this API is called `ns.ui.nano`. If we want to make `nano` not the default mode, maybe we can use `MonacoDefaultToVim`? However, in that case, we should call this API `ns.ui.openCodeEditor()`.

d0sboots: Yes, that is the setting I was thinking of. I had forgotten that it only applied to links, but I think it makes sense to do the function renaming and apply it here. I'm thinking of usecases like using someone else's script and having it respect your settings (without requiring explicit work).

src/ScriptEditor/NetscriptDefinitions.d.ts
d0sboots: I get the feeling that other options aside from editor type will be relevant in the future. It's just a feeling, and I might be wrong. But I'm also taking inspiration from the DOM apis, which have similar issues as us (even more strict API compatibility issues), and have very broadly moved to options structs even for single optional arguments. 

For the value itself, it can just be a bool. Having it in the struct gives it a more natural tristate (absent/false/true) than as an optional arg, even though technically you read them the same way. (We don't *really* want to distinguish between absent and undefined, even though that difference is detectable.)

ficocelliguy: Added an options object.

src/NetscriptFunctions/UserInterface.ts
ficocelliguy: If the options object doesn't have a `vim` property, I now fall back on the game setting.

PR #2832, opened 2026-06-01
MISC: Fix some typos and refactor some achievements

E9cipher:
I have edited some achievements' descriptions. They were inconsistent with each other, e.g some had "Bladeburners" capitalized while others had it lowercased. Same for other expressions, such as "Wall Street", "BitNode"... etc. I also removed the final periods of the titles and added some missing ones to the descriptions.

I have only edited titles and descriptions. The "ID" column remains untouched.


src/Achievements/AchievementData.json
```diff
@@ -388,13 +388,13 @@
     },
     "INDECISIVE": {
       "ID": "INDECISIVE",
-      "Name": "Too many options.",
-      "Description": "Spend 1h straight on the bitverse."
+      "Name": "Too many options",
+      "Description": "Spend 1h straight on the bitVerse."
```
catloversg:
```suggestion
      "Description": "Spend 1h straight on the BitVerse."
```

src/Achievements/AchievementData.json
```diff
@@ -519,7 +519,7 @@
     "EDITSAVEFILE": {
       "ID": "EDITSAVEFILE",
       "Name": "Exploit: edit",
-      "Description": "Acquire the EditSaveFile Source-File -1"
+      "Description": "Edit your savefile."
```
catloversg, line 522: The old description is better. The SF -1 achievement description should be vague. "Acquire the EditSaveFile Source-File -1" is not very vague, but it's still less clear than the new description.

src/Achievements/AchievementData.json
d0sboots: Agreed, camelCase is weird for prose.

src/Achievements/AchievementData.json
d0sboots: I think this one is OK. A lot of the other SF -1s are equally explicit. ("Open the dev menu," "Beat BN1 in megabyteburner 2000.") And the prose for this one flows naturally. All of the steam achievements are marked as hidden, so they don't reveal the description without you explicitly clicking on them to find out.

PR #2833, opened 2026-06-02
MISC: Request strict durability for save transactions

### LINKED ISSUE #2809: Error: Save data exists, but its type is invalid. (at "Loading Screen")
### How did this happen?

Please fill this information with details if relevant.

- [ ] Save file
- [ ] Minimal scripts to reproduce the issue
- [ ] Steps to reproduce

My PC just restarted (not sure what the cause is, I am just browsing on Chrome while running Bitburner). When I launched Bitburner after the restart, it is on recovery mode.

Soft reset doesn't work as well and I haven't saved recently so I am not keen on deleting the save.

### Environment

* Error: Error: Save data exists, but its type is invalid.
* Page: Loading Screen
* Version: v3.0.1 (3162fd2)
* Environment: Production
* Platform: Steam
* UserAgent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) bitburner/3.0.1 Chrome/146.0.7680.216 Electron/41.4.0 Safari/537.36
* Features: lang=en-US cookiesEnabled=true doNotTrack=null indexedDb=true
* Source: undefined

### Stack Trace
```
Error: Save data exists, but its type is invalid.
    at s.onsuccess (file:///C:/Program%20Files%20(x86)/Steam/steamapps/common/Bitburner/resources/app/dist/main.bundle.js:9:947900)
```

### React Component Stack
```
undefined
```

### Save
```
Copy your save here if possible

- I haven't saved for a while since I am not installing augmentations (using grafting). Can you please help me locate where the autosave is? Thanks.
```

Evan109:
## Change Description

Request strict IndexedDB transaction durability when saving game data. This asks the browser to flush outstanding changes to persistent storage before the save transaction reports completion, reducing the risk of an interrupted save leaving corrupted data behind.

Closes #2809

## Testing

- `npx tsc --noEmit`
- `npm run lint:report`
- `npx webpack --mode development`
- `npm run test` was run locally, but the existing Corporation and Sleeve floating-point assertions fail under Node `v23.11.0`. `CONTRIBUTING.md` documents these unrelated failures for Node versions older than v24.

## Screenshots

Not applicable: this change does not affect the UI.

### DIFF
--- a/src/db.ts
+++ b/src/db.ts
@@ -109,7 +109,7 @@ export async function save(saveData: SaveData): Promise<void> {
   }
   const db = await getDB();
   return new Promise((resolve, reject) => {
-    const transaction = db.transaction(["savestring"], "readwrite");
+    const transaction = db.transaction(["savestring"], "readwrite", { durability: "strict" });
     const rejectHandler = () => {
       reject(transaction.error ?? new Error("Cannot save game data. Transaction aborted or encountered an error."));
     };


catloversg:
Don't use `Closes #2809` to automatically close #2809. After we release a version containing this change, if we don't receive any similar reports for a few months, I'll close it manually.

catloversg (approved):

d0sboots:
From my reading of the docs, this *shouldn't* affect anything except in the case of power failures. However, I can easily see bugs in rarely-exercised parts of browser code causing there to be possibilities of data loss when the browser dies unexpectedly, when there should not be. This should significantly reduce the window of those, at the expense of worse performance during saving.

PR #2835, opened 2026-06-02
MISC: Change random server selection when generating coding contracts

catloversg:
The implementation of `getRandomServer` is awkward.

The documentation and the retry loop imply that coding contracts are intended to be generated only on NPC servers. However, with the current implementation, there is a very small chance that a contract will be generated on home, a cloud server, or a hacknet server.

The documentation is also incorrect. It states that the chosen server can be home, but `home.purchasedByPlayer` is true, so home can only be selected when the loop reaches its extremely unlikely fallback case.

### DIFF
--- a/src/CodingContract/ContractGenerator.ts
+++ b/src/CodingContract/ContractGenerator.ts
@@ -190,29 +190,18 @@ export function getRandomReward(): ICodingContractReward {
   return { type: getRandomIntInclusive(0, validRewardTypes.length - 1) };
 }
 
-function getRandomServer(): BaseServer | null {
-  const servers = GetAllServers().filter((server: BaseServer) => server.serversOnNetwork.length !== 0);
+export function getRandomServer(): BaseServer | null {
+  const servers = GetAllServers().filter(
+    (server: BaseServer) =>
+      server instanceof Server &&
+      !server.purchasedByPlayer &&
+      server.hostname !== SpecialServers.WorldDaemon &&
+      server.serversOnNetwork.length !== 0,
+  );
   if (servers.length === 0) {
     return null;
   }
-  let randIndex = getRandomIntInclusive(0, servers.length - 1);
-  let randServer = servers[randIndex];
-
-  // An infinite loop shouldn't ever happen, but to be safe we'll use
-  // a for loop with a limited number of tries
-  for (let i = 0; i < 200; ++i) {
-    if (
-      randServer instanceof Server &&
-      !randServer.purchasedByPlayer &&
-      randServer.hostname !== SpecialServers.WorldDaemon
-    ) {
-      break;
-    }
-    randIndex = getRandomIntInclusive(0, servers.length - 1);
-    randServer = servers[randIndex];
-  }
-
-  return randServer;
+  return servers[getRandomIntInclusive(0, servers.length - 1)];
 }
 
 /**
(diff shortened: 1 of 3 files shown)


src/CodingContract/ContractGenerator.ts
```diff
@@ -190,29 +190,18 @@ export function getRandomReward(): ICodingContractReward {
   return { type: getRandomIntInclusive(0, validRewardTypes.length - 1) };
 }
 
-function getRandomServer(): BaseServer | null {
-  const servers = GetAllServers().filter((server: BaseServer) => server.serversOnNetwork.length !== 0);
+export function getRandomServer(): BaseServer | null {
+  const servers = GetAllServers().filter(
+    (server: BaseServer) =>
+      server instanceof Server &&
```
d0sboots, line 196: Don't we have a helper that does this? I feel like we did, but I might be misremembering.

test/jest/CodingContract/ContractGenerator.test.ts
```diff
@@ -57,6 +59,30 @@ describe("Generator", () => {
...
+  test("getRandomServer", () => {
+    Player.sourceFiles.set(9, 3);
+    const ns = getNS();
+    for (let i = 0; i < ns.cloud.getServerLimit(); ++i) {
+      ns.cloud.purchaseServer(`pserver-${i}`, 2);
+    }
+    for (let i = 0; i < ns.hacknet.maxNumNodes(); ++i) {
+      ns.hacknet.purchaseNode();
+    }
+    for (let i = 0; i < 10000; ++i) {
```
d0sboots: Instead of this, which I don't think is enough loops to fail (with any reliability) on the old code, how about forcing random to 0? That always brings bugs out, and I believe will cause the old code to fail right away.

src/CodingContract/ContractGenerator.ts
catloversg: I couldn't find it with a quick check.

test/jest/CodingContract/ContractGenerator.test.ts
catloversg: I added a test with a mocked `Math.random` while keeping the existing loop. That loop can catch regression bugs if we accidentally change the selection conditions in `getRandomServer`.

For example:
```ts
  const servers = GetAllServers(true).filter(
    (server: BaseServer) =>
      !server.purchasedByPlayer &&
      server.hostname !== SpecialServers.WorldDaemon &&
      server.serversOnNetwork.length !== 0,
  );
```
Or
```ts
  const servers = GetAllServers().filter(
    (server: BaseServer) =>
      server instanceof Server &&
      !server.purchasedByPlayer
  );
```

test/jest/CodingContract/ContractGenerator.test.ts
```diff
@@ -57,6 +60,38 @@ describe("Generator", () => {
...
+    const testRandomServer = () => {
+      const server = getRandomServer();
+      if (server === null) {
+        throw new Error(`getRandomServer does not return a server`);
+      }
+      if (
+        !(server instanceof Server) ||
+        server.purchasedByPlayer ||
+        server.hostname === SpecialServers.WorldDaemon ||
+        server.serversOnNetwork.length === 0
```
hexagonrecursion, line 82: I fail to see the point of adding this test to the test suite. It just duplicates the code under test. If one has a bug, the other has the same bug.

catloversg: https://github.com/bitburner-official/bitburner-src/pull/2835#discussion_r3346454861

d0sboots: Yeah, it's not testing that that expression is correct (obviously). It tests that the surrounding code correctly correctly filters using that expression. It's a bit borderline, but still a valid test IMO.

PR #2841, opened 2026-06-03
BITNODE: Wrap BitNode SF text into a reusable function and edit BN8's description

### LINKED ISSUE #2840: BITVERSE: reuse "Destroying this BitNode will give you..." text and change BN8's description
This issue proposes two changes:
1. When you go to the BitVerse and click on a BitNode, the popup always has the text
```markdown
Destroying this BitNode will give you Source-File ??, or if you already have this Source-File, it will upgrade its level up to a maximum of 3.
```
The only thing that changes is the `??` which is the BitNode number. This proposal creates a function that returns this text, making it reusable for future BitNodes instead of copy-pasting it every time.

2. BN8's description in the BitVerse says "Your starting money is 250e6". I have changed the text to "Your starting money is $250 million" for better readability.

@catloversg You were the last editor of this code. Does this look good to you?

I also saw some pull request that did exactly the opposite of my step 2. If this gets merged, I would prefer adding a comment for future editions saying something like "this stays like this". Same thing if step 2 doesn't get merged

E9cipher:
Context in and closes #2840

### DIFF
--- a/src/BitNode/BitNode.tsx
+++ b/src/BitNode/BitNode.tsx
@@ -37,7 +37,7 @@ class BitNode {
 }
 
 export const BitNodes: Record<string, BitNode> = {};
-function replaceSourceFileTextForBN(sourceFileNum: number) {
+function upgradeTextForBN(sourceFileNum: number) {
   const sourceFileText = `Destroying this BitNode will give you Source-File ${sourceFileNum}, or if you already have this Source-File, it will upgrade its level up to a maximum of 3.`;
   return sourceFileText;
 }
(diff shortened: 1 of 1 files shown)


src/BitNode/BitNode.tsx
```diff
@@ -37,6 +37,10 @@ class BitNode {
 }
 
 export const BitNodes: Record<string, BitNode> = {};
+function replaceSourceFileTextForBN(sourceFileNum: number) {
```
d0sboots: This is sort of a mouthful. Also, the purpose of this function is in what it returns (it doesn't actually do any string replacing really), so I'd suggest
```suggestion
function upgradeTextForBN(sourceFileNum: number) {
```

PR #2843, opened 2026-06-03
BUGFIX: Removed WeakRef in ModuleCache, preventing duplicate module creation leading to a memory leak

Sphyxis:
A weakRef ends up being garbage collected, creating an undefined entry for cache lookups, forcing duplicate modules to be created.  In short, modules are not always being recycled/cached for reuse.  They are being forgotten about and then recreated.

This is the pattern:

- Compile a script, say script A
- Create blob URL A
- import(blob URL A), creating Module A
- Store only a WeakRef(LoadedModule A)
- Later, the script.mod is cleared.  This can happen for a multitude of reasons, including no scripts running and content changed (even if it's the same content)
- GC collects the LoadedModule A wrapper
- Module A is not recycled, it remains in memory
- Later, the same source is compiled again
- WeakRef is empty, or the whole cache entry is gone
- Create blob URL A2
- import(blob URL A2), creating Module A2
- Now, we have both module A and module A2 in existence at the same time.
- Repeat

The fix simply removes the WeakRef from the equation, along with the registered cleanup code - which does not cleanup modules anyway.

Below is the MRE to show this in action.  I alternate between 2 files, run them, wait for them to finish, and then rewrite them and do it over again.  Doing this in dev will slowly bring up the memory usage in the renderer.  Within 15 minutes it should start to hover around 580-640mb used.  After few hours, it's over 1000mb.  Doing this with the fix will show the memory usage hovering between 500-600mb over multiple hours.

Monitor script uptime in Active Scripts.  Since it takes a while to see, printing updates eats up memory as well and skews the results.

```js
/** @param {NS} ns */
  ns.disableLog("ALL")

  const victim = "module-cache-victim.js"
  const bodyA = `
  ns.writePort(1, "A:" + ns.args[0])
}
`;
  const bodyB = `
  ns.writePort(1, "B:" + ns.args[0])
}
`;

  ns.clearPort(1)
  let i = 0
  let startTime = performance.now()
  let timer = performance.now()

  while (true) {
    i++
    ns.write(victim, i % 2 === 0 ? bodyA : bodyB, "w")
    /**
     * The ns.exec commented out will also show that modules are not being recreated when written and then executed,
     * even when an argument is used,
     * but it will slowly increase memory usage as it create a unique Runscript key - they add up over time.
     * It would only be needed to verify module caching works regardless of arguments - just based on script.
     * I've alreadey verified this.
     * 
     * const pid = ns.exec(victim, "home", { threads: 1, temporary: true, ramOverride: 1.6 }), i) 
     */
    //
    const pid = ns.exec(victim, "home", { threads: 1, temporary: true, ramOverride: 1.6 })
    if (pid === 0) {
      ns.ui.openTail()
      ns.printf("Failed to start victim at iteration " + i)
      ns.printf("Run Time: " + ns.format.time(performance.now() - startTime))
      return
    }

    await ns.nextPortWrite(1)
    ns.readPort(1)

    /** 
     * This performance monitor creates an increase in memory usage over time.
     * I wouldn't suggest using it as it skews the results.
     */ 

    /*if (performance.now() - timer >= 200) {
      ns.clearLog()
      ns.printf("module-cache MRE iteration " + i + "\nRun Time: " + ns.format.time(performance.now() - startTime))
      timer = performance.now()
      await ns.sleep(0)
    }*/

    //Track the time through active scripts.
    if (performance.now() - timer >= 200) {
      timer = performance.now()
      await ns.sleep(0)
    }
  }
}
```


Sphyxis:
Specifically fixes this issue:

image

Sphyxis:
Closes: #2853

d0sboots:
I am not totally against changing this, but I want to understand more, because I can't reproduce the issue and your analysis is flawed.

Specifically, you say:
> Later, the script.mod is cleared. This can happen for a multitude of reasons, including no scripts running and content changed (even if it's the same content)

However, this is not true. The `script` in question here is a `Script` instance on the filesystem; it does not get cleared when an instance dies. Neither `WorkerScript` nor `RunningScript` have direct access to this field, and the only thing that invalidates it is changes to the content. Furthermore, content changes are checked to see if they are identical before invalidating.

As a result, you specifically have to *change* the content of a single file (and then import it) in order to create a new LoadedModule, which is what your test does. However, this is also the simplest way to force a fresh set of module-scoped variables to happen - in other words, the very behavior you are "fixing" is also used as a feature by people.

I tried loading the savegame in the linked issue. However, I couldn't get the error to happen. Also, running that game for a while did not cause memory to bloat noticably. Looking through the heap snapshots, nothing stood out as accumulating - the largest thing of note was the number of copies of the source of `darknet.jsx`. I'm not sure why there were multiple copies - a simple scp should do a simple assignment and preserve object identity. However, since that script is 600+kb, if lots of copies are being made it's a potential culprit for RAM increases.

catloversg:
This change is okay. I have actually been considering making a similar change, but for a different reason.

A few weeks ago, yichi reported a strange behavior related to module recompilation. I tracked it down to the use of `FinalizationRegistry` and `WeakRef`. I do not remember all of the details off the top of my head, but my MRE demonstrated that, in certain edge cases, module compilation behavior depends on when garbage collection occurs. In other words, the same script can behave differently each time it is run, even when everything in the script is deterministic. I will dig up that MRE and write a full report here.

Regarding the memory leak, I think it exists, but the MRE provided here is not sufficient. For a memory leak of this nature, a good MRE should reproduce the issue within a few seconds or minutes, not hours or days. I ran some tests yesterday using variations of this MRE, and the results were inconsistent. I can observe the leak, but the heap analysis results are inconsistent. I need more time to investigate this issue.

Another thing to note is the issue related to swc. I would like to see more evidence before concluding that it is related to the issue(s) this PR is attempting to resolve. For example, if you believe a memory leak is the cause, please provide an MRE that consistently triggers the error. That should be relatively straightforward: intentionally create a memory leak in the MRE and then try to run a jsx or tsx file.

Sphyxis:
Thank you for clarifying when a script gets invalidated.  Here is a better MRE to test with.  Just give yourself a lot of RAM on home.

```js
/** @param {NS} ns */
  ns.disableLog("ALL")

  // Increase if you have lots of RAM and want the failure faster.
  const MAX_LANES = 256

  // Heap pressure to encourage WeakRef targets to be collected on the old build.
  const PRESSURE_CHUNKS = 1024
  const PRESSURE_ARRAY_SIZE = 8192

  const freeRam = ns.getServerMaxRam("home") - ns.getServerUsedRam("home")
  const lanes = Math.max(1, Math.min(MAX_LANES, Math.floor((freeRam - 32) / 1.6)))

  function victimName(lane) {
    return `${"module-cache-victim/fast-"}${lane}.js`
  }

  function victimBody(lane, phase) {
    // This creates a finite hot set: lanes * 2 possible source strings.
    // On the fixed branch, module count should stop growing after this set is cached.
    return `/** @param {NS} ns */
  const marker = ${lane * 2 + phase}
  ns.writePort(ns.pid, marker)
}
`
  }

  async function readPidPort(pid) {
    const immediate = ns.readPort(pid)
    if (immediate !== "NULL PORT DATA") return immediate

    await ns.nextPortWrite(pid)
    return ns.readPort(pid)
  }

  async function applyGcPressure(cycle) {
    let junk = []
    for (let i = 0; i < PRESSURE_CHUNKS; i++) {
      const arr = new Array(PRESSURE_ARRAY_SIZE)
      for (let j = 0; j < arr.length; j++) arr[j] = cycle + i + j
      junk.push(arr)
    }

    await ns.sleep(0)
    junk = null
    await ns.sleep(0)
  }

  let cycle = 0

  while (true) {
    const phase = cycle & 1

    for (let lane = 0; lane < lanes; lane++) {
      ns.write(victimName(lane), victimBody(lane, phase), "w")
    }

    await applyGcPressure(cycle)

    const pids = []
    for (let lane = 0; lane < lanes; lane++) {
      const pid = ns.exec(victimName(lane), "home", { temporary: true })
      if (pid === 0) {
        ns.ui.openTail()
        ns.printf(`Failed to start lane ${lane} at cycle ${cycle}`)
        ns.printf(`Active lanes: ${lanes}`)
        return
      }
      pids.push(pid)
    }

    await Promise.all(pids.map(readPidPort))

    cycle++
    await ns.sleep(0)
  }
}
```

Sphyxis:
I tried looking for the garbage collection function to call it directly, but it looks like it's not exposed.  Since this leak requires garbage collection it tends to take a while.  In order to reproduce the module error my render process had to get to around 6gb+.  With this MRE though, it should be apparent fairly soon that memory is climbing in Dev and stabilized with this fix.

d0sboots:
So, to clarify, in your experience the swc error is directly tied to memory (inability to allocate in wasm or some such)? That would explain why I can't repro, since my tab will die for OOM before that point.

d0sboots:
I also wanted to mention the reason why the WeakRef exists in the first place. It's not to GC the modules, since even the current comments describe how that's impossible. Rather, it is to release the source code that makes up the key in the cache map. The source is actually surprisingly large compared to the compiled representation, when you look at heap dumps. 

But, I think the original source is also retained as a reference, so there isn't actually any savings.

Sphyxis:
Yes.  Once memory gets high enough it starts throwing the cannot generate module error.  That memory dump was roughly 7.1GB and it took over a day.

Is there a different, more reliable way to clear out the source code that's used as a key?  Could it be reliably transformed into a hash of some sort?  Shrunk down into a different format, like Hex?  Or would that just increase the memory load of the code?

d0sboots:
I don't think there is any point to trying, since I believe it is retained by the compiled code anyway.

I made a test that proves it. Here is the heap snapshot evidence:
image

Here is the test script:
```js
/** @param {NS} ns */
  ns.disableLog("ALL");
  let i = 0;
  const begin = "/*" + " ".repeat(2**20 - 10);
  let promises = [];
  while (true) {
    // We write to a new file every time to avoid triggering the code
    // involved with invalidating data on overwrite.
    const fname = "/testdata/" + i + ".js";
    ns.write(fname, begin + i + "*/", "w");
    promises.push(ns.dynamicImport(fname));
    // We don't need the file once we import, and we don't want the Script
    // to continue holding any references to the module.
    ns.rm(fname);
    i++;
    if (i % 100 == 0) {
      ns.print(i);
      // We don't want to keep a reference to the module either. The test
      // is if the module itself keeps a reference to the source code.
      await Promise.all(promises);
      promises = [];
    }
  }
}
```
BTW if you want to leak a lot of memory really fast, this is how you do it.

Sphyxis:
You mentioned this:
"As a result, you specifically have to change the content of a single file (and then import it) in order to create a new LoadedModule, which is what your test does. However, this is also the simplest way to force a fresh set of module-scoped variables to happen - in other words, the very behavior you are "fixing" is also used as a feature by people."

However, when done properly, there will never be a fresh set of module scoped variables - it will always be reused if it's the same code.  What used to be intermittent failures changes to it working the way you think it should, and it won't suddenly forget and create a new, duplicate module.  You are essentially forced to manage them properly rather than refreshing them.

Would additional testing of some sort help in any way?

d0sboots:
I think the only thing to note here is that this will be a user-visible behavior change, and some set of people will be surprised by it. We've (intentionally) not specified how module sharing between scripts works, so this change doesn't constitute an API break, but it will still be a surprise. Specifically, some people will start getting "old" modules when they were expecting "fresh" modules, mostly when it comes to manually debugging stuff.

But, I think that's fine on the balance. The chance to reduce memory usage for certain other scenarios makes it worth it.

PR #2844, opened 2026-06-04
MISC: Partially revert #2841

catloversg:
There are two issues in #2841:
- `${formatMoney(250e6)}` evaluates to `$$250.000m` (notice the double `$`).
- Using `formatMoney` does not work the way you are expecting. Please check my comments and https://github.com/bitburner-official/bitburner-src/pull/2453#issuecomment-3810566785 for more information.

Refactor our codebase to use `formatMoney` here is doable, but it's not worth the effort. I'll do it when there is sufficient benefit.

### DIFF
--- a/src/BitNode/BitNode.tsx
+++ b/src/BitNode/BitNode.tsx
@@ -2,7 +2,7 @@ import React from "react";
 import { Player } from "@player";
 import { AugmentationName, CityName, CompletedProgramName, FactionName } from "@enums";
 import { BitNodeMultipliers, replaceCurrentNodeMults } from "./BitNodeMultipliers";
-import { formatMoney } from "../ui/formatNumber";
+
 class BitNode {
   // A short description, or tagline, about the BitNode
   tagline: string;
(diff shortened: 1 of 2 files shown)


src/BitNode/BitNode.tsx
```diff
@@ -285,7 +286,10 @@ export function initBitNodes() {
         <br />
         In this BitNode:
         <ul>
-          <li>Your starting money is ${formatMoney(250e6)}.</li>
+          {/* Do NOT call formatMoney. formatMoney applies the player-defined currency symbol settings, but BitNode data
+          is initialized before the save data is loaded, so it always uses the default settings. If we cannot apply the
+          player's settings, just don't call formatMoney. */}
+          <li>Your starting money is 250e6.</li>
```
E9cipher: I would do this with player-friendly numbers, i.e. "$250 million" instead of using scientific notation. I don't see the point in using this notation if this is only meant for the player to read (right?)

catloversg: `250 million` is okay, but not `$250 million`.

PR #2847, opened 2026-06-04
UI: Unify default color handling between `ns.print` and `ns.tprint`

MageKing17:
In Ye Olden Days, both `print` and `tprint` decided which color a line should be by checking `.startsWith` against the various strings ("ERROR", "FAIL", "SUCCESS", "WARN", and "INFO"). While `tprint` still does this, `print` was at some point changed to use regular expressions designed to skip an optional prefix in brackets (basically, the optional timestamp). However, these regular expressions include the "start of string" marker inside the optional portion, which means they now flag a string that contains the relevant text *anywhere*, rather than just at the beginning.

This commit corrects this error in the regexps and splits the color-determination logic into a helper function that is now also used by `tprint` (and `tprintf`), so default color handling is once again the same whether you're printing to a tail log or the terminal.

Tested by simply running a test script that `print`ed and `tprint`ed the same five lines of text, before and after my changes.

`tprint` before:
image
`tprint` after:
image
`print` before:
image
`print` after:
image

### DIFF
--- a/src/Netscript/NetscriptHelpers.tsx
+++ b/src/Netscript/NetscriptHelpers.tsx
@@ -86,6 +86,7 @@ export const helpers = {
   hostReturnOptions,
   returnServerID,
   argsToString,
+  getTextColor,
   errorMessage,
   validateHGWOptions,
   checkEnvFlags,
@@ -365,6 +366,23 @@ function argsToString(args: unknown[]): string {
 
     return (out += String(nativeArg));
   }, "");
+}
+
+/** Determine what default color a string should have for tprint/print. */
+function getTextColor(str: string): "error" | "success" | "warn" | "info" | "primary" {
+  if (str.match(/^(\[[^\]]+\] )?ERROR/) || str.match(/(^\[[^\]]+\] )?FAIL/)) {
+    return "error";
+  }
+  if (str.match(/^(\[[^\]]+\] )?SUCCESS/)) {
+    return "success";
+  }
+  if (str.match(/^(\[[^\]]+\] )?WARN/)) {
+    return "warn";
+  }
+  if (str.match(/^(\[[^\]]+\] )?INFO/)) {
+    return "info";
+  }
+  return "primary";
 }
 
 function validateHGWOptions(ctx: NetscriptContext, opts: unknown): CompleteHGWOptions {
--- a/src/Netscript/NetscriptHelpers.tsx
+++ b/src/Netscript/NetscriptHelpers.tsx
@@ -370,6 +370,8 @@ function argsToString(args: unknown[]): string {
 
 /** Determine what default color a string should have for tprint/print. */
 function getTextColor(str: string): "error" | "success" | "warn" | "info" | "primary" {
+  // Match a tag at the start-of-line with an optional bracket part (primarily for timestamps)
+  // Example: "[13:10] ERROR: Too much regex"
   if (str.match(/^(\[[^\]]+\] )?ERROR/) || str.match(/^(\[[^\]]+\] )?FAIL/)) {
     return "error";
   }
--- a/src/Netscript/NetscriptHelpers.tsx
+++ b/src/Netscript/NetscriptHelpers.tsx
@@ -370,7 +370,7 @@ function argsToString(args: unknown[]): string {
 
 /** Determine what default color a string should have for tprint/print. */
 function getTextColor(str: string): "error" | "success" | "warn" | "info" | "primary" {
-  if (str.match(/^(\[[^\]]+\] )?ERROR/) || str.match(/(^\[[^\]]+\] )?FAIL/)) {
+  if (str.match(/^(\[[^\]]+\] )?ERROR/) || str.match(/^(\[[^\]]+\] )?FAIL/)) {
     return "error";
   }
   if (str.match(/^(\[[^\]]+\] )?SUCCESS/)) {
--- a/src/NetscriptFunctions.ts
+++ b/src/NetscriptFunctions.ts
@@ -430,20 +430,11 @@ export const ns: InternalAPI<NSFull> = {
       const str = helpers.argsToString(args);
       const color = helpers.getTextColor(str);
       switch (color) {
-        case "error": {
-          Terminal.error(`${ctx.workerScript.name}: ${str}`);
-          break;
-        }
-        case "success": {
-          Terminal.success(`${ctx.workerScript.name}: ${str}`);
-          break;
-        }
-        case "warn": {
-          Terminal.warn(`${ctx.workerScript.name}: ${str}`);
-          break;
-        }
+        case "error":
+        case "success":
+        case "warn":
         case "info": {
-          Terminal.info(`${ctx.workerScript.name}: ${str}`);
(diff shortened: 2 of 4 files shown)


src/Netscript/NetscriptHelpers.tsx
```diff
@@ -367,6 +368,23 @@ function argsToString(args: unknown[]): string {
   }, "");
 }
 
+/** Determine what default color a string should have for tprint/print. */
+function getTextColor(str: string): "error" | "success" | "warn" | "info" | "primary" {
+  if (str.match(/^(\[[^\]]+\] )?ERROR/) || str.match(/(^\[[^\]]+\] )?FAIL/)) {
```
catloversg:
```suggestion
  if (str.match(/^(\[[^\]]+\] )?ERROR/) || str.match(/^(\[[^\]]+\] )?FAIL/)) {
```

src/Terminal/Terminal.ts
```diff
@@ -176,8 +176,8 @@ export class Terminal {
     TerminalEvents.emit();
   }
 
-  print(s: string): void {
-    this.append(new Output(s, "primary"));
+  print(s: string, color: "primary" | "error" | "success" | "info" | "warn" = "primary"): void {
```
catloversg: Revert this change. More on this later.

src/NetscriptFunctions.ts
```diff
@@ -428,47 +428,16 @@ export const ns: InternalAPI<NSFull> = {
...
-        Terminal.warn(`${ctx.workerScript.name}: ${str}`);
-        return;
-      }
-      if (str.startsWith("INFO")) {
-        Terminal.info(`${ctx.workerScript.name}: ${str}`);
-        return;
-      }
-      Terminal.print(`${ctx.workerScript.name}: ${str}`);
+      const color = helpers.getTextColor(str);
+      Terminal.print(`${ctx.workerScript.name}: ${str}`, color);
```
catloversg: Currently, `Terminal.print("foo", "error")` is the same as `Terminal.error("foo")`, but that may not remain true in the future.

For example, if we add basic support for pipes and streams, we may also add support for stdout and stderr. In that case, `Terminal.print("foo", "error")` would no longer be equivalent to `Terminal.error("foo")`.

You should perform a switch-case on `color` and call the respective functions.

src/NetscriptFunctions.ts
```diff
@@ -428,47 +428,16 @@ export const ns: InternalAPI<NSFull> = {
...
-        Terminal.warn(`${str}`);
-        return;
-      }
-      if (str.startsWith("INFO")) {
-        Terminal.info(`${str}`);
-        return;
-      }
-      Terminal.print(`${str}`);
+      const color = helpers.getTextColor(str);
+      Terminal.print(`${str}`, color);
```
catloversg: Same comment about switch-case.

catloversg (commented):
Test code:

```js
/** @param {NS} ns */
  ns.ui.openTail();
  ns.clearLog();
  ns.ui.clearTerminal();
  test(ns, "Test");
  test(ns, "Test INFOrmation");
  test(ns, "INFO: Test");
  test(ns, "SUCCESS: Test");

  test(ns, "FAIL: Test");
  test(ns, "foo FAIL: Test");
  test(ns, "[tag] FAIL: Test");
  test(ns, "foo [tag] FAIL: Test");

  test(ns, "ERROR: Test");
  test(ns, "foo ERROR: Test");
  test(ns, "[tag] ERROR: Test");
  test(ns, "foo [tag] ERROR: Test");
}

function test(ns, str) {
  ns.print(str);
  ns.tprint(str);
}
```

In future PRs, please include the test code you used. Screenshots only provide a quick view of the results. I usually test non-trivial changes manually, so having access to the same test code is very useful.

src/Netscript/NetscriptHelpers.tsx
```diff
@@ -367,6 +368,23 @@ function argsToString(args: unknown[]): string {
   }, "");
 }
 
+/** Determine what default color a string should have for tprint/print. */
+function getTextColor(str: string): "error" | "success" | "warn" | "info" | "primary" {
+  if (str.match(/^(\[[^\]]+\] )?ERROR/) || str.match(/^(\[[^\]]+\] )?FAIL/)) {
```
d0sboots, line 375: I know this was (mostly) copied from elsewhere, but I find it's helpful to document regexes with what they're supposed to match. Since these are all similar, just doing one is fine.
```suggestion
  // Match a tag at the start-of-line with an optional bracket part (primarily for timestamps)
  // Example: "[13:10] ERROR: Too much regex"
  if (str.match(/^(\[[^\]]+\] )?ERROR/) || str.match(/^(\[[^\]]+\] )?FAIL/)) {
```

src/NetscriptFunctions.ts
```diff
@@ -428,47 +428,58 @@ export const ns: InternalAPI<NSFull> = {
...
+          break;
+        }
+        case "warn": {
+          Terminal.warn(`${str}`);
+          break;
+        }
+        case "info": {
+          Terminal.info(`${str}`);
+          break;
+        }
```
d0sboots: I don't know if TS will let you get away with this, but if it does, you can do this:
```suggestion
        case "error":
        case "success":
        case "warn":
        case "info":
          Terminal[color](`${str}`);
          break;
```

src/NetscriptFunctions.ts
```diff
@@ -428,47 +428,58 @@ export const ns: InternalAPI<NSFull> = {
...
-      if (str.startsWith("WARN")) {
-        Terminal.warn(`${ctx.workerScript.name}: ${str}`);
-        return;
-      }
-      if (str.startsWith("INFO")) {
-        Terminal.info(`${ctx.workerScript.name}: ${str}`);
-        return;
+      const color = helpers.getTextColor(str);
+      switch (color) {
+        case "error": {
```
d0sboots: Same thing here, if it works.

src/NetscriptFunctions.ts
MageKing17: As a matter of fact, it apparently does. Unsurprisingly, I suppose, since `color` is restricted by the `case` statements.

MageKing17:
Latest test results:

```js
/** @param {NS} ns */
  const test = (str) => {
    ns.print(str);
    ns.tprint(str);
  };
  ns.ui.openTail();
  ns.clearLog();
  ns.ui.clearTerminal();
  test("Test");
  test("Test INFOrmation");
  test("INFO: Test");
  test("SUCCESS: Test");

  test("FAIL: Test");
  test("foo FAIL: Test");
  test("[tag] FAIL: Test");
  test("foo [tag] FAIL: Test");

  test("ERROR: Test");
  test("foo ERROR: Test");
  test("[tag] ERROR: Test");
  test("foo [tag] ERROR: Test");

  test("WARNING: Test");
  test("foo WARNING: Test");
  test("[tag] WARNING: Test");
  test("foo [tag] WARNING: Test");

  test("INFO: Test");
  test("foo INFO: Test");
  test("[tag] INFO: Test");
  test("foo [tag] INFO: Test");
}
```

image

PR #2848, opened 2026-06-04
API: Add `ns.format.money`

MageKing17:
There is currently no way for a user script to format money and have the user-supplied currency symbol be used. This commit just adds `ns.format.money` as a wrapper for `formatMoney` to give users that functionality.


src/NetscriptFunctions/Format.ts
```diff
@@ -34,5 +34,10 @@ export function NetscriptFormat(): InternalAPI<Format> {
       const milliPrecision = !!_milliPrecision;
       return convertTimeMsToTimeElapsedString(milliseconds, milliPrecision);
     },
+    money: (ctx) => (_n, _exp) => {
+      const n = helpers.number(ctx, "n", _n);
+      const exp = !!_exp;
```
catloversg:
```suggestion
      const exp = helpers.boolean(ctx, "exp", _exp ?? false);
```
New APIs should use `helpers.boolean`.

catloversg (approved):

PR #2852, opened 2026-06-04
UI: Prevent accidental pastes when closing an editor tab with middle mouse button

### LINKED ISSUE #2849: Closing a script editor tab with MMB pastes the clipboard into the current script
Here's a clip of the problem in action:

https://github.com/user-attachments/assets/b5ddab57-09f1-4a56-beed-d2a6f6c4edbe

I'm on Fedora 44 KDE, running through Steam with no compatibility layers (native).

This is a problem, since the clipboard contents will not always be immediately script-breaking, possibly causing unexpected logic errors. This is further amplified by the fact that an autosave, or switching to a different tab (i.e. Terminal) will apply these unintended changes without an explicit save command.
What's even worse, clicking off the script editor tab resets the undo action buffer, making the new accidental paste un-undo-able and even harder to look for.

Expected behavior is that an MMB event on an editor's tab does not bubble up to the rest of the editor, and does not cause the (otherwise correct) MMB->Paste action.

I couldn't replicate the problem on the web version, could anyone running on Windows try and replicate this?

PatPL:
PR for issue #2849 

closes #2849 

Tested with a native bitburner-linux-x64 build artifact, first replicated the issue once more on freshly cloned repository, then confirmed the issue is gone with the fix applied.

The issue was caused by the paste action happening OS-side (on mouse up). Usually, clicking somewhere with MMB moves away the focus from the editor (on mouse down), so the (mouse up) does not usually paste anything.

For some reason, on the electron build, the MMB on the button element in the \<Tab\> component doesn't move the focus away from the editor (on mouse down), so the paste still happened in the editor once MMB was released (and after the tab was correctly closed).

This PR kills any MMB (on mouse up) events happening within the \<Tab\>, and \<Tabs\> components. Since MMB is the standard non-remappable keybind for close/open tab in the context of tab containers, no users should be negatively impacted (no impact on potential custom keybinds).

Before:
[Screencast_20260604_131628.webm](https://github.com/user-attachments/assets/344161b9-a921-4821-8fab-ae2d4218ac8e)

After:
[Screencast_20260604_180833.webm](https://github.com/user-attachments/assets/622722c9-71ad-4c94-a252-283fc185b500)

### DIFF
--- a/src/ScriptEditor/ui/Tabs.tsx
+++ b/src/ScriptEditor/ui/Tabs.tsx
@@ -116,12 +116,16 @@ export function Tabs({ scripts, currentScript, onTabClick, onTabClose, onTabUpda
               onWheel={handleScroll}
               onMouseUp={(e) => {
                 // Some Linux DEs do a "paste" action on MMB (mouse up).
-                // Closing a tab with the editor in focus would cause a paste in the editor
-                // This event handler prevents the paste in cases where the last tab is closed
+                // Closing a tab with the editor in focus would cause a paste in the editor.
+                // This event handler prevents the paste in cases where a tab is closed.
                 //
-                // It's important to mention that a regular MMB click on the tab container would
-                // also do nothing here, since the "mouse down" event changes the focus away from
+                // It's important to mention that a regular MMB click on the tab container itself
+                // (not one of the tabs) without this workaround would do nothing here,
+                // since the "mouse down" event changes the focus away from
                 // the editor before the "mouse up" attempts a paste, preventing it altogether.
+                //
+                // This handler also catches onMouseUp events originating from <Tab> (Tab.tsx), in cases
+                // where a tab is closed, and another one slides under the cursor before MMB is released.
                 if (e.button === 1) {
                   e.preventDefault();
                 }
--- a/src/ScriptEditor/ui/Tabs.tsx
+++ b/src/ScriptEditor/ui/Tabs.tsx
@@ -114,6 +114,18 @@ export function Tabs({ scripts, currentScript, onTabClick, onTabClose, onTabUpda
                 overflowX: "scroll",
               }}
               onWheel={handleScroll}
+              onMouseUp={(e) => {
+                // Some Linux DEs do a "paste" action on MMB (mouse up).
+                // Closing a tab with the editor in focus would cause a paste in the editor
+                // This event handler prevents the paste in cases where the last tab is closed
+                //
+                // It's important to mention that a regular MMB click on the tab container would
+                // also do nothing here, since the "mouse down" event changes the focus away from
+                // the editor before the "mouse up" attempts a paste, preventing it altogether.
+                if (e.button === 1) {
+                  e.preventDefault();
+                }
+              }}
             >
               {filteredScripts.map(({ script, originalIndex }, index) => {
                 const { path, hostname } = script;
--- a/src/ScriptEditor/ui/Tab.tsx
+++ b/src/ScriptEditor/ui/Tab.tsx
@@ -125,6 +125,14 @@ export function Tab({ provided, tabId, isActive, isExternal, isUnsaved, onClick,
               onClose();
             }
           }}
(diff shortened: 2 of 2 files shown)


catloversg:
I performed a quick test, and it seems we only need `onMouseUp` in `Tabs.tsx`. What happens on your machine if you remove `onMouseUp` in `Tab.tsx` (`Tab.tsx`, not `Tabs.tsx`)?

PatPL:
Added logging and retested and yeah, you're right. `Tab` bubbles to `Tabs`, and is stopped there anyways. The change in `Tab.onMouseUp` is redundant and can be removed.

There were originally two changes, because closing a tab in the middle, and the last were two different cases, element-under-the-cursor-on-mouse-up-wise.

image

[Screencast_20260604_224629.webm](https://github.com/user-attachments/assets/f7e5a473-e9bd-4fea-8fd7-5c2c14305228)

d0sboots:
TIL that middle-mouse button closes Monaco tabs.

src/ScriptEditor/ui/Tabs.tsx
```diff
@@ -114,6 +114,18 @@ export function Tabs({ scripts, currentScript, onTabClick, onTabClose, onTabUpda
                 overflowX: "scroll",
               }}
               onWheel={handleScroll}
+              onMouseUp={(e) => {
+                // Some Linux DEs do a "paste" action on MMB (mouse up).
+                // Closing a tab with the editor in focus would cause a paste in the editor
+                // This event handler prevents the paste in cases where the last tab is closed
```
d0sboots:
```suggestion
                // Some Linux DEs do a "paste" action on MMB (mouse up).
                // Closing a tab with the editor in focus would cause a paste in the editor.
                // This event handler prevents the paste in cases where the tab is closed.
```

src/ScriptEditor/ui/Tabs.tsx
```diff
@@ -114,6 +114,18 @@ export function Tabs({ scripts, currentScript, onTabClick, onTabClose, onTabUpda
...
+                //
+                // It's important to mention that a regular MMB click on the tab container would
+                // also do nothing here, since the "mouse down" event changes the focus away from
+                // the editor before the "mouse up" attempts a paste, preventing it altogether.
```
d0sboots:
```suggestion
                // It's important to mention that a MMB click on the tab container (not one of the tabs)
                // without this workaround would do nothing here, since the "mouse down" event
                // changes the focus away from the editor before the "mouse up" attempts a paste,
                // preventing it altogether. This handler is only placed here to catch bubbling mouseUps
                // where the mouseDown failed to change focus due to being involved in tab closing.
```

d0sboots (commented):
There's subtle stuff here, so I'm suggesting some comment refinements.

PatPL:
I've updated the comment with the previous change and your suggestions in mind.

I've checked it again and to be honest I'm not exactly certain what's the mechanism behind this problem. In the web version the focus stays in the original element the whole time (without pasting), so it's not fully focus-related. It's also the case in other monaco-adjacent editors like VSCode, or browser tabs in general; You can MMB-close a tab, and go back to writing straight away.

I'd rather leave it somewhat vague in the comment, rather than possibly mislead anyone.

A dedicated MMB click solely within the tab container will still move the focus to the container onMouseDown, which is the default, unchanged behavior. That's what I tried conveying in the comment.

catloversg:
I'll take a look at this later and rewrite the comments if needed. It may take a while, though.

By the way, please avoid rebasing after a PR has been reviewed. The change in this PR is very small, so it is not a problem here. However, for non-trivial PRs, a rebase can make it much harder to track what has changed. In some cases, we even have to review the entire PR from scratch.

PatPL:
Sorry for that, I'll stop it.

I have a habit of force-pushing very minor changes to the very last commit if I catch something quick enough that's not worth a whole new commit to fix.

A merge will squash it all anyways so I should have just fixed it in a new commit.

PR #2856, opened 2026-06-06
BASE: update several packages due to security

mctylr-gh:
Update various package dependencies to address multiple known security issues. Most are moderate severity, but at least one is high security by GitHub Advisory, and at least one is a potential remote DoS.

`jsdom` 20.0.3 -> 26.1.0 (used by jest-environment-jsdom)
Remove abab (due to jsdom 20.0.3) and domexception (jsdom @ 20.0.3) package dependencies. Both packages are *obsolete*.

npm warnings:

* Use your platform's native `atob()` and `btoa()` methods instead
* Use your platform's native `DOMException` instead

`dompurify` 3.2.7 -> 3.4.8 (used by monaco-editor)
Severity: *moderate*

* DOMPurify contains a Cross-site Scripting vulnerability - [GHSA-v2wj-7wpq-c8vv](https://github.com/advisories/GHSA-v2wj-7wpq-c8vv)
* DOMPurify ADD_ATTR predicate skips URI validation - [GHSA-cjmm-f4jc-qw8r](https://github.com/advisories/GHSA-cjmm-f4jc-qw8r)
* DOMPurify USE_PROFILES prototype pollution allows event handlers - [GHSA-cj63-jhhr-wcxv](https://github.com/advisories/GHSA-cj63-jhhr-wcxv)
* DOMPurify's ADD_TAGS function form bypasses FORBID_TAGS due to short-circuit evaluation - [GHSA-39q2-94rc-95cp](https://github.com/advisories/GHSA-39q2-94rc-95cp)
* DOMPurify: FORBID_TAGS bypassed by function-based ADD_TAGS predicate (asymmetry with FORBID_ATTR fix) - [GHSA-h7mw-gpvr-xq4m](https://github.com/advisories/GHSA-h7mw-gpvr-xq4m)
* DOMPurify has a SAFE_FOR_TEMPLATES bypass in RETURN_DOM mode - [GHSA-crv5-9vww-q3g8](https://github.com/advisories/GHSA-crv5-9vww-q3g8)
* DOMPurify: Prototype Pollution to XSS Bypass via CUSTOM_ELEMENT_HANDLING Fallback - [GHSA-v9jr-rg53-9pgp](https://github.com/advisories/GHSA-v9jr-rg53-9pgp)
* DOMPurify is vulnerable to mutation-XSS via Re-Contextualization - [GHSA-h8r8-wccr-v5f2](https://github.com/advisories/GHSA-h8r8-wccr-v5f2)

`launch-editor` 2.6.1 -> 2.14.1 (used by webpack-dev-server)
Severity: **high**

* launch-editor vulnerable to command injection via the crafted request on Windows - [GHSA-c27g-q93r-2cwf](https://github.com/advisories/GHSA-c27g-q93r-2cwf)

`qs` 6.14.2 -> 6.15.2 (used by body-parser -> express -> webpack-dev-server)
Severity: *moderate*

* qs has a remotely triggerable DoS: `qs.stringify` crashes with `TypeError` on null/undefined entries in comma-format arrays when encodeValuesOnly is set - [GHSA-q8mj-m7cp-5q26](https://github.com/advisories/GHSA-q8mj-m7cp-5q26)

`uuid` 8.3.2 -> 14.0.0 (used by sockjs)
Severity: *moderate*

* uuid: Missing buffer bounds check in v3/v5/v6 when buf is provided - [GHSA-w5hq-g745-h8pq](https://github.com/advisories/GHSA-w5hq-g745-h8pq)

`ws` 8.19.0 -> 8.21.0 (used by jsdom)
Severity: *moderate*

* ws: Uninitialized memory disclosure - [GHSA-58qx-3vcg-4xpx](https://github.com/advisories/GHSA-58qx-3vcg-4xpx)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8331,15 +8331,6 @@
         "url": "https://github.com/fb55/domhandler?sponsor=1"
       }
     },
-    "node_modules/dompurify": {
-      "version": "3.4.8",
-      "resolved": "https://registry.npmjs.org/dompurify/-/dompurify-3.4.8.tgz",
-      "integrity": "sha512-yb1cEmaOum7wFvOCSQxyfgVlv5D47Rc30iZWoMpbDIWTnJ6grDDQyu2KFJzB2k7u0pMuJcQ1zphH//fFnw2tjQ==",
-      "license": "(MPL-2.0 OR Apache-2.0)",
-      "optionalDependencies": {
-        "@types/trusted-types": "^2.0.7"
-      }
-    },
     "node_modules/domutils": {
       "version": "2.8.0",
       "resolved": "https://registry.npmjs.org/domutils/-/domutils-2.8.0.tgz",
@@ -14183,6 +14174,15 @@
       "peerDependencies": {
         "monaco-editor": ">= 0.31.0",
         "webpack": "^4.5.0 || 5.x"
+      }
+    },
+    "node_modules/monaco-editor/node_modules/dompurify": {
+      "version": "3.2.7",
+      "resolved": "https://registry.npmjs.org/dompurify/-/dompurify-3.2.7.tgz",
+      "integrity": "sha512-WhL/YuveyGXJaerVlMYGWhvQswa7myDG17P7Vu65EWC05o8vfeNbvNf4d/BOvH99+ZW+LlQsc1GDKMa1vNK6dw==",
+      "license": "(MPL-2.0 OR Apache-2.0)",
+      "optionalDependencies": {
+        "@types/trusted-types": "^2.0.7"
       }
     },
     "node_modules/monaco-vim": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -6482,6 +6482,13 @@
         "url": "https://github.com/sponsors/ljharb"
       }
     },
+    "node_modules/array-flatten": {
+      "version": "1.1.1",
+      "resolved": "https://registry.npmjs.org/array-flatten/-/array-flatten-1.1.1.tgz",
+      "integrity": "sha512-PCVAQswWemu6UdxsDFFX/+gVeYqKAod3D3UVm91jHwynguOwAvYPhx8nNlM++NqRcK6CxxpUafjmhIdKiHibqg==",
+      "dev": true,
+      "license": "MIT"
+    },
     "node_modules/array-includes": {
       "version": "3.1.9",
       "resolved": "https://registry.npmjs.org/array-includes/-/array-includes-3.1.9.tgz",
@@ -6939,31 +6946,6 @@
       "integrity": "sha512-XpNj6GDQzdfW+r2Wnn7xiSAd7TM3jzkxGXBGTtWKuSXv1xUV+azxAm8jdWZN06QTQk+2N2XB9jRDkvbmQmcRtg==",
       "dev": true,
       "license": "MIT"
-    },
-    "node_modules/body-parser": {
-      "version": "2.2.2",
-      "resolved": "https://registry.npmjs.org/body-parser/-/body-parser-2.2.2.tgz",
-      "integrity": "sha512-oP5VkATKlNwcgvxi0vM0p/D3n2C3EReYVX+DNYs5TjZFn/oQt2j+4sVJtSMr18pdRr8wjTcBl6LoV+FUwzPmNA==",
-      "dev": true,
-      "license": "MIT",
-      "dependencies": {
-        "bytes": "^3.1.2",
-        "content-type": "^1.0.5",
-        "debug": "^4.4.3",
-        "http-errors": "^2.0.0",
-        "iconv-lite": "^0.7.0",
-        "on-finished": "^2.4.1",
-        "qs": "^6.14.1",
-        "raw-body": "^3.0.1",
-        "type-is": "^2.0.1"
-      },
-      "engines": {
-        "node": ">=18"
-      },
-      "funding": {
-        "type": "opencollective",
-        "url": "https://opencollective.com/express"
-      }
     },
     "node_modules/bonjour-service": {
       "version": "1.3.0",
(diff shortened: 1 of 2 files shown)


d0sboots:
This does not look like how I would expect versions to be updated. Overrides are for pinning something at a particular version, usually because of some bug or incompatibility. It should not be required just to get newer versions in package.lock.

catloversg:
I'm not particularly against using `overrides`, but I think we should use it only as the last resort. I'm okay with using it to fix vulnerabilities in runtime dependencies, but most vulnerabilities in dev dependencies are "irrelevant", in the sense that they are not really dangerous or exploitable.

Among these changes, there are two that I want to highlight. I deliberately avoided making them in recent dependency update PRs.

With jsdom, you are removing jsdom from the direct dependency list. I don't remember whether we are using jsdom API directly. If we only use `jest-environment-jsdom` (in `FixJSDOMEnvironment.ts`) and do not use jsdom anywhere else, then it's okay.

You are also forcing an update of the jsdom version used by `jest-environment-jsdom`. Jumping 6 major versions (from v20 to v26) is usually very risky. If `jest-environment-jsdom` uses only a small subset of the jsdom API surface, it may continue to work without issues. I will not veto this change, but we should consider the trade-offs carefully. On one hand, we avoid very outdated transitive dependencies and eliminate vulnerability warnings. On the other hand, those vulnerabilities are likely irrelevant in practice, and we introduce the risk of subtle test failures or behavior changes. I'll leave the final decision to d0sboots.

With dompurify, you need to revert it. Monaco *declares* dompurify as its direct dependency, but it does not actually use the version installed in `node_modules`.

Suppose `foo` v1 depends on `bar` v2, normally, code in `foo` v1 would call code of `bar` v2 installed in `node_modules`. However, in monaco, dompurify is bundled inside `esm/vs/base/browser/dompurify`. As a result, overriding the dompurify version in `node_modules` has no effect. In fact, making that change can be considered risky because it creates a false sense of security. It appears that the vulnerable dependency has been updated, but the code that Monaco actually executes remains unchanged.

Ref:
- https://github.com/microsoft/monaco-editor/issues/4906
- https://github.com/microsoft/monaco-editor/issues/5067

Ironically, even in the monaco's repository, they incorrectly updated dompurify in https://github.com/microsoft/monaco-editor/pull/5333 and then immediately reverted it in https://github.com/microsoft/monaco-editor/pull/5347.

With other dependencies, I have not checked the changes.

mctylr-gh (commented):
I have tried to limit the `overrides` usage to cases where either the parent dependency cannot be updated without breaking changes, or when an update to the parent dependency does not exist. 

The "ideal" solution for `jsdom` would be to update Jest and `jest-environment-jsdom` to v30 (Jun 2025) or newer (e.g. 30.4.2 - May 2026)  from v29.7.0 (Sep 2023), but that is a breaking change. 

For jsdom being upgrades from v20.0.3 to v26.1.0, [v21](https://github.com/jsdom/jsdom/releases/tag/v21.0.0) and [v25](https://github.com/jsdom/jsdom/releases/tag/v25.0.0) and [v26](https://github.com/jsdom/jsdom/releases/tag/v26.0.0) has a potential breaking issues, the rests are dependency updates (minimum nodejs version), including a flip-flopping of `nwsapi` to `@asamuzakjp/dom-selector` and back to `nwsapi` again due to performance issues.


jsdom v21.0.0

A potentially-breaking bug fix:

* Fixed the window, document, location, and top properties of Window to be non-configurable. 


jsdom v25.0.0

This major release changes the prototype of a jsdom's EventTarget.prototype to point to the Object.prototype inside the jsdom, instead of pointing to the Node.js Object.prototype. Thus, the prototype chain of Window stays entirely within the jsdom, never crossing over into the Node.js realm.

This only occurs when runScripts is set to non-default values of "dangerously" or "outside-only", as with the default value, there is no separate Object.prototype inside the jsdom.


jsdom v26.0.0

* Breaking change: `canvas` peer dependency requirement has been upgraded from v2 to v3.

The update does not include jsdom [27.0.0](https://github.com/jsdom/jsdom/releases/tag/v27.0.0) or newer (29.1.1) as v27 includes a breaking change of renaming a method used by `jest-environment-jsdom` from `.sendTo()` to `.forwardTo()`, but I don't believe they are any other issues I'm aware of.

I didn't know about the DOMPurify being bundled as an ESM bundle, so I will revert that commit.

mctylr-gh (commented):
Just to be clear; the `jsdom` change passes `npm test` and appear to be functional on my testing via `npm run start:dev`.

mctylr-gh:
My local `npm test` works. And the CI/Test says it failed due to a timeout. 

Is this an issue with GitHub, or something I should be considered with?

d0sboots:
The test timeout failure is a flaky test we haven't run down yet.

d0sboots:
In general, I agree with catlover here, both in terms of the risk of updating versions, the undesirability of using overrides, and our reasons for being significantly more lax about security concerns.

However, when it comes to devDependencies I think we can be a little more aggressive about updating (most) things. These are the dependencies that are (only) used to build and test the app, so if it still builds, tests, and runs (as a smoke test), it *should* be fine. So, if everything still builds and (in particular) tests OK with jsdom removed as a direct dependency, that should be fine (for instance).

This same reasoning is why I don't think it's appropriate to use overrides here. Just update the relevant parent packages. If they have breaking changes that affect us, we'll find out because the tests or build will break. Using overrides has too much risk of leaving us pinned to older versions accidentally, because we won't know/remember why the specific pins were used.

catloversg:
With jsdom, ideally, we should update Jest and related dependencies, but that's not easy. Ref: #2579.

mctylr-gh:
So to silence all the current security `npm audit` warnings without using overrides would require the non-trivial updating both the Jest & et all from v29 to v30, and updating a not yet existing `webpack-dev-server` package and its associated `webpack` / `webpack-cli` packages.

It looks like the dependency updates I want are at least potentially addressed in an open webpack/webpack-dev-server#5652 pull request.

I'm not certain whether it addresses everything but it does contain the worst severity one, the `launch-editor` (command-injection on Windows), and I did confirm that the `sockjs` / `uuid` is not used by us, as it is a non-default configuration option that we don't appear to use.

In light of the pending `web-dev-server` updating, I don't mind reverting the overrides relating to `webpack-dev-server` child dependencies.

But I would like to keep the `jsdom` override, at least for the short term, since I think that update is beneficial to `npm test`'s performance if not behaviour. At least until we can update Jest to v30.

d0sboots:
If you've observed performance improvements in testing, I'll take that as a concrete win.

PR #2857, opened 2026-06-06
API: Let `singularity.destroyW0r1dD43m0n` travel to Bitverse

### LINKED ISSUE #2854: API: Let ns.singularity.destroyW0r1dD43m0n travel to the BitVerse
I recently implemented `ns.singularity.destroyW0r1d_d43m0n` on my scripts, and I noticed that `nextBN` is a mandatory field.

In my opinion, there should be an option to travel to the BitVerse, and let the player choose BN from there. I know there is the `ns.getBitNodeMultipliers()` to do so, but I was thinking to go kinda mid-way.

I know this will look kind of a blunder, but I think it can be added without too much trouble and no api breaks. The implementation is as simple as letting "0" be nextBN value for BitVerse. Like:
```javascript
  ns.singularity.destroyW0r1dD43m0n(0) // see explanation below
}
```
\* the other options (`callbackScript` and `bitNodeOptions`) would be dismissed here, as the game isn't designed to remember the callback script to launch when the player finally decides to choose BN. `bitNodeOptions` can be set through UI.

E9cipher:
Context in and closes #2854 
This PR makes `singularity.destroyW0r1dD43m0n` travel to Bitverse by setting `nextBN` to 0.
MRE:
1.1: **Classic:** Set yourself hacking exp enough to destroy w0r1d_d43m0n, root the server
1.2: **Bladeburner:** Join Bladeburner, complete all blackops.
2. Create a script on home and paste `ns.singularity.destroyW0r1dD43m0n(0)`. Run it.

You should travel to the bitverse. Once there, enter a BN of your choice to ensure that `nextBN` hasn't been set

Edit: Added MRE

### DIFF
--- a/src/NetscriptFunctions/Singularity.ts
+++ b/src/NetscriptFunctions/Singularity.ts
@@ -1155,31 +1155,31 @@ export function NetscriptSingularity(): InternalAPI<ISingularity> {
     },
     destroyW0r1dD43m0n: (ctx) => (_nextBN, _cbScript, _bitNodeOptions) => {
       helpers.checkSingularityAccess(ctx);
-      const nextBN = helpers.number(ctx, "nextBN", _nextBN);
+      const { shouldStopOnBitVerse, nextBN } = _nextBN
+        ? { shouldStopOnBitVerse: false, nextBN: helpers.number(ctx, "nextBN", _nextBN) }
+        : { shouldStopOnBitVerse: true, nextBN: 0 };
+      // Checks if nextBN is undefined and either 2nd or 3rd param have been declared
       if (!nextBN && (_cbScript || _bitNodeOptions)) {
         throw helpers.errorMessage(
           ctx,
           `Invalid call: When nextBN is undefined, all other params should not be declared`,
         );
-      } else {
-        const wd = GetServer(SpecialServers.WorldDaemon);
-        if (!(wd instanceof Server)) {
-          throw new Error("WorldDaemon is not a normal server. This is a bug. Please contact developers.");
-        } else {
-          wd.backdoorInstalled = true;
+      }
+      // Only runs if nextBN !== 0 (jumping to a valid BN)
+      let cbScript;
+      if (shouldStopOnBitVerse) {
+        if (!validBitNodes.includes(nextBN)) {
+          throw new Error(`Invalid BitNode: ${_nextBN}.`);
         }
-        Router.toPage(Page.BitVerse, { flume: false, quick: false });
-      }
-      if (!validBitNodes.includes(nextBN)) {
-        throw new Error(`Invalid BitNode: ${_nextBN}.`);
-      }
-      const cbScript = _cbScript
-        ? resolveScriptFilePath(helpers.string(ctx, "cbScript", _cbScript), ctx.workerScript.name)
-        : false;
-      if (cbScript === null) {
-        throw helpers.errorMessage(ctx, `Could not resolve file path. callbackScript is null.`);
+        cbScript = _cbScript
+          ? resolveScriptFilePath(helpers.string(ctx, "cbScript", _cbScript), ctx.workerScript.name)
+          : false;
+        if (cbScript === null) {
+          throw helpers.errorMessage(ctx, `Could not resolve file path. callbackScript is null.`);
+        }
       }
 
+      // General checks
       const wd = GetServer(SpecialServers.WorldDaemon);
       if (!(wd instanceof Server)) {
         throw new Error("WorldDaemon is not a normal server. This is a bug. Please contact developers.");
@@ -1204,9 +1204,13 @@ export function NetscriptSingularity(): InternalAPI<ISingularity> {
 
       wd.backdoorInstalled = true;
       calculateAchievements();
-      enterBitNode(false, Player.bitNodeN, nextBN, helpers.validateBitNodeOptions(ctx, _bitNodeOptions));
-      if (cbScript) {
-        setTimeout(() => runAfterReset(cbScript), 500);
+      if (shouldStopOnBitVerse) {
+        enterBitNode(false, Player.bitNodeN, nextBN, helpers.validateBitNodeOptions(ctx, _bitNodeOptions));
+        if (cbScript) {
+          setTimeout(() => runAfterReset(cbScript), 500);
+        }
+      } else {
+        Router.toPage(Page.BitVerse, { flume: false, quick: false });
       }
     },
     getCurrentWork: (ctx) => () => {
--- a/src/NetscriptFunctions/Singularity.ts
+++ b/src/NetscriptFunctions/Singularity.ts
@@ -1156,6 +1156,20 @@ export function NetscriptSingularity(): InternalAPI<ISingularity> {
     destroyW0r1dD43m0n: (ctx) => (_nextBN, _cbScript, _bitNodeOptions) => {
       helpers.checkSingularityAccess(ctx);
       const nextBN = helpers.number(ctx, "nextBN", _nextBN);
+      if (!nextBN && (_cbScript || _bitNodeOptions)) {
+        throw helpers.errorMessage(
+          ctx,
+          `Invalid call: When nextBN is undefined, all other params should not be declared`,
+        );
+      } else {
+        const wd = GetServer(SpecialServers.WorldDaemon);
+        if (!(wd instanceof Server)) {
+          throw new Error("WorldDaemon is not a normal server. This is a bug. Please contact developers.");
+        } else {
+          wd.backdoorInstalled = true;
+        }
+        Router.toPage(Page.BitVerse, { flume: false, quick: false });
+      }
       if (!validBitNodes.includes(nextBN)) {
         throw new Error(`Invalid BitNode: ${_nextBN}.`);
       }
--- a/src/NetscriptFunctions/Singularity.ts
+++ b/src/NetscriptFunctions/Singularity.ts
@@ -1155,21 +1155,25 @@ export function NetscriptSingularity(): InternalAPI<ISingularity> {
     },
     destroyW0r1dD43m0n: (ctx) => (_nextBN, _cbScript, _bitNodeOptions) => {
       helpers.checkSingularityAccess(ctx);
-      const { shouldStopOnBitVerse, nextBN } = _nextBN
-        ? { shouldStopOnBitVerse: false, nextBN: helpers.number(ctx, "nextBN", _nextBN) }
-        : { shouldStopOnBitVerse: true, nextBN: 0 };
+      let nextBN;
+      if (_nextBN == null) {
+        nextBN = 0;
+      } else {
+        nextBN = helpers.number(ctx, "nextBN", _nextBN);
+      }
+
       // Checks if nextBN is undefined and either 2nd or 3rd param have been declared
-      if (!nextBN && (_cbScript || _bitNodeOptions)) {
+      if (!_nextBN && (_cbScript || _bitNodeOptions)) {
         throw helpers.errorMessage(
           ctx,
           `Invalid call: When nextBN is undefined, all other params should not be declared`,
         );
       }
-      // Only runs if nextBN !== 0 (jumping to a valid BN)
-      let cbScript;
-      if (shouldStopOnBitVerse) {
+
+      let cbScript: ScriptFilePath | false | null | undefined;
+      if (!(_nextBN == null)) {
         if (!validBitNodes.includes(nextBN)) {
-          throw new Error(`Invalid BitNode: ${_nextBN}.`);
+          throw new Error(`Invalid BitNode: ${nextBN}.`);
         }
         cbScript = _cbScript
           ? resolveScriptFilePath(helpers.string(ctx, "cbScript", _cbScript), ctx.workerScript.name)
@@ -1204,7 +1208,7 @@ export function NetscriptSingularity(): InternalAPI<ISingularity> {
 
       wd.backdoorInstalled = true;
       calculateAchievements();
-      if (shouldStopOnBitVerse) {
+      if (_nextBN) {
         enterBitNode(false, Player.bitNodeN, nextBN, helpers.validateBitNodeOptions(ctx, _bitNodeOptions));
         if (cbScript) {
           setTimeout(() => runAfterReset(cbScript), 500);
--- a/src/NetscriptFunctions/Singularity.ts
+++ b/src/NetscriptFunctions/Singularity.ts
@@ -1156,7 +1156,7 @@ export function NetscriptSingularity(): InternalAPI<ISingularity> {
     destroyW0r1dD43m0n: (ctx) => (_nextBN, _cbScript, _bitNodeOptions) => {
       helpers.checkSingularityAccess(ctx);
       const nextBN = _nextBN != null ? helpers.number(ctx, "nextBN", _nextBN) : null;
-      if (nextBN != null) {
+      if (nextBN !== null) {
         // If _nextBN was provided, check that it is a valid BitNode.
         if (!validBitNodes.includes(nextBN)) {
           throw new Error(`Invalid BitNode: ${_nextBN}.`);
--- a/src/NetscriptFunctions/Singularity.ts
+++ b/src/NetscriptFunctions/Singularity.ts
@@ -1155,35 +1155,23 @@ export function NetscriptSingularity(): InternalAPI<ISingularity> {
     },
     destroyW0r1dD43m0n: (ctx) => (_nextBN, _cbScript, _bitNodeOptions) => {
       helpers.checkSingularityAccess(ctx);
-      let nextBN;
-      if (_nextBN == null) {
-        nextBN = 0;
-      } else {
-        nextBN = helpers.number(ctx, "nextBN", _nextBN);
-      }
-
-      // Checks if nextBN is undefined and either 2nd or 3rd param have been declared
-      if (!_nextBN && (_cbScript || _bitNodeOptions)) {
-        throw helpers.errorMessage(
-          ctx,
-          `Invalid call: When nextBN is undefined, all other params should not be declared`,
-        );
-      }
-
-      let cbScript: ScriptFilePath | false | null | undefined;
-      if (!(_nextBN == null)) {
+      const nextBN = _nextBN != null ? helpers.number(ctx, "nextBN", _nextBN) : null;
+      if (nextBN != null) {
+        // If _nextBN was provided, check that it is a valid BitNode.
         if (!validBitNodes.includes(nextBN)) {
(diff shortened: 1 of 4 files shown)


src/NetscriptFunctions/Singularity.ts
```diff
@@ -1156,7 +1156,10 @@ export function NetscriptSingularity(): InternalAPI<ISingularity> {
     destroyW0r1dD43m0n: (ctx) => (_nextBN, _cbScript, _bitNodeOptions) => {
       helpers.checkSingularityAccess(ctx);
       const nextBN = helpers.number(ctx, "nextBN", _nextBN);
-      if (!validBitNodes.includes(nextBN)) {
+      if (nextBN == 0) {
```
catloversg: This is the wrong way to implement what you want.

This change makes passing 0 become an undocumented behavior, i.e., BN 0 does not exist, but the API now accepts it and behaves unexpectedly. You should make `nextBN` an optional paramater and explicitly check whether the value is undefined/null. Updating TSDoc is also mandatory.

The logic is also incorrect. When destroying the BitNode, you need to either use the logic that follows in this function, which is currently skipped because of the early return, or use the logic in `GeneralDev.tsx`. If you call `Router.toPage` like this, you are only displaying the BitVerse as if the BitNode had been destroyed, but it has not actually been destroyed. This is likely to introduce SF–related bugs.

catloversg (commented):
By the way, I think we should reconsider the old suggestion of adding a Bladeburner API that allows players to complete the BitNode without SF4.

The original suggestion is at https://discord.com/channels/415207508303544321/415213435974975508/1223735679763091578.
I brought it up for discussion again at https://discord.com/channels/415207508303544321/415213435974975508/1248596388128751676.

Quote from Saynt_Garmo:
```
Wuoldnt that be the distinction between singularity and not. BB simply ends the node putting yuo to node selection, vs the utility of the singularity.  That's honestly the only truest point I was trying to make.  One that lets you end the node VIA bb, but that is it. And then the other, with all the singularity goodness.
```

Personally, I think this is a good suggestion. This API allows players to end the BitNode through Bladeburner, but it cannot be used to automate the game further. As a result, its purpose does not overlap with the Singularity APIs. In that sense, it is similar to `installBackdoor`, but for Bladeburner.

E9cipher:
If that is so, then we can simplify this a lot. As I see it, we have two options:

1. Continue with the plan (original purpose). I'll edit the files and apply your suggestions, and we rework the `destroyW0r1dD43m0n` API. We would have to move it out of singularity because of `Saynt_Garmo`'s quote, which in my opinion is totally right.
**Downside:** requires an API method's deprecation, players would have to edit their scripts or we can use auto-migration. Still, doesn't solve the problem for `dev` users. Personally, I don't think this is the right choice.

2. Create a new API in the bladeburner namespace (your suggestion). If that's the case, I'd like to discard this PR and create another one. We keep `installBackdoor` as method for traveling to the BitVerse and don't touch the `destroyW0r1dD43m0n` method, which is good by now.
**Downside:** we now have duplicated logic in `singularity` and `bladeburner` namespaces, but it's not so destructive as option 1.

I agree option 2 is the best choice here, but I'd like to see what other players/maintainers have in mind.

catloversg:
> I'll edit the files and apply your suggestions, and we rework the destroyW0r1dD43m0n API. We would have to move it out of singularity because of Saynt_Garmo's quote, which in my opinion is totally right.

I think you misunderstand something here. Why do you move `destroyW0r1dD43m0n` out of singularity? Why is it related to what Saynt_Garmo said?

> requires an API method's deprecation, players would have to edit their scripts or we can use auto-migration. Still, doesn't solve the problem for dev users. Personally, I don't think this is the right choice.

What do you mean? When the first param becomes optional, it's technically a behavior change, but it does not affect any reasonable code. Currently, `destroyW0r1dD43m0n` throws an error when `nextBN` is undefined, which means any code doing that is effectively nonsensical.

> we now have duplicated logic in singularity and bladeburner namespaces, but it's not so destructive as option 1.

The new API should not share too much logic with the current `destroyW0r1dD43m0n`. If it does, it's likely implemented incorrectly.

E9cipher:
Sorry for the ambiguity :)
Replying to your questions in order:
1. I move it out of the singularity namepace because if player doesn't own SF-4 then it's nonsense they can access `ns.singularity`namespace, unless there is actually some method or property that doesn't require SF-4.
2. Related to my 1st point, if we move it out of singularity then all scripts currently using the `destroyW0r1dD43m0n` method will not work.
3. I agree here that it shouldn't share too much logic. But the only thing that changes is the "trigger", the condition in which the player can destroy WD. The "move BN logic" is the same

catloversg:
I'm even more confused now. What is "it" that you want to move out of the singularity namespace? `destroyW0r1dD43m0n` API? That makes no sense to me. Do you want to discuss on Discord?

E9cipher:
Sure. I can create an account and I think I got the link to the server. What do I do, shall I post a msg like linking this PR? Give me 10min to do the processing

catloversg:
Just start chatting on the suggestion channel. You can post the link to this PR as context.

E9cipher:
Link to discord (for context): [https://discord.com/channels/415207508303544321/415213435974975508](https://discord.com/channels/415207508303544321/415213435974975508).
Also, I'll do a reset so I can start fresh on the `dev` branch. @catloversg that okay?
EDIT: doesn't let me revert, so I'll just reset it. Commit history will be lost, but the changes are not important. I made a backup branch just in case

E9cipher:
Sorry, don't know why it closed :( just made a commit and I can reopen this.

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -2887,11 +2887,11 @@ export interface Singularity {
    *   OR
    * Completed the final black op.
    *
-   * @param nextBN - BN number to jump to
+   * @param nextBN - BN number to jump to. undefined leads you to the BitVerse
```
```suggestion
   * @param nextBN - BN number to jump to. Undefined leaves you on the BitVerse selection screen.
```

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -2887,11 +2887,11 @@ export interface Singularity {
    *   OR
    * Completed the final black op.
    *
```
d0sboots, line 2889: I'm on my phone, so double check the syntax here.
```suggestion
   *
   * Although this function can be used to take you to the BitVerse screen, that can usually be accomplished more cheaply with @link{#installBackdoor}.
```

src/NetscriptFunctions/Singularity.ts
```diff
@@ -1156,6 +1156,20 @@ export function NetscriptSingularity(): InternalAPI<ISingularity> {
...
+        throw helpers.errorMessage(
+          ctx,
+          `Invalid call: When nextBN is undefined, all other params should not be declared`,
+        );
+      } else {
+        const wd = GetServer(SpecialServers.WorldDaemon);
+        if (!(wd instanceof Server)) {
+          throw new Error("WorldDaemon is not a normal server. This is a bug. Please contact developers.");
+        } else {
+          wd.backdoorInstalled = true;
```
d0sboots: This skips all the important eligibility checks below and is generally both duplicated code and in the wrong place.

src/NetscriptFunctions/Singularity.ts
```diff
@@ -1156,6 +1156,20 @@ export function NetscriptSingularity(): InternalAPI<ISingularity> {
     destroyW0r1dD43m0n: (ctx) => (_nextBN, _cbScript, _bitNodeOptions) => {
       helpers.checkSingularityAccess(ctx);
       const nextBN = helpers.number(ctx, "nextBN", _nextBN);
+      if (!nextBN && (_cbScript || _bitNodeOptions)) {
```
d0sboots: General structure: nextBN should probably be undefined if _nextBN is. If that makes things too ugly for ts, you can use 0 instead. This check makes sense here, but most of the rest you've added doesn't. You'll need to conditionally skip the valid bn check. The Router.toPage should conditionally happen right by where enterBitnode happens.

src/ScriptEditor/NetscriptDefinitions.d.ts
E9cipher: I would rather leave this blank intentionally for players to discover it on their own. If we give them the cheap API  (`installBackdoor`) directly we don't encourage them to dig into the docs. What do you think?

E9cipher:
This is the logic I have added. I think the logic is good, but somehow I get an error when `destroyW0r1dD43m0n()` has no Paramus. I need more time to track the error down, cause I seem to fail locating it. It says it's ~20 lines above where it should be. If you somehow spot it, please point it out to me.

Haven't generated docs nor reviewed lint, so tests are all failing. I'll finish this tomorrow for sure.

@catloversg are you okay with typing the jest tests?

src/ScriptEditor/NetscriptDefinitions.d.ts
d0sboots: What do you think digging into the docs is? If they are seeing this, they are already digging into the docs.

Although, I also disagree with the idea in general: The docs exist to make it easier to find (and use) what you're looking for. If there's something we want players to discover on their own, we explicitly document that it's undocumented.

src/NetscriptFunctions/Singularity.ts
```diff
@@ -1155,31 +1155,31 @@ export function NetscriptSingularity(): InternalAPI<ISingularity> {
     },
     destroyW0r1dD43m0n: (ctx) => (_nextBN, _cbScript, _bitNodeOptions) => {
       helpers.checkSingularityAccess(ctx);
-      const nextBN = helpers.number(ctx, "nextBN", _nextBN);
+      const { shouldStopOnBitVerse, nextBN } = _nextBN
```
d0sboots: You don't need a separate bool for this; after this point you can just check if _nextBN == null.

d0sboots: Test needs to be against `_nextBN == null`. Note the double equals, which is rare - almost always you want triple equals. This specific construct checks for null or undefined *only*. What you have now will also catch falsy things like 0 and "", which is no good.

src/NetscriptFunctions/Singularity.ts
```diff
@@ -1190,9 +1204,13 @@ export function NetscriptSingularity(): InternalAPI<ISingularity> {
 
       wd.backdoorInstalled = true;
       calculateAchievements();
-      enterBitNode(false, Player.bitNodeN, nextBN, helpers.validateBitNodeOptions(ctx, _bitNodeOptions));
-      if (cbScript) {
-        setTimeout(() => runAfterReset(cbScript), 500);
+      if (shouldStopOnBitVerse) {
```
d0sboots: Optional: you can reduce the size of the diff and the amount of indenting by doing the toPage branch first with an early return, and then the other branch can be left alone.

src/NetscriptFunctions/Singularity.ts
```diff
@@ -1155,17 +1155,31 @@ export function NetscriptSingularity(): InternalAPI<ISingularity> {
...
+          ctx,
+          `Invalid call: When nextBN is undefined, all other params should not be declared`,
+        );
       }
-      const cbScript = _cbScript
-        ? resolveScriptFilePath(helpers.string(ctx, "cbScript", _cbScript), ctx.workerScript.name)
-        : false;
-      if (cbScript === null) {
-        throw helpers.errorMessage(ctx, `Could not resolve file path. callbackScript is null.`);
+      // Only runs if nextBN !== 0 (jumping to a valid BN)
```
d0sboots: Remove this comment; it is just stating what the if statement does, and so it's redundant.

d0sboots (commented):
I see why the tests are failing, but IMO figuring out why yourself will help you. I will give you a hint: taking my suggestions will help fix it.

src/NetscriptFunctions/Singularity.ts
E9cipher: I'd rather leave it this way, if I may. It's just because both checks that go to BitVerse will have the same condition, `_nextBN == null`.

E9cipher (commented):
The logic is done. I've applied the suggestions (except the diff one from @d0sboots , I think the code is okay like this). Let me know if you have any other feedback.

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -2887,11 +2887,14 @@ export interface Singularity {
...
-   * @param nextBN - BN number to jump to
+   * Although this function can be used to take you to the BitVerse screen, that can usually be accomplished
+   * more cheaply with @link{#installBackdoor}.
```
d0sboots: Looks like the syntax was wrong before, and this is what it should be. I warned you that I was on my phone and you should double-check the syntax! XD
```suggestion
   * more cheaply with {@link Singularity.installBackdoor | installBackdoor}.
```

src/NetscriptFunctions/Singularity.ts
```diff
@@ -1155,17 +1155,35 @@ export function NetscriptSingularity(): InternalAPI<ISingularity> {
...
+        nextBN = helpers.number(ctx, "nextBN", _nextBN);
       }
-      const cbScript = _cbScript
-        ? resolveScriptFilePath(helpers.string(ctx, "cbScript", _cbScript), ctx.workerScript.name)
-        : false;
-      if (cbScript === null) {
-        throw helpers.errorMessage(ctx, `Could not resolve file path. callbackScript is null.`);
+
+      // Checks if nextBN is undefined and either 2nd or 3rd param have been declared
+      if (!_nextBN && (_cbScript || _bitNodeOptions)) {
```
d0sboots: This needs to actually check for undefined, otherwise, it'll trigger if you pass 0.
```suggestion
      if (_nextBN == null && (_cbScript || _bitNodeOptions)) {
```

src/NetscriptFunctions/Singularity.ts
```diff
@@ -1155,17 +1155,35 @@ export function NetscriptSingularity(): InternalAPI<ISingularity> {
...
+        throw helpers.errorMessage(
+          ctx,
+          `Invalid call: When nextBN is undefined, all other params should not be declared`,
+        );
+      }
+
+      let cbScript: ScriptFilePath | false | null | undefined;
+      if (!(_nextBN == null)) {
```
d0sboots:
```suggestion
      if (_nextBN != null) {
```

src/NetscriptFunctions/Singularity.ts
```diff
@@ -1155,17 +1155,35 @@ export function NetscriptSingularity(): InternalAPI<ISingularity> {
...
+
+      // Checks if nextBN is undefined and either 2nd or 3rd param have been declared
+      if (!_nextBN && (_cbScript || _bitNodeOptions)) {
+        throw helpers.errorMessage(
+          ctx,
+          `Invalid call: When nextBN is undefined, all other params should not be declared`,
+        );
+      }
+
+      let cbScript: ScriptFilePath | false | null | undefined;
```
d0sboots: We can cut down on the polymorphism here by a lot with some rearrangement below.
```suggestion
      let cbScript: ScriptFilePath | null = null;
```

src/NetscriptFunctions/Singularity.ts
```diff
@@ -1155,17 +1155,35 @@ export function NetscriptSingularity(): InternalAPI<ISingularity> {
...
+      if (!(_nextBN == null)) {
+        if (!validBitNodes.includes(nextBN)) {
+          throw new Error(`Invalid BitNode: ${nextBN}.`);
+        }
+        cbScript = _cbScript
+          ? resolveScriptFilePath(helpers.string(ctx, "cbScript", _cbScript), ctx.workerScript.name)
+          : false;
+        if (cbScript === null) {
+          throw helpers.errorMessage(ctx, `Could not resolve file path. callbackScript is null.`);
+        }
```
d0sboots:
```suggestion
        if (_cbScript != null) {
          cbScript = resolveScriptFilePath(helpers.string(ctx, "cbScript", _cbScript), ctx.workerScript.name);
          if (cbScript === null) {
            throw helpers.errorMessage(ctx, `Could not resolve file path. callbackScript is null.`);
          }
        }
```

src/NetscriptFunctions/Singularity.ts
```diff
@@ -1190,9 +1208,13 @@ export function NetscriptSingularity(): InternalAPI<ISingularity> {
 
       wd.backdoorInstalled = true;
       calculateAchievements();
-      enterBitNode(false, Player.bitNodeN, nextBN, helpers.validateBitNodeOptions(ctx, _bitNodeOptions));
-      if (cbScript) {
-        setTimeout(() => runAfterReset(cbScript), 500);
+      if (_nextBN) {
```
d0sboots: I wouldn't mention this normally, but since there are other comments:
I'd either do `if (nextBN) {` or `if (_nextBN != null) {`. By this point, those are both equivalent to what you wrote, but it's not entirely obvious that is the case, whereas the first two expressions are much more clearly correct.

d0sboots (commented):
I think the core logic is correct, what's left is some style and doc things.

src/ScriptEditor/NetscriptDefinitions.d.ts
E9cipher: Yeah, output seemed a bit suspicious but believe it or not, that was the only part I didn't doubt of! I'll change it later, it's just a minor docs thing.

catloversg:
Since this PR has already been reviewed several times and there are only a few minor issues left, I'll take care of the remaining work so we can move it forward.

E9cipher:
Yup, it's all yours. By the way, I have a last question: we added this for the bladeburner route, right? Then players who don't have SF-4 will be able to access  this API?

catloversg:
> we added this for the bladeburner route, right?

For posterity (so future contributors won't have to dig up old discussions on Discord), I'll quote my message regarding this change and the related bitflume suggestion.
```
Regarding bitflume, I don't think we need to do that. The change in destroyW0r1dD43m0n is already for a really niche feature, which is programmatically end the BN via the Bladeburner route while not jumping to the next BN.
To be honest, this looks too niche to me, but anyway, I won't veto it.
With bitflume, it's even more niche. When a player uses it, they want to programmatically abandon the current BN and wait until they manually choose another BN. I bet people will come up with some justification for that, but I'm sure I won't buy it.

My opinion: Changing destroyW0r1dD43m0n is not necessary, but I'll allow it because it provides an alternative to installBackdoor for the Bladeburner route. With bitflume, the player needs to bitflume manually or use terminal injection.
```

> Then players who don't have SF-4 will be able to access this API?

Players who don't have SF4 will *not* be able to access this API. This is the current rule, and this PR does not change it. No SF4 = No automation.

In other words, if they want to use the Bladeburner route and automate the ending of the BN, they still need SF4 to use `destroyW0r1dD43m0n`.

PR #2860, opened 2026-06-07
BUGFIX: Remove duplicate running script code from memory, fix missed module cache lookups

Sphyxis:
This PR aims to fix a few issues.

-Duplicate script content is present in memory when multiple scripts with the same content are running.
-Scripts can miss their module cache lookup due to their blob URL being unique
-Writing scripts could invalidate them and rewrite, even if the new content is identical

Lets start with the `newCode` variable.  It is created using `importedScript.mod.url`.  That string is unique, which can cause a cache lookup to fail.  To fix this, we add the `cacheKey` variable.  This represents a stable key for cache lookup.  This key can be used to lookup modules for multiple URL's.
In order to accommodate this new cacheKey variable, I've added it directly to LoadedModule.  We now set/get from the moduleCache using this cacheKey directly.

Next up is the scriptCodeCache.  I've added a get/set code function to Scripts.ts to retrieve the code.  The `refs` field is how many times it has been referenced.  Now, when you start a script, it's referenced code block gets cached, and each subsequent duplicate script (based on content) that is started adds to that count.  When a script is done, it is released which decrements the counter.  When the counter is at 0 the script code is released.  This should be a far cleaner life cycle path, which ends with the code being released, than relying on WeakRefs and GC.

The other minor fixes are mostly early returns, which show up as unneeded churn in the system.  If the contents have not changed when you write to it, don't invalidate and rewrite.

Pattern:

Pattern 1:  Same source code
Before
- New Script object / new string / new blob URL
- Cache sees a different identity
- New module or duplicated source
- Memory increases
After
- We have a stable source/cache identity key
- We reuse existing string/modules when possible by using this key
- We avoid invalidating anything when content has not changed
- Memory plateaus

We track 2 identities now.  Runtime still uses the blob URLs, but caches use the stable cacheKey

Patter 2: Each running script kept a copy of it's own source code in memory
Before
- Each running duplicate script would store it's own .code value
After
- Each running duplicate script stores 1 instance of it's code, and once the last script is done it is freed and can be GCd

The pattern for this becomes:
- script with source A runs
- Cache miss
- Store A, refs = 1
- script._code = A
- second script with source A runs
- cache hit
- refs = 2
- script._code = cached A

Before we could have 1000 Script objects with 1000 retained source strings.  Now, we can have 1000 Script objects with 1 retained source string and 1000 refs

Just added as a draft incase I think of something I've missed by the morning.

MRE:
```js
const SIZE = .5 //In MB
const COPIES = 32
const PREFIX = "mre-source-bloat/"
/**
 * Backup your save.  These tests can cause your game to fail to save or corrupt it
 * While I'm looking at fixing modules and caching code it still creates new Script instances
 * 
 * Key: Delete = 0, Unique = 1, Start = 2
 * 
 * General control:  0 = false, 1 = false, 2 = false
 * - Dev: memory will plateau
 * - Fix: same
 * 
 * Source/file object recreation:  0 = true, 1 = false, 2 = false
 * - Dev: Some churning, and memory should plateau at a higher level as each script uses it's own code
 * - Fix: Same, but less churning.  Memory should plateau earlier due to code sharing
 * - Increasing the file size or count should increase the gap between dev and this fix
 * 
 * Module Leak (Main test):  0 = true, 1 = false, 2 = true
 * - Dev: Memory should grow due to new modules being created from missed cache lookups
 * - Fix: Memory should plateau as module cache is now based on source/cacheKey
 * 
 * Save/file bloat: 0 = false, 1 = true, 2 = true
 * - Dev: Grows fast as new script code/modules accumulate
 * - Fix: Still grows from file/save data, but live source-string duplication should be lowered
 * - I ran this for both Dev and Fix and analyzed the memory dumps.  The results were great
 * - The fix had a lower dump size, ran longer, generated more unique names by over 4x, and a total of 1 Source-Code shared.
 *   The dev branch had 16k unique file names and 1,024 Source-Code shared
 * - The main bloat from this is the save data accumulation from all the files.
 */
const DELETE_BEFORE_SCP = false
const UNIQUE_FILE_NAMES = false
const START_SCRIPTS = false
/** @param {NS} ns */
  ns.disableLog("ALL")
  ns.ui.openTail()
  const files = Array.from({ length: COPIES }, (_, i) => `copy-${i}.js`)
  const targets = scanAll(ns)
    .filter((server) => ns.hasRootAccess(server))
    .filter((server) => ns.getServerMaxRam(server) >= 2)
  ns.atExit(() => {
    for (const target of targets)
      for (const file of ns.ls(target, PREFIX))
        ns.rm(file, target)
  })

  const victimBody = await makeLargeVictimScript(SIZE)
  let lastYield = performance.now()
  let generation = 0

  while (true) {
    generation++ //We can see this in the console
    const gen_prefix = UNIQUE_FILE_NAMES ? ns.sprintf(`-Gen-${generation}-`) : ""
    for (const file of files) {
      const fileName = `${PREFIX}${gen_prefix}${file}`
      ns.write(fileName, victimBody, "w")
      if (performance.now() - lastYield >= 200) {
        lastYield = performance.now()
        await ns.sleep(0)
      }
    }
    if (START_SCRIPTS) {
      //Give started scripts some time to finish.
      await ns.sleep(0)
    }
    for (const target of targets) {
      for (const file of files) {
        const fileName = `${PREFIX}${gen_prefix}${file}`
        if (DELETE_BEFORE_SCP) ns.rm(fileName, target)
        ns.scp(fileName, target, "home")
        if (START_SCRIPTS) {
          const threads = Math.floor(Math.max(0, ns.getServerMaxRam(target) - ns.getServerUsedRam(target) - target === "home" ? 32 : 0) / 1.6)
          if (threads) ns.exec(fileName, target, { temporary: true })
        }
        if (performance.now() - lastYield >= 200) {
          lastYield = performance.now()
          await ns.sleep(0)
        }
      }
    }
  }
}

function scanAll(ns) {
  const serverList = new Set(["home"])
  for (const server of serverList) {
    for (const connection of ns.scan(server)) {
      serverList.add(connection)
    }
  }
  return Array.from(serverList)
}

async function makeLargeVictimScript(sizeMb) {
  const targetBytes = sizeMb * 1024 * 1024;
  const line = "SOURCE_BLOAT_MRE_PAD_0123456789abcdef".repeat(32) + "\n"
  let pad = "";
  let lastYield = performance.now()

  while (pad.length < targetBytes) {
    pad += line;
    if (pad.length > 1024 * 1024 && pad.length * 2 < targetBytes) pad += pad
    if (performance.now() - lastYield >= 200) {
      lastYield = performance.now()
      await ns.sleep(0)
    }
  }

  pad = pad.slice(0, targetBytes)

  return `/*
${pad}
*/
  ns.disableLog("ALL");
  await ns.sleep(50)
}`
    ;
}
```

### DIFF
--- a/src/Netscript/killWorkerScript.ts
+++ b/src/Netscript/killWorkerScript.ts
@@ -119,4 +119,5 @@ function removeWorkerScript(workerScript: WorkerScript): void {
   if (rs.temporary === false) {
     AddRecentScript(workerScript);
   }
+  workerScript.code = "";
 }
--- a/src/NetscriptJSEvaluator.ts
+++ b/src/NetscriptJSEvaluator.ts
@@ -160,7 +160,11 @@ function generateLoadedModule(script: Script, scripts: Map<ScriptFilePath, Scrip
   // preventing the ranges for other imports from being shifted.
   importNodes.sort((a, b) => b.start - a.start);
   let newCode = scriptCode;
-  // Loop through each node and replace the script name with a blob url.
+  let cacheKey = scriptCode;
+
+  // Runtime code needs blob URLs, but the module cache key does not.
+  // Blob URLs are unique per creation, so using them as the cache key makes
+  // identical imports miss the cache and create permanent duplicate modules.
   for (const node of importNodes) {
     const importedScript = getModuleScript(node.filename, script.filename, scripts);
     for (const scriptInSeenStack of seenStack) {
@@ -175,9 +179,10 @@ function generateLoadedModule(script: Script, scripts: Map<ScriptFilePath, Scrip
     importedScript.mod = generateLoadedModule(importedScript, scripts, seenStack);
     seenStack.pop();
     newCode = newCode.substring(0, node.start) + importedScript.mod.url + newCode.substring(node.end);
+    cacheKey = cacheKey.substring(0, node.start) + importedScript.mod.cacheKey + cacheKey.substring(node.end);
   }
 
-  const cachedMod = moduleCache.get(newCode);
+  const cachedMod = moduleCache.get(cacheKey);
   if (cachedMod) {
     script.mod = cachedMod;
   } else {
@@ -213,8 +218,8 @@ function generateLoadedModule(script: Script, scripts: Map<ScriptFilePath, Scrip
     // directly return the module, without even attempting to fetch, due to the way
     // modules work.
     URL.revokeObjectURL(url);
-    script.mod = new LoadedModule(url, module);
-    moduleCache.set(newCode, script.mod);
+    script.mod = new LoadedModule(url, module, cacheKey);
+    moduleCache.set(cacheKey, script.mod);
   }
 
   addDependencyInfo(script, seenStack);
--- a/src/Script/Script.ts
+++ b/src/Script/Script.ts
@@ -7,10 +7,36 @@ import { RamCostConstants } from "../Netscript/RamCostGenerator";
 import type { ScriptFilePath } from "../Paths/ScriptFilePath";
 import { ContentFile } from "../Paths/ContentFile";
 
+//This caches script code for scripts with the same code but a different filename
+interface ScriptCodeCacheEntry {
+  code: string;
+  refs: number;
+}
+
+const scriptCodeCache = new Map<string, ScriptCodeCacheEntry>();
+
+function retainScriptCode(code: string): string {
+  const entry = scriptCodeCache.get(code);
+  if (entry) {
+    entry.refs++;
+    return entry.code;
+  }
+  scriptCodeCache.set(code, { code, refs: 1 });
+  return code;
+}
+
+function releaseScriptCode(code: string): void {
+  const entry = scriptCodeCache.get(code);
+  if (!entry) return;
+  entry.refs--;
+  if (entry.refs <= 0) scriptCodeCache.delete(code);
+}
+
 /** A script file as a file on a server.
  * For the execution of a script, see RunningScript and WorkerScript */
 export class Script extends ContentFile {
-  code: string;
+  private _code = "";
+  private codeRefRetained = false;
   filename: ScriptFilePath;
   server: string;
 
@@ -34,18 +60,31 @@ export class Script extends ContentFile {
     this.metadata.read();
     return this.code;
   }
+  get code(): string {
+    return this._code;
+  }
   set content(newCode: string) {
-    this.metadata.edit();
     if (this.code === newCode) return;
+    this.metadata.edit();
     this.code = newCode;
     this.invalidateModule();
   }
-
+  set code(newCode: string) {
+    if (this.codeRefRetained && this._code === newCode) return;
+    if (this.codeRefRetained) releaseScriptCode(this._code);
+    this._code = retainScriptCode(newCode);
+    this.codeRefRetained = true;
+  }
   constructor(filename = "default.js" as ScriptFilePath, code = "", server = "") {
     super();
     this.filename = filename;
     this.code = code;
     this.server = server; // hostname of server this script is on
+  }
+  dispose(): void {
+    if (this.codeRefRetained) releaseScriptCode(this._code);
+    this._code = "";
+    this.codeRefRetained = false;
   }
 
   /** Invalidates the current script module and related data, e.g. when modifying the file. */
@@ -93,6 +132,7 @@ export class Script extends ContentFile {
     if (this.server !== server.hostname || server.isRunning(this.filename)) return false;
     this.invalidateModule();
     server.scripts.delete(this.filename);
+    this.dispose();
     return true;
   }
 
--- a/src/Script/Script.ts
+++ b/src/Script/Script.ts
@@ -64,8 +64,8 @@ export class Script extends ContentFile {
     return this._code;
   }
   set content(newCode: string) {
-    if (this.code === newCode) return;
     this.metadata.edit();
+    if (this.code === newCode) return;
     this.code = newCode;
     this.invalidateModule();
   }
--- a/src/Server/BaseServer.ts
+++ b/src/Server/BaseServer.ts
@@ -190,6 +190,7 @@ export abstract class BaseServer implements IServer {
       if (this.isRunning(path)) return { res: false, msg: "Cannot delete a script that is currently running!" };
       script.invalidateModule();
       this.scripts.delete(path);
+      script.dispose();
       return { res: true };
     }
     if (hasProgramExtension(path)) {
@@ -256,6 +257,7 @@ export abstract class BaseServer implements IServer {
     // Check if the script already exists, and overwrite it if it does
     const script = this.scripts.get(filename);
     if (script) {
+      if (script.code === code) return { overwritten: false };
       // content setter handles module invalidation
       script.content = code;
       return { overwritten: true };
@@ -274,6 +276,7 @@ export abstract class BaseServer implements IServer {
     const existingFile = this.textFiles.get(textPath);
     // overWrite if already exists
     if (existingFile) {
+      if (existingFile.text === txt) return { overwritten: false };
       existingFile.content = txt;
       return { overwritten: true };
     }
--- a/src/Server/BaseServer.ts
+++ b/src/Server/BaseServer.ts
@@ -276,7 +276,7 @@ export abstract class BaseServer implements IServer {
     const existingFile = this.textFiles.get(textPath);
     // overWrite if already exists
     if (existingFile) {
-      if (existingFile.text === txt) return { overwritten: false };
+      if (existingFile.text === txt) return { overwritten: true };
       existingFile.content = txt;
       return { overwritten: true };
     }
--- a/src/Script/LoadedModule.ts
+++ b/src/Script/LoadedModule.ts
@@ -13,9 +13,11 @@ export interface ScriptModule {
 export class LoadedModule {
   url: ScriptURL;
   module: Promise<ScriptModule>;
+  cacheKey: string;
 
-  constructor(url: ScriptURL, module: Promise<ScriptModule>) {
+  constructor(url: ScriptURL, module: Promise<ScriptModule>, cacheKey: string) {
     this.url = url;
     this.module = module;
+    this.cacheKey = cacheKey;
   }
 }
--- a/src/TextFile.ts
+++ b/src/TextFile.ts
@@ -17,6 +17,7 @@ export class TextFile extends ContentFile {
     return this.text;
   }
   set content(text: string) {
+    if (this.text === text) return;
     this.metadata.edit();
     this.text = text;
   }


Sphyxis:
After a night long test with my darknet script, all of the issues that this PR aims to fix have been fixed.  Another memory leak was found though, but that will be fixed in another PR.

src/Script/Script.ts
```diff
@@ -34,19 +60,32 @@ export class Script extends ContentFile {
     this.metadata.read();
     return this.code;
   }
+  get code(): string {
+    return this._code;
+  }
   set content(newCode: string) {
-    this.metadata.edit();
     if (this.code === newCode) return;
+    this.metadata.edit();
```
d0sboots, line 67: This changes semantics. Even if we don't invalidate anything as a performance optimization, the write still occurred, and the metadata should reflect that.

src/Script/Script.ts
```diff
@@ -7,10 +7,36 @@ import { RamCostConstants } from "../Netscript/RamCostGenerator";
 import type { ScriptFilePath } from "../Paths/ScriptFilePath";
 import { ContentFile } from "../Paths/ContentFile";
 
+//This caches script code for scripts with the same code but a different filename
+interface ScriptCodeCacheEntry {
+  code: string;
+  refs: number;
+}
+
+const scriptCodeCache = new Map<string, ScriptCodeCacheEntry>();
```
d0sboots, line 16: This entire system seems overcomplicated to me; you are just redoing the work that the GC already does. 

A string will be naturally be released once it has no references; we don't need to explicitly track them. If the code we are tracking isn't equal, then this map won't accomplish anything, since it will end up with different map entries. If the code is all the same string instance, this map won't accomplish anything because all the memory will already be shared, and GC will already handle releasing it when the last reference is removed. (In practice, this will never happen because both the module map and the module itself keep a reference to the code.)

So the only way this accomplishes something is if we have multiple copies of the code that are equal, but different instances. Note that you cannot observe this in JavaScript - `===` and `Object.is` will return true in this case. It's strictly an implementation detail that you can only peer at with the memory inspector. Your writeup in the PR didn't explain to me if this is what's happening, and if so, what is causing it. But if it is, we should aim to fix it directly there.

src/Server/BaseServer.ts
```diff
@@ -274,6 +276,7 @@ export abstract class BaseServer implements IServer {
     const existingFile = this.textFiles.get(textPath);
     // overWrite if already exists
     if (existingFile) {
+      if (existingFile.text === txt) return { overwritten: false };
```
d0sboots: These are also breaking changes. You're changing the semantics of what "overwritten" means.

src/NetscriptJSEvaluator.ts
```diff
@@ -160,7 +160,11 @@ function generateLoadedModule(script: Script, scripts: Map<ScriptFilePath, Scrip
   // preventing the ranges for other imports from being shifted.
   importNodes.sort((a, b) => b.start - a.start);
   let newCode = scriptCode;
-  // Loop through each node and replace the script name with a blob url.
+  let cacheKey = scriptCode;
+
+  // Runtime code needs blob URLs, but the module cache key does not.
```
d0sboots, line 165: This is both unnecessary and invalid.

Yes, blob urls are unique on each creation. However, the entire purpose of the module cache (especially now that it uses full refs and not WeakRefs) is to avoid re-compiling modules and thus re-creating blob urls. If this function were to be called twice with the same code, it would get the same blob urls back both times, because the first time would compile the modules it needs into imports (thus creating new blob urls), and the second time those would exist as LoadedModules in the cache, and thus the same urls would be returned. This argument applies recursively: The result causes the generated code to be the same, and thus causing the 2nd time to pick up the LoadedModule that the first call created and put in the cache.

If there is something broken in that argument, then module loading has been *severely broken* for a long time. I have not seen any evidence of this.

The reason this change is invalid is because it *breaks* the invariants that the above argument relies upon. Instead of the key containing the blob urls, it would contain only the original filenames. This means that if `a.js` imports `b.js`, and `b.js` changes but `a.js` does not, the cacheKey for `a.js` would remain the same, and thus you would pick up the *old* code when importing `a.js`.

src/Script/Script.ts
Sphyxis: I can remove it in this and other places.

src/Script/Script.ts
Sphyxis: It is exactly for your second thought, multiple instances all the same copy of code, that this is for.  Each instance has it's own copy of script code.  This is a big issue for things like darknet scripts, where you can have 80+ of each controller running.  Or even hack/grow/weaken from a batcher.  With this change in place, I was able to bring my batches up to 300k Batches at once.  It brought me to 3.6gb of memory for the renderer.  It serves as an anchor for those scripts source code, making it shared across all instances.  When there are no more instances, it is released to GC.  This can have massive memory savings based on what you are doing.

src/Server/BaseServer.ts
Sphyxis: I will undo this.

src/NetscriptJSEvaluator.ts
Sphyxis: It has been broken for a long time.  Let me try to explain as best I can.

This fix does not contain only the original filenames.  It does not make the `cacheKey` something like `a.js` source + "./b.js",  It creates a stable cacheKey, based on code and imports.  We do not create stale states.
Example:
-`a.js` imports `b.js`
-`a.js` is compiled, `generateLoadedModule(a)` walks through the imports.  A dependency relationship is created then
-Once `b.js` gets a module, it does `addDependencyInfo()` that records that `a.js` depends on `b.js`
-At this point, internally `b.js` dependents contains `a.js`
-When you change a script, like `b.js`, it's `invalidateModule` runs
-Dependencies are cleared
-Because `b.js` had dependents, their `invalidateModule` is also called
-Now, both `a.js` and `b.js` have a null module

Now, the next time `a.js` is called,it cannot reuse `a.js`' original module.
-The first thing compile starts with is if the script.mod field is populated, it returns `script.mod.module`.  But if it's null, it calls `generateLoadedModule`
-This is a new lookup for `a.js`.  While rebuilding `a.js`, `b.js` gets rebuilt first as a dependency.
-Since `b.js`.mod is also null now, it also runs `generateLoadedModule`.
-This will produce a new `cacheKey` for `b.js`
-`a.js`' `cacheKey` includes `b.js`' `cacheKey`
-Since `cacheKey` is made up of `cacheKeys`, this will mean that the `cacheKey` for `a.js`' lookup is now different
--It is not simply: `a.js` source code with "./b.js", it becomes `a.js` source with `b.js`' current cacheKey
This causes the new `a.js` `cacheKey` lookup to be different.  So, if there is no cached entry for `a.js` as it is now where it's imported `b.js` as it is now, it will create a new `cacheKey` for this specific mutation and then cache it.

As soon as you invalidate a scripts module, it can never find it again.  That's because the old `cacheKey` was made using the `BLOB.url`.  That is unique, and when you recreate the `cacheKey` lookup for that script it changes.
This change also makes the cache file name agnostic.

Real examples that helped me find this:
-My darknet script, when run overnight with the previous WeakRef fix, showed over 7.5k identical module entries for all the files I write to a server.  With this change, it only found 1.
-Writing a helper script to a fresh darknet server does not create a module for it.  That happens when it is run for the first time.  Since it's lookup is based on it's BLOB.url, it never finds a cache and creates a new entry.  This can be easily replicated by writing a file from memory, running it, deleting it and then re-writing it.  You can also test unique filename with identical content in the same way.

I hope this explains it.

src/Script/Script.ts
Sphyxis: Newest commit returns the edit to the top.

src/Server/BaseServer.ts
Sphyxis: Newest commit returns overwritten: true early if there is nothing to change/invalidate.

Sphyxis:
I have been able to push my batcher up to 300k batches for a HWGW composition - and it's stable.  That's a 3x increase from baseline and it could likely go higher.

src/Script/Script.ts
d0sboots: OK, but you didn't address my key point, which is *why* is this happening? If you assign a string to another place, it does not create a new instance, and that is what should be happening with scp and such.

It is very *very* unlikely that I will merge this part, specifically. Manual reference-counting is extremely error-prone (even if you get it correct now, it is very easy for a change later to break it). JS makes this worse by not having destructors or any good way to verify correctness of reference counting schemes. That is why, if we *were* going to do this, it would best be done with `Symbol.for` and `Symbol.keyFor` - that provides a way to create globally unique objects that are mapped to strings, so you can guarantee uniqueness. The downside is that they can't be GC'd at all, but the modules can't be GC'd and they pin the code itself so that's not a big loss.

However, IMO that is a last resort. The better solution is to understand why we are getting multiple string instances, and to fix that. Note that we do not need to *guarantee* that all string instances will be the same - it is always possible for users to do something strange and step off a performance cliff. What matters is that it is *possible* to avoid the issue.

Sphyxis: I will dig more into what is causing the source code duplicating in the first place, see if I can find it's root and patch it there.

Sphyxis: Ok.  Tracing the source of the problem was easier than I thought it would be.

The duplication starts at the Script object level.  The game essentially says that one file on one server = one Script object.  Each Script object owns it's own code field.  Every script file saves its full source independently.  So every new file creates a new Script and scp just writes the source text into another servers' Script object.

This essentially means that the problem is at the file/save area since script has it's own code, which is used in it's own runtime, and each script keeps a separate instance of it's source code.

The cache allows all of these separate instances, each with their own source code area, to share a single source.  The ref count is ONLY there to allow it to be disposed of once all scripts are done, so the source code leaves memory.  If it's taken out, then each scripts will essentially write their source once to this cache and then always share it.  It will never be GC'd or have it's ref count messed up.  The performance gains however are huge.  Moving from 100k batches (400k scripts) to 300k batches (1200k scripts) with some room to grow is big.

The cache is setting the shared code at the point the Script.code is set.  In order to fix it before it reaches that point, which is something I've been thinking of doing, you need to change the save methods from this:
```js
{
"filename": "a.js",
"code": "Large source code A"
}
```
to something like:
```js
{
"filename": "a.js",
"code": "hash-or-id"
}
```

And then later, you will need a separate area to match it, like this, with a ScriptSource definition.
```js
{
   "sources": {
      "hash-or-id": "Large source code A"
   }
}
```
After this, each Script would have:
```js
source: ScriptSource
```
instead of:
```js
code: string
```


If you are alright with me tackling this type of change I could work on it.  Once it's in, we could migrate away from the cached script code entirely and just use the hash-or-id lookups.  Until then, I could remove the ref counts so they cannot get messed up or missed in another change.  Caching script code, from my testing, is alright as long as it's being referenced and not duplicated.

This will also have the benefit of creating a single source code entry in the save file for all duplicate scripts in the network.  This change should also lower save file size, loading time, etc.

src/NetscriptJSEvaluator.ts
d0sboots: > This can be easily replicated by writing a file from memory, running it, deleting it and then re-writing it. You can also test unique filename with identical content in the same way.

This does not match my testing. Am testing on latest dev. (So, after the removal of WeakRef in the cache.)

```js
/** @param {NS} ns */
  ns.disableLog("ALL");
  ns.ui.openTail();
  let i = 0;
  let p = [];
  while (true) {
    // Explicitly use an expression here to try and force different string instances
    ns.write("temp.js", "export function main() {/*" + " ".repeat(1000000) + "*/}", "w");
    p.push(ns.dynamicImport("temp.js"));
    ns.rm("temp.js");
    i++;
    if (i % 100 === 0) {
      ns.print(i);
      await Promise.all(p);
      p = [];
      if (i % 1000 === 0) {
        await ns.sleep(0);
      }
    }
  }
}
```

Result: 150k+ iterations in less than a minute. Memory remains constant. Looking at memory, there are exactly two copies of the string, one linked to the module and one being used as context within `compile`, I believe as part of its JIT. My guess is that the first is the first copy of the string that was created (the module is never recreated, so this would be the first string), whereas `compile` has to be called many times before it gets JIT-compiled, so the copy it uses is different.

Looking at perf, all of the time is spent in `generateLoadedModule`, specifically the acorn parsing. It *is* rerunning the code, but as I asserted, when given the same input it produces the same output, and thus the cache lookup returns the same `LoadedModule`.

src/Script/Script.ts
d0sboots: OK, I also thought of a way this could be implemented with WeakRef and FinalizationRegistry. It wouldn't be that much code, but it would still be better to avoid it if at all possible.

Sphyxis: I really don't like the idea of relying on WeakRef anywhere.  I've been thinking about this exact change for a few months now - a central point for all script code references that everything shares.  I could just take out the code caching method I have for now and focus on on the hash-or-id idea.  It does work wonders though as it is.

src/Netscript/killWorkerScript.ts
```diff
@@ -119,4 +119,5 @@ function removeWorkerScript(workerScript: WorkerScript): void {
   if (rs.temporary === false) {
     AddRecentScript(workerScript);
   }
+  workerScript.code = "";
```
d0sboots, line 122: This is unneeded. This function (should) have the last reference to `workerScript` at this point.

src/Script/Script.ts
d0sboots: > This essentially means that the problem is at the file/save area since script has it's own code, which is used in it's own runtime, and each script keeps a separate instance of it's source code.

This analysis is insufficient. Just because two separate variables (or in this case, two separate `Script`s) exist, it does not mean that they do not have a reference to the same object/string. I know full well every file is its own `Script`, each with its own `code` field. However, when you call `scp` (for instance), you are (through a chain of methods that involves the `content` setter) assigning the reference from one `Script.code` to another, and now they share the *same* reference. Or at least, they should.

The task is to figure out what is interfering in that process; why is it that what should be naturally shared references, are not?

d0sboots: > I really don't like the idea of relying on WeakRef anywhere.

This is a good instinct - you never want to *rely* on WeakRef, because in the worst case (and also in quite a few common cases) it will never get GC'd at all. And even in practical systems, it can be "arbitrarily long" (which can be minutes even) before you get the kind of GC that triggers WeakRef and FinalizationRegistry.

But, that is why we'd only be (weakly) relying on them for GC properties - i.e., to dispose of strings that no one is using anymore. The deduping properties would be strongly guaranteed by the regular map.

(Edit: Part of your distaste may come from fixing the previous "bug" with the other cache that used WeakRefs. However, that cache was working exactly as designed all along - technically speaking, it never had a bug. It's just that the design ended up being insufficient for the tasks we wanted it to perform.)

Sphyxis: I believe I can convert the script code cache to use a WeakRef then, and completely remove the refs counter.  The refs only purpose was to allow GC to trigger when nothing referred to it, but it can be done with a WeakRef, even if I have to wrap the code string in something and come up with a different reference key.

I can double check the scp path to see if the code string is referenced throughout.  It's when you have a file in the script itself that is written to the server that it gets it's own code section, as it has nothing to reference.

src/NetscriptJSEvaluator.ts
Sphyxis: You are using dynamic import there.  Use ns.exec instead and you should see the issue.

src/Netscript/killWorkerScript.ts
Sphyxis: I'd like to test this one out a bit more.  This was added after I had tested out the main patch against my darknet script overnight.  It still had a memory leak, which looked like it came from the script transform process (for .jsx, .ts, etc).  As soon as I put that in, I couldn't duplicate it anymore.

src/NetscriptJSEvaluator.ts
Sphyxis: You should also be able to duplicate the issue by giving the file a unique name, but the same content.

src/Script/Script.ts
Sphyxis: I have not looked at those paths yet, but I did have a thought.  You cannot create a reference for script code on a Script object.  If you do, you create a scenario where you have a master copy that changes all other copies.

Say you make `a.js`.  You then scp `a.js` everywhere.  If those copies had references to the content of `a.js`, then when you changed `a.js` only `a.js` has it's module invalidated.  All other copies would have their original modules, especially if they have been run, and because of that they will not search for a cached one.  But, they will run with the new code of the original `a.js`.  In order for a reference to work, you would have to re-reference all remaining scripts to one of their own copies of the code - which they don't have because they will all be referencing the new code of `a.js`.

Does this sound correct?

src/NetscriptJSEvaluator.ts
Sphyxis: Also, my MRE code has a test that will show this happening.

src/Script/Script.ts
d0sboots: This shows a fundamental misunderstanding of how strings and references work.

Like in most languages (that aren't C/C++), strings are immutable. So, you cannot change `a.js` in the sense of "changing" its `Script.code` - if you (say) change it by doing `myScript.code += " /* extra junk*/`, that is equivalent to `myScript.code = myScript.code + " /* extra junk*/`, which creates a *new* string that is the concatenation of the old and new parts, and sets the reference to the new string in `myScript.code`. So now `a.js` has a new code reference, all the other scripts still have the old reference (containing the old code), and everything works correctly.

(This is actually an oversimplification, because we (intentionally) cannot observe reference identity for strings in JS like we can for objects. So, in fact, strings *can* be mutable (and are), if the system can prove that nothing will be broken by doing this. It uses this to optimize loops where strings are repeatedly added to character-by-character, since those are typically done with new strings whose reference hasn't escaped outside that scope yet.)

src/NetscriptJSEvaluator.ts
d0sboots: I used your MRE, and cannot reproduce the leak either. I fixed some bugs (you weren't checking the result of exec, you had the wrong order of ops in your thread calc, you were deleting the scripts off of home such that nothing was being executed, etc.) and modified it slightly to print the number of generatons in the log window, since despite the comment it did not seem to be printing that anywhere. I also changed `DELETE_BEFORE_SCP` to true and `START_SCRIPTS` to true, since that seemed to be the test you were trying to perform. I also took out a bunch of sleeps so it doesn't take so long to run. I'll post the whole thing below, since the total modifications ended up being non-trivial, but they didn't touch the core of the tests.

As of 180000 generations, memory usage has stayed totally flat. Looking at the memory profile, there are exactly two instances of your 500kb code in memory. One is shared by all the Scripts and all the WorkerScripts. The other is the copy that is augmented with source map, that is referenced by the single compiled module. I see zero evidence of any of the kind of leaks you were talking about.

image

```js
const SIZE = .5 //In MB
const COPIES = 32
const PREFIX = "mre-source-bloat/"
/**
 * Backup your save.  These tests can cause your game to fail to save or corrupt it
 * While I'm looking at fixing modules and caching code it still creates new Script instances
 * 
 * Key: Delete = 0, Unique = 1, Start = 2
 * 
 * General control:  0 = false, 1 = false, 2 = false
 * - Dev: memory will plateau
 * - Fix: same
 * 
 * Source/file object recreation:  0 = true, 1 = false, 2 = false
 * - Dev: Some churning, and memory should plateau at a higher level as each script uses it's own code
 * - Fix: Same, but less churning.  Memory should plateau earlier due to code sharing
 * - Increasing the file size or count should increase the gap between dev and this fix
 * 
 * Module Leak (Main test):  0 = true, 1 = false, 2 = true
 * - Dev: Memory should grow due to new modules being created from missed cache lookups
 * - Fix: Memory should plateau as module cache is now based on source/cacheKey
 * 
 * Save/file bloat: 0 = false, 1 = true, 2 = true
 * - Dev: Grows fast as new script code/modules accumulate
 * - Fix: Still grows from file/save data, but live source-string duplication should be lowered
 * - I ran this for both Dev and Fix and analyzed the memory dumps.  The results were great
 * - The fix had a lower dump size, ran longer, generated more unique names by over 4x, and a total of 1 Source-Code shared.
 *   The dev branch had 16k unique file names and 1,024 Source-Code shared
 * - The main bloat from this is the save data accumulation from all the files.
 */
const DELETE_BEFORE_SCP = true
const UNIQUE_FILE_NAMES = false
const START_SCRIPTS = true
/** @param {NS} ns */
  ns.disableLog("ALL")
  ns.ui.openTail()
  const files = Array.from({ length: COPIES }, (_, i) => `copy-${i}.js`)
  const targets = scanAll(ns)
    .filter((server) => ns.hasRootAccess(server))
    .filter((server) => ns.getServerMaxRam(server) >= 2)
  ns.tprint(targets);
  ns.atExit(() => {
    for (const target of targets)
      for (const file of ns.ls(target, PREFIX)) {
        if (!ns.rm(file, target)) {
          ns.print(`Couldn't clean up ${target}: ${file}`);
        }
      }
  })

  const victimBody = makeLargeVictimScript(SIZE)
  let lastYield = performance.now()
  let generation = 0

  while (true) {
    ns.print(generation);
    generation++ //We can see this in the console
    const gen_prefix = UNIQUE_FILE_NAMES ? ns.sprintf(`-Gen-${generation}-`) : ""
    if (performance.now() - lastYield >= 200) {
      lastYield = performance.now()
      await ns.sleep(0)
    }
    for (const target of targets) {
      for (const file of files) {
        const fileName = `${PREFIX}${gen_prefix}${file}`
        if (DELETE_BEFORE_SCP) {
          ns.rm(fileName, target)
        }
        if (target === "home") {
          ns.write(fileName, victimBody, "w")
        } else {
          ns.scp(fileName, target, "home")
        }
        if (START_SCRIPTS) {
          const threads = Math.floor(Math.max(0, ns.getServerMaxRam(target) - ns.getServerUsedRam(target) - (target === "home" ? 32 : 0)) / 1.6)
          if (threads) {
            if (!ns.exec(fileName, target, { temporary: true })) {
              ns.print("ERROR: Running on " + target);
            }
          }
        }
      }
      if (performance.now() - lastYield >= 200) {
        lastYield = performance.now()
        await ns.sleep(0)
      }
    }
  }
}

function scanAll(ns) {
  const serverList = new Set(["home"])
  for (const server of serverList) {
    for (const connection of ns.scan(server)) {
      serverList.add(connection)
    }
  }
  return Array.from(serverList)
}

function makeLargeVictimScript(sizeMb) {
  const targetBytes = sizeMb * 1024 * 1024;
  const line = "SOURCE_BLOAT_MRE_PAD_0123456789abcdef".repeat(32) + "\n"
  let pad = "";

  while (pad.length < targetBytes) {
    pad += line;
    if (pad.length > 1024 * 1024 && pad.length * 2 < targetBytes) pad += pad
  }

  pad = pad.slice(0, targetBytes)

  return `/*
${pad}
*/
  ns.disableLog("ALL");
  await ns.sleep(50)
}`
    ;
}
```

Sphyxis: Now I can't replicate it either.  Something is messing with me somewhere.

src/Netscript/killWorkerScript.ts
Sphyxis: I've only been able to track the use/change/reading of workerScript.code to 2 places:

WorkerScript.ts:27  Which is  `code = "";`
WorkerScrip0t.ts:101  Which is `this.code = script.code`

I can't find anything else that references it.  Checking the execution path, it seems that we gets the script from calling `workerScript.getScript()`.  That script is then compiled, we get the module and we start.  Could you double check to make sure I'm not missing something?

If it's a dead area, I'll remove the .code part entirely in a separate PR.  I've already made up a quick one with the property removed, and it passes lint without issues and runs scripts with no errors.

src/NetscriptJSEvaluator.ts
Sphyxis: Something must have glitched out on my end.  Everything reads fine now.  I'll strip out the cacheKey section entirely.

src/Script/Script.ts
Sphyxis: And now that my tests are all coming back negative for memory leaks, I am still able to go up to 300k batches.  I'm thinking something did glitch out on me after that WeakRef fix.  Now, I'm not getting massive script copies in memory for 300k batches, I'm getting around 850 of each worker.  I'm going to run some more tests, but I think that WeakRef fix actually fixed this too.

Sphyxis:
I apologize for wasting your time.  It seems all of the issues I was trying to fix were already fixed by the last WeakRef fix.  Thank you for the knowledge on Strings though.  I came from C, where I had to specifically create a duplicate string in memory to assign, so it wouldn't get overridden by changing a different one.  I did find a few other areas that could likely benefit though.  I'll test them out in a different PR.

d0sboots:
I'm glad in the end that you got the same results too! It would have been really frustrating for everyone if it was a memory leak localized entirely to your setup.

PR #2861, opened 2026-06-07
MISC: Add missing description for exe programs

E9cipher:
This PR adds a more detailed description to the .exe programs (concretely the port openers).

### DIFF
--- a/src/NetscriptFunctions.ts
+++ b/src/NetscriptFunctions.ts
@@ -613,11 +613,11 @@ export const ns: InternalAPI<NSFull> = {
       return false;
     }
     if (!server.sqlPortOpen) {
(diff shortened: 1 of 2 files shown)


catloversg:
The tooltip changes are okay, but I don't think we need to change the port number.

Port 1433 is the default port for Microsoft SQL Server, while port 3306 is the default port for MySQL. Changing it would effectively mean preferring MySQL over Microsoft SQL Server. There is no right or wrong choice here. I'm okay with using 3306 or any other port if this is a new feature, but Daniel (the original developer) chose this port number years ago. I don't want to change that.

E9cipher (commented):
Done. I thought 3306 was the standard for MySQL, didn't think of Microsoft's server. I understand and respect your and the original developer's idea, so it's changed to 1433

PR #2863, opened 2026-06-08
UI: Add an easter egg to the glitch

d0sboots:
After ~25 seconds in the glitch, the user will be dumped in The Void. Just a little spooky easter-egg flavor.

Basic manual tests indicate it seems to work.

### DIFF
--- a/src/Locations/ui/SpecialLocation.tsx
+++ b/src/Locations/ui/SpecialLocation.tsx
@@ -10,11 +10,12 @@
  * This subcomponent creates all of the buttons for interacting with those special
  * properties
  */
-import React, { useCallback, useState } from "react";
+import React, { useCallback, useState, useEffect } from "react";
 import Typography from "@mui/material/Typography";
 import Button from "@mui/material/Button";
 
-import { Location } from "../Location";
+import type { Location } from "../Location";
+import { Locations } from "../Locations";
 import { CreateCorporationModal } from "../../Corporation/ui/modals/CreateCorporationModal";
 import { AugmentationName, CompletedProgramName, FactionName, LocationName, ToastVariant } from "@enums";
 import { Factions } from "../../Faction/Factions";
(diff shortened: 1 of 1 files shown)


src/Locations/ui/SpecialLocation.tsx
```diff
@@ -350,7 +351,19 @@ export function SpecialLocation(props: SpecialLocationProps): React.ReactElement
...
+    useEffect(() => {
+      let delay = 0;
+      // This is a sum of 25 exponential random variables, which is equivalent
+      // to one Erlang-distributed random variable with mean 25sec and stddev 5sec.
+      for (let i = 0; i < 25; ++i) {
+        delay += -1000 * Math.log(1 - Math.random());
+      }
+      const id = setTimeout(() => Router.toPage(Page.Location, { location: Locations[LocationName.Void] }), delay);
+      return () => clearTimeout(id);
+    });
```
catloversg, line 365:
```suggestion
    }, []);
```
Is the missing dependency array intentional? Without it, this effect runs after every render and reschedules the timeout each time. While this component is unlikely to rerender, adding [] would make it clear that the timeout should only be scheduled once on mount.

d0sboots: It was intentional. I'm once again fuzzy on exactly what will count as a mount of this component, but I'm more concerned about the timeout being reset too little than too much.

catloversg: This component only rerenders when its parent rerenders after joining Bladeburner or purchasing dnet exe, but when that happens, this component has not been rendered yet. This means the component effectively never rerenders.

If you want to reschedule the timeout on every rerender, it's fine to omit the dependency array.

d0sboots: Yeah, I was pretty sure that right now there was no difference, it was more about hypothetical future changes. In most scenarios that I can think of, if something causes this to rerender it should reset the timer.

PR #2871, opened 2026-06-10
BUGFIX: ns.singularity.destroyW0r1dD43m0n validates BN options too late

catloversg:
MRE:
- Set hack level to WD required hack level and nuke WD.
- `ns.singularity.destroyW0r1dD43m0n(1, undefined, {})`. The script throws an error because the options are invalid.
- Save game and reload. The BitVerse screen is shown.

The BN options are validated too late, so `wd.backdoorInstalled` can be set to true even when `_bitNodeOptions` is invalid.


E9cipher:
Conflicts with #2857?

catloversg:
Depends on what "conflict" means. It will conflict with the Jest tests in that PR, but the goal of this PR is unrelated. This bug can be used to simulate the "stay-on-BitVerse" effect, but it's still a bug.

d0sboots:
I'm gonna merge this 2nd, as a favor to E9cipher so they don't have to figure out how to resolve merges.

E9cipher:
Thanks, but in my PR there are still jest tests to do, and I think this one is ready to merge. You can merge this one first and I'll try to resolve merge conflicts, if you could help me there I'd really appreciate it.

catloversg:
I'll handle #2857 from now on, so it's up to d0sboots to decide which PR should be merged first. I'll deal with the merge conflicts.

PR #2873, opened 2026-06-11
feat(ui): add ns.ui.openPage/closePage for script-created sidebar pages

iSteed:
## Summary

This PR adds two new functions to the `ns.ui` namespace that allow scripts to create named pages in the game sidebar:

- **`ns.ui.openPage(title, content?)`** — creates a page in the sidebar (or updates its content if it already exists). Calling this in a loop is the intended update pattern.
- **`ns.ui.closePage(title)`** — removes the page. Pages are also automatically removed when the owning script exits or is killed.

Both functions cost **0 GB RAM**.

### Motivation

Currently, scripts that want to display persistent status information have to use `ns.ui.openTail()` and a floating log window. This works, but tail windows:
- Float over game content
- Don't feel like a native part of the UI
- Have to be manually positioned/resized
- Show script log output rather than structured content

`openPage` lets scripts render content as a first-class game page, appearing in the sidebar just like Terminal, Stats, or Hacknet.

### Example usage — plain string

Strings are rendered in a `<pre>` block using the player's configured monospace font, so column-aligned text works the same as in tail windows:

```js
  while (true) {
    const lines = [
      `Money:  ${ns.formatNumber(ns.getServerMoneyAvailable("home"))}`,
      `RAM:    ${ns.getServerUsedRam("home").toFixed(1)} / ${ns.getServerMaxRam("home")} GB`,
      `Hacks:  ${ns.getHackingLevel()}`,
    ].join("\n");

    ns.ui.openPage("Status", lines);
    await ns.sleep(1000);
  }
}
```

### Example usage — React JSX (*.jsx / *.tsx)

`content` also accepts a `ReactNode`, giving full control over layout and styling — the same model as `ns.printRaw` / `ns.tprintRaw`:

```jsx
  while (true) {
    const money = ns.getServerMoneyAvailable("home");
    ns.ui.openPage("Status",
      <div style={{ padding: "1rem" }}>
        <h2>Home Server</h2>
        <p>Money: {ns.formatNumber(money)}</p>
      </div>
    );
    await ns.sleep(1000);
  }
}
```

## Implementation

| File | Change |
|------|--------|
| `src/ui/CustomPageManager.ts` | New singleton registry — stores pages by title, notifies React subscribers on mutation. `content` typed as `ReactNode`. |
| `src/ui/CustomPages/CustomPageRoot.tsx` | New React component — subscribes to the manager and re-renders on content updates. Strings render in a Typography `pre`; React nodes render directly in a Box. |
| `src/ui/Enums.ts` | `ComplexPage.ScriptPage = "Script Page"` |
| `src/ui/Router.ts` | `PageContext<ScriptPage> = { id: string }`, added to `PageWithContext` union |
| `src/ui/GameRoot.tsx` | `case Page.ScriptPage` renders `<CustomPageRoot>`, passes full `pageWithContext` to sidebar |
| `src/Sidebar/ui/SidebarRoot.tsx` | `ScriptPageItems` — dynamic section hidden when no pages exist, shows one item per registered page |
| `src/Netscript/killWorkerScript.ts` | `CustomPageManager.closePagesByPid(pid)` called on script termination |
| `src/NetscriptFunctions/UserInterface.ts` | `openPage` and `closePage` implementation. Strings passed through; other values wrapped with `wrapUserNode` for error-boundary protection. |
| `src/Netscript/RamCostGenerator.ts` | `openPage: 0, closePage: 0` |
| `src/ScriptEditor/NetscriptDefinitions.d.ts` | Full JSDoc public API types with both string and JSX examples |

## Test plan

- [ ] `run test_page.js` creates a sidebar entry and shows updating content
- [ ] Killing the script removes the sidebar entry automatically
- [ ] `ns.ui.closePage()` removes the entry from a running script
- [ ] Calling `openPage` with the same title updates content without creating a duplicate
- [ ] Multiple scripts can register different pages simultaneously
- [ ] Navigating to another page and back preserves content
- [ ] Plain string content renders in monospace pre-formatted style
- [ ] React JSX content (in a .jsx file) renders with full layout control
- [ ] TypeScript types are correct and JSDoc appears in the script editor

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/Netscript/RamCostGenerator.ts
+++ b/src/Netscript/RamCostGenerator.ts
@@ -457,6 +457,8 @@ const ui = {
   alias: 0,
   unalias: 0,
   getAllAliases: 0,
+  openPage: 0,
+  closePage: 0,
 } as const;
 
 // Grafting API
--- a/src/Netscript/killWorkerScript.ts
+++ b/src/Netscript/killWorkerScript.ts
@@ -11,6 +11,7 @@ import { AddRecentScript } from "./RecentScripts";
 import { handleUnknownError } from "../utils/ErrorHandler";
 import { roundToTwo } from "../utils/helpers/roundToTwo";
 import { BaseServer } from "../Server/BaseServer";
+import { CustomPageManager } from "../ui/CustomPageManager";
 
 export function killWorkerScript(ws: WorkerScript): void {
   stopAndCleanUpWorkerScript(ws);
@@ -119,4 +120,5 @@ function removeWorkerScript(workerScript: WorkerScript): void {
   if (rs.temporary === false) {
     AddRecentScript(workerScript);
   }
+  CustomPageManager.closePagesByPid(workerScript.pid);
 }
--- a/src/NetscriptFunctions/UserInterface.ts
+++ b/src/NetscriptFunctions/UserInterface.ts
@@ -16,6 +16,7 @@ import { hasTextExtension } from "../Paths/TextFilePath";
 import { errorMessage } from "../Netscript/ErrorMessages";
 import { addGlobalAlias, addAlias, removeAlias, Aliases, GlobalAliases, aliasRegex } from "../Alias";
 import { assertStringWithNSContext } from "../Netscript/TypeAssertion";
+import { CustomPageManager } from "../ui/CustomPageManager";
 
 /** Converts the provided value to a string and ensures it satisfies the alias condition, throwing if it is not  */
 export function parseAsAlias(ctx: NetscriptContext, argName: string, v: unknown): string {
@@ -277,5 +278,22 @@ export function NetscriptUserInterface(): InternalAPI<IUserInterface> {
 
       return returnMap;
     },
+
+    openPage:
+      (ctx) =>
+      (_title, _content = "") => {
+        const title = helpers.string(ctx, "title", _title);
+        const content = helpers.string(ctx, "content", _content);
+        if (!title) throw helpers.errorMessage(ctx, "'title' cannot be empty.");
+        CustomPageManager.openPage(ctx.workerScript.pid, title, content);
+      },
+
+    closePage:
+      (ctx) =>
+      (_title) => {
+        const title = helpers.string(ctx, "title", _title);
+        if (!title) throw helpers.errorMessage(ctx, "'title' cannot be empty.");
+        CustomPageManager.closePage(title);
+      },
   };
 }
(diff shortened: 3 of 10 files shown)


ficocelliguy:
the page content should probably be jsx, not a string, in the same way that printRaw and tPrintRaw expect react components or HTML

iSteed:
Thanks, that makes sense. I’ll take a look at changing this over so the page content accepts JSX/React components instead of just a string, like you suggested. 

Just to be transparent: I had Claude do most of the heavy lifting on the implementation here. I know enough to get something working for my use case, but I’m still getting comfortable with the codebase and PR-level changes. Since this works well for what I needed, I figured I’d open it up in case the general idea is useful and the implementation can be adjusted into the right shape.

catloversg:
You can render the sidebar item with the built-in hooks (`sidebar-extra-hook-<number>`). A built-in API would need to have a lot of optional parameters to let players customize the item. Using the hooks lets them render whatever they want. That's also why we have hooks in the overview window.

The missing part is an API to render a custom page in the main content area. This will be added in #2875. I think this is a better solution. It's extremely simple on our side (e.g., there is no need for a manager to keep track of state), and players have maximum customizability.

iSteed:
Thanks for taking a look and for #2875 — that's a much simpler approach than what I did.

I think the main thing I'd miss with `renderPage` is that `openPage` adds a sidebar entry, so the player can navigate back to it after clicking away. With `renderPage` the content disappears as soon as you leave. Also since it's a named slot, multiple scripts can each have their own page without stepping on each other.

Not sure if that's worth the extra complexity — just wanted to flag it in case it's useful!

d0sboots:
I think I'm with catlover on this one - UI is a very diverse space, so it's best to provide simple building blocks rather than rigid ones. The trade-off is more complexity on the use side, but it's always going to be complex, and we have a pretty good stepping stone with printRaw and tail windows for people to get their feet wet.

iSteed:
thanks for taking a look, while i made this PR to be a low friction way for people to get info on the screen, i have gone back and fourth and decided maybe it doesn't belong in this repo. Part of the spirit of bitburner, is the tools are just barely good enough to get by(such as a big tail window you render to), and annoy you into building your own tooling and do some real learning

catlover's approach is much more aligned with that idea. Where I slightly disagree is for people who are very fresh to coding. For someone just starting out, finding the right UI hook can feel out of reach before they even get to the experimenting part. the persistent page maybe isn't worth the extra overhead, and catlovers way is a great way to learn something and push yourself. 

Maybe the better compromise is not adding this API directly, but improving the docs around the overview hooks. For example, we could include small examples for rendering into the main area and the sidebar, so new players have a clear starting point without fully abstracting away the learning.

d0sboots:
> Maybe the better compromise is not adding this API directly, but improving the docs around the overview hooks. For example, we could include small examples for rendering into the main area and the sidebar, so new players have a clear starting point without fully abstracting away the learning.

Yes, I think a help topic that's specifically focused around the diversity of UI options we have, and how to use them, is a missing spot right now.

PR #2874, opened 2026-06-11
API: Follow up to #2806

catloversg:
While running `npm run doc`, api-extractor prints some warnings related to the new `ns.ui.alias` API. While fixing the TSDoc for that API, I found some issues in #2806. This PR is basically the review that vivian-wei requested in that PR.

The biggest issue is this part in the TSDoc:
```
   * This function throws an error if alias/substitution are empty strings after leading and trailing whitespace are removed. It also throws if
   * alias has any invalid characters (not alphanum or `|!%,@-`).
```
Currently, the substitution can actually be an empty string. For example, running `alias foo="    "` in the terminal will alias `foo` to an empty string. This may not be the desired behavior, but as long as this API works nearly identically to the CLI command, the docs should accurately describe the current behavior.

```
alias has any invalid characters (not alphanum or `|!%,@-`)
```
This part is also incorrect.

The next issue is the implementation:
https://github.com/bitburner-official/bitburner-src/blob/763bd6c93ac297ffd30fcb66220676a0d91fb68c/src/NetscriptFunctions/UserInterface.ts#L230-L238

The first check is unnecessary. `parseAsAlias` already handles the empty string case.

The second check is wrong and outdated. As I said above, `substitution` should be able to be an empty string. If we want to change this behavior, we need to do that in a separate PR.

The phrase "only contain whitespace" is outdated. In https://github.com/bitburner-official/bitburner-src/pull/2806/changes/31a01c230c0fc3bd44aaa94f7d2a4c48ee7eba3c, there was a `trim()` call when parsing `_substitution`. That call was removed later due to https://github.com/bitburner-official/bitburner-src/pull/2806/changes/31a01c230c0fc3bd44aaa94f7d2a4c48ee7eba3c#r3365850789. However, the error message was not updated.

### DIFF
--- a/markdown/bitburner.userinterface.alias.md
+++ b/markdown/bitburner.userinterface.alias.md
@@ -9,7 +9,7 @@ Programmatically sets an alias.
 **Signature:**
 
 ```typescript
-alias(alias: string, substitution: string, global?: boolean): void;
+alias(alias: string, substitution: string, isGlobal?: boolean): void;
 ```
 
 ## Parameters
@@ -42,7 +42,7 @@ string
 
 </td><td>
 
-The keyword to set.
+The alias name to set.
 
 
 </td></tr>
@@ -64,7 +64,7 @@ The substitution to run.
 </td></tr>
 <tr><td>
 
-global
+isGlobal
 
 
 </td><td>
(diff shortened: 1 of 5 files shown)


d0sboots:
I think the current state is fine. What's useful to note is that this command and the shell do *not* have exactly the same behavior, because of shell processing. As a result, the ns function can set aliases that contain leading/trailing whitespace (or whitespace-only aliases). The command form can set empty aliases. With more robust shell quoting, the shell would also be able to do what the command can do, but I don't think it's a priority.

PR #2875, opened 2026-06-11
API: Add ns.ui.renderPage to render React node in main content area

catloversg:
#2651 added hooks to the sidebar for players to attach custom content. This custom content is usually a button that opens a custom page or HUD. Currently, players usually create a HUD using the overview window or tail log windows. However, they cannot render a custom page in the main content area.

This PR adds a new API for that purpose. It's basically the same as `ns.printRaw`, but for the main content area instead of a tail log window.

Test code:
```jsx
/** @param {NS} ns */
  ns.ui.openTail();
  ReactDOM.render(<SidebarItem ns={ns} text={"foo"} />, document.getElementById("sidebar-extra-hook-0"));
  ReactDOM.render(<SidebarItem ns={ns} text={"bar"} />, document.getElementById("sidebar-extra-hook-1"));
  return new Promise(() => { });
}

function Timer({ text }) {
  const [seconds, setSeconds] = React.useState(0);

  React.useEffect(() => {
    const interval = setInterval(() => {
      setSeconds((seconds) => seconds + 1);
    }, 1000);
    return () => {
      clearInterval(interval);
      console.log("Unmount Timer");
    };
  }, []);

  return <div>{text} Seconds: {seconds}</div>;
}

/** @param {{ ns: NS }} */
function SidebarItem({ ns, text }) {
  return (
    <span
      onClick={() => {
        ns.ui.renderPage(<Timer text={text} />);
      }}
    >
      {text}
    </span>
  );
}
```

https://github.com/user-attachments/assets/b98cf41d-632c-4eeb-9c6b-3ff1f6733615


src/ui/Enums.ts
```diff
@@ -56,4 +56,5 @@ export enum ComplexPage {
   // is constructed as a ComplexPage with no PageContext, and thus toPage() cannot be used (since no overload will fit
   // it).
   LoadingScreen = "Loading Screen",
+  CustomPage = "Custom Page",
```
d0sboots: Because of the long comment and its special status, let's keep LoadingScreen last.

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -7080,6 +7080,16 @@ interface UserInterface {
    * @returns A map of alias names to an object containing the substitution string and if the alias was set to global.
    */
   getAllAliases(): Map<string, { substitution: string; isGlobal: boolean }>;
+
+  /**
+   * Render a ReactNode in the main content area.
```
```suggestion
   * Renders a ReactNode in the main content area.
```

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -7080,6 +7080,16 @@ interface UserInterface {
...
+   *
+   * @remarks
+   * RAM cost: 0 GB
+   *
```
d0sboots, line 7089: There needs to be a lot more details in the remarks - even I didn't understand what this was gonna do until I read the source. 

Things to include: 
* What is the "main content area?" (It is the space that shows the terminal, city page, editor, etc.)
* That this function will effectively switch to a new custom "page", as if you had navigated via the sidebar. Calling it again replaces the contents of this page.

PR #2877, opened 2026-06-12
REWORK: Only create script title when needed

Sphyxis:
This PR aims to save a bit of memory by removing the creation of a script title on run, and to unclutter some of the debug logs.  This has no impact on the player or API calls, as the title is created on read or reference.

Memory Savings:
A default title, depending on characters + overhead, can take around 90-150 bytes per title (For a hack script in a batcher)
This means, for 300k scripts (Or the standard 100k batches as HGW) this can take between 27mb to 45mb (90 bytes to 150 bytes * 300k)
V8 can store some in ways that increase it's size though, like two-byte strings.
This brings the range closer 60-80mb, and that's just the title.
So, if you are pushing things and starting hundreds of thousands of scripts, this starts to add up.
It also removes the overhead of having memory saturated with a title for every script when it's not needed the majority of the time.
We are also removing the full copy of args that WorkerScript was keeping.  It does not need it as it can access it already through scriptRef
Estimated cost/savings for args:
Each script args from a batcher should take up between 48 bytes and 96 bytes.
For 300k scripts, this ends up being a savings of 14.4mb - 28.8mb
The savings is a bit lower due to it being an array, rather than a full and new joined string.

This also fixes a minor bug in ui\React\LogBoxManager.tsx related to titles:
`const title_str = script.title === "string" ? script.title : `${script.filename} ${script.args.join(" ")}`;`
-Note the missing typeof
to
`const titleText = typeof displayTitle === "string" ? displayTitle : defaultTitle;`

MRE to test API functionality:
```js
/** @param {NS} ns */
  ns.disableLog("ALL")
  //3 API entry points found to get the title.
  //Ensure at least 1 script in recents.
  //In all examples, .title will not be there until it's looked at.
  //Setting the title also works just fine.

  ns.tprintf("Testing ns.self()")
  const me = ns.self()
  debugger
  ns.tprintRaw(JSON.stringify(me, null, 2))
  ns.tprintf("\nTesting getRunningScript()")
  const running = ns.getRunningScript(ns.pid)
  debugger
  ns.tprintRaw(JSON.stringify(running, null, 2))
  ns.tprintf("\nTesting getRecentScripts()")
  const recent = ns.getRecentScripts()
  debugger
  ns.tprintRaw(JSON.stringify(recent, null, 2))
  ns.ui.openTail()
  await ns.sleep(2000)
  ns.ui.setTailTitle("Test 1")
  await ns.sleep(2000)
  ns.ui.setTailTitle("Test 2")
  await ns.sleep(2000)
}
```

EDIT:
-I tested my batcher and darknet scripts out for a few hours with 0 issues.
-I also tested saving with an active script without a title, closing and reopening the game.  No issues.


src/Script/RunningScript.ts
```diff
@@ -157,7 +158,7 @@ export class RunningScript {
         ...this,
         dataMap: Object.fromEntries(this.dataMap.entries()),
       },
-      typeof this.title === "string" ? includedProperties : includedPropsNoTitle,
+      typeof this.title === "string" && this.title !== "" ? includedProperties : includedPropsNoTitle,
```
Sphyxis: Quick thought, this can be a React element too.  Should it only check for string here?

src/Netscript/NetscriptHelpers.tsx
```diff
@@ -856,10 +856,15 @@ function getCannotFindRunningScriptErrorMessage(ident: ScriptIdentifier): string
  */
 function createPublicRunningScript(runningScript: RunningScript, workerScript?: WorkerScript): IRunningScript {
   const logProps = runningScript.tailProps;
+  //Create a reference to the strings, not the runningScript
+  const args = runningScript.args.slice();
```
d0sboots: This shouldn't do anything. If you can prove that it *does* do something, it needs a much more involved comment explaining how.

src/Netscript/NetscriptHelpers.tsx
```diff
@@ -882,7 +887,9 @@ function createPublicRunningScript(runningScript: RunningScript, workerScript?:
             fontSize: logProps.fontSize ?? Settings.styles.tailFontSize,
             minimized: logProps.minimized,
           },
-    title: runningScript.title,
+    get title(): string | React.ReactElement {
```
d0sboots: This is (minorly) api-breaking and just not worth it. This would change the publicly exposed RunningScript object from a POD to something with a single member that can't be assigned, which I'm sure would break someone somewhere. It's also super irrelevant - the functions like getRunningScripts are not in the hot path of the memory concerns here.

src/Script/RunningScript.ts
```diff
@@ -73,9 +73,7 @@ export class RunningScript {
   // Access to properties of the tail window. Can be used to get/set size, position, etc.
   tailProps = null as LogBoxProperties | null;
 
-  // The title, as shown in the script's log box. Defaults to the name + args,
-  // but can be changed by the user. If it is set to a React element (only by the user),
-  // that will not be persisted, and will be restored to default on load.
+  // Custom title for the script's log box. Undefined means use the default filename + args title only when looked at.
```
d0sboots: (Mostly) revert this, you deleted important info.
```suggestion
  // The title, as shown in the script's log box. Null means use the default filename + args,
  // but the text is only created on first access. If it is set to a React element (only by the user),
  // that will not be persisted, and will be restored to default on load.
```

src/Script/RunningScript.ts
```diff
@@ -73,9 +73,7 @@ export class RunningScript {
...
   title = "" as string | React.ReactElement;
```
d0sboots: If we're defaulting to null, we should actually default to it. (And I suggest we use null instead of undefined, as a subtle codebase consistency point.)
```suggestion
  title = null as string | React.ReactElement | null;
```

src/Script/RunningScript.ts
d0sboots: That's *why* it was checking for string - if it's a string, we store it in the save, otherwise it must be a React element so we don't.

With changing this to include null, it should be this (i.e. revert):
```suggestion
      typeof this.title === "string" ? includedProperties : includedPropsNoTitle,
```
We don't need to include the property if it's null, since it will default to that on load (we can save some space).

src/Script/RunningScript.ts
```diff
@@ -167,7 +168,6 @@ export class RunningScript {
     const validEntries = Object.entries(runningScript.dataMap).filter(isValidDataMapEntry);
     runningScript.dataMap = new Map(validEntries);
     if (!runningScript.scriptKey) runningScript.scriptKey = scriptKey(runningScript.filename, runningScript.args);
-    if (!runningScript.title) runningScript.title = `${runningScript.filename} ${runningScript.args.join(" ")}`;
```
d0sboots, line 164: There is a subtle compatibility thing here that I believe actually makes this a bugfix of sorts.

This exists to set a proper title for older saves, or for cases where it had a react element and thus the title was not saved. However, if you set the title to "", that *would* get saved, but that's falsey so it would get overwritten here on load.

Now, all those cases will remain the default of null, and get filled in on-demand, as we want. But a string of "" will be retained.

src/NetscriptFunctions.ts
```diff
@@ -847,10 +847,10 @@ export const ns: InternalAPI<NSFull> = {
     return allFilenames.filter((filename) => ("/" + filename).includes(substring)).sort();
   },
   getRecentScripts: () => (): RecentScript[] => {
-    return recentScripts.map((rs) => ({
-      timeOfDeath: rs.timeOfDeath,
-      ...helpers.createPublicRunningScript(rs.runningScript),
-    }));
+    return recentScripts.map((rs) =>
```
d0sboots: Revert - this doesn't matter IMO

src/Script/RunningScript.ts
```diff
@@ -96,7 +94,10 @@ export class RunningScript {
     this.server = script.server;
     this.ramUsage = ramUsage;
     this.dependencies = script.dependencies;
-    this.title = `${this.filename} ${args.join(" ")}`;
+  }
+
+  getDefaultTitle(): string {
```
d0sboots, line 104: What I was actually expecting was, similar to what you were trying to do with the public version of RunningScript, centralizing responsibility for this behind a getter. Specifically:

* Change the property name to `title_`. (Private would also work.)
* You'll need to massage the name to<->from `title` in the save/load hooks, to avoid breakage. This is the most annoying part, and could be avoided by keeping the name the same, but that comes with other downsides. (Specifically, the risk of future code accidentally referring directly to `title`.)
* Add a getter that initializes `title_` if needed and then returns it.
* A trivial setter is also needed.

src/ui/React/LogBoxManager.tsx
```diff
@@ -94,6 +94,7 @@ export class LogBoxProperties {
 interface Log {
   id: number; // The PID of the script *when the window was first opened*
   script: RunningScript;
+  defaultTitle: string;
```
d0sboots: If you use a getter for title, nothing in LogBoxManager needs to change.

src/Netscript/NetscriptHelpers.tsx
Sphyxis: It's for further down, where it creates and send out a title.  Rather than reference runningScript for the slice - which I think can keep runningScript alive longer than it needs to be.  That's it really.  Just to prevent GC from potentially skipping it.

src/Netscript/NetscriptHelpers.tsx
Sphyxis: If you look at it in the debugger, it is not assigned to anything.  It's just (...)  But when you go to look at the assigned property, that's when it populates.  I'm not sure if that means it cannot be referenced though.  If it does, I'll have to rethink it.

d0sboots: Let me show you what I mean with a quick JS snippet:
```js
let foo = { get title() { return "a" } }
> undefined
foo.title
> 'a'
foo.title = "bar"
> 'bar'
foo.title
> 'a'
```
This is what I meant by this being an API-breaking change. Even if you add a setter, it still differs in subtle ways from an object with a data property.

src/Netscript/NetscriptHelpers.tsx
d0sboots: This is showing a fundamental misunderstanding of how garbage collection works. `runningScript` is one of the arguments to this function, so until this function returns it *definitely* has at least once live reference that prevents it from being collected. If GC were to run in the middle of this function (an unlikely occurrence that we shouldn't be optimizing for anyway), that wouldn't be able to be collected no matter what.

Zooming out a bit further, as long as the script is alive, the RunningScript will be referenced via the RunningScriptsMap. So it definitely won't be collectable in the short-term.

src/Script/RunningScript.ts
Sphyxis: Now that I know the creation on demand does not work properly I'll be changing it to that style.

Sphyxis: No setter or getter needed.  I'm creating the title on request of a publicRunningScript.  If the main script does not have a title, it provides a default.  If it does, it uses that.

Sphyxis:
Let me know if this is a better approach.  There are a few other areas I want to do this too.

src/Netscript/NetscriptHelpers.tsx
Sphyxis: Should be fixed now

src/ui/React/LogBoxManager.tsx
Sphyxis: Not using a getter or setter now.

src/Script/RunningScript.ts
d0sboots: Yes, a getter is not *needed*, but I think you want one (on the internal API). Let me expand.

Encapsulation and reusability are key tenets of OOP. A `RunningScript` has a title which is either a string or a ReactElement; it is merely an *implementation detail* that we wish to store null for the case where the title is the default, to avoid some work and memory up-front. So `RunningScript` should be the one to hand out titles and manage dealing with if the internal value is undefined or not; otherwise, that same logic would have to be duplicated across every site that uses it (which is what you are currently doing).

If you accept that, then the issue becomes how to actually name/implement it. All the possibilities have their own downsides:
* You could call such a function `getTitle()` and leave the existing `title` field alone. However, this causes an asymmetry at call-sites between setting the field (by writing directly) and getting the field (calling the function). More importantly, it is very easy to mess up callers and use `.title` directly, causing issues.
* To solve the problem about, you can hide the `title` field either by renaming to `title_` or making it private. At this point, it becomes very natural to rename `getTitle()` to just `get title()`, i.e. a getter. A setter is also required (regardless of if it's a JS Setter or a separate function) to be able to set the title from outside. This prevents accidental misuse, but has the downside of requiring care with the savegame, as I talked about above.

Because of the issues with misuse, catlover did exactly the getter/setter approach with `content` for `TextFile`s and `ScriptFile`s, which you can use as a guide if you want. However, potentially TS will catch misuse for us since `title`s type is wider than just `string | ReactElement`, so maybe we can get away with the simpler first approach.

Sphyxis:
I think this is closer to what you were wanting.  Due to the possibility of a null value I used the _ version.

src/Script/RunningScript.ts
```diff
@@ -173,7 +187,8 @@ export class RunningScript {
 }
 const includedProperties = getKeyList(RunningScript, {
   removedKeys: ["logs", "dependencies", "logUpd", "pid", "parent", "tailProps"],
-});
+  // Persist the title under its canonical name, keeping its original key position.
+}).map((key) => (key === "title_" ? "title" : key));
 const includedPropsNoTitle = includedProperties.filter((x) => x !== "title");
```
Sphyxis, line 192: Only spot I'm not really sure on.  If it's saved as title it's the same as before, and on load it should populate the title_ field.

d0sboots: Ah, I see. The `fromJSON` operates through the setter now. That's more elegant than what I had been thinking of.

Sphyxis:
Thank you.  I'm trying to delegate authority to RunningScript here.  WorkingScript should look to it for everything possible I think.

PR #2878, opened 2026-06-12
CLEANUP: Remove duplicate scriptKey

Sphyxis:
`RunningScript` currently has a duplicated `scriptKey` entry in it.  This PR aims to reclaim that little bit of memory that the duplicate `scriptKey` is taking up.  This PR does not affect any API's or displayed info in game.

`getRunningScriptsByArgs` is mentioned as needing this.  However, that function eventually calls the helper `scriptKey()`.  It no longer uses the `scriptKey` entry in `RunningScript`.

The memory savings for this are not as large as before.
Example script key: `grow.js*["n00dles",49449.62645586807,"QUIET"]`
If this key is removed from 300k running scripts, it can save around 20-30mb.  If the file is a bit longer, such as `SphyxOS/basic/grow.js`, then it comes to around 27-35mb.

I've run my batcher and darknet script for a while.  Restarted, no issues.

This will create a conflict with my previous PR, which I can easily resolve.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -108,6 +108,15 @@
         "node": ">=24"
       }
     },
+    "node_modules/@aashutoshrathi/word-wrap": {
+      "version": "1.2.6",
+      "resolved": "https://registry.npmjs.org/@aashutoshrathi/word-wrap/-/word-wrap-1.2.6.tgz",
+      "integrity": "sha512-1Yjs2SvM8TflER/OD3cOjhWWOZb58A2t7wpE2S9XfBYTiIl+XFhQG2bjy4Pu1I+EAlCNUzRDYDdFwFYUKvXcIA==",
+      "dev": true,
+      "engines": {
+        "node": ">=0.10.0"
+      }
+    },
     "node_modules/@asamuzakjp/css-color": {
       "version": "3.2.0",
       "resolved": "https://registry.npmjs.org/@asamuzakjp/css-color/-/css-color-3.2.0.tgz",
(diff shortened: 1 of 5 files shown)


Sphyxis:
I did have to update the snapshots.

d0sboots:
Note that this does not save the memory you are claiming. The key itself is also stored in the RunningScriptMap, and thus will still be retained. The only memory that is actually saved here is the 8 bytes for the slot storage in RunningScript. However, if we don't need it, it's still worth removing.

----

Due diligence:
The original reason this was added was to cache the calculation. By computing it always, in the common case we are calling scriptKey one more time (once at start and once at death, instead of only once at start). I benchmarked to see if this had an adverse performance impact:
```js
/** @param {NS} ns */
    const perf = globalThis.performance;
    let runTime = 1;  // Slow start, assume 1ms time
    let speedMax = 0, p;
    // Do n iterations
    for (let i = 0; i < (Number(ns.args[0]) || 10); ++i) {
      const start = perf.now();
      // Warm up in 200ms chunks
      let end;
      let cycles = 0;
      for (let j = 0; j < 5; ++j) {
        const iters = (cycles + (200 / runTime)) | 0;
        for (; cycles < iters; ++cycles) {
          p = ns.run("lib/noop.js", "a", "b", "c", "d", "e", "f");
        }
        end = perf.now();
        // Add 0.05 because of minimum timing precision
        runTime = (end + 0.05 - start) / (200 * (j+1));
      }
      const speed = cycles * 1000 / (end - start);
      speedMax = speed > speedMax ? speed : speedMax;
      ns.tprintf("Iter %2d: %.3f/sec", i, speed);
      await ns.sleep(0);
    }
    // Force result to be used
    ns.tprintf("Best: %.3f/sec", speedMax + p * 1e-21);
}
```
lib/noop.js:
```js
export function main() {}
```
Performance was unchanged, or if anything, slightly better by removing the key. This is not completely surprising.

Issue #2741, opened 2026-05-06
CodingContractSignature error for "Square Root"


fluxxdog:
https://github.com/bitburner-official/bitburner-src/blob/a7b34b110c6a684e9ab3d125881d150f7aa64f16/markdown/bitburner.codingcontractsignatures.md?plain=1#L39

This odd bit came up while reading through some of the NetscriptDefinitions.d.ts file. All the other signatures have a 2 part tuple, with [0] showing the expected data type given by the contract and [1] showing the expected datatype for the answer. But the line for Square Root is 3 part. The contract itself says it will provide a bigint and it expects a string so it should read ```[bigint, string]```

d0sboots:
The signature isn't actually wrong, but it is confusing. More on this when I'm at a keyboard.

catloversg:
The signature contains 3 parts:
- Data: The type of data you get from `ns.codingcontract.getContract().data` or `ns.codingcontract.getData()`.
- Answer: The expected answer type.
- State: The type of the internal data structure. State is usually the same as Data, but not always.

In this case, `[bigint, bigint, [string, string]]` means:
- The input is a bigint.
- The expected answer is a bigint.
- The internal data structure, which is irrelevant to players, is a string array.

There are 3 issues here:
- The answer is expected to be a bigint, but the description makes it sound like players should submit a string.[1]
- The signature documentation does not clearly explain the meaning of the three parts.[2]
- The internal data structure should not be exposed in public APIs.[3]

[1]: The description was written that way to help players avoid submitting an invalid string answer through the UI (many players mistakenly submitted a string containing the trailing "n"). I'll update the description of this cct and `codingcontracts.md`.

[2]: Doc update.

[3]: I'm not sure how to do this in an elegant way. This is not urgent, so I'll keep an eye on it.

d0sboots:
Technically the internal data structure is a pair containing two strings. But yes, thank you for the correct explanation that is more complete than the one I forgot to return to my keyboard to write. :)

PR #2881, opened 2026-06-13
DOCUMENTATION: Remove internal state type definitions and clarify CodingContractSignatures

catloversg:
This PR implements [2] and [3] at https://github.com/bitburner-official/bitburner-src/issues/2741#issuecomment-4414669163.


src/CodingContract/ContractTypes.ts
```diff
@@ -68,7 +68,16 @@ type CodingContractDefinitions<Signatures extends Record<string, [unknown, unkno
     ? CodingContractComplexType<Signatures[T][0], Signatures[T][1], Signatures[T][2]>
     : CodingContractSimpleType<Signatures[T][0], Signatures[T][1]>;
 };
-export type CodingContractTypes = CodingContractDefinitions<CodingContractSignatures>;
+
+type CodingContractStates = {
```
d0sboots: Please add a brief comment here about what this is doing and (more importantly) why it has to be done for Square Root in particular.

PR #2882, opened 2026-06-13
DEPS: Update @electron/packager

catloversg:
On Node 24.16.0 and Node 26.3.0, electron-packager cannot package the Electron version due to an issue with an old version of yauzl used by extract-zip, which is a dependency of electron-packager. Updating electron-packager fixes this issue.

Ref:
- https://github.com/electron/electron/issues/51619
- https://github.com/electron/packager/pull/1917

electron-packager has enabled ASAR by default since v19, so we need to configure it as well.

Tested on Windows 10 and Debian 13.

### DIFF
--- a/tools/package-electron.sh
+++ b/tools/package-electron.sh
@@ -16,15 +16,15 @@ cp -r .app/* .package
 cp -r electron/* .package
 
 packageWin() {
-  electron-packager .package bitburner --platform win32 --arch x64,arm64 --out .build --overwrite --icon .package/icon.ico --app-copyright "Copyright (C) 2025 Bitburner"
+  electron-packager .package bitburner --platform win32 --arch x64,arm64 --out .build --asar.unpack='**/node_modules/@catloversg/steamworks.js/**/*.{dll,node}' --overwrite --icon .package/icon.ico --app-copyright "Copyright (C) 2025 Bitburner"
 }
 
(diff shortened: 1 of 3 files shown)


tools/package-electron.sh
```diff
@@ -16,15 +16,15 @@ cp -r .app/* .package
 cp -r electron/* .package
 
 packageWin() {
-  electron-packager .package bitburner --platform win32 --arch x64,arm64 --out .build --overwrite --icon .package/icon.ico --app-copyright "Copyright (C) 2025 Bitburner"
+  electron-packager .package bitburner --platform win32 --arch x64,arm64 --out .build --asar.unpack='**/node_modules/@catloversg/steamworks.js/**/*.{dll,node}' --overwrite --icon .package/icon.ico --app-copyright "Copyright (C) 2025 Bitburner"
```
d0sboots, line 19: I know you'd like to keep this PR small, but the level of duplication here is getting to the point where I think it would make sense to pull most of this commandline out into one place. The `asar` line is a bit tricky since it varies per-arch, but ideally most of it would be shared, I don't have a great idea on that.

d0sboots:
Although I would like to see my comment addressed, at the same time since this blocks building with latest node, I also wanted to get the fix in ASAP.

PR #2883, opened 2026-06-13
OPTIMIZATION: Script.code interning and transform caching

Sphyxis:
This is a proof of concept for the benefits of interning large, reused strings such as Script.code.

My tests on darknet, which is my largest script, is as follows:

Darknet - big script - big savings over ~180 darknet servers
Observed over time
(dev) - stable at around 720-820mb. Jumps to 1150mb at times.
(fix) - stable at around 280-420mb. Jumps to 770mb at times.

All you need to do is test it with a large number identical scripts all running on their own server.  Since each Script instance has it's own .code field, every script that is started from that instance essentially shares the same .code, but it adds up over 180 darknet servers.

### DIFF
--- a/src/NetscriptJSEvaluator.ts
+++ b/src/NetscriptJSEvaluator.ts
@@ -93,7 +93,7 @@ function generateLoadedModule(script: Script, scripts: Map<ScriptFilePath, Scrip
     case FileType.JSX:
     case FileType.TS:
     case FileType.TSX:
-      ({ scriptCode, sourceMap } = transformScript(script.code, fileType));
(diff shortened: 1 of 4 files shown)


d0sboots:
As discussed on Discord, the main issue is savegame loading, so we'll focus on deduplication at the save/load level, which will translate into the memory savings that this PR is aiming for. However, if another category of duplication pops up that is independent of save/load, we can dust this off.

Sphyxis:
I'm just glad something I put together conveyed my thoughts on what my problem was.  I'll keep it around, but once the filesystem is itself deduped on save and load I don't think it will be needed.

PR #2888, opened 2026-06-17
UI: Follow up to #2886

catloversg:
I noticed some bugs and issues in #2886, but I hadn't finished my review before it was merged.

### The "Active Scripts" page is inaccessible.

MRE: Click on that sidebar item:
```
Uncaught Error: Can't handle click on Page Active Scripts
    at eval (SidebarRoot.tsx:235:13)
    at first.<computed> (SidebarAccordion.tsx:40:32)
    at HTMLUnknownElement.callCallback (react-dom.development.js:3945:14)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:3994:16)
    at invokeGuardedCallback (react-dom.development.js:4056:31)
    at invokeGuardedCallbackAndCatchFirstError (react-dom.development.js:4070:25)
    at executeDispatch (react-dom.development.js:8243:3)
    at processDispatchQueueItemsInOrder (react-dom.development.js:8275:7)
    at processDispatchQueue (react-dom.development.js:8288:5)
    at dispatchEventsForPlugins (react-dom.development.js:8299:3)
```

### Incorrect max number of icons

Capture

There are 3 icons here.

https://github.com/bitburner-official/bitburner-src/blob/0ebefa2c4591cb09580560a1f9c7f13ce04d3768/src/DarkNet/ui/ServerSummary.tsx#L161
This should be 3, not 2. Notice how `componentsToShow` is changed in #2886.

### Use pointer cursor when hovering the icon of running scripts

The UI should indicate that this icon is clickable; otherwise, nobody knows this feature exists.

### Change icon order

> The script icon on darknet server summaries is now always on the right side, to maintain consistency in finding it for quickly checking script count or getting to the Active Scripts page for that server

The goal is good (maintaining consistency), but the implementation is not (putting the icon on the right).

This does not look too bad.

However, without reading #2886's description, most people would think this is a bug. The icon looks like its left margin is set to a random large value.

Bitburner UI is left-to-right in most places (except for some popups in IPvGO; they weirdly place the "main" button on the right, but it's a separate issue), so the icon that is always present should be on the left, not the right.

### DIFF
--- a/src/DarkNet/ui/ServerSummary.tsx
+++ b/src/DarkNet/ui/ServerSummary.tsx
@@ -70,7 +70,7 @@ export function ServerSummary({
   const ramBlocked = showDetails ? ramBlockedDetails : formatNumber(server.blockedRam, 0);
 
   const components = [
-    <Tooltip key="runningScript" title={<>{runningScriptsTooltip}</>}>
+    <Tooltip key="runningScript" title={<>{runningScriptsTooltip}</>} style={{ cursor: "pointer" }}>
       <Typography
         color={runningScriptCount > 0 ? "primary" : "secondary"}
         onClick={() => Router.toPage(ComplexPage.ActiveScripts, { serverName: server.hostname })}
@@ -158,13 +158,11 @@ export function ServerSummary({
       </Tooltip>,
     );
   }
-  const maxIcons = showDetails ? components.length : 2;
+  const maxIcons = showDetails ? components.length : 3;
   const componentsToShow = components.slice(0, maxIcons);
 
   return (
-    <div
-      style={{ display: "inline-flex", flexDirection: "row-reverse", width: "100%", justifyContent: "space-between" }}
-    >
+    <div style={{ display: "inline-flex", flexDirection: "row", width: "100%", justifyContent: "space-between" }}>
       {componentsToShow}
     </div>
   );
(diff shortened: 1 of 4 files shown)


src/DarkNet/ui/ServerSummary.tsx
```diff
@@ -158,13 +158,11 @@ export function ServerSummary({
...
   }
-  const maxIcons = showDetails ? components.length : 2;
+  const maxIcons = showDetails ? components.length : 3;
   const componentsToShow = components.slice(0, maxIcons);
 
   return (
-    <div
-      style={{ display: "inline-flex", flexDirection: "row-reverse", width: "100%", justifyContent: "space-between" }}
-    >
+    <div style={{ display: "inline-flex", flexDirection: "row", width: "100%", justifyContent: "space-between" }}>
```
d0sboots, line 165: This reverts the purpose of fico's change to have the active scripts in a consistent spot. I'm going to merge this as-is, because I was also a bit leary of switching the flexDirection and your screenshot showed that I should have paid more attention to that instinct, but probably something else will need to be done here.

d0sboots:
Thank you for being thorough, as always. I'm too lazy to manually test each PR ;_;

ficocelliguy:
Apologies for not testing the prior PR thoroughly enough - I should have caught most of these myself.

src/DarkNet/ui/ServerSummary.tsx
catloversg: > This reverts the purpose of fico's change to have the active scripts in a consistent spot.

The icon is still in a consistent spot. The difference is that it's consistently on the left, not on the right.

Before:





After:





If it's on the right, we have this:

PR #2889, opened 2026-06-17
SAVE: Change save format for significantly better memory and file efficiency

d0sboots:
TODO: Write a full description

This introduces a general method for deduping strings in the savefile, and uses it for several known sources of repeats: files, hostnames and filenames. It can be trivially extended to more strings as needed (at the cost of a save format bump).

The code here is working. Tests are still in progress.

Tested with using this as the baseline, when creating many scripts it is 300KB on dev and 16KB with this change. (It's pretty impressive how small it gets, given that it always has at least 1 1MB script.) Loading on dev is ~2GB, vs no real movement with this change.
[bitburnerSave_1781691382_BN15x3_manyScripts_base.json.gz](https://github.com/user-attachments/files/29044288/bitburnerSave_1781691382_BN15x3_manyScripts_base.json.gz)


Sphyxis:
I'm only seeing the strings table in the save data.  Could you make it persistent and destroyed/recreated on save?  That way, I could make a helper that adds a read only reference from it to items that match it, like hostnames in arguments.  Keeping it alive shouldn't really take up much memory at all due to everything being a reference anyway.

Sphyxis:
Specifically, using something like this:
```js
//Strings map instead
const namePool = new Map<string, string>();

/** Return the canonical instance if pooled; otherwise return the original */
export function dedupeIfInterned<T extends string>(value: T): T {
  const existing = namePool.get(value);
  return existing !== undefined ? (existing as T) : value;
}
```

d0sboots:
I was actually going to add deduping for strings in arguments as part of this PR already. The only slight complication is because numbers are a valid type for args, so it's a ~10 line change instead of a 5 line change.

The explicit design goal of this approach is to *not* have any persistent maps outside of the save/load process. IMO they are unnecessary and only serve as a potential source of bugs and/or memory leaks. If we tried to make this permanent, it would because *radically* more complicated, because you would have to track ownership info for everything in the table, as well as dynamically juggling index values - and this is all stuff that the garbage collector already handles for us.

Sphyxis:
I'm just saying, let it persist across the load and save functions.  Let it be recreated in those processes.  The most important thing is already there - the ability to see if a string exists in pool and assign it as a read only ref to something.  The pool can come and go, but it needs to be there to make this assignment.  Just have it contain the string as the key and as the value so it can be used as a reference.  Let it be rebuilt, it's not used for real time tracking this way.

d0sboots:
If you're rebuilding it on every save, there is no reason to keep it around past that point. Negative reason even, since then it's occupying memory that it doesn't need to.

I've implemented deduping for script args. It ended up being more like 40 lines, not the 10 I thought. I also did deduping for the filenames that are stored in the keys of the JSONMaps.

Here is the script I am using to test `RunningScript` and args-related things:
```js
/** @param {NS} ns */
  ns.disableLog("ALL");
  const source = "export function main(ns) {return ns.sleep(5000)}"
  ns.write("wait5.js", source, "w");
  let launched = 0;
    launched += ns.run("wait5.js", 1, source, source, 12, source, -1, "0", source, source, source, "ablab") != 0;
  }
}
```

I profiled launching both 10000 and 100000 scripts. With 10000, the compressed savefile is 37kb, uncompressed is 4.7MB. This is from the many RunningScript copies in the data, plus the 1MB script from the other test that's still present.

For 100000 scripts, I profiled save/load performance.
image
Launching the scripts takes ~5 seconds, split across ~770ms to actually launch them all, 140ms to print results via `tprint`, and 4160ms to asynchronously start all the tasks after they're been launched.
Saving the game in this state takes ~900ms, split across 380ms to run all toJSON impls, 312ms to do raw JSON encoding, and 208ms for compression.

Then I switched back to head and profiled the same thing. Launching the scripts takes roughly the same time, but saving is up to 1200ms, with about 450ms in toJSON impls and 366ms in compression. This is only one sample, but shows that (at minimum) the increase in processing is offset by the gains from encoding less data, when dealing with lots of repeated data (the common case for large savegames).

d0sboots:
The tests should now be fixed. There was a really subtle issue with a missing Blob implementation in our test utils.

catloversg:
The idea and general structure look good, but I have only skimmed through the changes so far. I'll try to review the implementation this week.

I found a bug. MRE:
- Load a clean save file.
- Create a corp.
- Expand into an industry.
- Save.
```
Cannot get save data. Error: Error: Tried to serialize game objects outside of save().
Stack:
Error: Tried to serialize game objects outside of save()
    at stringDataIdx (webpack://bitburner/./src/utils/JSONContext.ts?:20:11)
    at JSONMap.toJSON (webpack://bitburner/./src/Types/Jsonable.ts?:32:85)
    at JSON.stringify (<anonymous>)
    at getSaveData (webpack://bitburner/./src/SaveObject.ts?:155:26)
    at saveGame (webpack://bitburner/./src/SaveObject.ts?:212:22)
    at save (webpack://bitburner/./src/ui/GameRoot.tsx?:644:60)
```

d0sboots:
> I found a bug. MRE:
>
> * Load a clean save file.
> * Create a corp.
> * Expand into an industry.
> * Save.

I clearly didn't manually test this week enough, and also our automated tests are obviously lacking here.

Don't bother manually testing further atm, there is a *big* class of bugs like that based on my naive assumption that serializing Player was "simple", as opposed to where a lot of the data is (like Corp).

I'll figure out what to do and add more tests, but it shouldn't affect most of the rest of the code.

Sphyxis:
When you are ready for testing I'll throw everything I have at it.  Just
post here or DM me in some way.

PR #2890, opened 2026-06-17
CLEANUP: Removing unneeded items from WorkerScript

Sphyxis:
WorkerScript already has access to the hostname property from its RunningScript.  The only real game interaction I found was in renaming servers.  I ran a simple script below and got the expected output.  The new server name was returned when I got the running script by pid.

Test Script:
```js
/** @param {NS} ns */
  //Give yourself lots of money first.
  const hostname = "test-"
  const startingRam = 256
  const filename = "workerscript_hostname_stability_test.js"

  ns.write(filename, testFile, "w")

  for (const server of ns.cloud.getServerNames()) {
    ns.cloud.deleteServer(server)
  }

  let count = 0
  while (ns.cloud.purchaseServer(hostname + String(count >= 10 ? count++ : "0" + String(count++)), startingRam) !== "") { }

  for (const server of ns.cloud.getServerNames()) {
    ns.scp(filename, server)
    ns.exec(filename, server)
  }

  await ns.sleep(100)

  for (const server of ns.cloud.getServerNames()) {
    ns.cloud.renameServer(server, server + "-rename")
  }

  await ns.sleep(100)

  for (const server of ns.cloud.getServerNames()) {
    const processes = ns.ps(server)
    for (const proc of processes)
      ns.tprintf(ns.getRunningScript(proc.pid).server)
  }
}
const testFile = `
  ns.disableLog("ALL");
  await ns.sleep(20000)
}`;
```

### DIFF
--- a/src/Netscript/WorkerScript.ts
+++ b/src/Netscript/WorkerScript.ts
@@ -54,11 +54,10 @@ export class WorkerScript {
    */
   loadedFns: Record<string, boolean> = {};
 
-  /** Filename of script */
-  name: ScriptFilePath;
-
-  /** Script's output/return value. Currently not used or implemented */
-  output = "";
+  /** Filename of script. Mirrors the RunningScript, so we don't store a second copy. */
+  get name(): ScriptFilePath {
+    return this.scriptRef.filename;
+  }
 
   /**
    * Process ID. Must be an integer. Used for efficient script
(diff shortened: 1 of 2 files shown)


d0sboots:
FWIW, I don't think this will be any significant amount of memory savings, but I do support it for code cleanliness reasons.

Sphyxis:
I agree.  No memory savings really, just a cleanup.

d0sboots:
lol i just keep making conflicts

Sphyxis:
I knew this would happen when I decided to do them in small bites.

Sphyxis:
Crap.  It kept the wrong thing

PR #2892, opened 2026-06-17
REWORK: lazy assign disableLogs on every script to save memory

Sphyxis:
This only creates the disableLogs object on each WorkerScript if it needs to.  Saving ~10mb for 300k WorkerScripts.  It will only create it when a specific disableLog("specific") is called.  If ALL is specified, it still gets the reference object and skips the disabling of all other logs.

Tested by running some scripts with disableLogs("ALL"), and just for "exec".  No issues.

### DIFF
--- a/src/Netscript/WorkerScript.ts
+++ b/src/Netscript/WorkerScript.ts
@@ -32,8 +32,8 @@ export class WorkerScript {
   /** Holds the Promise reject() function while the script is "blocked" by an async op */
   delayReject: ((reason?: ScriptDeath) => void) | undefined = undefined;
 
-  /** Stores names of all functions that have logging disabled */
-  disableLogs: Record<string, boolean> = {};
+  /** Stores names of all functions that have logging disabled. Allocated lazily; null means none disabled. */
+  disableLogs: Record<string, boolean> | null = null;
 
   /**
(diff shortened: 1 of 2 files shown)


d0sboots:
I don't think this one is worth it. The usecase of a script that calls *zero* `ns` functions is basically nonexistent. As a result, this is optimizing for a scenario that just doesn't happen. (Unlike the others, such as "no atExist use" or "doesn't open a tail window," which are pretty common cases, especially for high-volume scripts.)

Sphyxis:
If you look at how `ns.disableLog("ALL")` works it is.  If you do not call any ns function that write to the log, or if your first call is `ns.disableLog("ALL")`, it doesn't create a new object, it references the default disableLogsAll (or whatever it's called) static entry.  Meaning, all scripts using `ns.disableLogs("ALL")` will automatically share a reference and receive this memory benefit.

d0sboots:
If your first call is to `ns.disableLogs("ALL")`, the empty object is immediately replaced by the `allDisabled` shared reference. Meaning they don't incur any permanent memory usage from the temporary new object. In terms of how many scripts you can run at once, it should be a no-op.

Sphyxis:
True.  I'll close this one.

PR #2894, opened 2026-06-17
FIX: Cover move RecoveryRoot cases with the newer version check

d0sboots:
In cases of save data changes, the errors we throw can be varied - and SaveObject itself has lots of paths that just throw a plain Error. This broadens the test to check for newer versions first, regarless of Error type.

Here is what loading the data from my other PR looks like now, if I artificially drop the current version to emulate a release:
image

Unfortunately, this won't make it in soon enough to help with beta testers for *this* release, but it will help going forward.

### DIFF
--- a/src/ui/React/RecoveryRoot.tsx
+++ b/src/ui/React/RecoveryRoot.tsx
@@ -109,14 +109,8 @@ export function RecoveryRoot({ softReset, crashReport, resetError }: IProps): Re
         Please update your browser.
       </Typography>
     );
-  } else if (sourceError instanceof InvalidSaveData) {
-    instructions = (
-      <Typography variant="h4" color={Settings.theme.warning}>
-        Your save data is invalid. Please import a valid backup save file.
-      </Typography>
-    );
   } else if (
-    (sourceError instanceof JSONReviverError && isSaveDataFromNewerVersions(loadedSaveObjectMiniDump.VersionSave)) ||
+    isSaveDataFromNewerVersions(loadedSaveObjectMiniDump.VersionSave) ||
     sourceError instanceof IndexedDBVersionError
   ) {
     instructions = (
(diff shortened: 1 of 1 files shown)


src/ui/React/RecoveryRoot.tsx
```diff
@@ -109,14 +109,8 @@ export function RecoveryRoot({ softReset, crashReport, resetError }: IProps): Re
         Please update your browser.
       </Typography>
     );
-  } else if (sourceError instanceof InvalidSaveData) {
```
catloversg, line 112: Moving this check to after the version check is not exactly wrong, but I don't think it helps much.

`InvalidSaveData` is only thrown when the save data is "unreadable" or "unparsable" (e.g., null, undefined, empty string, invalid json string, or there is an error when decompressing). This happens before any code that parses the save data. The only exception is `sourceError instanceof IndexedDBVersionError`, which can happens before everything else.

src/ui/React/RecoveryRoot.tsx
```diff
@@ -109,14 +109,8 @@ export function RecoveryRoot({ softReset, crashReport, resetError }: IProps): Re
...
-    instructions = (
-      <Typography variant="h4" color={Settings.theme.warning}>
-        Your save data is invalid. Please import a valid backup save file.
-      </Typography>
-    );
   } else if (
-    (sourceError instanceof JSONReviverError && isSaveDataFromNewerVersions(loadedSaveObjectMiniDump.VersionSave)) ||
```
catloversg, line 119: Can you explain a bit more about why you moved this check? Aside from bad save editing, `JSONReviverError` can only be thrown when loading newer save data. Moving it into the same check as `InvalidSaveData` does not make sense to me.

d0sboots: If `JSONReviverError` is thrown due to loading newer save data, it will be caught by the `isSaveDataFromNewerVersions` check - the same check, just now broader.

Otherwise, it will be caught lower, and print the "Invalid save data" line. Presumably, that case is due to bad save editing, as opposed to a mistake on our part as devs.

src/ui/React/RecoveryRoot.tsx
d0sboots: It's possible its use could be expanded in the future. Unlikely, but still. Semantically, it seems more correct to me that the version check come first as a catch-all for "if it's from a newer version, all bets are off." Only otherwise, can we make reasonable guesses about what the various exceptions *actually* mean.

src/ui/React/RecoveryRoot.tsx
catloversg: Optional: Add a comment in the `sourceError instanceof JSONReviverError` block documenting the rationale for this check. For example:
```js
    // JSONReviverError can only be thrown when the save data is from a newer version, which is already checked in the
    // previous if statement, or when the save data is corrupted by bad save editing.
```

d0sboots: I took the opportunity to comment on the branches in general.

PR #2895, opened 2026-06-17
BASE: Preemptive update for npm security changes

mctylr-gh:
This is a preemptive update to opt-in to future
[npm v12 security-related changes](
https://github.blog/changelog/2026-06-09-upcoming-breaking-changes-for-npm-v12/).
Which is expected to occur in
July 2026. Users currently with npm v11.16.0 and higher will notice
`npm` warnings about `--allowScripts` on packages with scripts for
pre-, during, and post-install.

For Bitburner that includes:

*  `@swc/core` (postinstall: node postinstall.js)
*  `core-js-pure` (postinstall: node -e "try{require('./postinstall')}catch(e){}")
*  `electron` (postinstall: node install.js)

This commit adds those scripts in those package to the `package.json`
as the only scripts allowed to be executed by `npm install`.

Since we don't use, I have not enabled `--allow-git`, `--allow-file`,
or `--allow-remote`.

This is to reduce our project's (and individual contributors) exposure
to various `npm` and **GitHub** supply-chain attacks such as the
on-going *Shai-Hulud* and *TeamPCP* [related attacks](
https://www.securityweek.com/npm-12-will-change-script-execution-behavior-to-prevent-supply-chain-attacks/),
as we do run CI/CD runner scripts on GitHub triggered on contributor's
pull request submissions.

Note: Only commit 5979ec9 is only change. I didn't rebase my personal fork with an accidental (local) automated pull-request merge, and my attempt to revert the merge. So the merges are *dumb* fast-forwards for the bitburner `origin` updates.


package.json
```diff
@@ -145,5 +145,10 @@
     "version": "sh ./tools/build-release.sh && git add --all",
     "postversion": "git push -u origin dev && git push --tags",
     "changelog": "node tools/fetch-changelog/index.js --from=$(cat last_changelog_hash) > changelog.md"
+  },
```
d0sboots: A minor thing, but since this is under our control, and we do hand-edit this file, let's move this to right after `overrides`, so it's next to all the other dependency-related info.

mctylr-gh: Okay.

catloversg:
Among these packages, we should only allow scripts from `electron`.
- `electron`: This package is only used when running `npm run start`, but this command is rarely useful. Nonetheless, I don't mind keeping this package and `npm run start` as-is at the moment. The postinstall script is necessary in this case.
- `core-js-pure`: The postinstall script only conditionally prints a banner.
- `@swc/core`: The postinstall script validates `@swc/core`'s native binary installation and falls back to `@swc/wasm` if necessary. This provides no benefit for us.

catloversg:
Optional: Explicitly set those two to false. It makes our intention clear and suppresses the warnings.

PR #2896, opened 2026-06-18
REWORK: Collapse environment wrapper into WorkerScript

Sphyxis:
The env property in WorkerScripts does not need to be there.  While the data inside of it is needed, the wrapper is not.  Removing the wrapper will save a little bit of memory (~7mb for 300k scripts) and GC churn - as currently you have to create a `new Environment()` for every WorkerScript.  It also comes with the benefit of removing the entire class for Environment, and the file associated with it, as it's only used in WorkerScript and when creating a dummy WorkerScript.

This does affect WorkerScripts in a base way, so I wrote up a test for the 5 issues I could see.  It's a bit ugly, but everything seems to pass.
```js
/**
 * Verifies the behaviours touched by collapsing Environment into WorkerScript:
 *   1. NS binding (vars)         2. Kill / stopFlag        3. runningFn cleared
 *   4. Concurrency guard         5. Parent/child exp transfer
 *
 * @param {NS} ns
 */
  ns.disableLog("ALL");
  ns.ui.openTail();
  ns.ui.setTailTitle("Environment-merge test  (running...)");
  ns.clearLog();

  const PORT = 1;
  const CHILD = "env-merge-test-child.js";

  // One helper child
  const CHILD_SRC = `/** @param {NS} ns */
  const mode = ns.args[0];
  const port = ns.args[1];

  if (mode === "loop") {
    while (true) await ns.sleep(1000);

  } else if (mode === "clear") {
    let ok = false;
    ns.atExit(() => ns.writePort(port, "clear:" + (ok ? "PASS" : "FAIL")));
    await ns.sleep(50);
    await ns.sleep(50);
    ok = true;

  } else if (mode === "concurrency") {
    let detected = true;
    ns.atExit(() => ns.writePort(port, "concurrency:" + (detected ? "PASS" : "FAIL")));
    ns.sleep(1000); // sets runningFn (intentionally not awaited)
    ns.sleep(1000); // should throw synchronously -> killed -> atExit reports
    detected = false; // reached only if the guard failed

  } else if (mode === "exp") {
    ns.nuke("n00dles");
    await ns.grow("n00dles");
    ns.writePort(port, "exp:" + ns.getRunningScript().onlineExpGained);
  }
}
`;
  ns.write(CHILD, CHILD_SRC, "w");

  const log = (m) => ns.print(m);
  const results = [];
  let n = 0;
  const start = (name) => { n++; log(""); log("[" + n + "/5] " + name); };
  const done = (pass, detail) => {
    log("      => " + (pass ? "PASS" : "FAIL") + (detail ? "  (" + detail + ")" : ""));
    results.push(!!pass);
  };

  // Poll the port (destructive read) until a message with the given prefix shows up.
  // onTick() fires ~every 2s so slow waits still look alive in the log.
  const waitForMsg = async (prefix, timeoutMs, onTick) => {
    const end = Date.now() + timeoutMs;
    let lastTick = Date.now();
    while (Date.now() < end) {
      const v = ns.readPort(PORT);
      if (typeof v === "string" && v.startsWith(prefix)) return v;
      if (onTick && Date.now() - lastTick > 20000) { onTick(); lastTick = Date.now(); }
      await ns.sleep(200);
    }
    return null;
  };

  ns.clearPort(PORT);

  // 1) NS binding (vars)
  start("NS binding (vars)");
  {
    const host = ns.getHostname();
    log("      ns.getHostname() = " + host);
    done(typeof host === "string" && host.length > 0);
  }
  await ns.sleep(500);

  // 2) Kill aborts a blocked script (stopFlag)
  start("Kill aborts blocked script (stopFlag)");
  {
    log("      launching victim (sleeps forever)...");
    const pid = ns.run(CHILD, 1, "loop", PORT);
    if (!pid) done(false, "could not launch victim");
    else {
      await ns.sleep(400);
      const before = ns.isRunning(pid);
      log("      running before kill: " + before);
      log("      ns.kill(pid)...");
      ns.kill(pid);
      await ns.sleep(400);
      const after = ns.isRunning(pid);
      log("      running after kill:  " + after);
      done(before && !after);
    }
  }
  await ns.sleep(500);

  // 3) runningFn cleared after an async op (run in a child so a regression can't kill us)
  start("runningFn cleared after async op");
  {
    ns.clearPort(PORT);
    log("      child does two sequential awaited sleeps...");
    const pid = ns.run(CHILD, 1, "clear", PORT);
    if (!pid) done(false, "could not launch child");
    else {
      const msg = await waitForMsg("clear:", 8000);
      log("      child reported: " + (msg || "(timeout)"));
      done(msg === "clear:PASS");
    }
  }
  await ns.sleep(500);

  // 4) Concurrency guard (runningFn)
  start("Concurrency guard (runningFn)");
  {
    ns.clearPort(PORT);
    log("      child fires two un-awaited blocking calls...");
    const pid = ns.run(CHILD, 1, "concurrency", PORT);
    if (!pid) done(false, "could not launch child");
    else {
      const msg = await waitForMsg("concurrency:", 8000);
      log("      child reported: " + (msg || "(timeout)"));
      done(msg === "concurrency:PASS");
    }
  }
  await ns.sleep(500);

  // 5) Parent gains child's exp on finish (parent.stopFlag check)
  start("Parent gains child exp on finish");
  {
    ns.clearPort(PORT);
    const before = ns.getRunningScript().onlineExpGained;
    log("      parent exp before: " + before);
    log("      launching child to grow('n00dles') (this can take a bit)...");
    const pid = ns.run(CHILD, 1, "exp", PORT);
    if (!pid) done(false, "could not launch child");
    else {
      const t0 = Date.now();
      const tick = () => log("      ...still growing (" + Math.round((Date.now() - t0) / 1000) + "s)");
      const msg = await waitForMsg("exp:", 1200000, tick);
      const end = Date.now() + 5000;
      while (ns.isRunning(pid) && Date.now() < end) await ns.sleep(100);
      await ns.sleep(150);
      const after = ns.getRunningScript().onlineExpGained;
      const delta = after - before;
      log("      child reported: " + (msg || "(timeout)"));
      log("      parent exp after:  " + after + "   (delta " + String(delta) + ")");
      done(!!msg && delta > 0);
    }
  }

  const passed = results.filter(Boolean).length;
  const allPass = passed === results.length;
  log("");
  log("========================================");
  log("  RESULT: " + passed + "/" + results.length + " passed  ->  " + (allPass ? "ALL PASS" : "SOME FAILED"));
  log("========================================");
  ns.ui.setTailTitle("Env-merge test: " + (allPass ? "ALL PASS" : passed + "/" + results.length));

  ns.rm(CHILD);
  ns.tprint(allPass ? "Environment-merge test: ALL PASS (see log window)" : "Environment-merge test: " + passed + "/" + results.length + " passed (see log window)");
}
```


src/Netscript/WorkerScript.ts
```diff
@@ -41,8 +40,14 @@ export class WorkerScript {
...
 
-  /** Netscript Environment for this script */
-  env: Environment;
+  /** Whether or not this script is stopped */
+  stopFlag = false;
+
+  /** The currently running function */
+  runningFn = "";
+
+  /** Netscript API bound to this script (the script's environment variables) */
```
d0sboots:
```suggestion
  /** Netscript API bound to this script */
```
The extra bit isn't correct. I don't know how to add more and have it still be correct, without adding a *lot* more. (An instance of the NS API is an NSProxy that dynamically binds functions to a set of helpers to ensure dynamic ram usage and concurrency limits remain enforced.)

d0sboots:
Environment always bothered me, but never enough to do something about.

PR #2897, opened 2026-06-18
REWORK: Caching Transforms

Sphyxis:
Currently, whenever a script that needs to be transformed is run, it always goes through the transform process.  It then uses that transformed scriptCode to look up it's module.  This can cause some churn as even though the return values from the transform can be GC'd, it still goes through the process of actually transforming the code every time.

This expands the moduleCache to keep everything in one place.  It adds a transform map to it along with the modules map.  The newly added map is keyed by filetype along with the source code to ensure that we get the proper transform for the various file types.

The consequences of this is that the transforms for any file will be kept around until game close.  Since these transforms are needed anyways for your scripts to run, it's likely not going to be an issue unless you are doing something intentionally destructive.


src/NetscriptJSEvaluator.ts
```diff
@@ -33,12 +33,32 @@ export const config = {
...
+// Cache transforms by exact source and file type, and LoadedModules by the
+// dependency-resolved code, so both can be reused across servers and files.
 // Modules can never be garbage collected by Javascript, so it's good to try
 // to keep from making more than we need.
-const moduleCache = new Map<string, LoadedModule>();
+type CachedTransform = ReturnType<typeof transformScript>;
+
+const moduleCache = {
+  modules: new Map<string, LoadedModule>(),
+  transforms: new Map<string, Map<FileType, CachedTransform>>(),
```
d0sboots: I think this is more complicated than it needs to be. I also think it's a bit memory-wasteful; `Map` instances aren't that small, and having a new one for every source string seen strikes me as a bit wasteful.

Much simpler and also smaller I believe is what I proposed in the thread earlier: use the existing cache, with a key formed by a combination of extension + pre-transform code. This will make for less per-script memory overhead, and (more importantly) it's far simpler to deal with. All it relies on is the guarantee that transforms are idempotent, which we're already more-or-less relying on.

src/NetscriptJSEvaluator.ts
```diff
@@ -93,7 +100,9 @@ function generateLoadedModule(script: Script, scripts: Map<ScriptFilePath, Scrip
     case FileType.JSX:
     case FileType.TS:
     case FileType.TSX:
-      ({ scriptCode, sourceMap } = transformScript(script.code, fileType));
+      transformCacheKey = `\0${fileType}\0${script.code}`;
```
Sphyxis, line 103: The `\0` is not a valid character allowed in the game (I think).  You have to modify your save file to get it.

src/NetscriptJSEvaluator.ts
Sphyxis: I think this revision is exactly what you are looking for.

d0sboots:
Apologies for taking a long time on this, and more critically, for not having thought deeply enough about it before I made suggestions earlier.

I said this area of our codebase was one of the most complex, and I am remembering why now. In working through the options regarding your PR, I uncovered a long-standing bug with memory usage calculation that's easy to repro but wouldn't come up in usual conditions, for instance.

Let me brain dump what I've determined so far, so it's not just sitting in my head:
## What are we trying to do here? (What is the goal?)

I believe we've pretty well established that there aren't any (demonstrated) memory leaks in this code, aside from the ones we know about: We permanently leak every new module because that's how JS works, and we're permanently leaking a copy of the transformed code as part of the key to the module cache, which may or may not be sharable with the actual code for the module, depending on how V8 does things. In any case, adding more caches here will only hurt memory usage, not help it.

So, the goal is targeting performance: The fact that when many copies of (especially transformed) scripts are launched on different servers or with different filenames, it takes a long time.

A concrete repro script for this is here:
```js
/** @param {NS} ns */
  ns.disableLog("ALL");

  const source = "export function main(ns) {}"

  function forEachFile(fn) {
      const fname = `noop-${i}.js`;
      fn(fname);
    }
  }

  forEachFile(fname => ns.write(fname, source, "w"));
  let launched = 0;
  forEachFile(fname => { launched += ns.run(fname) != 0 });
  // Wait for all scripts to launch and finish
  await ns.sleep(0);
  forEachFile(fname => ns.rm(fname));
}
```
Notably, this has a significant pause even though it's not doing any transforming - more on that later. You can change the filename to ".jsx" to make the pause larger.

Now, approaches to solve this:
Well, first, a transform cache can't solve this entirely, because as demonstrated the problem still exists even when there is no script transforming involved. This is because of two other steps that also occur in this case: The ast-walk to find imports, and the ram calculation (which does its own ast-walk, and is the larger share of the time). Ultimately, both would have to be avoided to have truly fast execution.

Of course, the transform step would *also* have to be avoided, since it is also at least as slow as the other two, and sometimes slower.

One conclusion I've come to is that I don't think a transform cache is a good answer, even to just the transform part. This is because when code is transformed, it produces both the transformed code *and* a source map. If you save only the transformed code, and then use that cached copy later, it will lack a source map and so debugging will be poor. (And it will be very unclear to the user under what circumstances a source map will exist - also not a good situation.) However, if you cache the source map as well, that's a lot of extra space per script - adding the transformed code and the source map on top of the original code is probably ~3x the size. That doesn't seem like a good solution either.

I haven't given up here, but for the moment I don't have any concrete good ideas. I wanted to share my thoughts now though, so it wasn't just radio silence.

Sphyxis:
One thing I did try was a WeakRef version of the transform cache, which seemed to work pretty well.  My original thought was for it to have its own map, but I should be able to convert the attachment to the moduleCache to use a WeakRef.  That should be OK to use that for, as it does not get rid of the module reference itself.  This way, you permanently cache the module with it's reference, and allow the transforms to happen once as long as the script is currently running.  Then when the last one stops GC is given a chance to clean up the transform information only.

PR #2898, opened 2026-06-18
UI: Fix sell division window profit display

TomasVaner:
image
image

The profit shown in the Sell division windows was 10 times lower than the actual profit.

The source of the bug was that the profit (`divisionToSell.lastCycleRevenue - divisionToSell.lastCycleExpenses`) was divided by 10. But the calculation for revenue/expenses already includes cycle length: https://github.com/bitburner-official/bitburner-src/blob/09e85517704ea81adb448a9f5345af031fe90ed7/src/Corporation/Division.ts#L173

### DIFF
--- a/src/Corporation/ui/modals/SellDivisionModal.tsx
+++ b/src/Corporation/ui/modals/SellDivisionModal.tsx
@@ -62,10 +62,7 @@ export function SellDivisionModal(props: IProps): React.ReactElement {
           rows={[
             [
               "Profit:",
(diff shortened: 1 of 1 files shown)


d0sboots:
This looks good to me, but I'm going to wait on CatLover (the corp master) just in case.

TomasVaner:
image
image

Also the comparison after the fix applied

catloversg (approved):

Issue #2899, opened 2026-06-18
Share bonus is applied inconsistently


TomasVaner:
RAM share bonus is applied to faction work types inconsistently. 

For Hacking job it multiplies the result: 
``` js
export function getHackingWorkRepGain(p: IPerson, favor: number): number {
  return (
    ((p.skills.hacking + p.skills.intelligence / 3 + getDarknetCharismaBonus(p, 0.1)) / CONSTANTS.MaxSkillLevel) *
    p.mults.faction_rep *
    calculateIntelligenceBonus(p.skills.intelligence, 1) *
    mult(favor) *
    calculateCurrentShareBonus()
  );
}
```

For Security job it multiplies Hackings and Intelligence skills:
``` js
export function getFactionSecurityWorkRepGain(p: IPerson, favor: number): number {
  const t =
    (0.9 *
      (p.skills.strength +
        p.skills.defense +
        p.skills.dexterity +
        p.skills.agility +
        getDarknetCharismaBonus(p, 0.3) +
        (p.skills.hacking + p.skills.intelligence) * calculateCurrentShareBonus())) /
    CONSTANTS.MaxSkillLevel /
    4.5;
  return t * p.mults.faction_rep * mult(favor) * calculateIntelligenceBonus(p.skills.intelligence, 1);
}
```

For Field Work job it multiplies Hacking and Intelligence skills but also includes BN15 bonus:
``` js
export function getFactionFieldWorkRepGain(p: IPerson, favor: number): number {
  const t =
    (0.9 *
      (p.skills.strength +
        p.skills.defense +
        p.skills.dexterity +
        p.skills.agility +
        p.skills.charisma +
        (p.skills.hacking + p.skills.intelligence + getDarknetCharismaBonus(p, 0.3)) * calculateCurrentShareBonus())) /
    CONSTANTS.MaxSkillLevel /
    5.5;
  return t * p.mults.faction_rep * mult(favor) * calculateIntelligenceBonus(p.skills.intelligence, 1);
}
```

I suggest changing it to the way Hacking job works (make it reputation gain multiplier). It would make the description of RAM share on the Faction page consistent with what it does. Making the description to reflect the current situation would not make it very clear.

Also this change would make RAM share viable in BNs that nerf hacking skill severely (like BN9). The values of the bonus are not gonna be that big (since the bonus is logarithmic) so it won't break anything.

[...]

PR #2900, opened 2026-06-18
MISC: Change how share bonus affects security work and faction work

TomasVaner:
PR to resolve a #2899

### DIFF
--- a/src/PersonObjects/formulas/reputation.ts
+++ b/src/PersonObjects/formulas/reputation.ts
@@ -5,22 +5,23 @@ import { Person as IPerson } from "@nsdefs";
 import { calculateIntelligenceBonus } from "./intelligence";
 import { Player } from "@player";
 
-function mult(favor: number): number {
+function mult(p: IPerson, favor: number): number {
   let favorMult = 1 + favor / 100;
   if (isNaN(favorMult)) {
     favorMult = 1;
   }
-  return favorMult * currentNodeMults.FactionWorkRepGain;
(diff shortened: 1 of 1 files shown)


src/PersonObjects/formulas/reputation.ts
```diff
@@ -5,36 +5,39 @@ import { Person as IPerson } from "@nsdefs";
...
       (p.skills.strength +
         p.skills.defense +
         p.skills.dexterity +
         p.skills.agility +
-        getDarknetCharismaBonus(p, 0.3) +
-        (p.skills.hacking + p.skills.intelligence) * calculateCurrentShareBonus())) /
+        p.skills.hacking +
+        p.skills.intelligence +
+        getDarknetCharismaBonus(p, 0.3)) *
+      mult(p, favor)) /
```
catloversg: You are applying `mult(p, favor)` twice.

catloversg (commented):
Consideration: The scale of charisma and how effective three types of work are in rep farming.

Due to dnet, charisma now scales much more strongly than the other stats even after all recent nerfs. This means that, in some cases (I'm not sure how many), security work and field work may become significantly better than hacking work for reputation farming. This is a balance change. This PR further increases that potential gap.

It's not exactly an issue caused by this PR, but it's still something to keep in mind. I usually ask for an analysis of this kind of balance change, but I'll skip it this time. If I notice a significant gap between these types of work, I'll rebalance them later.

PR #2901, opened 2026-06-18
TOOLS: Convert Electron packaging script to JS

catloversg:
While refactoring `package-electron.sh` as requested in https://github.com/bitburner-official/bitburner-src/pull/2882#discussion_r3408950931, I remembered an issue that new contributors on Windows often encounter. When running tasks that require bash (e.g., doc, build, build:dev, electron), cmd and powershell print:
```
> bitburner@3.0.2 doc
> bash ./tools/doc.sh

Windows Subsystem for Linux has no installed distributions.
Distributions can be installed by visiting the Microsoft Store:
https://aka.ms/wslstore
```
Inexperienced contributors usually either do not know what this means or think that they need to install WSL. However, all they actually need is Git Bash, which is very likely already available if they have Git installed.

I initially considered updating the contributing documentation to point out this issue, but I now think it would be better to convert these scripts to JS instead.

If this idea is accepted, I'll convert the remaining scripts later.

Tested on Windows 10.
Sample run: https://github.com/catloversg/bitburner-src/actions/runs/27756468639


d0sboots:
I personally like bash more for this application, but writing good bash scripts has a lot of hidden gotchas. Add in the Windows issues, and overall this seems like a better approach for us.

PR #2905, opened 2026-06-19
FIX: Make skillMaxUpgradeCount work in all cases

### LINKED ISSUE #1693: API: skillMaxUpgradeCount returns 1 when SP is too small and current level is too high

d0sboots:
This bug has been lingering a long time. A proper fix required the ability to find the next greater/smaller floating-point number. Fixes: #1693

I also changed the format on the big comment block away from LaTeX, for better readability. LaTeX is readable-enough for formulas that will be rendered, but comments don't get rendered.

The tests pass. I didn't do any manual testing.

### DIFF
--- a/src/Bladeburner/Skill.ts
+++ b/src/Bladeburner/Skill.ts
@@ -35,12 +35,16 @@ export class Skill {
     for (const [multName, mult] of getRecordEntries(params.mults)) this.mults[multName] = mult;
   }
 
-  calculateCost(currentLevel: number, count = 1 as PositiveInteger): number {
+  getActualUpgradeCount(currentLevel: number, count: number): number {
     // This avoids an exploit where at high currentLevels, skills can be bought more cheaply by manipulating
     // rounding. Adding "count" to "currentLevel" will round in these circumstances because not all integer values
     // are representable; this leads to potentially getting more levels than we asked for. By performing the same
     // rounding here, we calculate the cost for the number of levels we will actually get.
-    const actualCount = currentLevel + count - currentLevel;
+    return currentLevel + count - currentLevel;
+  }
+
+  calculateCost(currentLevel: number, count = 1 as PositiveInteger): number {
+    const actualCount = this.getActualUpgradeCount(currentLevel, count);
     /**
      * The cost of the next level: (baseCost + currentLevel * costInc) * mult. The cost needs to be an integer, so we
      * need to use Math.floor or Math.round.
@@ -132,7 +136,15 @@ export class Skill {
      *
      * x_2 = -Delta - m
      *
-     * a, c and y are always greater than 0, so x_2 is always less than 0. Therefore, x_1 is the only solution.
+     * Since a, c, and y are all greater than 0:
+     *
+     * Delta = sqrt(m^2 + 2y/(a*c)) > |m|
+     *
+     * Therefore:
+     *
+     * x_2 = -Delta - m < -|m| - m <= 0
+     *
+     * Since the first inequality is strict, x_2 < 0. Thus, x_1 is the only non-negative solution.
      *
      * However, calculating it directly in this form will lead to catastrophic cancellation when m gets large.
      * So, multiply top and bottom by (Delta + m):
@@ -182,9 +194,149 @@ export class Skill {
     return r0;
   }
 
+  calculateMaxUpgradeCountNew(currentLevel: number, cost: PositiveNumber): number {
+    // Sanity checks for bnMult and skill settings. Callers should already check currentLevel and cost.
+    if (currentNodeMults.BladeburnerSkillCost <= 0 || this.baseCost <= 0 || this.costInc <= 0) {
+      throw new Error(
+        `Invalid bnMult or skill settings. BladeburnerSkillCost: ${currentNodeMults.BladeburnerSkillCost}, baseCost: ${this.baseCost}, costInc: ${this.costInc}`,
+      );
+    }
+
+    // Find the smallest count that registers a real, nonzero level change. The dead zone below this can span more than
+    // one count, so this needs its own search, not a single fixed check at count = 1.
+    let lo = 0;
+    let hi = 1;
+    while (this.getActualUpgradeCount(currentLevel, hi) === 0) {
+      lo = hi;
+      hi *= 2;
+    }
+    // At extreme magnitudes, the gap between adjacent representable doubles can exceed 1, so "hi - lo > 1" would never
+    // become false and the loop would spin indefinitely. The "mid === lo || mid === hi" check would break the loop in
+    // that case.
+    while (hi - lo > 1) {
+      const mid = lo + Math.floor((hi - lo) / 2);
+      // Granularity floor, can't narrow further
+      if (mid === lo || mid === hi) {
+        break;
+      }
+      if (this.getActualUpgradeCount(currentLevel, mid) === 0) {
+        lo = mid;
+      } else {
+        hi = mid;
+      }
+    }
+    const countMin = hi;
+
+    const isAffordable = (count: number) =>
+      count >= countMin && this.calculateCost(currentLevel, count as PositiveInteger) <= cost;
+
+    // Return early if even the cheapest viable request is not affordable.
+    if (!isAffordable(countMin)) {
+      return 0;
+    }
+
+    /**
+     * Define:
+     * - x = count
+     * - a = currentNodeMults.BladeburnerSkillCost
+     * - b = this.baseCost
+     * - c = this.costInc
+     * - d = currentLevel
+     * - y = cost
+     *
+     * We have:
+     *
+     * y = round(x*a*(b + c*(d + (x - 1)/2)))
+     *
+     * To simplify the calculation, let's ignore the Math.round part:
+     *
+     * y = x*a*(b + c*(d + (x - 1)/2))
+     *
+     * Divide by a and c to simplify things, multiply by 2 to help complete the square:
+     *
+     * 2y/(a*c) = 2x*(b/c + d + (x - 1)/2)
+     *          = x^2 + 2x*(b/c + d - 1/2)
+     *
+     * Solve for x in terms of everything else:
+     *
+     * Define:
+     *
+     * m = b/c + d - 1/2
+     *
+     * 2y/(a*c) = x^2 + 2x*m
+     * m^2 + 2y/(a*c) = x^2 + 2xm + m^2 = (x + m)^2
+     *
+     * Define:
+     *
+     * Delta = sqrt(m^2 + 2y/(a*c))
+     *
+     * Solutions:
+     *
+     * x_1 = Delta - m
+     *
+     * x_2 = -Delta - m
+     *
+     * Since a, c, and y are all greater than 0:
+     *
+     * Delta = sqrt(m^2 + 2y/(a*c)) > |m|
+     *
+     * Therefore:
+     *
+     * x_2 = -Delta - m < -|m| - m <= 0
+     *
+     * Since the first inequality is strict, x_2 < 0. Thus, x_1 is the only non-negative solution.
+     *
+     * However, calculating it directly in this form will lead to catastrophic cancellation when m gets large.
+     * So, multiply top and bottom by (Delta + m):
+     *
+     * x_1 = (Delta - m)(Delta + m)/(Delta + m)
+     *     = (Delta^2 - m^2)/(Delta + m)
+     *     = (m^2 + 2y/(a*c) - m^2)/(Delta + m)
+     *     = (2y/(a*c))/(Delta + m)
+     */
+    const m = this.baseCost / this.costInc + currentLevel - 0.5;
+    const yac = (2 * cost) / (currentNodeMults.BladeburnerSkillCost * this.costInc);
+    const delta = Math.sqrt(m * m + yac);
+    const numericallyStableEstimate = yac / (delta + m);
+
+    let seed = Math.round(numericallyStableEstimate);
+    if (!Number.isFinite(seed) || seed < countMin) {
+      seed = countMin;
+    }
+
+    // Expand outward from the seed to bracket the true boundary.
+    // After this step, it's guaranteed that lower is affordable and upper is not.
+    let lower = countMin;
+    let upper = seed;
+    let step = Math.max(1, seed - countMin);
+    // This loop cannot be infinite. If upper is a too high finite value (not affordable) or Infinity, isAffordable will
+    // return false.
+    while (isAffordable(upper)) {
+      lower = upper;
+      upper += step;
+      step *= 2;
+    }
+
+    // Binary search the exact boundary.
+    // Same comment about the "upper - lower > 1" case and the "mid === lower || mid === upper" check as above.
+    while (upper - lower > 1) {
+      const mid = lower + Math.floor((upper - lower) / 2);
+      if (mid === lower || mid === upper) {
+        break;
+      }
+      if (isAffordable(mid)) {
+        lower = mid;
+      } else {
+        upper = mid;
+      }
+    }
+
+    return lower;
+  }
+
   canUpgrade(bladeburner: Bladeburner, count = 1): Availability<{ actualCount: number; cost: number }> {
     const currentLevel = bladeburner.skills[this.name] ?? 0;
-    const actualCount = currentLevel + count - currentLevel;
+    const actualCount = this.getActualUpgradeCount(currentLevel, count);
     if (actualCount === 0) {
       return {
         error: `Cannot upgrade ${this.name}: Due to floating-point inaccuracy and the small value of specified "count", your skill cannot be upgraded.`,
--- a/src/Bladeburner/Skill.ts
+++ b/src/Bladeburner/Skill.ts
@@ -44,7 +44,11 @@ export class Skill {
   }
 
   calculateCost(currentLevel: number, count = 1 as PositiveInteger): number {
+    if (!isFinite(count) || !isFinite(currentLevel)) return Number.POSITIVE_INFINITY;
+
     const actualCount = this.getActualUpgradeCount(currentLevel, count);
+    // Edge case: If currentLevel is too high, the calculation below can result in NaN.
+    if (actualCount === 0) return 0;
     /**
      * The cost of the next level: (baseCost + currentLevel * costInc) * mult. The cost needs to be an integer, so we
      * need to use Math.floor or Math.round.
@@ -95,6 +99,19 @@ export class Skill {
   }
 
   calculateMaxUpgradeCount(currentLevel: number, cost: PositiveNumber): number {
+    // Check edge cases.
+    if (Number.isNaN(currentLevel) || Number.isNaN(cost)) {
+      return Number.NaN;
+    }
+    if (cost < 1) {
+      return 0;
+    }
+    if (!Number.isFinite(cost)) {
+      return Number.POSITIVE_INFINITY;
+    }
+    if (!Number.isFinite(currentLevel)) {
+      return 0;
+    }
     /**
      * Define:
      * - x = count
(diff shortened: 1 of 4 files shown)


src/utils/NextAfter.ts
```diff
@@ -0,0 +1,40 @@
...
+  }
+  if (x === y) {
+    // Subtle -0 case here.
+    return y;
+  }
+  if (x === 0) {
+    // y can't be 0, from above
+    return y > 0 ? Number.MIN_VALUE : -Number.MIN_VALUE;
+  }
+  const f64 = Float64Array.of(x, y);
```
hexagonrecursion:
```suggestion
  const f64 = Float64Array.of(x);
```

src/utils/NextAfter.ts
```diff
@@ -0,0 +1,40 @@
...
+  const u32 = new Uint32Array(f64.buffer);
```
hexagonrecursion: The byte order of Float64Array and Uint32Array depends on the platform. Maybe we should replace both Float64Array and Uint32Array with DataView

d0sboots: DataView makes the code a bit more awkward, but sure. I'm not changing the tests because of the readability hit, and because there is zero chance we will ever be running them on IBM z/Architecture mainframes XD

catloversg:
This new approach is similar to the old one in that it assumes there is a "window" that always contains the true result, and that the true result is always either at one edge of the window (r0 or r2) or at a specific value within it (r1). The old version assumed that ±1 was sufficient, but we later found counterexamples showing that it was not. This new version attempts to construct a more accurate window, but I also found counterexamples showing that this window is not correct in all cases.

Counterexample 1:
baseCost = 1, costInc = 2.5, mult = 1
currentLevel = 1e18, cost = 320000000000000070000 (`nextafter(3.2e20, 1e99)`)

In your Jest test, you expect the true result to be 128, but it's actually 191.
- r0 = 0, r1 = 128, r2 = 256
- calculateCost(1e18, 128) = 320000000000000000000
- calculateCost(1e18, 129) = 320000000000000000000
- calculateCost(1e18, 191) = 320000000000000000000
- calculateCost(1e18, 192) = 640000000000000100000

191 is at the exact boundary, not 128.

This is acceptable in our case because `1e18 + 128 === 1e18 + 191`. In other words, calling `upgradeSkill` with either value produces the same final skill level. However, that will not be true when the true result falls outside the expected [r0, r2] window.

Counterexample 2:
baseCost = 1, costInc = 1, mult = 1
currentLevel = 2048, cost = 2.049e32

- r0 = 20243517480910204, r1 = 20243517480910210, r2 = 20243517480910212
- calculateCost(2048, 20243517480910212) = 2.0490000000000017e+32
- calculateCost(2048, 20243517480910210) = 2.0490000000000006e+32

Since neither r2 nor r1 is affordable, the implementation assumes that r0 is the correct answer. However, `calculateCost(2048, 20243517480910204)` is 2.0490000000000002e+32, which is also not affordable.
The true result is 20243517480910200 (`calculateCost(2048, 20243517480910200) = 2.0489999999999988e+32`).

Counterexample 3:
baseCost = 1, costInc = 1, mult = 1
currentLevel = 1, cost = 1e308

This approach assumes that a valid window can be constructed from the numerically stable estimate. However, if the estimate itself is NaN, the approach breaks down. In this example, the estimate is NaN, which causes the computed window to become invalid as well.

Based on these examples, I think we should stop pursuing approaches that rely on constructing a window and assuming that the true result lies within it. Such an approach would be fine if we could prove that assumption, but in the last two attempts, we handwaved the proof, and we have repeatedly found counterexamples afterward.

I submitted an alternative implementation that uses binary search to find the true result directly. I also added additional Jest tests. The new function is called `calculateMaxUpgradeCountNew`. You can switch between the existing `calculateMaxUpgradeCount` implementation and the new one for testing.

d0sboots:
Before I make a detailed reply, I am adding a *lot* more tests. I'll reply once that's done.

src/utils/NextAfter.ts
d0sboots: So, funny thing: I talked with my friend today who used to work on Chrome. He said that people there were basically split 50/50 on handling endian issues; some of them did, but a lot of them felt it was a waste of time. In other words, there is basically 0 chance that Chrome *itself* will ever run on a big-endian system. XD (I imagine the same is likely true for Firefox.)

d0sboots:
> This new approach is similar to the old one in that it assumes there is a "window" that always contains the true result, and that the true result is always either at one edge of the window (r0 or r2) or at a specific value within it (r1). The old version assumed that ±1 was sufficient, but we later found counterexamples showing that it was not. This new version attempts to construct a more accurate window, but I also found counterexamples showing that this window is not correct in all cases.

This is true, and you're right that the window approach was flawed. More below.

> Counterexample 1: baseCost = 1, costInc = 2.5, mult = 1 currentLevel = 1e18, cost = 320000000000000070000 (`nextafter(3.2e20, 1e99)`)
> 
> In your Jest test, you expect the true result to be 128, but it's actually 191.
> 
> * r0 = 0, r1 = 128, r2 = 256
> * calculateCost(1e18, 128) = 320000000000000000000
> * calculateCost(1e18, 129) = 320000000000000000000
> * calculateCost(1e18, 191) = 320000000000000000000
> * calculateCost(1e18, 192) = 640000000000000100000
> 
> 191 is at the exact boundary, not 128.
> 
> This is acceptable in our case because `1e18 + 128 === 1e18 + 191`. In other words, calling `upgradeSkill` with either value produces the same final skill level. However, that will not be true when the true result falls outside the expected [r0, r2] window.

This is not a counterexample, this is by design. (Which is why I was testing for this value explicitly.) In a similar way to how non-integers are not valid amounts to pass to `calculateCost` and related functions, amounts that can't be added to level in a representable fashion aren't valid either. It's not an *error* to use them (that would be too punishing to users), but they way we calculate makes it clear that, semantically, they're not valid amounts. Therefore, it is not appropriate to return them from `skillMaxUpgradeCount`.

> Counterexample 2: baseCost = 1, costInc = 1, mult = 1 currentLevel = 2048, cost = 2.049e32
> 
> * r0 = 20243517480910204, r1 = 20243517480910210, r2 = 20243517480910212
> * calculateCost(2048, 20243517480910212) = 2.0490000000000017e+32
> * calculateCost(2048, 20243517480910210) = 2.0490000000000006e+32
> 
> Since neither r2 nor r1 is affordable, the implementation assumes that r0 is the correct answer. However, `calculateCost(2048, 20243517480910204)` is 2.0490000000000002e+32, which is also not affordable. The true result is 20243517480910200 (`calculateCost(2048, 20243517480910200) = 2.0489999999999988e+32`).

This is a true counterexample, and I added it to the tests.

> Counterexample 3: baseCost = 1, costInc = 1, mult = 1 currentLevel = 1, cost = 1e308
> 
> This approach assumes that a valid window can be constructed from the numerically stable estimate. However, if the estimate itself is NaN, the approach breaks down. In this example, the estimate is NaN, which causes the computed window to become invalid as well.

This was a case where the old code failed, but not due to the method being flawed (although it was.) Rather, this was due to insufficient care being taken with edge cases. Your approach also has these types of issues. This was revealed after I added tests of this nature.

> Based on these examples, I think we should stop pursuing approaches that rely on constructing a window and assuming that the true result lies within it. Such an approach would be fine if we could prove that assumption, but in the last two attempts, we handwaved the proof, and we have repeatedly found counterexamples afterward.
> 
> I submitted an alternative implementation that uses binary search to find the true result directly. I also added additional Jest tests. The new function is called `calculateMaxUpgradeCountNew`. You can switch between the existing `calculateMaxUpgradeCount` implementation and the new one for testing.

The first thing I did was to *significantly* beef up the tests, including adding a random-based stress test. The last one, in particular, ended up being invaluable not for uncovering bugs in my eventual algorithm (the algorithm turned out to be fine), but rather bugs in my assumptions that led to the test code itself being flawed. Because of that, I have much higher confidence in the final correctness of the result, now.

I reverted your test changes, because mine ended up overlapping them, and also because you changed the method name under test. IMO the right way to test alternate implementations is not to change it in the tests, but rather to rename the main method, so you're certain to test everything end-to-end.

I changed my implementation from a hardcoded window to a linear search. The correctness of this should be much more straightforward to see, even more than your binary search. The *speed* of it is still based on a handwaved assertion that the approximation ends up very close to the true result, but both intuition and testing support this.

My new tests showed failures (in many places) on my old code, and obviously pass now. They also fail on your binary search method, producing at least one infinite loop that I didn't bother to track down.

d0sboots:
Oh, one other thing I was going to mention but forgot in my previous comment: Getting floating-point library code right is *hard*. Having mistakes in a prior version is not necessarily evidence that the prior approach is intrinsically flawed (although it still could be). If you look through the source of libraries that implement things like `exp`, `sin`, `hypot`, etc., you will see comments and techniques that are very similar or identical to the things I'm doing here - the reasoning involved can be tricky, but it is by no means unprecedented.

src/Bladeburner/Skill.ts
```diff
@@ -72,15 +85,39 @@ export class Skill {
...
+    // Sanity checks for bnMult and skill settings.
+    if (currentNodeMults.BladeburnerSkillCost <= 0 || this.baseCost <= 0 || this.costInc <= 0) {
+      throw new Error(
+        `Invalid bnMult or skill settings. BladeburnerSkillCost: ${currentNodeMults.BladeburnerSkillCost}, baseCost: ${this.baseCost}, costInc: ${this.costInc}`,
+      );
+    }
+    if (Number.isNaN(currentLevel) || Number.isNaN(cost)) {
+      return Number.NaN;
+    }
+    if (cost < currentNodeMults.BladeburnerSkillCost * this.baseCost) {
```
catloversg: Can you explain why we need to check this case beforehand?

Before my commit, this line checked `cost < this.baseCost`, but there are cases where this check causes the function to return the wrong value. I added a test for that.

test/jest/Bladeburner/Skill.test.ts
```diff
@@ -1,18 +1,178 @@
...
-import type { PositiveNumber } from "../../../src/types";
+import { Skill } from "../../../src/Bladeburner/Skill";
+import { nextafter } from "../../../src/utils/NextAfter";
+import type { PositiveInteger, PositiveNumber } from "../../../src/types";
+import { currentNodeMults } from "../../../src/BitNode/BitNodeMultipliers";
 
 const hyperdrive = Skills[BladeburnerSkillName.Hyperdrive];
 
+// Given two Uint32s (presumably randomly generated), return a floating point
+// value for testing. The first part of the tuple is an integer, and the 2nd
```
catloversg: Optional (IIUC):
```js
// Given two random Uint32 values, construct a test value represented as
// (integer, multiplier), where the numeric value is integer * multiplier.
// The first part of the tuple is an integer ...
```

test/jest/Bladeburner/Skill.test.ts
```diff
@@ -1,18 +1,178 @@
...
+      expect(result).toBe(Infinity);
+    });
+
+    it("Inf arg both", () => {
+      const result = hyperdrive.calculateCost(Infinity, Infinity as PositiveInteger);
+      expect(result).toBe(Infinity);
+    });
+
+    it("Last possible upgrade", () => {
+      const level = 2 ** 537 * 1.6 - 2 ** 486;
```
catloversg, line 78: Can you write a bit more detail about the goal of this test and what kind of error it catches? These numbers look very specific, so I want to know how and why you came up with them.

test/jest/Bladeburner/Skill.test.ts
```diff
@@ -1,18 +1,178 @@
...
+    it("should return 0 when currentLevel is too high for floating-point precision", () => {
+      // At level 1e18, the smallest gap between numbers is 128. The cost of
+      // 128 levels will be 128 * (1 + 2.5 * 127/2 + 2.5 * 1e18) = 128 * (2.5e18) = 3.2e20.
+      const result = hyperdrive.calculateMaxUpgradeCount(1e18, nextafter(3.2e20, 0) as PositiveNumber);
+      expect(result).toBe(0);
+    });
+
+    it("should return correct count at high currentLevel", () => {
+      const result = hyperdrive.calculateMaxUpgradeCount(1e18, nextafter(3.2e20, 1e99) as PositiveNumber);
+      expect(result).toBe(128);
```
catloversg, line 255: Optional: Copy the comment referencing this PR from above to here.

Expecting the value of 128 is very debatable, IMO. I'll explain why in detail later.

test/jest/Bladeburner/Skill.test.ts
```diff
@@ -1,18 +1,178 @@
...
+        baseCost: 1,
+        costInc: 1,
+        mults: {},
+      });
+      const result = testSkill.calculateMaxUpgradeCount(2048, 2.049e32 as PositiveNumber);
+      expect(result).toBe(20243517480910200);
+    });
+
+    it("should return 0 when currentLevel is too high for floating-point precision", () => {
+      // At level 1e18, the smallest gap between numbers is 128. The cost of
```
catloversg: Optional: Explain what "smallest gap" means in this context. I assume you mean it's the representable value of `1000000000000000100 - 1e18` (more about it later).

catloversg:
> This is not a counterexample, this is by design. (Which is why I was testing for this value explicitly.) In a similar way to how non-integers are not valid amounts to pass to calculateCost and related functions, amounts that can't be added to level in a representable fashion aren't valid either. It's not an error to use them (that would be too punishing to users), but they way we calculate makes it clear that, semantically, they're not valid amounts. Therefore, it is not appropriate to return them from skillMaxUpgradeCount.

This argument is not wrong, but I think it needs some clarification. The first time I read it, it sounded contradictory to me. Only after rereading it many times and considering it from different perspectives, I agree with it to some extent (I still don't agree with it entirely, though).

For posterity's sake, I'll write down what I think about your argument here, and why I accept the value of 128 even though I don't think it's the *only* correct value (depending on the perspective). This is basically the clarification that I think the argument needs.

First, these are 3 different perspectives.

```js
const budget = 320000000000000070000;
const count = calculateMaxUpgradeCount(1e18, budget);
upgradeSkill(1e18, count);
```
[1]: What you do is *exactly* what you get.
If count is `foo`, then `skillLevel` must be *exactly* `skillLevel + foo`. For example:
- count = 65 => skillLevel = 1000000000000000065 (impossible)
- count = 100 => skillLevel = 1000000000000000100
- count = 128 => skillLevel = 1000000000000000128 (impossible)
- count = 191 => skillLevel = 1000000000000000191 (impossible)

[2]: I guarantee you cannot find a higher `count` such that `upgradeSkill(1e18, count)` is successful.
If count is `foo`, then there is no `foo + PositiveInteger` such that calling `upgradeSkill(1e18, foo + PositiveInteger)` is successful. For example:
- count = 65 => `upgradeSkill(1e18, 66)` is an error (not affordable) (wrong)
- count = 128 => `upgradeSkill(1e18, 129)` is an error (not affordable) (wrong)
- count = 191 => `upgradeSkill(1e18, 192)` is an error (not affordable) (correct)

[3]: Suppose `foo` is the highest *representable* level that you can upgrade to with `budget`, `count` is the *representable* value of `foo - currentLevel`.
At currentLevel = 1e18 and budget = 320000000000000070000, you cannot upgrade to any *representable* level higher than 1000000000000000100. Therefore, `count` is the *representable* value of `1000000000000000100 - currentLevel`.

The first time I read your argument, I thought you were saying that 191 is not an appropriate value because `1e18 + 191` is clearly not `1000000000000000191` (this is not a representable value). This is where I thought it was contradictory. How can 128 be the correct value? In [1], both 128 and 191 are the "wrong" values, and the only correct value is 100.

Given floating-point inaccuracy, [1] is clearly too "strict", and after reading your implementation, I think I understand what's going on. The new approach focuses on [3] while the binary search focuses on [2]. In other words, we have different perspectives about what the true answer should be.
- With [2], I focus on finding the highest `count` such that calling `upgradeSkill` is successful.
- With [3], you focus on finding the highest *representable* affordable level and then return the *representable* gap between it and the current level.

With [3] in mind, I accept the value of 128. 128 is clearly not the highest `count` such that `upgradeSkill(1e18, count)` is successful (from the perspective of [2]), but it's the *representable* value of `1000000000000000100 - 1e18`, and 1000000000000000100 is the highest *representable* level that you can upgrade to with `budget`.

> Your approach also has these types of issues. This was revealed after I added tests of this nature.

What's the issue? I only see the infinite loop that you mentioned, but it only happens because you pass values that it considers invalid. Specifically, that code assumes `currentLevel` and `cost` are always finite (`// Sanity checks for bnMult and skill settings. Callers should already check currentLevel and cost.`). "Callers" are the functions in `src/NetscriptFunctions/Formulas.ts`. Checking these infinite cases in `calculateMaxUpgradeCount` is simple, so it's not a big deal.

The stress test shows cases in which the binary search returns an unexpected value (e.g., `With parameters {level:1.8449903992772817e+76, cost:6.652694954817565e+151, prevCost:6.652694954817564e+151, amount:1.3899679065382986e+75}`), but that's the same case with the 128 vs 191 case.

> I reverted your test changes, because mine ended up overlapping them, and also because you changed the method name under test. IMO the right way to test alternate implementations is not to change it in the tests, but rather to rename the main method, so you're certain to test everything end-to-end.

Those tests are the "stepping stones" for the discussion, not the final tests for merging. I assumed you would switch the calls to different implementations manually for testing. In other words, they are WIP tests for the discussion phase. I set them to the new implementation to demonstrate what I said.

> They also fail on your binary search method, producing at least one infinite loop that I didn't bother to track down.

I explained the infinite loop above.

d0sboots:
I think perspectives [1] and [3] are actually the same, although the way I'm thinking about it is closer to 3.

> ```js
> const budget = 320000000000000070000;
> const count = calculateMaxUpgradeCount(1e18, budget);
> upgradeSkill(1e18, count);
> ```
> 
> [1]: What you do is _exactly_ what you get. If count is `foo`, then `skillLevel` must be _exactly_ `skillLevel + foo`. For example:
> 
> * count = 65 => skillLevel = 1000000000000000065 (impossible)
> * count = 100 => skillLevel = 1000000000000000100
> * count = 128 => skillLevel = 1000000000000000128 (impossible)
> * count = 191 => skillLevel = 1000000000000000191 (impossible)

1e18 + 100 is *not* 1000000000000000100. (Mathematically it is, obviously, but we're talking floating point here.) It is 1000000000000000128. You can be thrown off on this point because JS will only print as many digits as it needs for a precise representation, which is typically 17, and fill in the rest with zeros.

You typically need to use programs that support arbitrary precision, and use the binary representation of the float, if you want an exact decimal representation.

src/Bladeburner/Skill.ts
d0sboots: I was being defensive but ended up being too lax. Checking <= 0 would be fine.

test/jest/Bladeburner/Skill.test.ts
d0sboots: I'm not sure how to explain this more/better without explaining the entirety of floating point representation. 

The fundamental property of how floating point numbers work is that they have a mantissa and an exponent. Since the value is mantissa * 2^exponent, there is a gap between representable numbers which is a power of 2 based on the size of the exponent.

d0sboots:
Ah, you can actually see this relatively simply directly in JS:
```
> BigInt(1e18)
1000000000000000000n
> BigInt(1e18+100)
1000000000000000128n
```

d0sboots:
> What's the issue? I only see the infinite loop that you mentioned, but it only happens because you pass values that it considers invalid. Specifically, that code assumes currentLevel and cost are always finite (// Sanity checks for bnMult and skill settings. Callers should already check currentLevel and cost.). "Callers" are the functions in src/NetscriptFunctions/Formulas.ts. Checking these infinite cases in calculateMaxUpgradeCount is simple, so it's not a big deal.

One of the issues is an out-of-range problem in `calculateCost` that produced NaN despite valid values. I didn't 100% track down the logic of your binary search, but it seemed unlikely to me that it would have produced the correct result with that unfixed.

test/jest/Bladeburner/Skill.test.ts
d0sboots: Done.

test/jest/Bladeburner/Skill.test.ts
d0sboots: Done

test/jest/Bladeburner/Skill.test.ts
d0sboots: Done.

d0sboots:
Your defensive check made me realize there was yet another edge case, which I added a test and a fix for.

d0sboots:
My previous fix left a bad taste in my mouth, and going back I was able to find another edge case with it. This fix is more general.

All of these later class of bugs I am 90% sure would not occur with the parameters we use in the game, but they are technically valid parameters, so the function should work with them.

PR #2909, opened 2026-06-21
DNET: Check darknet API access for all dnet methods

### LINKED ISSUE #2908: Some darknet functions do not require DarkscapeNavigator.exe
MRE:
Buy or create a program. Then, go to the terminal and remove it. A popup will show alerting the user that removing a program cannot be undone. If you click yes, the program will be removed and not avaliable until the next soft reset. However, the effects of the program (depending on which one) are still applied. If you remove `DarkscapeNavigator.exe` you will still be able to use the API and the DarkNet page on the sidebar. Also, the "Create Program" page will show you already own the program.

The alert feature is intentional, however I think instead of showing an alert and removing it permanently there should be a way to re-create the program without recurring to a soft reset. It's true that this cannot be done accidentally, but being able to re-buy or re-create the program should be possible to prevent any future bugs or requiring soft resets (which loose all progress) to the players.

ficocelliguy:
This PR adds a check to ensure that players have the DarknetNavigator.exe to the remaining dnet namespace methods that did not yet have such a check.

Closes #2908

### DIFF
--- a/src/NetscriptFunctions/Darknet.ts
+++ b/src/NetscriptFunctions/Darknet.ts
@@ -500,7 +500,6 @@ export function NetscriptDarknet(): InternalAPI<DarknetAPI> {
     },
     isDarknetServer: (ctx) => (_host) => {
       const targetHost = helpers.string(ctx, "host", _host ?? ctx.workerScript.hostname);
-      expectDarknetAccess(ctx);
       const server = GetServer(targetHost);
       if (!server) {
         return false;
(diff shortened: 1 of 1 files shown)


E9cipher:
Closes #2908?

src/NetscriptFunctions/Darknet.ts
```diff
@@ -497,6 +500,7 @@ export function NetscriptDarknet(): InternalAPI<DarknetAPI> {
     },
     isDarknetServer: (ctx) => (_host) => {
       const targetHost = helpers.string(ctx, "host", _host ?? ctx.workerScript.hostname);
+      expectDarknetAccess(ctx);
```
d0sboots: For consideration: it might make sense to leave this one accessible. Darkweb is a darknet server that is always accessible, and you can manually route to it even without probe.

If we leave this one available, it should be documented.

ficocelliguy: It was intended to not need the Navigator (it's already documented that way), I just didn't remember that on the first pass of this change.

PR #2913, opened 2026-06-22
DNET: Harden logic around darkweb connections to home to prevent errors

ficocelliguy:
In the past there have been ways to get invalid networks, where darknet is connected to home but not vice-versa. This can still be done, technically, in the latest dev branch by starting a new save and hitting "generate new darknet network" in the dev menu without having a tor router or the darkscape navigator.

This PR hardens the logic around clearing the darknet to avoid causing uni-directional links. It also hides the dev menu options that aren't meaningful if the game doesn't have darknet initialized yet.

### DIFF
--- a/src/DarkNet/controllers/NetworkGenerator.ts
+++ b/src/DarkNet/controllers/NetworkGenerator.ts
@@ -122,7 +122,7 @@ export const clearDarknet = () => {
   }
   const darkwebRoot = GetServer(SpecialServers.DarkWeb);
   if (darkwebRoot) {
-    darkwebRoot.serversOnNetwork = [Player.getHomeComputer().hostname];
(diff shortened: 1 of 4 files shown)


hexagonrecursion:
I think the dev menu options should be disabled instead of being hidden. A UI rule of thumb: a disabled button has better discoverability than a hidden one

ficocelliguy:
I considered disabling them and adding tooltips. However, since there's only one user flow from that point (and you discover it as soon as you interact with it), and since this is a tool only for a handful of devtools power users, it's simpler to not render them if they aren't applicable. It's also easier to add to it in the future, if needed, without requiring that contributor to add special disabled logic and custom tooltips.

For a player-facing feature? Absolutely I would use disabled treatment instead, yeah

d0sboots:
I'm also for disabling them. I don't think we need a tooltip for a dev tool, but slapping a disable prop on them is low effort.

PR #2918, opened 2026-06-24
CCT: Adding Info to Largest Rectangle in a Matrix

UltraGaming1001:
This commit slightly modifies the description given for the Coding Contract "Largest Rectangle in a Matrix", by adding more information about the corners. Since the problem statement confused even me, I believe this change would clear things up a bit, by telling that the corners to be given in the answer are the top-left and bottom-right corner respectively, in [row,column] order.

### DIFF
--- a/src/CodingContract/contracts/LargestRectangle.ts
+++ b/src/CodingContract/contracts/LargestRectangle.ts
@@ -16,7 +16,8 @@ export const largestRectangle: Pick<CodingContractTypes, CodingContractName.Larg
 ${gridString}
 ]
 
-Your task is to find the two corners of the largest rectangle ([[r1,c1],[r2,c2]]) that does not contain any 1s.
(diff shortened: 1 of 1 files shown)


d0sboots:
The corners don't have to be the top-left and bottom-right corner, any opposing corners will work.

Was it not apparent from the [r1,c1] notation that the coordinates were row, column? That is very common notation.

Misha279-UA:
there is actually no difference between entering top-left, bottom-right and top-right, bottom-left, unless d0s or catlover changed something after me

UltraGaming1001:
> The corners don't have to be the top-left and bottom-right corner, any opposing corners will work.
> 
> Was it not apparent from the [r1,c1] notation that the coordinates were row, column? That is very common notation.

Oh I never actually knew that :P
From the examples given I thought it had to specifically be the top-left and bottom-right corners, in that specific order.

PR #2919, opened 2026-06-25
REFACTOR: Uncurry netscript InternalAPI

d0sboots:
Warning: You'll want to look at the diff with `--ignore-space-change` (there's a toggle for it in the GitHub UI).

The internal implementation of the NS API has been built as a two-level structure of `(ctx) => (...args) => { body }` since forever. This is both awkward and inefficient; inside `NSProxyHandler` this results in two closures per function called, one via `wrappedFunction` and an anonymous one via the two-level structure.

This flattens the structure to a single call with `ctx` as the first parameter. Because the remaining arguments are `unknown`, the typing is not quite as strict as before, so some errors can fall through the cracks - in particular, TS will not warn about dropping the `ctx` argument if it is unused. (`(arg1, arg2) => { body }` will be allowed, since the impl of `body` will treat arg1 and arg2 as unknown and so even though arg1 is a NetscriptContext it won't be an issue, from a type-safety perspective.) The correct thing in this situation is to use an underscore argument for the unused context - only a few functions need this treatment.

The translation was almost entirely mechanical. Only `Flags` required any remotely non-trivial adjustment.

### Testing

The following script runs many copies of a script that lives for long enough to measure the elevated memory pressure. Several formulas functions are run to increase usage of the NSProxyHandler.
```js
/** @param {NS} ns */
  ns.disableLog("ALL");

  const source = `export function main(ns) {
  ns.disableLog("ALL");
  ns.formulas.mockPerson();
  ns.formulas.mockPlayer();
  ns.formulas.mockServer();
  return ns.sleep(30000);
}`;
  const fname = "waitlong.js";
  ns.write(fname, source, "w");
  let launched = 0;
  let last;
    const current = ns.run(fname, {temporary: true});
    if (current != 0) {
      launched++;
      last = current;
    }
    if (count % 5000 === 0) {
      await 0;
    }
  }
  // Wait for all scripts to launch and finish
  while (ns.getRunningScript(last)) {
    await ns.sleep(100);
  }
  ns.rm(fname);
  ns.tprint("Done");
}
```
Results: The script was run as `run testScriptMem.js 100000` to make the results obvious. Memory consumed by NSProxyHandler dropped from 1.72kB to 1.42kB per instance, and NSProxyHandler accounts for a significant portion of the RAM increase while running scripts in this scenario - somewhere in the 25-50% range. So, it is a significant although not earth-shattering reduction. The memory numbers should be exact, i.e. not subject to measurement errors.

Runtime on my machine dropped from 12.22s to 11.68s. This could be a fluke, but at least we can say with reasonable confidence that this does not slow down script launching.

I didn't get numbers for FF.


src/Netscript/APIWrapper.ts
```diff
@@ -4,8 +4,11 @@ import { helpers } from "./NetscriptHelpers";
 
 /** Permissive type for the documented API functions */
 type APIFn = (...args: any[]) => unknown;
-/** Type for internal, unwrapped ctx function that produces an APIFunction */
-type InternalFn<F extends APIFn> = (ctx: NetscriptContext) => ((...args: unknown[]) => ReturnType<F>) & F;
+// This helper can't be inlined, it needs to be its own generic to function
```
catloversg: What does "to function" mean?

src/Netscript/APIWrapper.ts
```diff
@@ -4,8 +4,11 @@ import { helpers } from "./NetscriptHelpers";
...
+type Unknownify<T> = { [K in keyof T]: unknown };
```
catloversg: We defined this type in `src/types.ts`. That version is a bit different:
```ts
export type Unknownify<T> = {
  [key in keyof T]?: unknown;
};
```
I think the version here without `?` is more appropriate. How about removing `?` in the defined type in `src/types.ts` and then reusing it here?

src/Netscript/APIWrapper.ts
```diff
@@ -73,13 +76,12 @@ class NSProxyHandler<API extends GenericAPI<API>> {
       const functionPath = arrayPath.join(".");
       const ctx = { workerScript: this.ws, function: key, functionPath };
       // Only do the context-binding once, instead of each time the function
-      // is called.
-      const func = field(ctx) as (...args: unknown[]) => unknown;
+      // is called. It is stored in the closure.
```
catloversg: I think we should remove this comment now. It describes creating `func` as a context-bound version of `field` by calling `field(ctx)` once and reusing the result. After this refactor, `field` is called directly with `ctx` as the first argument, so there is no separate context binding step anymore.

src/Netscript/APIWrapper.ts
d0sboots: "to work"

Perhaps a less ambiguous wording would be "it needs to be its own generic type to function"?

src/Netscript/APIWrapper.ts
d0sboots: They are used for different purposes, and I think the version with `?` makes sense for the object use that is imported from types. Specifically, those users (mostly in NetscriptHelpers) are taking a type and saying, "We don't know anything about the shape of this, not even that these fields exist. But we want to be able to access the fields without TS complaining." Here we want to map over an array, so `?` is inappropriate.

src/Netscript/APIWrapper.ts
d0sboots: Done.

d0sboots:
Because resolving conflicts with this PR is a *huge* pain, I'm not going to bother unless/until I hear that it's generally good to go and it's ready to submit.

src/Netscript/APIWrapper.ts
catloversg: To avoid naming confusion, let's define:
```ts
export type PartialUnknownify<T> = {
  [key in keyof T]?: unknown;
};
export type Unknownify<T> = {
  [key in keyof T]: unknown;
};
```
Can you clarify your opinion a bit more? I agree they are used for different purposes, but I don't understand why `PartialUnknownify` is useful in our current use cases, especially in `NetscriptHelpers`(`foo as Unknownify<Foo>`). I've tried replacing `PartialUnknownify` with `Unknownify` in those validators, and I don't see any issues. Can you give an example where it would make a difference?
> Specifically, those users (mostly in NetscriptHelpers) are taking a type and saying, "We don't know anything about the shape of this, not even that these fields exist. But we want to be able to access the fields without TS complaining."

```ts
export function validate(input: object) {
  const a = input as Unknownify<{foo: number}>;
  const b = input as PartialUnknownify<{foo: number}>;
  a.foo;
  b.foo;
}
```
We can access `a.foo` and `b.foo`, and both of them are typed as `unknown`.

The difference between `Unknownify` and `PartialUnknownify` mainly shows up in assignability. For example:
```ts
interface Foo {
  a: number;
  b: string;
}
const p: PartialUnknownify<Foo> = {
  a: 123,
};
// OK

const q: Unknownify<Foo> = {
  a: 123,
};
// Error: missing b
```
Another legitimate use case for `PartialUnknownify`:
```ts
function mergeDefaults(
  overrides: PartialUnknownify<Foo>,
): Foo {
    ...
}
```
where callers can write:
```ts
mergeDefaults({});
mergeDefaults({ a: 1 });
```

However, our current use cases are different. We use `const options = value as Unknownify<Foo>;`, where the type is only used to enable property access after a runtime type check. In that pattern, I don't see an observable difference between `Unknownify` and `PartialUnknownify`.

I think it makes sense to define both `Unknownify` and `PartialUnknownify` in `src/types.ts`. Whether the validators in `NetscriptHelpers` should use `Unknownify` or `PartialUnknownify` is still open to discussion, but I don't currently see a practical advantage to `PartialUnknownify` there.

d0sboots: OK. I think our original usage was under some of the same mistaken impressions I was.

However, there's still a naming issue here. I don't like PartialUnknownify because it sounds like it's only applying partially - that only some of the fields will be unknown'd. I honestly can't think of any good name for the second one. This is because both have a valid claim to be called "Unknownify" - one makes the fields unknown, the other also makes the *shape* unknown. 

I will try just changing the definition, because I think the best case is to only have one.

d0sboots: Yup, looks like that works. Any other concerns?

catloversg: No. That's the last concern.

catloversg:
It looks good to me. You can resolve the conflicts now.

d0sboots:
I did a basic smoke-test after the merge (ran `eval.js` to ensure that my fix to setRemovedFunctions was working right.)

PR #2921, opened 2026-06-25
UI: Add support for `cat` printing to terminal

E9cipher:
This PR adds support for `cat` being able to print to the terminal via setting the `-t` flag. This can prove useful later, for example if pipe support is added. Tested manually.

Note that this only acts on the terminal; i.e. the singularity API for cat has not been altered. I can add support for this in the API as well but imo it's fine being only in the terminal.


d0sboots:
This is already part of the pipes PR; that changes cat to go to a pipe if its output is piped.

E9cipher:
Oh! Well, in that case I'll close this one and wait for the pipes support!

ficocelliguy:
For context: #2525

PR #2926, opened 2026-06-27
NETSCRIPT: Add top-level Netscript Port APIs for empty and full checks

NagaOuroboros:
## Motivation
Currently, it is only possible to directly check if a port is full or empty by manually grabbing a `PortHandle` first, then calling `full()`/`empty()`.  This PR adds those checks into the top-level NS API.

## Scope
Simply adds the required methods:
- `portIsFull(portNum: number): boolean`
- `portIsEmpty(portNum: number): boolean`

## Considerations
- With this change, there will be no need to create a `PortHandle` unless the player deliberately wants/needs one.
- Players can now be directed to the top-level APIs by default, and `PortHandle` can be reserved for its specialized functionality, instead of needing to mix the two just to get access to all port-related methods.

Tested on latest version of 3.0.2dev


catloversg:
These APIs should be called `isFullPort` and `isEmptyPort`.

NagaOuroboros:
> These APIs should be called `isFullPort` and `isEmptyPort`.

Done!  Dunno why I didn't abide by that convention in the first place

PR #2928, opened 2026-06-27
BUGFIX: Darknet server connections are drawn incorrectly

### LINKED ISSUE #2920: Darknet UI draws connections to hidden servers (again)
#2868 tried to fix this, but I'm still experimenting the issue. I have the latest dev and I have definetly soft-reseted after the PR got merged.

Image

I can't seem to reproduce this again, I have tried a few webstorms but nothing so far. I will continue trying and post any updates here.

catloversg:
This PR fixes 2 issues.

MRE 1: Load `bug-1.gz` and open the dnet tab.

When drawing the connections, we compare the server depth to `netDisplayDepth`, but with the lab server, `server.depth` is always -1, so we need to check `labDepth` instead.

MRE 2:
- Apply the patch in `networkCanvas.ts`, but revert the patch in `NetworkDisplayWrapper.tsx`:
  - `drawOnCanvas(canvas.current, netDisplayDepth, labDepth)`
  - `}, [rerender, netDisplayDepth, labDepth]);`
- Load `bug-2.gz`.
- Open the dnet tab. Make sure you can see both `d0s_slippers` and the lab server.
- Run `GetServer("d0s_slippers").hasAdminRights=false;DarknetEvents.emit()` to simulate a network mutation that resets `d0s_slippers`.
- Do NOT touch the UI. Notice the wrong connections.
- Touch the UI. Notice how the wrong connections disappear.

https://github.com/user-attachments/assets/2ab9222d-407a-4497-ae33-34f6d2d00f80

I'm not sure how to summarize this issue and the fix. You would need to add tons of `console.log` to `updateDisplay` and the following `useEffect` block to see why it happens.

Test save files:
[bug-1.gz](https://github.com/user-attachments/files/29412581/bug-1.gz)
[bug-2.gz](https://github.com/user-attachments/files/29412578/bug-2.gz)

Closes #2920.

### DIFF
--- a/src/DarkNet/ui/networkCanvas.ts
+++ b/src/DarkNet/ui/networkCanvas.ts
@@ -12,7 +12,7 @@ import { NET_WIDTH } from "../Enums";
 import type { DarknetServer } from "../../Server/DarknetServer";
 import { getDarknetServerOrThrow } from "../utils/darknetServerUtils";
 
-export const drawOnCanvas = (canvas: HTMLCanvasElement, netDisplayDepth: number) => {
+export const drawOnCanvas = (canvas: HTMLCanvasElement, netDisplayDepth: number, labDepth: number) => {
   const ctx = canvas?.getContext("2d");
   if (!ctx || !canvas) {
     console.error("Could not get canvas context");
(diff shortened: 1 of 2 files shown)


ficocelliguy (approved):

src/DarkNet/ui/networkCanvas.ts
```diff
@@ -23,17 +23,20 @@ export const drawOnCanvas = (canvas: HTMLCanvasElement, netDisplayDepth: number)
...
       (!server.hasAdminRights && !server.serversOnNetwork.find((s) => getDarknetServerOrThrow(s).hasAdminRights))
     ) {
       continue;
     }
 
-    // draw a line between each server and its connected servers
+    // Draw a line between each server and its connected servers
     for (const connectedServerName of server.serversOnNetwork) {
       const connectedServer = getDarknetServerOrThrow(connectedServerName);
+      // With labyrinth servers, server.depth is always -1, so we need to check labDepth.
```
d0sboots, line 36: Could we change this? It seems very weird to have the lab claim that it's depth -1.

catloversg: Maybe, but I'll need @ficocelliguy to confirm his design rationale. Based on what I understand from the code, fico uses `depth` for different things.

We have `DarknetState.Network`, which is a 2D map of "normal" dnet servers (non-darkweb, non-lab). This map is used whenever we need to work with a server's coordinates (e.g., mutating the network or drawing the UI). The darkweb and all lab servers are special servers and are not on this map so that they don't "collide" with non-lab servers. The "real" lab depth is usually used when we draw the UI or calculate balancing factors (e.g., `requiredCharismaSkill`).

This is more of a design choice than a real limitation, so changing it is fine. However, we'll need to be very careful. It seems like a lot of code assumes that lab servers receive special treatment and are not on the grid (`DarknetState.Network`).

d0sboots: Yes, I wasn't expecting to put lab servers on the network itself. Unless there's an assumption that server.depth !== -1 means it's on the network?

ficocelliguy: yeah, I use -1 for special servers not intended to be on the grid (all the labs are there, even though only one is "active"), and servers not yet placed on the grid (e.g. created but not yet moved in)

We could make lab be informed of its own depth, but I suspect that would require changes in some other places too, since its depth is over the max and also I use the depth -1 in some places iirc

PR #2930, opened 2026-06-27
DOCUMENTATION: Clarify that port handle methods remain usable after the ns instance is dead

catloversg:
This behavior is not documented currently. Before I finish the port documentation and the technical deep dive into how "dead" scripts work in Bitburner, let's add a small clarification to the port API documentation.

### DIFF
--- a/markdown/bitburner.netscriptport.md
+++ b/markdown/bitburner.netscriptport.md
@@ -6,6 +6,8 @@
 
 Object representing a port. A port is a serialized queue.
 
+All methods in this interface can be used even after the ns instance is dead.
+
 **Signature:**
 
 ```typescript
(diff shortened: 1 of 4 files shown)


markdown/bitburner.netscriptport.md
```diff
@@ -6,6 +6,8 @@
 
 Object representing a port. A port is a serialized queue.
 
+All methods in this interface can be used even after the ns instance is dead.
```
d0sboots: Suggested here and everywhere:
```suggestion
All methods in this interface can be used while the ns instance is "busy" (they avoid the concurrency check), or even when it is dead.
```
Avoiding the concurrency check is often the more relevant part.

PR #2931, opened 2026-06-27
DOCUMENTATION: Mark host parameter of `ns.ls` as optional

E9cipher:
This PR makes `ns.ls()` default to current host if the `host` param is undefined. Tested manually

### DIFF
--- a/src/NetscriptFunctions.ts
+++ b/src/NetscriptFunctions.ts
@@ -827,7 +827,7 @@ export const ns: InternalAPI<NSFull> = {
     }
     return noFailures;
   },
-  ls: (ctx) => (_host, _substring) => {
+  ls: (ctx) => (_host = ctx.workerScript.hostname, _substring) => {
     const [server] = helpers.getServer(ctx, _host);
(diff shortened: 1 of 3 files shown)


src/NetscriptFunctions.ts
```diff
@@ -827,25 +827,27 @@ export const ns: InternalAPI<NSFull> = {
...
   },
-  ls: (ctx) => (_host, _substring) => {
-    const [server] = helpers.getServer(ctx, _host);
-    const substring = helpers.string(ctx, "substring", _substring ?? "");
-    if (!server) {
-      return [];
-    }
+  ls:
+    (ctx) =>
+    (_host = ctx.workerScript.hostname, _substring) => {
```
d0sboots: This won't work for `ls(undefined, "foo")`. Do it the same way substring works.

catloversg:
There is no need to change the implementation of `ls`. The first parameter is already optional (check how `helpers.getServer` is implemented).

E9cipher:
Oh yes, you're right: `getServer` already handles the logic. So I'll update the definitions and documentation then.

PR #2937, opened 2026-06-30
UI: Make coding contract successful attempts print to terminal

E9cipher:
MRE: Create a dummy coding contract. Solve it manually via the UI.
There is no message indicating the success, although if you fail it there is an error that indicates failure.

I have modified the printing logic via ternary operators (cleanest), but I can do it as an if-else block if you wish me to. Tested manually.


E9cipher:
Let's try to merge #2939 instead of this one as a favor to @ultragaming1001. I'll leave it open as a fallback

PR #2939, opened 2026-06-30
UI: Always print result to terminal when manually solving contract succeeds

UltraGaming1001:
This commit/pull request removes the requirement for the Terminal confirmation for Coding Contracts solved manually to be only for normally-generated/natural Coding Contracts. Now, when successfully solving a Dummy Coding Contract manually, instead of the .cct file being removed with no confirmation that the user has actually correctly solved it, it now says:
```
Contract SUCCESS - No reward for this contract
```


catloversg:
Between this PR and #2937, I prefer this one. `gainCodingContractReward` already handles the null reward case, so there is no need to check for it beforehand.

PR #2944, opened 2026-07-02
DNET: Remove hard-coded styles

### LINKED ISSUE #2851: Some hardcoded white text is unreadable in light theme
This happens in some dialogs and possibly elsewhere.  There are some `<div>` elements with an inline style of `color: white;` set.  This is unreadable in light themes.

For now, I'm applying a user agent stylesheet like this as a workaround.

```
[style*="color: white;"] {
    color: unset !important;
}
```

ficocelliguy:
This PR removes the hard-coded styles in the Darknet component and uses theme-based colors instead.

This allows the flashbang theme to still have the text be readable, and standardizes the look and feel to better match the rest of the UI for other themes.

It also adds an icon and border color for frozen servers, and improves the coloring of servers that have a connection but need authentication still.

Closes #2851 

image

image

image


d0sboots:
Someday I intend to change how styling works in dnet, because I investigated it and I think it should be using one of 6 other ways styling can happen, instead. XD But that will wait for another day.

PR #2945, opened 2026-07-02
API: Expose GangTaskName to NSEnums

E9cipher:
This PR exposes `GangTaskName` to NSEnums. Test code:
```javascript
ns.tprint(Object.keys(ns.enums.gangTaskName))
ns.gang.setMemberTask(ns.enums.gangTaskName.humanTrafficking)
```

@catloversg do you prefer merging this first and refactoring the gang code in a separate PR or commit here and merge it all together?


hexagonrecursion:
I am confused. In your example code it's `ns.enums.gangTaskName` but in `src/NetscriptFunctions.ts` it's `ns.enums.GangTaskName` (the first letter is capitalized). `GangTaskName` looks correct to me because it is consistent with all other properties of `ns.enums`

catloversg:
This needs some changes. I'll commit directly to your branch.

E9cipher:
> I am confused. In your example code it's ns.enums.gangTaskName but in src/NetscriptFunctions.ts it's ns.enums.GangTaskName (the first letter is capitalized).

Yes, sorry. I don't know why I capitalized it (the correct version) and then lowercased it in the enums :( . I don't want to interfere with catlover's current work, so maybe I'll leave up to catlover?

src/Gang/Enums.ts
```diff
@@ -0,0 +1,29 @@
+export enum GangTaskName {
```
d0sboots: I thought for new exported enums we were doing plain objects, not TS enums? Or was I misremembering?

E9cipher: I looked at the Programs/ enum and it's a plan TS enum like this...
Speaking of which, I was thinking of a PR changing all the enum to `as const`. What do you think?

d0sboots: Changing all the internal enums would be a lot. However, I thought we were introducing new ones as const objects. (This is a question for CatLover.)

catloversg: > I thought for new exported enums we were doing plain objects, not TS enums? Or was I misremembering?

I don't remember whether we established that rule. Anyway, I'm fine with that.

catloversg: > I was thinking of a PR changing all the enum to as const. What do you think?

Don't do that. You can check #2001 to see why.

PR #2946, opened 2026-07-02
UI: Allow access to grafting UI outside New Tokyo

catloversg:
Previously, the grafting UI and the `ns.grafting.graftAugmentation` API could only be used when players were in New Tokyo. As discussed on Discord, it's fine to remove this requirement for the grafting UI, so players can now open that UI to check grafting information from anywhere.

Note that the location requirement is still enforced in the API.

I also added a tooltip to the grafting UI and fixed a race condition.

@d0sboots There are two things to consider.

First, when players work for VitaLife, the Jobs tab shows the "Enter the Secret Lab" button. This button implies that they are in New Tokyo (VitaLife is in that city), so it seems strange to show the button when players are not in New Tokyo. This isn't really an issue now that we've removed the location requirement. The Bladeburner UI works the same way (the NSA job always shows the "Enter Bladeburner Headquarters" button).

Second, when players are not in New Tokyo, we throw an error, but we return false when they do not satisfy other requirements. I think we should always return false to be consistent here. It's a minor breaking change, but I don't think it will be a problem.

### DIFF
--- a/src/PersonObjects/Grafting/ui/GraftingRoot.tsx
+++ b/src/PersonObjects/Grafting/ui/GraftingRoot.tsx
@@ -1,11 +1,12 @@
 import type { Augmentation } from "../../../Augmentation/Augmentation";
 
 import { Player } from "@player";
-import { AugmentationName } from "@enums";
+import { AugmentationName, CityName } from "@enums";
 
 import React, { useState } from "react";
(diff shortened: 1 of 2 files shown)


d0sboots:
> Second, when players are not in New Tokyo, we throw an error, but we return false when they do not satisfy other requirements. I think we should always return false to be consistent here. It's a minor breaking change, but I don't think it will be a problem.

I'm fine with that. I merged this so you can do it as a separate PR.

PR #2948, opened 2026-07-03
API: Add ns.ui.createConnectLink to create clickable server links

### LINKED ISSUE #1640: API Feature Request: After access to AutoLink.exe allow script use of links

hexagonrecursion:
* closes #1640

Adds `ns.ui.createConnectLink(connectPath: string[], linkText?: string) -> ReactNode`. This function receives an array of host names or ip addresses and returns a clickable link. Clicking on the link is equivalent to typing a sequence of `connect` commands, except that if it encounters an error it will return you to the server you started from.

## Demo

```tsx
  function find(path: string[]) {
    const host = path.at(-1)!;
    if(path.length > 3 || visited.has(host)) return;
    visited.add(host);
    map.push(path);
    for(const s of ns.scan(host)) {
      find([...path, s]);
    }
  }
  const visited = new Set();
  const map: string[][] = [];
  find(['home']);
  const space = (d: number) => {
    let res = '';
    while(--d >= 0) {
      res += ' ';
    }
    return res
  }
  ns.tprintRaw(
    <>
      {map.map((p) => <div>{space(2*p.length-2)}{ns.ui.createConnectLink(p)}</div>)}
    </>
  );
}
```

image

## ~The player does not need AutoLink.exe in order to use ns.ui.createServerLink()~

~Justification: we want built-in tools (like scan-analyze) to be clunky and limited in order to encourage players to write their own.~

Update:  AutoLink.exe is now required, but DeepscanV1.exe and DeepscanV2.exe are not required.

## ~We only check that the target server exists when the link is clicked~

1. When the API is called, we check the provided array of host names and throw an error if they do not exist.
2. When the link is clicked, we check, whether the sequence of connects commands is valid.

* ~It is possible to create a link to a server that does not exist yet.~
* ~We check whether the target server exists when the link is clicked.~

~justification:~
1. ~Checking the target server when the link is clicked will be necessary no matter what because a server can be removed.~
3. ~I think that creating a link to a server that does not yet exist is a niche, but valid use case.~

## The link can not be clicked programmatically

This is to avoid making `ns.singularity.connect()` obsolete.

This will print an error to the terminal:
```ts
  const div = globalThis['document'].createElement('div');
  ReactDOM.render(
    ns.ui.createConnectLink(['home']),
    div,
  );
  div.childNodes[0].click();
}
```

## Optional linkText

The text on the link does not have to match the host name or IP address of the target server:

```ts
  ns.tprintRaw(ns.ui.createConnectLink(['home'], 'here be dragons'));
}
```

## ~Doubts about special servers~

~Currently, `ns.ui.createServerLink()` can be used to connect to any server as long as the player knows its IP or host name:~

We now validate the full path to the destination.


catloversg:
I'll review this PR after d0sboots gives his opinion. My high level comments:

> We only check that the target server exists when the link is clicked

I prefer checking it immediately, but I'll let d0sboots make the call.

> The link can not be clicked programmatically

I won't veto this, but I don't think it's necessary.

> Currently, ns.ui.createServerLink() can be used to connect to any server as long as the player knows its IP or host name. I am not sure whether this is a problem or not.

It's absolutely an issue.

This API is a convenient tool, so it needs to be less powerful than what players can build with singularity APIs and simple DOM interaction. Without serious exploits such as prototype manip or module raid, players cannot access isolated servers (pre-TRP WD and non-TOR darkweb) and isolated dnet servers (dnet servers that have no connection). This means this tool must not be able to do so either.

My guideline for this kind of API is to check what the equivalent CLI command can do (`scan-analyze` in this case), then make the API do that or less, not more.

In terms of implementation, you can call `helpers.getServer(ctx, _host)` to filter out non-existent servers and isolated non-dnet servers. Then filter out non-darkweb dnet servers.

hexagonrecursion:
> > The link can not be clicked programmatically
> 
> I won't veto this, but I don't think it's necessary.

I could remove the check and instead add "Programmatically simulating a click on the link created by this function is considered an exploit" to the function documentation (once I write it). Would this be better?

> It's absolutely an issue.

I have added access-control to the links.
1. If a server is a DarknetServer other than `darkweb`, you can connect to it iff it has a stasis link or a backdoor.
1. For all other servers: you can connect to it iff `server.serversOnNetwork` is not empty. This prevents connections to w0r1d_d43m0n without the red pill and connections to `darkweb` without tor

d0sboots:
High level: The overall API sounds good. I agree with the points catlover raised. You shouldn't even be able to get a link to world daemon until it's accessible, because it's not on the network yet. 

Much like the existing ls functionality, I think the "get a link" part should be restricted by what you can *currently* do - I'd propose any server that you have a direct connection+auth to, in dnet terms. (This is irrelevant for the normal network, where you can reach anywhere.) And then the "does the link work" part is only limited by the server still existing. 

This would allow for a more "permanent" type of dnet backdoor based on printing links after getting auth, with the trade-off that it's strictly manual. That feels like it fills a hole, to me.

catloversg:
> Much like the existing ls functionality, I think the "get a link" part should be restricted by what you can currently do - I'd propose any server that you have a direct connection+auth to, in dnet terms. (This is irrelevant for the normal network, where you can reach anywhere.) And then the "does the link work" part is only limited by the server still existing.

> This would allow for a more "permanent" type of dnet backdoor based on printing links after getting auth, with the trade-off that it's strictly manual. That feels like it fills a hole, to me.

Can you clarify a bit about the "direct connection+auth" condition? I think we need to be very precise about the condition here. Let's say the network looks like this:
```
home => darkweb => foo (rooted) => - bar (rooted)
                                   - baz (rooted + backdoor)
```
Players can connect to both bar and baz normally via the terminal or the singularity API. However, if bar and baz lose all their connections and become isolated, players will be able to connect only to baz (due to the backdoor).

IMO, being manual only is not really a big downside, so I'm strongly against anything that does more than the terminal `connect` and `ns.singularity.connect` allow.

ficocelliguy:
Finding the path to a server is an interesting puzzle for players (and good practice navigating trees, if they don't skip that part of puzzle with abusing scan[0]). I really dislike the idea of giving players the option to just skip that entirely with just a `ns.ui.createServerLink()` call. Remember we are a puzzle game at heart - we give players interesting problems and a cool sandbox to explore them in. We don't want to remove puzzles via quality-of-life changes.

If it were up to me I would veto this entirely. We want to give players tools (like printRaw + singularity connect), not solutions wholesale. At the very least, this should go under a source-file unlocked API like singularity, so that players can earn it later and not skip puzzles from day 1.

I also will insist that this does not work on darkweb servers. Navigating the darkweb is another puzzle that shouldn't be skipped with a pre-packaged ns call.

hexagonrecursion:
tldr:
my current opinion:
1. dnet links that connect w/o backdoors/statis links (suggested by d0sboots) are not OK. Not even manual-only
2. dnet links that connect only to backdoored/statis linked servers are OK because the `connect` terminal command would succeed too

---

> I also will insist that this does not work on darkweb servers

The current implementation of `ns.ui.createServerLink()` allows links to dark net servers, but the links _only_ work if the target is backoored and/or stasis linked at the point in time when the link is clicked. Clicking on a link to a backoored/stasis linked dark net server succeeds, which seams reasonable to me because a connect command would also succeed (both clicking a link and typing a command are manual actions).

> This would allow for a more "permanent" type of dnet backdoor based on printing links after getting auth, with the trade-off that it's strictly manual. That feels like it fills a hole, to me.

After giving it some thought, I agree with catloversg: finding ways to connect your terminal to air gapped servers deep in the dark net and/or finding ways to interact with them through your scripts without needing a connection appears to be a major part of the challenge of the dark web. There are many fun solutions to this puzzle. You could, for example, have scripts on each dnet server check a port and create a stasis link on-demand. Or you could do something more involved: create a custom UI in the tail window that lets you give commands to the script without needing a terminal connection. I fear that clickable links that always work as long as the destination still exists would trivialize most of the challenge. Manual-only is not a sufficient downside to justify "permanent type of dnet backdoor based on printing links" because when you want to connect your terminal to a dnet server it's usually so that you can do something manual on the server (e.g manually stop or start a script).

> You shouldn't even be able to get a link to world daemon until it's accessible, because it's not on the network yet.

I fail to see the harm. `ns.ui.createServerLink()` allows the player to create a link to any string of letters. It does not have to be a real host name. Clicking on the link will tell the player that the server does not exist. Until you get the red pill, clicking a link to w0r1d_d43m0n behaves exactly as if w0r1d_d43m0n does not exist.

> Finding the path to a server is an interesting puzzle for players

That is a good point. I could change the API to require the caller to supply an array that describes the full path from `home` instead of just the destination host name. This preserves the puzzle. What do you think?

ficocelliguy:
I really like requiring a nav path - in fact I would be fine with connecting to darknet servers too given that (if there is a chain to get there), since it would be a fun puzzle to get that info. (Darknet backdoor and stasis link are both temporary, so they aren't really stable/meaningful restrictions). It also solves the "permanent link" issue since its only valid as long as the provided path is valid.

Backdoors also could be used as links on the chain to jump around, since you can connect directly to backdoored servers via singularity or the terminal.

For consistency with other API methods, giving players feedback immediately if the hostname(s) they input is invalid is a good idea, IMO, rather than waiting for them to actually click on it. It could also validate that the chain presented is valid / connected.

hexagonrecursion:
@d0sboots @catloversg I'll implement the path-based API if you say it's OK (don't want to spend time re-doing this if there is a risk you might tell me to re-do it again). If you agree that the path-based API is a good idea, I will also try to refactor `ns.singularity.connect()` and `commands/connect.ts` to reduce code duplication.

catloversg:
To be honest, this API is in an awkward situation. On one hand, it violates the "DIY" philosophy of Bitburner (we give players tools, namely NS APIs, and it's their job to use those tools). On the other hand, printing clickable links that utilize the connect CLI/API is not exactly easy, especially for players with very limited programming knowledge. Personally, I don't think we need to add this API, but I'm not opposed to it strongly enough to veto it immediately.

If we add it, we should put some constraints on it. The simplest constraint is to gate it behind other features (SF), but doing that defeats the entire goal of this API, which is to help new players. Forcing players to provide a path to the destination looks like a good enough constraint (they need to prove that they have a scanner), but I would like to hear d0sboots's opinion.

Some notes:
- The path should be a string array. We will validate this path at both creation time (when the API is called) and runtime (when players click on the link).
- You say that you will refactor ns.singularity.connect() and commands/connect.ts to reduce code duplication. This makes me worry a bit. There is certainly some duplicate code between them, but refactoring them is not what you should do in this PR. If your new API needs to use that code, you can do so, but you should ensure that you actually need to. The connect CLI and API only connect to an adjacent server, so their use case is different from this API.
- I prefer leaving out the entire dnet network in this feature. Personally, I consider dnet an advanced and optional feature. Although players can access a simple version of it pre-endgame, it's not an easy challenge. There are many things about dnet that I don't want to "dumb down". If players want to conveniently navigate dnet via a clickable link, they should learn how to use `printRaw` and `connect` (or use terminal injection). We only need to help them in the main network.

d0sboots:
Lots of activity here while I was taking a mini vacation!

So, for starters I think you should be able to form/follow a link to any server you have backdoored, which includes stasis links (since they are also backdoored). IMO this should be uncontroversial; you can directly connect to them from anywhere, so this is not adding *any* power or subverting any challenges. 

Beyond that, to me it makes sense that you can form a link to any regular (non-dnet) server that is "accessible" (i.e. on the network). Why? Because in order to get the hostname for that server, you must have recursively scanned to it *somehow* - whether via your own program, it the built-in programs. And the purpose of this functionality is to allow people to make a better version of scan-analyze, which already makes these links. IMO it does not make sense to "prove" that you know the path from home - no other API we have has such a needless requirement. Knowing the server's name is always sufficient.

I would be fine with locking this behind the autolink program, which is what gates this for scan-analyze and is available for not too much $$$ in early game.

d0sboots:
Another thing worth noting (and possibly considering for changes): `scan-analyze` with autolink works with darknet servers just fine currently:
image
You can connect to all of them, regardless of auth status. Probably autolinks should be restricted to servers you have authed, or servers that are backdoored. Another potentially reasonable change would be for `scan-analyze -a` to only follow authed links on dnet servers.

I *don't* think scan-analyze should be completely removed for dnet. I also don't think this is evidence that the feature proposed here needs to work for all dnet servers - even in its current (IMO broken) form, scan-analyze is limited to a depth of 10, and `createServerLink` does not/cannot have such a restriction.

catloversg:
> Because in order to get the hostname for that server, you must have recursively scanned to it somehow - whether via your own program, it the built-in programs.

Nitpick: You can get the hostname of all non-isolated servers in the main network via autocomplete (data.servers). 

I get your points that in most cases, if players know a hostname, they should already have a way to get the path to that host, either manually or programmatically. I agree that requiring a path to use the API is not the best way to restrict it, but this API is too useful (it bypasses the need to learn how to use printRaw and connect), so I still want it to be restricted somehow.

Gating it behind the autolink program is okay, but I want to have some more restrictions, even if those restrictions are not ideal or somewhat questionable. How about raising the RAM cost to a high value? My point is that if players want to enjoy the convenience of using this API instead of learning UI stuff, they must pay a price.

I don't like the idea of letting this API work with dnet servers. If this API also works with dnet servers (authed/backdoored only), I would like the "price" to be even higher.

> Another thing worth noting (and possibly considering for changes): scan-analyze with autolink works with darknet servers just fine currently.

I almost never use `-a`, so this is the first time I have noticed that `scan-analyze` can scan and allow connecting to other dnet servers. I agree that autolinks should be restricted to servers you have authed or servers that are backdoored.

d0sboots:
> I get your points that in most cases, if players know a hostname, they should already have a way to get the path to that host, either manually or programmatically. I agree that requiring a path to use the API is not the best way to restrict it, but this API is too useful (it bypasses the need to learn how to use printRaw and connect), so I still want it to be restricted somehow.
> 
> Gating it behind the autolink program is okay, but I want to have some more restrictions, even if those restrictions are not ideal or somewhat questionable. How about raising the RAM cost to a high value? My point is that if players want to enjoy the convenience of using this API instead of learning UI stuff, they must pay a price.
> 
> I don't like the idea of letting this API work with dnet servers. If this API also works with dnet servers (authed/backdoored only), I would like the "price" to be even higher.

Hmm. We are at a bit of an impasse, because I feel the opposite in a pretty fundamental way.

To me, this API is not that useful at all. It is a strictly manual tool - in order to make use of it, the player must *click* on the links it provides. This puts it on the same level as the "connect" command from the terminal, which is completely free and available from the start of the game, in my mind.

The main difference is the "backdoor-like" functionality: That the link it provides is a *direct* link, as opposed to connect requiring a path through servers. You can circumvent this either with the backdoor command (again, available at the start of the game), the singularity function, or by constructing a recursive path and making "aliases" so that you can easily connect directly to a server without having to type the whole path.

To me, the fact that this is a UI API *is* a cost. I would personally not use it because I find UI stuff annoying; using this requires working with `printRaw` or `tprintRaw`. To me, creating a connect string is simpler. And I think that's a good thing! Having a diversity of approaches is what we strive for in this game.

d0sboots:
Despite that, I'm not against making this somewhat pricey. Like maybe 5GB level - a chunk of change for 8GB starting server, but not out of reach, just requires some planning around.

catloversg:
We agree on two things: gating this API behind autolink.exe and setting the RAM cost to a high value (5GB).

Regarding the dnet servers, what is the requirement? Previously, we talked about being authed or having a backdoor. In this example:
```
home => darkweb => foo (rooted) => - bar (rooted)
                                   - baz (rooted + backdoor)
```
If bar and baz lose all their connections and become isolated, players will be able to connect only to baz (due to the backdoor). I suggest only allow autolinking to dnet servers that have a backdoor (darkweb is an exception; players can always autolink to it).

Technically, we can check if a dnet server is truly reachable by checking paths from darkweb and other backdoored servers, but we can skip that and use the backdoor-only limitation as an additional restriction for dnet servers. I think this is a good compromise between our differing opinions.

ficocelliguy:
Note that the backdoor / stasis link will have to also be checked when the link is clicked, since backdoors can be removed from darknet servers.

I still personally think that asking for a connection chain is a good idea. It's an interesting puzzle (different than just getting the list of server names), gracefully solves the issues with potentially allowing this in darknet, and contains the added QOL to the UI aspects and not the mechanical ones.

d0sboots:
> Regarding the dnet servers, what is the requirement? Previously, we talked about being authed or having a backdoor. In this example:
> 
> ```
> home => darkweb => foo (rooted) => - bar (rooted)
>                                    - baz (rooted + backdoor)
> ```
> 
> If bar and baz lose all their connections and become isolated, players will be able to connect only to baz (due to the backdoor). I suggest only allow autolinking to dnet servers that have a backdoor (darkweb is an exception; players can always autolink to it).

Yes, that is the conclusion I came to as well - that for this API and dnet, it only makes sense to support servers that are backdoored. I hadn't considered the wrinkle that backdoor can be removed, but like other checks it makes sense that it would need to be checked both at creation time and follow time.

I have a compromise to propose on the "prove a path" API. Instead of needing a full path, take only a single "parent" hostname. The way the check would actually be implemented is that "parent" contains "server" in its connections - this naturally handles regular connections, backdoors (via home), and WD.

For most situations, this requires you to use scan, although not recursively. IMO requiring a path-following scan is too specific/limiting of a requirement for this, and that is also a puzzle that will be solved for other purposes anyway. I know you are concerned about people "skipping" that puzzle, but I think having the flexibility to approach the game in different ways is important. Anyone who skips network scan definitely won't be able to tackle darknet; what I'm considering is not the experience of people who will play to the end and pick up everything anyway, but those who are quitting partway through (probably before endgame). Assuming the game wasn't *completely* against their taste, I want them to have fun too.

UltraGaming1001:
This would be way too overpowered for any `scan-analyze` greater than 10 (considering the fact that it can even access forbidden/non-existent servers). I haven't read all the changes, nor most of the comments, but I think the depth to which you can click should be limited by the DeepScan program, the same way as how it goes for scan-analyze. And if the player does not have the DeepScan program, then it should be limited to a depth of 3 (again, the same how `scan-analyze` is).

I see use in this in the fact that you can create a direct clickable link to a server instead of using `scan-analyze 10` and having to slowly scroll and carefully look at each and every server to try and connect to the right one.

Also, I think this should ONLY work if you have AutoLink.exe
Just straight access to this function would not really encourage new players to make an auto-connect function/script of their own, but instead just depend on this one.

hexagonrecursion:
@UltraGaming1001 

> considering the fact that it can even access forbidden/non-existent servers

We have already agreed to change that. Read the comments.

> I think the depth to which you can click should be limited by the DeepScan program, the same way as how it goes for scan-analyze. And if the player does not have the DeepScan program, then it should be limited to a depth of 3 (again, the same how scan-analyze is).

I strongly disagree: the `scan-analyze` command is intentionally limited in order to encourage players to write their own better tools. I do agree with gating this API behind AutoLink.exe, but I think that preventing links to servers at depths greater than 10 is too big of a restriction.

@d0sboots

> I have a compromise to propose on the "prove a path" API. Instead of needing a full path, take only a single "parent" hostname

This sounds link the worst of both worlds because there is no puzzle: if you have the host name of the destination, it is trivial to obtain a host name of an adjacent server.
   ```ts
     const destination = 'avmnite-02h';
     const maybeNeighbor: string | undefined = ns.scan(destination)[0];
     const neighbor = maybeNeighbor ?? 'home'; // for dnet servers with backdoors and/or stasis links
     const linkText = destination;
     ns.ui.createServerLink(neighbor, destination, linkText);
   }
   ```
In my understanding, a puzzle is something that is at least a tiny bit non-trivial. Like this, for example:
   ```ts
   // Does not handle dnet
     const destination = 'avmnite-02h';
     const path = [];
     for(let h = destination; h !== 'home'; h = ns.scan(h)[0]) {
       path.unshift(h);
     }
     path.unshift('home');
     const linkText = destination;
     ns.ui.createServerLink(path, linkText);
   }
   ```

d0sboots:
> > I think the depth to which you can click should be limited by the DeepScan program, the same way as how it goes for scan-analyze. And if the player does not have the DeepScan program, then it should be limited to a depth of 3 (again, the same how scan-analyze is).
> 
> I strongly disagree: the `scan-analyze` command is intentionally limited in order to encourage players to write their own better tools. I do agree with gating this API behind AutoLink.exe, but I think that preventing links to servers at depths greater than 10 is too big of a restriction.

I agree. I don't think any of us have been considering an arbitrary numeric limit, because it doesn't make sense.

For one thing, this is an NS api, so where would you even make the limit be *from*? It can't really be the server the player is connected to, because this is intended to work without singularity, where you have no knowledge of that through NS. Having it be a limit from home makes it *more* restricted than scan-analyze. Having it be a limit from the running server is both a pain to implement, and also means you can still link the whole net by executing scripts on key servers (which only requires port crackers).

> > I have a compromise to propose on the "prove a path" API. Instead of needing a full path, take only a single "parent" hostname
> 
> This sounds link the worst of both worlds because there is no puzzle: if you have the host name of the destination, it is trivial to obtain a host name of an adjacent server.

That is the point: it's meant to be trivial. It's only meant to teach the existence of scan and how the first result is a parent link. If you think it's a bad idea, well, I don't entirely disagree, but that's also how I feel about requiring a path in the api.

Every other coding puzzle we have in this game, you use the api to solve the puzzle. You never do the backwards thing: solving the puzzle to use the api. The closest we come to that is darknet, where we intentionally don't document the return type of the lab functions, because you're meant to find the shape by observation. But even that is a far cry from this proposal, where the api is not part of the puzzle being solved, but merely gated by it.

I'm having a hard time expressing in words my visceral dislike here. Perhaps it comes, at the root, from the fact that it is *a bad api*. If you look at all our other APIs, they are designed like you would design APIs for any other library: Name the parameters sensibly, make them optional when it makes sense, use best practices for options structs, document types, etc. (We'll ignore the historical warts.) There are game-specific choices about *what* functions we expose, and *when*, and cost is obviously unique to us. But the API design itself is standard. 

The darknet lab APIs dent that slightly. They are still sensible APIs, but the types/documentation are bad, to mimic what you often find in the real world. The upcoming infil APIs are in a similar place: they have intentionally awkward APIs, in a way that is also meant to mimic the real world. 

This is a radical step beyond. The hostname is obviously the only information the API needs to take. In any UI library (or user library), it is the only parameter that would be taken. The extra arguments exist... just to make it harder to call. That is bad api design.

catloversg:
> Perhaps it comes, at the root, from the fact that it is a bad api.

This is the elephant in the room. It may be too late to say it now, but when #1640 was created, I already thought it was not a good idea. I agree that it is useful for new players with very limited programming experience, but it contradicts Bitburner's design philosophy: we provide the tools to build solutions, not the solutions themselves. Because of that contradiction, we are now at an impasse. Everyone draws the line differently on what is acceptable, so our choices are limited.

The first choice is to veto this entire idea. I know it's bad to do that now, especially since d0sboots did not veto it in the past and we have already had two implementation attempts. However, accepting this idea means we will also need to carefully evaluate similar proposals in the future. This kind of API falls into a gray area: it is not problematic enough to be vetoed outright, but it is also not a clear fit for the game's design philosophy.

The second choice is to accept significant compromises. I agree that requiring a path is not ideal. Most other APIs do not require solving part of the problem before they can be used. However, I don't have any better suggestions.

ficocelliguy:
We could build up around the new API to help the player understand why connections are limited on a low level to adjacent or backdoored servers, which is why a full route is required. e.g. make it clear in the documentation / flavor for both this API and existing ones that both terminal and singularity connect are limited this same way, and scan-analyze has to build a tree (and route you through it on click) to make connections available to more places.

We could call the new API something like connectionRouteLink or staticRouteLink, and clarify in its docs it doesn't make a link to some distant server directly, but rather (tries to) route you through the hops you specify. (You need to like `ip route add` each connection you want to use in order to get places, because there isn't network routing tables in this dystopia.) In other words, the value-add of the API is to make a clickable link that works after the creating script dies, not to make a pseudo-backdoor that lets you jump directly - you have to do the legwork on that bit.

We could also add support to singularity.connect to accept a route (list of hostnames) too, to establish consistency / precedence, which would be identical to calling it multiple times in a row (so it's just syntactic sugar, not new capability).

d0sboots:
> > Perhaps it comes, at the root, from the fact that it is a bad api.
> 
> This is the elephant in the room. It may be too late to say it now, but when #1640 was created, I already thought it was not a good idea.

To be absolutely clear, I do not think an API of taking a hostname and returning a react link is a bad api. It might be a bad fit for bitburner (that is debatable, for the reasons you state), but as an *api* it is correctly constructed. 

What I'm saying is a bad api is one that takes a connected path, since that api requires extra information for the sole purpose of making it harder to call.

d0sboots:
> We could call the new API something like connectionRouteLink or staticRouteLink, and clarify in its docs it doesn't make a link to some distant server directly, but rather (tries to) route you through the hops you specify.

You saying this made it all click for me, and I think I see a way through. 

From an API perspective, keep the array of hostnames. However, they are not strictly a path - rather, it's just a list of strings that will be fed to "connect". If you give only a single target, that link will work iff you are adjacent to the target when you click it. A link to "home" will always work. You can bounce back and forth between two servers if you want to. Links to dnet work iff you can connect manually through them.

This becomes "singularity lite" - it lets you do something sing currently does (access to connect), but with *significant* restrictions: 
* Activating it is strictly manual
* No programmatic access to the currently connected host
* The connect path is static, not dynamic at click-time

But it also has the significant upside from singularity that the link works even without a running script. 

From a validation perspective, we would still validate that all the arguments are valid hostname/ips, but no more than that. Whether it will be valid to make a connection between them is too complicated to properly check, and also dynamic anyway. 

IMO this fills a gap that has been persistently present. (People are *always* asking about ways to automate connecting to servers, and currently the best that can be done without exploits is constructing a connect string and putting it in the paste buffer.) It is also equivalent in power to that currently-best method.

Edit: An additional upside is that this would make connecting to dnet servers possible in a balanced way. Right now, the net shifts too quickly to manually connect through it. You *need* to have a script generate a connect path to have any hope of getting to a server properly. A link can be updated in a tail window so that it stays correct (as long as it is possible to do so).

d0sboots:
Another upside to that approach: whatever we do here, IMO we want scan-analyze to share the impl. If scan-analyze used this scheme, it could continue to link dnet servers like it currently does without issues. (The links would just become broken over a short timespan.)

hexagonrecursion:
> From an API perspective, keep the array of hostnames. However, they are not strictly a path - rather, it's just a list of strings that will be fed to "connect".

I guess I am just very bad at communicating. This is pretty much the mental model I had in mind when I said that I would try to refactor `ns.singularity.connect()` and `commands/connect.ts`

catloversg:
It seems we only need to find an appropriate name for this API (e.g., `createRouteLink`). Regarding the other restrictions, I would require it to have a high RAM cost (5 GB or so), as discussed. Gating it behind Autolink.exe is optional, but I prefer keeping it.

d0sboots:
How about `createConnectLink`? IMO "connect" gets at the heart of what it does.

I think 5GB and requiring AutoLink remains appropriate for the revised formulation as well. I wouldn't want to go much over 5, since then it couldn't be used on home in early-game.

catloversg:
That sounds good to me.

src/NetscriptFunctions/UserInterface.tsx
```diff
@@ -18,6 +18,12 @@ import { addGlobalAlias, addAlias, removeAlias, Aliases, GlobalAliases, aliasReg
 import { assertStringWithNSContext } from "../Netscript/TypeAssertion";
 import { Router } from "../ui/GameRoot";
 import { Page } from "../ui/Router";
+import React from "react";
```
d0sboots: Style: This should come first

src/NetscriptFunctions/UserInterface.tsx
```diff
@@ -278,5 +284,31 @@ export function NetscriptUserInterface(): InternalAPI<IUserInterface> {
...
+        throw errorMessage(
+          ctx,
+          `connectPath must be an array. Current type is ${getFriendlyType(_connectPath)}.`,
+          "TYPE",
+        );
+      }
+      const connectPath = _connectPath.map((s) => helpers.string(ctx, "connectPath", s));
+      for (const host of connectPath) {
+        const [s] = helpers.getServer(ctx, host);
+        if (s == null || (s.hostname === SpecialServers.DarkWeb && s.serversOnNetwork.length === 0)) {
```
d0sboots: I think this whole check should be removed. 

Offline servers shouldn't throw. The check for serversOnNetwork is not one that should be made here; if it's important that you can't access darkweb pre-tor then that should be fixed in getServer.

src/NetscriptFunctions/UserInterface.tsx
```diff
@@ -278,5 +284,31 @@ export function NetscriptUserInterface(): InternalAPI<IUserInterface> {
...
+          throw errorMessage(ctx, `Invalid host: '${host}'`);
+        }
+      }
+      const last = connectPath.at(-1);
+      if (last == null) {
+        throw errorMessage(ctx, `connectPath must not be empty.`, "TYPE");
+      }
+      const linkText = _linkText == null ? last : helpers.string(ctx, "linkText", _linkText);
```
d0sboots:
```suggestion
      const linkText = helpers.string(ctx, "linkText", _linkText ?? last);
```

src/NetscriptFunctions/UserInterface.tsx
```diff
@@ -278,5 +284,31 @@ export function NetscriptUserInterface(): InternalAPI<IUserInterface> {
...
+      }
+      const connectPath = _connectPath.map((s) => helpers.string(ctx, "connectPath", s));
+      for (const host of connectPath) {
+        const [s] = helpers.getServer(ctx, host);
+        if (s == null || (s.hostname === SpecialServers.DarkWeb && s.serversOnNetwork.length === 0)) {
+          throw errorMessage(ctx, `Invalid host: '${host}'`);
+        }
+      }
+      const last = connectPath.at(-1);
+      if (last == null) {
```
d0sboots: Optional: Is this really invalid? An empty array should just do nothing, right? Off the top of my head, I can't think of anything it would be *useful* for, but it feels iffy to ban it.

src/NetscriptFunctions/UserInterface.tsx
```diff
@@ -278,5 +284,31 @@ export function NetscriptUserInterface(): InternalAPI<IUserInterface> {
...
+        throw errorMessage(ctx, `Requires AutoLink.exe to run.`);
+      }
+      if (!Array.isArray(_connectPath)) {
+        throw errorMessage(
+          ctx,
+          `connectPath must be an array. Current type is ${getFriendlyType(_connectPath)}.`,
+          "TYPE",
+        );
+      }
+      const connectPath = _connectPath.map((s) => helpers.string(ctx, "connectPath", s));
```
d0sboots:
```suggestion
      // Enforce validation of server and return resolved hostname
      const connectPath = _connectPath.map((s) => helpers.getServer(helpers.string(ctx, "connectPath", s)))[1];
```
With my suggestion below, everything can be collapsed into the map

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -7128,6 +7123,31 @@ interface UserInterface {
    * @param node - The node to be rendered.
    */
   renderPage(node: ReactNode): void;
+
+  /**
+   * Allows programmatic use of AutoLink.exe.
```
d0sboots, line 7128: Good short description.

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -7128,6 +7123,31 @@ interface UserInterface {
...
+   *
+   * @remarks
+   * RAM cost: 5 GB
+   *
+   * This function uses AutoLink.exe to create a clickable link. Clicking on the link is approximately
```
```suggestion
   * This function uses AutoLink.exe to create a clickable link. Clicking on the link is
```
If it's not equivalent in all the ways that matter, we did something wrong.

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -7128,6 +7123,31 @@ interface UserInterface {
...
+   * equivalent to typing a sequence of "connect" commands into the terminal.
+   *
+   * * @example
+   * ```js
+   * export async function main(ns) {
+   *   // Prints a link to the terminal. Clicking on it is approximately equivalent
```
```suggestion
   *   // Prints a link to the terminal. Clicking on it is equivalent
```

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -7128,6 +7123,31 @@ interface UserInterface {
...
+   *   // to typing `connect joesguns; connect zer0; connect  silver-helix`.
+   *   ns.tprintRaw(ns.ui.createConnectLink(['joesguns', 'zer0', 'silver-helix']));
+   * }
+   * ```
+   *
+   * @param connectPath - host names or ip addresses of servers to connect to.
+   * @param linkText - the text to display on the link. Defaunts to the last host name or ip in connectPath.
```
```suggestion
   * @param linkText - the text to display on the link. Defaults to the last host name or ip in connectPath.
```

src/Server/ServerHelpers.ts
```diff
@@ -256,6 +256,26 @@ export function prestigeHomeComputer(homeComp: Server): void {
   }
 }
 
+type ConnectionResult =
```
d0sboots: This should be an actual Result (import from nsdefs)

src/NetscriptFunctions/Singularity.ts
```diff
@@ -475,39 +474,15 @@ export function NetscriptSingularity(): InternalAPI<ISingularity> {
...
-      if (target.backdoorInstalled || target.purchasedByPlayer) {
-        Terminal.connectToServer(host, true);
+      const result = validateConnections(Player.getCurrentServer(), [target.hostname]);
+      if (result.status === "ok") {
+        Terminal.connectToServer(result.destination.hostname, true);
         return true;
       }
-
-      // Failure case
       return false;
```
d0sboots: Should log the failing msg from the result. (This is new, but that's the whole point of having a Result, and it's non-breaking.)

src/Terminal/commands/connect.ts
```diff
@@ -13,39 +11,22 @@ export function connect(args: (string | number | boolean)[], server: BaseServer)
...
-        new Error(
-          `${server.serversOnNetwork[i]} is on the network of ${server.hostname}, but we cannot find its data.`,
-        ),
+  switch (result.status) {
+    case "server not found":
+      Terminal.error(`Invalid hostname: '${hostname}'`);
+      return;
+    case "no connection":
+      Terminal.error(
+        `Cannot directly connect to ${hostname}. Make sure the server is backdoored or adjacent to your current server`,
```
d0sboots: These cases will move directly into validateConnections. There's no conflict since sing didn't do *any* with reporting before.

src/Terminal/ui/ConnextLink.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+interface IConnectLinkProps {
+  path: string[];
+  text: string;
+}
+
+export function ConnectLink(props: IConnectLinkProps): React.ReactElement {
+  const onClick = (event: React.MouseEvent<HTMLAnchorElement>) => {
+    const { nativeEvent } = event;
+    if (!(nativeEvent instanceof Event) || !nativeEvent.isTrusted) {
```
d0sboots: I'd remove this. It can be bypassed with some extra effort, and IMO we should reward players who want to get cheeky. There are a few puzzles that we've hardened this way for extra challenge, but I don't think this needs to be one of them.

src/Terminal/ui/ConnextLink.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { Player } from "@player";
+import { validateConnections } from "../../Server/ServerHelpers";
+
+interface IConnectLinkProps {
+  path: string[];
+  text: string;
+}
+
+export function ConnectLink(props: IConnectLinkProps): React.ReactElement {
+  const onClick = (event: React.MouseEvent<HTMLAnchorElement>) => {
```
d0sboots: Optional: A new handler gets created on every eval of the function. You can use useCallback to eliminate that. It becomes easier to do if you destructure props into {path, text} in the signature.

test/jest/Netscript/UserInterface.test.ts
```diff
@@ -178,3 +184,94 @@ test("alias", () => {
   expect(() => ns.ui.alias("   ", "bar")).toThrow();
   expect(() => ns.ui.alias("^", "bar")).toThrow();
 });
+
+test("createConnectLink", () => {
+  const ns = getNS();
+
+  expect(() => ns.ui.createConnectLink([SpecialServers.Home])).toThrow();
```
d0sboots:
```suggestion
  expect(() => ns.ui.createConnectLink([SpecialServers.Home])).toThrow("Requires AutoLink");
```

test/jest/Netscript/UserInterface.test.ts
```diff
@@ -178,3 +184,94 @@ test("alias", () => {
...
+  Player.getHomeComputer().pushProgram(CompletedProgramName.autoLink);
+  ns.ui.createConnectLink([SpecialServers.Home]);
+
+  expect(() => ns.ui.createConnectLink([])).toThrow();
+
+  expect(() => ns.ui.createConnectLink([SpecialServers.WorldDaemon])).toThrow();
```
d0sboots: Here and everywhere: expect an identifying part of the error message.

test/jest/Netscript/UserInterface.test.ts
```diff
@@ -178,3 +184,94 @@ test("alias", () => {
...
+    //   ┃ ┗ b
+    //   ┗ c
+    //     ┗ d
+    connectServers(home, a);
+    connectServers(home, c);
+    connectServers(a, b);
+    connectServers(c, d);
+    return { home, a, b, c, d };
+  };
+  it("connects by host name", () => {
```
d0sboots: This ought to be written as a data-parametrized test.

d0sboots (commented):
Despite the large number of comments, this looks good! The structure is right, it's just down to details

hexagonrecursion:
I will finish this tomorrow.

src/Netscript/NetscriptHelpers.tsx
```diff
@@ -551,12 +552,18 @@ function scriptIdentifier(ctx: NetscriptContext, scriptID: unknown, _host: unkno
 export function getServer(ctx: NetscriptContext, _host: unknown): [BaseServer | null, string] {
   const host = helpers.string(ctx, "host", _host ?? ctx.workerScript.hostname);
   const server = GetServer(host);
-  if (server != null && (server.serversOnNetwork.length > 0 || server instanceof DarknetServer)) {
-    return [server, host];
-  }
-  if (DarknetState.offlineServers.has(host)) {
-    log(ctx, () => `Server ${host} is offline.`);
-    return [null, host];
+  if (server == null) {
```
d0sboots: Why did you re-order this if statement? IMO it's harder to follow now, without the early return, and no more correct.

catloversg: It seems they are trying to fix an issue in this function. According to the TSDoc, it throws when `_host` is darkweb and the player does not have TOR (`pre-TOR darkweb`), but the implementation does not check that case. I agree that the change is hard to follow. I would keep the current structure while making only a small change like this:
```ts
  if (
    server != null &&
    (server.serversOnNetwork.length > 0 ||
      (server instanceof DarknetServer && server.hostname !== SpecialServers.DarkWeb))
  ) {
    return [server, host];
  }
```
`server instanceof DarknetServer` => `(server instanceof DarknetServer && server.hostname !== SpecialServers.DarkWeb)`

hexagonrecursion:
* I have changed the rules that the `connect` command and `ns.singularity.connect()` follow: it is now possible to connect to a server that you are already connected to. e.g. `connect n00dles` will succeed when you are already connected to n00dles. Previously `connect n00dles` would fail when you are already connected to n00dles (unless n00dles has a backdoor). I made this change in preparation for https://github.com/bitburner-official/bitburner-src/pull/2995

Here is my thought process:
1. I want the links that scan-analyze creates while you are connected to home to always work no matter where you are (assuming the path from home is still valid). Therefore the links have to contain a sequence of commands that starts with `connect home`
2. For consistency I made scan-analyze always produce links that contain a sequence of commands that starts with `connect <server you executed scan-analyze from>`, but this in turn requires allowing connections to the server that you are already connected to

src/Netscript/NetscriptHelpers.tsx
hexagonrecursion: Thanks. I think I found a nicer-looking way to express the same check.

hexagonrecursion:
I think I have addressed all review comments. I am not 100% sure though because github keeps occasionally hiding some of the comments from me

src/Netscript/NetscriptHelpers.tsx
d0sboots: I still think catlover's suggestion is best. Creating a new method on Server would be best if we wanted to re-use that logic from multiple places, but we *don't* want that - we want everyone to be calling GetServer instead. So it makes the most sense to have the logic be encapsulated there, and not require jumping through other files to understand.

d0sboots:
I agree with the dicision to allow self-connect and the change to scan-analyze. All comments are addressed, except for the last comment thread.

test/jest/Netscript/UserInterface.test.ts
```diff
@@ -227,13 +228,13 @@ test.each([
     throw new Error("home is not a Server");
   }
   prestigeAllServers();
-  home.ip = "1.1.1.1";
+  home.ip = "1.1.1.1" as IPAddress;
```
hexagonrecursion, line 241: Strange. I _did_ run the linter. Why did I not get the error? Why did continuous integration not detect this?

catloversg: These are errors reported by tsc. ESLint won't report them. I configured the lint task to also perform a tsc check, but that is still in my WIP changes. I usually detect this kind of error, but that's because I manually test PRs with a custom setup.

catloversg: @d0sboots Do you want me to upstream the mentioned change (run tsc in lint task)? It's currently included in my WIP PR of upgrading eslint, but that PR will take a while to finish.

hexagonrecursion:
In #2995 d0sboots suggested adding tooltips. I have created a new branch for it: https://github.com/hexagonrecursion/bitburner-src/tree/link-tooltip . I would have just merged the commit into hexagonrecursion:link instead of creating a new branch, but I wasn't sure if it would be ok because it looks like you are getting ready to merge #2948.

catloversg:
That tooltip can be implemented in #2995. This PR is nearly complete. You can merge this branch to 2995's branch and start working there.

test/jest/Netscript/UserInterface.test.ts
d0sboots: Yeah, let's do that as a separate thing.

PR #2953, opened 2026-07-04
DOCS: Clarify ns.corporation.limitMaterialProduction()

hexagonrecursion:
In some industries (e.g. Agriculture) output materials take up more space than input materials. I thought that recent changes to how material production limits work unintentionally create a new optimal strategy: discarding some of the plants you produce to be able to produce more food than your warehouse space would otherwise allow (or discarding food to have more space for plants, depending on which of the two is more useful to you). It turns out, I was wrong: materials that are discarded by a production limit still require empty warehouse space.

I have updated the docs to clarify this.

### DIFF
--- a/src/ScriptEditor/NetscriptDefinitions.d.ts
+++ b/src/ScriptEditor/NetscriptDefinitions.d.ts
@@ -10302,11 +10302,19 @@ export interface WarehouseAPI {
    * @remarks
    * RAM cost: 20 GB
    *
-   * This limit applies only to output; it does not affect input consumption.
+   * This limit does not reduce input consumption.
    *
-   * For example, in Agriculture, assume the division's raw production is 1000. You need to consume 500 Water and 200
-   * Chemicals to produce 1000 Plants and 1000 Food. If you set the limits for Plants and Food to 200 and 100
-   * respectively, you will still consume 500 Water and 200 Chemicals, but only produce 200 Plants and 100 Food.
+   * For example, in Agriculture, assume the division's raw production is 1000 and assume the division
+   * is bottlenecked by raw production (as opposed to lack of input materials or warehouse space).
+   * You need to consume 500 Water and 200 Chemicals to produce 1000 Plants and 1000 Food. If you set
+   * the limits for Plants and Food to 200 and 100 respectively, you will still consume 500 Water
+   * and 200 Chemicals, but only produce 200 Plants and 100 Food.
+   *
+   * The limit does not make production more space-efficient.
+   *
+   * For example, If you set the limit for Plants to 0, the formula effectively becomes
+   * 0.5 Water + 0.2 Chemicals ⟹ 1 Food, but it still requires empty warehouse space as
+   * if it were 0.5 Water + 0.2 Chemicals ⟹ 1 Plants + 1 Food.
    *
    * With industries that produce both materials and products, the material production limits do not affect product
    * production.
--- a/src/ScriptEditor/NetscriptDefinitions.d.ts
+++ b/src/ScriptEditor/NetscriptDefinitions.d.ts
@@ -10302,19 +10302,27 @@ export interface WarehouseAPI {
    * @remarks
    * RAM cost: 20 GB
    *
-   * This limit does not reduce input consumption.
+   * This limit applies only to output. It does not reduce input consumption. The excess output is discarded after being
+   * produced.
    *
-   * For example, in Agriculture, assume the division's raw production is 1000 and assume the division
-   * is bottlenecked by raw production (as opposed to lack of input materials or warehouse space).
-   * You need to consume 500 Water and 200 Chemicals to produce 1000 Plants and 1000 Food. If you set
-   * the limits for Plants and Food to 200 and 100 respectively, you will still consume 500 Water
-   * and 200 Chemicals, but only produce 200 Plants and 100 Food.
+   * For example: Assume the division's raw production is 1000 and the warehouse has enough input materials. In
+   * Agriculture, for each unit of raw production, you need 0.045 units of free space (check corporation documentation
+   * for further explanation). Therefore, with RawProduction = 1000, you need 45 units of free space to avoid being
+   * bottlenecked by insufficient free space.
    *
-   * The limit does not make production more space-efficient.
+   * Case 1: Enough free space (Free space = 45). You need to consume 500 Water and 200 Chemicals to produce 1000
+   * Plants and 1000 Food.
    *
-   * For example, If you set the limit for Plants to 0, the formula effectively becomes
-   * 0.5 Water + 0.2 Chemicals ⟹ 1 Food, but it still requires empty warehouse space as
-   * if it were 0.5 Water + 0.2 Chemicals ⟹ 1 Plants + 1 Food.
+   * If you set the limits for Plants and Food to 200 and 100 respectively, you will still consume 500 Water and 200
+   * Chemicals, but only produce 200 Plants and 100 Food (the excess 800 Plants and 900 Food are discarded after being
+   * produced). The free space after production is 67 (800 Plants and 900 Food are discarded).
+   *
+   * Case 2: Insufficient free space (Free space = 22.5). The available free space is only 50% of the required free
+   * space. Therefore, RawProduction is scaled down to 500.
+   *
+   * You need to consume 250 Water and 100 Chemicals to produce 500 Plants and 500 Food. If you set the Food limit to 0,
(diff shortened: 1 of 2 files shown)


src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -10302,11 +10302,19 @@ export interface WarehouseAPI {
    * @remarks
    * RAM cost: 20 GB
    *
-   * This limit applies only to output; it does not affect input consumption.
+   * This limit does not reduce input consumption.
```
```suggestion
   * This limit applies only to output. It does not reduce input consumption.
```

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -10302,11 +10302,19 @@ export interface WarehouseAPI {
...
    *
-   * For example, in Agriculture, assume the division's raw production is 1000. You need to consume 500 Water and 200
-   * Chemicals to produce 1000 Plants and 1000 Food. If you set the limits for Plants and Food to 200 and 100
-   * respectively, you will still consume 500 Water and 200 Chemicals, but only produce 200 Plants and 100 Food.
+   * For example, in Agriculture, assume the division's raw production is 1000 and assume the division
```
```suggestion
   * For example, in Agriculture, assume the division's raw production is 1000 and the warehouse has enough input
   * materials and free space. You need to consume 500 Water and 200 Chemicals to produce 1000 Plants and 1000 Food. If
   * you set the limits for Plants and Food to 200 and 100 respectively, you will still consume 500 Water and 200
   * Chemicals, but only produce 200 Plants and 100 Food.
```
- Don't repeat "assume".
- I think mentioning bottlenecks like this is actually more confusing. If you consider "raw production" as a bottleneck, then the production is *always* bottlenecked by raw production.

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -10302,11 +10302,19 @@ export interface WarehouseAPI {
...
+   * is bottlenecked by raw production (as opposed to lack of input materials or warehouse space).
+   * You need to consume 500 Water and 200 Chemicals to produce 1000 Plants and 1000 Food. If you set
+   * the limits for Plants and Food to 200 and 100 respectively, you will still consume 500 Water
+   * and 200 Chemicals, but only produce 200 Plants and 100 Food.
+   *
+   * The limit does not make production more space-efficient.
```
catloversg: This part is not necessary. Players should be able to infer it after reading "This limit applies only to output. It does not reduce input consumption.", and the new example is almost the same as the previous one (it only changes the limit of Plants to 0).

We don't need to spell out everything explicitly in TSDoc.

catloversg: If we really need to clarify this, I suggest doing this:
```js
   * This limit applies only to output. It does not reduce input consumption. The excess output is discarded after being
   * produced.
   *
   * For example, in Agriculture, assume the division's raw production is 1000 and the warehouse has enough input
   * materials and free space. You need to consume 500 Water and 200 Chemicals to produce 1000 Plants and 1000 Food. If
   * you set the limits for Plants and Food to 200 and 100 respectively, you will still consume 500 Water and 200
   * Chemicals, but only produce 200 Plants and 100 Food (the excess 800 Plants and 900 Food are discarded after being
   * produced).
```

hexagonrecursion: > and the new example is almost the same as the previous one (it only changes the limit of Plants to 0).

Perhaps I need to come up with a better example. The intention was for the two examples to cover two different scenarios:
1. There is enough warehouse space, RawProduction is the limiting factor
2. There is enough RawProduction, warehouse space is the limiting factor

Both scenarios assume that you buy exactly enough input materials

catloversg: I think explicitly mentioning the discarded excess output is already more than enough.

d0sboots: I don't understand what you were confused by. Without that, it's hard to understand the goal of this PR.

d0sboots: OK, reading the comments as well as the changes together more carefully, I think I understand the confusion now. It centers around this line that catlover suggested adding: "The excess output is discarded after being produced."

There are two ways you can mentally model a production "limit," and without that line both are valid possibilities, but have different effects. The way it actually works is that we produce as normal, and then discard any excess that was after the limit. Another sensible approach would be to only produce up to the limit, *even though all input materials are still consumed*.

The difference between the two comes when space is constrained. Suppose you have 195 space left, and set a production limit of 0 plants. IIUC you will produce 150 food and 150 plants, consuming 75 water and 19.5 chemicals, and then toss the plants, leaving 150 space free. By the other interpretation, you would produce 650 food and 0 plants, consuming 325 water and 130 chemicals, leaving 0 space free.

d0sboots: Given the amount of code diving I had to do to come up with this example, I think it (or something like it) would be worth adding. It not only clears the confusion, it also clarifies points about how the production amount is calculated when space is tight and there are limits.

catloversg: For each unit of raw production, you need 0.045 unit of free space (0.05 + 0.03 - (0.5 * 0.05 + 0.2 * 0.05). With RawProduction = 1000, you need 45 units of free space to avoid being bottlenecked by insufficient free space.

An example of the case in which there is not enough free space:
Given:
- RawProduction = 1000
- Enough input materials
- Free space = 22.5
- Food limit = 0

Calculation:
- 1 RawProduction requires 0.045 units free space. Therefore, 1000 RawProduction requires 45 units free space.
- The available free space is only 50% of the required free space. Therefore, RawProduction is scaled down to 500.
- Given 0.5 Water + 0.2 Chemicals => 1 Plants + 1 Food, we have:
  - Consume: 250 Water + 100 Chemicals
  - Produce: 500 Plants + 500 Food
- Given Food limit = 0, we have:
  - Produce: 500 Plants + 0 Food
- Free space: 15 (500 Food are discarded)

I'll add a Jest test for this example and explain in details how raw production is scaled down when there is not enough free space in corp documentation. For this PR, I recommend using this explanation:
```ts
   * This limit applies only to output. It does not reduce input consumption. The excess output is discarded after being
   * produced.
   *
   * For example: Assume the division's raw production is 1000 and the warehouse has enough input materials. In
   * Agriculture, for each unit of raw production, you need 0.045 units of free space (check corporation documentation
   * for further explanation). Therefore, with RawProduction = 1000, you need 45 units of free space to avoid being
   * bottlenecked by insufficient free space.
   *
   * Case 1: Enough free space (Free space = 45). You need to consume 500 Water and 200 Chemicals to produce 1000
   * Plants and 1000 Food.
   *
   * If you set the limits for Plants and Food to 200 and 100 respectively, you will still consume 500 Water and 200
   * Chemicals, but only produce 200 Plants and 100 Food (the excess 800 Plants and 900 Food are discarded after being
   * produced). The free space after production is 67 (800 Plants and 900 Food are discarded).
   *
   * Case 2: Insufficient free space (Free space = 22.5). The available free space is only 50% of the required free
   * space. Therefore, RawProduction is scaled down to 500.
   *
   * You need to consume 250 Water and 100 Chemicals to produce 500 Plants and 500 Food. If you set the Food limit to 0,
   * you will still consume 250 Water and 100 Chemicals, but only produce 500 Plants (the excess 500 Food are discarded
   * after being produced). The free space after production is 15 (500 Food are discarded).
```

hexagonrecursion: > The difference between the two comes when space is constrained. Suppose you have 195 space left, and set a production limit of 0 plants. IIUC you will produce 150 food and 150 plants, consuming 75 water and 19.5 chemicals, and then toss the plants, leaving 150 space free. By the other interpretation, you would produce 650 food and 0 plants, consuming 325 water and 130 chemicals, leaving 0 space free.

@d0sboots, I agree with your post, except your math is wrong.
Firstly, you assumed that every material has a size of 1. Their actual sizes:
|material|size|
|-|-|
|Water | 0.05 |
|Chemicals|0.05|
|Food|0.03|
|Plants|0.05|

Secondly, 150*0.2 is 30. I am not sure how you got 19.5

hexagonrecursion:
@catloversg Thank you. I have applied your suggestion

src/ScriptEditor/NetscriptDefinitions.d.ts
d0sboots: Yeah, I'm not surprised my math was wrong XD

But the example made sense, which was the more important bit.

PR #2954, opened 2026-07-04
TUTORIAL: Discoverable example scripts on npc servers

ficocelliguy:
In order to reward discovery and add more ways for players to learn about the game and its tools, this PR adds somewhat-broken example scripts that spawn on npc servers. The idea is to practice debugging, reward exploration, hint as to what kinds of tools the game gives you, and flesh out the world and backstory/lore more.

The specific scripts included all fail to run initially, requiring players to fix various flaws. Some are more subtle than others, and some have more flaws than others. The more potentially powerful scripts have structural/logical flaws in addition to the syntactic ones, to ensure players do not get ready-made solutions for anything useful from these scripts.

Details on the specific scripts:

1:  A simple game using ns.prompt. Demonstrates simple control flow, user interaction, ns.ui.openTail. Establishes jump3r's real name and some history.  Bugged (doesn't properly stop when the right answer is found)
_guessing_game.wip.js_

2: Basic script that demonstrates that cloud servers can be upgraded. Shows how typescript can be helpful. Only upgrades one server. Bugged: uses the wrong arguments to upgrade, and wrong ram value, and forgotten parens on a function call. 
_upgrade_cloud_server.wip.ts_

3: Port sender and receiver scripts. Demonstrates detecting other scripts, starting child scripts, waiting for child scripts to start, writing data to ports, receiving the data in another script, coloring text in logs. Uses hard-coded single port. Hints about daedalus in the lore. Bugged: doesn't pass port number to nextPortWrite and writePort
_port_receiver.wip.js_
_port_sender.wip.js_

4: A broken script that thinks about calling scan repeatedly to find servers (but doesn't do it.) Bugged: does not actually iterate over array, does not pass ns into function, wrong capitalization for a scan call, pushes full result array instead of concatenating, doesn't filter out unique results from scan
_server_finder.wip.js_

5: Extremely minimal 10-line hack-controller script. Starts looping child scripts against a hard-coded target (n00dles). Demonstrates parallel execution of child scripts, script args, using the return value of exec to detect launch success. This "loop algorithm" style is simpler (and less effective) than protobatching, and isn't mentioned in the Hacking Algorithms document anymore, so I wanted to give it a place for players to find it.
Bugged: incorrect use of exec, doesn't pass target into weaken script, some of the child scripts are missing. Doesn't use other servers, just sorta fills the current one. Hard-codes 1 thread per child script and equal numbers of hack loops, grow loops, and weaken loops.
_hack_controller.wip.js_

6: Basic UI interaction: adding text to the overview, printRaw basic html content in script logs, ads a clickabe button using player theme coloring. Lore includes flavor (but no details) about endgame. Bugged: forgotten await on ns.asleep, missing close curly brace on click handler, wrong capitalization on ns method call
_escaping_your_chains.wip.jsx_

7: Custom UI creation: adding entire new basic page with the new UI, custom buttons with click handler, css styling, react state variable. jump3r continues to deteriorate in the lore. Bugged: extra semicolon breaking function call, forgotten await
_custom_content.wip.jsx_


src/Literature/DiscoverableScriptContents/hack_controller.wip.js
```diff
@@ -0,0 +1,47 @@
+/*
+                        ** Shotgun-style Batcher **
+                (c) 2074 Fulcrum Tech (Secret Lab Division)
+                             Author: jump3
```
E9cipher:
```suggestion
                             Author: jump3r
```

src/Literature/DiscoverableScriptContent/escaping_your_chains.wip.jsx
```diff
@@ -0,0 +1,46 @@
...
+  A true hacker is one who works to learn the secrets of every tool and tech they use.
+  It's not about cracking security or pwning servers - it's a craftsman's mindset, the application of ingenuity.
+
+  Whether you are doing and-dirty patchwork or elegant architecture, know your tools, understand the problem, and
+  apply your craft to solve it. Don't just use someone else's script as-is or ask a rogue AI to write it for you!
+  Piece it together yourself from whatever you can find, and make it your own. Break it, fix it, and understand it.
+
+   This world is not what it seems. You can break its boundaries and push what is possible. Never be complacent!
+
+   If you find this, it means I'm already gone. Gone ahead into what comes next. I hope I've left enough hints behind.
```
E9cipher: > it means I'm already gone

What does this suggest? On discord you said she would leave the BN before the player, but she isn't dead, is she?

src/Literature/DiscoverableScriptContents/port_sender.wip.js
```diff
@@ -0,0 +1,44 @@
+/*
+            **  Data Transfer Protocol RFC 1149  **
+                (c) 2069 Helios Labs / jump3r
+
+           See port_receiver.wip.js for full details.
+
+
+     > is this finally good enough? it's perfect code - no flaws!
+     > I know that a high hacking skill is the key to finally
+     > getting an in with daedalus
```
E9cipher: Isn't that a spoiler?

catloversg:
First, for context, I'll repost my general comments that I posted on Discord.

General comments:
- Do not provide code that is actually optimal. For example, in order to access document, just use document. Do not use globalThis["document"].
- All scripts should always be unusable as-is. Broken scripts are actually "good".
- The line between giving hints and hand-holding is very thin.

Writing corp documentation taught me a valuable lesson: you cannot satisfy everybody when it comes to usefulness, even when it's "only" documentation. People complain when there are no docs. Then other people complain when there are docs but they contain too many hints or too much advice.

These scripts are basically hints for new players about topics that are asked about most frequently. In other words, they are a new FAQ.md in the form of JS code. They are very useful, but if they become too useful, they start turning into almost ready to use solutions. We should draw a fine line between giving hints and giving out working code. This is why I said, "All scripts should always be unusable as is." Players are expected to inspect these scripts and get hints from them, not run them.

I'll review this PR carefully later. Currently, I have only skimmed through your description and the example scripts. Among them, I notice two scripts:
- `server_finder.wip.js`: As I said on Discord, this script is very close to a ready-to-use solution for an issue that everybody must solve in Bitburner. Most other scripts introduce some optional features of Bitburner, so not everybody has to write them. However, I have never seen any players who do not write a server scanner. This means you are giving players something that everybody writes on their own. Even after adding more flaws to it, I still think this is not something we should provide.
- `batch_hack.wip.js`: I did not notice this script back when I checked the thread on Discord. I'll be clear with you, I'm very strongly against this script. Not just the implementation, but the entire idea. Batcher is one of the most powerful tools in this game, so it's usually the best bang-for-buck script. Players can spend their time to write a barely usable batcher and then enjoy the massive benefit everywhere else. With `additionalMsec` and the much better `hackingalgorithms.md`, batcher is nowhere near the dark magic it was in the past. I'm fine with improving `hackingalgorithms.md` even further to explain shotgun batcher in more detail, but I'm against any attempts to provide code of a shotgun batcher, no matter how many artificial flaws you put in it.

I can find a compromise if too many people like that scanner, but I won't budge on the shotgun batcher (note that I know this script is nowhere near an actual runnable shotgun batcher).

E9cipher (commented):
Just a typo and a few things. Did I do it properly? (first time reviewing :)

src/Literature/DiscoverableScriptContent/escaping_your_chains.wip.jsx
ficocelliguy: She knows about the bitverse somehow, implied from the existing lit files/ lore. She can jump, after all!

src/Literature/DiscoverableScriptContents/port_sender.wip.js
ficocelliguy: added redaction to hide spoiler (but still hint at helios <-> daedalus connection)

src/Literature/DiscoverableScriptContent/build_new_worlds.wip.jsx
```diff
@@ -0,0 +1,82 @@
...
+ */
+
+/** @param {NS} ns */
+export async function main(ns) {
+  ns.ui.openTail();
+  addToSidebar;(ns, "֍   jump3r's secrets", 'this world is not what it seems');
+  ns.tprint('Custom content added to sidebar');
+
+  // Prevent the script from exiting, so react components can still use ns methods
+  ns.asleep(1e20);
```
catloversg: Optional: Replace the missing await bug here with another bug somewhere else.

Although forgetting to use `await` is a common mistake, in normal cases, it's very obvious because we throw a concurrent error that explicitly hints to players that they are missing `await`. However, in this case, players only see an error message in the tail log saying `ui.renderPage: Failed to run due to script being killed.`. I think new players will have trouble figuring out what went wrong.

src/Literature/DiscoverableScriptContent/escaping_your_chains.wip.jsx
```diff
@@ -0,0 +1,46 @@
...
+    ns.toast('Was that really a good idea?');
+
+
+  ns.printRaw(
+    <button style={{ borderColor: borderColor, backgroundColor: '#111', color: '#FFF' }} onClick={handleClick}>
+      Click Me!
+    </button>,
+  );
+
+  ns.asleep(1e20);
```
catloversg: Same comment about the missing await bug.

You can also introduce the "never-resolving promise" trick here (`return new Promise(() => {});`). Passing a large number to sleep/asleep works in most cases, but most people don't know the quirks of `setTimeout`, so they will be surprised when they pass numbers that hit edge cases of `setTimeout`.

src/Literature/DiscoverableScriptContent/port_sender.wip.js
```diff
@@ -0,0 +1,44 @@
+/*
+            **  Data Transfer Protocol RFC 1149  **
+                (c) 2069 Helios Labs / jump3r
+
+           See port_receiver.wip.js for full details.
+
+
+     > is this finally good enough? it's perfect code - no flaws!
+     > I know that a high hacking skill is the key to finally
+     > getting an in with d██████s
```
catloversg: What is this word supposed to be? daemons?

src/Literature/DiscoverableScriptContent/hack_controller.wip.js
```diff
@@ -0,0 +1,44 @@
...
+
+ RULE 1: Timing is Everything
+ Hack calls blow up security upon completion!
+ grows and weakens must be started BEFORE the hack finishes, to avoid security debuffs.
+
+ RULE 2: The Rich Get Richer
+ Grow is like interest - the more money the server has, the more that grow adds!
+ Make sure the server stays near max money. Don't steal big chunks of money at once.
+
+ RULE 3: The Botnet Must Grow
```
catloversg: Remove this "rule".

The other three "rules" are advice related to how HGW operations work. These things are not exactly easy to discover, so although I don't like giving them out here, I won't veto them. However, this rule is simply too good to be here.

This advice is repeated too many times on Discord, so people tend to forget or underestimate how valuable it is for new players. It may sound simple to you, but most new players don't know how important RAM is. This is something they should discover on their own.

catloversg (commented):
Regarding the change to jumper's lore, I know you want to improve the lore, but I think this is the wrong way to do that.

In terms of the development timeline, Jumper is the most ancient "being" in Bitburner. We added Jumper's messages even before we implemented TRP. You can think of Jumper as a mysterious character whose exact nature nobody knows. The lore about Jumper and the Enders is still vague, and that's actually a good thing.

If nobody knows who or what Jumper is, they can be anything. They can be an Ender. They can be a rogue Ender. They can be the AI that controls BitVerse under the Enders' command. They can be the AI that controls BitVerse, but that AI became a rogue one. They can be a mere trapped human mind that fails to escape the sandbox. They can be the first human mind to escape the sandbox and want to help other people's minds. Because the lore is vague, people can freely interpret it in any way they want. There is no limitation. If you create canonical lore like this, you basically kill all other interpretations.

If you want to "flesh out" the lore and add more lore, you should create a new character. That's your character, so you are free to "design" them without interfering with other people's interpretations of the most mysterious character.

Note that when you create this character, although you are free to set up their story, you should still be reasonably vague. For example, don't kill them off or make them destroy WD. If their lore is too specific, future contributors may run into trouble when their lore contradicts yours. You should also avoid making them become Jumper v2 (no flight or other jumper's stuff). In general, being sufficiently vague is a good thing in Bitburner lore. Bitburner is not a game written by one person and doesn't need detailed lore from day one. We should provide enough room for future contributors when they add new lore.

Another note: you are hinting at WD in a lot of places. You should be very careful about that. I saw you removed the mention of Daedalus, but Daedalus (the faction name) is a tiny spoiler compared to WD.

ficocelliguy:
Part of the inspiration for this project is how frustratingly little there is to find out about jump3r, outside of the fact that they can get into your computer and they hint at the WD. Adding some hints to a past doesn't preclude them from being mysterious, or later in the game's timeline (when the player appears) awakening as any of the ideas you proposed. I repeatedly have seen questions from players asking "who is jumper?" and there just isn't much to say, and what there _is_ to talk about is a spoiler. Adding a little depth gives structure - a launching point for players' imagination, or future contributors' work.

src/Literature/DiscoverableScriptContent/port_sender.wip.js
ficocelliguy: daedalus - their server is owned by Helios and I wanted to add an extra tie here. I changed the redaction slightly to make it more readable after you know the word

src/Literature/DiscoverableScriptContent/hack_controller.wip.js
ficocelliguy: I swapped it for another on-theme rule

catloversg:
> I repeatedly have seen questions from players asking "who is jumper?" and there just isn't much to say, and what there is to talk about is a spoiler.

If someone asks, "Who is jumper?", the only "correct" answer is, "Nobody knows." The better but longer answer is, "Nobody knows. After you enter the endgame, there are more hints about who or what jumper might be. You are free to imagine their identity."

In the current state of this PR, you are implying a lot of things about jumper. You give them a name (Junie "jumper" Perez). You give them a "human" personality (look at the tone of the messages in `guessing_game.wip.js` and other scripts). You imply they either die or escape the sandbox ("If you find this, it means I'm already gone. Gone ahead into what comes next."). All these things add more details about who jumper is, which is likely what you want. However, as I said, I think that's the wrong way to improve the lore.

You don't need to add more details about jumper. That's a mysterious character. You can think of them as an omniscient being in other stories. There is no need to make a god look like a human. You can, but you don't have to.

You can argue that this new information is only about the "avatar" of this character. For example, jumper can be the AI controlling BitVerse. This AI is sentient and becomes bored after maintaining BitVerse for billions of years. Now, just for fun, it decides to create an avatar and "help" some trapped human minds "escape" the sandbox. With jumper, it only gives the toy (the player) a few hints because it wants some entertainment. Escaping the sandbox? "Good job," jumper thinks. "Welcome to the next prison."

In this plot, having more "human" information about jumper is fine because that's only an avatar. However, how do players know that the "human" information about jumper is merely for the avatar? When they read all these messages, they only see a human jumper, so naturally they will think jumper is human. Without more literature files, the entire plot of jumper being an AI that is tricking the player for fun just disappears. And this is only one plot that I randomly came up with. My point is that jumper does not need an identity. In theory, we can add more information about jumper, but jumper's mysterious nature should be protected as much as possible. Currently, jumper only appears in `.msg` files with very vague messages and on the BitVerse screen. They are basically a character that you cannot reach. Suddenly, with this PR, jumper is just a mere human with very good technical skills who likely knows the truth for some reason. This is like introducing a god in chapter 1 of a story and then later revealing that this "god" is actually just another very strong human. This is not exactly wrong, and it can work in some stories. However, in Bitburner, IMO, this is inappropriate.

In short, I don't think we need to add more information to clarify who or what jumper is. Nobody knows. And nobody will know.

Having more information about jumper does not help players' imagination; it limits their imagination. If they have a human name and a human personality, it's clear they are human. We can claim that everything we say about jumper is merely about their avatar, but then we need to explain this avatar plotline somewhere. And if we do that, we just create a very specific plot about jumper.

I don't think it's frustrating when there is little information about a mysterious "core" character. Jumper is just like Tom Bombadil. Despite Tolkien writing a massive number of stories to expand the world of Middle-earth, he still intentionally never explained who or what Tom Bombadil ultimately was. This is fine. A mysterious character can remain mysterious forever.

hexagonrecursion:
Consider this scenario:
1. Alice is a new player.
2. Alice gets a message from jump3R that contains fl1ght.exe.
3. Alice encounters more references to this mysterious jump3R.
4. The references to jump3R give Alice an impression that she will find out more about jump3R over the course of the game. This is a very common pattern in other games Alice plays.
5. Alice becomes hyped for the eventual reveal of more information about jump3R.
6. The anticipation of the reveal becomes a major part of the motivation to play the game.
7. Alice is hyped for a reveal that never comes.
8. Alice becomes increasingly frustrated: Does she have to beat all the bitnodes? Is there some puzzle that Alice failed to solve? Some easter egg she failed to find?
9. Eventually someone tells Alice that there is no reveal, that jump3R is anything you want jump3R to be. I do not expect Alice to enjoy finding this out.

* This definitely isn't what every player experiences. I, for example, did not get the impression that jump3R's identity or motives will be revealed, but I do expect some players to get that impression.
* Once Alice thinks that there will be a big reveal of jump3R lore, I do not see how we can break her out of the mindset without breaking the 4th wall.
* A more experienced storyteller than me might give some advice on how to reduce the risk of giving Alice the wrong impression in the first place.

catloversg:
I don't have any suggestions about not giving Alice a false expectation of a nonexistent jumper reveal. What I do know is that we should not give Alice a false impression of jumper's identity (a human in this case).

ficocelliguy:
I've removed the references to jump3r, and restyled Perez as a new/original character.

d0sboots:
> Regarding the change to jumper's lore, I know you want to improve the lore, but I think this is the wrong way to do that.
> 
> In terms of the development timeline, Jumper is the most ancient "being" in Bitburner. We added Jumper's messages even before we implemented TRP. You can think of Jumper as a mysterious character whose exact nature nobody knows. The lore about Jumper and the Enders is still vague, and that's actually a good thing.
> 
> If nobody knows who or what Jumper is, they can be anything. They can be an Ender. They can be a rogue Ender. They can be the AI that controls BitVerse under the Enders' command. They can be the AI that controls BitVerse, but that AI became a rogue one. They can be a mere trapped human mind that fails to escape the sandbox. They can be the first human mind to escape the sandbox and want to help other people's minds. Because the lore is vague, people can freely interpret it in any way they want. There is no limitation. If you create canonical lore like this, you basically kill all other interpretations.
> 
> If you want to "flesh out" the lore and add more lore, you should create a new character. That's your character, so you are free to "design" them without interfering with other people's interpretations of the most mysterious character.

To be frank, I completely disagree with this. I think it's fine to add additional characters, but I think it's also fine to flesh out what already exists. Storytelling is fundamentally about *removing* possibilities: Before the story, anything can happen. After the story, only what happened could have happened. Is the state beforehand better, because it had more possibilities? I would emphatically say "no," because the state beforehand had nothing.

In terms of lore jump3r's messages are very obviously a reference to The Matrix. There, the messages were from Trinity (a mysterious hacker) who rapidly becomes a (human) character who is vital to the plot. So, there's certainly parallels to "do more" with this character rather than just leave them hanging.

But I think the most important thing to consider is that the reason BitBurner is very light on lore is simply that all the devs and contributors (all the way back to Daniel) are simply better at coding than writing prose. Which really should not a surprise at all. XD Because of that, I at least am very supportive of just about any additions to the lore. Yes, care still needs to be taken, but IMO Fico has absolutely met the bar needed to be playing with the core characters and concepts (such as they exist) in this game.

catloversg:
I think we have different perspectives here. What you said is not wrong, and I agree that it's technically okay to improve the lore of core characters. However, jumper has been a mysterious character for a very long time, and until now, there has been no canonical interpretation of who jumper is. Naturally, people can have different interpretations of a mysterious character.

Fico thinks of jumper as a human with very good technical skills and a specific personality (very human-like and arrogant). Suppose I think of jumper as an Ender or an AI with nearly omniscient abilities that sees the entire BitVerse as a playground, and the cool "escaping the sandbox" is merely a game involving some trapped minds of an inferior race (the human race). Who is wrong here? Is fico wrong? Am I wrong? The current plot takes heavy inspiration from The Matrix, but that does not mean everything needs to be the same. There is nothing wrong with thinking of jumper as a non-human entity.

Let's say we have two different PRs submitted at the same time. One from fico, and one from another contributor. Both are good, but both give jumper widely different identities. What will we do? This is the same issue with the fandom of any story. When there is a vague concept, everybody has their own interpretation, and it's simply impossible to say which one is "correct".

You are right that Bitburner's lore is very light because most contributors are developers, not writers. However, I want to expand on that a bit more. Bitburner's lore is very light, and this is actually a good thing, IMO. It's light, so it's inherently vague in many places. Because it's vague, people can come up with their own interpretations. If we keep it vague in *appropriate* places, we can avoid the flamewars that always happen in other fandoms. Of course, this does not mean we should make everything vague. The new character from fico is cool, and it's a good addition.

src/Literature/DiscoverableScriptContent/build_new_worlds.wip.jsx
```diff
@@ -0,0 +1,82 @@
...
+  ReactDOM.render(<SidebarItem ns={ns} sidebarLabel={sidebarLabel} pageText={pageText} />, containerNode);
+}
+
+/**
+ * Example component: a sidebar item that opens a new ContentPage when clicked.
+ * @param {NS} ns - netscript api
+ * @param {string} sidebarLabel - the label for the sidebar button
+ * @param {string} pageText - the text to display in the new page
+ * @returns {React.ReactElement}
+ * @constructor
```
catloversg: Don't use `@constructor` for React components.

The syntax for the other params is also incorrect. You need to use:
```js
/**
 * @param {{ ns: NS, sidebarLabel: string, pageText: string }} props The component props.
 */
function SidebarItem({ ns, sidebarLabel, pageText }) {
}
```
Or
```js
/**
 * @param {Object} props The component props.
 * @param {NS} props.ns foo
 * @param {string} props.sidebarLabel bar
 * @param {string} props.pageText baz
 */
function SidebarItem({ ns, sidebarLabel, pageText }) {
}
```

src/Literature/DiscoverableScriptContent/build_new_worlds.wip.jsx
```diff
@@ -0,0 +1,82 @@
...
+      {sidebarLabel}
+    </div>
+  );
+}
+
+/**
+ * Example component: a simple page that displays some text and a button that changes the background color
+ * @param {string} text - text to show on the page
+ * @returns {React.ReactElement}
+ * @constructor
```
catloversg: Same comment

catloversg:
Except for a minor issue with the JSDoc, this PR looks good to me.

Both d0sboots and I stated our opinions about whether jumper's identity should be changed and decided in a PR. Although we have differing opinions on that, it does not have to block this PR. Fico creates a new character, and I'm happy with it. If there are any other suggestions or discussions about the lore, let's do that later or in another PR.

ficocelliguy:
As a note , I made all of the added lore dated from before 2077 (when the game takes place) to help ensure it does not limit what happens going forward in the player's personal story, who any of the characters later become, what happens to you when you enter the bitverse, etc, for maximum imagination / future additions

d0sboots:
> Let's say we have two different PRs submitted at the same time. One from fico, and one from another contributor. Both are good, but both give jumper widely different identities. What will we do? This is the same issue with the fandom of any story. When there is a vague concept, everybody has their own interpretation, and it's simply impossible to say which one is "correct".

Yes, we definitely have different perspectives. In my view, this is not dissimilar to the issue of what happens when two PRs are submitted at the same time for a code issue (which has also happened.)

My answer to that is: Resolving that is what maintainers are *for*. You can't write anything (be it code or prose) entirely by consensus. There will inevitably be conflicts (hopefully rare) that split opinions down the middle, and at these times a central decision needs to made. More broadly, in most projects it's the maintainer's job to provide an overall guidance and continuity so that everything hangs together as a cohesive whole instead of being a bunch of random patches tacked on based on contributor's whims. 

I feel that this is just as relevant when it comes to lore as code. I don't view "old" lore as untouchable any more than old code. In a fandom, many different interpretations can exist, some of them mutually contradictory. But the author/maintainer sets the canon. It is not uncommon for later books/series/etc. to contradict earlier fanworks, and sometimes those continue off on their own parallel histories of headcanon.

It would be *awesome* if these types of issues were ones we had to seriously consider. However, right now Bitburner's lore is so skimpy that there's practically no fandom to speak of. There's a grand total of one hit on AO3 for bitburner, and it's a she-ra fic that's BB-inspired, as opposed to actually playing with our lore.

> You are right that Bitburner's lore is very light because most contributors are developers, not writers. However, I want to expand on that a bit more. Bitburner's lore is very light, and this is actually a good thing, IMO. It's light, so it's inherently vague in many places. Because it's vague, people can come up with their own interpretations. If we keep it vague in _appropriate_ places, we can avoid the flamewars that always happen in other fandoms. Of course, this does not mean we should make everything vague. The new character from fico is cool, and it's a good addition.

I do not view "avoiding flamewars" as an inherently interesting goal. A dead space has no flamewars, and that's essentially where we are at right now. If we actually want something interesting enough to have fans, we will deal with the consequences of having fans.

src/Literature/DiscoverableScriptContent/upgrade_cloud_server.wip.ts
```diff
@@ -19,6 +19,7 @@
    Now written in TypeScript: for even fewer bugs! Probably!
 */
 
+/** @param {NS} ns */
```
hexagonrecursion: This looks wrong to me. You have added `@param {NS} ns` to a typescript file. `@param {NS} ns` is useful in a .js file to get autocomplete, but here `main()` already has a typescriot-style type annotation. `@param {NS} ns` is redundant

ficocelliguy: I was messing with tsconfig and put this in the wrong branch (here instead of my scripts project.) good catch!

catloversg:
I don't think the flamewar is as far off as you think. The differing opinions about who jumper can be in this PR could already be considered a form of flamewar, at least IMO. It's not as heated as the flamewars you see in larger fandoms, but it's fundamentally the same kind of debate. If we ever decide on jumper's identity, these are the things I would insist on.

First, and most importantly, jumper must be a unique character. Don't make jumper turn out to be a copy of Trinity, Morpheus, or Neo. Although Bitburner uses many references to The Matrix, it is not a cheap knockoff. Bitburner already has more than enough original plotlines, such as the Enders and the BitVerse, to create unique and interesting characters. I love The Matrix series, but I'm not interested in reading a copy of it.

Second, jumper must be powerful. Currently, jumper is the most mysterious and powerful character that the player cannot reach. You can "demystify" them by giving them a background story, but don't make them become just another awakened human without anything special. In stories about "escaping the sandbox", there is typically a "guidance" character. This character can be good by nature and genuinely want to help the protagonist (e.g., the Oracle in The Matrix or the Ellimist in Animorphs). This character can also be deceitful and have no real interest in helping (e.g., my previous example of an AI that sees the entire BitVerse as a playground and only "helps" the player for its own entertainment). You are free to choose either archetype, but this character should exist on a meta level rather than simply being someone who awakened earlier and helps the protagonist. For example, in The Matrix, Trinity (pre-Resurrections) and Morpheus are powerful and important to the plot, but they are nowhere near the level I want jumper to be. In my mind, jumper is at least on the level of the Oracle. My previous example of the AI manipulating the player is another example of the level of power I want jumper to have.

catloversg:
To clarify the second requirement in my previous message, I don't want jumper to be just another awakened human without anything special. However, this does not mean jumper cannot be human or have once been human. You can still make jumper a human (now or in the past) as long as they are unique, interesting, and powerful. The BitVerse UI shows the jumper's "name" (JUMP3R), so there are a lot of interesting potential plots for jumper and the BitVerse. For example, this is a plot I came up with.

The simulation machine created by the Enders is much more sophisticated than what we think it is.

Jumper, who was an awakened human, somehow discovered the truth. They found WD, figured out how it worked, and then had an idea. If WD is hacked, BN will encounter a non-recoverable error (the `Restarting BitNode - [ERROR] FAILED TO AUTOMATICALLY REBOOT BITNODE` message implies the system uses a failsafe mechanism to reboot and fix the error, but it fails), and everybody (not just the awakened humans) will be free. However, after jumper does this, they realize that's not as simple as they think. WD is actually only a machine-in-machine. They break the first simulation layer just to enter the second layer. I.e., after the BN is "destroyed", the system simply spins up a new BN instance and puts everybody's mind back to that new instance.

After going back to the BN, they feel devastated. However, they soon discover a subtle difference: they feel stronger in almost every aspect. They wonder if this is just the placebo effect, so they break the BN again to test it. They find out they actually become stronger every time they break the BN. However, this effect only happens the first 3 times. They realize the simulated world is not a simple digital prison; it has many secrets, and they should carefully explore the entire simulated world.

After years of exploration, they find a lot of hints about *different* kinds of BNs. Their BN is only one of many BNs. Each BN is basically a separate "universe" with very different "rules" and different "rewards". This is the eureka moment for them. What if they "jump" to other BNs, destroy them and obtain powerful "power ups" from them? With enough power, there is a chance they can actually break the second simulation layer. Every time they break the BN, they have a chance to observe how the system works. Based on their observation, jumping to other BNs is theoretically possible, but it's not easy.

They try to break the BN countless times to gain more knowledge of the system. The more knowledge they gain, the more confidence they have in the idea of jumping to other BNs. However, with more knowledge of the system, they understand that this idea is impractical. It's simply impossible to "jump" in a system that does not define what "jump" means. The system does not specifically disallow this action, but it needs to be "implemented" beforehand.

Now they come up with a crazy idea. If the system does not have this feature, why don't they just "implement" this feature? They are excited by this idea and start finding a way to do that. After countless attempts, they finally find a way to do that, but it comes with a price. Just as a ship needs a navigation system, jumping between BNs requires a "map". This map is not a simple map like the paper maps that ships used centuries ago to navigate. This map is an incredibly complicated orchestration program running inside the system. The system does not have this program because there is no need for it, but it allows the creation of this program.

They find a secret way to implement this program, but the cost is that they become the "brain" of this program. In other words, they have to become part of the system. The system has many restrictions, so they cannot write a random program that breaks the system itself. However, as long as they operate within these restrictions, they can change how the system works in a subtle way. It won't break the system, but it gives future awakened humans a better chance of breaking the system.

*The BitVerse screen (not the BitVerse concept) is the front-end of a program running inside the simulation machine. It makes the impossible possible. It opens new paths for humanity.*

*This is jumper. No longer human. Now a program. The one giving humanity hope.*

I hope this plot highlights what I want in jumper's identity. They are not merely a human who awoke before the player and wanted to help them. It's fine if they awoke earlier and wanted to help, but that alone is not enough; it's too boring. In the example plot, although being awakened is still important, it alone is not enough to break the simulation. When you hack WD, you simply escape your current prison only to be placed back into the same prison again. There may have been countless humans who awoke before jumper and the player, but they could not do anything. Jumper is special because they are not an ordinary awakened human. They found a path for future awakened humans and sacrificed themselves to open that path.

Obviously, there are thousands of potential plots for jumper, including the plot about the AI that tricks the player for entertainment. My point is that the mysterious and powerful jumper must be an original and interesting character. When writing about jumper's origins, the new details should deepen their mystique rather than reduce them to "someone who figured it out before the player".

ficocelliguy:
We want these scripts to not compile on run, or have similarly glaring problems, which is pretty far from that cool vibe of a mystical, powerful character. Given those lore goals, I'll leave Perez as-is and not link jumper's name to these intentionally weak, broken code snippets.

catloversg:
Yeah. I'm mostly discussing the potential lore changes. I'll submit a PR focused on jumper's lore later. This PR is good as is. While writing this example plot, I deliberately structured it so that we can introduce many more awakened humans in the future. Perez in this PR is one example.

src/Literature/DiscoverableScriptContent/build_new_worlds.wip.jsx
```diff
@@ -0,0 +1,78 @@
...
+  ns.ui.openTail();
+  addToSidebar;(ns, "֍   The Next Step", 'this world is not what it seems');
+  ns.tprint('Custom content added to sidebar');
+
+  // Prevent the script from exiting, so react components can still use ns methods
+  await new Promise(resolve => {});
+}
+
+/**
+ * Creates a clickable SidebarItem and adds it to the game's left-hand menu.
```
d0sboots: Optional: These function comments are good for understanding. If possible, reword the top-level bits here to be more in Junie's "voice".

src/Literature/DiscoverableScriptContent/escaping_your_chains.wip.jsx
```diff
@@ -0,0 +1,46 @@
+/*
+  A true hacker is one who works to learn the secrets of every tool and tech they use.
+  It's not about cracking security or pwning servers - it's a craftsman's mindset, the application of ingenuity.
+
+  Whether you are doing and-dirty patchwork or elegant architecture, know your tools, understand the problem, and
```
d0sboots:
```suggestion
  Whether down-and-dirty patchwork or elegant architecture, know your tools, understand the problem, and
```
You're mixing sayings here and it doesn't end up grammatical. I suggested one way out, there are others.

src/Literature/DiscoverableScriptContent/escaping_your_chains.wip.jsx
```diff
@@ -0,0 +1,46 @@
...
+  apply your craft to solve it. Don't just use someone else's script as-is or ask a rogue AI to write it for you!
+  Piece it together yourself from whatever you can find, and make it your own. Break it, fix it, and understand it.
+
+   This world is not what it seems. You can break its boundaries and push what is possible. Never be complacent!
+
+   If you find this, it means I'm already gone ahead. I hope I've left enough hints behind.
```
d0sboots:
```suggestion
   If you find this, it means I've already gone ahead. I hope I've left enough hints behind.
```

src/Literature/DiscoverableScriptContent/hack.wip.js
```diff
@@ -0,0 +1,8 @@
+/** @param {NS} ns */
```
d0sboots: Optional: Maybe remove the boilerplate comment line here?

It would go with this being a very small snippet, and (maybe) people expanding on this would get a hint about how the lack of it makes typing worse.

src/Literature/DiscoverableScriptContent/hack_controller.wip.js
```diff
@@ -0,0 +1,45 @@
+/*
+                        ** Hack Loop Controller **
+                (c) 2074 Fulcrum Tech (Secret Lab Division)
+                             Author: J. Perez
+
+ This script launches child scripts that loop just one of either hack, grow, or
+ weaken. They run alongside each other in order to constantly keep the server
+ near maximum money (to increase the power of hack and grow) and minimum security
+ (to reduce the penalties for high security on runtime and money hacked)
```
d0sboots:
```suggestion
 This script launches children that loop just one of either hack, grow, or
 weaken. They run alongside each other in order to constantly keep the server
 near max cash (to increase the power of hack and grow) and minsec
 (to reduce the penalties for high security on runtime and money hacked)
```
use jargon to sound more knowledgeable (but not so much that it's too obfuscated)

src/Literature/DiscoverableScriptContent/port_receiver.wip.js
```diff
@@ -0,0 +1,47 @@
...
+        A proposal for a new standard data transfer protocol using Netscript ports.
+
+  Coordinating actions and data between scripts is a key part of network architecture. This script demonstrates
+  a simple way to start a child process, let it do work for some time, and have it send its results back.
+
+  This allows a parent script to delegate tasks to other scripts, which allows for concurrent processing,
+  and reduces the continuous RAM load required of calling expensive netscript methods.
+ */
+
+export const DATA_PORT = 31337;
```
d0sboots, line 14: 👌

src/Literature/DiscoverableScriptContent/port_receiver.wip.js
```diff
@@ -0,0 +1,47 @@
...
+function senderScriptIsRunning(ns) {
+  const runningScripts = ns.ps();
+  return runningScripts.find((script) => script.filename === 'port_sender.wip.js');
+}
+
+export function colorText(text, colorHexCode) {
+  const r = parseInt(colorHexCode.slice(1, 3), 16);
+  const g = parseInt(colorHexCode.slice(3, 5), 16);
+  const b = parseInt(colorHexCode.slice(5, 7), 16);
+  const colorCode = `\x1b[38;2;0x${r};0x${g};0x${b}m`;
```
d0sboots, line 46:
```suggestion
  // ansi color ftw
  const colorCode = `\x1b[38;2;0x${r};0x${g};0x${b}m`;
```
The magic escape sequence probably deserves *some* comment - I attempted to match how I feel about her voice XD

src/Literature/DiscoverableScriptContent/port_sender.wip.js
```diff
@@ -0,0 +1,44 @@
...
+
+/** @param {NS} ns */
+export async function main(ns) {
+  ns.ui.openTail();
+  ns.ui.moveTail(600, 80);
+
+  if (!receiverScriptIsRunning(ns)) {
+    ns.print(colorText('Starting data receiver script...', '#8855FF'));
+    ns.run('port_receiver.wip.js');
+    await ns.sleep(10); // let go of the thread so the other script can actually start
```
d0sboots:
```suggestion
    await ns.sleep(0); // let go of the thread so the other script can actually start
```
Show people that you can sleep for 0 (and it actually has a purpose)

src/Literature/Enums.ts
```diff
@@ -34,3 +34,17 @@ export enum LiteratureName {
...
+  PortSender: "port_sender.wip.js",
+  PortReceiver: "port_receiver.wip.js",
+  ServerFinder: "server_finder.wip.js",
+  CustomLogScript: "escaping_your_chains.wip.jsx",
+  CustomContentScript: "build_new_worlds.wip.jsx",
+  HackController: "hack_controller.wip.js",
+  HackChildScript: "hack.wip.js",
+} as const;
+
+export type DiscoverableScriptName = (typeof DiscoverableScriptName)[keyof typeof DiscoverableScriptName];
```
d0sboots: I think we should use the same naming schema as NetscriptDefinitions for enums we build this way.

src/Server/ServerHelpers.ts
```diff
@@ -376,10 +379,22 @@ export function initForeignServers(homeComputer: Server): void {
...
-    for (const filename of metadata.literature || []) {
+    for (const filename of metadata.literature ?? []) {
       server.messages.push(filename);
     }
 
+    for (const scriptName of metadata.discoverableScripts ?? []) {
+      const path = resolveScriptFilePath(scriptName);
+      const content = discoverableNetworkScripts[scriptName].content;
+      if (!path || !content) {
+        console.error(
```
d0sboots: I'd go with a throw for now. This seems like a serious error, and if it occurs it should happen right on startup.

d0sboots (commented):
Looks good overall!

I think the changes you made leave the possibility of Junie == Jump3r open without explicitly confirming it. IMO that's a good balance of the "mystery" that CatLover was talking about.

src/Literature/Enums.ts
catloversg: One advantage of defining both the as const object and the corresponding type with the same name is that consumers can simply write `import { DiscoverableScriptName } from "@enums";`. Using separate names, in the same style as `NetscriptDefinitions.d.ts`, is also fine (it's clear which one is the type and which one is the object). I'm okay if you want to do that.

This is one of the topics I'll cover in the WIP style guide I'm writing, so we should decide which style we want to adopt going forward. I'll make the change after your confirmation.

src/Literature/DiscoverableScriptContent/build_new_worlds.wip.jsx
ficocelliguy: I've re-worded them to be less dry-api-doc-style

src/Literature/Enums.ts
catloversg: @ficocelliguy Don't switch back to enum. For new enum-like data, we will use as const object. What I'm discussing with d0sboots is the naming convention for the object and the corresponding type.

ficocelliguy: reverted.

d0sboots: I think using the same name for two separate things, although legal, is confusing. Also, IMO it's a bad idea to have a separate style from what we've established in NetscriptDefinitions.d.ts.

The downside is worse naming and the fact that we end up with split internal styles unless/until we migrate internal enums.

PR #2956, opened 2026-07-05
UI: Change how server summary displays dnet-specific information

catloversg:
The server summary currently displays at most three pieces of dnet-specific information. This can easily lead to unnecessary debates about which pieces should be prioritized. The box has enough space for four pieces, so we should display four instead of three. If there are more, the fourth item will be an icon that shows all remaining pieces on hover.

I also fixed an issue introduced in #2944. The current code assumes that a frozen server cannot have a stasis link, but that's not true. Players can set the stasis link on a server, then freeze it. That's a stupid thing to do, but it can happen.

Before:

before

After:

after

### DIFF
--- a/src/DarkNet/ui/ServerSummary.tsx
+++ b/src/DarkNet/ui/ServerSummary.tsx
@@ -184,7 +184,7 @@ export function ServerSummary({
       ? components
       : [
           ...components.slice(0, 3),
-          <Tooltip key="others" title={components.slice(3)}>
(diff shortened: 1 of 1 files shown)


ficocelliguy:
Can you make the extra icons on the tooltip horizontal instead of vertical? there's a bit more room that way, and it's consistent with the first few visible on the display.

catloversg:
Capture

ficocelliguy (approved):

PR #2957, opened 2026-07-05
UI: Do not show Entropy virus notice when grafting Congruity Implant

E9cipher:
First off, sorry for the typo in the commit :)
This PR modifies the grafting popup when the augmentation is the Congruity Implant it the text "grafting an augmentation will increase the potency of the entropy virus" changes to "This augmentation will remove the Entropy virus effect".

### DIFF
--- a/src/PersonObjects/Grafting/ui/GraftingRoot.tsx
+++ b/src/PersonObjects/Grafting/ui/GraftingRoot.tsx
@@ -201,16 +201,14 @@ export const GraftingRoot = (): React.ReactElement => {
                   <Typography component="div" paddingBottom="1rem">
                     Cancelling grafting will <b>not</b> save grafting progress, and the money you spend will <b>not</b>{" "}
                     be returned.
-                    {!Player.hasAugmentation(AugmentationName.CongruityImplant) && (
-                      <>
-                        <br />
-                        <br />
-                        {selectedAug !== AugmentationName.CongruityImplant &&
-                          "Additionally, grafting an augmentation will increase the potency of the Entropy virus."}
-                        {selectedAug === AugmentationName.CongruityImplant &&
-                          "This augmentation will remove the Entropy virus effect."}
-                      </>
-                    )}
+                    {!Player.hasAugmentation(AugmentationName.CongruityImplant) &&
+                      selectedAug !== AugmentationName.CongruityImplant && (
+                        <>
+                          <br />
+                          <br />
+                          Additionally, grafting an augmentation will increase the potency of the Entropy virus.
+                        </>
+                      )}
                   </Typography>
                 }
               />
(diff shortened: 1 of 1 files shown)


src/PersonObjects/Grafting/ui/GraftingRoot.tsx
```diff
@@ -205,7 +205,10 @@ export const GraftingRoot = (): React.ReactElement => {
                       <>
                         <br />
                         <br />
-                        Additionally, grafting an augmentation will increase the potency of the Entropy virus.
+                        {selectedAug !== AugmentationName.CongruityImplant &&
```
catloversg: This condition should be combined with the `!Player.hasAugmentation(AugmentationName.CongruityImplant)` check above.

src/PersonObjects/Grafting/ui/GraftingRoot.tsx
```diff
@@ -205,7 +205,10 @@ export const GraftingRoot = (): React.ReactElement => {
...
+                          "Additionally, grafting an augmentation will increase the potency of the Entropy virus."}
+                        {selectedAug === AugmentationName.CongruityImplant &&
```
catloversg: The "Additionally, grafting ..." message above is a reminder just in case players don't notice the Entropy virus effect. This message is unnecessary. If they graft this augmentation, they should already know what it does.

E9cipher:
I'll update it so the messsage "Additionally..." doesn't show up when grafting the Congruity Implant, then?

catloversg:
Yes.

src/PersonObjects/Grafting/ui/GraftingRoot.tsx
```diff
@@ -201,16 +201,14 @@ export const GraftingRoot = (): React.ReactElement => {
...
-                      <>
-                        <br />
-                        <br />
-                        {selectedAug !== AugmentationName.CongruityImplant &&
-                          "Additionally, grafting an augmentation will increase the potency of the Entropy virus."}
-                        {selectedAug === AugmentationName.CongruityImplant &&
-                          "This augmentation will remove the Entropy virus effect."}
-                      </>
-                    )}
+                    {!Player.hasAugmentation(AugmentationName.CongruityImplant) &&
```
catloversg: On second thought, this condition can be simplified to only `selectedAug !== AugmentationName.CongruityImplant`. If players can select this augmentation, `Player.hasAugmentation()` is false.

E9cipher: Yeah, that's true. I'll remove that extra check then

E9cipher: Wait. There's actually a misfire here. What if the player already has the Congruity Implant? Then the condition `selectedAug !== AugmentationName.CongruityImplant` is true, but the player has it installed. This check cannot be removed.
I'll submit a commit to revert `#73cd682`

catloversg (approved):

d0sboots:
Code doesn't match the PR description anymore really, but that's OK

PR #2958, opened 2026-07-06
ELECTRON: Add command line parameter to disable Steam integration

aef123:
Adding Command Line Parameter to Disable Steam Integration
CATEGORY: MISC

This PR enables the '--no-steam' command line parameter. If set, it will not initialize Steamworks.

Steam integration is optional. However, force loading Steamworks can cause issues for some players. In some cases these issues manifest as an error dialog saying Steam is not installed. In other cases, users may find Steam does not allow them to run both BitBurner and Steam games at the same time. While neither of these issues are fatal, they can be annoying to users who do hit them. 

This settings is off by default. If the command line parameter isn't set, it will run the game like normal with steam integration.

### DIFF
--- a/electron/steamworksUtils.js
+++ b/electron/steamworksUtils.js
@@ -54,17 +54,13 @@ function verifyLibraryFiles() {
   return { success: true };
 }
 
-/** Steam integration can be disabled entirely with the --no-steam command-line flag. When disabled, we never load
- * steamworks.js or connect to the Steam client, and we don't set global.steamworksError, so the "Could not connect
- * to Steam" dialog is skipped. Steam Cloud and achievements are already no-ops when steamworksClient is undefined. */
-const steamworksDisabled = process.argv.includes("--no-steam");
-
+/** Steam integration is optional. Players can disable it by passing the --no-steam flag. */
 let steamworks;
 try {
-  if (!steamworksDisabled) {
+  if (!process.argv.includes("--no-steam")) {
     steamworks = require("@catloversg/steamworks.js");
   } else {
-    log.info("Steam integration disabled by --no-steam flag");
+    log.info("Steam integration was disabled by --no-steam flag");
   }
 } catch (error) {
   log.error(error);
--- a/electron/steamworksUtils.js
+++ b/electron/steamworksUtils.js
@@ -54,9 +54,18 @@ function verifyLibraryFiles() {
   return { success: true };
 }
 
+/** Steam integration can be disabled entirely with the --no-steam command-line flag. When disabled, we never load
+ * steamworks.js or connect to the Steam client, and we don't set global.steamworksError, so the "Could not connect
+ * to Steam" dialog is skipped. Steam Cloud and achievements are already no-ops when steamworksClient is undefined. */
+const steamworksDisabled = process.argv.includes("--no-steam");
+
 let steamworks;
 try {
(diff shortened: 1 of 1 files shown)


electron/steamworksUtils.js
```diff
@@ -54,9 +54,18 @@ function verifyLibraryFiles() {
   return { success: true };
 }
 
+/** Steam integration can be disabled entirely with the --no-steam command-line flag. When disabled, we never load
+ * steamworks.js or connect to the Steam client, and we don't set global.steamworksError, so the "Could not connect
+ * to Steam" dialog is skipped. Steam Cloud and achievements are already no-ops when steamworksClient is undefined. */
```
catloversg:
```suggestion
/** Steam integration is optional. Players can disable it by passing the --no-steam flag. */
```

electron/steamworksUtils.js
```diff
@@ -54,9 +54,18 @@ function verifyLibraryFiles() {
...
+const steamworksDisabled = process.argv.includes("--no-steam");
```
catloversg: Inline this variable and its comments.

electron/steamworksUtils.js
```diff
@@ -54,9 +54,18 @@ function verifyLibraryFiles() {
...
+
 let steamworks;
 try {
-  steamworks = require("@catloversg/steamworks.js");
+  if (!steamworksDisabled) {
+    steamworks = require("@catloversg/steamworks.js");
+  } else {
+    log.info("Steam integration disabled by --no-steam flag");
```
catloversg:
```suggestion
    log.info("Steam integration was disabled by --no-steam flag");
```

catloversg (commented):
I agree with this change because Steam integration is optional. If players run the game on a machine that does not have Steam installed, we should add a way to launch the game without showing the Steam integration error popup.

However, you need to fact-check the claim that Steam blocks running other Steam games while the Electron version of Bitburner is running. Running multiple Steam games at the same time is a very common use case. For example, many people play multiple idle Steam games simultaneously. I just tested this by launching Bitburner from Steam, then launching another Steam game. There was no issue. If you have an issue, then it's likely specific to your system or to those other games.

I suggest rewriting the description to highlight that this is an issue on your system.

aef123:
> I agree with this change because Steam integration is optional. If players run the game on a machine that does not have Steam installed, we should add a way to launch the game without showing the Steam integration error popup.
> 
> However, you need to fact-check the claim that Steam blocks running other Steam games while the Electron version of Bitburner is running. Running multiple Steam games at the same time is a very common use case. For example, many people play multiple idle Steam games simultaneously. I just tested this by launching Bitburner from Steam, then launching another Steam game. There was no issue. If you have an issue, then it's likely specific to your system or to those other games.
> 
> I suggest rewriting the description to highlight that this is an issue on your system.

Looks like the issue may not be running multiple games at once on the same machine. It's due to running multiple games on different machines. I've been trying to use my Steamdeck while BitBurner is running on my gaming PC. Apparently running steam games on multiple machines at the same time is the issue. 

Whatever the root cause, I've updated the comment to be more inclusive.

aef123:
Updated based on PR suggestions.

catloversg (approved):

d0sboots:
Documenting this parameter would be good, but it doesn't need to be this PR.

PR #2959, opened 2026-07-06
DOCS: Improve CONTRIBUTING and README

E9cipher:
I added some more instructions to the CONTRIBUTING document so it's easier for first-time contributors to read it and understand it.
I have also fixed some relative paths (`/src` to `./src`), clarified a few things, updated the npm LTS version and added links for the discord channels. I also added a TL;DR section with quick instructions to get started as fast as possible (and still check the whole document, but only when needed, one step at a time)


catloversg:
@d0sboots Don't merge this PR. I'll rewrite it.

catloversg:
I don't think we need to add the TLDR section, but I'll let d0sboots make the call.

d0sboots:
> @d0sboots Don't merge this PR. I'll rewrite it.

Oops, I was ignoring this PR thinking you were going to open a different one.

PR #2960, opened 2026-07-07
UI: Add achievement for visiting The Void

E9cipher:
This adds an achievement for when the user enters The Void.

### DIFF
--- a/src/Locations/ui/TheVoid.tsx
+++ b/src/Locations/ui/TheVoid.tsx
@@ -1,5 +1,4 @@
 import { LocationName } from "@enums";
-import { gotoLocation } from "src/PersonObjects/Player/PlayerObjectGeneralMethods";
 import { Player } from "@player";
 import React, { useEffect } from "react";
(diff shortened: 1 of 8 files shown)


src/Locations/ui/TheVoid.tsx
```diff
@@ -0,0 +1,10 @@
+import { LocationName } from "@enums";
+import { gotoLocation } from "src/PersonObjects/Player/PlayerObjectGeneralMethods";
+import { Player } from "@player";
+import React, { useEffect } from "react";
+export function TheVoid(): React.ReactElement {
+  useEffect(() => {
+    Player.gotoLocation(LocationName.Void);
```
E9cipher: I had to add this because the `Player.location` was "Travel Agency". I suspect that the location remains unchanged because of the `Router.toPage`, which changes the UI but not the player's location. Let me know if there's a more efficient and clean approach.

catloversg (approved):

PR #2965, opened 2026-07-08
ARCADE: Add Zork I, II, and III as playable cabinets in the New Tokyo Arcade

aef123:
Adds Zork I, II, and III as playable machines in the New Tokyo Arcade, next to the Megabyte Burner 2000. These are the original games: Microsoft open-sourced them under MIT (github.com/historicalsource), and the interpreter is the Parchment project's pure-JS Z-machine (ZVM from ifvms.js, also MIT). All third-party material lives in `src/Arcade/Zork/` with a scoped `LICENSE.md`; the rest of Bitburner's licensing is untouched.

- No new npm dependencies, no jQuery, no WASM. The display is a custom React terminal styled from `Settings.theme`, with an inverse-video status line and command history.
- Keyboard input stays inside the cabinet, so playing doesn't trigger Bitburner hotkeys.
- In-game `save`/`restore` persist to localStorage (works in web and Electron). Nothing touches the Bitburner save schema.
- Pure flavor: no gating, no rewards, nothing to balance.
- Bundle impact is ~600KB raw (interpreter + story files); story files load only when a cabinet is opened.
- Eleven jest unit tests run in existing CI, including one that headlessly plays the Zork I opening through the real interpreter and round-trips a save.

## Screenshots

Arcade menu with the Zork cabinets

Zork I running in the arcade


catloversg:
Bitburner is not a game hub for playing other games.

d0sboots:
To expand on catlover's response: This is the sort of thing that can be done as a user script, and has been. (I think someone made a userscript version of the Doom-in-JS port at one point.) At that level, it's something we very much encourage. But it's not something we're going to build in to the base game, precisely *because* we want to encourage people adding things like this via user scripting.

aef123:
No problem. Maybe I'll port it to a user script, but probably not. I'll probably just keep it in my fork of the game.

PR #2967, opened 2026-07-08
JEST: Add more tests for material production limit

catloversg:
Context: https://github.com/bitburner-official/bitburner-src/pull/2953#discussion_r3538209669

The current tests assume there is always enough free space. This PR adds tests for cases where there is insufficient free space.

### DIFF
--- a/test/jest/Corporation.test.ts
+++ b/test/jest/Corporation.test.ts
@@ -364,7 +364,7 @@ describe("production", () => {
     ns.corporation.bulkPurchase(div1, city, "Drugs", 1e6);
     ns.corporation.makeProduct(div2, city, "0", 1e10, 1e10);
     ns.corporation.sellProduct(div2, city, "0", "MAX", "MP", true);
-    /* eslint-disable @typescript-eslint/no-non-null-assertion */
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
     corp.divisions.get(div2)!.products.get("0")!.developmentProgress = 99.99;
     processCycles(10);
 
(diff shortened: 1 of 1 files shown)


test/jest/Corporation.test.ts
```diff
@@ -373,4 +373,69 @@ describe("production", () => {
...
+    setUpDivision(ns, div);
+    ns.corporation.bulkPurchase(div, city, "Chemicals", 1e6);
+    ns.corporation.bulkPurchase(div, city, "Water", 1e6);
+    ns.corporation.bulkPurchase(div, city, "Real Estate", 1e6);
+    processCycles(2);
+
+    ns.corporation.limitMaterialProduction(div, city, "Plants", 20);
+    ns.corporation.limitMaterialProduction(div, city, "Food", 10);
+    // Manually set the free space.
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
d0sboots: Optional: You could assign
```
const warehouse = corp.divisions.get(div)?.warehouses[city];
expect(warehouse).toBeDefined();
```
Up at the top and avoid the ugly comments

test/jest/Corporation.test.ts
```diff
@@ -373,4 +373,69 @@ describe("production", () => {
     expect(Number.isFinite(ns.corporation.getProduct(div2, city, "0").effectiveRating)).toBe(true);
     expect(Number.isFinite(corp.totalAssets)).toBe(true);
   });
+
+  test("limitMaterialProduction 3", () => {
```
d0sboots: Since these tests are all very similar, can you add a comment at the top explaining what scenario the specific parameters are for?

test/jest/Corporation.test.ts
catloversg: Jest functions do not narrow the type, so after `expect(warehouse).toBeDefined()`, from tsc's perspective, `warehouse` can still be undefined. Instead of using `toBeDefined`, I use a normal falsy check.

PR #2971, opened 2026-07-10
JEST: Allow setting hostname when creating mocked NS context

catloversg:
After #2890, `workerScript.hostname` is a getter, so we can no longer set it via `ctx.workerScript.hostname`. This still works in Jest tests because they use a mocked context instead of a real `WorkerScript` instance. Instead of suppressing lint errors, we should create a proper context.

I also added a TODO comment. Fico works around the flaky test by increasing the timeout, but we should investigate why it happens in the first place.

### DIFF
--- a/test/jest/Netscript/Darknet.test.ts
+++ b/test/jest/Netscript/Darknet.test.ts
@@ -320,7 +320,7 @@ describe("home", () => {
     await ns.singularity.installBackdoor();
     // Can exec from home
     expect(ns.exec(scriptPath, dnetServerHostname)).toBeGreaterThan(0);
-  }, 8000);
+  }, 8000); // TODO: Check why this test is flaky.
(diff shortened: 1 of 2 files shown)


d0sboots:
I fixed this test in a different way that I believe has been outstanding for a while.

catloversg:
Do you mean you have an alternative fix in a WIP PR that you haven't opened?

d0sboots:
I thought it was in one I *had* opened. I'll have to go through my branches and double-check.

catloversg:
I'm starting to review your other PRs. While doing that, I found what you said. Your fix is part of #2889. Besides the errors in `Darknet.test.ts`, I found another type error in `AllServers.test.ts`. I'll fix all these errors in the PR that extends the lint task.

PR #2972, opened 2026-07-10
UI: Fix several modal dialogues in corporation UI sometimes showing wrong values

### LINKED ISSUE #416: Corps - Quality of life issues (10 key enter fails, and previous sell not offered as default)

hexagonrecursion:
* closes #416

## Proposed solution
https://react.dev/learn/preserving-and-resetting-state#option-2-resetting-state-with-a-key

### DIFF
--- a/src/Corporation/ui/MaterialElem.tsx
+++ b/src/Corporation/ui/MaterialElem.tsx
@@ -123,7 +123,7 @@ export function MaterialElem(props: IMaterialProps): React.ReactElement {
           </Tooltip>
         </Box>
 
-        <Box sx={{ "& button": { width: "100%" } }}>
+        <Box key={`${division.name}-${city}`} sx={{ "& button": { width: "100%" } }}>
           <Tooltip
             title={tutorial ? <Typography>Purchase your required materials to get production started!</Typography> : ""}
           >
--- a/src/Corporation/ui/MaterialElem.tsx
+++ b/src/Corporation/ui/MaterialElem.tsx
@@ -123,7 +123,7 @@ export function MaterialElem(props: IMaterialProps): React.ReactElement {
           </Tooltip>
         </Box>
 
-        <Box key={`${division.name}-${city}`} sx={{ "& button": { width: "100%" } }}>
+        <Box sx={{ "& button": { width: "100%" } }}>
           <Tooltip
             title={tutorial ? <Typography>Purchase your required materials to get production started!</Typography> : ""}
           >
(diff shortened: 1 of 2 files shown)


d0sboots:
Although this might end up being the most pragmatic solution, to me it smells like we're messing up our state in a deeper way with these Corp dialogs. The box you are adding the key to does not have state of its own; instead you are using the side-effect of it being a new instance to deal with state issues in the subcomponents.

hexagonrecursion:
-  I could instead add a `key` to each modal, if you think this makes it easier to understand the intent
- Semantically a SellMaterialModal for Plants in Sector-12 and SellMaterialModal for Plants in Aevum are two separate things, but without a `key` or without rendering a component in different positions (see link in my previous post) the react reconciliation algorithm has no way to obtain this semantic information
- In addition to option 1 and option 2 from the provided link, there is an option to change corporation modals to nest all state inside `@mui/material/Modal`, which unmounts its children when the modal is closed

catloversg:
I think this kind of modal should belong in a higher level parent component. Currently, each division has one `PurchaseMaterialModal` for each material, but we actually only need one modal owned by the division component or the corp component. When clicking the "Buy" button, the child component opens the modal via a callback received from the parent component. However, this is not how all modals in corp are designed, so I'm fine with setting the key.

Note that the key should be on the modal, not on the wrapper (`Box` in the case of `PurchaseMaterialModal`).

hexagonrecursion:
> Note that the key should be on the modal, not on the wrapper

@catloversg Thank you. Done

PR #2974, opened 2026-07-10
CORPORATION: Explain how production scales due to insufficient storage space

catloversg:
Context: https://github.com/bitburner-official/bitburner-src/pull/2953#discussion_r3538209669

### DIFF
--- a/src/Documentation/doc/en/advanced/corporation/division-raw-production.md
+++ b/src/Documentation/doc/en/advanced/corporation/division-raw-production.md
@@ -13,9 +13,29 @@ Each division has a number that I call "Division raw production". This raw value
 - Calculating how much input material that we need. It's multiplied by the input material's coefficient to find the required quantity of that input material.
 - Calculating how much material/product that division can produce. It's multiplied by `ProducibleFrac`. `ProducibleFrac` starts at 1 and is reduced if there are not enough input materials.
 
-For example, with Agriculture, if `RawProduction` is 1000, we need 500 units of Water and 200 units of Chemicals. With these input materials, we can produce 1000 units of Plants and 1000 units of Food.
+For example, with Agriculture, Suppose `RawProduction` is 1000, we consume 500 units of Water and 200 units of Chemicals
+and produce 1000 units of Plants and 1000 units of Food.
 
-### Formula
+## Production scaling due to insufficient storage space
+
+`RawProduction` will be scaled down if there is insufficient free space in the warehouse. The games calculates the net
+change in storage space, then scale down `RawProduction` based on the net change and the free space.
+
+For example, with Agriculture, for each unit of `RawProduction`, we consume 0.5 units of Water and 0.2 units of Chemicals
+to produce 1 unit of Plants and 1 unit of Food. The sizes of these materials:
+
+- Water: 0.05
+- Chemicals: 0.05
+- Plants: 0.05
+- Food: 0.03
+
+The net change in storage is: `0.05 + 0.03 - (0.5 * 0.05 + 0.2 * 0.05) = 0.045`.
+
+This means for each unit of `RawProduction`, we need 0.045 units of free space. Suppose `RawProduction` is 1000 and
+free space is 22.5. We need 45 units of free space for 1000 units of `RawProduction`, but there are only 22.5 units of
+free space, so the effective `RawProduction` is scaled down to 500.
+
+## Formula
 
 `RawProduction` is the product of 4 multipliers:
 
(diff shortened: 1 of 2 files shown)


hexagonrecursion (approved):

src/Documentation/doc/en/advanced/corporation/division-raw-production.md
```diff
@@ -13,9 +13,29 @@ Each division has a number that I call "Division raw production". This raw value
...
 - Calculating how much material/product that division can produce. It's multiplied by `ProducibleFrac`. `ProducibleFrac` starts at 1 and is reduced if there are not enough input materials.
 
-For example, with Agriculture, if `RawProduction` is 1000, we need 500 units of Water and 200 units of Chemicals. With these input materials, we can produce 1000 units of Plants and 1000 units of Food.
+For example, with Agriculture, Suppose `RawProduction` is 1000, we consume 500 units of Water and 200 units of Chemicals
+and produce 1000 units of Plants and 1000 units of Food.
 
-### Formula
+## Production scaling due to insufficient storage space
+
+`RawProduction` will be scaled down if there is insufficient free space in the warehouse. The games calculates the net
```
d0sboots:
```suggestion
`RawProduction` will be scaled down if there is insufficient free space in the warehouse. The game calculates the net
```

src/Documentation/doc/en/advanced/corporation/division-raw-production.md
```diff
@@ -13,9 +13,29 @@ Each division has a number that I call "Division raw production". This raw value
...
+change in storage space, then scale down `RawProduction` based on the net change and the free space.
```
d0sboots:
```suggestion
change in storage space, then scales down `RawProduction` based on the net change and the free space.
```

src/Documentation/doc/en/advanced/corporation/division-raw-production.md
```diff
@@ -13,9 +13,29 @@ Each division has a number that I call "Division raw production". This raw value
...
+
+For example, with Agriculture, for each unit of `RawProduction`, we consume 0.5 units of Water and 0.2 units of Chemicals
```
d0sboots, line 24:
```suggestion
For example, with Agriculture, for each unit of `RawProduction` we consume 0.5 units of Water and 0.2 units of Chemicals
```

d0sboots (approved):
nits

src/Documentation/doc/en/advanced/corporation/division-raw-production.md
catloversg: This seems like a matter of style. Some style guides consider the comma here optional and say it can be omitted, depending on clarity and rhythm. Personally, I think using the comma here is clearer.

d0sboots: Interesting, do you have a citation for that? Some quick googling didn't find me anything.
If you are thinking of the <https://en.wikipedia.org/wiki/Serial_comma>, that doesn't apply here. This is not a list of noun-phrases. Rather, this is commas separating clauses, and as I understand it the rules operate like this:

* For each A we do B.
* For X, for each A we do B.
* For X, for Y, for each A we do B.

catloversg: I'm referring to commas that separate introductory elements from the main clause, not commas that separate items in a list. The main clause is "we consume ...", and it is preceded by three introductory elements: "For example", "with Agriculture", and "for each unit of RawProduction". The comma after "RawProduction" marks the end of the introductory elements and the beginning of the main clause. Some style guides recommend using it after introductory phrases, especially longer ones, though some allow omission for short phrases. For example, https://owl.purdue.edu/owl/general_writing/punctuation/commas/extended_rules_for_commas.html: "Use commas after introductory a) clauses, b) phrases, or c) words that come before the main clause."

In this case, I consider "for each unit of RawProduction" an introductory phrase, like "For X" and "For Y" in your example. Likewise, I interpret "for each A" in your example as an introductory phrase modifying the main clause "we do B". That's why I think the comma is appropriate here, especially given the other introductory elements that precede it. For a short introductory phrase like "for each A", omitting the comma is acceptable, but I find it clearer to use the comma in this longer sentence.

d0sboots: Thanks for that link, that's exactly the kind of style guide I was looking for! (AP doesn't have their full guide online.)

In this case we're dealing with prepositional phrases, which they say
> Common introductory phrases that should be followed by a comma include participial and infinitive phrases, absolute phrases, nonessential appositive phrases, and long prepositional phrases (over four words).

Here it's five words, so right on that line, but they're with you. Ofc additional wrinkles come in with weighting "too many commas" and then you start getting into "maybe there's a better way to word this entirely" - which is why I'm not an editor, I'm just playing one in the comments. XD

PR #2977, opened 2026-07-11
UI: Add reminder for backing up save data

catloversg:
We sometimes receive reports about corrupted save data. In some cases, players reported that it happens after events such as power cut or browser crashes. In other cases, they do not remember whether anything unusual happened. Anyway, these reports show us:
- Generally, browsers try to ensure data durability, but it may fail. Sometimes, browsers may prioritize other things such as performance over durability (https://developer.chrome.com/blog/indexeddb-durability-mode-now-defaults-to-relaxed).
- Many players underestimate the importance of data backup.

Currently, we give players a small benefit when they export save data every day, but this is not very effective at encouraging them to back up their data. I also don't want to grant more in-game rewards. Instead of trying to encourage players by increasing the rewards, which I think is a bad idea, we should remind them if they have not backed up their data for a long time. If they still don't do that even after we remind them, that's on them.

This PR adds a reminder that shows a warning toast if players have not backed up their data for over 24 hours. The toast appears every hour and lasts 30 seconds. This should be noticeable without being too intrusive. There is a setting to disable the reminder. Disabling it requires confirmation.

I also removed unused counters and renamed the `mechanicProcess` counter to `bladeburnerProcess`.

### DIFF
--- a/src/GameOptions/ui/SystemPage.tsx
+++ b/src/GameOptions/ui/SystemPage.tsx
@@ -147,6 +147,19 @@ export const SystemPage = (): React.ReactElement => {
         tooltip={<>If this is set, there will be no warning triggered when auto-save is disabled (at 0).</>}
       />
       <OptionSwitch
+        checked={Settings.DisableSaveDataBackupReminder}
+        onChange={(newValue) => {
+          Settings.DisableSaveDataBackupReminder = newValue;
+        }}
+        promptOptions={{
+          // Only require confirmation if the player is disabling the reminder.
+          shouldShowPrompt: (switchNewValue) => switchNewValue,
+          text: "Are you sure you want to disable the reminder?",
+        }}
+        text="Disable save data backup reminder"
+        tooltip={<>If enabled, we will not remind you to back up your save data.</>}
+      />
+      <OptionSwitch
         checked={Settings.SaveGameOnFileSave}
         onChange={(newValue) => (Settings.SaveGameOnFileSave = newValue)}
         text="Save game on file save"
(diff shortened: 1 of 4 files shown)


E9cipher:
In my opinion, a toast can be easily missed or ignored (in case of autopilot scripts that use toasts to notify events). What about using an alert (a yes/no prompt)? Hitting "yes" could redirect the user to the Augmentations tab where they can backup their save as normally
It may be intrusive at first, but the player can disable it on settings so that shouldn't be a problem.

catloversg:
My goal is to reduce the chance they disable the reminder as much as possible. A disabled reminder is a useless one, so I want it to be as unobtrusive as possible. People tend to disable anything they find annoying.

E9cipher:
Yeah, but there's not too much point in alerting the player with something that can be missed easily. I understand that your goal is to be as unobtrusive as possible, but sometimes you have to be a little intrusive in order for things to actually reach the player. Plus, if they disable the option in settings because it's annoying, they're on their own, as you said.

catloversg:
That toast appears every hour and lasts 30 seconds. It's really hard to miss it.

d0sboots:
I think this hits the right balance of "obvious but not so annoying it gets disabled."

src/GameOptions/ui/SystemPage.tsx
```diff
@@ -146,6 +146,19 @@ export const SystemPage = (): React.ReactElement => {
         text="Suppress Auto-Save Disabled Warning"
         tooltip={<>If this is set, there will be no warning triggered when auto-save is disabled (at 0).</>}
       />
+      <OptionSwitch
+        checked={Settings.DisableSaveDataBackupReminder}
```
d0sboots: I think it would be better (both in code and player settings comprehension) to word this in the positive, i.e. `SaveDataBackupReminder` or `EnableSaveDataBackupReminder`.) The downside is that it will require a bit more custom code to set it to true if not present. Feel free to disagree if you think that downside makes it not worth the clarity upside.

PR #2980, opened 2026-07-12
IPVGO: Add the inital board state to ns.go.getMoveHistory() on new game

Core-27:
Add the initial board state to previousBoards (which feeds `ns.go.getMoveHistory()` ) on resetBoardState to improve the discoverability of entirely removing `ns.go.getGameState()` from the player's script to save on RAM.

It is possible already using the return value of `ns.go.resetBoardState()` to get the dead nodes of the new board, but it took me making this PR to realize this ¯\\_(ツ)_/¯

Documentation: 
getmovehistory.md does not mention that it currently returns an empty array on new game state, and this PR will prevent it, so no change needed.

### DIFF
--- a/src/Go/effects/netscriptGoImplementation.ts
+++ b/src/Go/effects/netscriptGoImplementation.ts
@@ -320,7 +320,11 @@ export function getGameState() {
 }
 
 export function getMoveHistory(): SimpleBoard[] {
-  return Go.currentGame.previousBoards.map((boardString) => simpleBoardFromBoardString(boardString));
+  if (Go.currentGame.previousBoards.length > 0) {
+    return Go.currentGame.previousBoards.map((boardString) => simpleBoardFromBoardString(boardString));
+  } else {
+    return [simpleBoardFromBoard(Go.currentGame.board)];
+  }
 }
 
 /**


d0sboots:
This doesn't seem correct. Once you make a move, you will have the first board in the previousBoards twice.

Core-27:
Thanks I missed that. It also fails to add the initial state to the history when a new game is started by manually clicking in the IPvGo tab. I'll give it another attempt after a good night sleep.

d0sboots:
Having touched code in this general area before, I'll say that I think this is probably a bad idea, but I won't veto it if you find a way to make it work.

Core-27:
I think I got it. The condition to skip adding a duplicate of the initial board to `previousBoards` feels unelegant, but it behaves correctly on scripted and manual new game, no AI opponent and cheated first moves, as far as I can tell.

I'm not sure if this is worth merging, but I'm happy for the learning experience.

ficocelliguy:
I think it would be easier and cleaner to change getMoveHistory to detect an empty history and return the current state, rather than trying to actually edit the history.

ficocelliguy (approved):

d0sboots:
When you make your first move, getMoveHistory will contain the initial board state. If your first move was a pass, the current board state will be equal to the previous board state. Thus, what you have done makes it impossible to distinguish (using getMoveHistory() and getBoardState()) between the start of the game and an initial pass. 

I think this will definitely break various scripts that currently use getMoveHistory() and rely on its length equaling the number of moves in the game.

PR #2981, opened 2026-07-12
API: Add optional hostname argument to ns.getFileMetadata()

### LINKED ISSUE #2941: Add optional hostname argument to ns.getFileMetadata()
Currently, `ns.getFileMetadata()` only accepts one input - the name of the file (on the current server). It cannot see the details of files from other servers. You cannot scp the file locally to see it either, because scp changes the file metadata. 

We should add an optional hostname argument as a second function parameter. The function documentation will also need to be updated accordingly.

Code pointer:
https://github.com/bitburner-official/bitburner-src/blob/dev/src/NetscriptFunctions.ts#L1136

Discussion / context on Discord:
https://discord.com/channels/415207508303544321/415213435974975508/1518403369595703327

Core-27:
I'm giving a go at #2941, I basically copy pasted from scp() while making sure I understood the code. Testing in game gives the expected results or errors, `npm run test` is clean.

If I'm not mistaken there is no safeguards on most file actions as long as the target server is connected to the network. No auth or root requirement, except to run `ls`.  `scp` only check the destination server. So there is no extra safeguard on this implementation either.

closes #2941

### DIFF
--- a/src/ScriptEditor/NetscriptDefinitions.d.ts
+++ b/src/ScriptEditor/NetscriptDefinitions.d.ts
@@ -8656,6 +8656,7 @@ export interface NS {
    * RAM cost: 0 GB
    *
    * This function returns the metadata associated with the specified file.
+   * If the file does not exist or the server is offline, returns null. It will throw if the path or host is malformed.
    *
    * @param filename - Name of the file to read the metadata from. It must be a text file (.txt, .json, .css) or a script
    * (.js, .jsx, .ts, .tsx).
--- a/src/ScriptEditor/NetscriptDefinitions.d.ts
+++ b/src/ScriptEditor/NetscriptDefinitions.d.ts
@@ -8580,7 +8580,8 @@ export interface NS {
    *
    * @param ram - The new RAM limit to set.
    * @returns The new static RAM limit, which will be the old one if it wasn't changed.
-   * This means you can use no parameters to check the current ram limit.
+   * This means you can use no
+   * parameters to check the current ram limit.
    */
   ramOverride(ram?: number): number;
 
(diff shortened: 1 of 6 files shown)


d0sboots:
@catloversg
I think we made a mistake when we first implemented this function in #2199.

You actually specifically called it out, but looking at it now I don't agree with your reasoning:
> Make `ns.getFileMetadata` throw an error when the filename is invalid instead of returning undefined. It seems that Hoekstraa reused the "style" of `ns.read`, but I think the behavior of `ns.read` is bad. We have been too lenient when it comes to param checking. Returning an empty string when the param is invalid just hides the programming error of the player. It may be convenient for some people, especially newbies who hate seeing errors, but those errors will eventually come back and wreck havoc in other parts of their scripts. Changing that behavior is a big breaking change, so I won't do anything now, but I don't want to see that behavior in new APIs.

I would agree if specifying an incorrect path were a programmer error (exclusively), but IMO it can also be a data error. The distinction comes from if the error comes from code or the state of the system: With hostnames, the error can only come from code (if we exclude dnet), because they are static and thus an invalid hostname means you spelled it wrong. However, a missing file can come about because the file got deleted in-between when it was entered and when you tried to read it/look up its metadata. In other words, it can happen because *the state of the system changed*, not because you specified it wrong in your code.

Our standard (as much as we have one) is to throw on the first type of error, but return an invalid state for the second type.

The reason I bring this up is because it is becoming much more urgently relevant with this change. *Because* we are currently throwing on all errors, this PR implemented "offline dnet server" as a throw. However, it would be the *only* NS function to do so - every other NS function returns an invalid value in this case, and IMO it's wrong to throw here, for reasons we discovered during dnet testing. But if we return undefined here for an offline server, it seems really nonsensical to throw on an invalid path.

catloversg:
I'm fine with returning null if the host is offline or the file does not exist.

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -8659,9 +8659,10 @@ export interface NS {
    *
```
d0sboots, line 8660:
```suggestion
   * If the file does not exist or the server is offline, returns null. It will throw if the path or host is malformed.
   *
```
You get stuck with adding docs since this happens on your watch.

PR #2985, opened 2026-07-13
DEVMENU: Disable accordions instead of not displaying them

E9cipher:
This PR disables the `AutoExpandAccordion` in the devmenu instead of not displaying them. I also remove redundant checks (that the new code no longer needs).

The stock market change is different than the others because it does not check for the feature access in the file itself like the others do. Instead, it is passed a parameter `disabled` and the check occurs in the `DevMenu.tsx` file.

Linter and jest come back clean, I also tested it manually and I encountered no errors.


catloversg:
I disagree with this change. Why bother rendering things that you cannot use? In some components, you are just moving the null check.

E9cipher:
Imo, disabled is better than not rendering it. Same thing happened with the darknet section. The only thing I would say is missing is a tooltip explaining that it should be enabled on the general tab, but I see it as an optional change (as you see).

d0sboots:
I agree with the spirit of this change - that it is (usually) better to disable things than remove them. A disabled UI element tells you that it will become available in the future, or under different circumstances. If it is missing, then you might think the feature does not exist.

Sometimes, this is what you want - for instance, many incremental games use this intentionally to reveal mechanics over time. However, for a dev menu I think it makes sense to know all the tools that are available, even if some of them are unusable due to the current state.

I merged a change from CatLover that means that the *implementation* here would need to be different.

E9cipher:
Thanks. If Catlover agrees, I will deal with the conflicts  (change the impl)

catloversg:
If d0sboots think it's fine to disable these tools, then I'm okay with it.

Note that you need to be consistent.
```ts
<BladeburnerDev bladeburner={Player.bladeburner ?? null} />
<GangDev />
<StockMarketDev disabled={canAccessStockMarket()} />
```
- `Player.bladeburner ?? null` is redundant because `Player.bladeburner` is already nullable.
- `BladeburnerDev` receives `Player.bladeburner` as a prop, whereas `GangDev` accesses `Player.gang` directly.
- If you want the components to perform these checks themselves, then do so consistently. For example, don't compute `canAccessStockMarket()` here and pass the result to `StockMarketDev`.

E9cipher:
Sorry for taking so long. This first commit makes the logic, I'll deal with merge conflicts on the second one (see if I can finish it by today)

E9cipher:
Hmmmm nice, actually github UI is much better than I thought at resolving conflicts : )

Edit: never mind

E9cipher:
@d0sboots @catloversg does this seem correct to you?

src/DevMenu/ui/BladeburnerDev.tsx
```diff
@@ -14,15 +14,34 @@ import Select, { SelectChangeEvent } from "@mui/material/Select";
...
-export function BladeburnerDev({ bladeburner }: { bladeburner: Bladeburner }): React.ReactElement {
+export function BladeburnerDev(): React.ReactElement {
+  const bladeburner = Player.bladeburner;
+  const [skillName, setSkillName] = useState(BladeburnerSkillName.BladesIntuition);
+  const [contractTarget, setContract] = useState<BladeburnerContractName>(
+    bladeburner?.contracts.Tracking.name ?? BladeburnerContractName.Tracking,
+  );
+  const [operationTarget, setOperation] = useState<BladeburnerOperationName>(
+    bladeburner?.operations.Investigation.name ?? BladeburnerOperationName.Investigation,
+  );
```
catloversg:
```suggestion
  const [contractTarget, setContractTarget] = useState(BladeburnerContractName.Tracking);
  const [operationTarget, setOperationTarget] = useState(BladeburnerOperationName.Investigation);
```
`Player.bladeburner.contracts.Tracking.name` is `BladeburnerContractName.Tracking`. Same with Operation.

Remember to change callers of `setContract` and `setOperation`.

src/DevMenu/ui/SleevesDev.tsx
```diff
@@ -11,6 +11,15 @@ import { Adjuster } from "./Adjuster";
...
 
 export function SleevesDev(): React.ReactElement {
+  if (!(Player.sleeves.length > 0))
+    return (
+      <AutoExpandAccordion cacheKey="DEVMENU_SleevesDev" unmountOnExit={true} disabled={true}>
+        <AccordionSummary expandIcon={<ExpandMoreIcon />}>
+          <Typography>Sleeves</Typography>
+        </AccordionSummary>
+      </AutoExpandAccordion>
+    );
```
catloversg:
```suggestion
  if (Player.sleeves.length === 0) {
    return (
      <AutoExpandAccordion cacheKey="DEVMENU_SleevesDev" unmountOnExit={true} disabled={true}>
        <AccordionSummary expandIcon={<ExpandMoreIcon />}>
          <Typography>Sleeves</Typography>
        </AccordionSummary>
      </AutoExpandAccordion>
    );
  }
```

src/DevMenu/ui/StanekDev.tsx
```diff
@@ -10,8 +10,17 @@ import ExpandMoreIcon from "@mui/icons-material/ExpandMore";
...
 
 export function StanekDev(): React.ReactElement {
+  if (!Player.augmentations.some((aug) => aug.name === AugmentationName.StaneksGift1))
+    return (
+      <AutoExpandAccordion cacheKey="DEVMENU_StanekDev" unmountOnExit={true} disabled={true}>
+        <AccordionSummary expandIcon={<ExpandMoreIcon />}>
+          <Typography>Stanek's Gift</Typography>
+        </AccordionSummary>
+      </AutoExpandAccordion>
+    );
```
catloversg:
```suggestion
  if (!Player.augmentations.some((aug) => aug.name === AugmentationName.StaneksGift1)) {
    return (
      <AutoExpandAccordion cacheKey="DEVMENU_StanekDev" unmountOnExit={true} disabled={true}>
        <AccordionSummary expandIcon={<ExpandMoreIcon />}>
          <Typography>Stanek's Gift</Typography>
        </AccordionSummary>
      </AutoExpandAccordion>
    );
  }
```

src/DevMenu/ui/BladeburnerDev.tsx
```diff
@@ -14,15 +14,34 @@ import Select, { SelectChangeEvent } from "@mui/material/Select";
...
 import { Skills as AllSkills } from "../../Bladeburner/data/Skills";
-import { Bladeburner } from "../../Bladeburner/Bladeburner";
 import { getEnumHelper } from "../../utils/EnumHelper";
 import { AutoExpandAccordion } from "../../ui/AutoExpand/AutoExpandAccordion";
 
 const bigNumber = 1e27;
 
-export function BladeburnerDev({ bladeburner }: { bladeburner: Bladeburner }): React.ReactElement {
+export function BladeburnerDev(): React.ReactElement {
+  const bladeburner = Player.bladeburner;
```
catloversg: Move this line to after all useState calls. You don't need `bladeburner` for them.

src/DevMenu/ui/BladeburnerDev.tsx
```diff
@@ -14,15 +14,34 @@ import Select, { SelectChangeEvent } from "@mui/material/Select";
...
+  const [skillName, setSkillName] = useState(BladeburnerSkillName.BladesIntuition);
+  const [contractTarget, setContract] = useState<BladeburnerContractName>(
+    bladeburner?.contracts.Tracking.name ?? BladeburnerContractName.Tracking,
+  );
+  const [operationTarget, setOperation] = useState<BladeburnerOperationName>(
+    bladeburner?.operations.Investigation.name ?? BladeburnerOperationName.Investigation,
+  );
+
+  if (!bladeburner) {
```
catloversg: Optional:
```ts
if (!Player.bladeburner) {
  return ...
}
const bladeburner = Player.bladeburner;
```
You already use this pattern in other files.

PR #2986, opened 2026-07-13
API: Rename ns.singularity.exportGameBonus()

### LINKED ISSUE #2984: Rename ns.singularity.exportGameBonus()
exportGameBonus() is a getter function that returns a boolean. It breaks the naming convention most other `ns` APIs use: its name should either start with "get" or with "is".

Naming things is widely recognized as one of the two hard things in computer science ;). Therefore I have opened this tread for people to discuss possible better names for this API.

If we decide to rename exportGameBonus(), the old name will be deprecated, but will still continue to work (for now) to avoid breaking player code. The old name may be a candidate to be removed in v4.

catloversg:
Closes #2984.

Test code:
```js
/** @param {NS} ns */
  ns.tprint(ns.singularity.exportGameBonus());
  ns.tprint(ns.singularity.hasExportGameBonus());
}
```

Capture


hexagonrecursion (approved):

d0sboots:
Open question before I merge this: Since the change is mechanical and straightforward, does it make more sense to use the upgrade machinery to do an auto-migration now, even though it's not a major version? That has the benefit that it will potentially save a lot of users effort, although it may be unexpected. Or, we can defer that until we remove the deprecated function.

catloversg:
I'm neutral on that.

d0sboots:
In that case, although it's slightly weird, I feel like the most user-friendly thing to do is to auto-migrate now, and then also again when we deprecate.

d0sboots:
Thanks. I suppose it's obvious, but I only asked for that because of your familiarity with the upgrade machinery - otherwise it would be a lot for such a small change.

catloversg:
Should we perform auto-migration for future deprecations? I have some on my list. If we decide to make this a standard procedure, I'll do that in future PRs.

d0sboots:
If they're straightforward auto-migrations, I think it makes sense to for the same reasons given here.

Also, I realized I didn't catch this before I merged, but the next release will be 3.1.0. It's not a big deal, because there will be other places to change too.

PR #2987, opened 2026-07-13
UI: Improve accessibility of wire cutter minigame

catloversg:
#2950 addressed the remaining issues with the light theme, except for the wire cutter minigame. A while ago, a player reported having trouble distinguishing red and yellow wires. They suggested adding the ability to change the wire colors. I considered adding more theme colors and using them for the wires. However, I now think we should change the color palette instead.

Allowing players to change the colors is a nice idea, but it seems like the XY problem. What players actually want is to distinguish the wire colors more easily. Adding new theme colors and telling players to adjust them on their own is unnecessary. Instead, I think we should choose a good palette that helps colorblind players while also solving the issue with the light theme.

This PR uses colors from the Okabe-Ito color palette (https://jfly.uni-koeln.de/color/), which is well known for addressing this kind of issue. It also forces a black background to improve contrast and ensure that unusual theme backgrounds do not introduce new color related issues.

Default theme:
Before:
before-default-theme

After:
after-default-theme

Iron Man theme:
Before:
before-red

After:
after-red

Light theme:
Before:
before-light

After:
after-light

Have the `SoA - Knowledge of Apollo` augmentation:
Capture

### DIFF
--- a/src/Infiltration/model/WireCuttingModel.ts
+++ b/src/Infiltration/model/WireCuttingModel.ts
@@ -18,13 +18,13 @@ const difficultySettings = {
   Brutal: { timer: 4000, wiresmin: 9, wiresmax: 9, rules: 4 },
 };
 
-const colors = ["red", "#FFC107", "blue", "white"] as const;
+const colors = ["#D55E00", "#F0E442", "#0072B2", "#FFFFFF"] as const;
 
 const colorNames = {
-  red: "RED",
-  "#FFC107": "YELLOW",
-  blue: "BLUE",
-  white: "WHITE",
+  "#D55E00": "RED",
+  "#F0E442": "YELLOW",
+  "#0072B2": "BLUE",
+  "#FFFFFF": "WHITE",
 } as const;
 
 interface Wire {
--- a/src/Infiltration/model/WireCuttingModel.ts
+++ b/src/Infiltration/model/WireCuttingModel.ts
@@ -18,10 +18,13 @@ const difficultySettings = {
   Brutal: { timer: 4000, wiresmin: 9, wiresmax: 9, rules: 4 },
 };
 
+// These colors, except white, come from the Okabe-Ito palette. White is used because it provides high contrast on a
+// black background.
+// Reference: https://jfly.uni-koeln.de/color/
 const colors = ["#D55E00", "#F0E442", "#0072B2", "#FFFFFF"] as const;
 
 const colorNames = {
-  "#D55E00": "RED",
+  "#D55E00": "RED", // Okabe-Ito Vermilion
   "#F0E442": "YELLOW",
   "#0072B2": "BLUE",
   "#FFFFFF": "WHITE",
--- a/src/Infiltration/model/WireCuttingModel.tsx
+++ b/src/Infiltration/model/WireCuttingModel.tsx
@@ -1,3 +1,4 @@
+import React from "react";
 import type { InfiltrationStage, KeyboardLikeEvent } from "../InfiltrationStage";
 import type { Infiltration } from "../Infiltration";
 import { interpolate } from "./Difficulty";
@@ -36,14 +37,14 @@ interface Wire {
 }
 
 interface Question {
-  toString: () => string;
+  render: () => React.ReactNode;
   shouldCut: (wire: Wire, index: number) => boolean;
 }
 
 function randomPositionQuestion(wires: Wire[]): Question {
   const index = Math.floor(Math.random() * wires.length);
   return {
-    toString: (): string => {
+    render: () => {
       return `Cut wire number ${index + 1}.`;
     },
     shouldCut: (_wire: Wire, i: number): boolean => {
@@ -56,8 +57,12 @@ function randomColorQuestion(wires: Wire[]): Question {
   const index = Math.floor(Math.random() * wires.length);
   const cutColor = wires[index].colors[0];
   return {
-    toString: (): string => {
-      return `Cut all wires colored ${colorNames[cutColor]}.`;
+    render: () => {
+      return (
+        <>
+          Cut all wires colored <span style={{ color: cutColor }}>{colorNames[cutColor]}</span>.
+        </>
+      );
     },
     shouldCut: (wire: Wire): boolean => {
       return wire.colors.includes(cutColor);
--- a/src/Infiltration/ui/WireCuttingGame.tsx
+++ b/src/Infiltration/ui/WireCuttingGame.tsx
@@ -4,7 +4,6 @@ import type { WireCuttingModel } from "../model/WireCuttingModel";
 import { Box, Paper, Typography } from "@mui/material";
 import { AugmentationName } from "@enums";
 import { Player } from "@player";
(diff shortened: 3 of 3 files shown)


src/Infiltration/model/WireCuttingModel.ts
```diff
@@ -18,13 +18,13 @@ const difficultySettings = {
...
 
-const colors = ["red", "#FFC107", "blue", "white"] as const;
+const colors = ["#D55E00", "#F0E442", "#0072B2", "#FFFFFF"] as const;
 
 const colorNames = {
-  red: "RED",
-  "#FFC107": "YELLOW",
-  blue: "BLUE",
-  white: "WHITE",
+  "#D55E00": "RED",
```
d0sboots: This is not cleanly red anymore. In addition, it's suggested to use the palette in order to maximize contrast. So, I'd suggest replacing this with Orange and also changing the label.

src/Infiltration/model/WireCuttingModel.tsx
```diff
@@ -18,13 +18,13 @@ const difficultySettings = {
   Brutal: { timer: 4000, wiresmin: 9, wiresmax: 9, rules: 4 },
 };
 
-const colors = ["red", "#FFC107", "blue", "white"] as const;
+const colors = ["#D55E00", "#F0E442", "#0072B2", "#FFFFFF"] as const;
```
d0sboots, line 25:
```suggestion
// These colors (aside from White, which is assumed since it was designed for print) come from
// the Okabe-Ito palette, which is considered the gold standard for accessible colors for categorical data.
// See more at https://vizcept.com/blog/okabe-ito-palette-guide
const colors = ["#D55E00", "#F0E442", "#0072B2", "#FFFFFF"] as const;
```
Reword or replace the link as you like.

src/Infiltration/ui/WireCuttingGame.tsx
```diff
@@ -13,6 +12,7 @@ interface IProps {
 
 export function WireCuttingGame({ stage }: IProps): React.ReactElement {
   const hasAugment = Player.hasAugmentation(AugmentationName.KnowledgeOfApollo, true);
+  const wrongWireColor = "#CC79A7";
```
d0sboots, line 16:
```suggestion
  // Okabe-Ito "Reddish Purple"
  const wrongWireColor = "#CC79A7";
```

src/Infiltration/ui/WireCuttingGame.tsx
```diff
@@ -28,11 +28,13 @@ export function WireCuttingGame({ stage }: IProps): React.ReactElement {
...
             columnGap: 3,
             justifyItems: "center",
+            background: "black",
+            padding: "10px 20px",
           }}
         >
           {Array.from({ length: stage.wires.length }, (_, i) => {
             const isCorrectWire = stage.cutWires[i] || stage.wiresToCut.has(i);
-            const color = hasAugment && !isCorrectWire ? Settings.theme.disabled : Settings.theme.primary;
+            const color = hasAugment && !isCorrectWire ? wrongWireColor : "#009E73";
```
d0sboots:
```suggestion
            const color = hasAugment && !isCorrectWire ? wrongWireColor : "#009E73"; // Okabe-Ito "Bluish Green"
```
Maybe consider "Sky Blue" instead, for better contrast vs. wrongWireColor?

d0sboots (commented):
Overall idea is great, solving the root of the problem instead of the symptoms.

catloversg:
I usually mark comments as resolved after merging, so let's discuss the colors here so the discussion is not hidden in the collapsed UI elements.

For reference (from https://jfly.uni-koeln.de/color/):

pallete

For the four wire colors, I chose Vermilion instead of Orange because Orange and Yellow are difficult to distinguish for people with tritanopia. In the Okabe and Ito simulation, Orange appears pink and Yellow appears light pink.

For the wrong wire color versus the correct wire color, I chose Bluish Green instead of Sky Blue because Sky Blue and Reddish Purple may be difficult to distinguish for people with protanopia, according to the simulated colors in the image above. Note that the simulated colors may differ depending on the simulation tool and how accurately the monitor reproduces colors. For example, when I tested Sky Blue on https://rgblind.com/color-blind-color-picker, it appeared more purple than in the Okabe and Ito paper under the protanopia simulation.

d0sboots:
Sky Blue vs Bluish Green: K, that's compelling.

For Orange vs Vermillion, it is more complicated. I agree that color-wise, Yellow/Blue/Vermillion has better contrast across all the groups than Yellow/Blue/Orange. However, an additional dimension (that I don't think Okabe-Ito tested, but we do have to consider) is the accessibility of associating names to colors - for *all* groups.

Honestly, I have no idea how well the labels we have will work for various types of color deficiency. I'm not sure if research has even been done on it. But, I do know that the Vermillion color is not that close to the center of what non-colorblind people consider "Red," creating a potential labeling issue there that I suspect may translate over to colorblind people. Given that there still *is* contrast even for tritanopes, and that tritan-type colorblindness is much rarer, my gut says that choosing a color that gives a clearer label might be better.

d0sboots:
Very optional: There's one other change that would improve accessibility (for everyone, but especially for colorblind folks): Extend the "black background/fixed scheme" section to include the cutting instructions, so that the color names in the instructions can also be colored with their color.

This would be a significant extension to the change, so only do it if you want to. However, if you did this I think it would be a lot less important for the color labels to line up, i.e. it would be fine to stick with Red/Vermillion.

catloversg:
> But, I do know that the Vermillion color is not that close to the center of what non-colorblind people consider "Red," creating a potential labeling issue there that I suspect may translate over to colorblind people.

Do you mean that people without color blindness could be confused when they see the label "RED" but the color is vermilion? I.e., when they see the vermilion color, they don't think it's reddish enough? When I see vermilion, I think it's reddish (with a bit of orange), so that's not confusing for me, but that's just my opinion.

Anyway, extending the background to the cutting instructions and setting the text color there too are good ideas. I'll do that.

d0sboots:
> Do you mean that people without color blindness could be confused when they see the label "RED" but the color is vermilion? I.e., when they see the vermilion color, they don't think it's reddish enough? When I see vermilion, I think it's reddish (with a bit of orange), so that's not confusing for me, but that's just my opinion.

Let me try to explain my potential worry with more words. 

When I see the current color, I think "that's reddish-orange" or "orangish-red", but I recognize that the label "RED" surely applies to it without any issues. 

However, from previous discussions with colorblind people I know that color labels can be challenging for them. (And a lot of times they don't even realize this.) Since they're learning to map labels to colors but the colors they experience have reduced dimensionality, the specific bits that separate different color labels for them (i.e. what makes "red" vs "green" to a red-green colorblind person) are not what we would necessarily expect. You might naively think that they would have uncertainty about some of the labels, but this is very much not the case! Instead, you're more likely to get into arguments about how "that isn't green." XD

Anyway, my vague concern was that this issue could be worse the father we are from the "center" of the label. I'm not nearly so concerned when we put the color on the text of the label.

PR #2989, opened 2026-07-14
UI: Fix description case for MONEY_1Q achievement

ndc:
To reduce the number of questions about not getting the achievement

### DIFF
--- a/src/Achievements/AchievementData.json
+++ b/src/Achievements/AchievementData.json
@@ -144,7 +144,7 @@
     "MONEY_1Q": {
       "ID": "MONEY_1Q",
       "Name": "Here comes the money!",
-      "Description": "Have $1Q on your home computer."
+      "Description": "Have $1Q (not $1q) on your home computer."
     },
     "MONEY_M1B": {
       "ID": "MONEY_M1B",


src/Achievements/AchievementData.json
```diff
@@ -144,7 +144,7 @@
     "MONEY_1Q": {
       "ID": "MONEY_1Q",
       "Name": "Here comes the money!",
-      "Description": "Have $1Q on your home computer."
+      "Description": "Have $1Q (not $1q) on your home computer."
```
d0sboots, line 147:
```suggestion
      "Description": "Have $1Q ($1e18) on your home computer."
```
This might be better? I am honestly unsure. It also hints that there's more to this value, and is more informative, but people that are struggling with this might have never encountered scientific notation.

ndc: Hmm I don't know if 10^18 is quadrillion or quintillion, and is it a small q or a big Q, so $1e18 wouldn't help me.

E9cipher: If it is to be read by humans, I would say $1q fits better

catloversg: I think using `1e18` is better. It's good to show the required amount of money in both scientific notation and suffix form. Some players prefer scientific notation, while others prefer suffix notation. If they somehow don't recognize that it's `1Q` instead of `1q`, `1e18` should also clarify it. If they don't know what `1e18` means, they can google it.

To be honest, if they can't tell the difference between `1Q` and `1q`, don't know what `1e18` means, and don't want to learn what it means, then they don't deserve this achievement.

E9cipher: Why not just put it in both? Plus, this PR's objective is to clarify it. Imo, using scientific notation does not clarify things at all...
Though I agree, if players can't tell the difference between $1Q and $1q they should learn to identify the units...

catloversg: > If it is to be read by humans, I would say $1q fits better

This implies the suffix form is easier for humans to read, which is not true. That purely depends on what you're familiar with. If the suffix form were superior, then no incremental game would add the option to display numbers using other notations.

E9cipher: I think most players will prefer to read it in suffix form. However, this is not all players so maybe both forms is better (or based on settings)?

d0sboots: Hehe I am not surprised by the conflicting opinions. However, I am merging it as-is because the responses I'm getting confirmed my suspicions that adding scientific notation would be more helpful to people who already don't need the help.

PR #2991, opened 2026-07-15
UI: Accounting for ellipsis length when shortening hostnames

aebabis:
# Accounting for ellipsis length when shorting hostnames in the Active Scripts panel

The `truncatedHostname` in the Active Scripts panel are 3 characters too long, causing the corresponding RAM meter to be misaligned

before

# Bug fix

Subtracting 3 from the length to account for the length of the ellipsis fixes this

after

Code tested locally using the same save in which I observed the bug

### DIFF
--- a/src/ui/ActiveScripts/ServerAccordion.tsx
+++ b/src/ui/ActiveScripts/ServerAccordion.tsx
@@ -22,9 +22,10 @@ export function ServerAccordion({ server, scripts, startOpen }: ServerAccordionP
   // Accordion's header text
   const longestHostnameLength = 26;
   // Use spread operator to get accurate length on servers with UTF-16 names
-  const paddedName = [...server.hostname].length > longestHostnameLength
-      ? [...server.hostname].slice(0, longestHostnameLength - 3).join('') + "..."
-      : [...`${server.hostname}${" ".repeat(longestHostnameLength)}`].slice(0, longestHostnameLength).join('');
+  const hostnameChars = [...server.hostname];
+  const paddedName = hostnameChars.length > longestHostnameLength
+      ? hostnameChars.slice(0, longestHostnameLength - 3).join('') + "..."
+      : server.hostname + " ".repeat(longestHostnameLength - hostnameChars.length);
   const barOptions = {
     progress: server.ramUsed / server.maxRam,
     totalTicks: 30,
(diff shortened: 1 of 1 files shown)


src/ui/ActiveScripts/ServerAccordion.tsx
d0sboots: This got changed during the Darknet PR, which was huge, so I missed it. It's actually still got a (smaller) bug in it (it switches to ellipsis too soon). Let's switch to a simpler impl which doesn't hide bugs so easily (2nd try on this comment, with the correct range of lines)
```suggestion
const paddedName = server.hostname.length > longestHostnameLength
      ? server.hostname.slice(0, longestHostnameLength - 3) + "..."
      : `${server.hostname}${" ".repeat(longestHostnameLength)}`.slice(0, longestHostnameLength);
```

hexagonrecursion: This uses .length and .slice(). Both work in terms of utf-16 code units. You will have similar issues to the ones you solved in safelyReverseString():
1. You will calculate length incorrectly. For some strings hostname.length is bigger than the number of graphemes in the string
2. hostname.slice() will happily cut a string in the middle of a surrogate pair

aebabis: Good call. I've fixed it using spreading. I also simplified it to avoid a slice, but I can revert the last commit if we prefer the old style:

https://github.com/bitburner-official/bitburner-src/pull/2991/changes/b96fe10da4e60e1f2d43d4f347a45fabfeb88b54
https://github.com/bitburner-official/bitburner-src/pull/2991/changes/c5474bcd11ccd89a62845aba4a2b17979ec4a46a

aebabis:
Here's a screenshot showing that it now supports emoji too

image

Disclosure: It still doesn't support non-monospace characters.

image

but I don't think there's a solution without writing some dynamic sizing JS+CSS.

aebabis:
Just to be sure, I tried this:
```typescript
  // Use spread operator to get accurate length on servers with UTF-16 names
  const hostnameChars = [...server.hostname];
  const cjkCount = [...server.hostname.matchAll(/[\u4E00-\u9FFF\u3400-\u4DBF]/g)].length;
  const width = hostnameChars.length + cjkCount;
  const paddedName = width > longestHostnameLength
      ? hostnameChars.slice(0, longestHostnameLength + cjkCount - 3).join('') + "..."
      : server.hostname + " ".repeat(longestHostnameLength - width);
```
 
It looks a little better:
image

but it proves that we'd need to either wrap CJK characters in a different-font span or mess with the width or kerning to make it work

d0sboots:
Good catch on the UTF-16 issues.

I'm OK with accepting this, after npm format. If you're up for it, you can look into using CSS `text-overflow: ellipsis`, which is really the correct solution and how we deal with this elsewhere.

PR #2992, opened 2026-07-17
REFACTOR: Change Aug impl for greater sanity

d0sboots:
The augmentation code was a bit of a mess, with NFG tacked on as a special-case all over the place. This reworks that in preparation for the "socks" PR, so that it doesn't need to add a pile more special-cases.

The weirdness where augs are unique when installed, but can have duplicates in the queued list is now handled in a uniform way. I thought about changing that, but we'd need to keep supporting duplicates in the queued list ~forever anyway, because of old saves. But now there's a *lot* more comments for the weird sections.

This also cleans up the way multipliers are calculated, as augs are the primary user of that code.

# Testing

I tested augs manually using the dev menu. I should check if we have automated tests, but got lazy and wanted to get this out there.

@ficocelliguy you can see the plan now. Also you can poke me if I forget about this again.

### DIFF
--- a/src/Augmentation/AugmentationHelpers.ts
+++ b/src/Augmentation/AugmentationHelpers.ts
@@ -5,6 +5,7 @@ import { AugmentationName } from "@enums";
 
 import { CONSTANTS } from "../Constants";
 import { Player } from "@player";
+import type { Multipliers } from "@nsdefs";
 import { prestigeAugmentation } from "../Prestige";
 
 import { dialogBoxCreate } from "../ui/React/DialogBox";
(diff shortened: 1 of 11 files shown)


src/PersonObjects/Multipliers.ts
```diff
@@ -68,74 +36,29 @@ export const defaultMultipliers = (): Multipliers => {
...
-    work_money: (m0.work_money - 1) * v + 1,
-    crime_success: (m0.crime_success - 1) * v + 1,
-    crime_money: (m0.crime_money - 1) * v + 1,
-    dnet_money: (m0.dnet_money - 1) * v + 1,
-    bladeburner_max_stamina: (m0.bladeburner_max_stamina - 1) * v + 1,
-    bladeburner_stamina_gain: (m0.bladeburner_stamina_gain - 1) * v + 1,
-    bladeburner_analysis: (m0.bladeburner_analysis - 1) * v + 1,
-    bladeburner_success_chance: (m0.bladeburner_success_chance - 1) * v + 1,
-  };
+export const scaleMultipliers = (result: Multipliers, v: number): void => {
```
ficocelliguy, line 47: "scaleAllMultipliers" might be a better name, especially as it's a mutation now

ficocelliguy (approved):

d0sboots:
TODO: I need to test Stanek, since this introduces a small change to those numbers

gmcew:
Was there anything specifically you wanted to check with Stanek? Install and charging in this PR seem to work the same as stable (including the fact that stats don't update on removal of pieces)

d0sboots:
@gmcew The stats for sleeves should be marginally different in the case of multipliers on effects where "less is more". I also needed to check that I didn't break that whole system.

src/Augmentation/AugmentationHelpers.ts
```diff
@@ -36,33 +37,55 @@ export function getGenericAugmentationPriceMultiplier(): number {
...
-  if (aug.name === AugmentationName.NeuroFluxGovernor && !reapply && ownedNfg) {
-    ownedNfg.level = aug.level;
-    return;
+// Update the multipliers in "mults" from previousLevel to aug.level. This should be used in most cases
+// instead of mergeMultipliers, since it handles special cases like NFG uniformly. (It works for both
+// queued and installed augs.)
+export function updateMultipliers(mults: Multipliers, aug: PlayerOwnedAugmentation, previousLevel: number): void {
+  const staticAugmentation = Augmentations[aug.name];
+  if (aug.level <= previousLevel) {
+    throw new Error(`Trying to downlevel/relevel aug {aug.name} from {previousLevel} to {aug.level}!`);
```
catloversg, line 82:
```suggestion
    throw new Error(`Trying to downlevel/relevel aug ${aug.name} from ${previousLevel} to ${aug.level}!`);
```

src/Augmentation/AugmentationHelpers.ts
```diff
@@ -36,33 +37,55 @@ export function getGenericAugmentationPriceMultiplier(): number {
...
   }
-
-  // Push onto Player's Augmentation list
-  if (!reapply) {
-    const ownedAug = new PlayerOwnedAugmentation(aug.name);
-
-    Player.augmentations.push(ownedAug);
+  if (aug.name !== AugmentationName.NeuroFluxGovernor && (aug.level !== 1 || previousLevel !== 0)) {
+    throw new Error(`Unexpected levels for {aug.name}: Leveling {previousLevel} to {aug.level}!`);
```
catloversg, line 85:
```suggestion
    throw new Error(`Unexpected levels for ${aug.name}: Leveling ${previousLevel} to ${aug.level}!`);
```

src/PersonObjects/Multipliers.ts
```diff
@@ -68,74 +36,29 @@ export const defaultMultipliers = (): Multipliers => {
...
+    //
+    // However, inverse mults apply in a reciprocal fashion, so we have to
+    // take that into effect here. For instance, scaling a 0.5x mult (/2) by
+    // 1.5 gives a 0.4x mult (/2.5). This has the same effect on stats, but a
+    // more complicated formula. We assume any mult that is less than 1 is
+    // applying this way.
+    if (effect >= 1) {
+      result[key] = (effect - 1) * v + 1;
+    } else {
+      result[key] = 1 / ((1 / effect - 1) * v + 1);
```
catloversg, line 61: This is a behavior change.

This function is currently used only when scaling mults of Sleeve augs when Sleeves have ZOE installed. Let's say Sleeve 0 has an aug that gives foo_mult = 0.8, and there are 4 Sleeves that have installed ZOE. We have mult = 0.8, and we want that mult to be weakened by 50% (`3 / (zoeAmt + 2)`). There are two ways to do that:

Linear scaling (old code):
0.8 = −20% effect
↓ 50% weaker
0.9  = −10% effect

This treats the mult as a percentage modifier around the neutral point 1. For example, 1.2 means +20% and 0.8 means −20%.

Reciprocal scaling (new code):
0.8 = 1/1.25
↓ 50% weaker
0.8889 = 1/1.125

This treats the mult as an inverse factor, and scales the reciprocal of the mult instead.

Switching to reciprocal scaling makes the beneficial effect stronger for mults below 1. For example, let's say 0.8 is the hacknet purchase cost mult. With linear scaling, it becomes 0.9, while with reciprocal scaling it becomes 0.8889. Since 0.8889 is a lower cost mult, the reciprocal scaling produces a larger discount.

I'm neutral on switching to reciprocal scaling. On one hand, I don't see anything wrong with using linear scaling. On the hand, I also don't see anything wrong with using reciprocal scaling. Note that, even though this is a behavior change, it actually does not change anything in practice. The only mults that are less than 1 are ones that affect hacknet costs, but sleeves cannot do anything with hacknet. This is why I'm okay with either scaling.

Anyway, I think you should mention this behavior change in the PR description.

src/Augmentation/AugmentationHelpers.ts
d0sboots: Done.

src/Augmentation/AugmentationHelpers.ts
d0sboots: Done

src/PersonObjects/Multipliers.ts
d0sboots: Yes, I made this change as a cleanup/bugfix due to how hacknet and related augs mults work.

Most multipliers are positive, and we apply linearly scaled modifiers to them when such circumstances arise. For instance, something labeled as +40% to hacking will apply a (1 + 0.4)x hacking mult. We used to have bugs where we incorrectly applied linear scaling for decreasing multipliers, such that a 40% boost to hacknet would be applied as (1 - 0.4)x hacknet multiplier - which is a larger decrease than the correct 1/(1 + 0.4) effect.

Because of that, I've been on a quest to ensure that negative mults always function in the correct inverse fashion to positive mults. As a slight kludge, this is using "negative multiplier value" as a proxy for "affecting a decreasing multiplier".

I've made a note in the description.

catloversg (approved):

PR #2993, opened 2026-07-17
BUGFIX: Sleeve "Take on contracts" option displaying empty submenu

### LINKED ISSUE #2949: Empty menu if all bladeburner contracts are already taken by a sleeve
MRE: Get 4 sleeves. Assign 3 of them to perform bladeburner actions > take on contracts and choose a contract for each one. Now do the same for the fourth. You'll notice there is an empty select instead of the contracts list.

Image
In other actions, the main action disappears if there are no "child" actions avaliable.

E9cipher:
Context in and closes #2949
This makes the "Take on Contracts" option not display if all the contracts are already taken by other sleeves.

### DIFF
--- a/src/PersonObjects/Sleeve/ui/TaskSelector.tsx
+++ b/src/PersonObjects/Sleeve/ui/TaskSelector.tsx
@@ -110,7 +110,7 @@ function possibleContracts(sleeve: Sleeve): BladeburnerContractName[] | ["------
       contracts = contracts.filter((x) => x != w.actionId.name);
     }
   }
-  return contracts;
+  return contracts.length !== 0 ? contracts : ["------"];
 }
 
 const tasks: {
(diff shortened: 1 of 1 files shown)


d0sboots:
I want to hear what @catloversg thinks too, but IMO we want to disable the option in some way, rather than remove it.

E9cipher:
The sleeve menu is consistent, it doesn't display the options if their conditions do not "match".  imo, if we're gonna change this (which as you know from #2985 I am in favor of disabling things rather than not rendering them) we should disable all the options, not just this one, and maybe? add a tooltip to explain why it is disabled.

catloversg:
I'm neutral on this. If the Sleeve UI currently removes options that are not applicable, then I prefer doing the same here to be consistent. I'm okay with disabling the options, but that should be done in a separate PR and affect all actions, not just Bladeburner actions. If d0sboots prefers doing that, I'll do it later.

Note that no matter what we come up with, this PR should be minimal. In other words, don't touch anything unrelated to Bladeburner actions. I have not touched #2949 because `TaskSelector.tsx` and its parent need to be refactored. Currently, we can see many React warnings when accessing the Sleeve UI, and even without the warnings, it's fairly clear that the Sleeve UI code is messy just by looking at it. Any changes here will likely become irrelevant after my refactor, especially if I decide to implement the change that d0sboots mentioned (disabling instead of hiding).

d0sboots:
Based on the fact that the Sleeve UI is consistently like this, this seems fine. The area is clearly ripe for refactoring, but there is value in having the full complexity of how it *ought* to be modeled, before doing that.

PR #2995, opened 2026-07-18
CLI: make scan-analyze less powerful

hexagonrecursion:
Marking this pull request as a draft because it depends on #2948, which hasn't been merged yet.

* d0sboots pointed out that there are issues with how `scan-analyze` with the `-a` flag interacts with the dark net.
 https://github.com/bitburner-official/bitburner-src/pull/2948#issuecomment-4898943573
* In my opinion there are two issues:
  1. `scan-analyze 10 -a` often shows more servers than can be seen on the "Dark net" page.
  2. If you have AutoLink.exe, `scan-analyze 10 -a` creates clickable links to all the dark net servers that it displays. The links continue to function even if the destination becomes airgapped.
* d0sboots suggested that links created by `scan-analyze` should work the same as links created by `ns.ui.createConnectLink()` https://github.com/bitburner-official/bitburner-src/pull/2948#issuecomment-4938841268

### more servers than UI
Now scan-analyze -a reveals less information than the UI because it no longer recurses into dnet servers that you do not have admin rights on.

### AutoLink.exe
* Every link created by scan-analyze now contains a sequence of `connect` commands that it tries it execute when it is clicked.
* Example:
  * You have AutoLink.exe
  * You are connected to home
  * You execute scan-analyze
  * scan-analyze produces a link to n00dles
  * Clicking on the link is equivalent to typing `connect home; connect n00dles`, except that the sequence of connect commands is atomic: it will either successfully connect to the destination or it will not move you at all.
* Note that the sequence of connect commands always starts with the server you were connected to when you executed scan-analyze

This has several consequences:
1. Links to dnet servers stop working when server movement makes the path invalid.
2. Links to normal servers do not always work. Example:
   * Step 1: You have AutoLink.exe
   * Step 2: You are connected to n00dles
   * Step 3: n00dles is not backdoored
   * Step 4: You execute scan-analyze
   * Step 5: scan-analyze produces a links that contain commands that start with `connect n00dles`
   * Step 6: You connect to a server that is not adjacent to n00dles
   * Step 7: If you now click on any of the links produced in step 5, it will tell you that it is unable to connect to n00dles

### DIFF
--- a/src/Terminal/ui/ConnectLink.tsx
+++ b/src/Terminal/ui/ConnectLink.tsx
@@ -19,17 +19,20 @@ export function ConnectLink({ path, text }: IConnectLinkProps): React.ReactEleme
     }
     Terminal.connectToServer(result.destination);
   }, [path]);
-  const first = path[0];
-  const last = path.at(-1);
   let tooltip: string;
-  if (last == null) {
-    tooltip = "";
-  } else if (path.length === 1) {
-    tooltip = `connect ${first}`;
-  } else if (path.length === 2) {
-    tooltip = `connect ${first}; connect ${last}`;
-  } else {
-    tooltip = `connect ${first}; ...; connect ${last}`;
+  switch (path.length) {
+    case 0:
+      tooltip = "";
+      break;
+    case 1:
+      tooltip = `connect ${path[0]}`;
+      break;
+    case 2:
+      tooltip = `connect ${path[0]}; connect ${path[1]}`;
+      break;
+    default:
+      tooltip = `connect ${path[0]}; ...; connect ${path.at(-1)}`;
+      break;
   }
   return (
     <Tooltip title={tooltip}>
--- a/src/Terminal/ui/ConnectLink.tsx
+++ b/src/Terminal/ui/ConnectLink.tsx
@@ -1,5 +1,6 @@
 import React, { useCallback } from "react";
-import { Link as MuiLink } from "@mui/material";
+import Link from "@mui/material/Link";
+import Tooltip from "@mui/material/Tooltip";
 import { Terminal } from "../../Terminal";
 import { Player } from "@player";
 import { validateConnections } from "../../Server/ServerHelpers";
@@ -18,5 +19,21 @@ export function ConnectLink({ path, text }: IConnectLinkProps): React.ReactEleme
     }
     Terminal.connectToServer(result.destination);
   }, [path]);
-  return <MuiLink onClick={onClick}>{text}</MuiLink>;
+  const first = path[0];
+  const last = path.at(-1);
+  let tooltip: string;
+  if (last == null) {
+    tooltip = "";
+  } else if (path.length === 1) {
+    tooltip = `connect ${first}`;
+  } else if (path.length === 2) {
+    tooltip = `connect ${first}; connect ${last}`;
+  } else {
+    tooltip = `connect ${first}; ...; connect ${last}`;
+  }
+  return (
+    <Tooltip title={tooltip}>
+      <Link onClick={onClick}>{text}</Link>
+    </Tooltip>
+  );
 }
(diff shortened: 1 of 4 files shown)


hexagonrecursion:
Alternatively I could run two depth first searches:
1. A search from `home` to find a stable path to every non-dnet server: a path that always works no matter where you are because it begins with `connect home`.
2. A search from the current server to build the tree for scan-analyze to print

hexagonrecursion:
I think I have skipped past an important discussion: how clunky do we want scan-analyze to be? When it prints a link to a non-dweb server, do we want the link to always work no matter which server you were connected to when you executed scan-analyze and no matter which server you are currently connected to? If not, do we want the links printed while you are connected to home to always work?

d0sboots:
I like your proposed solution, despite (or possibly because of) the additional clunkiness it adds to scan-analyze. I think most people run it from home, so they won't notice.

A possible UI addition to help with the clunkiness would be to add a tooltip to connectlinks that shows the initial server; this would be useful for both scan-analyze and programmatic ones. 

@catloversg @ficocelliguy thoughts?

catloversg:
That looks okay to me. The consequence for normal servers is not an issue. For dnet servers, any downside is acceptable. I consider `scan-analyze -a` working with dnet servers a bug/oversight, so any clunkiness when using `scan-analyze` for dnet servers is actually a good thing in my opinion.

Adding a tooltip is also a good idea.

hexagonrecursion:
1. #2948 was squashed. I rebased hexagonrecursion:scan-2 to remove the squashed commits
2. added tooltips

This pull request is ready for review.

src/Terminal/ui/ConnectLink.tsx
```diff
@@ -18,5 +19,21 @@ export function ConnectLink({ path, text }: IConnectLinkProps): React.ReactEleme
     }
     Terminal.connectToServer(result.destination);
   }, [path]);
-  return <MuiLink onClick={onClick}>{text}</MuiLink>;
+  const first = path[0];
+  const last = path.at(-1);
+  let tooltip: string;
+  if (last == null) {
+    tooltip = "";
```
d0sboots: Do empty tooltips not appear? I feel like we either want that, or a "does nothing" label, but probably the former if an empty link has any purpose.

src/Terminal/ui/ConnectLink.tsx
```diff
@@ -18,5 +19,21 @@ export function ConnectLink({ path, text }: IConnectLinkProps): React.ReactEleme
     }
     Terminal.connectToServer(result.destination);
   }, [path]);
-  return <MuiLink onClick={onClick}>{text}</MuiLink>;
+  const first = path[0];
+  const last = path.at(-1);
+  let tooltip: string;
+  if (last == null) {
```
d0sboots: Optional: This can be a switch on path.length.

d0sboots (commented):
This looks good, I'll wait for catlover to double-check if he wants to.

src/Terminal/ui/ConnectLink.tsx
hexagonrecursion: I tried it. The empty tooltips do not appear.

d0sboots:
It looks like catlover doesn't want to double-check.

catloversg:
Sorry for being late. I have been busy for the last few days, so I have not reviewed many PRs. This looks okay to me.

By the way, when I'm requested to review a PR, I always perform the review eventually. Sometimes I even review PRs that were merged before I had time to check them. If, for some reason, I refuse to review a PR that I was requested to review, I'll explicitly say that. Do you want me to self-request reviews on PRs that I was requested to review (without using GitHub's "request review" feature) but have not finished reviewing?

d0sboots:
> By the way, when I'm requested to review a PR, I always perform the review eventually.

I think this is where the confusion was. AFAIK, you weren't explicitly requested to review, rather I was holding off on submitting in case you *wanted* to review. Since it seemed like you didn't want to (or were busy), I went ahead and merged. 

For the ones where I have asked for your review (like my own PRs), I'm happy to wait until you can get to it.

PR #2999, opened 2026-07-18
DOCUMENTATION: Clarify ns.singularity.softReset and soft reset button tooltip

branchtheory:
Thanks Hallowed for the idea 💫

I've corrected and expanded the documentation and UI tooltip text for soft resets. The new version:

- corrects 'any augmentations installed' to 'augmentations queued'
- gives some extra context so it's clear what a 'reset' actually involves

### DIFF
--- a/markdown/bitburner.singularity.softreset.md
+++ b/markdown/bitburner.singularity.softreset.md
@@ -56,5 +56,5 @@ void
 
 RAM cost: 5 GB \* 16/4/1
 
-This function performs the same reset as when you install Augmentations, but it can be performed even when you don't have Augmentations queued. If you do have Augmentations queued, it will install them.
+Performs the same reset as when you install Augmentations. This can be used even when no Augmentations are queued. Installs any queued Augmentations.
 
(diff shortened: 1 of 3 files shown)


hexagonrecursion (approved):

markdown/bitburner.singularity.softreset.md
```diff
@@ -56,5 +56,5 @@ void
 
 RAM cost: 5 GB \* 16/4/1
 
-This function will perform a reset even if you don’t have any augmentation installed.
+This function performs the same reset as when you install Augmentations, but it can be performed even when you don't have Augmentations queued. If you do have Augmentations queued, it will install them.
```
ficocelliguy: "If you do have Augmentations queued, it will install them." can be simplified to something like  "Installs any queued augmentations." Same with the tooltip

branchtheory: Doc
> Performs the same reset as when you install Augmentations. This can be used even when no Augmentations are queued. Installs any queued Augmentations.

Tooltip
> Perform the same reset as when you install Augmentations. This can be used even when no Augmentations are queued. Installs any queued Augmentations.

(only difference is no 's' on perform, to match the other tooltips)

That any better? Feels a lot crisper to me.

PR #3000, opened 2026-07-19
MISC: Fix WHRNG description wrong

E9cipher:
Before:
```ts
export class WHRNG implements RNG {
  s1 = 0;
  s2 = 0;
  s3 = 0;

  constructor(totalPlaytime: number) {
    // This one is seeded by the players total play time.
    const v: number = (totalPlaytime / 1000) % 30000;
    this.s1 = v;
    this.s2 = v;
    this.s3 = v;
  }
  ```
  The comment is wrong. The most cases use `Date.now().getTime()`, not the player's total playtime. I moved the comment to `SlotMachine.tsx` which does use the player's total playtime. I also renamed the param to `randomizer`, though I think it could be better renamed to `seed`.

### DIFF
--- a/src/Casino/RNG.ts
+++ b/src/Casino/RNG.ts
@@ -42,9 +42,8 @@ export class WHRNG implements RNG {
   s2 = 0;
   s3 = 0;
 
-  constructor(totalPlaytime: number) {
-    // This one is seeded by the players total play time.
-    const v: number = (totalPlaytime / 1000) % 30000;
+  constructor(randomizer: number) {
+    const v: number = (randomizer / 1000) % 30000;
     this.s1 = v;
     this.s2 = v;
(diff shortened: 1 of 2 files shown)


catloversg:
I won't veto this change, but I would like to keep this comment. It teaches people that they should never blindly trust the comment. I saw some people make this exact mistake on Discord when they tried to cheat the casino, and I really like it. To me, it's more of an Easter egg than a bug.

d0sboots:
I don't agree with catlover here (I like our comments to not be stale), but I understand the perspective. 

It was undoubtedly a bad comment to begin with (a class never controls how it is used, and naming the argument `totalPlaytime` would have been enough, if it were correct.) But I don't have strong feelings about it.

PR #3001, opened 2026-07-19
CODEBASE: Update parameter name and comment in WHRNG constructor

catloversg:
The current comment is technically incorrect, but I want to keep it for the reason I explained in #3000. I think the new comment strikes a good balance. For contributors, the parameter name is now correct (`seed` instead of `totalPlaytime`), and the "Trying to cheat?" bait clearly signals that the comment is not intended for them. For players trying to cheat the casino, it still serves as a lesson.

### DIFF
--- a/src/Casino/RNG.ts
+++ b/src/Casino/RNG.ts
@@ -42,9 +42,9 @@ export class WHRNG implements RNG {
   s2 = 0;
   s3 = 0;
 
-  constructor(totalPlaytime: number) {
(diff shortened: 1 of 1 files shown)


E9cipher:
What about #3000? Do I close it?

catloversg:
Yes.

d0sboots:
Cheeky. Strikes a good balance of deception but not outright lying.

PR #3004, opened 2026-07-20
MISC: Add hints for debugging infinite loop

catloversg:
Accidentally writing an infinite loop is one of the most common mistakes. Players with little programming experience usually don't know how to debug their scripts. Some players have never even heard of the word "debugger". This PR adds several hints to help players troubleshoot this common issue.

The example script is placed on the n00dles server, which is likely to be discovered by new players. "Koike-san" is a recurring character in many manga created by Fujiko Fujio. This character loves ramen.

### DIFF
--- a/src/Documentation/doc/en/programming/game_frozen.md
+++ b/src/Documentation/doc/en/programming/game_frozen.md
@@ -2,10 +2,18 @@
 
 ## Infinite Loop in Scripts
 
-If your game is frozen or stuck in any way, then the most likely culprit is an infinitely running loop in your script.
-To get past the freezing, run the game with `?noScripts` in the URL:
+### Overview
+
+If your game is frozen or stuck in any way, then the most likely culprit is an infinite loop in your script.
+To recover from the freeze, run the game with `?noScripts` in the URL:
 
 [Link to no freeze](https://bitburner-official.github.io?noScripts)
+
+If you are playing the Steam version, you can reload the game without running scripts by using the menu option: Reloads => Reload & Kill All Scripts.
+
+If you run the executable file from the command line, you can use the `--no-scripts` parameter.
+
+### Fix the infinite loop
 
 Then, to fix your script, make sure you have a `sleep()` or any other timed function like `hack()` or `grow()` in any infinite loops:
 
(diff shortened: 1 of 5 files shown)


src/Documentation/doc/en/programming/game_frozen.md
```diff
@@ -49,6 +57,28 @@ If the player does not currently have enough money to purchase a server, the `if
...
+- Steam version:
+  - Menu: Debug => Activate
+  - Shortcut:
+    - Windows/Linux: F12
+    - macOS: Cmd + Option + I
+
+In the developer tools:
+
+- Open the Sources tab (Chrome/Safari) or the Debugger tab (Firefox).
+- Click the Pause button. It usually looks like this: ⏸ and is located on the right side of the Sources/Debugger tab.
```
E9cipher, line 78: I think everyone knows how a pause button is. If they do not, I don't see how they're playing bitburner at all...

catloversg: If I can assume what people should know based on common sense, I'll veto and revert a lot of PRs, including ones I made. I'll be frank with you, a lot of things that we made are unnecessary if people use search engines efficiently and use common sense.

However, that assumption is sometimes incorrect, and based on my experience, it's better to limit the assumptions. For example, you assume nobody will ask where the Pause button is, but I'm sure someone will. People think `ls` is `1s`. People think `1Q` is `1q`. Are you sure nobody will fail to find a button that uses an icon instead of the text "Pause", especially when they are overwhelmed by a lot of stuff in the debugger UI, which they have never seen before?

E9cipher: I think people should learn to use the dev tools _before_ attempting to do anything with bitburner. My suggestion is on the sentence "it usually looks like this", which I think is not necessary. The rest is fine for me.

d0sboots: The whole point of this PR is for people who have starting playing bitburner without learning anything about the dev tools (or, likely, even knowing they exist.) And I strongly disagree that that knowledge should be a prerequisite.

catloversg: Yeah. I agree that learning how to use dev tools is optional. This is a game with a diverse group of players, so it's normal for people not to know how to use a debugger to troubleshoot efficiently. We should introduce the tools, but gatekeeping the game behind knowledge of dev tools is not okay.

Issue #2997, opened 2026-07-18
BN-9 challenge text / criteria


branchtheory:
## Problems

The text of the achievement currently says:

> BN9: Challenge
> Destroy BN9 without using hacknet servers or hacknet nodes.

There are two issues here.

(1) The text is weird. 
* 'Hacknet nodes' is irrelevant because you can't use them anyway.
* Even if you don't actively _use_ the intial hacknet server (to upgrade or spend hashes), you still fail the achievement for _passively benefiting from_ its money production once its cache maxes out.

(2) It doesn't match the condition in the code. 

Here's the challenge in Achievements.ts:

```
CHALLENGE_BN9: {
    ...achievementData.CHALLENGE_BN9,
    Icon: "BN9+",
    Visible: () => canAccessBitNodeFeature(9),
    Condition: () =>
      Player.bitNodeN === 9 &&
      isBitNodeFinished() &&
      Player.moneySourceB.hacknet === 0 &&
      Player.moneySourceB.hacknet_expenses === 0,
  },
```

So you can spend hashes. You just can't produce money from or spend money on hacknet servers. So if you wanted you could get a lot of value out of your starting hacknet.

## Suggestions

(1) Update the challenge text to be clearer. 

Either:

> BN9: Challenge
> Destroy BN9 without using hacknet servers.

Or maybe better:

> BN9: Challenge
> Destroy BN9 without the help of hacknet servers.

Because it's possible to be aided by the starting hacket server (through passive money production) without strictly using it.

(2) Update the condition in the challenge.

gmcew:
FWIW, I think the achievement condition should definitely check hashes spent as well as the above wording clarification, rather than just monetary gain/spend. That seems vital to a "don't use hashnet" challenge!

catloversg:
The achievement description can be improved for clarity, but what you said is missing some important details.

First, before we added BN options, getting this achievement was really difficult if you had a very high SF12 level. With too many NFGs at the start, the hash value exceeds the max cache and is converted to money immediately, so you cannot get this achievement. I don't remember whether there was a workaround, but even if there was, earning this achievement was still too cumbersome. Now you can override the SF12 level or disable hacknet servers with BN options.

Second, with BN options, you can use hacknet nodes again even after achieving SF9. This means the "hacknet nodes" part is relevant. In fact, I specifically added that phrase to clarify the condition. If you disable hacknet servers and use hacknet nodes, you won't get this achievement.

[...]

PR #3005, opened 2026-07-20
MISC: Clarify conditions for "BN9: Challenge" achievement

catloversg:
Context: #2997.

The alternative approach is to change the conditions. I'll open a separate PR for that.


d0sboots:
This is an improvement, but I think the true fix for 2997 involves changing the condition.

hexagonrecursion:
The new description makes it sound like you are allowed to buy and/or upgrade hacknet servers, but (assuming I understood Achievements.ts correctly) this isn't actually the case

catloversg:
Adding more text to clarify the current conditions is unnecessary at this point. I'll change them anyway.

PR #3008, opened 2026-07-22
API: Convert a contract into a dummy contract and return the expected answer

hexagonrecursion:
When a contract self-destructs, we tell the players what the correct answer was. This is very helpful for players who are trying to figure out what is wrong with their solution, but this feature has poor discoverability: you only find out about it if you _manually_ submit enough incorrect answers to make the contract self-destruct.

I propose an API that would let players get the correct answer for a contract at the expense of turning the contract into a dummy contract that gives no reward.


d0sboots:
Not all contracts are capable of generating answers, and that is by design. I feel like this might overpromise.

If we were going to do an API like this, I feel like it would be better to just immediately kill the contract and return the answer, and have it fail if it's not a type that supports answers. That wouldn't have to care if it's a dummy or not. 

But, I'm still not sure how I feel about this overall.

gmcew:
I was unconvinced by the change that failed ccts tell the player the answer, this makes ccts even more hand-holding.

I believe both changes are/were unnecessary for a couple of reasons:
- you can generate test data using the dummy ccts to get a solver working
- real ccts don't expire over time, so it's possible to fix an edge case without losing the reward by fixing the code bug and simply not repeatedly trying a wrong answer. If you lose them through reset you didn't need the reward anyway 

Cct generation with darknet, hash, and the recent 3x rate change is more frequent than it's ever been, with a way to test them effectively for free. I don't feel they need to be any easier.

catloversg:
I veto this change. d0sboots and gmcew already explained the reasons.

PR #3012, opened 2026-07-25
UI: Automatically connect when clicking Remote API status icon

E9cipher:
This PR makes the remote API icon on the overview try a connect attempt when it is clicked. Regardless of the result, it still routes the player to the options tab.
I added this because it's bothering that to connect (when you already have the settings) you have to navigate to the options page, resetting the UI on whatever you were doing.

### DIFF
--- a/src/GameOptions/ui/RemoteFileApiConnectionStatus.tsx
+++ b/src/GameOptions/ui/RemoteFileApiConnectionStatus.tsx
@@ -10,6 +10,7 @@ import { Settings } from "../../Settings/Settings";
 import { Router } from "../../ui/GameRoot";
 import { Page } from "../../ui/Router";
 import { RemoteFileApiConnectionEvents } from "../../RemoteFileAPI/Remote";
+import { isValidConnectionHostname, isValidConnectionPort } from "src/Settings/SettingsUtils";
 
 export const RemoteFileApiConnectionStatus = ({ showIcon }: { showIcon: boolean }): React.ReactElement => {
   const [rfaConnectionStatus, setRfaConnectionStatus] = useState(getRemoteFileApiConnectionStatus());
@@ -43,8 +44,8 @@ export const RemoteFileApiConnectionStatus = ({ showIcon }: { showIcon: boolean 
           onClick={() => {
             if (
               !isRemoteFileApiConnectionLive() &&
-              Settings.RemoteFileApiAddress !== "" &&
-              Settings.RemoteFileApiPort > 0
+              isValidConnectionHostname(Settings.RemoteFileApiAddress) &&
+              isValidConnectionPort(Settings.RemoteFileApiPort)
             ) {
               newRemoteFileApiConnection();
             }
--- a/src/GameOptions/ui/RemoteFileApiConnectionStatus.tsx
+++ b/src/GameOptions/ui/RemoteFileApiConnectionStatus.tsx
@@ -1,6 +1,10 @@
 import React, { useState, useEffect } from "react";
 import { Box, IconButton, Tooltip, Typography } from "@mui/material";
-import { getRemoteFileApiConnectionStatus } from "../../RemoteFileAPI/RemoteFileAPI";
+import {
+  getRemoteFileApiConnectionStatus,
+  isRemoteFileApiConnectionLive,
+  newRemoteFileApiConnection,
+} from "../../RemoteFileAPI/RemoteFileAPI";
 import OnlinePredictionIcon from "@mui/icons-material/OnlinePrediction";
 import { Settings } from "../../Settings/Settings";
 import { Router } from "../../ui/GameRoot";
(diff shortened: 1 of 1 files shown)


src/GameOptions/ui/RemoteFileApiConnectionStatus.tsx
```diff
@@ -34,7 +38,19 @@ export const RemoteFileApiConnectionStatus = ({ showIcon }: { showIcon: boolean
   return (
     <Box style={{ display: "flex", flex: 1, justifyContent: "flex-start", alignItems: "center" }}>
       {showIcon ? (
-        <IconButton aria-label="Remote API status" onClick={() => Router.toPage(Page.Options, { tab: "Remote API" })}>
+        <IconButton
+          aria-label="Remote API status"
+          onClick={() => {
+            if (
+              !isRemoteFileApiConnectionLive() &&
+              Settings.RemoteFileApiAddress !== "" &&
```
catloversg: Use `isValidConnectionHostname` and `isValidConnectionPort`.

catloversg (commented):
This feature is debatable. Let's see what other people think about it.

E9cipher:
Ok, I have no idea of why this failed

catloversg:
Wrong import.

d0sboots:
I think clicking on an icon should do one thing - in this case, either go to the options page *or* try to reconnect, depending on what we decide.

My gut says that this is trying to fix the wrong problem - i.e. it is trying to fix "I need to reconnect frequently" by making it an easier UI action, when the correct fix is to use a lower reconnect timeout and/or to fix whatever is making the connection flaky in the first place. But, I don't use the remote API myself, so I could be misunderstanding the usecase here.

E9cipher:
My case pretty much is that I need a quick shortcut to enable remote API. The way I see it, this can be a very useful feature. Maybe we can add an optional setting for it, leave it off by default.

The reconnect timeout can be bothering - at least for me. It tries to reconnect no matter what, if you are playing the game, but don't want remote API it's still there, trying to reconnect. Higher timeouts mean more waiting when you want to connect.

You mentioned that an icon should do one thing. Adding another icon or button seems unnecessary to me, but I'm not totally against it, that's why I'd say to do this in the remote API icon, it preserves functionality the same.

d0sboots:
OK, but *why* do you need a quick shortcut to enable remote api? *Why* is it a problem if it quickly tries to reconnect? 

Remote API was designed to be a long-term session, not a frequent connect-and-disconnect, so if you want to support a different way we need to understand the usecase.

E9cipher:
I'll do my best to explain it better 👍
Basically I need a way for the remote API button to not drive me out of the current page. A popup/dialog with the connect opts could also work.

My setup is: I open Bitburner, do some stuff. Then, open vscode and I activate the extension, which is just a click and it's set up. With Bitburner, I have to click the API button, and then click connect. This is tedious on a daily basis.

The problem with reconnect is that on that "do some stuff" timestamp, it's continuously spam-toasting me. Which is good for the use case of the reconnect feature, it's exactly what it's been built for. But it's not build for this kind of setup (which is pretty normal imo)

Hope that clarifies my point.

d0sboots:
That sounds like what should have been fixed by https://github.com/bitburner-official/bitburner-src/pull/2535. Toasts should not spam you when it is trying to reconnect, you should only get one, so that it is ergonomic to just leave the remote API enabled all the time.

E9cipher:
Ooops sorry, I was using the wrong build :(
True, toast spam is over.
There's however another problem, but it's more of a non-standard case, so I'll leave it up to you whether to merge this or close it.
Open bitburner, set reconnect timeout to 5s (for example). Open vscode, connect to the remote API using Fico's extension or another plugin. Then, click "connect" to close the connection.
After the 5s, bitburner will reconnect again, even if you were trying to connect to another instance of bitburner. The solution is to set the reconnect timeout to 0, click "connect" to disconnect, then connect to the other instance using same method, then go back to your original bitburner and set again the reconnect timeout.

It would have been easier to just click the button to disconnect, then connect to the other instance using same method. Two clicks.

Also, the "connect" button should switch from "connect" to "disconnect", but I'll treat that in another PR.

Sorry for insisting so much :)

d0sboots:
I didn't follow all of the steps there, but if I understand correctly you are talking about the friction experienced when juggling two instances of BitBurner, both of which want to talk to the same remote API endpoint? 

Although that is unusual, I'm suspecting that others have similar usecases. So now it just comes down to getting the UX right.

d0sboots:
@catloversg I am spitballing ideas here, let me know if this send like good UI to you. 

I am thinking of turning the icon into a toggle - clicking on it while connected/reconnecting will set the state to disconnected, and clicking while disconnected will go to reconnecting. To avoid silently doing nothing, if the hostname/port is unconfigured a click would go to the settings (like in the current state of the PR).

Discoverability of this will be poor, but I think that's OK? We could also add "Click to X" in the tooltips to help with that part.

E9cipher:
Didn't ask me, but I think it's a good idea :)

catloversg:
I made a PoC for your idea, and it seems okay to me. I'll submit a PR for it.

E9cipher:
Already done by #3050

Issue #2943, opened 2026-07-01
Prevent running port openers against servers you own


ficocelliguy:
Currently, the command-line programs used for gaining sever access (BruteSSH, FTPCrack, relaySMTP, HTTPWorm, SQLInject, and Nuke) all work on the player's home server and other owned servers. However, this "punishes" players for experimenting with them and running them as soon as they first make them, showing a concerning message that implies your home server is now more vulnerable. 

To better reward exploration, and also make it more clear how they are intended to be used, the command-line versions of these scripts could log a descriptive error when run on player-owned servers.

The API versions of these can be left as-is, but they could also be changed to return `false` when run on player-owned servers to be consistent.

Context:
https://discord.com/channels/415207508303544321/415207839246581781/1510696198112284823

E9cipher:
Are you working on a PR? I could take this if you want.

ficocelliguy:
Nope, I'm putting these here for folks who want them to grab

catloversg:
Similar to #2942, I think this is an issue that should be resolved by improving the documentation instead of changing how things work.

First, CLIs and NS APIs should behave identically. For example, if `ns.brutessh("home")` returns true, then running `run BruteSSH.exe` on home should also succeed. There is no reason to make them behave differently. If running `run BruteSSH.exe` on home fails, we would need to introduce a breaking change to the `brutessh` API, and I don't think that breaking change is necessary.

Second, although the original question on Discord is not a textbook example of the XY problem, it has the same underlying issue: the question is based on an incorrect assumption. They asked how to close a port after running a port cracker. Although they did not state why they wanted to do that, they likely wanted to make their home computer secure again by closing the ports they had opened, based on the assumption that opening ports on your home computer has security implications in real life. However, this is currently not true in Bitburner. Opening ports on your own machines has no negative consequences.

I agree that showing a message that implies the home server becomes vulnerable is not ideal, but this can be resolved by improving the documentation. Instead of changing how CLIs and NS APIs work, we only need to clarify that opening ports on your own machine is harmless.

Besides updating the documentation, I'll also open a PR that resets the five port properties on home when prestiging. Optionally, we can also tweak the message to make it clear that this action is harmless when it is performed on player owned machines.

hexagonrecursion:
1. Improving documentation is inferior to improving the user interface, because people, who read carefully read documentation, are an exception rather than the norm.
[...]

PR #3013, opened 2026-07-25
UI: Add clarifying message when running port openers on home

E9cipher:
Context in #2943 
I added a message, suggestion from catlover in https://github.com/bitburner-official/bitburner-src/issues/2943#issuecomment-4861979179 .
Note that the message does not return and is after the "port opened" check, therefore only appears when _opening_ the port. I can tweak the message as you see.

### DIFF
--- a/src/Programs/Programs.ts
+++ b/src/Programs/Programs.ts
@@ -94,7 +94,7 @@ export const Programs: Record<CompletedProgramName, Program> = {
         Terminal.print("SSH Port (22) is already open!");
         return;
       }
-      if (server.hostname === "home" || server.purchasedByPlayer) {
+      if (server.purchasedByPlayer) {
         Terminal.print("Opening ports on your own machines has no negative consequences in the game.");
       }
 
@@ -122,7 +122,7 @@ export const Programs: Record<CompletedProgramName, Program> = {
         Terminal.print("FTP Port (21) is already open!");
         return;
       }
-      if (server.hostname === "home" || server.purchasedByPlayer) {
+      if (server.purchasedByPlayer) {
         Terminal.print("Opening ports on your own machines has no negative consequences in the game.");
       }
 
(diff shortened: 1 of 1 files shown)


catloversg:
Nitpick:
- This is not a warning. We only warn players when they are performing a potentially dangerous action. This *in-game* action is harmless.
- Opening ports *does have* an effect on home. The effect is that, well... the port is open.

The message should indicate that action is harmless. E.g., "Opening ports on your own machines has no negative consequences in the game." This message should apply to all servers owned by players, not just home.

Note that the "in the game" part is important. It's not ideal because it breaks the immersion, but if we only say that opening ports is harmless, naive players may think this is also true in the real life. Breaking the immersion is acceptable in this case. I don't want to mislead people when it comes to security.

And don't close that issue with this PR. The most important part is port documentation, which I'm writing.

E9cipher:
Gotcha. I'll update description and the message

E9cipher:
@catloversg I'm not sure how to write the message.  "Opening ports on this kind of server does not affect anything in-game"?

catloversg:
"Opening ports on your own machines has no negative consequences in the game."
That example is in my previous message.

catloversg:
- No need to say "INFO". Just use that message as-is.
- `server.purchasedByPlayer`

catloversg:
Why do you still check for `server.hostname === "home"`? `server.purchasedByPlayer` already covers every case.

E9cipher:
Sorry

d0sboots:
> Sorry

Catlover can appear brusque or even angry at times, but that's (almost) never his intent. So, you don't need to apologize, and don't feel bad if he or I nitpick you a lot in code reviews - we are both just picky reviewers. :D

E9cipher:
Good to know, thanks, though I can be veeery.. well, a very tedious person to work with sometimes (especially when I'm on autopilot) :)

PR #3015, opened 2026-07-28
API: Typesafety for AugmentationName

E9cipher:
This is a (minor) breaking change.

I added typesafety for `AugmentationName`. This should not be a problem except for linter. My suggestion is we add `AugmentationName` to NSEnums, but skip a few intentionally (e.g. TRP).

I will add examples as they come to me, but I've tested this code (which would be a perfectly valid code from players) and the change does not affect it (Intellisense detects that `Neural Wit Amplifier`) is a member of `AugmentationName`).
```ts
  const augName = "Neural Wit Amplifier";
  ns.grafting.graftAugmentation(augName);
}
```
Typos however, get caught now. `augName = "Neural Wit Amplifie"` wasn't caught by eslint before, now it does.


src/ScriptEditor/NetscriptDefinitions.d.ts
E9cipher, line 5112: Gang equimpent is still a `string`, because `AugmentationName` is a string, and `string | AugmentationName` is irrelevant. If in future PRs someone decides to use their own type for `equipName`, this will have to be changed to `| AugmentationName` as well.

E9cipher, line 5149: Same equipment comment.

d0sboots:
I don't think it makes sense to expose a TS type for these without a corresponding runtime enum type. And although I won't rule that out completely, per the discussion in development it seems like most of us agree that there are unique things about augs that make this not worth doing.

PR #3016, opened 2026-07-29
API: improve how ns.print() and ns.tprint() print exceptions

hexagonrecursion:
I sometimes write code like this:
```ts
  ns.ui.openTail();
  try {
    JSON.parse('{{');
  } catch(e) {
    // prints: {}
    ns.print(e);
  }
}
```

Unfortunately it prints a very unhelpful `{}`. With this change it prints `SyntaxError: JSON.parse: expected property name or '}' at line 1 column 2 of the JSON data`

### DIFF
--- a/src/Netscript/NetscriptHelpers.tsx
+++ b/src/Netscript/NetscriptHelpers.tsx
@@ -348,7 +348,7 @@ function argsToString(args: unknown[]): string {
     if (nativeArg instanceof Set) {
       return (out += setToString(nativeArg));
     }
-    if (typeof nativeArg === "object" && !(nativeArg instanceof Error)) {
+    if (typeof nativeArg === "object") {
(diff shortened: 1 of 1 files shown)


src/Netscript/NetscriptHelpers.tsx
```diff
@@ -348,7 +348,7 @@ function argsToString(args: unknown[]): string {
     if (nativeArg instanceof Set) {
       return (out += setToString(nativeArg));
     }
-    if (typeof nativeArg === "object") {
+    if (typeof nativeArg === "object" && !(nativeArg instanceof Error)) {
```
d0sboots: This should be with the Promise check so it works with Errors embedded in objects, too. Add a comment about it.

hexagonrecursion: Good idea. Done.

PR #3019, opened 2026-07-29
CLEANUP: delete src/ThirdParty/JSInterpreter.js

hexagonrecursion:
While reading the code I have noticed that we have a file called JSInterpreter.js. It appears to be a JavaScript interpreter written in JavaScript. I was curious to see where how and why we are using it. Based on the output of `git grep JSInterpreter` and `git log -S JSInterpreter` I think that we do not need it anymore.

## Testing
The jest testsuite passes.


d0sboots:
That was NS1 stuff, and the last usages of it died when we ripped out NS1.

PR #3022, opened 2026-08-01
REFACTOR: Reduce redundant reviver references

d0sboots:
PR #2889 is going to need to touch all the sites where toJSON/fromJSON are happening. To avoid dealing with that multiple times, I'm doing this first: Let's get rid of most of the boilerplate by not having to write toJSON/fromJSON except for the exceptional cases.

This replaces the toJSON built-in method with a generic Replacer function that complements the existing generic Reviver. Both functions call Generic_[to|from]JSON automatically if the appropriate jsonReplacer/jsonReviver function is not defined.

includedKeys handling is built-in generically. All common cases are handled with a single line of setup per serializable class, with the next-most-common case being classes that need to define jsonReviver to do custom load processing.

This *should* have 0 effect on the binary format of any saved or loaded game. *Any* changes are a bug in the implementation.

# Testing
- Unit tests pass, obviously.
- Smoke test of saving/loading a game that includes Sleeves and Bladeburner (although I haven't played with bladeburner much, it's possible there are still hidden issues)
- TODO: Manual test of RunningScript with different title states


src/utils/GenericReviver.ts
```diff
@@ -10,11 +49,44 @@ export class JSONReviverError extends Error {
   }
 }
 
+type JsonableInstance = {
+  jsonReplacer?: () => IReviverValue;
+  // This is not really part of the interface. It exists to catch accidental
+  // definitions of toJSON and error due to mismatched types. Classes should
+  // be defining jsonReplacer instead, and this way the mistake is caught at
+  // build time.
+  toJSON?: (badArgument: number) => number;
```
catloversg: It may be okay to do this, but what do you think about using `toJSON?: never;` instead of `toJSON?: (badArgument: number) => number;`?

IIUC, you want to catch this kind of mistake at build time:
```ts
export class Foo {
  bar: number = 0;
  toJSON() {
    return { ctor: "Foo", data: { something } };
  }
  static includedKeys = makeSerializable("Foo", Foo);
}
```

With the new system, no jsonable class needs to define `toJSON`. If a class needs custom save handling, they need to define `jsonReplacer`. The one somewhat exceptional case is `BaseServer`, which defines `toJSONBase`, so that its subclasses can call `toJSONBase` in their `jsonReplacer`. However, `toJSONBase` is also not `toJSON`.

In other words, if a class defines `toJSON`, it's likely a mistake regardless of how its `toJSON` is implemented. Using `toJSON?: never;` would express this intent directly and catch any definition of `toJSON` at build time. This looks like a better choice unless there is a legit use case of `toJSON` in the new system.

src/utils/GenericReviver.ts
```diff
@@ -37,9 +109,42 @@ export function Reviver(_key: string, value: unknown): any {
     throw new JSONReviverError(`Could not locate constructor named ${value.ctor}.`, value.ctor);
   }
 
-  const obj = ctor.fromJSON(value);
+  const obj =
+    ctor.jsonReviver === undefined
+      ? Generic_fromJSON<Record<string, unknown>>(ctor, value.data, ctor.includedKeys)
```
catloversg:
```suggestion
      ? Generic_fromJSON(ctor, value.data, ctor.includedKeys)
```

src/utils/GenericReviver.ts
```diff
@@ -37,9 +109,42 @@ export function Reviver(_key: string, value: unknown): any {
...
   if (ctor.validationData !== undefined) {
     validateObject(obj, ctor.validationData);
   }
   return obj;
 }
+
+/**
+ * A generic "smart replacer" function.
+ * Looks for object values that we know are specially serializable.
+ * If it finds them, it either callse Generic_toJSON, or hands it off
```
catloversg:
```suggestion
 * If it finds them, it either calls Generic_toJSON, or hands it off
```

src/utils/GenericReviver.ts
```diff
@@ -37,9 +109,42 @@ export function Reviver(_key: string, value: unknown): any {
...
+  // the map.
+  // It's important that we don't ever overwrite `constructor` for our serializable classes.
+  // (Data objects are fine, since they would end up finding a value that fails the map lookup.)
+  const ctorName = reverseMap.get(value.constructor);
+  if (ctorName === undefined) {
+    return value;
+  }
+  // It being in the map asserts that it is the correct type.
+  const ctor = value.constructor as JsonableClass;
+  // Typescript no-op: This just defines an optional field.
```
catloversg: "defines an optional field" sounds like a strange way to explain this type cast. How about rewriting the comment and the variable like this?
```ts
  // The constructor is registered as a JsonableClass, so this value may have a jsonReplacer method.
  const jsonable = value as JsonableInstance;
  if (jsonable.jsonReplacer === undefined) {
    return Generic_toJSON(ctorName, value, ctor.includedKeys);
  }
  return jsonable.jsonReplacer();
```

src/utils/GenericReviver.ts
```diff
@@ -37,9 +109,42 @@ export function Reviver(_key: string, value: unknown): any {
...
+  const maybeReplacer = value as JsonableInstance;
+  if (maybeReplacer.jsonReplacer === undefined) {
+    return Generic_toJSON(ctorName, value as Record<string, unknown>, ctor.includedKeys);
```
catloversg:
```suggestion
    return Generic_toJSON(ctorName, value, ctor.includedKeys);
```

src/utils/JSONReviver.ts
```diff
@@ -30,16 +15,19 @@ export const constructorsForReviver: Partial<Record<string, JsonableClass>> = {
 export function Generic_toJSON<T extends Record<string, any>>(
   ctorName: string,
   obj: T,
-  keys?: readonly (keyof T)[],
+  keys: readonly string[] | null,
 ): IReviverValue {
-  const data = {} as T;
+  if (keys == null) {
```
catloversg: Optional:

```suggestion
  if (keys === null) {
```
`keys` cannot be undefined, so it may be better to use strict equality here.

src/utils/JSONReviver.ts
```diff
@@ -53,29 +41,24 @@ export function Generic_toJSON<T extends Record<string, any>>(
...
-export function Generic_fromJSON<T extends Record<string, any>>(
+export function Generic_fromJSON<T extends object>(
   ctor: new () => T,
   data: unknown,
-  keys?: readonly (keyof T)[],
+  keys: readonly string[] | null,
 ): T {
   assertObject(data);
   const obj = new ctor();
+  if (keys == null) {
```
catloversg: Optional:

```suggestion
  if (keys === null) {
```

src/utils/JSONReviver.ts
```diff
@@ -53,29 +41,24 @@ export function Generic_toJSON<T extends Record<string, any>>(
...
+    // This is an unsafe assignment. We may load data with wrong types at runtime.
+    return Object.assign(obj, data) as T;
```
catloversg:
```suggestion
    return Object.assign(obj, data);
```

src/SaveObject.ts
```diff
@@ -268,12 +267,14 @@ export async function importGame(
...
-        const settings: unknown = JSON.parse(parsedSaveData.data.SettingsSave);
+        const settings: unknown = JSON.parse(parsedSaveData.data.SettingsSave, Reviver);
         assertObject(settings);
         // Modify setting
         settings.SyncSteamAchievements = overrideSettings.SyncSteamAchievements;
         // Save modified data back to saveData
-        parsedSaveData.data.SettingsSave = JSON.stringify(settings);
+        parsedSaveData.data.SettingsSave = JSON.stringify(settings, Replacer);
+        // This does *not* use Replacer, because parsedSaveData is in
+        // IReviverData form already (it was never properly Revived).
```
catloversg:
```suggestion
        // IReviverValue form already (it was never properly Revived).
```

catloversg (commented):
I think we should pass `Reviver` to `JSON.parse` calls in these locations:
- `src/DarkNet/effects/SaveLoad.ts`: `loadDarkNet`
- `src/Go/SaveLoad.ts`: `loadGo`
- `src/Infiltration/SaveLoadInfiltration.ts`: `loadInfiltrations`
- `src/Settings/SettingsUtils.ts`: `loadSettings`
- `src/Alias.ts`: `loadAliases` and `loadGlobalAliases`

This ensures we won't have any issues if these data structures use the "jsonable" system in the future.

src/utils/GenericReviver.ts
d0sboots: Done for this and `fromJSON`. Tested that it works properly.

src/utils/GenericReviver.ts
d0sboots: Done.

src/utils/GenericReviver.ts
d0sboots: Done.

src/utils/GenericReviver.ts
d0sboots: Done.

src/utils/GenericReviver.ts
d0sboots: Done.

src/utils/JSONReviver.ts
d0sboots: Done.

src/utils/JSONReviver.ts
d0sboots: Done.

src/utils/JSONReviver.ts
d0sboots: Done.

src/SaveObject.ts
d0sboots: Done.

d0sboots:
> I think we should pass `Reviver` to `JSON.parse` calls in these locations:
> 
> * `src/DarkNet/effects/SaveLoad.ts`: `loadDarkNet`
> * `src/Go/SaveLoad.ts`: `loadGo`
> * `src/Infiltration/SaveLoadInfiltration.ts`: `loadInfiltrations`
> * `src/Settings/SettingsUtils.ts`: `loadSettings`
> * `src/Alias.ts`: `loadAliases` and `loadGlobalAliases`
> 
> This ensures we won't have any issues if these data structures use the "jsonable" system in the future.

Done.

Smoke-test: This still loads my dev save correctly.

catloversg (approved):

PR #3026, opened 2026-08-02
DNET: populate server logs with noise upon heartbleed

### LINKED ISSUE #3024: DNet: random logs do not spawn
Dnet servers are supposed to generate random logs every `logTrafficInterval` seconds. It is supposed to be possible to retrieve the logs by using `ns.dnet.heartbleed()` (if your charisma is high enough), but it usually returns no logs, no matter how long you wait. The only action I have found that causes the random logs to spawn is manually opening the server details modal.

## To reproduce
1. Enter bitnide 15
2. Run this on `home`:
    ```ts
      ns.clearLog();
      ns.ui.openTail();

      const host = 'darkweb';
      const {logTrafficInterval} = ns.dnet.getServerDetails(host);

      // discard logs that may have been generated previously
      await ns.dnet.heartbleed(host, {logsToCapture: 1000}); 
      // sleep for twice the logTrafficInterval
      await ns.asleep(2000 * logTrafficInterval);
      const {logs} = await ns.dnet.heartbleed(host, {logsToCapture: 1000});
      if(logs.length >= 1) {
        ns.print(`SUCCESS got ${logs.length} logs`);
      } else {
        ns.print('ERROR Expected one or more logs, but got 0');
      }
    }
    ```
## Expected result:
Two or three logs are generated

## Actual result:
No logs are generated

## Notes
1. The example uses darkweb, but I tried the same with other servers too: authenticated, not authenticated, with session and without. No logs were generated. To do this I executed the script on the `darkweb` server and targeted an adjacent server. I made sure that the server remains adjacent to darkweb by using the `connect` command because dnet does not move the server that your terminal is currently connected to. I made sure that I have enough charisma for the target server.
2. If you run the same script while looking at the server details modal for the target server, the logs are generated as expected.

## Game version
v3.0.1 (3162fd2) and v3.0.2 (f34116c)

ficocelliguy:
Closes #3024 

Currently, server logs only get their random garbage log entries that they are due when you look at the server in the UI or when you authenticate via the API. This PR also updates the logs with any missing noise entries upon heartbleed, to make the noise more consistent based on the server's noise internal stats, and to make it more intuitive to players how the noise works.

### DIFF
--- a/src/NetscriptFunctions/Darknet.ts
+++ b/src/NetscriptFunctions/Darknet.ts
@@ -50,6 +50,7 @@ import { type DarknetServerData, getDarknetServerOrThrow } from "../DarkNet/util
 import { shuffle } from "lodash";
 import { getSharedChars } from "../DarkNet/utils/darknetAuthUtils";
 import { freezeServer } from "../DarkNet/controllers/NetworkMovement";
+import { populateServerLogsWithNoise } from "../DarkNet/models/packetSniffing";
 
 type CompleteHeartbleedOptions = {
   peek: boolean;
@@ -284,6 +285,7 @@ export function NetscriptDarknet(): InternalAPI<DarknetAPI> {
           };
         }
         const serverState = getServerState(server.hostname);
+        populateServerLogsWithNoise(server);
 
         logger(ctx)(`Extracted log data from ${server.hostname}... (Gained ${formatNumber(xpGained, 1)} cha xp)`);
 


hexagonrecursion:
Can confirm: bug fixed.

d0sboots:
The whole way that this works leaves a sour taste of premature optimization in my mouth, but I'm not gonna insist that it get folded in to processDarknet at this point.

PR #3028, opened 2026-08-03
UI: add price history charts to Stock Market

AGuyWhoSkis:
Adds visual price history to the 4S Data Market Access upgrade

Linked issues: none

Quick preview
image

# Documentation

New feature: Stock price history visual charts
  * **Requires 4S Market Data Access upgrade**
  * Stock price history is now recorded (in-memory only)
    *  _All price history is erased on game reload/exit, or after a max duration_
  * Expanded stock rows render a full chart
    * Mouse hover on expanded chart to see time-specific price detail
  * Collapsed stock rows render a compressed chart
    * Mouse hover on collapsed chart to see recent price detail

New options: _Interface > Stock Market price charts_
* Toggle price history
* Select chart type: Line or Candlestick 
* Price history duration, in minutes (default 15)
* Market ticks per candle (default 5)
* Toggle price chart preview in collapsed ticker rows

New options
image

### More screenshots

Candlestick charts (preview option enabled)
image

Candlestick charts (preview option disabled)
image

Line charts (preview option enabled)
image

Line charts (preview option disabled)
image

Before purchasing `4S Market Data Access` upgrade
image

After purchasing `4S Market Data Access` upgrade
image

Compressed chart tooltip
image

### Non-obvious things that have been accommodated
* Jest
* Window resizing
* Themes
* Currency symbol

#### The provably resolved "memory leak" concern
Since recording a price point per stock per tick sounded risky for long-running sessions, the default test suite now carries a test that asserts total retained points stay exactly flat at symbols x maxHistoryLength (once the price history buffer fills)

The same test scales to a full 14-day run (~201k ticks) via `STOCK_SOAK=1 npx jest test/jest/StockMarket/PriceHistory.test.ts` - which passes.

BTW: `npm run format` and `npm run lint` were both run before pushing

### DIFF
--- a/src/GameOptions/ui/InterfacePage.tsx
+++ b/src/GameOptions/ui/InterfacePage.tsx
@@ -1,16 +1,34 @@
 import React, { useState } from "react";
-import { TextField, Tooltip, Typography } from "@mui/material";
+import { Box, MenuItem, Select, SelectChangeEvent, TextField, Tooltip, Typography } from "@mui/material";
 import { Settings } from "../../Settings/Settings";
+import { StockChartTypeSetting } from "../../Settings/SettingEnums";
+import { clearStockPriceHistories } from "../../StockMarket/PriceHistory";
 import { OptionSwitch } from "../../ui/React/OptionSwitch";
 import { GameOptionsPage } from "./GameOptionsPage";
+import { OptionsSlider } from "./OptionsSlider";
 import { formatTime } from "../../utils/helpers/formatTime";
 
 export const InterfacePage = (): React.ReactElement => {
   const [timestampFormat, setTimestampFormat] = useState(Settings.TimestampsFormat);
+  const [stockChartType, setStockChartType] = useState(Settings.StockChartType);
+  const [stockHistoryDisabled, setStockHistoryDisabled] = useState(Settings.DisableStockPriceHistory);
 
   function handleTimestampFormatChange(event: React.ChangeEvent<HTMLInputElement>): void {
     setTimestampFormat(event.target.value);
     Settings.TimestampsFormat = event.target.value;
+  }
+
+  function handleStockChartTypeChange(event: SelectChangeEvent): void {
+    const chartType = event.target.value as StockChartTypeSetting;
+    setStockChartType(chartType);
+    Settings.StockChartType = chartType;
+  }
+
+  function handleStockHistoryDisabledChange(newValue: boolean): void {
+    setStockHistoryDisabled(newValue);
+    Settings.DisableStockPriceHistory = newValue;
+    // Disabling is a promise that nothing is retained, not just that nothing new is recorded.
+    if (newValue) clearStockPriceHistories();
   }
   return (
     <GameOptionsPage title="Interface">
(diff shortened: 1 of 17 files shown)


d0sboots:
Like most AI-gen PRs, this will take a while to review because of its size.

Thank you for including screenshots, that makes things much faster to evaluate. A few top-level comments: 

I think you're adding too many options. Most of these things don't need to be configurable. I don't think having two chart styles is worth it, from code complexity and options - we should choose the style that works the best and just use that. IMO that would probably be the line chart, because it can fit all the data as it comes without needing to aggregate like the candlestick chart. Also, it is more intuitive. 

I think the only option you'll end up needing is "stock market history size," with 0 indicating disabled. 

Other changes: I think it makes more sense to have the time axis be "time" rather than "time since data point." That lines up with how traditional stock displays work, and when the data updates the label on the point won't change.

AGuyWhoSkis:
Thanks for your comments!

Apologies for the size of this PR - I get that large changes tend to be more painful to review. The complexity came from good intentions, but simplifying makes a lot of sense.

Here's my suggestion to reduce the burden on reviewers:

1. Gather feature-level (UX) feedback first; no code reviews for now
2. Remove complexity
3. Now code reviews make the most sense

I'll keep an eye on this thread. If a consensus or verdict on (2) becomes obvious I'm happy to make those changes.

AGuyWhoSkis:
With maintainability as a priority, here's the **minimum** changes that are currently planned
- Remove Candlestick chart option
- Replace all options with a single slider for "stock market history size" (0m-60m)
- Chart preview for collapsed stocks remains

Here are more features **not** currently planned (RE: complexity) but may be worth it anyway:
- Render buy/sell orders made by the player as red/green points on the chart (with tooltip summary)
- Add red/green line segments corresponding to stock forecast at that time (Sparkline enhancement)
  - _currently the entire graph switches between red/green according latest forecast_

---

> Other changes: I think it makes more sense to have the time axis be "time" rather than "time since data point." That lines up with how traditional stock displays work, and when the data updates the label on the point won't change.

I'm not sure I see your point because fixed timestamps complicate UX slightly: it requires the player does some math. If authenticity is a priority, maybe Candlestick charts are the better default. But what do you think? Between "stock market authenticity" and simplified UX is there a clear priority?

d0sboots:
> > Other changes: I think it makes more sense to have the time axis be "time" rather than "time since data point." That lines up with how traditional stock displays work, and when the data updates the label on the point won't change.
> 
> I'm not sure I see your point because fixed timestamps complicate UX slightly: it requires the player does some math. If authenticity is a priority, maybe Candlestick charts are the better default. But what do you think? Between "stock market authenticity" and simplified UX is there a clear priority?

This suggestion was actually all about UX, not about authenticity. 

If labels are set as "ticks/time since now," then they change on *every* update. Looking at any given piece of data requires constant math to understand where it stands, since the labels keep changing. 

This is why it's good to give each data point a fixed reference, such as the time it occurred. "Ticks since game start" would also work, but it's less natural and (as you say) would involve mental math.

PR #3029, opened 2026-08-03
FIX: Preserve script editor state when switching tabs

AGuyWhoSkis:
Leaving the Script Editor for another tab and coming back used to reset the editor. This PR fixes several symptoms that come from the same bug, plus a semi-related focus issue.

### What was wrong

Undo/redo history in Script Editor was wiped after switching tabs in-game.
Collapsed (folded) code sections sprang back open.

### Why

Switching tabs fully unmounts the editor, and its cleanup disposed every Monaco model - both the open scripts and the ~380 KB of type definitions for the ns API and React. All that state lives in the models, so the next visit started from scratch and the type definitions had to be re-parsed from nothing.

The models were being disposed because the type definitions were re-created unconditionally on mount, and Monaco refuses to create a model for a file it already knows about. Clearing everything on the way out kept that from erroring.

### The fix

The type definitions are now created once and reused, so there's no longer a reason to throw the models away. Cleanup disposes only the editor itself; the models for open scripts stay alive until their tab is closed, which already disposed them.

Folding needed one more piece: it lives in the editor's view state, and only the cursor position was being saved. Each open script now remembers its full view state, so folds survive switching pages and switching between script tabs - the latter was broken too.

#### Bonus fixes

The first script opened in a session didn't get the text cursor; you had to click into the editor before typing. The editor is created inside a hidden container on that first open, and focusing something hidden does nothing. Focus is now applied once the editor is actually visible.

Using Script Editor to open files with JSDoc comments causes JSDoc-related errors to appear _briefly_ before they're resolved within ~1-2 seconds. This behavior remains, but now it's reduced because JSDoc is not being loaded every time the Script Editor is opened.

### Testing

Manually: fold a region and type in a script, switch to Terminal and back - folds, undo history, and cursor position are intact. nano a script on a fresh page load - the cursor is live immediately.

"BEFORE" - Script Editor behavior in Dev

before

"AFTER" - Script Editor behavior in this PR

after

### DIFF
--- a/src/ScriptEditor/ui/Editor.tsx
+++ b/src/ScriptEditor/ui/Editor.tsx
@@ -44,13 +44,16 @@ export function Editor({ onMount, onChange, onUnmount }: EditorProps) {
      * monaco editor has a bug that makes function definitions appear as duplicate ones. For more information, please
      * check: https://github.com/microsoft/monaco-editor/issues/3580 and https://github.com/microsoft/monaco-editor/pull/4544.
      */
-    monaco.editor.createModel(
-      netscriptDefinitions.replace(/^export /gm, ""),
-      "typescript",
-      monaco.Uri.file("netscript.d.ts"),
-    );
-    monaco.editor.createModel(reactTypes, "typescript", monaco.Uri.file("react.d.ts"));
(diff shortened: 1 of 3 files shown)


d0sboots:
This one is going to have to wait on @catloversg, because all changes to the script editor are subtle as hell.

PR #3032, opened 2026-08-05
CODEBASE: Use population helper functions in random Bladeburner events

### LINKED ISSUE #2994: Random Bladeburner city population increase not clamped
A city's synthoid population in Bladeburner can go past the MAX_SAFE_INTEGER in random events.

In Bladeburner, city population changes are normally handled by `changePopulationCount()` and `changePopulationByPercentage()`, which use `clampInteger()` on the result and limit the result to MAX_SAFE_INTEGER.
However, the population changes in `randomEvent()` don't use the functions mentioned above and directly modify a city's population. Example:

https://github.com/bitburner-official/bitburner-src/blob/257f477fbb71c5104577d23406580f4c4b6845e0/src/Bladeburner/Bladeburner.ts#L616-L618

This means that a city's population can randomly push past the integer limit, until a player action influences the population and it immediately gets clamped back down to the limit, potentially resulting in a huge population loss. Since all random events operate on percentages, it can't go into the negatives, but there's no upper limit.
Does it matter? Not much, but it can influence action success change unpredictably. Since a larger population positively affects bladeburner action success chance, you can lose a success factor in the size of several orders of magnitude after a single action that influences city population.

It would be pretty easy to let random event population changes be handled by `changePopulationCount()` and `changePopulationByPercentage()`, but there's a few things to consider.

- If we grow a population that's already at the limit, the user is still informed about a synthoid change that doesn't actually happen.
- Does the user just miss out on a random event's effects if the luck of the draw lands on a maxed-out city? Do they get a re-roll?
- When synthoids migrate, they effectively vanish from the source city and never arrive in the destination city, resulting in a net overall population decrease. This can also result in an "empty" synthoid community forming in the destination city.

gmcew:
closes #2994 
Potentially avoids exceeding MAX_SAFE_INTEGER in niche situations, also looks cleaner and is more consistent.

Tested by sitting and watching numbers go up and down for a while of bonus time; I've never seen the MAX_SAFE_INTEGER issue come up in dev or normal play, but this appears to behave as before for the typical ranges of population.

If there's a better way to test this I'm down!

### DIFF
--- a/src/Bladeburner/Bladeburner.ts
+++ b/src/Bladeburner/Bladeburner.ts
@@ -579,8 +579,8 @@ export class Bladeburner implements OperationTeam {
       ++destCity.comms;
     }
     const count = Math.round(sourceCity.pop * percentage);
-    sourceCity.pop -= count;
-    destCity.pop += count;
+    sourceCity.changePopulationByCount(-count);
+    destCity.changePopulationByCount(count);
     if (destCity.pop < BladeburnerConstants.PopGrowthCeiling) {
       destCity.pop += BladeburnerConstants.BasePopGrowth;
     }
@@ -615,7 +615,7 @@ export class Bladeburner implements OperationTeam {
       ++sourceCity.comms;
       const percentage = getRandomIntInclusive(10, 20) / 100;
       const count = Math.round(sourceCity.pop * percentage);
-      sourceCity.pop += count;
+      sourceCity.changePopulationByCount(count);
       if (sourceCity.pop < BladeburnerConstants.PopGrowthCeiling) {
         sourceCity.pop += BladeburnerConstants.BasePopGrowth;
       }
@@ -629,7 +629,7 @@ export class Bladeburner implements OperationTeam {
         ++sourceCity.comms;
         const percentage = getRandomIntInclusive(10, 20) / 100;
         const count = Math.round(sourceCity.pop * percentage);
-        sourceCity.pop += count;
+        sourceCity.changePopulationByCount(count);
         if (sourceCity.pop < BladeburnerConstants.PopGrowthCeiling) {
           sourceCity.pop += BladeburnerConstants.BasePopGrowth;
         }
@@ -643,8 +643,8 @@ export class Bladeburner implements OperationTeam {
         // Change pop
         const percentage = getRandomIntInclusive(10, 20) / 100;
         const count = Math.round(sourceCity.pop * percentage);
-        sourceCity.pop -= count;
-        destCity.pop += count;
+        sourceCity.changePopulationByCount(-count);
+        destCity.changePopulationByCount(count);
         if (destCity.pop < BladeburnerConstants.PopGrowthCeiling) {
           destCity.pop += BladeburnerConstants.BasePopGrowth;
         }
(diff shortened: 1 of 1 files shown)


src/Bladeburner/Bladeburner.ts
```diff
@@ -578,9 +578,15 @@ export class Bladeburner implements OperationTeam {
       --sourceCity.comms;
       ++destCity.comms;
     }
-    const count = Math.round(sourceCity.pop * percentage);
-    sourceCity.pop -= count;
-    destCity.pop += count;
+    // community based population changes logically require a non-zero population change
+    sourceCity.changePopulationByPercentage(-percentage, {
```
d0sboots: This doesn't do the same thing. (Here and elsewhere.)

Before, it was transferring population based on a percentage of source. Now, it is adjusting populations independently in a percentagewise fashion. The obvious example is if source is 0 and dest is large, before there would be no change, now dest grows.

gmcew: That was a stupid oversight, sorry. I'll fix this evening by using the population change by value method, and base that value on a calculated percentage as before

catloversg:
I run a Monte Carlo simulation to see how long until population reaches `Number.MAX_SAFE_INTEGER`.
Code:
```js
/** @param {NS} ns */
  const iters = 500;
  const results = [];

  for (let i = 0; i < iters; ++i) {
    results.push(foo(ns));
  }

  results.sort((a, b) => a - b);

  const total = results.reduce((sum, value) => sum + value, 0);
  const mean = total / iters;

  const median =
    iters % 2 === 0
      ? (results[iters / 2 - 1] + results[iters / 2]) / 2
      : results[Math.floor(iters / 2)];

  const variance = results.reduce((sum, value) => sum + (value - mean) ** 2, 0) / iters;
  const stddev = Math.sqrt(variance);
  const standardError = stddev / Math.sqrt(iters);

  console.log(`min:             ${results[0]}`);
  console.log(`max:             ${results[results.length - 1]}`);
  console.log(`mean:            ${mean}`);
  console.log(`median:          ${median}`);
  console.log(`stddev:          ${stddev}`);
  console.log(`standard error:  ${standardError}`);
}

/** @param {NS} ns */
function foo(ns) {
  for (const city of Object.values(ns.enums.CityName)) {
    Player.bladeburner.cities[city].pop = 1.25e9;
    Player.bladeburner.cities[city].popEst = Player.bladeburner.cities[city].pop;
    Player.bladeburner.cities[city].comms = 77;
    Player.bladeburner.cities[city].chaos = 0;
  }

  let events = 0;
  while (Player.bladeburner.cities["Sector-12"].pop <= Number.MAX_SAFE_INTEGER) {
    Player.bladeburner.randomEvent();
    ++events;
  }

  return events;
}
```
Output:
```
min:             6222
max:             12489
mean:            9005.896
median:          8958.5
stddev:          1027.9317765221583
standard error:  45.97050657071336
```
Based on this simulation, the expected time is roughly 9000 random events. An event happens every ~420 seconds, so this would take about 43.75 days of continuous runtime in the BN. Unless the player is running a long int farming session, they are unlikely to encounter this in normal gameplay. Therefore, I don't think the concerns raised in #2994 require a complicated solution. A simple fix using `changePopulationByCount` or `changePopulationByPercentage` should be sufficient.

Regarding the current fix, there are two issues. D0sboots pointed out the first one. Another issue is that `percentage`, which is calculated like this: `getRandomIntInclusive(foo, bar) / 100`, is a fraction, but `changePopulationByPercentage` expects a "percentage" value (5 instead of 0.05).

In the latest commit, I used `changePopulationByCount` instead of `changePopulationByPercentage`.

PR #3034, opened 2026-08-07
DNET: add lab depth and migration progress

ficocelliguy:
Player feedback tweaks to the values returned by dnet api methods:

- For dnet.getServerDetails(), return lab server depth as it appears in the UI, rather than its internal value of -1
- for dnet.induceServerMigration(), include the progress fraction in the response object, not just the script logging

### DIFF
--- a/src/NetscriptFunctions/Darknet.ts
+++ b/src/NetscriptFunctions/Darknet.ts
@@ -26,6 +26,7 @@ import {
   getLabyrinthLocationReport,
   getSurroundingsVisualized,
   isLabyrinthServer,
+  labData,
 } from "../DarkNet/effects/labyrinth";
 import { getPhishingAttackSpeed, handlePhishingAttack } from "../DarkNet/effects/phishing";
 import { handleRamBlockRemoved } from "../DarkNet/effects/ramblock";
@@ -401,7 +402,9 @@ export function NetscriptDarknet(): InternalAPI<DarknetAPI> {
       const localServer = ctx.workerScript.getServer();
       const isConnected = isDirectConnected(localServer, targetServer);
       const hasSession = isAuthenticated(targetServer, ctx.workerScript.pid);
-      const depth = isLabyrinthServer(targetServer.hostname) ? getLabyrinthDetails().depth : targetServer.depth;
+      const depth = isLabyrinthServer(targetServer.hostname)
+        ? labData[targetServer.hostname].depth
+        : targetServer.depth;
       return {
         isOnline: true,
         isConnectedToCurrentServer: isConnected,
@@ -419,8 +422,9 @@ export function NetscriptDarknet(): InternalAPI<DarknetAPI> {
         isStationary: targetServer.isStationary,
       } satisfies ReturnType<DarknetAPI["getServerDetails"]>;
     },
-    induceServerMigration: (ctx, _host): Promise<DarknetResult & { progress?: number }> => {
+    induceServerMigration: (ctx, _host): Promise<DarknetResult & { progress: number }> => {
       const targetHost = helpers.string(ctx, "host", _host);
+      const currentProgress = DarknetState.migrationInductionServers.get(targetHost) ?? 0;
       const serverCheck = checkDarknetServer(ctx, targetHost, {
         requireDirectConnection: true,
         preventUseOnStationaryServers: true,
@@ -430,6 +434,7 @@ export function NetscriptDarknet(): InternalAPI<DarknetAPI> {
           success: false,
           code: serverCheck.code,
           message: serverCheck.message,
+          progress: currentProgress,
         }));
       }
       const hostOfCurrentServer = !isIPAddress(targetHost)
@@ -442,6 +447,7 @@ export function NetscriptDarknet(): InternalAPI<DarknetAPI> {
           success: false,
           code: ResponseCodeEnum.DirectConnectionRequired,
           message: message,
+          progress: currentProgress,
         }));
       }
       const server = serverCheck.server;
@@ -458,6 +464,7 @@ export function NetscriptDarknet(): InternalAPI<DarknetAPI> {
             success: false,
             code: serverCheck.code,
             message: serverCheck.message,
+            progress: DarknetState.migrationInductionServers.get(targetHost) ?? 0,
           }));
         }
         const server = serverCheck.server;
@@ -710,7 +717,7 @@ export const getDarknetPropertiesForDeprecationSupport = (dnetServer: DarknetSer
   depth: {
     identifier: "ns.getServer().depth",
     message: "Use ns.dnet.getServerDetails().depth instead.",
-    value: dnetServer.depth,
+    value: isLabyrinthServer(dnetServer.hostname) ? labData[dnetServer.hostname].depth : dnetServer.depth,
   },
   modelId: {
     identifier: "ns.getServer().modelId",
--- a/src/NetscriptFunctions/Darknet.ts
+++ b/src/NetscriptFunctions/Darknet.ts
@@ -401,6 +401,7 @@ export function NetscriptDarknet(): InternalAPI<DarknetAPI> {
       const localServer = ctx.workerScript.getServer();
       const isConnected = isDirectConnected(localServer, targetServer);
       const hasSession = isAuthenticated(targetServer, ctx.workerScript.pid);
+      const depth = isLabyrinthServer(targetServer.hostname) ? getLabyrinthDetails().depth : targetServer.depth;
       return {
         isOnline: true,
         isConnectedToCurrentServer: isConnected,
@@ -414,11 +415,11 @@ export function NetscriptDarknet(): InternalAPI<DarknetAPI> {
         blockedRam: targetServer.blockedRam,
         difficulty: targetServer.difficulty,
         requiredCharismaSkill: targetServer.requiredCharismaSkill,
-        depth: targetServer.depth,
+        depth: depth,
         isStationary: targetServer.isStationary,
       } satisfies ReturnType<DarknetAPI["getServerDetails"]>;
     },
-    induceServerMigration: (ctx, _host): Promise<DarknetResult> => {
+    induceServerMigration: (ctx, _host): Promise<DarknetResult & { progress?: number }> => {
       const targetHost = helpers.string(ctx, "host", _host);
       const serverCheck = checkDarknetServer(ctx, targetHost, {
         requireDirectConnection: true,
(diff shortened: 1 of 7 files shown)


src/NetscriptFunctions/Darknet.ts
```diff
@@ -401,6 +401,7 @@ export function NetscriptDarknet(): InternalAPI<DarknetAPI> {
       const localServer = ctx.workerScript.getServer();
       const isConnected = isDirectConnected(localServer, targetServer);
       const hasSession = isAuthenticated(targetServer, ctx.workerScript.pid);
+      const depth = isLabyrinthServer(targetServer.hostname) ? getLabyrinthDetails().depth : targetServer.depth;
```
catloversg: I don't think this is correct.
```js
/** @param {NS} ns */
  console.log(ns.dnet.getServerDetails("th3_l4byr1nth").depth);
  console.log(ns.dnet.getServerDetails("cru3l_l4byr1nth").depth);
  console.log(ns.dnet.getServerDetails("m3rc1l3ss_l4byr1nth").depth);
  console.log(ns.dnet.getServerDetails("ub3r_l4byr1nth").depth);
  console.log(ns.dnet.getServerDetails("et3rn4l_l4byr1nth").depth);
  console.log(ns.dnet.getServerDetails("end13ss_l4byr1nth").depth);
  console.log(ns.dnet.getServerDetails("f1n4l_l4byr1nth").depth);
  console.log(ns.dnet.getServerDetails("b0nus_l4byr1nth").depth);
}
```
If you run this code without SF15, it always prints 5. With SF15 but no "completed" lab, it prints 7. This means the hostname does not matter; the API always returns the depth of the current lab, and if there is no accessible lab (no SF15), it returns a default value (5).

Using `getLabyrinthDetails` in internal code is usually okay because callers only care about the current lab, but with this API, the caller (i.e., the player) wants the detail of a specific lab, not the current lab.

I think you should use `labData[targetServer.hostname].depth`.

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -4745,9 +4745,9 @@ export interface Darknet {
    * RAM cost: 4 GB
    *
    * @param host - Hostname/IP of the connected server to migrate.
-   * @returns A promise that resolves to a {@link DarknetResult} object.
+   * @returns A promise that resolves to a {@link DarknetResult} object with an additional `progress` field indicating the migration progress between 0 and 1.
    */
-  induceServerMigration(host: string): Promise<DarknetResult>;
+  induceServerMigration(host: string): Promise<DarknetResult & { progress?: number }>;
```
catloversg: In `heartbleed`, `logs` is not optional; you specify an empty array for `logs` when the call fails. To be consistent, `progress` should not be optional. It would be 0 when the call fails.

If you want to say that `progress` exists when `DarknetResult.success` is true, then you need to use a discriminated union like this:
```ts
export type DarknetMigrationResult =
  | (DarknetResult & { success: true; progress: number })
  | (DarknetResult & { success: false });
induceServerMigration(host: string): Promise<DarknetMigrationResult>;
```

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -4745,9 +4745,9 @@ export interface Darknet {
    * RAM cost: 4 GB
    *
    * @param host - Hostname/IP of the connected server to migrate.
-   * @returns A promise that resolves to a {@link DarknetResult} object.
+   * @returns A promise that resolves to a {@link DarknetResult} object with an additional `progress` field indicating the migration progress between 0 and 1.
```
```suggestion
   * @returns A promise that resolves to a {@link DarknetResult} object with an additional `progress` field indicating the migration progress in the range [0, 1].
```
Using `[0, 1]` or `(0, 1]` depends on whether `progress` is optional (see my other comment).

src/NetscriptFunctions/Darknet.ts
d0sboots: I think the code here is correct, but you have exposed a more fundamental bug in Darknet.

```
[home /]> run eval.js "ns.tprint(ns.getServer('th3_l4byr1nth'), ns.getServer('b0nus_l4byr1nth'))"
Running script with 1 thread, pid 49 and args: ["ns.tprint(ns.getServer('th3_l4byr1nth'), ns.getServer('b0nus_l4byr1nth'))"].
eval.js: {"hostname":"th3_l4byr1nth","ip":"9.67.250.37","sshPortOpen":false,"ftpPortOpen":false,"smtpPortOpen":false,"httpPortOpen":false,"sqlPortOpen":false,"hasAdminRights":false,"cpuCores":1,"isConnectedTo":false,"ramUsed":0,"maxRam":128,"organizationName":"","purchasedByPlayer":false,"backdoorInstalled":false,"isOnline":true,"depth":-1,"modelId":"(The Labyrinth)","hasStasisLink":false,"blockedRam":0,"staticPasswordHint":"You have discovered a dark, mysterious maze. Your footsteps echo eerily in the silence.","passwordHintData":"","difficulty":10,"requiredCharismaSkill":300,"logTrafficInterval":9007199254740991,"isStationary":true}{"hostname":"b0nus_l4byr1nth","ip":"201.161.44.245","sshPortOpen":false,"ftpPortOpen":false,"smtpPortOpen":false,"httpPortOpen":false,"sqlPortOpen":false,"hasAdminRights":false,"cpuCores":1,"isConnectedTo":false,"ramUsed":0,"maxRam":128,"organizationName":"","purchasedByPlayer":false,"backdoorInstalled":false,"isOnline":true,"depth":-1,"modelId":"(The Labyrinth)","hasStasisLink":false,"blockedRam":0,"staticPasswordHint":"You have discovered a dark, mysterious maze. Your footsteps echo eerily in the silence.","passwordHintData":"","difficulty":10,"requiredCharismaSkill":3200,"logTrafficInterval":9007199254740991,"isStationary":true}
```

There can only be one labyrinth per node instance. The fact that you can access them all indicates that they have been improperly added to AllServers - only the one that exists should be there.

src/ScriptEditor/NetscriptDefinitions.d.ts
d0sboots: Even though it's a lot, I'd be inclined to inline the type instead of defining a one-off `DarknetMigrationResult`, because of the annoyance of clicking through to see the type. But only if our tools (intellisense etc.) can properly parse it.

src/NetscriptFunctions/Darknet.ts
d0sboots: You need to update getDarknetPropertiesForDeprecationSupport too.

d0sboots:
This is an API-breaking change, although one that I don't expect to cause much fuss. You should add an informational message on `getServerDetails` (I'm 95% sure that string is unique by itself) about the change; sadly I don't think there's any way to restrict this to usages of depth. Technically this is also breaking for the deprecated .depth accessor in `getServer` as well, but IMO warning on that would be too spammy and exceed its purpose, and you're already warned via the deprecation notice.

src/NetscriptFunctions/Darknet.ts
catloversg: When I see all lab servers appearing in the server list at the same time, I have always thought it was a weird design.

@ficocelliguy I guess it's not intentional?

ficocelliguy: It is intentional that all the labs exist. It's following the same pattern of darkweb and w0rld_d34m0n - they're always there and only connected situationally. I assumed there was good reason to have key servers set up that way, I suppose, but it doesn't have to be

ficocelliguy: I've updated this to grab the details of the lab they ask for by name.

src/ScriptEditor/NetscriptDefinitions.d.ts
ficocelliguy: I updated this to make progress non-optional

src/NetscriptFunctions/Darknet.ts
d0sboots: If you keep all the lab servers in AllServers, then you have to make a compromise somewhere else that is unwanted. 

darkweb and w0rld_d34m0n are treated specially because they can be dynamically added. Long before pservers and all of that, AllServers was static so that's how it was handled. Also, certain properties of w0rld_d34m0n are set up right at the start of the node, so it would require reorganization to do it dynamically. 

However, both of those are servers that *can* exist within a particular install. Only one of the labs can exist (and it always exists), so only one should be added. This maintains the vital invariant that `getServer` returns objects for existing servers only (and also tombstones for dead dnet servers).

ficocelliguy: should getServerDetails() return the lab details for the current lab when given ANY lab's hostname, or should it return an offline server object for any labs that don't currently exist? Being able to consistently get the lab's current depth by querying the details of `th3_l4byr1nth` seems useful, but it's a slight misrepresentation.

catloversg: If the player requests the details of a nonexistent lab server (one that does not exist in `AllServers`), the API should throw an error. We return the offline server object because normal dnet servers can go offline at any time while scripts are still running. However, lab servers do not suddenly go offline. They only "disappear" when the player prestiges (another lab server is connected to dnet instead of it).

Ideally, I think we should add only the relevant lab server instead of all lab servers to `AllServers`. This would require a save data migration. I can handle that for you if you want.

ficocelliguy: This branch has been re-tooled to only insert the current lab into the network. I don't think we need an explicit save migration, as the change to only include the current lab in AllServers will be applied anyway on the next aug install or bitnode completion.

d0sboots: Agreed that we don't need a migration, this will be cleaned up on next install and it's ok to be (the existing state of) slightly weird until then.

d0sboots:
Looks good, just waiting on the last comment:
> This is an API-breaking change...

(In case it wasn't clear, I meant add a message via the api-breaks infra)

ficocelliguy:
I added a message to `breakingChanges302` indicating the change to the .depth value for lab servers.

src/utils/APIBreaks/3.0.2.ts
```diff
@@ -17,5 +17,14 @@ export const breakingChanges302: VersionBreakingChange = {
         "Use ns.singularity.hasExportGameBonus() instead.",
       showWarning: false,
     },
+    {
+      brokenAPIs: [
+        {
+          name: "ns.dnet.getServerDetails",
```
d0sboots:
```suggestion
          name: "getServerDetails",
```
I believe this string is unique, so it's better to cast a wider net (for people who assign namespaces to variables, etc.)

PR #3037, opened 2026-08-08
BUGFIX: Preserve generated logs during authentication

### LINKED ISSUE #3030: ns.dnet.authenticate() prevents random log entries from spawning
In #3026 @ficocelliguy claimed that random garbage log entries can be added when you authenticate via the API, but the API appears to do the opposite.

## To reproduce
Note: do not open the server details modal while trying to reproduce this bug.
1. Enter bitnide 15
1. Run this on home:
   ```ts
     ns.clearLog();
     ns.ui.openTail();
     const host = 'darkweb';
     const { logTrafficInterval } = ns.dnet.getServerDetails(host);
     // Discard logs that may have been generated previously.
     await ns.dnet.heartbleed(host, { logsToCapture: 1000 });
   
     // Sleep for twice the logTrafficInterval.
     // We expect 2 or 3 random logs to be generated while we sleep.
     await ns.asleep(2000 * logTrafficInterval);
     // This will generate a log, containig the string 'passwordAttempted'.
     await ns.dnet.authenticate(host, '');
     // One more random log might be generated after authenticate()
     // adds its log entry and before heartbleed() retrieves the logs.
     const { logs } = await ns.dnet.heartbleed(host, { logsToCapture: 1000 });
   
     const ramdom = logs.filter(f =>
       !f.includes('passwordAttempted')
     ).length;
     if (ramdom >= 2) {
       ns.print(`SUCCESS got ${ramdom} random logs`);
     } else {
       ns.print(`ERROR got ${ramdom} random logs`);
     }
     ns.print('logs:');
     ns.print(logs.join('\n'));
   }
   ```

## Expected result
Two or three random logs are generated

## Actual result
Usually no random logs are generated. I suspect that in 6653b8d38 up to one random log could be generated because it could be generated after `ns.dnet.authenticate()` is done.

## Notes
I think this is the relevant code. It looks like `logPasswordAttempt()` calls `populateServerLogsWithNoise()` and then discards the logs it generates.

   ```ts
     const serverState = getServerState(server.hostname);
     const serverLogs = serverState.serverLogs;
     populateServerLogsWithNoise(server);
     // [...]
     serverState.serverLogs = [logMessage, ...serverLogs].slice(0, MAX_LOG_LINES);
   };
   // [...]
   export const populateServerLogsWithNoise = (server: DarknetServer) => {
     // [...]
       serverState.serverLogs = [...noiseArray, ...serverState.serverLogs].slice(0, MAX_LOG_LINES);
     // [...]
   };
   ```

## I reproduced this bug in the following game versions:
v3.0.1 (3162fd2): ERROR got 0 random logs
v3.0.2 (f34116c): ERROR got 0 random logs
the dev branch (6653b8d38) - a version that includes that fix from #3026: ERROR got 0 random logs

BreezeDelegate:
Fixes #3030.

Authentication logs no longer discard noise logs generated immediately before them.

### DIFF
--- a/test/jest/Darknet/PacketSniffing.test.ts
+++ b/test/jest/Darknet/PacketSniffing.test.ts
@@ -1,6 +1,7 @@
-import { getDefaultPasswordConfig, serverFactory } from "../../../src/DarkNet/controllers/ServerGenerator";
 import { getAuthResult } from "../../../src/DarkNet/effects/authentication";
 import { getServerState } from "../../../src/DarkNet/models/DarknetState";
+import { SpecialServers } from "../../../src/Server/data/SpecialServers";
+import { GetServerOrThrow } from "../../../src/Server/AllServers";
 import { initGameEnvironment, setupBasicTestingEnvironment } from "../Utilities";
 
 beforeAll(() => {
@@ -18,7 +19,7 @@ afterEach(() => {
 test("authentication preserves generated server logs", () => {
   jest.spyOn(Math, "random").mockReturnValue(0.99);
 
-  const server = serverFactory(getDefaultPasswordConfig, 1, 0, 0);
+  const server = GetServerOrThrow(SpecialServers.DarkWeb);
   server.logTrafficInterval = 1;
 
   const serverState = getServerState(server.hostname);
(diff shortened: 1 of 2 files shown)


test/jest/Darknet/PacketSniffing.test.ts
```diff
@@ -0,0 +1,31 @@
...
+
+beforeEach(() => {
+  setupBasicTestingEnvironment();
+});
+
+afterEach(() => {
+  jest.restoreAllMocks();
+});
+
+test("authentication preserves generated server logs", () => {
```
catloversg, line 19: Optional: Add this test to `test/jest/Darknet/Darknet.test.ts` instead of a new file and run the test on `darkweb` so you don't need to create a testing server.

PR #3046, opened 2026-08-11
MISC: Create issue templates

E9cipher:
This avoids the "help" titled issues and related stuff. Small commits because GitHub Mobile doesn't let me batch them all.

As you see, more templates can come. GitHub docs recommends a feature request, bug and security vulnerability. I'd say only the first two apply here. 

I will also create a config.yml file to disable blank issues. Instead, issuers will select the "blank issue" template, which contains basic instructions.

This may also require a collaborator/owner to select it in settings to be used. I'm not sure about this, though (gh docs say it applies when it's merged, but I had to do some setup on my fork other than allowing issues)

You can test this on my (wonderful) [blackjack project](https://github.com/E9cipher/blackjack)! For limited time only :)

I'll finish wiring things up tomorrow.


catloversg:
I'll do this myself.

E9cipher:
Well, now that this is here, you could just review it and share your thoughts...

PR #3047, opened 2026-08-12
MISC: Follow up to #3037

catloversg:
The fix in #3037 is correct, but there are many things in the current codebase that I want to change.

### Reassign serverState.serverLogs
I notice Fico tends to create a new array and reassign it every time they want to change the array contents. For example, when pushing a new item at the start of the array, they usually do this: `arr = [item, ...arr]` instead of `arr.unshift(item)`.

`logPasswordAttempt` was implemented like this:
```ts
  const serverState = getServerState(server.hostname);
  const serverLogs = serverState.serverLogs;
  populateServerLogsWithNoise(server);
  // Truncated code
  serverState.serverLogs = [logMessage, ...serverLogs].slice(0, MAX_LOG_LINES);
};
```
Capturing `serverState.serverLogs` before calling `populateServerLogsWithNoise` is arguably questionable, but I don't think it's a bug per se. `populateServerLogsWithNoise` is supposed to insert log noise to `serverState.serverLogs`, so this code still makes sense to me. It creates a bug because `populateServerLogsWithNoise` reassigns a new array to serverState.serverLogs.

At the end of `logPasswordAttempt`, we see this pattern again. Suppose we have:
```ts
const serverLogs = getServerState(server.hostname).serverLogs;
logPasswordAttempt(server, passwordResponse, pid);
foo(serverLogs);
```
At first glance, it seems foo would receive a log array that contains the password authentication log and noise logs, but it actually does not. In other words, with this pattern, all users of `serverState.serverLogs` must ensure that they always get `serverState.serverLogs` again after calling any functions that potentially change `serverState.serverLogs`.

I don't think this is a good approach. I refactored all code that followed this pattern, changing it to mutate the array instead of reassigning a new one.

### Wrong order of noise logs
`serverState.serverLogs` contains logs that are ordered from newest to oldest, but the noise logs are generated in the reversed order.

Another minor issue is that `noiseArray` is unnecessarily long if the gap between now and the last log time is very long.

### Duplicate code in heartbleed
That function can use `getServerLogs`.

### Bug in the `authentication preserves generated server logs` test
That test incorrectly uses `GetServerOrThrow` instead of `getDarknetServerOrThrow`. This type of error will be caught by our CI after I change the lint task.

### Other things for discussion
`deleteDarknetServer` is implemented like this:
```ts
export const deleteDarknetServer = (server: DarknetServer, force = false): void => {
  // truncated code
  DeleteServer(server.hostname);
  const serverState = getServerState(server.hostname);
  serverState.authenticatedPIDs = [];
  serverState.serverLogs.length = 0;
};
```
This looks questionable to me. If the server was deleted, I think we should remove the corresponding entry from `DarknetState.serverState`. This block does not reset `lastLogTime`, so it can cause subtle issues.

@d0sboots What do you think about it?


ficocelliguy (approved):

d0sboots:
I agree with the idea that we should be deleting from ServerState as well, unless there's something even more subtle going on.

src/DarkNet/models/packetSniffing.ts
```diff
@@ -122,33 +121,38 @@ export const logPasswordAttempt = (server: DarknetServer, passwordResponse: Pass
...
+  const now = Date.now();
   if (!serverState.lastLogTime) {
-    serverState.serverLogs = [
-      getLogNoise(server, new Date(new Date().getTime() - interval * 1000)),
-      getLogNoise(server, new Date(new Date().getTime() - interval * 2000)),
-    ];
-    serverState.lastLogTime = new Date();
+    serverState.serverLogs.length = 0;
+    serverState.serverLogs.push(
+      getLogNoise(server, new Date(now - interval * 1000)),
```
d0sboots, line 137: These probably should just be timestamps instead of whole Date objects, but that's not for this fix.

src/DarkNet/models/packetSniffing.ts
```diff
@@ -122,33 +121,38 @@ export const logPasswordAttempt = (server: DarknetServer, passwordResponse: Pass
...
-    serverState.lastLogTime = new Date(lastLogTime + missingLogs * interval * 1000);
+  const lastLogTime = serverState.lastLogTime.getTime();
+  const totalMissingLogs = Math.floor((now - lastLogTime) / (interval * 1000));
+  if (totalMissingLogs > 0) {
+    const missingLogs = Math.min(totalMissingLogs, MAX_LOG_LINES);
+    const latestLogTimeMs = lastLogTime + totalMissingLogs * interval * 1000;
+    const noiseArray = Array.from({ length: missingLogs }, (_, i) =>
+      getLogNoise(server, new Date(latestLogTimeMs - interval * 1000 * i)),
+    );
+    serverState.serverLogs.unshift(...noiseArray);
```
d0sboots, line 152: The server logs are really stored backwards, with the newest data at the front?!

catloversg: Yes. What do you think about reversing the order? The current design is fairly weird because we have to call `unshift` every time we add a new log entry.

d0sboots: Seems like a reasonable refactor. Hopefully no save implications...

PR #3053, opened 2026-08-15
TOOLS: Add type checking task

catloversg:
Context: https://github.com/bitburner-official/bitburner-src/pull/2948#discussion_r3623215222 and https://github.com/bitburner-official/bitburner-src/pull/2971#issuecomment-5049658122.


package.json
```diff
@@ -135,8 +135,9 @@
     "start:dev": "webpack serve --progress --mode development",
     "build": "bash ./tools/build.sh production",
     "build:dev": "bash ./tools/build.sh development",
-    "lint": "eslint --fix --ext js,mjs,jsx,ts,mts,tsx --max-warnings 0 src test",
-    "lint:report": "eslint --ext js,mjs,jsx,ts,mts,tsx --max-warnings 0 src test",
+    "typecheck": "tsc --noEmit && tsc --noEmit -p test/tsconfig.json",
+    "lint": "eslint --fix --ext js,mjs,jsx,ts,mts,tsx --max-warnings 0 src test && npm run typecheck",
```
d0sboots: This might just be my personal preference, but: For me, lint is one of the slowest tasks (not as slow as tests, but slower than prettier). I would rather have something I can run in parallel (which is what `typecheck` is already) and have it be checked in parallel via a new git task.

package.json
```diff
@@ -135,8 +135,9 @@
     "start:dev": "webpack serve --progress --mode development",
     "build": "bash ./tools/build.sh production",
     "build:dev": "bash ./tools/build.sh development",
-    "lint": "eslint --fix --ext js,mjs,jsx,ts,mts,tsx --max-warnings 0 src test",
-    "lint:report": "eslint --ext js,mjs,jsx,ts,mts,tsx --max-warnings 0 src test",
+    "typecheck": "tsc --noEmit && tsc --noEmit -p test/tsconfig.json",
```
d0sboots: Minor thing, but in my normal dev flow, typecheck for the regular code is covered by running the dev server. So, for faster turnarounds it would be good to check the tests first.

PR #3056, opened 2026-08-17
DNET: Removed darknet stock volatility influence from stock forecast

ficocelliguy:
Currently, the rolled volatility of a stock is influenced by darknet promoteStock, which was intended. However, that rolled volatility influences the stocks' forecast afterwards (including the influence from promoteStock) which is not intended. This PR removes the influence of darknet from the forecast updates.

I am opening this PR primarily as a conversation starter; @catloversg originally identified there was an issue and planned to think on it, but AFAIK has not yet made any changes to it.

Some more context: https://discord.com/channels/415207508303544321/415207923506216971/1526139835478179870

### DIFF
--- a/src/StockMarket/StockMarket.ts
+++ b/src/StockMarket/StockMarket.ts
@@ -265,8 +265,8 @@ export function processStockPrices(numCycles = 1): void {
   for (const name of Object.keys(StockMarket)) {
     const stock = StockMarket[name];
     if (!(stock instanceof Stock)) continue;
-    const volatility = stock.mv * getDarknetVolatilityMult(stock.symbol);
-    let av = (v * volatility) / 100;
+    const maxVolatility = stock.mv * getDarknetVolatilityMult(stock.symbol);
+    let av = (v * maxVolatility) / 100;
     if (isNaN(av)) {
       av = 0.02;
     }
@@ -304,7 +304,9 @@ export function processStockPrices(numCycles = 1): void {
       processOrders(stock, OrderType.StopSell, PositionType.Long, processOrderRefs);
     }
 
-    let otlkMagChange = stock.otlkMag * av;
+    // Remove darknet influence from stock volatility, so darknet only affects current price, and not the forecast
+    const normalizedVolatility = av / getDarknetVolatilityMult(stock.symbol);
+    let otlkMagChange = stock.otlkMag * normalizedVolatility;
     if (stock.otlkMag < 5) {
       if (stock.otlkMag <= 1) {
         otlkMagChange = 1;


catloversg:
This change assumes the issue is that dnet stock promotion unintentionally affects stock forecast, and it's still okay to change the volatility that affects stock prices. However, as I pointed out on Discord, the issue is that volatility is not supposed to change at all. That factor determines how fast a stock price can change in a given time period.

First, let's see how dnet stock promotion can affect volatility.

The easiest way to experience this is by importing the save file RandomRogue attached on Discord. After importing, buy 4S UI data access, and observe the volatility of Four Sigma. You will see it skyrocket from 1.08% to 4.3% in 2 minutes. 4.3 / 1.08 is ~3.98.

Dnet effect is calculated like this:
```js
export const getDarknetVolatilityMult = (symbol: string) => {
  const charges = DarknetState.stockPromotions[symbol] ?? 0;
  const growthRate = 0.001;
  return 1 + (1 - Math.exp(-growthRate * charges) + 2 * (1 - Math.exp(-growthRate * 0.15 * charges)));
};
```
Desmos link: https://www.desmos.com/calculator/wuaeqewmho

The mult value approaches the softcap value (4) very fast.

Let's test how fast we can gain `charges`:
```js
// a.js
/** @param {NS} ns */
  Player.exp.charisma = 0;
  Player.gainCharismaExp(1e6);
  GetServer("darkweb").maxRam = 1024;
  const freeRAM = ns.getServerMaxRam() - ns.getServerUsedRam();
  console.log(freeRAM)
  ns.run("b.js", { threads: Math.floor(freeRAM / ns.getScriptRam("b.js")) });
}

// b.js
/** @param {NS} ns */
  DarknetState.stockPromotions.FSIG = 0;
  while (1) {
    await ns.dnet.promoteStock("FSIG");
    console.log(DarknetState.stockPromotions.FSIG);
  }
}
```
Getting 20000 charges (mult = 3.9) takes less than 3 minutes with this set up.

Conclusion: Getting 4x volatility is very easy with dnet.

Second, how important is volatility?

The volatility of a stock is randomized on initialization, and it never changes until prestige. On average, the volatility is 1.0454545454545454 (this is a percentage value; 1.045 means 1.045%).
```js
/** @param {NS} ns */
  const avgMv =
    InitStockMetadata.reduce((sum, stock) => {
      if (typeof stock.mv.min !== "number"
        || typeof stock.mv.max !== "number"
        || typeof stock.mv.divisor !== "number") {
        throw new Error();
      }
      const mv = (stock.mv.min + stock.mv.max) / 2 / stock.mv.divisor;
      return sum + mv;
    }, 0) / InitStockMetadata.length;
  console.log(avgMv)
}
```
Now let's run my main test code (you need to apply my attached patch):
Patch:
[test stock market.patch](https://github.com/user-attachments/files/31147199/test.stock.market.patch)
```js
/** @param {NS} ns */
  console.clear();
  StockMarket.storedCycles = 1e111;
  globalThis.disableFlip = true;
  const cycleMult = 2;
  const a = foo(ns, cycleMult, 0, 1.0454545454545454);
  const b = foo(ns, cycleMult, 40000, 1.0454545454545454 * 4);
  console.log(b / a);
}

function foo(ns, cycleMult, charge, constMv) {
  let avg = 0;
  const iters = 1000;
  for (let i = 0; i < iters; ++i) {
    const price = bar(ns, cycleMult, charge, constMv);
    // console.log(`final price: ${price}`);
    avg += price;
  }
  console.log(`avg: ${(avg / iters).toExponential()}`);
  return avg;
}

function bar(ns, cycleMult, charge, constMv) {
  globalThis.constMv = constMv;
  DarknetState.stockPromotions.FSIG = charge;
  StockMarket['Four Sigma'].b = true;
  StockMarket['Four Sigma'].price = 4605359.323000276;
  StockMarket['Four Sigma'].lastPrice = 4559003.692339174;
  StockMarket['Four Sigma'].otlkMag = 47.293385817073876;
  StockMarket['Four Sigma'].otlkMagForecast = 98.45794720733174;
  StockMarket['Four Sigma'].playerAvgPx = 2979632.86517832;
  StockMarket['Four Sigma'].shareTxForMovement = 44156;
  StockMarket['Four Sigma'].shareTxUntilMovement = 33305;
  for (let i = 0; i < 75 * cycleMult; ++i) {
    // processStockPrices(1);
    // processStockPricesPR(1);
    processStockPricesConstMv(1);
    // console.log(`price: ${ns.stock.getPrice("FSIG")}`);
  }
  return ns.stock.getPrice("FSIG");
}
```
With cycleMult = 1 (75 cycles), the buff is ~2.8.
With cycleMult = 2 (150 cycles; a flip is rolled every 75 cycles, so let's assume we're lucky and the flip does not happen after the first 75 cycles), the buff is ~7.7.

As we can see here, although the absolute change in volatility looks very small (3 percentage points, from 1% to 4%), the relative change is huge (300%), and it makes the stock price increase much faster than intended.

The general consensus on Discord is that the stock market lets you double your money every hour (I have not verified this claim). With this bug, players can potentially make several times as much profit over the same period. In effect, this is a massive increase in the rate at which players can make money from the stock market.

In general, I don't think changing volatility is a good approach for influencing the stock market through any mechanic (not just dnet). In theory, we can tune `getDarknetVolatilityMult`, but I think we should find a different approach instead. Volatility is like NFG. It looks very small, but when compounded over many cycles, it gets out of control really fast. I don't have a good suggestion right now (that's why I have not rebalanced this part), though. I'll think more about it after reviewing the save data format PRs.

d0sboots:
I agree with catlover: volatility was always meant to be static, and I overlooked that in the initial dnet change. Dnet boosting can potentially use a "temporary" volatility by doing movements that are scaled by an extra multiplier.

PR #3051, opened 2026-08-14
UI: Adding BN multipliers to sleeves augmentations modal

branchtheory:
EDIT: See [this comment](https://github.com/bitburner-official/bitburner-src/pull/3051#issuecomment-5539783109) for an an up-to-date account of what's changed in the final version. Everything in this current comment is out of date.

----

I've updated the "More Stats" modal.

Changed to:
1. include BN multipliers
2. have nice styling (nearly all of the new / updated code was copied from CharacterStats.tsx, with minor tweaks to make it work in the context of sleeves)
3. be all visible without any scrolling

On 2., some of this stuff is now duplicated. There are lots of minor differences though, which make me think it's not worthwhile to make CharacterStats.tsx and MoreStatsModal.tsx pull from a shared resource. An example is the tooltip, where a couple of word have changed ("your" > "this sleeve's"). I don't feel like it's worth the hassle to spin that out and have both files to pull from it.

On 3., the two column format looks crap, and is grouped differently to the player's stats page. But the advantage is you'll never have to scroll. I think it's the least-bad option.

Credit to armok the bunny, avatar of hate for the idea

# Before
image

With Bonus Time only visible if you scroll down a bit.

Image captured at 100% zoom, on my laptop screen, with no changes to the default text size etc.

# After

image

(slightly out of date -- I've corrected the gap to 1 (from 2) and increased the h6s to h5s)

# Alternatives in case people prefer them

## 2 columns compact

image

## 1 column, but where stuff can disappear off the bottom

image

image

## 1 column but less spacing so more stuff is visible on first glance

image

PR #3057, opened 2026-08-19
UI: Add price history chart to stock market

catloversg:
#3028 adds price history to the stock market UI. This kind of QoL feature usually needs to be considered case by case. On one hand, as a game, it's good to improve UI/UX. On the other hand, Bitburner has a unique design philosophy, so we sometimes intentionally do *not* add more QoL features or only add barely usable features. After considering the trade-offs, I'm fine with adding this feature, but I want to do it with a different approach.

For context, I will quote my own message about Bitburner's UI design philosophy in #3051:

In general, Bitburner UI is intentionally simple. We give players tools (NS APIs) and encourage them to build anything they want. This is not an excuse to have bad UI/UX (except some special cases in which we intentionally make the features bad enough so most players have to build their own custom solution, such as the server scan tool), so we will still try to improve the UI/UX. This means sometimes it's hard to determine whether a UI change is necessary, and it needs to be discussed case by case.

With this in mind, we need to answer these questions when designing the UI/UX for this feature:
- Do players actually need this feature? I.e., is the price history a mandatory feature or a purely cosmetic one? [1]
- If it's a mandatory feature, how good should the built-in implementation be? What kind of intentional restrictions should we put in place? [2]

There are two phases: pre-4S and post-4S. In pre-4S, players need to find a way to predict stock price movements. In order to do that, they need to record the price history and analyse it to estimate the hidden `otlkMag`. In post-4S, the hidden `otlkMag` can be queried directly, so players can easily find the most profitable stocks and decide whether to buy/sell each cycle. In other words, the stock market is practically a "solved" challenge in post-4S.

Therefore, the answer to question [1] is that, in pre-4S, the price history is a mandatory feature, but in post-4S, it's mostly a cosmetic feature.

For question [2], we need to consider:
- Should we lock the history access behind something such as the 4S data access?
- How many data points can the history contain?
- Can the history be queried by NS APIs?
- How easily can players check the history of a stock via the UI?
- Can players easily check the history of many stocks at the same time via the UI?

As I pointed out above, in post-4S, with access to the hidden `otlkMag`, the history is mostly a cosmetic feature. It's only useful in pre-4S. This is why I think we should let players access it right at the start.

The price history is not just a visualization tool; it's an important source of information that players need before they can analyze anything. Collecting information is very easy, but analyzing it is not. Naturally, players may think that they should analyze as many data points as possible. However, in this *game feature*, analyzing too many data points is actually counterproductive.

For example, let's say we have this timeline:
- Cycle 1: bull
- Cycle 2: bull
...
- Cycle 74: bull
- Cycle 75: flip bull/bear (bull => bear) and `otlkMagForecast`
- Cycle 76: bear
...
- Cycle 86: bear
- Cycle 96: bear

At cycle 96, if they use the data from the last 10 cycles, they may realize that the market is bear. With 20 cycles, they have a better chance of doing so. Naturally, they may think it's best to use as many collected data points as possible, potentially including all 96 cycles. However, if they do that, it's very likely that they will conclude that the current market is bull, which is wrong. Ideally, the history should contain *enough* data points to correctly infer the market state, but not too many points; otherwise, they will just infer the old and irrelevant market state (due to the flip cycle).

How many data points are *enough*? In theory, we can collect as many as possible (with a hard cap to avoid performance issues) and let players decide how many data points they want to look at. However, should we do that? This is where I want to mention the Bitburner's design philosophy again. Collecting data points is something players can and should do; it's part of their toolset for the stock market feature. We only need to provide a very basic chart so that players can use it to manually infer the market state. They will realize this is a mandatory feature (in pre-4S) and the built-in tool is very limited, so naturally they need to build their own tool.

Another important thing to note is that by observing a long history of many stocks, players can detect when the flip cycle happens without too much trouble. This is actually an advanced pre-4S technique. As shown above, if players can detect the flip around cycle 75 (with a small margin of error), they can collect a relatively long history to accurately guess the market state without worrying about the flip cycle. This is also why we randomize the first flip cycle. If the UI lets players access the long history of all stocks, they can guess the flip cycle too easily.

All things considered, I decided:
- Do *not* lock the history chart behind 4S data access.
- Only collect 10 data points.
- No NS APIs.
- Intentionally inconvenient UI: Players need to click on a stock, then click another button to view the history chart of that stock.
- No convenient way to view the history of all stocks at the same time.
- Nothing can be configured.

Other things to consider:
- Save the history in the save data. We only collect 10 data points, so it would not increase the save data size significantly. I think it's okay to do that, but I can change it if d0sboots wants.
- #3028 also fixes an edge case in `cycleForecast`. Currently, given the values of `changeAmt` used by `processStockPrices`, `otlkMag` cannot become greater than 50 after the `this.otlkMag < 0` block, so the fix is unnecessary with the current range of `changeAmt`. However, it preserves the `otlkMag` <= 50 invariant if `changeAmt` is ever increased significantly, so I incorporated it into this PR as well.

Screenshots:

### DIFF
--- a/src/StockMarket/Stock.ts
+++ b/src/StockMarket/Stock.ts
@@ -123,6 +123,8 @@ export class Stock {
    */
   readonly totalShares: number;
 
+  readonly priceHistory: { time: number; price: number }[];
+
   constructor(p: IConstructorParams = defaultConstructorParams) {
     this.name = p.name;
     this.symbol = p.symbol;
@@ -148,6 +150,8 @@ export class Stock {
     // Max Shares (Outstanding shares) is a percentage of total shares
     const outstandingSharePercentage = 0.2;
     this.maxShares = Math.round((this.totalShares * outstandingSharePercentage) / 1e5) * 1e5;
+
+    this.priceHistory = [];
   }
 
   /** Safely set the stock's second-order forecast to a new value */
@@ -188,11 +192,11 @@ export class Stock {
       this.otlkMag += changeAmt;
     }
 
-    this.otlkMag = Math.min(this.otlkMag, 50);
     if (this.otlkMag < 0) {
       this.otlkMag *= -1;
       this.b = !this.b;
     }
+    this.otlkMag = Math.min(this.otlkMag, 50);
   }
 
   /**
--- a/src/StockMarket/Stock.ts
+++ b/src/StockMarket/Stock.ts
@@ -123,7 +123,7 @@ export class Stock {
    */
   readonly totalShares: number;
 
-  readonly priceHistory: { time: number; price: number }[];
+  readonly priceHistory: { timeMs: number; price: number }[];
 
   constructor(p: IConstructorParams = defaultConstructorParams) {
     this.name = p.name;
--- a/src/StockMarket/ui/PriceHistoryChart.tsx
+++ b/src/StockMarket/ui/PriceHistoryChart.tsx
@@ -1,0 +1,94 @@
+import React from "react";
+import Typography from "@mui/material/Typography";
+import type { Stock } from "../Stock";
+import { useCycleRerender } from "../../ui/React/hooks";
+import { Settings } from "../../Settings/Settings";
+import { Money } from "../../ui/React/Money";
+
+export function PriceHistoryChart({ stock }: { stock: Stock }) {
+  const [hoveredDataPoint, setHoveredDataPoint] = React.useState<number | null>(null);
+
+  useCycleRerender();
+
+  const prices = stock.priceHistory.map((v) => v.price);
+  if (prices.length === 0) {
+    return null;
+  }
+
+  const width = 800;
+  const height = 500;
+  const padding = 20;
+
+  const minPrice = Math.min(...prices);
+  const maxPrice = Math.max(...prices);
+  const range = Math.max(maxPrice - minPrice, Number.EPSILON);
+
+  const points = stock.priceHistory.map((entry, index) => {
+    const x = padding + (index / Math.max(prices.length - 1, 1)) * (width - padding * 2);
+    const y = height - padding - ((entry.price - minPrice) / range) * (height - padding * 2);
+    return { x, y };
+  });
+
+  return (
+    <div style={{ padding: 10 }}>
+      <svg
+        viewBox={`0 0 ${width} ${height}`}
+        style={{
+          background: Settings.theme.backgroundprimary,
+          marginBottom: "10px",
+        }}
+      >
+        <polyline
+          points={points.map(({ x, y }) => `${x},${y}`).join(" ")}
+          fill="none"
+          stroke={Settings.theme.primary}
+          strokeWidth={2}
+        />
+
+        {/* Wide invisible hit area that snaps the cursor to the nearest data point. */}
+        <polyline
+          points={points.map(({ x, y }) => `${x},${y}`).join(" ")}
+          fill="none"
+          stroke="transparent"
+          strokeWidth={20}
+          onMouseMove={(event) => {
+            const svg = event.currentTarget.ownerSVGElement;
+            if (svg == null) {
+              return;
+            }
+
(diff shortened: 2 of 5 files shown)


d0sboots:
Before I get into the implementation, there's one part of your assumptions that I don't agree with:

> * Intentionally inconvenient UI: Players need to click on a stock, then click another button to view the history chart of that stock.
> * No convenient way to view the history of all stocks at the same time.

I don't think these are adding anything other than frustration. I think this is because I have a subtly different take on the Bitburner philosophy: 

We provide tools and UI that are fit for their purpose, but not necessarily fit for *all* purposes - that is what we have the NS APIs for. We sometimes intentionally limit the *tools* we provide (scan, the commandline), but we don't cripple the UI. This is because we expect our players to build their own tools, but not necessarily to build their own UIs.

In this specific case, what does the player gain/lose from being able to easily see multiple stocks at once? The primary thing is detection of the flip. IMO this is a concept that we *should* be making more visually obvious, instead of burying in docs. With a good UI, players have the potential to discover the primary mechanic of the market naturally, on their own.

catloversg:
New UI:

Wide screen:

Capture

Mobile:

mobile

The entire Stock Market UI now has a maximum width of 1200px. This is the maximum width used by many other pages.

Recently, there was a discussion on Discord about how inconsistent the UI is when it comes to horizontal space. I mentioned that pages with fixed content, such as the Stock Market, should have a consistent maximum width of 1200px. I remembered this discussion when I wondered how wide the chart could and should be. With a very limited number of data points, we don't need a very wide chart, so capping the width of the entire page at 1200px, as planned, makes sense here.

src/StockMarket/ui/PriceHistoryChart.tsx
```diff
@@ -0,0 +1,138 @@
...
+              <>
+                <div>{new Date(stock.priceHistory[hoveredDataPoint].time).toLocaleTimeString()}</div>
+                <div>
+                  <Money money={stock.priceHistory[hoveredDataPoint].price} />
+                </div>
+              </>
+            }
+            placement="right"
+          >
+            <span
```
d0sboots: All of this seems rather complicated. Can you not just give a tooltip to the SVG circle elements? If not, that train should live in a comment.

src/StockMarket/ui/PriceHistoryChart.tsx
```diff
@@ -0,0 +1,138 @@
...
+  const maxPrice = prices.length > 0 ? Math.max(...prices) : 0;
+  const range = Math.max(maxPrice - minPrice, Number.EPSILON);
+
+  const points = stock.priceHistory.map((entry, index) => {
+    const x = padding + (index / Math.max(prices.length - 1, 1)) * (width - padding * 2);
+    const y = height - padding - ((entry.price - minPrice) / range) * (height - padding * 2);
+    return { x, y };
+  });
+
+  const hoveredPoint = hoveredDataPoint !== null ? points[hoveredDataPoint] : null;
```
d0sboots: If I'm not mistaken, when the data updates this means that the tooltip will jump to the next item. 

I'm not sure what behavior makes the most sense in that case, but I don't think this is it. I suspect we want to keep the previous point until the mouse is moved.

src/StockMarket/Stock.ts
```diff
@@ -123,6 +123,8 @@ export class Stock {
    */
   readonly totalShares: number;
 
+  readonly priceHistory: { time: number; price: number }[];
```
d0sboots:
```suggestion
  readonly priceHistory: { timeMs: number; price: number }[];
```
Time formatted as numbers should have units for clarity.

src/StockMarket/ui/PriceHistoryChart.tsx
catloversg: It looks like this now:

https://github.com/user-attachments/assets/8070170e-b105-4893-b353-c087794dc0da

PR #3064, opened 2026-08-24
BUGFIX: Invalidate cached RAM when imported scripts change

### LINKED ISSUE #2907: We cache the wrong RAM usage under certain circumstances
1. Create the following scripts:
`testRamUsage1.js`
```js
import { func } from "testRamUsage2.js"

/** @param {NS} ns */
  func(ns); 
  ns.tprint(ns.ramOverride(0));
}
```
`testRamUsage2.js`
```js
/** @param {NS} ns */
export function func(ns) {
  //ns.singularity.connect("home");
}
```
2. Save the game
3. Reload, to get a fresh state with no cached ram usage or LoadedModules
4. run `mem testRamUsage1.js`
> This script requires 1.60GiB of RAM to run for 1 thread(s)
> 1.60GiB | baseCost (misc)
5. edit testRamUsage2.js and un-comment the line, saving the file. Do *not* run any scripts. Do *not* edit testRamUsage1.js.
6. run `mem testRamUsage1.js`
At this point the bug is confirmed.
> This script requires 1.60GiB of RAM to run for 1 thread(s)
> 1.60GiB | baseCost (misc)
7. run `run testRamUsage1.js`
An error will pop up due to insufficient dynamic ram.

----

The bug is due to the way we cache ram usage via `Script.getRamUsage()`. If downstream scripts are changed, it can make the cached valid invalid without actually invalidating the cached value. If there is a LoadedModule (due to running a script), the compilation process tracks dependency chains such that the ram usage *will* get invalidated correctly; this combined with the fact that only oddball methods like the `mem` command cause one-off ram calculation are why this bug went unnoticed.

eviaaaaa:
Fixes #2907.

A RAM lookup can be the first operation that traverses a script's import graph. For example:

```js
// main.js
import { func } from "b.js";
export function main(ns) { return func(ns); }

// b.js
export function func(ns) { return ns.args.length; }
```

Running `mem main.js` before either script has been executed caches `main.js` at 1.6 GB without compiling a `LoadedModule`. If `b.js` is then changed to call a higher-cost API such as `ns.hack()`, another `mem main.js` still returns 1.6 GB instead of recalculating the importer.

`calculateRamUsage` already parses every imported module, but its module-level import edges were discarded with the calculation. The reverse `Script.dependents` edges used for invalidation were only registered by `NetscriptJSEvaluator` during module compilation. In this RAM-only path both scripts still have `mod === null`, so editing `b.js` clears its own RAM data but has no edge through which to invalidate `main.js`. The early return in `invalidateModule` for scripts without a compiled module made the same assumption.

Successful RAM calculations now return the direct import edges and the modules they parsed. `Script` installs those edges as a RAM-owned dependency graph alongside the existing compile-owned graph. The two owners are tracked separately so a RAM recalculation cannot remove an edge still needed by a compiled module, and compile cleanup cannot remove an edge still needed by a valid RAM cache. A failed calculation does not commit a partial replacement graph.

Invalidation clears the current node and detaches its outgoing edges before recursing through importers. That ordering makes circular imports terminate and prevents deleted, recreated, or renamed scripts from leaving old `Script` objects in another script's `dependents` set. Recalculation also removes edges for imports that are no longer present, so later edits to the old dependency do not invalidate an unrelated cache.

The automated regression calls `Script.getRamUsage` directly, which is the same cache path used by both `mem` and `ns.getScriptRam`; it does not drive the browser terminal UI. On the unpatched tree, changing an uncompiled dependency leaves the importer cached at the stale 1.6 GB value, and the same failure appears through transitive and circular import graphs. The permanent tests cover that RAM-only path, dependency replacement and script lifecycles, and shared RAM/compile ownership. The targeted file reports 14 passed, the related script and RAM suites report 167 passed and 2 skipped, and the full Jest suite reports 4,560 passed and 4 skipped.

### DIFF
--- a/src/NetscriptJSEvaluator.ts
+++ b/src/NetscriptJSEvaluator.ts
@@ -47,6 +47,7 @@ export function compile(script: Script, scripts: Map<ScriptFilePath, Script>): P
   try {
     script.mod = generateLoadedModule(script, scripts, []);
   } catch (error) {
+    script.clearModuleDependencies();
     throw new Error(`Cannot generate module ${script.filename}`, { cause: error });
   }
   return script.mod.module;
@@ -60,7 +61,7 @@ export function compile(script: Script, scripts: Map<ScriptFilePath, Script>): P
 function addDependencyInfo(script: Script, seenStack: Script[]) {
   if (!script.mod) throw new Error(`addDependencyInfo called without a LoadedModule (${script.filename})`);
   if (seenStack.length) {
-    script.dependents.add(seenStack[seenStack.length - 1]);
+    seenStack[seenStack.length - 1].registerModuleDependency(script);
     for (const dependent of seenStack) dependent.dependencies.set(script.mod.url, script);
   }
   // Add self to dependencies (it's not part of the stack, since we don't want
--- a/src/Script/RamCalculations.ts
+++ b/src/Script/RamCalculations.ts
@@ -31,9 +31,13 @@ export interface RamUsageEntry {
   cost: number;
 }
 
+export type RamDependencyEdge = [dependent: ScriptFilePath, dependency: ScriptFilePath];
+
 export type RamCalculationSuccess = {
   cost: number;
   entries: RamUsageEntry[];
+  dependencyEdges: RamDependencyEdge[];
+  parsedModules: ScriptFilePath[];
   errorCode?: never;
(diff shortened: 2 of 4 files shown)


E9cipher:
I just saw this while scrolling on gh. Was this ever reviewed?

PR #3065, opened 2026-08-24
RFA: Add file size to metadata API responses

### LINKED ISSUE #3061: RFA: bulk methods accept no bound, and unsupported params are silently ignored
## Summary

The Remote File API's bulk methods accept no bound of any kind, and `isFileServer` silently ignores any parameter a caller adds in the hope of one. A tool built on RFA must therefore fetch a server's entire file list or make one call per file, and a caller who guesses at a `limit` parameter receives the full payload with no indication their request was ignored.

Both observations are read from the handler source rather than inferred from the method list.

## 1. The bulk methods take exactly one parameter

`getFileNames`, `getAllFiles` and `getAllFileMetadata` all route through `validateServerParams`, which validates against `isFileServer`. That interface is:

```ts
export interface FileServer {
  server: string;
}
```

The handler bodies contain no `slice`, `limit`, `offset`, `take` or `cursor`. There is no ranged or paged variant elsewhere in `RFARequestHandler`.

**Measured consequence.** On a save with 466 files on `home`, `getAllFiles` produces a response that exceeds the default `ws` payload limit and closes the connection with code 1006. `getAllFileMetadata` — which carries no file contents at all — still produces roughly 74 KB of JSON for that server. A tool wanting a partial view has only two options: fetch everything and discard, or issue 466 individual `getFile` calls.

## 2. ⚠️ An unsupported parameter is accepted and silently ignored

```ts
export function isFileServer(p: unknown): p is FileServer {
  const pf = p as FileServer;
  return typeof pf.server === "string";
}
```

This validates the presence of `server` and rejects nothing else. A caller sending:

```json
{ "method": "getAllFileMetadata", "params": { "server": "home", "limit": 50 } }
```

passes validation, has `limit` **silently discarded**, and receives the complete unbounded list. There is no error, no warning, and no field in the response indicating a parameter was not understood.

This is the harder half to notice. A rejected request announces itself; a request that succeeds while ignoring half of what was asked does not, and the only signal is a response larger than expected — which for a bulk method is exactly what a caller would expect anyway.

## Suggested directions

Either would resolve the first point, and they are not exclusive:

1. **Accept an optional bound** on the bulk methods — a `limit`, or a `limit`/`offset` pair — and echo the applied value plus a total in the response so a caller can tell whether more remains.
2. **Reject unknown parameters** in the `isFileServer` family, or echo the parameters actually honoured. Even without adding a bound, this converts a silent misunderstanding into a visible one.

The second is valuable on its own. A caller who guesses wrong should learn that they guessed.

## Context

Found while building an MCP server that bridges a running game to a language model over RFA. The tool now caps its own output and returns a per-directory rollup, so this is not a request to unblock anything — the workaround exists and works. The report is about the silent-parameter behaviour being difficult for any RFA consumer to discover, and about bulk methods being all-or-nothing on saves that have grown large.

Versions: read against the current `bitburner-src` working tree, `src/RemoteFileAPI/MessageHandlers.ts` and `src/RemoteFileAPI/MessageDefinitions.ts`.

catloversg:
Context: #3061.

I'm fine with adding pagination support, but I don't think we need to reject unsupported parameters. We would need to define schemas specifying which parameters are allowed for each API, then validate the input parameters against those schemas. Doing that is not hard, but it's also not trivial. IMO, it's not worth the effort, especially since being permissive as we are currently is not really an issue.

If a tool wants to know whether pagination is supported, it needs to read the documentation rather than guess based on the response. I updated the RFA documentation to explicitly mention this.

Note:
- The types defined in `src/RemoteFileAPI/MessageDefinitions.ts` are not exactly correct. For example, `RFARequest.params` was `FileDescription` (`FileDescription & PaginationParams` now). `FileDescription` is `FileData | FileContent | FileLocation | FileServer`, but `FileContent` is not a param type for any API's request. I think we can improve other things in `MessageDefinitions.ts` as well, but let's do that later. I want to keep the changes minimal.
- `getFileNames` returns `textFiles` before `scripts`, while the other two APIs return `scripts` before `textFiles`. Their behavior is now consistent.

Closes #3061.

Test code:
```js
/** @param {NS} ns */
  for (const filename of ns.ls("home")) {
    if (!filename.startsWith("script-")) {
      continue;
    }
    ns.rm(filename, "home");
  }
    ns.write(`script-${i}.js`, `/** @param {NS} ns */
}`, "w");
  }
}
```
RFA client:
```js
import EventEmitter from "node:events";
import { WebSocketServer } from "ws";

export const EventType = {
  MessageSent: "MessageSent",
};

const eventEmitter = new EventEmitter();
let messageCounter = 0;

const wss = new WebSocketServer({ port: 12525 });
wss.on("connection", function connection(ws) {
  ws.on("error", console.error);
  ws.on("message", (msg) => {
    const response = JSON.parse(msg.toString());
    if (response.error) {
      console.error(response);
      return;
    }
    if (response.total == null) {
      throw new Error("missing total");
    }
    console.log(
      response.result.map((v) => {
        if (v.filename != null) {
          return v.filename;
        }
        if (v.hostname != null) {
          return v.hostname;
        }
        return v;
      }),
    );
  });
  eventEmitter.on(EventType.MessageSent, (msg) => {
    ws.send(JSON.stringify(msg));
  });
  eventEmitter.emit(EventType.MessageSent, {
    jsonrpc: "2.0",
    method: "getFileNames",
    params: {
      server: "home",
      limit: null,
    },
    id: messageCounter++,
  });
  eventEmitter.emit(EventType.MessageSent, {
    jsonrpc: "2.0",
    method: "getFileNames",
    params: {
      server: "home",
      offset: null,
    },
    id: messageCounter++,
  });
  eventEmitter.emit(EventType.MessageSent, {
    jsonrpc: "2.0",
    method: "getFileNames",
    params: {
      server: "home",
      limit: -1,
    },
    id: messageCounter++,
  });
  eventEmitter.emit(EventType.MessageSent, {
    jsonrpc: "2.0",
    method: "getFileNames",
    params: {
      server: "home",
      limit: "foo",
    },
    id: messageCounter++,
  });
  eventEmitter.emit(EventType.MessageSent, {
    jsonrpc: "2.0",
    method: "getFileNames",
    params: {
      server: "home",
    },
    id: messageCounter++,
  });
  for (let i = 0; i < 5; ++i) {
    eventEmitter.emit(EventType.MessageSent, {
      jsonrpc: "2.0",
      method: "getFileNames",
      params: {
        server: "home",
        limit: 3,
        offset: i,
      },
      id: messageCounter++,
    });
  }
  for (let i = 0; i < 5; ++i) {
    eventEmitter.emit(EventType.MessageSent, {
      jsonrpc: "2.0",
      method: "getAllFiles",
      params: {
        server: "home",
        limit: 3,
        offset: i,
      },
      id: messageCounter++,
    });
  }
  for (let i = 0; i < 5; ++i) {
    eventEmitter.emit(EventType.MessageSent, {
      jsonrpc: "2.0",
      method: "getAllFileMetadata",
      params: {
        server: "home",
        limit: 3,
        offset: i,
      },
      id: messageCounter++,
    });
  }
});
```

### DIFF
--- a/src/Documentation/doc/en/programming/remote_api.md
+++ b/src/Documentation/doc/en/programming/remote_api.md
@@ -43,6 +43,8 @@ All APIs use a request/response format similar to the JSON RPC 2.0 protocol.
 
 Unknown parameters in requests are ignored.
 
+Pagination is not supported.
+
 Request:
 
         {
--- a/src/Documentation/doc/en/programming/remote_api.md
+++ b/src/Documentation/doc/en/programming/remote_api.md
@@ -68,36 +68,6 @@ Error Response:
             "error": string
         }
 
-### Pagination
-
-Some APIs support pagination. These APIs accept the following optional parameters in `params`:
-
-- `limit`: The maximum number of items to return. If omitted, there is no limit.
-- `offset`: The number of items to skip before returning items. If omitted, defaults to 0.
-
-If specified, `limit` and `offset` must be a non-negative integer. `null` is not a valid value.
-
-        {
-            "jsonrpc": "2.0",
-            "id": number,
-            "method": string,
-            "params": {
-                "limit": number;
-                "offset": number;
-            }
-        }
-
-Paginated responses include a `total` field containing the total number of items before applying `limit` and `offset`:
-
-        {
-            "jsonrpc": "2.0",
-            "id": number,
-            "result": any,
-            "total": number
-        }
-
-In the next section, APIs that support pagination will say "This API supports pagination". For brevity, their request and response schemas do not include the optional `limit` and `offset` parameters or the `total` field. The pagination rules described above still apply.
-
 ### API list
 
 #### pushFile
@@ -172,9 +142,10 @@ Response:
             "id": number,
             "result": {
                 "filename": string,
-                "atime": string,
-                "btime": string,
-                "mtime": string
+                "size": number,
+                "atime": number,
+                "btime": number,
+                "mtime": number
             }
         }
 
@@ -206,8 +177,6 @@ Response:
 
 List all file names on a server.
 
-This API supports pagination.
-
 Request:
 
         {
@@ -230,8 +199,6 @@ Response:
 #### getAllFiles
 
 Get the content of all files on a server.
-
-This API supports pagination.
 
 Request:
 
@@ -259,8 +226,6 @@ Response:
 
 Get the content of all files on a server.
 
-This API supports pagination.
-
 Request:
 
         {
@@ -279,9 +244,10 @@ Response:
             "id": number,
             "result": {
                 "filename": string,
-                "atime": string
-                "btime": string,
-                "mtime": string,
+                "size": number,
+                "atime": number,
+                "btime": number,
+                "mtime": number
             }[]
         }
 
--- a/src/Documentation/doc/en/programming/remote_api.md
+++ b/src/Documentation/doc/en/programming/remote_api.md
@@ -37,7 +37,11 @@ Links:
 
 ## API specification
 
+### Overview
+
 All APIs use a request/response format similar to the JSON RPC 2.0 protocol.
+
+Unknown parameters in requests are ignored.
 
 Request:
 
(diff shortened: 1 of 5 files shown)


d0sboots:
I am very uncertain that we should support this. Pagination comes with a host of insidious issues:
* The order that results are returned goes from not mattering to being very important, and most likely, part of the API contract.
* Note that there is no purpose to pagination if you are going to grab everything, so it's *only* relevant for getting filtered results, but not results that are filtered in a more structured way (i.e. by search query), since that would require a different API.
* There are subtle race conditions involved with fetching paginated results. If a file is deleted from the first page of results between when the first and second page is fetched, you will miss the (previously) first item on the second page and never know it.

catloversg:
I'm neutral on this, but I think you should make the call now. If we won't support pagination, I will update the documentation to point out that: 1. unknown parameters are ignored, and 2. pagination is not supported.

d0sboots:
Given the further follow-up on the initial issue, I don't think pagination is a real need we have. The main todo (aside from saying it's not supported) is adding file sizes to getMetadata.

PR #3066, opened 2026-08-24
MISC: Sorting algorithms and text

branchtheory:
Currently we use the OOTB sort for the output of `ls` commands in the terminal. This sorts by ASCII value (technically UTF-16). The problem is that that separates capitals from lower case.

image

This PR sorts things truly alphabetically:

image

Also added some small misc changes to sorting:
* added caseFirst: upper to all localeCompares, because it's more Bitburner (the exception is where we sort md files) 
* changed some bizarre text to do with sorting ("Acquirement", "in Order")

### DIFF
--- a/src/Augmentation/ui/InstalledAugmentations.tsx
+++ b/src/Augmentation/ui/InstalledAugmentations.tsx
@@ -40,7 +40,7 @@ export function InstalledAugmentations(): React.ReactElement {
 
   if (Settings.OwnedAugmentationsOrder === OwnedAugmentationsOrderSetting.Alphabetically) {
     filteredSourceAugs.sort((aug1, aug2) => {
-      return aug1.name.localeCompare(aug2.name);
+      return aug1.name.localeCompare(aug2.name, undefined, { caseFirst: "upper" });
     });
   }
 
--- a/src/Augmentation/ui/InstalledAugmentations.tsx
+++ b/src/Augmentation/ui/InstalledAugmentations.tsx
@@ -40,7 +40,7 @@ export function InstalledAugmentations(): React.ReactElement {
 
   if (Settings.OwnedAugmentationsOrder === OwnedAugmentationsOrderSetting.Alphabetically) {
     filteredSourceAugs.sort((aug1, aug2) => {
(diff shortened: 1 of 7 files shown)


d0sboots:
This requires some thought... On Linux, this is controlled by LC_COLLATE. It's probable that the right answer for us would be a setting (that would potentially/eventually apply to more than just ls)

branchtheory:
> This requires some thought... On Linux, this is controlled by LC_COLLATE. It's probable that the right answer for us would be a setting (that would potentially/eventually apply to more than just ls)

Can you expand on this a bit? I can't make it out.

I don't think LC_COLLATE is used. As far as I can tell, the chain of calls goes: localCompare -> uses Intl.collate -> uses ICU Collation API -> uses CLDR. LC_COLLATE doesn't seem to be involved.

Also, exactly which algorithm / sorting methods are used seems to be decided by the browser, rather than the operating system.

The bit about settings confused me too. To me it doesn't feel like the sort of thing most users would want to customise.

d0sboots:
Sorry, I was taking about the system utility "ls" that our command is based off of. On Linux, its behavior is determined by LC_COLLATE, which determines how C strings are compared when using `strcoll`.

Historically, unix ls functioned the way our ls works right now. However, with typical Unicode settings, it works the way you're proposing.

branchtheory:
Ah, ok, I'm with you. Looks like lower-case first is the standard tie-breaker with `ls` too. I'll update accordingly.

(Will also remove some pointless changes.)

src/Augmentation/ui/InstalledAugmentations.tsx
```diff
@@ -59,14 +59,14 @@ export function InstalledAugmentations(): React.ReactElement {
       <Paper sx={{ p: 1 }}>
         <Typography variant="h5">Installed Augmentations</Typography>
         <Box sx={{ display: "grid", gridTemplateColumns: "1fr 1fr" }}>
-          <Tooltip title={"Sorts the Augmentations alphabetically in numeral order"}>
+          <Tooltip title={"Sorts the Augmentations alphabetically"}>
```
d0sboots, line 62: Scope creep. This is a different PR entirely, although I'll let it slide.

PR #3069, opened 2026-08-27
SIDEBAR: Minor refactor

branchtheory:
Minor refactor.

* `<ListItem button>` is deprecated. Changed to `<ListItemButton>`.
* Removed a redundant MakeStyles with an empty listitem property (??), which in turn was being used by a few redundant useMemos (?!), which I've also removed, along with all the redundant class references (?!!)
* Moved the only useful bit of the MakeStyles thing into an sx{{}} at the relevant element

### DIFF
--- a/src/Sidebar/ui/SidebarAccordion.tsx
+++ b/src/Sidebar/ui/SidebarAccordion.tsx
@@ -1,5 +1,6 @@
 import React, { useMemo, useState } from "react";
 import Collapse from "@mui/material/Collapse";
+import ListItem from "@mui/material/ListItem";
 import ListItemButton from "@mui/material/ListItemButton";
 import ListItemIcon from "@mui/material/ListItemIcon";
 import ListItemText from "@mui/material/ListItemText";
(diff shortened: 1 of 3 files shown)


d0sboots:
Once again, refactoring should be split from functional changes. Especially in cases like this, where it's less clear if the refactoring is supposed to be 100% equivalent.

branchtheory:
Sorry to make you say that again! 

I've updated. It's all refactoring now.

E9cipher:
Does this change the UI appearance? If so, could you provide some screenshots?

branchtheory:
Zero change to the UI 🙂 It's only a refactor. Just removes some clutter, and rationalised what's left.

That said, there's a much, much better way of doing this. Let me get back to this ...

branchtheory:
Actually I'm going to withdraw this altogether. The use of useMemo is much cleverer than I realised and I'm just not good enough at React yet to be tinkering with it. ListItem does need replacing with ListItemButton, but I feel like I should leave this to someone who understands the code better.

E9cipher:
You could create an issue, see if anyone wants to take it

PR #3050, opened 2026-08-13
UI: Rework RFA connection indicator and RFA setting page

catloversg:
Context: https://github.com/bitburner-official/bitburner-src/pull/3012#issuecomment-5122597027

Closes #3018.

Demo:

https://github.com/user-attachments/assets/4fee4f20-1e20-487d-9590-cb1f92cf50f4

PR #3070, opened 2026-08-30
RFA: Pressing enter key on textfields attempts to connect

E9cipher:
While CatLover made it easier to connect within the Overview in #3050 , it has still to be done manually in the RFA settings page. This PR makes pressing enter on the "hostname"/"port" textfields attempt to connect. I didn't add this for the reconnection field because it seemed irrelevant to me, but I can add it if you want.

### DIFF
--- a/src/GameOptions/ui/RemoteAPIPage.tsx
+++ b/src/GameOptions/ui/RemoteAPIPage.tsx
@@ -81,6 +81,11 @@ export const RemoteAPIPage = (): React.ReactElement => {
     setReconnectionDelayError("");
   }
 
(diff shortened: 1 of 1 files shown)


d0sboots:
I'm not sure that I agree with the goal of this PR. Now that we have an explicit connect/disconnect, it feels like you you should explicitly connect when you are entering the address details.

E9cipher:
I think it's easier and more intuitive. Normally, in most forms, when you press enter on any of the inputs the form attempts a submit event. I was confused that it didn't work here, I'm sure it has happened as well to more people than me.

E9cipher:
Hey! Do you think this needs anything else before a review?

PR #3072, opened 2026-08-30
API: Remove ns.getHostname() RAM cost to match ns.self().server

Traster101:
ns.getHostname() has a RAM cost of 0.05 GB, but after the addition of ns.self(), the same value is available via ns.self().server with 0 RAM cost. This PR removes the arbitrary RAM cost difference by making ns.getHostname() cost 0. 
The 0.05 GB cost still exists on ns.getIP(), since this is not available to ns.self(), though it too should probably also have its RAM cost removed to stay consistent.


E9cipher:
I would make `self` 0.05 and leave getHostname to 0.

Also, a madness idea: remove `getIP` and make getHostname return the hostname and IP (as separate props). This avoids two almost-identical APIs being scattered around (which is cleaner imo)

d0sboots:
> I would make `self` 0.05 and leave getHostname to 0.

The ram cost of `self` can't be increased, not without breaking lots of people's scripts. And `getHostname` is 0.05, "leaving" it to 0 doesn't make logical sense.

> Also, a madness idea: remove `getIP` and make getHostname return the hostname and IP (as separate props). This avoids two almost-identical APIs being scattered around (which is cleaner imo)

This would be an even more breaking change.

src/Netscript/RamCostGenerator.ts
```diff
@@ -28,7 +28,8 @@ export const RamCostConstants = {
   Scp: 0.6,
   Kill: 0.5,
   HasRootAccess: 0.05,
-  GetHostname: 0.05,
+  GetHostname: 0,
```
d0sboots: These constants are annoying and not really helpful, we are slowly passing them out. Just delete this one.

ficocelliguy:
There are a lot of different ways to get data from the NS api, and making each one cost the same amount or RAM is not really a goal we have in general. Rather, it is a place where players can optimize their scripts and be rewarded for reading the docs or exploring the API.

That's not a reason to prevent this particular change - the ram cost is negligible either way - but I would not want to see a broader pass trying to equalize a bunch of other comparable situations like getServer and the specific sub-methods etc

PR #3077, opened 2026-09-02
GH: Remove multiple OS selection from bug report template

E9cipher:
I think it's nice to have a title of the issue type, now that issue templates have been added. I also removed the `multiple:true` statement in the OS dropdown, it doesn't make sense in this context (selecting multiple OSs).

The titles are an h3, same as the preflight checklist. I choosed that over h2 because imo h2 is too big, but I can adjust it if you wish.

### DIFF
--- a/.github/ISSUE_TEMPLATE/1_bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/1_bug_report.yml
@@ -2,6 +2,10 @@
 name: Bug Report
 description: Report a bug
 body:
+  - type: markdown
+    attributes:
+      value: |
+        ### Bug Report
   - type: checkboxes
     attributes:
       label: Preflight Checklist
@@ -43,7 +47,6 @@ body:
   - type: dropdown
     attributes:
       label: What operating system(s) are you using?
-      multiple: true
       options:
         - Windows
         - macOS
--- a/.github/ISSUE_TEMPLATE/2_feature_request.yml
+++ b/.github/ISSUE_TEMPLATE/2_feature_request.yml
@@ -1,6 +1,10 @@
 name: Feature request
 description: Suggest ideas for in-game features or other changes to the project.
 body:
+  - type: markdown
+    attributes:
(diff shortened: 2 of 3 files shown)


catloversg:
The additional row is unnecessary. A bug report is clearly ... a bug report without an additional row in the description. The same applies to feature request issues. With the blank issue, it's really weird to say "Blank issue" in the description. "Blank issue" is merely a template. No issue is blank.

If the reporter wants to be clear, they can change the title.

> I also removed the multiple:true statement in the OS dropdown, it doesn't make sense in this context (selecting multiple OSs).

Why does it not make sense? A bug may affect only one OS or multiple OSs at the same time. That's useful information.

E9cipher:
I think the new row makes it easier to see what the PR is about at first glance. The "blank issue" title could be changed, yes. But the rest I think it's a good idea.

About the OS dropdown: the label indicates what OS(s) are you _using_. I see your point that a bug can happen on multiple OSs, then the label should be changed to "what operating system(s) are affected". I originally saw the label and the multiple dropdown and I thought that it made no sense to be running bitburner on multiple OSs at the same time.

.github/ISSUE_TEMPLATE/1_bug_report.yml
```diff
@@ -43,7 +47,6 @@ body:
   - type: dropdown
     attributes:
       label: What operating system(s) are you using?
```
d0sboots:
```suggestion
      label: What operating system are you using?
```

d0sboots (commented):
I don't think the type of template is particularly valuable information after the fact, such that we need to capture it like this. The templates exist to help guide the reporters, not to (directly) change how maintainers look at issues. 

However, I agree that the multi select is probably not worth it on the bug report template. Yes, it's possible to be running BB on multiple OSes, but that is very rare compared to the dominant case, which is "we want to make it as simple as possible for users who know nothing to report the info we *might* need."

PR #3079, opened 2026-09-04
TUTORIAL: Text and UI changes

branchtheory:
This is ready for review!

To get a feel for what's changed, I'd recommend going through the current tutorial first.

You can try the new version at https://a-branch-of-branch.github.io/bitburner-src/.

### What I've added

The 2 additional steps are:
* checking script RAM cost in the Script Editor
* scp-ing our hack script

And there are 3 new bits of information:
* the bit about Tech centres in the World step (adapted from the "you can buy extra ram on home" bit in the original)
* these two lines in the Script Editor step:
    * "When using scripts to hack, you don't need to be connected to the target server. Scripts can target any server, no matter where you are."
    * "Almost everything we do in this tutorial can be scripted using functions available to you right away." 

Despite that, the new version is 331 words / 21% shorter (1262 words compared to 1593).

### What I've removed

* Hacknet
* A lot of wordiness

### Less obvious changes

All the changes should be visible on the surface just from doing the tutorial, with the exception of these, which only appear circumstantially:
* Flashing the Script Editor tab without incrementing the step. In the current code, if you flash a sidebar item, the associated useEffect increments the tutorial step. The new `incrementTutorialStep` variable lets you choose whether or not to increment. So now we flash the Script Editor tab to help you get back to it if, for example, you follow the API documentation links.
* The changes to error messages when interacting with the Terminal incorrectly (wrong command, wrong step etc.).
* The change to the dialogue message you get when you save a different version of the code to the one from the tutorial.

### Refactoring

Tried to do this all in the other PR but missed some things:
* I've done a refactor of the switch statement in Terminal.ts. The flow is now **way** clearer and it should be easier to review.
* I've added an isRunning check to SidebarRoot. 
* I've changed references to "n00dles.js" to tutorialScriptName where possible.

### One irrelevant change

There's one irrelevant change -- to src/Sidebar/ui/SidebarAccordion.tsx -- which was collateral from running `npm run format`. Not intended but I can't seem to get rid of it.


src/Documentation/doc/en/help/getting_started.md
```diff
@@ -19,11 +19,11 @@ Since this guide is written as a basic introduction to Bitburner, it will not ex
 ## First Steps
 
 I'm going to assume you followed the introductory tutorial when you first began the game.
-In this introductory tutorial, you created a [Script](../basic/scripts.md) called `n00dles.js` and ran it on the `n00dles` server.
+In this introductory tutorial, you created a [Script](../basic/scripts.md) called `hacking.js` and ran it on the `home` server.
```
E9cipher, line 22: (optional) I think that the original script name is fine.

```suggestion
In this introductory tutorial, you created a [Script](../basic/scripts.md) called `hacking.js` and ran it on the `n00dles` server.
```
The script was ran on home, yes, but it's confusing for the player, so I'd leave it to n00dles

src/InteractiveTutorial.ts
```diff
@@ -21,16 +17,23 @@ enum iTutorialSteps {
...
-  TerminalEditScript, // Script Editor page - Edit script and then save & close
+  ScriptEditorEdit, // Script Editor page - Edit script
+  ScriptEditorRam, // Script Editor page - Check RAM usage
+  ScriptEditorGoToTerminal,
   TerminalFree, // Using 'Free' Terminal command
   TerminalRunScript, // Running script using 'run' Terminal command
   TerminalGoToActiveScriptsPage,
   ActiveScriptsPage,
   ActiveScriptsToTerminal,
   TerminalTailScript,
```
E9cipher: Now that we're at it, what do you think of adding comments to those?

src/InteractiveTutorial.ts
```diff
@@ -21,16 +17,23 @@ enum iTutorialSteps {
...
+  TerminalLs, // Using 'ls' Terminal command
+  TerminalScp, // using the 'scp' Terminal command
+  TerminalHelp, // Using 'help' Terminal command
+
+  // Finishing off
+  GoToCharacterPage, // Click on 'Stats' page
+  CharacterPage, // Introduction to 'Stats' page
+  GoToWorldPage,
   WorldDescription,
   DocumentationPageInfo,
```
E9cipher: Comments as well?

src/Terminal/Terminal.ts
```diff
@@ -347,7 +348,7 @@ export class Terminal {
         throw new Error("Could not get n00dles server");
       }
       const errorMessageForBadCommand =
-        "Bad command. Please follow the tutorial or click 'Exit Tutorial' if you'd like to skip it.";
+        "Wrong command. Try again, or if you'd like to skip the tutorial click Exit Tutorial";
```
E9cipher: Smh I feel the original was more natural. Normally, a wrong command is said to be a "bad command". The rest of the changes I feel that they do not sound as natural.

src/Terminal/Terminal.ts
```diff
@@ -393,14 +394,14 @@ export class Terminal {
...
+          if (commandArray[0] === "connect") {
             if (
-              commandArray[0] === "connect" &&
+              commandArray.length === 2 &&
               (commandArray[1] === "n00dles" || commandArray[1] === n00dlesServ.hostname)
             ) {
               iTutorialNextStep();
             } else {
-              this.error("Wrong command! Try again!");
+              this.error("Wrong command. Try again!");
```
E9cipher: Hmmm...

src/Terminal/Terminal.ts
```diff
@@ -417,7 +418,10 @@ export class Terminal {
           }
           break;
         case iTutorialSteps.TerminalNuke:
-          if (commandArray.length === 2 && commandArray[0] === "run" && commandArray[1] === "NUKE.exe") {
+          if (
+            (commandArray.length === 1 && commandArray[0] === "NUKE.exe") ||
```
E9cipher: Can't see the point in here

src/Terminal/Terminal.ts
```diff
@@ -417,7 +418,10 @@ export class Terminal {
...
+            (commandArray.length === 2 && commandArray[0] === "run" && commandArray[1] === "NUKE.exe")
+          ) {
             iTutorialNextStep();
```
E9cipher: Optional (idk if it is a design choice): use inline `if`

src/Terminal/Terminal.ts
```diff
@@ -417,7 +418,10 @@ export class Terminal {
           }
           break;
         case iTutorialSteps.TerminalNuke:
-          if (commandArray.length === 2 && commandArray[0] === "run" && commandArray[1] === "NUKE.exe") {
+          if (
```
E9cipher: Optional (only if the next comment suggestion is not applied): swap the sides of the operator to reduce diff?

src/Terminal/Terminal.ts
```diff
@@ -463,23 +472,51 @@ export class Terminal {
...
+              return;
+            }
+          } else {
+            this.error(errorMessageForBadCommand);
+            return;
+          }
+          break;
         default:
-          this.error("Please follow the tutorial or click 'Exit Tutorial' if you'd like to skip it");
+          this.error("Make sure you're following the tutorial. If you'd like to skip it click Exit Tutorial");
```
E9cipher: Same as before, I feel the previous one was better

src/Terminal/Terminal.ts
```diff
@@ -447,8 +451,13 @@ export class Terminal {
           }
           break;
         case iTutorialSteps.TerminalCreateScript:
-          if (commandArray.length === 2 && commandArray[0] === "nano" && commandArray[1] === "n00dles.js") {
-            iTutorialNextStep();
+          if (commandArray[0] === "nano") {
+            if (commandArray.length === 2 && commandArray[1] === tutorialScriptName) {
+              iTutorialNextStep();
+            } else {
+              this.error("Wrong command. Try again!");
```
E9cipher:
```suggestion
              this.error("Bad command! Try again.");
```

E9cipher (commented):
General review. @d0sboots you might want to check it as well, I have some suggestions that need your approval.

src/Documentation/doc/en/help/getting_started.md
branchtheory: 1. I think the script name `n00dles.js` sets the wrong expectations. It suggests that you need to create a different script for every server - foodnstuff.js, harakiri-sushi.js, and so on. 
2. The following lines in getting_started.md are about killing the script. Saying the script was run on home links to that -- it ran on home, so we need to kill it on home. I know "run on" has different meanings (it can mean "targeted"), but in context the meaning is the location it was run.

src/InteractiveTutorial.ts
branchtheory: 🤷 I can, it just felt like adding comments for the sake of it

src/Terminal/Terminal.ts
branchtheory: The command may not be bad though, exactly. `connect foodnstuff` is perfectly correct ... it's just not the command you need to enter in the moment. In other words, it's the wrong command.

To me, "Please follow the tutorial" sounds a bit passive-aggressive-yoga-teacher-y. "Try again" is more optimistic and relevant to the specific action.

Edit: Reverted to "Bad command." We could add logic to check whether the command is 'bad' (syntax) or 'wrong' (syntactically correct but not following the tutorial) buuuuut ... keep it simple.

src/Terminal/Terminal.ts
branchtheory: If I'm going to exclaim at people, I want to exclaim encouragement, not criticism.

src/Terminal/Terminal.ts
branchtheory: It means that `NUKE.exe` works too, not just `run NUKE.exe`. A tiny QOL. We still tell the player to enter `run NUKE.exe` though.

src/Terminal/Terminal.ts
branchtheory: Usually you don't want to use exclamation marks in error text. But in this case the error should be very obvious and easy to recover from. So the player isn't likely to feel 'got at' or frustrated. So I think "Try again!" just about functions as giving energy and encouragement, rather than just shouting at people.

I'm trying to be a bit more strategic with how we use exclamation marks in general. The current tutorial uses them a lot, and often for stuff that doesn't feel like a big win. It's kind of charming and energising, but it means that the significant moments lose their weight.

Here are the exclamation marks from the original:

* Welcome to Bitburner, a cyberpunk-themed incremental RPG!
* You are now connected to another machine! What can you do now? You can hack it!
* You now have root access!
* Let's head home before creating our first script!
* ... you don't want to be hacking manually the entire time. You can automate your hacking by writing scripts!
* This log might not show much because the script just started... but check back again in a few minutes!
* For now, let's move on to something else!
* You just purchased a Hacknet Node! 
* There's a lot of content out in the world, make sure you explore and discover!

And here are the ones that I've used:

* The network's much bigger than that though!
* Now you've got root access, you can hack the server! 
* ... you don't want to be hacking manually the entire time. You need a way to automate it!
* You have 8GB of free RAM, enough to run your script!
* This log might not show much because the script just started... but check back again in a few minutes!
* There's a lot of other content out in the world, so make sure you explore and discover!

I've tried to make sure the exclamation marks land mostly on light-bulb moments, or moments where the player's finally finished a long sequence of steps to set something up.

src/InteractiveTutorial.ts
branchtheory: Oh dammit you are right though. And some of the step names should be updated too.

src/Terminal/Terminal.ts
branchtheory: Hmmm, I need to have a think about this actually.

On reflection, the wrong/bad thing is moot. We have no idea whether their command is syntactically invalid (bad), or whether it's a correct command but not what the tutorial wants (wrong). Implementing a check for that would be overkill.

I'm also not sure about the stuff about exiting the tutorial either. We've already told the player they can exit the tutorial, and they've got an Exit Tutorial button staring them in the face. To tell them they can exit the tutorial _again_ makes it almost sound like we think they should.

I wonder if we should just simplify things and make all error responses the same: "Wrong/Bad command. Try again." I still prefer wrong, but I'm happy with either. (The exception is the 1s/ls thing, which really does need to be different.)

branchtheory: @E9cipher , check the new version. I think it's waaaaay cleaner, both for the player and for us.

E9cipher: Yeah, short and precise. I like it!

src/Terminal/Terminal.ts
E9cipher: Hmmm then why allowing this secret case? If we teach players to use `run`, make them use it. If we wanna keep this second option we have to tell players too. I think eliminating the second case is easier for players and for us. What do you think about it?

src/Documentation/doc/en/help/getting_started.md
E9cipher: I'll have to get used to tutorial 2.0 then 😆😅

src/Terminal/Terminal.ts
branchtheory: Yeah you're right. It's pretty pointless. I've already removed the other secret cases that I added. I'll remove this one too.

Thanks for the nudge.

src/Terminal/Terminal.ts
branchtheory: Thanks!

PR #3080, opened 2026-09-04
MISC: grow and weaken terminal command text output

branchtheory:
**Weaken** – Just fixing a typo.

**Grow** – The aim to give players richer feedback on what grow's accomplished. 

The grow command will mostly be used by people who are brand new, and helps them understand what grow accomplishes before they start scripting it. 

Seeing the server's money would be really helpful for that. It gives them a sense of how much money servers hold, and what grow's done.

It also creates a better emotional feedback loop. A % increase is nice. But seeing the actual money is more rewarding.

I don't think this overpowers grow – there are other faster and more informative terminal commands that show you a server's available money.

### DIFF
--- a/src/Terminal/commands/grow.ts
+++ b/src/Terminal/commands/grow.ts
@@ -4,7 +4,7 @@ import { BaseServer } from "../../Server/BaseServer";
 import { Server } from "../../Server/Server";
 import { HacknetServer } from "../../Hacknet/HacknetServer";
 import { Player } from "@player";
-import { formatExp, formatPercent, formatSecurity } from "../../ui/formatNumber";
+import { formatExp, formatPercent, formatSecurity, formatMoney } from "../../ui/formatNumber";
 import { calculateHackingExpGain, calculateGrowTime } from "../../Hacking";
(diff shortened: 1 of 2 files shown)


E9cipher:
I don't really agree with the change in grow:

Imo, giving players the money makes their life with the terminal easier, and we don't really want that. The terminal should be the first kick, and scripts should be the way to get that info.

A player will surely want to know how much money is there, so eventually they'll create a script to check it, which will lead to more complicated ones. If we give them the data directly here, that whole process becomes kinda pointless.

branchtheory:
I see where you're coming from, but I think that manually `grow`ing in the terminal is excruciating enough to push players towards scripting. I doubt that being able to see the available money will change that.

I think that the 'players will eventually create a script to check server money' reason for not including the money is looking at things from the wrong perspective. Yes, players can eventually create a script. But during the period of the game when you're using `grow` manually, you don't have that ability yet! 

The point of this change is to motivate players and help them get a feel for the game faster. If we can do that, we can get them scripting sooner.

For the reasons I mentioned in the first post, I don't think anyone will use this as their main way to check how much money is on a server. There are much better terminal commands for doing that, like `analyze`.

PR #3081, opened 2026-09-04
REFACTOR: Tutorial styling

branchtheory:
Supports #3079.

### DIFF
--- a/src/ui/InteractiveTutorial/InteractiveTutorialRoot.tsx
+++ b/src/ui/InteractiveTutorial/InteractiveTutorialRoot.tsx
@@ -63,8 +63,8 @@ export function InteractiveTutorialRoot(): React.ReactElement {
         <>
           <Typography>Let's start by heading to the Stats page. Click</Typography>
           <ListItem>
-            <EqualizerIcon sx={{ color: "error.main" }} />
-            <Typography sx={{ color: "error.main" }}>Stats</Typography>
+            <EqualizerIcon color={"error"} />
+            <Typography color={"error"}>Stats</Typography>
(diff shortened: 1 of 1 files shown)


E9cipher:
Why the separate PR? Name #3079 "refactor tutorial" and include these changes there.

branchtheory:
Señor Boot's previously asked if I can separate out refactoring and functional changes.

src/ui/InteractiveTutorial/InteractiveTutorialRoot.tsx
```diff
@@ -13,8 +11,7 @@ import HelpIcon from "@mui/icons-material/Help";
 import AccountTreeIcon from "@mui/icons-material/AccountTree";
 import StorageIcon from "@mui/icons-material/Storage";
 import LocationCityIcon from "@mui/icons-material/LocationCity";
-import { Theme } from "@mui/material/styles";
-import { makeStyles } from "tss-react/mui";
+import { styled } from "@mui/material/styles";
```
d0sboots: We use styled a grand total of 3 times in 2 files across the whole codebase, currently.

I'm not objecting to it on a technical basis. However, there are a *lot* of potential ways to style components, and we use most of them. Unless we're making a broader push to switch to this one by getting rid of one of the others (which would require a bigger discussion of pros and cons), I don't think we should introduce yet another way.

branchtheory: Thanks, I didn't really like that either. Also realised that the function to create TerminalText was executing with every rerender! Corrected both issues.

PR #3086, opened 2026-09-06
UX: Show properly mapped line numbers in errors thrown by TS player scripts

ficocelliguy:
Previously, the error modals shown to players had incorrect line numbers in the stacktrace. This was because the error lines were pulled from the transpiled js code, and not mapped to the original ts. 

This has caused some players frustration trying to track down errors : https://discord.com/channels/415207508303544321/415207923506216971/1545843857344700427

This PR now retrieves the source mapping if an error is thrown, and uses that to re-map the line and column numbers in the error stacktrace shown to the user to correspond correctly to the original file.

Implementation notes:
- the source map is retrieved on-demand when an error is thrown, to avoid memory overhead of storing it for lots of scripts
- the `source-map-js` library used to apply mapping was already a child dependency of the project
- this works with .jsx and .tsx as well!


catloversg:
I would like to handle this in #1951. I don't remember the exact details off the top of my head, but there are many edge cases and even bugs in our codebase that need to be dealt with.

ficocelliguy:
Should I leave this PR as a draft for reference, or just close it in favor of that one?

catloversg:
Leave it open. Using source map to improve the stack trace is easy, but due to how we cache modules, there are many edge cases. Having another approach and a set of tests is useful for this kind of PR.

d0sboots:
I also have https://github.com/bitburner-official/bitburner-src/pull/2916 open which cuts a lot of the complexity out of this space, hopefully removing some of those edge-cases.

Issue #3018, opened 2026-07-29
Remote API "Connect" button does not disconnect nor update its text accordingly


E9cipher:
There are a few bugs here, I'm not sure if they can be fixed in a minimal way.
### Bug 1: the button does not update its text accordingly
MRE: 
1. Set up the Remote API on vscode/your editor (not in bitburner). Start the server.
2. Go to the Remote API page on bitburner, click "Connect". Should successfully connect, but the button still says "connect" instead of disconnect.

### Bug 2: you can never disconnect the API from bitburner
MRE:
1. Follow steps from bug 1.
2. Try to disconnect the API. Unless you stop it from the other end, bitburner won't let you. You have to edit the port/hostname config and click connect again.

catloversg:
It seems you imply there should be a "Disconnect" feature, but that's debatable. Let's see what other people think about it.

d0sboots:
I'm for this. When I read this, I was actually shocked (due to not having poked remote api recently) that "disconnected" is not a state we currently support. I understand why we'd want to always auto-connect on start, though.

gmcew:
In case there's questions - I labelled this as an enhancement because how we handle the change (if any) is for review; it's not that this current behaviour is unintended per se.

PR #3087, opened 2026-09-07
REFACTOR: Consistency in the RFA codebase

### LINKED ISSUE #3071: CODEBASE/RFA: Naming is inconsistent
### Preflight Checklist

- [x] I have read the [Contributing Guidelines](https://github.com/bitburner-official/bitburner-src/blob/dev/CONTRIBUTING.md).
- [x] I have searched the [issue tracker](https://github.com/bitburner-official/bitburner-src/issues) for an existing report of this issue and found none.
- [x] I understand that the issue tracker is for bugs in the game itself and feature requests, not for issues with my own scripts.

### Description

I could not come up with a better title to describe this issue, sorry.

In the RFA files, names are inconsistent. For example, some functions name RFA as `remoteFileApi`, others as `RFA`. There is also `isValidConnectionHostname` vs `isValidRFAConnectionPortSetting`. Another example is `canCreateRemoteFileApiConnection` (`can`) vs `isRemoteFileApiConnectionLive` (`is`). 

I think we should agree on using consistent naming for those functions. If this gets approved, I will PR my draft branch.

E9cipher:
Context in and closes #3071 
For now, I have used RFA consistently across the codebase. Note that filenames remain untouched (they all used `remoteFileApi` so I don't think a change is necessary).
I may be refactoring some other names, such as using `can` or `is`

### DIFF
--- a/src/GameOptions/ui/RemoteAPIPage.tsx
+++ b/src/GameOptions/ui/RemoteAPIPage.tsx
@@ -2,36 +2,32 @@ import React, { useEffect, useState } from "react";
 import { Button, TextField, Tooltip, Typography } from "@mui/material";
 import { GameOptionsPage } from "./GameOptionsPage";
 import { Settings } from "../../Settings/Settings";
-import { isValidConnectionHostname, isValidRFAConnectionPortSetting } from "../../Settings/SettingsUtils";
-import { RemoteFileApiConnectionStatus } from "./RemoteFileApiConnectionStatus";
+import { isValidRFAHostname, isValidRFAPort } from "../../Settings/SettingsUtils";
+import { RFAConnectionStatus } from "./RemoteFileApiConnectionStatus";
 import {
-  canCreateNewRemoteFileApiConnection,
-  closeRemoteFileApiConnection,
-  isRemoteFileApiConnectionLive,
-  newRemoteFileApiConnection,
+  canCreateNewRFAConnection,
+  closeRFAConnection,
+  isRFAConnectionLive,
+  newRFAConnection,
 } from "../../RemoteFileAPI/RemoteFileAPI";
 import { OptionSwitch } from "../../ui/React/OptionSwitch";
 import { DocumentationLink } from "../../ui/React/DocumentationLink";
-import { RemoteFileApiConnectionEvents, RemoteFileApiConnectionSettingEvents } from "../../RemoteFileAPI/Remote";
+import { RFAConnectionEvents, RFAConnectionSettingEvents } from "../../RemoteFileAPI/Remote";
 import { useRerender } from "../../ui/React/hooks";
 
 export const RemoteAPIPage = (): React.ReactElement => {
-  const [remoteFileApiHostname, setRemoteFileApiHostname] = useState(Settings.RemoteFileApiAddress);
-  const [hostnameError, setHostnameError] = useState(
-    isValidConnectionHostname(Settings.RemoteFileApiAddress).message ?? "",
-  );
-  const [remoteFileApiPort, setRemoteFileApiPort] = useState(Settings.RemoteFileApiPort.toString());
-  const [portError, setPortError] = useState(isValidRFAConnectionPortSetting(Settings.RemoteFileApiPort).message ?? "");
-  const [remoteFileApiReconnectionDelay, setRemoteFileApiReconnectionDelay] = useState(
-    Settings.RemoteFileApiReconnectionDelay.toString(),
-  );
+  const [rfaHostname, setRFAHostname] = useState(Settings.RFAAddress);
+  const [hostnameError, setHostnameError] = useState(isValidRFAHostname(Settings.RFAAddress).message ?? "");
+  const [rfaPort, setRFAPort] = useState(Settings.RFAPort.toString());
+  const [portError, setPortError] = useState(isValidRFAPort(Settings.RFAPort).message ?? "");
+  const [rfaReconnectionDelay, setRFAReconnectionDelay] = useState(Settings.RFAReconnectionDelay.toString());
   const [reconnectionDelayError, setReconnectionDelayError] = useState("");
 
   const rerender = useRerender();
 
   useEffect(
     () =>
-      RemoteFileApiConnectionEvents.subscribe(() => {
+      RFAConnectionEvents.subscribe(() => {
         rerender();
       }),
     [rerender],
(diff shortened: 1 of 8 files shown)


catloversg:
I'm not really fond of this kind of change. It's good to be consistent, but changing `remoteFileApi` to `RFA` or vice versa is a bit meaningless. I won't veto it, but you should keep this kind of change minimal. The codebase has a massive amount of this type of inconsistency. The benefit of refactoring is very small. This is especially true if you refactor more of the "wording" of variable/function names ("can", "is", etc., like you said).

Note that it's only fine if the changes are correct. Even without testing, I'm fairly sure the changes in `Settings.ts` are wrong. The keys are included in the save data. If you rename them like that, you need to migrate save data, which is definitely overcomplicated. It's best not to touch anything in that file. Note that this means we accept that it's fine to be inconsistent in some cases. This is also why I think we don't really need to refactor stuff like this. It's risky, and sometimes it can still be inconsistent after making changes.

E9cipher:
I first thought of this PR when I tried #3018. I saw that everything had different names, i.e. imo it was messy. I think consistency can help a lot more understanding the codebase.

I didn't realize the settings were included in the save, so I will revert that change

d0sboots:
I agree with catlover. I won't veto this change either, but I think it's going in the wrong direction. That is, if we're going to try and use a single term in the code, I think it should be RemoteFileApi and not RFA. TLAs are frequently opaque.

E9cipher:
Should be ready by now. Anything else to tweak?

src/GameOptions/ui/RemoteFileApiConnectionStatus.tsx
```diff
@@ -14,19 +14,21 @@ import { RemoteFileApiConnectionEvents, RemoteFileApiConnectionSettingEvents } f
 import { useRerender } from "../../ui/React/hooks";
 
 export const RemoteFileApiConnectionStatus = ({ showIcon }: { showIcon: boolean }): React.ReactElement => {
-  const [rfaConnectionStatus, setRfaConnectionStatus] = useState(getRemoteFileApiConnectionStatus());
+  const [remoteFileApiConnectionStatus, setRemoteFileApiConnectionStatus] = useState(
```
d0sboots: This line here is a great example of how what you've done doesn't necessarily lead to better code.

The function here is named `RemoteFileApiConnectionStatus`. This is a component for showing the status to the user. Within it, we had the state `rfaConnectionStatus`, which holds the concrete current status. Now it is renamed to `remoteFileApiConnectionStatus`, which is almost identical to the function - it only differs by the capitalization of the first letter.

You could argue that the naming wasn't great here to begin with. However, there's no doubt that renaming it like this makes the situation worse.

src/Settings/SettingsUtils.ts
```diff
@@ -63,23 +63,23 @@ export function isValidConnectionHostname(hostname: string): Result {
...
 /**
- * Checks whether the input is a valid RFA port configuration value.
+ * Checks whether the input is a valid RemoteFileApi port configuration value.
  *
- * Port 0 is normally invalid, but it is used to disable RFA, so this function treats 0 as valid.
+ * Port 0 is normally invalid, but it is used to disable RemoteFileApi, so this function treats 0 as valid.
  */
-export function isValidRFAConnectionPortSetting(port: number): Result {
-  // 0 is not a valid port, but it is a valid configuration value that disables RFA.
+export function isValidRemoteFileApiConnectionPortSetting(port: number): Result {
```
d0sboots: This is just a really cumbersomely long name. Not just to write, but also to read. It kinda reeks of `AbstractBeanFactoryImplDelegate`.

The acronym was serving a useful function here, I think - it not only condensed, but also grouped these words together into a signal object. Before, you knew it was the *RFA* connection port. After, you could potentially read this as "Is ValidRemoteFileApi ConnectionPortSetting?" I.e. is this the ConnectionPortSetting for the ValidRemoteFileApi?

src/Settings/SettingsUtils.ts
```diff
@@ -21,7 +21,7 @@ import { Settings } from "./Settings";
  * - Use non-http schemes in the hostname: "ftp://a.com"
  * - etc.
  */
-export function isValidConnectionHostname(hostname: string): Result {
+export function isValidRemoteFileApiHostname(hostname: string): Result {
```
d0sboots: Although this is currently only used for RFA, the way it is constructed is generic. For instance, it could be used to validate `wget` hostnames in the future. I don't think this rename is correct.

src/GameOptions/ui/RemoteAPIPage.tsx
```diff
@@ -182,7 +184,7 @@ export const RemoteAPIPage = (): React.ReactElement => {
           RemoteFileApiConnectionSettingEvents.emit();
         }}
         text="Use wss"
-        tooltip={<>Use wss instead of ws when connecting to RFA clients.</>}
+        tooltip={<>Use wss instead of ws when connecting to RemoteFileApi clients.</>}
```
d0sboots: I wanted to call out this specifically as a good change - messages to users shouldn't use acronyms unless there's no other choice.

src/RemoteFileAPI/MessageDefinitions.ts
```diff
@@ -1,7 +1,7 @@
 import type { SaveData } from "../types";
 import type { BaseServer } from "../Server/BaseServer";
 
-abstract class RFAMessage {
+abstract class RemoteFileApiMessage {
```
d0sboots: Looking at the uses in this file, I don't think expanding them helps (more than it hurts). It's a narrow thing - none of these become that cumbersomely long - but the longer name isn't better, just longer. And since 100% consistency isn't doable for reasons I mention elsewhere, I don't think there's a driving reason to change these.

d0sboots (commented):
Sorry it took a while on this one. I needed to be at my compute to type up comments, and then it slipped my mind.

After looking over it carefully, I think the bulk of the changes here shouldn't be done. However, there are some that I'm neutral on, and a few that I think are definitely positive.

But the overall conclusion I have is: Consistency in this way isn't possible. We use acronyms in some places it serves our purposes. Instead, we should be asking which of the two makes the code more readable on a case-by-case basis.

src/GameOptions/ui/RemoteAPIPage.tsx
E9cipher: Yeah, this one snuck by

PR #3088, opened 2026-09-07
JEST: Add tests for save/load system

catloversg:
We will soon merge two PRs that change the save/load system: #3022 and #2889. These are risky PRs, so we need to improve our test suite.


test/jest/Save.test.ts
```diff
@@ -181,3 +210,315 @@ test("load/saveAllServers pruning RunningScripts", () => {
   const result = saveAllServers();
   expect(JSON.stringify(JSON.parse(result), null, 2)).toMatchSnapshot();
 });
+
+function initTestSaveData() {
```
d0sboots, line 216:
```suggestion
function initTestSaveData() {
// This interacts with most of the game mechanics in some way, to create a state that will exercise most of our save/load functionality.
```

test/jest/Save.test.ts
```diff
@@ -181,3 +210,315 @@ test("load/saveAllServers pruning RunningScripts", () => {
...
+  expect(InfiltrationState.lastChangeTimestamp).toBe(946684800000);
+
+  expect(Companies["ECorp"].playerReputation).toBe(1000);
+  expect(Factions["Slum Snakes"].playerReputation).toBe(1000);
+  expect(Aliases.get("foo")).toBe("bar");
+  expect(GlobalAliases.get("fooGlobal")).toBe("barGlobal");
+  expect(Settings.AutosaveInterval).toBe(1000);
+}
+
+describe("Save/Load system", () => {
```
d0sboots, line 382:
```suggestion
// This functions in a complimentary way to the snapshot tests we have above, and I'm FullSave.test. The snapshot tests check for an exact save format result, but are more rigid and less comprehensive as a result. These tests check the result after loading, which both tests the load system and allows for more flexible tests. Both are important.
describe("Save/Load system", () => {
```

d0sboots (commented):
Everything looks good, just wanted to add some explanation about the test philosophy/overlap for the future

PR #3099, opened 2026-09-12
MISC: Hacknet text

branchtheory:
* Added a missing `<br>` (been annoying me for ages)
* Reduced wordiness. It could be much shorter still, but I've erred on the side of preserving what's there. 
* Changed most singulars to plurals ("Here you can purchase a Hacknet Node" > "Here you can purchase Hacknet Nodes") for more idiomatic English

# Visuals

### Before

image

### After

image

### Before

image

### After

image

### DIFF
--- a/src/Hacknet/ui/GeneralInfo.tsx
+++ b/src/Hacknet/ui/GeneralInfo.tsx
@@ -40,7 +40,7 @@ export function GeneralInfo(props: IProps): React.ReactElement {
           <br />
           <Typography>
             Hacknet Servers can be used to run scripts, just like regular servers. However, running scripts on a Hacknet
-            Server reduces its hash rate by the percentage of its RAM being used.
+            Server reduces its hash rate by the percentage of its RAM you use.
           </Typography>
         </>
       )}
--- a/src/Hacknet/ui/GeneralInfo.tsx
+++ b/src/Hacknet/ui/GeneralInfo.tsx
@@ -14,36 +14,33 @@ export function GeneralInfo(props: IProps): React.ReactElement {
   return (
     <>
       <Typography>
-        The Hacknet is a global, decentralized network of machines. It is used by hackers all around the world to
-        anonymously share computing power and perform distributed cyberattacks without the fear of being traced.
+        The Hacknet is a global, decentralized network of machines. It is used by hackers around the world to perform
+        cyber attacks without the fear of being traced.
+        {!props.hasHacknetServers &&
+          " The machines that the Hacknet runs on are called Hacknet Nodes, specialised rigs that can share computing power anonymously. " +
+            "Hackers who use the network distribute a small percentage of their profits to the owners of Hacknet Nodes."}
       </Typography>
       <br />
       {!props.hasHacknetServers ? (
         <>
+          <Typography>You can purchase Hacknet Nodes to passively earn money.</Typography>
+          <br />
           <Typography>
-            {`Here, you can purchase a Hacknet Node, a specialized machine that can connect ` +
-              `and contribute its resources to the Hacknet network. This allows you to take ` +
-              `a small percentage of profits from hacks performed on the network. Essentially, ` +
-              `you are renting out your Node's computing power.`}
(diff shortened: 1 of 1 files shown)


src/Hacknet/ui/GeneralInfo.tsx
```diff
@@ -14,36 +14,32 @@ export function GeneralInfo(props: IProps): React.ReactElement {
...
-            {`Hacknet Servers can also be used as servers to run scripts. However, running scripts ` +
-              `on a server will reduce its hash rate (hashes generated per second). A Hacknet Server's hash ` +
-              `rate will be reduced by the percentage of RAM that is being used by that Server to run ` +
-              `scripts.`}
+            Here you can purchase Hacknet Servers, an upgraded version of Hacknet Nodes. Hacknet Servers perform
+            computations and operations on the network, earning you hashes. Hashes can be spent on a variety of
+            upgrades.
+            <br />
+            <br />
+            Hacknet Servers can be used to run scripts, just like regular servers. However, running scripts on a Hacknet
```
E9cipher, line 41:
```suggestion
            Hacknet Servers can also be used to run scripts, just like regular servers. However, running scripts on a Hacknet
```

src/Hacknet/ui/GeneralInfo.tsx
```diff
@@ -14,36 +14,32 @@ export function GeneralInfo(props: IProps): React.ReactElement {
   return (
     <>
       <Typography>
-        The Hacknet is a global, decentralized network of machines. It is used by hackers all around the world to
-        anonymously share computing power and perform distributed cyberattacks without the fear of being traced.
+        The Hacknet is a global, decentralized network of machines. It is used by hackers around the world to perform
+        cyber attacks without the fear of being traced.
```
E9cipher:
```suggestion
        cyberattacks without the fear of being traced.
```
The most common English dictionaries accept both ways. I kinda like the no-space variant, but I'd use whatever the current codebase is using.

src/Hacknet/ui/GeneralInfo.tsx
```diff
@@ -14,36 +14,32 @@ export function GeneralInfo(props: IProps): React.ReactElement {
   return (
     <>
       <Typography>
-        The Hacknet is a global, decentralized network of machines. It is used by hackers all around the world to
-        anonymously share computing power and perform distributed cyberattacks without the fear of being traced.
+        The Hacknet is a global, decentralized network of machines. It is used by hackers around the world to perform
```
E9cipher, line 17:
```suggestion
        The Hacknet is a global, decentralized network of machines. It is used by hackers all around the world to perform
```
Without the "all" the sentence sounds awkward to me. Same for you?

src/Hacknet/ui/GeneralInfo.tsx
branchtheory: Oh fair! Just Googled and you're right. Cyberattacks it is.

src/Hacknet/ui/GeneralInfo.tsx
branchtheory: Personally I think most readers will think that the paragraph break implies a change of topic without needing an 'also'.

It's just a guess though. I'm ok with either option.

src/Hacknet/ui/GeneralInfo.tsx
branchtheory: Both sound natural and idiomatic to me, so I went with the version with less words 🤷 

I'm ok with either option though.

src/Hacknet/ui/GeneralInfo.tsx
E9cipher: Re-reading makes me doubt 🤣, I'm gonna end up flipping a coin :)

src/Hacknet/ui/GeneralInfo.tsx
E9cipher: Let's see what d0sboots and CatLover think of it, right now I'm 50/50

PR #3100, opened 2026-09-12
MISC: Ellipses in UI text

branchtheory:
We do ellipses in three ways:
* `word... word`
* `word...word`
* `word ... word`

I've formalised all uses around the first option. The second option is just not right. And the third feels less Bitburner.

Regexes for finding usages:
* `[a-zA-Z0-9]\.\.\.[a-zA-Z0-9]`
* `[a-zA-Z0-9]\.\.\.\s`
* `\s\.\.\.[a-zA-Z0-9]`
* `\s\.\.\.\s`, `\.\.\.\s` and  `\s\.\.\.`

Also looked for just `...` and manually checked the list. I think I've got everything, but there are probably a few that are evading me.


catloversg:
I'll let d0sboots decide whether this style is appropriate.

Note:
- Revert the change in `Enums.ts`. Never change the values of enums and enum-like objects unless you have a very good reason. That's a breaking change. In some cases, such as objects included in save data, even changing their keys can be a breaking change.
- `Allison "Mother" Stanek: ...can ...you hear them too... ?` Don't put whitespace before "?".

PR #3102, opened 2026-09-13
DOCS: Update ns.prompt documentation

branchtheory:
<h1>NetscriptDefinitions.ts</h1><p>The current definition of prompt is confusing.</p>

<p>It's also wrong in a few places:</p>
<ol>
<li><p>It says a few times that</p>
<blockquote>The script's execution is halted until the player presses either the "Yes" or "No" button."</blockquote>
<p>or similar. But that's only true if the player awaits it. If they don't, the script continues.</p>
</li>
<li>
<p>It says it has the signature:</p>
<blockquote><code>options?: { type?: "boolean" | "text" | "select"; choices?: string[] },</code></blockquote>
<p>But <code>string[]</code> is wrong. It's effectively <code>(string | number)[]</code>. There's no typing on the actual prompt function in NetscriptFunctions, but the helpers.string used in the prompt function casts any numbers to strings. Other data types like object or array are ignored. The prompt then gets picked up by PromptManager, which expects a string array. So while there isn't any strict typechecking, strings and numbers are good, and nothing else is.</p>
</li>
<li>The docs say that things other than strings and numbers can be passed into the choices array. But if you try that, you get a type error.</li>
<li>
<p>It says that it returns</p>
<blockquote>True if the player clicks “Yes”; false if the player clicks “No”; or the value entered by the player.</blockquote>
<p>But that's only true if the options.type is "boolean". If it's "select" or "text", the default return value is `""`.</p>
<h1>NetscriptFunctions.ts</h1><p>NetscriptFunctions.ts defines validTypes, but then writes out the validTypes array again. I've refactored that.</p>


E9cipher:
I've just skimmed through the changes, but I noticed you are using `Array<string | number>` for the signature. Afaik, the current codebase is using `string[] | number[}`. Pending to confirm

branchtheory:
Hey E9, you mean the Bitburner codebase uses `string[] | number[]` as its standard way of expressing a string/number array union?

branchtheory:
Looks like `(x | y)[]` is the convention. I'll switch to that. Thanks for the heads up.

`Array<[a-z]+\s\|\s[a-z]+>` - 3 results
`\([a-z]+\s\|\s[a-z]+\)\[\]` - 20 results

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -9025,34 +9021,33 @@ export interface NS {
...
-   *   Choosing type `"select"` will require an array to be passed via the
-   *   `options.choices` property. The array can be an array of strings, an array of
-   *   numbers (not BigInt numbers), or a mixture of both numbers and strings. Any other
-   *   types of array elements will result in an error or an undefined/unexpected
-   *   behavior. The `options.choices` property will be ignored if `options.type` has a
-   *   value other than `"select"`. The script's execution is halted until the player
-   *   chooses one of the provided options and presses the "Confirm" button.
+   * If the prompt API is called again while the old dialog box still exists, the old dialog
+   * box will be replaced with a new one, and the old promise will be resolved with the
+   * default value.
```
```suggestion
   * Prompts the player with a dialog box and returns a promise.
   *
   * The prompt and the return value depend on the `options.type` and the player's action.
   *
   * If the `options.type` is `"boolean"`, not provided, or `undefined`:
   * - Prompt options: Yes/No
   * - Return value: `true` if the player selects Yes, or `false` if they click No or cancel the prompt.
   *
   * If the `options.type` is `"select"`:
   * - Prompt options: Dropdown list of options
   * - Return value: The value selected by the player, or `""` if they cancel the prompt.
   *
   * If the `options.type` is `"text"`:
   * - Prompt options: Free-form text field
   * - Return value: The value entered by the player, or `""` if they cancel the prompt.
   *
   * The `options.choices` property is an optional array whose elements can be strings or numbers (excluding BigInt). It
   * is only needed for the "select" prompt. It's ignored otherwise.
   *
   * Note that when the player selects an option, the selected value will always be converted to a string. For example,
   * if the choice array is `[1, "2"]` and the player chooses `1` (the number value), the promise resolves to `"1"` (the
   * string value).
   *
   * The player can cancel the prompt by clicking the X button in the top-right or clicking outside the dialog box.
   *
   * If the prompt API is called again while the old dialog box still exists, the old dialog box will be replaced with a
   * new one, and the old promise will be resolved with the default value.
```

src/ScriptEditor/NetscriptDefinitions.d.ts
```diff
@@ -9081,11 +9076,15 @@ export interface NS {
    *
    * @param txt - Text to appear in the prompt dialog box.
    * @param options - Options to modify the prompt the player is shown.
-   * @returns True if the player clicks “Yes”; false if the player clicks “No”; or the value entered by the player.
+   * @returns Return value depends on the player action and `options.type`. If `options.type`
+   * is "boolean" or `undefined`, the return value is `true` if the player clicks Yes, and
+   * `false` if they click "No" or cancels the dialogue box. If the `options.type` is "text"
+   * or "select", the return value is the one selected or entered by the player, or `""` if
+   * they cancel the dialogue box.
```
```suggestion
   * @returns Return value depends on the player action and `options.type`. If `options.type` is "boolean" or
   * `undefined`, the return value is `true` if the player clicks Yes, and `false` if they click "No" or cancel the
   * prompt. If the `options.type` is "text" or "select", the return value is the one selected or entered by the player,
   * or `""` if they cancel the prompt.
```

src/ScriptEditor/NetscriptDefinitions.d.ts
branchtheory: Thanks CatLover. Basically implemented this but with slight revisions, like flipping this sentence round ...

> The `options.choices` property is an optional array whose elements can be strings or numbers (excluding BigInt). It is only needed for the "select" prompt. It's ignored otherwise.

... so you find out whether this applies to your use case first, and only then then find out how it works.

> The `options.choices` property is an optional array, only needed for the "select" prompt and ignored otherwise. Its elements can be strings or numbers (excluding BigInt).

PR #3103, opened 2026-09-14
MISC: Adding a heading option to ns.prompt

branchtheory:
Many modals should really have headings. This adds that option.

image

This doesn't impact any current uses of PromptManager. I checked the following and they're both working fine:
* The prompt you get when quitting a script without saving it
* The prompt you get when joining the Bladeburner division if you have Stanek and Source File 7.3


E9cipher:
I don't think this is really necessary. Prompts are a very basic way to ask the player for a thing. It is used primarily in bitburner's way of telling something with a nice view (and to replace `stdin` at runtime). They may also be used by new players wanting a simple choice maker (e.g. my backdoor script prompts a yes/no selection based on the time it takes to do its job).

Imo, if you want customized prompts, build a jsx script to do it, which can have a lot of additional cool features (css of your own, structure, callback functions...)

catloversg:
I don't think this is the proper way to let players customize the prompt. If you add a header, people will ask for a footer, then they ask for ways to customize the header and footer, then they ask for more options. We either make this API minimal (i.e., provide very basic but still useful options, but no more) or let the player customize *everything* by passing a `ReactElement` as `txt` (`txt: string | ReactElement`). This is also why we have `printRaw` and `tprintRaw` instead of adding more and more options to `print` and `tprint`.

Personally, I prefer to keep `prompt` minimal.

branchtheory:
OK, makes sense to me. The real value from this would be that PromptManager is used by the game to display prompts too. Some of those could likely do with headings. I'll wait till I have a use case though before expanding the prompt interface though.

hexagonrecursion:
> The real value from this would be that PromptManager is used by the game to display prompts too

Adding the ability to display headings to PromptManager for our internal purposes may be a good idea, but I think we should not expose this capability to `ns.prompt()` for reasons explained by catlover

PR #3104, opened 2026-09-14
DOCS: Code blocks now read the language (from ```lang)

branchtheory:
Screenshot of JS, angularjs, and text: 

image


catloversg:
I'll handle this in #3106. That PR has more info and considerations for discussion. It also contains the proper fix.

PR #3045, opened 2026-08-11
UI: Limit max width of Documentation page

branchtheory:
### The motivation for the change
There's no restriction on the width of the docs pane in the Documentation tab. Long lines of text are hard to read though. I checked a few websites for documentation and they had an approximate character count per line of between 95 (Microsoft) and 115 (Stripe) characters. I've gone for 95. That's closer to what readability research suggests is optimal, while still making things relatively easy to skim.

### What this changes
(1) The pane width is now 920px. Running the game on the default settings, this gives you a maximum of 95 characters per line for regular text.

As a result of that change, three of the articles have code blocks where the code overflows the right-hand side of the block container. You can scroll horizontally, but it's not a great experience. 

To handle this ... 

(2) I changed some comments to span multiple lines. All text in code blocks will be visible for all typefaces for font sizes of 18px or less.

Let me know if we need more of a safety margin to handle larger fonts sizes still.

PR #3106, opened 2026-09-14
BUGFIX: Markdown renderer ignores specified code block language

catloversg:
When using react-syntax-highlighter, we always pass "javascript" to the language property, so all code blocks are rendered as js code, even when they contain plain text or code in other languages. This causes the issue mentioned in #3045.

Considerations for future PRs:

[1]: hljs highlighter (the default import) does not support jsx and tsx. When rendering jsx and tsx, it still performs some syntax highlighting, but it may produce incorrect highlighting. Currently, I don't see any obvious problems, so I don't change anything here. We may need to consider use the Prism highlighter (react-syntax-highlighter supports it). If we do that, we also need to choose a different "theme" (monokaiSublime is an hljs "theme").

[2]: ~The default language I used here is js, but I may switch to plain text in a future PR.~ Currently, many pages contain code blocks like these (notice that we don't use triple backticks):
```
    $ home
    $ nano early-hack-template.js
```
```
    if (ns.fileExists("BruteSSH.exe", "home")) {
        ns.brutessh(target);
    }

    ns.nuke(target);
```
Everything is rendered as js code even when it's not. ~To minimize the impact of this PR, I'll keep this behavior for now. I'll open another PR to make all these code blocks use triple backticks and set the correct language value.~

Edit: I opened #3107, so I set the default language to text instead of js.


branchtheory:
> `style={monokaiSublime}`

Does this mean that JS / TS will be highlighted according to the custom theme that players can set in the Script Editor options?

catloversg:
No. That's a predefined theme in hljs.

d0sboots:
Everything looks good here, but it's waiting on #3107. Ping me if I forget.

PR #3107, opened 2026-09-15
DOCUMENTATION: Consistently use triple ticks when writing code blocks

catloversg:
Context: #3106.


src/Documentation/doc/en/basic/autocomplete.md
```diff
@@ -94,7 +94,7 @@ In that example typing `run script.js` and pressing tab would initially suggest
 
 This is a function that works nearly identically to `ns.flags()`. The only difference is that it allows unknown options. For example:
 
-```js
+```javascript
```
d0sboots: "js" is the more natural way to write this, and I believe it's accepted by GH and other markdown processors. I don't think these should be changed, although it's fine to work it out in full (or short form)where it was unspecified before.

PR #3108, opened 2026-09-15
GH: Update bug report template

catloversg:
Using `render: markdown` makes text in this section be surrounded by a ```` ```markdown ```` block. For example, if the reporter writes `foo`, the description in the created issue will contain:
````
```markdown
foo
```
````

Tested manually.


E9cipher:
Yepp I was just gonna issue that!

PR #3112, opened 2026-09-17
UI: Active Scripts pagination controls made more visible

branchtheory:
Inspired by [suggestions on Discord](https://discord.com/channels/415207508303544321/459097632896188436/1550245989573730315) that if the pagination was more visible, there would be less confusion.

I've tripped up on this a couple of times myself, wondering why in the world 'my scripts aren't executing properly' when actually I'm only showing 10 servers.

In this video the search bar and pagination controls are switched. I've actually switched them back -- they're now in their original places.

https://github.com/user-attachments/assets/841a0038-b4ea-4448-b4c3-ba99e9f37fe8


catloversg:
Don't move the search box to the right border like that. If you don't realize the pagination tool exists, it's likely because you don't look at the right side of the UI. Moving the search box to the right basically means you are swapping the locations of that tool and the pagination tool. Now the issue is the same; it's just changing the "target". Instead of missing the pagination tool, players miss the search tool.

Just make everything render from left to right.

654274240-236b00f0-1221-4a19-b8c8-a460f1c25a26

catloversg:
This is what it looks like in the latest commit. Is this finished?

Capture

branchtheory:
I think so. I tried moving the search bar to the end of the row and to me it ends up looking cluttered.

If the search bar isn't on the far-right, and isn't at the end of the row, the only place left is where it is now. 

Personally I've never used the search bar. If I were designing the UI only for myself, I'd delete it (and use Ctrl+F if for some reason I ever needed to find stuff). But I don't know how representative I am. My guess is that it's not used much, but 🤷

image

catloversg:
The search bar is useful. Removing it is definitely out of the question.

branchtheory:
Cool, works for me

d0sboots:
Here's what my active scripts looks like rn:
image
I agree that adding controls to the bottom is an improvement, but leaving them aligned to the right like this seems like a missed opportunity. Your mock-up based on catlover's suggestion, with the search bar on the right (but not right-aligned) feels a lot better to me?

branchtheory:
Huh, with a weekend's space it looks a lot better. Alright, agree with you both. Committed this version:

image

PR #3113, opened 2026-09-19
MISC: Be less spoilery

E9cipher:
I noticed that the error messages for API access (singularity, bladeburner...) do not account for early-game players, which can spoil a few endgame things (e.g. SFs). I have added `knowAboutBitverse()` as a check to change the text if the player has not yet completed BN1.1, which displays text differently.


src/NetscriptFunctions/Bladeburner.ts
```diff
@@ -34,7 +34,11 @@ export function NetscriptBladeburner(): InternalAPI<INetscriptBladeburner> {
   const getBladeburner = function (ctx: NetscriptContext): Bladeburner {
     const apiAccess = canAccessBitNodeFeature(7) || canAccessBitNodeFeature(6);
     if (!apiAccess) {
-      throw helpers.errorMessage(ctx, "You have not unlocked the Bladeburner API.", "API ACCESS");
+      throw helpers.errorMessage(
+        ctx,
+        `You have not unlocked ${knowAboutBitverse() ? "the Bladeburner API" : "this special feature"}.`,
```
branchtheory: The current response only tells the player what they already know: that there's a Bladeburner API.

catloversg:
This is a questionable change, and I'm fairly against it. It's not a bad thing to do, but it's not useful in many cases. Avoiding spoilers is a gray area in which sometimes it's not worth the effort to do so.

The only real "spoiler" in Bitburner is the existence of the BitVerse concept (the reality is just a simulation, and you can jump between different simulations). To avoid spoiling that concept, we also avoid mentioning things such as TRP or WD that directly lead to the BitVerse. Technically, there are some other things that can be considered "spoilers", and people have suggested finding ways to hide them. However, I vetoed some of those suggestions. One of the most common suggestions is the faction enum. Daedalus sells TRP, so some people argue that `ns.enums.FactionName.Daedalus` is a spoiler. However, is it really a spoiler? IMO, it's not. Knowing a faction called "Daedalus" is exactly the same as knowing CSEC exists; it's just a name. That value in the enum does not say anything special about Daedalus, so it does not spoil anything.

When deciding whether we need to hide a term, we need to consider:
- How likely is the player to see that term?
- How likely is that term to reveal the actual thing we want to avoid spoiling (i.e., the BitVerse concept)?

In #2952, I also hid BN/SF terms, but I did that because the stock market is available pre-endgame, so the player can accidentally see these terms. All endgame-exclusive mechanics are in their own namespace, so it's less likely that players will accidentally see these terms.

Most, if not all, error messages in endgame-exclusive mechanics do not need to care about spoilers for two reasons. I explained the first one above. The second one is that seeing BN/SF terms does not immediately reveal the BitVerse concept. If the player accidentally uses these APIs (e.g., searches for `installBackdoor` and uses the Singularity API for that), they will see those terms, but what do those terms mean to them? I think it's nearly impossible to figure out Bitburner's plot just from seeing the "Source-File" phrase. Some players who are familiar with the Matrix series and similar stories may make a guess when they see "BitNode" and a number following it, but that's still just a guess.

In short, I don't think we need to search all error messages in endgame-exclusive mechanics for endgame terms and add a check just to hide them.

Note that I'm talking about hiding terms such as BN/SF in endgame-exclusive mechanics. With pre-endgame mechanics such as the stock market, hiding endgame terms is fine.

By the way, in the current commit, the change in Singularity is questionable as explained above. The change in Bladeburner is definitely unnecessary. It does not make sense to hide the "Bladeburner" term. In order to see this error message, the player needs to interact with Bladeburner APIs via `ns.bladeburner`. Why bother hiding the "Bladeburner" term if they already type `bladeburner` themselves (i.e., they already know something is called "bladeburner")?

E9cipher:
> How likely is the player to see that term?

In discord: a lot. Imagine person A, an average player in early-game, who has seen the term `singularity` somewhere (e.g. discord). They search the docs and see some corruptible text that seems to be singularity. They try to write `ns.singularity.`  and suddenly the editor shows a lot of props/functions. They see some (e.g. `destroyW0r1dD43m0n`) and try one of them. Then this popup appears. If we say "source-file", I at least would search what this term means. If we say that's an endgame feature, Player A knows that if they keep searching, they will spoil themselves, so we give them the choice.

Yes, if player A sees `singularity` in discord, the most obvious thing to do is ask "what is singularity". Although, some people might not do that (they might wanna test it for themselves), or might have heard the term in other places.

> How likely is that term to reveal the actual thing we want to avoid spoiling (i.e., the BitVerse concept)?

On the previous example, I'd say the chance is high if they keep searching.

Bladeburner: the player might know something is called bladeburner, and know it's a spoiler. I agree that it's pointless to hide the term bladeburner, but I would still like to mention it is a spoiler (just to be sure).

catloversg:
> I'd say the chance is high if they keep searching.

If they search what singularity means, it's likely they will only know it's an endgame feature that allows automation. Technically, it's a spoiler, but it's so tiny that it's not a problem. If they keep searching until they know the entire plot of Bitburner, to be honest, that's on them. The current message is fairly good at providing information (the real condition) while not revealing too much (the plot).

This is the same with other terms. By default, the game hides the endgame documentation by using corruptible text. This is not a security feature. It's a hint for the player that these are not things they are supposed to know yet. The player is still free to search for them, but they need to be careful. We already warn them (both with the corruptible text and the vague error message).

Regarding the bladeburner message, I'll be frank with you, that's a pointless change.

d0sboots:
I'm vetoing this.

```
This singularity function requires Source-File 4 to run. A power up you obtain later in the game. It will be very obvious when and how you can obtain it.
```
This is how the message currently reads, and it is basically the gold standard for how to *not* spoil things in messaging. If you get this message before reaching endgame, what is "spoiled?"

* That you aren't in endgame yet. But this should have been obvious.
* That there will be additional content later. Again, this should have been obvious, or at least extremely probable.
* That something called a "Source-File" exists, and that they are numbered in some way. Without further context, this tells you basically nothing. It doesn't even directly mention "bitnode."

> I agree that it's pointless to hide the term bladeburner, but I would still like to mention it is a spoiler (just to be sure).

Sometimes, things that are not spoilers *become* spoilers if you draw too much attention to them.

Our rules for spoilers exist on the Discord for a good reason - because that is where *other people* are most likely to spoil you. If you are just playing the game by yourself, you will not spoil yourself *unintentionally*. (In particular, if you start diving around in the source code, you should expect to get spoilers, especially if you start researching mechanics that you don't have yet.)

branchtheory:
Had a wider look. The sleeve one is the only one I noticed that has spoilers. 

> sleeve.getSleeve: You do not have access to the Sleeve API. This is either because you are not in **BitNode-10** or because you do not have Source-File 10.

Should really be something like:

> sleeve.getSleeve: The Sleeve API is not yet available. It requires Source-File 10 to use.

Not really sure how explain the "need to be in BN-10 _or_ have SF 10" nuance without actually saying "in a BN". Maybe best to just leave the whole thing out:

> sleeve.getSleeve: You have not yet unlocked the Sleeve API.

catloversg:
The messages that mention BN, such as the Sleeve's and other mechanics' messages, are correct. The Singularity message is technically incorrect because you can use Singularity APIs in BN4 without SF4. "Fixing" it is debatable because mentioning BN is technically a spoiler.

I think it's fine to keep everything as-is. With Singularity, it's a bit incorrect, but it strikes a good balance between technical correctness and not revealing too much information. With other mechanics, the messages are correct; mentioning BN is not ideal, but I don't see it as a problem for the reasons I explained previously.

